PR #2, opened 2018-11-21
fix failing overlayfs unmounting + minor refactoring

Rezney:


artmello:
LGTM

radomil:
Checked, LGTM, as well.

pirat89:
leapp-ci build

pirat89:
Created the follow up issue, that will finish this work. This is partial fix, but still valuable.

PR #4, opened 2018-11-26
Add description of phases to the workflow

bocekm:
- having the description of the phases right in the workflow definition file makes it easier for the actor developer to choose the right one without the need to go to the documentation.
- right now it's just a copy&paste from https://leapp-to.github.io/upgrades and needs to be updated to be more clear what each phase is about

### DIFF
--- a/repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
+++ b/repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
@@ -8,20 +8,32 @@ from leapp.tags import IPUWorkflowTag, FactsPhaseTag, ChecksPhaseTag, \
     LateTestsPhaseTag, PreparationPhaseTag, RPMUpgradePhaseTag, ApplicationsPhaseTag, ThirdPartyApplicationsPhaseTag, \
     FinalizationPhaseTag, FirstBootPhaseTag, ReportPhaseTag
 
+
 class IPUWorkflow(Workflow):
     name = 'InplaceUpgrade'
     tag = IPUWorkflowTag
     short_name = 'ipu'
-    description = '''No description has been provided for the InplaceUpgrade workflow.'''
+    description = '''The IPU workflow takes care of an in-place upgrade (IPU) of RHEL 7 to RHEL 8.'''
 
     class FactsCollectionPhase(Phase):
-        name = 'Facts collection'
+        '''Get information (facts) about the system (e.g. installed packages, configuration, ...).
+
+        No decision should be done in this phase. Scan the system to get information you need and provide
+        it to other actors in the following phases.
+        '''
+        name = 'FactsCollection'
         filter = TagFilter(FactsPhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
                             Policies.Retry.Phase)
         flags = Flags()
 
     class ChecksPhase(Phase):
+        '''Check upgradability of the system, produce user question if needed and produce output for the report.
+
+        Check whether it is possible to upgrade the system and detect potential risks. It is not expected to get
+        additional information about the system in this phase, but rather work with data provided by the actors from
+        the FactsCollection. When a potential risk is detected for upgrade, produce messages for the Reports phase.
+        '''
         name = 'Checks'
         filter = TagFilter(ChecksPhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
@@ -29,11 +41,14 @@ class IPUWorkflow(Workflow):
         flags = Flags()
 
     class ReportsPhase(Phase):
+        '''Provide user with the result of the checks.'''
         name = 'Reports'
         filter = TagFilter(ReportPhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
                             Policies.Retry.Phase)
         flags = Flags()
+
+    # The following two phases have been removed but are kept here for reference or in case they are needed again
 
     #class AttachPackageReposPhase(Phase):
     #    name = 'AttachPackageRepos'
@@ -52,6 +67,7 @@ class IPUWorkflow(Workflow):
     #    flags = Flags()
 
     class DownloadPhase(Phase):
+        '''Download data needed for the upgrade and prepare RPM transaction for the upgrade.'''
         name = 'Download'
         filter = TagFilter(DownloadPhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
@@ -59,6 +75,7 @@ class IPUWorkflow(Workflow):
         flags = Flags()
 
     class InterimPreparationPhase(Phase):
+        '''Prepare an initial RAM file system (if required). Setup bootloader.'''
         name = 'InterimPreparation'
         filter = TagFilter(InterimPreparationPhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
@@ -66,11 +83,14 @@ class IPUWorkflow(Workflow):
         flags = Flags(request_restart_after_phase=True)
 
     class InitRamStartPhase(Phase):
+        '''Boot into the upgrade initramfs, mount disks, etc.'''
         name = 'InitRamStart'
         filter = TagFilter(InitRamStartPhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
                             Policies.Retry.Phase)
         flags = Flags()
+
+    # The following two phases have been removed but are kept here for reference or in case they are needed again
 
     #class NetworkPhase(Phase):
     #    name = 'Network'
@@ -89,6 +109,7 @@ class IPUWorkflow(Workflow):
     #    flags = Flags()
 
     class LateTestsPhase(Phase):
+        '''Last tests before the RPM upgrade that have to be done with the new kernel and systemd.'''
         name = 'LateTests'
         filter = TagFilter(LateTestsPhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
```diff
@@ -16,118 +16,158 @@ class IPUWorkflow(Workflow):
...
                             Policies.Retry.Phase)
         flags = Flags()
 
     class ChecksPhase(Phase):
+        '''Inhibitors - Blockers, Shared Objects, Binaries, Languages,
+        Packages, Load Balance Support, User Management, Version Control
+        Systems, FreeRADIUS, IPA Bind Dyndb LDAP, SSSD, NTP, OpenLDAP, YPSERV,
+        Network, Databases, SELinux, Storage, System, HTTPD, OpenSSH, Postfix,
+        SQUID, DoveCot, Bind9 Configuration, RSYSLOG.
+        '''
```
pirat89: The description is vague. This is not only about inhibitors. Inhibitors are small part of actions that should happened here. Here should be put all user questions, should be checked what should be done and it should be planned. basically, it's subset of actions we usually do in PA during the check of the system. Just no actions should be done here.

repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
```diff
@@ -16,118 +16,158 @@ class IPUWorkflow(Workflow):
     description = '''No description has been provided for the InplaceUpgrade workflow.'''
 
     class FactsCollectionPhase(Phase):
+        '''Operating System Version Detection, IP, Hostname, RPMs, Services,
+        Augeas.
+        '''
```
pirat89: Get info about the system - what is installed, settings, ... IOW, scan the system to get facts about it.

repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
```diff
@@ -16,118 +16,158 @@ class IPUWorkflow(Workflow):
...
     class AttachPackageReposPhase(Phase):
+        '''Make sure the repositories of the new system are available.'''
         name = 'AttachPackageRepos'
         filter = TagFilter(AttachPackageReposPhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
                             Policies.Retry.Phase)
         flags = Flags()
 
     class PlanningPhase(Phase):
+        '''RPM Install transaction (feasibility checks).'''
```
pirat89: This phase looks completely useless from the point of the current design.

repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
```diff
@@ -16,118 +16,158 @@ class IPUWorkflow(Workflow):
...
         name = 'Planning'
         filter = TagFilter(PlanningPhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
                             Policies.Retry.Phase)
         flags = Flags()
 
     class DownloadPhase(Phase):
+        '''Download of all things needed (RPMs, Kernel images, Anaconda image
+        (very big maybe), etc.)
+        '''
```
pirat89: here we do everything connected to the transaction - get target rpm repositories, check the rpm transaction, download all packages... 
From that point, we should rename the phase in future to fit better to what we do here.

repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
```diff
@@ -16,118 +16,158 @@ class IPUWorkflow(Workflow):
...
         name = 'Download'
         filter = TagFilter(DownloadPhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
                             Policies.Retry.Phase)
         flags = Flags()
 
     class InterimPreparationPhase(Phase):
+        '''Preparation of initial ramdisk (if required).
+        Setup bootloaderGrub.
+        '''
```
pirat89: What about:,,Preparation of system to be able to proceed the upgrade after reboot. E.g. prepare initial ramdisk, setup bootloader, ...).

repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
```diff
@@ -16,118 +16,158 @@ class IPUWorkflow(Workflow):
...
         name = 'InterimPreparation'
         filter = TagFilter(InterimPreparationPhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
                             Policies.Retry.Phase)
         flags = Flags(request_restart_after_phase=True)
 
     class InitRamStartPhase(Phase):
+        '''Reboot into the next system therefore not a real step, however the
+        before and post parts are interesting to have for hooking.
+        '''
```
pirat89: From the desciprtion, really not sure what about it is. Anyway, this phase is about the boot - not reboot, that's action connected to previous phase only. You are booting here.

repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
```diff
@@ -16,118 +16,158 @@ class IPUWorkflow(Workflow):
...
     class StoragePhase(Phase):
+        '''Mount all storage points.'''
         name = 'Storage'
         filter = TagFilter(StoragePhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
                             Policies.Retry.Phase)
         flags = Flags()
 
     class LateTestsPhase(Phase):
+        '''Tests that need to be executed later.'''
```
pirat89: What about: last tests before the upgrade itself, that has to be done with the new kernel.

repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
```diff
@@ -16,118 +16,158 @@ class IPUWorkflow(Workflow):
...
         name = 'LateTests'
         filter = TagFilter(LateTestsPhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
                             Policies.Retry.Phase)
         flags = Flags()
 
     class PreparationPhase(Phase):
+        '''YUM Configurations. Backup not upgradable configurations (No Verify
+        Configs).
+        '''
```
pirat89: No Verify Configs != not upgradable configuration

repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
```diff
@@ -16,118 +16,158 @@ class IPUWorkflow(Workflow):
...
         name = 'RPMUpgrade'
         filter = TagFilter(RPMUpgradePhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
                             Policies.Retry.Phase)
         flags = Flags()
 
     class ApplicationsPhase(Phase):
+        '''Application of changes on configuration files, move various
+        directores, etc.
+        '''
```
pirat89: Provide all neccessary steps to finish upgrade of particular native applications. Including move/rename of configuration files (if required), modify configuration of applications to be able to run correctly (if possible, to be set up as close as possible to original behaviour), ....

repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
```diff
@@ -16,118 +16,158 @@ class IPUWorkflow(Workflow):
...
+        upgraded.
+        '''
         name = 'ThirdPartyApplications'
         filter = TagFilter(ThirdPartyApplicationsPhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
                             Policies.Retry.Phase)
         flags = Flags()
 
     class FinalizationPhase(Phase):
+        '''Plan SELinux relabeling. Labels Bootloader settings.'''
```
pirat89: Actually, we probably will do SELinux relabeling during this phase in future. What about: "Rest of actions that should be done before the reboot into the ugpraded system. E.g. selinux relabeling, ...."

repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
```diff
@@ -16,118 +16,158 @@ class IPUWorkflow(Workflow):
...
+        before and post parts are interesting to have for hooking.
+        '''
         name = 'InitRamStart'
         filter = TagFilter(InitRamStartPhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
                             Policies.Retry.Phase)
         flags = Flags()
 
     class NetworkPhase(Phase):
+        '''Bring up the network.'''
```
vinzenz: Might be unnecessary here, it's possible that we will have this done already before we start the workflow

repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
```diff
@@ -16,118 +16,158 @@ class IPUWorkflow(Workflow):
...
         name = 'Network'
         filter = TagFilter(NetworkPhaseTag)
         policies = Policies(Policies.Errors.FailPhase,
                             Policies.Retry.Phase)
         flags = Flags()
 
     class StoragePhase(Phase):
+        '''Mount all storage points.'''
```
vinzenz: Might be unnecessary here, it's possible that we will have this done already before we start the workflow

repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
vinzenz: and new systemd

repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
vinzenz: I think that this should be called PreTransaction and should consider to apply workarounds etc that are needed to make the transaction successful for example.

pirat89:
@vinzenz I keep rename of `PreparationPhase` for another PR or separate commit. For now, just kept as it is for review.

repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
pirat89: commented out already.

bocekm:
Squashed.

PR #5, opened 2018-11-28
Add check for available space on /boot

bocekm:
Part of that is removing the leapp-provided kernel and initramfs before the RPM upgrade phase in order to minimize the /boot free space needed.


bocekm:
The implementation has been done for some time and it's ready for review, I'm missing tests though.

repos/system_upgrade/el7toel8/actors/addupgradebootentry/tests/unit_test.py
```diff
@@ -0,0 +1,23 @@
...
+    monkeypatch.setenv('LEAPP_DEBUG', '1')
+    monkeypatch.setattr(stdlib, 'call', call_mocked())
+
+    def get_boot_file_paths_mocked():
+        return '/abc', '/def'
+    monkeypatch.setattr(library, 'get_boot_file_paths', get_boot_file_paths_mocked())
+
+    library.add_boot_entry()
+
+    assert library.get_boot_file_paths.args.join(' ') == ('/usr/sbin/grubby --add-kernel=/abc --initrd=/def'
```
vinzenz: wouldn't that be rather  `'  '.join(library.get_boot_file_paths.args) == ...`

bocekm: Indeed. This was like "pushing whatever I had before calling it a day" without really running the tests locally. I guess the thing is that I switch between the branches all the time causing conflicts when I have some stuff drafted and git stash is just a partial solution. But I'll try to run the tests before next pushes.

repos/system_upgrade/el7toel8/actors/removeupgradebootentry/libraries/library.py
```diff
@@ -0,0 +1,26 @@
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.stdlib import api, call
+from leapp.models import BootContent
+
+
+def remove_boot_entry():
+    kernel_filepath = get_upgrade_kernel_filepath()
+    call([
```
vinzenz: I guess we should split this off into its own actor. Because this is failing quite often when there's something unmountable in FSTAB e.g. an NFS share. What you think? (Anyway this is not to block this PR but about kicking of the thinking) 
Mainly this is irrelevant to the task the actor does anyway, so it's misleading IMHO.

repos/system_upgrade/el7toel8/actors/copyinitramfstoboot/libraries/library.py
```diff
@@ -0,0 +1,51 @@
...
+    return path.join('/boot', filename)
+
+
+def copy_files(files_to_copy):
+    for filename in files_to_copy.keys():
+        try:
+            copyfile(files_to_copy[filename]['src_path'], files_to_copy[filename]['dst_path'])
+        except IOError as err:
+            raise StopActorExecutionError('Could not copy {0} to /boot'.format(files_to_copy[filename]['src_path']),
+                                          details={'Details': str(err)})
```
vinzenz: here the 'details' make more sense. And I would use a lower case D for the key.

repos/system_upgrade/el7toel8/actors/checkbootavailspace/libraries/library.py
```diff
@@ -0,0 +1,31 @@
...
+    return boot_stat.f_frsize * boot_stat.f_bavail
+
+
+def is_additional_space_required(avail_bytes):
+    return avail_bytes < MIN_AVAIL_BYTES_FOR_BOOT
+
+
+def inhibit_upgrade(avail_bytes):
+    additional_mib_needed = (MIN_AVAIL_BYTES_FOR_BOOT - avail_bytes) / 2**20
+    api.produce(CheckResult(
```
vinzenz: Use now the Inhibitor type

repos/system_upgrade/el7toel8/actors/removeupgradebootentry/libraries/library.py
bocekm: I can see you wrote the original actor. Is the `mount -a` necessary before calling _grubby_ anyway?

repos/system_upgrade/el7toel8/actors/checkbootavailspace/libraries/library.py
bocekm: Done.

repos/system_upgrade/el7toel8/actors/copyinitramfstoboot/libraries/library.py
bocekm: Done.

repos/system_upgrade/el7toel8/actors/copyinitramfstoboot/libraries/library.py
```diff
@@ -0,0 +1,51 @@
...
+        files_to_copy[filename] = {'src_path': get_src_filepath(filename), 'dst_path': get_dst_filepath(filename)}
+    return files_to_copy
+
+
+def get_src_filepath(filename):
+    src_filepath = api.get_file_path(filename)
+    if src_filepath is None:
+        raise StopActorExecutionError('Could not find {0} in the following paths: {2}'
+                                      .format(filename, ' '.join(api.files_paths())),
+                                      details={'Details': 'You may want to try to reinstall'
```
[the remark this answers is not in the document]
bocekm: Haha, yeah. I used `Details` key everywhere where the dict for the `details` was needed just to satisfy the requirement that it's supposed to be dict, not string. But when you mention it, then yeah, it's more of a hint. Using `hint` key.

repos/system_upgrade/el7toel8/actors/removebootfiles/tests/unit_test.py
```diff
@@ -0,0 +1,39 @@
...
+    assert library.remove_file.files_to_remove == ['/abc', '/def']
+
+    # No BootContent message available
+    def consume_no_message_mocked(*models):
+        yield None
+    monkeypatch.setattr('leapp.libraries.stdlib.api.consume', consume_no_message_mocked)
+    monkeypatch.setattr(library, 'remove_file', remove_file_mocked())
+
+    with pytest.raises(StopActorExecution):
+        library.remove_boot_files()
```
bocekm: The test is failing on this line. I need to monkeypatch `api.current_logger()` because it returns _NoneType_.

repos/system_upgrade/el7toel8/actors/removeupgradebootentry/libraries/library.py
bocekm: @vinzenz ping :)

vinzenz: @bocekm mount -a is what should IMHO happen first thing, but maybe reversed order is better in case mount fails, there wouldn't be a reboot loop :-)

PR #7, opened 2018-11-29
Timestamp fix2

pirat89:
This solution is more simple and optimal. W8 for builds before merge. Thanks @vojtechsokol for investigation.


vojtechsokol:
Closing this as it attemps to solve the same thing as #8 and it is not working.

PR #8, opened 2018-11-29
Timestamp fix

vojtechsokol:
Simpler fix for persistent TIMESTAMP (#6)


pirat89:
I see now where I did mistake. Anyway, fix please even lines with `$(MAKE)`.
```
... __TIMESTAMP=$(TIMESTAMP) $(MAKE) ...`
```

pirat89:
We found this is not working as every target is run as separate shell and the substitution is evaluated separately. so this is not working solution.

PR #11, opened 2018-11-30
Final persistant timestamp solution

pirat89:
Previous solution was incorrect. This is more simple and working solution. However, sometimes issues with print_release target could appear. For this purposes you need to run `make prepare` manually  first.


pirat89:
Closing as it is resolved by the PR #12

PR #12, opened 2018-11-30
Final™ Timestamp fix

vojtechsokol:
Simpler (and working) fix for persistent TIMESTAMP (#6)


pirat89:
Nope. This solution really is not working. Try this:
```patch
diff --git a/Makefile b/Makefile
index 8dd14af..1422d25 100644
--- a/Makefile
+++ b/Makefile
@@ -68,10 +68,13 @@ clean:
        @find . -name '__pycache__' -exec rm -fr {} +
        @find . -name '*.pyc' -exec rm -f {} +
        @find . -name '*.pyo' -exec rm -f {} +
+       echo $(TIMESTAMP)
+       sleep 60
 
 prepare: clean
        @echo "--- Prepare build directories ---"
        @mkdir -p packaging/{sources,SRPMS}/
+       echo $(TIMESTAMP)
 
 list_builds:
        @copr --config $(_COPR_CONFIG) get-package $(_COPR_REPO) \

```
and call `make prepare` to see the behaviour.

vojtechsokol:
@pirat89 Of course it was not working, I pushed the original version instead of the fixed one. Now it works.

PR #13, opened 2018-12-05
Use the call function from the leapp.libraries.stdlib

radomil:
Can be merged after the https://github.com/oamg/leapp/pull/388 is merged into leapp.
Several actors refactored to reference the new _leapp.libraries.stdlib_ `call` function.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/firewalldisable/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/firewalldisable/actor.py
@@ -1,22 +1,7 @@
 from leapp.actors import Actor
 from leapp.models import FirewallDecisionM, CheckResult, SystemFacts
 from leapp.tags import IPUWorkflowTag, ApplicationsPhaseTag
-import six
-import subprocess
-import os
-
-
-def call(args, split=True):
-    ''' Call external processes with some additional sugar '''
-    r = None
-    with open(os.devnull, mode='w') as err:
-        if six.PY3:
-            r = subprocess.check_output(args, stderr=err, encoding='utf-8')
-        else:
-            r = subprocess.check_output(args, stderr=err).decode('utf-8')
-    if split:
-        return r.splitlines()
-    return r
+from leapp.libraries.stdlib import call
 
 
 class FirewallDisable(Actor):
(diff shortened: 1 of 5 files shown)


repos/system_upgrade/el7toel8/actors/firewalldisable/actor.py
```diff
@@ -29,35 +14,35 @@ class FirewallDisable(Actor):
 
     def stop_firewalld(self):
         ''' Stops FirewallD '''
-        call(['systemctl', 'stop', 'firewalld'])
+        check_output(['systemctl', 'stop', 'firewalld'])
```
bocekm: I think that wherever you don't check the output, it doesn't makes sense to call `check_output`. Use `subprocess.call` instead.
It's 5 times in this file and one time in prepareupgradetransaction/actor.py.

pirat89: What you wrote is reason to use `check_output`, right? :) From help:
>  If the return code was non-zero it raises a CalledProcessError

pirat89:
leapp-ci build

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
```diff
@@ -7,7 +7,7 @@
 from leapp.tags import IPUWorkflowTag, DownloadPhaseTag
 
 from subprocess import CalledProcessError, call
-
+from leapp.libraries.stdlib import check_output
```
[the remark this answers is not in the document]
radomil: updated with call

repos/system_upgrade/el7toel8/actors/firewalldisable/actor.py
bocekm: Well, it's nice that it raises CalledProcessError but actor catches the error. If this error is raised it would generate nice and shiny traceback for the user.
But this rant is not that relevant to this PR. We'll change the stdlib call in future to use the @shaded-enmity's https://github.com/oamg/leapp/pull/387 under the hood. Then, we'll need to update all actors again to accept the new return values.

bocekm:
@vinzenz, please re-review.

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
```diff
@@ -22,7 +22,7 @@ def produce_error(self, error):
     def is_system_registered_and_attached(self):
         # TODO: put this to different actor and process it already during check
         # + phase
-        out = preparetransaction.call(['subscription-manager', 'list', '--consumed'], True)
+        out = call(['subscription-manager', 'list', '--consumed'], True)
```
vinzenz: please use split=True for the split, so that a change in the signature won't break it

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
```diff
@@ -212,7 +212,7 @@ def process(self):
             return
 
         # TODO: Find a better place where to run this (perhaps even gate this behind user prompt/question)
-        preparetransaction.call(['/usr/bin/dnf', 'clean', 'all'])
+        call(['/usr/bin/dnf', 'clean', 'all'])
```
vinzenz: split=False, since it is not used, so it's a waste of resources to actually try to split it

radomil:
leapp-ci build

jmikovic:
leapp-ci build

PR #14, opened 2018-12-05
Add a doc about actor tests

bocekm:


docs/tests.md
```diff
@@ -0,0 +1,18 @@
+# Tests for actors
+
+You can implement two types of tests for an actor in the leapp-repository - functional and unit tests.
+
+## Functional tests
+   - These tests provide fabricated input messages for the actor, check the output messages and check for side effects stated in the actor's description.
+   - These tests should consider the actor to be a black box, so ideally they should written by somebody who doesn't know the code.
```
artmello:
```suggestion
   - These tests should consider the actor to be a black box, so ideally they could be written by somebody who doesn't know the code.
```

artmello: Even if I would really like this division between test/actor writer, in real world I believe this would not happen. I propose this change to change the meaning to something like "make sure your actor is well described in a way someone else can write the actor".

pirat89:
leapp-ci build

docs/tests.md
```diff
@@ -0,0 +1,18 @@
+# Tests for actors
+
+You can implement two types of tests for an actor in the leapp-repository - functional and unit tests.
+
+## Functional tests
+   - These tests provide fabricated input messages for the actor, check the output messages and check for side effects stated in the actor's description.
```
bocekm: From @amahdal: "check the output messages and check for side effects stated in the actor's description" .. side effect is something else, e.g. processor overheating. Mention just "check the outputs stated in the actor's description".

docs/tests.md
bocekm: From @AloisMahdal: do not mention black box, it may be confusing. We would need to define grey, white or black boxes.

docs/tests.md
```diff
@@ -0,0 +1,18 @@
+# Tests for actors
+
+You can implement two types of tests for an actor in the leapp-repository - functional and unit tests.
+
+## Functional tests
```
bocekm: From @AloisMahdal: do not use word _functional_. All tests we'll have for Leapp will be functional. Non-functional tests are stress tests, performance tests, load tests, etc. Use _component_ instead.

docs/tests.md
```diff
@@ -0,0 +1,18 @@
...
+   - Example of a functional test: https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/actors/redhatsignedrpmcheck/tests/test_redhatsignedrpmcheck.py
+
+## Unit tests
+   - These tests deal with individual actor's functions/methods.
+   - You can't unit test any method/function within the actor.py. You can write unit tests only for the functions/methods within actor's libraries.
+   - Thus, to be able to write unit tests for an actor, ideally the only thing in the actor.py's _process()_ method is calling the entry-point function of the actor's library python module.
+   - Tutorial on how to write unit tests: https://leapp.readthedocs.io/en/latest/unit-testing.html
+
+
+Both types of tests need to reside in the actor's _tests_ folder.
```
bocekm: Require only component tests, not unit tests. It may happen that an actor includes logic only in the actor's _process()_ method and then there would be little or no distinction between component and unit tests.

docs/tests.md
AloisMahdal: I googled E2E testing some time ago and generally the term seems to be defined in a very high-level way, meaning something along lines of: the process of carrying out very thorough and complete set of testing activities.

(Even the way we use it in OAMG-QE is sometimes a bit too specific; sometimes we use it just to refer to "automated upgrade tests".)

So I would prefer *component test*.

pirat89: Agree with Lojza. Component test fits better to me.

GAZDOWN: I am voting here for the name component test as well, it seems to be clear, plus, we are considering actor as a component of the whole system. Having the name E2E here would require more distinguishing and deeper specification. Personally, I do not want to go that way as there are other things to solve than definition of the words used here.

bocekm: _component_ it is then..

docs/tests.md
bocekm: Reworded to get rid of "black box".

docs/tests.md
bocekm: What is required and what not should be mentioned in the contribution guidelines. This docs it about describing the two test levels.

bocekm:
@AloisMahdal, @artmello, @shaded-enmity, the PR is ready for a re-review. Thanks.

pirat89:
leapp-ci build

pirat89:
Keeping unmerged several hours to give you time to react guys :)

PR #16, opened 2018-12-07
To remove python inotify

vinzenz:
Previously upgrades were failing when python-inotify was installed
in RHEL7. This adds it to the list of packages to remove during the
upgrade transaction until the PES data will solve the problem.
    
Additionally the entries have been sorted for easier looking up
things in the list.


vojtechsokol:
leapp-ci build

pirat89:
Merging as fix of tests will be done probably tomorrow. The failing tests is not related to this PR and it is more important to have this merged already.

Issue #23, opened 2018-12-12
`rhsm release --unset` in initramdisk


pirat89:
We hitting issue now because the network is down and rhsm is trying to connect to the server. We need to fix it properly to be able to do upgrades when people have set the release and do not block upgrades at all.

vinzenz:
We'll close this, because it's not going to happen anyway - But we'll take care of it in the FirstBoot phase

pirat89:
Will be resolved by https://github.com/oamg/leapp-repository/pull/260

PR #24, opened 2018-12-12
Comment out rhsm release unset to make upgrades possible now

pirat89:
We hitting issue now because the network is down and rhsm is trying to connect to the server. Commenting this out for now so people will not be affected in case they do not have set a release and we will have time to fix it properly.

Issue: #23 

Needs to be merged after PR #21 and rebase.


radomil:
LGTM

PR #26, opened 2018-12-13
Reduce the IPUWorkflow workflow

pirat89:
We found that several phases are not usable from our perspective of in-place upgrades. Such phases are commented out to make clear they should not be used and keeping info what was our original idea
in case we think about them again in future.

Documentation should be modified as well: https://leapp-to.github.io/upgrades

### DIFF
--- a/repos/system_upgrade/el7toel8/tags/attachpackagerepos.py
+++ b/repos/system_upgrade/el7toel8/tags/attachpackagerepos.py
@@ -1,5 +1,0 @@
-from leapp.tags import Tag
-
-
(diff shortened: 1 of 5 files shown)


pirat89:
leapp-ci build

vinzenz:
@pirat89 if you comment the phases out and comment out the import of the tags, I wouldn't delete those tags. Either, comment the class inside or leave it as is. Since you have those tags and their usage everywhere commented out, it's confusing in my opinion that after this commit there won't be the corresponding tags available.

pirat89:
@vinzenz  Prepared for review :)

PR #27, opened 2018-12-13
Invalid SELinux model fix and provided information

radomil:
SELinux model has been updated to expect only 'targeted', 'minimum', 'mls' policies as these are the only possible values.
SELinux facts gathering in SystemFacts library has been modified to a compact solution with libselinux-python library.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
+++ b/repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
@@ -5,10 +5,8 @@ import json
 import os
 import pwd
 import re
-import six
 import logging
 from subprocess import CalledProcessError
-from leapp.libraries.stdlib import call
 
 
 try:
@@ -20,6 +18,8 @@ from leapp.libraries.stdlib import call
 from leapp.models import SysctlVariablesFacts, SysctlVariable, ActiveKernelModulesFacts, ActiveKernelModule, \
     KernelModuleParameter, UsersFacts, User, GroupsFacts, Group, RepositoriesFacts, RepositoryFile, RepositoryData, \
     SELinuxFacts, fields, FirewallStatus, FirewallsFacts
+
+import six
 
 
 def aslist(f):
@@ -113,7 +113,7 @@ def _get_active_kernel_modules(logger):
                     parameter_dict[param] = fp.read().strip()
             except IOError as exc:
                 # Some parameters are write-only, in that case we just log the name of parameter
-                # and the module and continue 
+                # and the module and continue
                 if exc.errno in (errno.EACCES, errno.EPERM):
                     msg = 'Unable to read parameter "{param}" of kernel module "{name}"'
                     logger.warning(msg.format(param=param, name=name))
@@ -214,23 +214,30 @@ def get_repositories_status():
 
 def get_selinux_status():
     ''' Get SELinux status information '''
-    def asbool(x):
-        return x == 'enabled'
 
-    def identity(x):
-        return x
+    try:
+        import selinux
+    except ImportError:
+        api.report_error("SELinux Import Error", details="libselinux-python package must be installed.")
+        return
 
-    key_mapping = (('SELinux status', 'enabled', asbool),
-                   ('Policy MLS status', 'mls_enabled', asbool),
-                   ('Loaded policy name', 'policy', identity),
-                   ('Current mode', 'runtime_mode', identity),
-                   ('Mode from config file', 'static_mode', identity),)
+    outdata = dict({'enabled': selinux.is_selinux_enabled() == 1})
+    outdata['mls_enabled'] = selinux.is_selinux_mls_enabled == 1
 
-    outdata = {}
-    for item in call(['sestatus']):
-        for (key, mapped, cast) in key_mapping:
-            if item.startswith(key):
-                outdata[mapped] = cast(item[len(key)+1:].strip())
+    try:
+        outdata['runtime_mode'] = "enforcing" if selinux.security_getenforce() == 1 else "permissive"
+        enforce_mode = selinux.selinux_getenforcemode()
+        if enforce_mode >= 0:
+            outdata['static_mode'] = "enforcing" if enforce_mode == 1 else "permissive"
+        else:
+            outdata['static_mode'] = "disabled"
+        outdata['policy'] = selinux.selinux_getpolicytype()[1]
+    except OSError:
+        # This happens when SELinux is disabled
+        # [Errno 2] No such file or directory
+        outdata['runtime_mode'] = 'permissive'
+        outdata['static_mode'] = 'disabled'
+        outdata['policy'] = 'targeted'
 
     return SELinuxFacts(**outdata)
 
@@ -255,11 +262,11 @@ def get_firewalls_status():
             logger.debug('The %s service is likely not enabled nor running' % service_name)
 
         return FirewallStatus(
-                    active=active,
-                    enabled=enabled,
-                    )
+            active=active,
+            enabled=enabled,
+            )
 
     return FirewallsFacts(
-            firewalld=_get_firewall_status('firewalld'),
-            iptables=_get_firewall_status('iptables')
-            )
+        firewalld=_get_firewall_status('firewalld'),
+        iptables=_get_firewall_status('iptables'),
+        )
--- a/repos/system_upgrade/el7toel8/models/selinuxfacts.py
+++ b/repos/system_upgrade/el7toel8/models/selinuxfacts.py
@@ -5,10 +5,8 @@ from leapp.topics import SystemFactsTopic
 class SELinuxFacts(Model):
     topic = SystemFactsTopic
 
-    # FIXME: fixme properly regarding the issue:
-    # # https://github.com/oamg/leapp-repository/issues/20
     runtime_mode = fields.Nullable(fields.StringEnum(['enforcing', 'permissive']))
     static_mode = fields.StringEnum(['enforcing', 'permissive', 'disabled'])
     enabled = fields.Boolean()
-    policy = fields.String()
+    policy = fields.StringEnum(['targeted', 'minimum', 'mls'])
     mls_enabled = fields.Boolean()


repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -212,18 +212,38 @@ def asbool(x):
     def identity(x):
         return x
 
+    def _get_config_file_values():
+        _outdata = {}
+        _key_mapping = (('SELINUX=', 'static_mode', identity),
+                       ('SELINUXTYPE=', 'policy', identity),)
```
pirat89: indentation

repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -212,18 +212,38 @@ def asbool(x):
...
+
+        _outdata['enabled'] = False
+        _outdata['mls_enabled'] = False
```
pirat89: setup `runtime_mode` to `None` explicitly

repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -212,18 +212,38 @@ def asbool(x):
...
+
+        for item in call(['cat', '/etc/selinux/config']):
+            for (key, mapped, cast) in _key_mapping:
+                if item.startswith(key):
+                    _outdata[mapped] = cast(item[len(key):].strip())
+
+        return _outdata
```
pirat89: Information about `static_mode` and `policy` is taken from two sources of truth (`sestatus` in case selinux is enabled, config file in case selinux is disabled). I would prefer to use only one source of data (config file always) or maybe to setup `None` when data are not provided by the `sestatus` command? (@oamg/developers, guys what do you think? Another opinion is welcomed). Another think is that selinux can be disabled from the kernel commandline by command `selinux=0` - how to handle such situation? Probably topic for another PR (I prefer to resolve it separately as this is resolving the current problem we have now).

repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -212,18 +212,38 @@ def asbool(x):
...
 
+    def _get_config_file_values():
+        _outdata = {}
+        _key_mapping = (('SELINUX=', 'static_mode', identity),
+                       ('SELINUXTYPE=', 'policy', identity),)
+
+        _outdata['enabled'] = False
+        _outdata['mls_enabled'] = False
+
+        for item in call(['cat', '/etc/selinux/config']):
```
pirat89: the `open` function is enough. Do not use `call` here.

repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
shaded-enmity: We should definitely involve the selinux guys into this as there might be some special interactions among the various modes of enablement or disablement

bocekm: Two days ago on os-devel ML announcement from SELinux team introducing mtgs to discuss issues/questions/ideas. Too bad the first one is in late January.

pirat89: I guess that purpose of such mtgs is little different to the topic we are resolving here.

radomil: Regarding SELinux related system facts gathering. I did a research on python "selinux" library.
I propose replacing current solution which is using combination of sestatus command and a /etc/selinux/config file reading with the above mentioned library usage which covers both runtime and static - config information.

enabled <- selinux.is_selinux_enabled()
mls_enabled <- selinux.is_selinux_mls_enabled()
runtime_mode <- selinux.security_getenforce()
static_mode <- selinux.selinux_getenforcemode()
policy <- selinux.selinux_getpolicytype()

pirat89: That solution looks good to me. The library is not part of the std python library, but it is part of the minimal installation setup and it is required by other important applications (like firewalld). So I think this would be ok to use those functions. What from the point of other guys? @oamg/developers

shaded-enmity:
I think we should also add dependency in the spec file on selinux-python (or whatever the actual name is)

pirat89:
> I think we should also add dependency in the spec file on selinux-python (or whatever the actual name is)

Yes, the dependency in the SPEC file really should be added. And probably even into the `requirements.txt` file, for run of CI.

repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -206,23 +207,17 @@ def _parse(r):
...
-                outdata[mapped] = cast(item[len(key)+1:].strip())
+    outdata['enabled'] = True if selinux.is_selinux_enabled() == 1 else False
+    outdata['mls_enabled'] = True if selinux.is_selinux_mls_enabled == 1 else False
+    outdata['runtime_mode'] = "enforcing" if selinux.security_getenforce() == 1 else "permissive"
+    enforce_mode = selinux.selinux_getenforcemode()
+    if enforce_mode >= 0:
+        outdata['static_mode'] = "enforcing" if enforce_mode == 1 else "permissive"
+    else:
+        outdata['static_mode'] = "disabled"
+    outdata['policy'] = selinux.selinux_getpolicytype()[1]
```
pirat89: regarding the thing we cannot set dependency in the SPEC file, we are not sure the library is installed.  From that point on systems without the `libselinux-python` RPM the script will fail unhandled. Move import here and use `try` and catch `ImportError` to inform user that the rpm has to be installed. additionally, put commented out requires in the SPEC file so we will know we need them but it will not be used.

```
try:
    import selinux
    outdata...
except ImportError:
    self.report_error(...)
```

pirat89: See https://github.com/oamg/leapp/pull/405 - to be able to use report_error in the library

radomil: @pirat89 , copy.

repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -206,25 +207,23 @@ def _parse(r):
...
+        if enforce_mode >= 0:
+            outdata['static_mode'] = "enforcing" if enforce_mode == 1 else "permissive"
+        else:
+            outdata['static_mode'] = "disabled"
+        outdata['policy'] = selinux.selinux_getpolicytype()[1]
 
-    return SELinux(**outdata)
+        return SELinux(**outdata)
+    except ImportError as error:
+        api.report_error("SELinux Import Error", details="libselinux-python package must be installed.")
```
radomil: api.report_error will keep failing till leapp #405 is not merged into the master.

repos/system_upgrade/el7toel8/models/systemfacts.py
```diff
@@ -60,12 +60,10 @@ class ActiveKernelModule(Model):
 class SELinux(Model):
     topic = SystemInfoTopic
 
-    # FIXME: fixme properly regarding the issue:
-    # # https://github.com/oamg/leapp-repository/issues/20
     runtime_mode = fields.Nullable(fields.StringEnum(['enforcing', 'permissive']))
     static_mode = fields.StringEnum(['enforcing', 'permissive', 'disabled'])
     enabled = fields.Boolean()
-    policy = fields.String()
+    policy = fields.StringEnum(['targeted', 'minimum', 'mls'])
```
vinzenz: This field needs to be Nullable when selinux is disabled how it looks like

pirat89: Hmm...when selinux is disabled, still there would be set policy in the configuration file. Maybe instead of that, I would suggest use default policy (`targeted`), in case we cannot get it even from the config file. Not 100% about behaviour of the selinux function in such case.

pirat89: I mean, when selinux is disabled, it doesn't not mean it  is permanent state.

radomil: Right, these `'targeted', 'minimum', 'mls'` are the only values we should receive in whatever state SELinux is in.

vinzenz: see https://github.com/oamg/leapp-repository/issues/39

pirat89: Yes, that's purpose of this PR, to resolve #39  :) Btw. I checked functionality of `selinux.selinux_getpolicytype()`. So here are various results based on various invalid values in config file:
```python
# SELINUXTYPE=foo
>>> selinux.selinux_getpolicytype()
[0, 'foo']

# SELINUXTYPE=
>>> selinux.selinux_getpolicytype()
[0, '']

#
>>> selinux.selinux_getpolicytype()
[0, 'targeted']
```

Which is expected behaviour. Just in case of empty string, I would go with `targeted` anyway. In case someone will use custom policy, upgrade will not be enabled and we will now that we have to take care of it anyway. (in the meaning, that the value is probably important only in case selinux is enabled.. )

radomil:
leapp-ci

pirat89:
leapp-ci build

repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -206,25 +207,23 @@ def _parse(r):
...
+        import selinux
+        outdata['enabled'] = True if selinux.is_selinux_enabled() == 1 else False
+        outdata['mls_enabled'] = True if selinux.is_selinux_mls_enabled == 1 else False
+        outdata['runtime_mode'] = "enforcing" if selinux.security_getenforce() == 1 else "permissive"
+        enforce_mode = selinux.selinux_getenforcemode()
+        if enforce_mode >= 0:
+            outdata['static_mode'] = "enforcing" if enforce_mode == 1 else "permissive"
+        else:
+            outdata['static_mode'] = "disabled"
+        outdata['policy'] = selinux.selinux_getpolicytype()[1]
```
pirat89:
```suggestion
        if outdata['policy'] == '':
                outdata['policy'] = 'targeted'
```

radomil: @pirat89, as we discussed, this case of policy in the config file being empty may be rather a red flag for a leapp user.

repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -208,11 +208,14 @@ def _parse(r):
 def get_selinux_status():
     ''' Get SELinux status information '''
 
-    try:
-        import selinux
-    except ImportError:
-        api.report_error("SELinux Import Error", details="libselinux-python package must be installed.")
-        return
+    import sys
+    if not 'selinux' in sys.modules:
```
vinzenz: what is this solving? I can't think of anything. If it is already imported, then it won't import it again anyway. So the question at hand here, is what does this solve?

radomil: @vinzenz,  My intention: I want to ensure that when I mock selinux function in an unit test (PR #56) this mocking will not be overridden by local import in any way.
However, as I had dug deeper, it doesn't seem to be much helpful (in taking care of local imports in this way) as the module's name presence in sys.module doesn't actually mean it's visible in this function and then it takes as "local variable 'selinux' referenced before assignment". I'll rather remove it, then.

vinzenz: Well you can do this if you need to know if it's already there:
```python
if not 'selinux' in globals():
...
```

radomil:
@vinzenz, regarding that `if not 'selinux' in sys.modules:`
 It's a weird python behavior when I incorporate any of these conditions even if selinux is imported globally

- if not 'selinux' in globals() and/or
- if not 'selinux'sys.models conditions

I get UnboundLocalError: local variable 'selinux' referenced before assignment

To pass through the following reference of selinux successfully I had to put `global selinux` beforehand.

I just removed that last commit with the if condition! So, let's forget it for now.

Can it be merged?

repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -5,10 +5,11 @@
 import os
 import pwd
 import re
-import six
 import subprocess
 import logging
 from subprocess import CalledProcessError
+from leapp.libraries.stdlib import api
+import six
```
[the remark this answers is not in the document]
radomil: I made this adjustment because of 
pylint systemfacts.py | grep six
No config file found, using default configuration
C:  9, 0: standard import "import subprocess" should be placed before "import six" (wrong-import-order)
C: 10, 0: standard import "import logging" should be placed before "import six" (wrong-import-order)
C: 11, 0: standard import "from subprocess import CalledProcessError" should be placed before "import six" (wrong-import-order)

packaging/leapp-repository.spec
```diff
@@ -19,6 +19,11 @@ Requires:   %{name}-data = %{version}-%{release}
 # IMPORTANT: everytime the requirements are changed, increment number by one
 # - same for Provides in deps subpackage
 Requires:   leapp-repository-dependencies = 1
+%if 0%{?fedora} || 0%{?rhel} > 7
+Requires:   systemd-container
+%endif
+# Required by SystemFacts Actor (library) to gather system facts about SELinux
+# Requires: libselinux-python
```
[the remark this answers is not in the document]
vinzenz: packaging/leapp-el7toel8-deps.spec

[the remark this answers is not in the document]
radomil: to doublecheck
packaging/leapp-el7toel8-deps.spec describes packages needed in RHEL 8, is that right?
 there is a replacement
in RHEL 7 libselinux-python is needed | in RHEL 8 it's python3-libselinux
Isn't the RHEL7's package libselinux-python meant to be in packaging/leapp-repository.spec?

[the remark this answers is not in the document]
vinzenz: and again you are right, I am lost. However you did not increase the number:
```
#IMPORTANT: everytime the requirements are changed, increment number by one
# - same for Provides in deps subpackage
```

PR #30, opened 2019-01-02
WIP: Add template for bug reporting

pirat89:


Rezney:
All my concerns/suggestions already communicated offline. LGTM.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,44 @@
...
+2. setup '....'
+3. run '....'
+   (when run `leapp`, use always the `--debug` option to provide more data)
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**System information (please complete the following information):**
+ - OS and version: [e.g. Fedora 29]
```
shaded-enmity: I'd just ask for the output of `uname -a` here

vinzenz: Maybe we should also ask for the leapp-repository NVRA or the git commit id hash if from repo?

pirat89: discussed with guys. uname is good thing but it will not tell you what system you think you are using. I would rather require both info.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,44 @@
+---
+name: Bug report
+about: Create a report to help us improve
+labels: bug
+
+---
+
+**Describe the bug**
+A clear and concise description of what the bug is.
```
shaded-enmity: I'd drop this section and add `Actual behavior` to complement the `Expected behavior` section.

.github/ISSUE_TEMPLATE/bug_report.md
shaded-enmity: ok, so the output of `uname -a` and `cat /etc/redhat-release`?

shaded-enmity:
LGTM

(Slight nitpick, in some instances shell commands are prepended with either `$`, `#` or nothing -- can we use just one format?)

pirat89:
> LGTM
> 
> (Slight nitpick, in some instances shell commands are prepended with either `$`, `#` or nothing -- can we use just one format?)

Fixed. Now refer to commands processed by root (`#`).

PR #31, opened 2019-01-03
Demonstration of handling of spurious failures

vojtechsokol:
When connection goes down or there is no free space on the disk during the upgrade, the process fails without user knowing that there is no problem with the upgrade itself.

This PR tries to address this issue by introducing new function that wraps subprocess.Popen and handles process's stdout, return codes and error messages.

This function also accepts optional arguments (callables) that can preform various checks. Currently `connection_checker` (for checking internet connection) and `space_checker` (for checking disk free space) are implemented and used .

The approach is demonstrated on `PrepareUpgradeTransaction` actor, where most of the
subprocess calls in `actor.py` are modified to use the introduced functions.


vojtechsokol:
leapp-ci build

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
```diff
@@ -208,54 +206,72 @@ def process(self):
...
-                self.produce_error(error)
+                error.summary = 'Cannot set minor release version.'
+                preparetransaction.produce_error(self, error)
                 error_flag = True
 
         # clean #
         error = preparetransaction.umount_overlayfs(ofs_info)
         if error:
-            self.produce_error(error)
+            preparetransaction.produce_error(self, error)
```
vinzenz: why?

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
```diff
@@ -208,54 +206,72 @@ def process(self):
...
-                        )
+                ofs_info,
+                '/etc/yum.repos.d/redhat.repo',
+                '/etc/yum.repos.d/',
+                'redhat.repo.upgrade',
+            )
 
         if error:
-            self.produce_error(error)
+            preparetransaction.produce_error(self, error)
```
vinzenz: why?

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
```diff
@@ -179,20 +177,20 @@ def dnf_shell_rpm_download(self, overlayfs_info):
...
             script.flush()
 
-        error = preparetransaction.check_container_call(overlayfs_info, [
-            '--',
-            '/bin/bash',
-            '-c',
-            ' '.join(dnf_command + ['/var/lib/leapp/dnf-script.txt'])])
+        # FIXME: fails on insufficient permissions
+        cmd = ['--', '/bin/bash', '-c', ' '.join(dnf_command + ['/var/lib/leapp/dnf-script.txt'])]
+        _, error = preparetransaction.guard_container_call(overlayfs_info, cmd, checkers=(preparetransaction.connection_checker, preparetransaction.space_checker))
```
vinzenz: This line seems to be above 120

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
```diff
@@ -208,54 +206,72 @@ def process(self):
                          ' to proceed the upgrade. Register your system with the'
                          ' subscription-manager tool and attach'
                          ' it to proper SKUs to be able to proceed the upgrade.'))
-            self.produce_error(error)
+            preparetransaction.produce_error(self, error)
             return
 
         # TODO: Find a better place where to run this (perhaps even gate this behind user prompt/question)
-        preparetransaction.call(['/usr/bin/dnf', 'clean', 'all'])
+        # FIXME: fails on insufficient permissions
```
vinzenz: Insufficient permissions? When do you encounter that? Just out of curiosity

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
```diff
@@ -179,20 +177,20 @@ def dnf_shell_rpm_download(self, overlayfs_info):
...
-            '/bin/bash',
-            '-c',
-            ' '.join(dnf_command + ['/var/lib/leapp/dnf-script.txt'])])
+        # FIXME: fails on insufficient permissions
+        cmd = ['--', '/bin/bash', '-c', ' '.join(dnf_command + ['/var/lib/leapp/dnf-script.txt'])]
+        _, error = preparetransaction.guard_container_call(overlayfs_info, cmd, checkers=(preparetransaction.connection_checker, preparetransaction.space_checker))
 
         if os.environ.get('LEAPP_DEBUG', '0') == '1':
             # We want the debug data available where we would expect it usually.
+            # FIXME: no exception is caught, retcode is not checked
```
vinzenz: well rm -rf shouldn't be really raising an exception call was used here originally because we don't care if either of those call fail - We might want to log a warning maybe

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
```diff
@@ -179,20 +177,20 @@ def dnf_shell_rpm_download(self, overlayfs_info):
...
             call(['rm', '-rf', os.path.join('/tmp', 'download-debugdata')])
             call(['cp', '-a', os.path.join(overlayfs_info.merged, 'debugdata'), '/tmp/download-debugdata'])
```
vinzenz: That might be rather the one we care about failing, but we shouldn't cause it to break the upgrade - since this is debug data only

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
```diff
@@ -208,54 +206,72 @@ def process(self):
...
         if error:
-            self.produce_error(error)
+            preparetransaction.produce_error(self, error)
             preparetransaction.remove_overlayfs_dirs(container_root)
             return
 
         # switch EngID to use RHEL 8 subscriptions #
-        prev_rhsm_release = preparetransaction.call(['subscription-manager', 'release'])[0]
+        cmd = ['subscription-manager', 'release']
+        prev_rhsm_release, error = preparetransaction.guard_call(
```
vinzenz: I wonder if guard_call couldn't be rather a context manager for more generic usage:
```python
with call_guard(checkers=....):
    do_call()
```

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
pirat89: @vinzenz selinux would forbid the access possibly?

pirat89: @vojtechsokol As I wrote, even when user is root, it can be hit that it will not be able to open a file for write because of selinux or another security technology.

vojtechsokol: @pirat89 However in this case this problem can happen to virtually all subprocess calls. Is it necessary then to keep those FIXME comments? If yes, they should be added to all subprocess calls to be consistent...

vojtechsokol:
@vinzenz
Yes, the guard functions (and other related code) are definitely useful for other actors and it is intended to move it eventually to stdlib, but this was just demonstration of reporting spurious failures on this particular actor (as it is most susceptible to spurious failures). If this approach is deemed appropriate, I'll open a PR where it will be moved to stdlib.

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
pirat89: @vojtechsokol: nope. But maybe rather we should detect something like that later in the call function itself. Or by the guard funsction.

vojtechsokol:
leapp-ci build

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
```diff
@@ -71,11 +70,19 @@ def update_rhel_subscription(self, overlayfs_info):
             # + would be problem later). Needs a little investigation here about the impact, even
             # + in the container.
             for cert in existing_certs:
-                error = preparetransaction.check_container_call(overlayfs_info, ['rm', '-f', cert])
+                # FIXME: fails on insufficient permissions
+                cmd = ['rm', '-f', cert]
+                _, error = preparetransaction.guard_container_call(overlayfs_info, cmd)
```
vinzenz: please don't use _ eventually this will be used for localization, use _unused, the underscore infront of it will avoid warnings about unused variables and

vojtechsokol: fixed

PR #33, opened 2019-01-07
SPEC: add missing dirs into the files and rm tests

pirat89:
Some directories have not been part of the rpm. That caused messing directories have been kept on the system after removal of the rpm.

Additionally remove tests from buildroot as those tests are not expected to deliver with rpm and just messing fs.

Note: w8 for builds and check the file list.


vinzenz:
We also should remove the Makefiles which are only useful for the tests

pirat89:
Everything should fine now. Prepared for final review.

PR #34, opened 2019-01-07
Switch to dnf plugin

Rezney:
In order to have better control of transaction we switch to a dnf plugin
instead of dnf shell command.


pirat89:
Anyway, it works good regarding to my upgrades, just the *file* thing above should be fixed.

pirat89:
Another thing is, that part of this PR should be change of the SPEC file. The rpm should require newer version of dnf

Rezney:
@pirat89 your comments should be now addressed...

pirat89:
leapp-ci build

pirat89:
leapp-ci build

dirgim:
leapp-ci build

PR #35, opened 2019-01-08
SCTP Actors: unblacklist SCTP kernel modules on upgrade if needed

marceloleitner:
Some modules can be autoloaded by user-triggered actions, including
SCTP stack. To enhance system security, we will disable that for some
modules, including SCTP.

These Actors are responsible for identifying that SCTP is being used
on the system being upgraded and then re-enable it automatically.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/sctpchecks/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/sctpchecks/actor.py
@@ -1,0 +1,23 @@
+from leapp.actors import Actor
+from leapp.models import SCTPConfig, RpmTransactionTasks
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+class SCTPChecks(Actor):
+    """
+    Parses collected SCTP information and take necessary actions.
+
+    The only action performed by this actor is to request the installation of the
+    kernel-modules-extra rpm package, based on if SCTP is being used or not which
+    is collected on SCTPConfig message. If yes, it then produces a RpmTransactionTasks
+    requesting to install the package.
+    """
+    name = 'sctp_checks'
+    consumes = (SCTPConfig,)
+    produces = (RpmTransactionTasks, )
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        for sctpconfig in self.consume(SCTPConfig):
+            if sctpconfig.wanted:
+                self.produce(RpmTransactionTasks(to_install=['kernel-modules-extra']))
+                break
--- a/repos/system_upgrade/el7toel8/actors/sctpconfigread/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/sctpconfigread/actor.py
@@ -1,0 +1,21 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.sctplib import is_sctp_wanted
+from leapp.models import ActiveKernelModulesFacts, SCTPConfig
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+
+
+class SCTPConfigRead(Actor):
+    """
+    Determines whether or not the SCTP kernel module might be wanted.
+
+    This actor determines whether or not the SCTP is currently used by this machine or has been quite
+    recently used (1 month timeframe). In case it has been used it will issue a SCTPConfig message that
+    defines the decision whether or not the SCTP module should be removed from the module blacklist on RHEL8.
+    """
+    name = 'sctp_read_status'
+    consumes = (ActiveKernelModulesFacts,)
+    produces = (SCTPConfig,)
+    tags = (FactsPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        self.produce(SCTPConfig(wanted=is_sctp_wanted()))
--- a/repos/system_upgrade/el7toel8/actors/sctpconfigread/libraries/sctpdlm.py
+++ b/repos/system_upgrade/el7toel8/actors/sctpconfigread/libraries/sctpdlm.py
@@ -1,0 +1,64 @@
+#
+# Functions for probing SCTP usage by DLM
+#
+import re
+
+from six.moves.configparser import SafeConfigParser
+
+from leapp.libraries.stdlib import api
+
+
+def check_dlm_cfgfile():
+    """Parse DLM config file"""
+    fname = "/etc/dlm/dlm.conf"
+
+    try:
+        with open(fname, 'r') as fp:
+            cfgs = '[dlm]\n' + fp.read()
+    except (OSError, IOError):
+        return False
+
+    cfg = SafeConfigParser()
+    try:
+        cfg.read_string(cfgs)
+    except AttributeError:
+        # Python2 ConfigParser doesn't have cfg.read_string
+        from cStringIO import StringIO
+        cfg.readfp(StringIO(cfgs))
+
+    if not cfg.has_option('dlm', 'protocol'):
+        return False
+
+    proto = cfg.get('dlm', 'protocol').lower()
+    return proto in ['sctp', 'detect', '1', '2']
+
+
+def check_dlm_sysconfig():
+    """Parse /etc/sysconfig/dlm"""
+    regex = re.compile('^[^#]*DLM_CONTROLD_OPTS.*=.*(?:--protocol|-r)[ =]*([^"\' ]+).*', re.IGNORECASE)
+
+    try:
+        with open('/etc/sysconfig/dlm', 'r') as fp:
+            lines = fp.readlines()
+    except (OSError, IOError):
+        return False
+
+    for line in lines:
+        if regex.match(line):
+            proto = regex.sub('\\1', line).lower().strip()
+            if proto in ['sctp', 'detect']:
+                return True
+
+    return False
+
+
+def is_dlm_using_sctp():
+    if check_dlm_cfgfile():
+        api.current_logger().info('DLM is configured to use SCTP on dlm.conf.')
+        return True
+
+    if check_dlm_sysconfig():
+        api.current_logger().info('DLM is configured to use SCTP on sysconfig.')
+        return True
+
+    return False
--- a/repos/system_upgrade/el7toel8/actors/sctpconfigupdate/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/sctpconfigupdate/actor.py
@@ -1,0 +1,21 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import sctpupdate
+from leapp.models import SCTPConfig
+from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
+
+
+class SCTPConfigUpdate(Actor):
+    """
+    Updates the kernel module blacklist for SCTP.
+
+    If the SCTP module is wanted on RHEL8 the modprobe configuration gets updated to remove SCTP from the black listed
+    kernel modules.
+    """
+    name = 'sctp_config_update'
+    description = 'This actor updates SCTP configuration for RHEL8.'
+    consumes = (SCTPConfig,)
+    produces = ()
+    tags = (ApplicationsPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        sctpupdate.perform_update()
--- a/repos/system_upgrade/el7toel8/models/sctpconfigmodel.py
+++ b/repos/system_upgrade/el7toel8/models/sctpconfigmodel.py
@@ -1,0 +1,6 @@
+from leapp.models import Model, fields
+from leapp.topics import SCTPConfigTopic
+
+class SCTPConfig(Model):
+    topic = SCTPConfigTopic
+    wanted = fields.Boolean(default=False)
--- a/repos/system_upgrade/el7toel8/actors/sctpconfigread/libraries/sctplib.py
+++ b/repos/system_upgrade/el7toel8/actors/sctpconfigread/libraries/sctplib.py
@@ -1,0 +1,107 @@
+#
+# Helper functions
+#
+
+from os.path import isfile
+from subprocess import CalledProcessError
+
+from leapp.libraries.actor import sctpdlm
+from leapp.libraries.stdlib import api, call
+from leapp.models import ActiveKernelModulesFacts
+
+
+def anyfile(files):
+    """
+    Determines if any of the given paths exist and are a file.
+    
+    :return: True if any of the given paths exists and it is a file.
+    :rtype: bool
+    """
+    for f in files:
+        try:
+            if isfile(f):
+                return True
+        except OSError:
+            continue
+    return False
+
+
+def is_module_loaded(module):
+    """
+    Determines if the given kernel module has been reported in the ActiveKernelModuleFacts as loaded.
+
+    :return: True if the module has been found in the ActiveKernelModuleFacts.
+    :rtype: bool
+    """    
+    for fact in api.consume(ActiveKernelModulesFacts):
+        for active_module in fact.kernel_modules:
+            if active_module.filename == module:
+                return True
+    return False
+
+
+def is_sctp_used():
+    """
+    Logic function that decides whether SCTP is being used on this machine.
+
+    :return: True if SCTP usage was detected.
+    :rtype: bool
+    """
+
+    # If anything is using SCTP, be it for listening on new connections or
+    # connecting somewhere else, the module will be loaded. Thus, no need to
+    # also probe on sockets.
+    if is_module_loaded('sctp'):
+        return True
+
+    # Basic files from lksctp-tools. This check is enough and checking RPM
+    # database is an overkill here and this allows for checking for
+    # manually installed ones, which is possible.
+    lksctp_files = ['/usr/lib64/libsctp.so.1',
+                    '/usr/lib/libsctp.so.1',
+                    '/usr/bin/sctp_test']
+    if anyfile(lksctp_files):
+        api.current_logger().debug('At least one of lksctp files is present.')
+        return True
+
+    if sctpdlm.is_dlm_using_sctp():
+        return True
+
+    return False
+
+
+def was_sctp_used():
+    """
+    Determines whether SCTP has been used in the path month, by checking the journalctl.
+
+    :return: True if SCTP usage has been found.
+    :rtype: bool
+    """
+    try:
+        call(['check_syslog_for_sctp.sh'])
+    except CalledProcessError:
+        api.current_logger().debug('Nothing regarding SCTP was found on journal.')
+        return False
+    api.current_logger().debug('Found logs regarding SCTP on journal.')
+    return True
+
+
+
+def is_sctp_wanted():
+    """
+    Decision making funtion that decides based on the current or past usage of SCTP, the SCTP module is wanted
+    on the new system.
+
+    :return: True if SCTP seems to be in use or has been recently used.
+    :rtype: bool
+    """
+    if is_sctp_used():
+        api.current_logger().info('SCTP is being used.')
+        return True
+
+    if was_sctp_used():
+        api.current_logger().info('SCTP was used.')
+        return True
+
+    api.current_logger().info('SCTP is not being used and neither wanted.')
+    return False
--- a/repos/system_upgrade/el7toel8/actors/sctpconfigread/tools/check_syslog_for_sctp.sh
+++ b/repos/system_upgrade/el7toel8/actors/sctpconfigread/tools/check_syslog_for_sctp.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+/usr/bin/journalctl --system -S '1 month ago' | /usr/bin/grep -q -m1 -w sctp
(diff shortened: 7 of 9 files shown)


vinzenz:
Thank you Marcelo for this contribution. To be honest we quite impressed how far you got already!

There are some improvements we need you to apply, but this is already a quite impressive first state!
Please give us time to go over with it.

BTW you are the first external contributor to our team with this PR, so even congratulations on that one :-)

repos/system_upgrade/el7toel8/actors/sctpconfigread/actor.py
```diff
@@ -0,0 +1,159 @@
...
+    proto = cfg.get('dlm', 'protocol').lower()
+    if proto in ['sctp', 'detect', '1', '2']:
+        return True
+
+    return False
+
+def check_dlm_sysconfig():
+    """Parse /etc/sysconfig/dlm"""
+    try:
+        #fp = open('/etc/sysconfig/dlm', 'r')
```
marceloleitner: Timothy noticed I left a test code here. Will need a respin for this one at least.

repos/system_upgrade/el7toel8/actors/sctpconfigread/actor.py
```diff
@@ -0,0 +1,159 @@
...
+        return True
+
+    if is_dlm_using_sctp():
+        return True
+
+    return False
+
+def was_sctp_used():
+    from subprocess import check_output
+    cmd = ['/bin/sh', '-c', "/usr/bin/journalctl --system -S '1 month ago' | /usr/bin/grep -qw sctp"]
```
vinzenz: Please try to avoid /bin/sh -c  here, if you really need a shell script, the best is to add a 'tools' folder within your actor folder and put a script into there, that is executable (a+x flag) and call by name. The path to the tools folder is added to the PATH variable before the actor is executed so you can call it by referring to its name only.

Additionally you tell grep -q which means exit code 0 or 1 this code will raise a CalledProcessError on check_output if it didn't match anything. Which seems according to your code to be not what you wanted.

repos/system_upgrade/el7toel8/actors/sctpconfigread/actor.py
```diff
@@ -0,0 +1,159 @@
...
+    try:
+        output = check_output(cmd)
+        if len(output):
+            logging.debug('Found logs regarding SCTP on journal.')
+            return True
+        logging.debug('Nothing regarding SCTP was found on journal.')
+    except Exception as err:
```
vinzenz: Please do not Catch exceptions this generic, they preferably should be restricted to the exceptions that should be thrown at that time. Here it could be an `OSError` or a `subprocess.CalledProcessError` due to check_output

repos/system_upgrade/el7toel8/actors/sctpconfigread/actor.py
```diff
@@ -0,0 +1,159 @@
...
+    tags = (FactsPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        global logging
+        logging = self.log
+
+        try:
+            sctpconfig = SCTPConfig()
+            sctpconfig.wanted = is_sctp_wanted()
+        except Exception as e:
```
vinzenz: Again, too wide exception catching here

repos/system_upgrade/el7toel8/actors/sctpconfigupdate/actor.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def enable_sctp(self):
+        from subprocess import check_call
+        self.log.info('Enabling SCTP.')
+        try:
+            # FIXME: Improve error logging
+            check_call(['/usr/bin/sed', '-i', 's/^\s*blacklist.*sctp/#&/',
+                        '/etc/modprobe.d/sctp_diag-blacklist.conf',
+                        '/etc/modprobe.d/sctp-blacklist.conf'])
+        except Exception as err:
```
vinzenz: Too wide exception handling here.

repos/system_upgrade/el7toel8/actors/sctpconfigread/actor.py
```diff
@@ -0,0 +1,159 @@
...
+
+def is_sctp_used():
+    # If anything is using SCTP, be it for listening on new connections or
+    # connecting somewhere else, the module will be loaded. Thus, no need to
+    # also probe on sockets.
+    if check_loaded_module('sctp'):
+        return True
+
+    # Basic files from lksctp-tools. This check is enough and checking RPM database
+    # is an overkill here.
```
vinzenz: You can also consume the `leapp.models.InstalledRPM` Model and check against them if the lksctp-tools are installed. But this should be ok, if this is safe enough for your test.

marceloleitner:
On Mon, Jan 14, 2019 at 07:10:20AM -0800, vinzenz wrote:
> vinzenz requested changes on this pull request.
> 
> I have left a few things to be addressed. 
> 
> There are two things in general I would like you to change. Please move the code that is global rather into a separate python module into `repos/system_upgrade/el7toel8/actors/sctpconfigread/libraries/` This module you can import via `from leapp.libraries.actor import mymodule` and use it further out. 

Ok.

> Also we usually try to have tests for the actors and their libraries. Especially in the case of the kernel modules, it's easier to fake the presence of a module, if you would use the message instead of the check you do yourself.
> We can work the tests out together once you applied the other comments.

Right. Yeah I'll probably need some help with the tests.

> 
> Anyway once more, already great work. Just we need to get the PR in shape :-)

Thanks!

> 
> > @@ -0,0 +1,159 @@
> +from leapp.actors import Actor
> +from leapp.models import SCTPConfig
> +from leapp.tags import FactsPhaseTag, IPUWorkflowTag
> +
> +#
> +# Generic functions
> +#
> +def check_loaded_module(module):
> +    """Check if a given module is loaded."""
> +    import os
> +    try:
> +        os.stat('/sys/module/%s' % module)
> 
> Instead of checking for the loaded modules yourself, we'd prefer if you would consume the message that reports active modules.
> 
> To accomplish this you would consume the [`leapp.models.SystemFacts`](https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/models/systemfacts.py) Model and use the `kernel_modules` field from it. See: 

Ok. Moved it to a class method then in order to have access to
'self.consume'.

>  
> 
> > +
> +#
> +# Generic functions
> +#
> +def check_loaded_module(module):
> +    """Check if a given module is loaded."""
> +    import os
> +    try:
> +        os.stat('/sys/module/%s' % module)
> +    except OSError:
> +        logging.debug('Module %s is not loaded.' % module)
> +        return False
> +    logging.debug('Module %s is loaded.' % module)
> +    return True
> +
> +def check_any_file_present(files):
> 
> just FYI this is basically the same as: `any(map(os.path.exists,
> files))` if you want to be sure that they are files you use
> `os.path.isfile` instead of `os.path.exists`, so I am not sure if
> this really needs an own function

Mostly agree. The difference is that the manual for will end as soon
as the first element is found. Pretty much like the 'anyendswith' &
cia helpers from systemfacts/libraries/systemfacts.py.
Here, we also only need to know if at least 1 element matched.

Created a libraries/sctplib.py, with this function as 'anyfile' in
there.

> 
> > +    for f in files:
> +        try:
> +            if os.stat(f):
> +                logging.debug('File %s is present.' % module)
> +                return True
> +        except OSError:
> +            continue
> +    return False
> +
> +#
> +# Functions for probing for SCTP usage
> +#
> +def check_dlm_cfgfile():
> +    """Parse the config file"""
> +    fname = "test.conf"
> +    try:
> 
> Prefer using the context manager when writing code, that allows for safer code, and you can't easily forget about closing the handle.
> 
> I would write it like this:
> 
> ```python
> # You may put this on top of the actor.py
> from six.moves.configparser import SafeConfigParser
> 
> if not os.path.isfile(fname):
>     return False
> 
> # Read the content
> with open(fname) as f:
>     cfgs = '[dlm]\n' + fp.read()
> 
> cfg = SafeConfigParser()
> if hasattr(cfg, 'read_string'): 
>     # Python 3
>     cfg.read_string(cfgs)
> else:
>     # Python 2 
>     from cStringIO import StringIO
>     cfg.readfp(StringIO(cfgs))
> ```

Ok.

> 
> 
> > +
> +    cfgs = '[dlm]\n' + fp.read()
> +    del fp
> +
> +    try:
> +        from configparser import SafeConfigParser
> +        cfg = SafeConfigParser()
> +        cfg.read_string(fp)
> +    except:
> +        # Python2 ConfigParser doesn't have cfg.read_string
> +        from cStringIO import StringIO
> +        from ConfigParser import SafeConfigParser
> +        cfg = SafeConfigParser()
> +        cfg.readfp(StringIO(cfgs))
> +
> +    proto = cfg.get('dlm', 'protocol').lower()
> 
> You probably should check whether or not the protocol variable exists to be safe:
> ```cfg.has_option('dlm', 'protocol')```

Yep.

> 
> > +    cfgs = '[dlm]\n' + fp.read()
> +    del fp
> +
> +    try:
> +        from configparser import SafeConfigParser
> +        cfg = SafeConfigParser()
> +        cfg.read_string(fp)
> +    except:
> +        # Python2 ConfigParser doesn't have cfg.read_string
> +        from cStringIO import StringIO
> +        from ConfigParser import SafeConfigParser
> +        cfg = SafeConfigParser()
> +        cfg.readfp(StringIO(cfgs))
> +
> +    proto = cfg.get('dlm', 'protocol').lower()
> +    if proto in ['sctp', 'detect', '1', '2']:
> 
> since you don't do anything further with it, it's fine to do:
> ```return proto in ['sctp', 'detect', '1', '2']``` 

Ok. I usually prefer not to do so because I think the interpreter will
optimize that for us and we can add other conditions without chaging
the surronding lines (for git blame). But ok.

> 
> > +        from cStringIO import StringIO
> +        from ConfigParser import SafeConfigParser
> +        cfg = SafeConfigParser()
> +        cfg.readfp(StringIO(cfgs))
> +
> +    proto = cfg.get('dlm', 'protocol').lower()
> +    if proto in ['sctp', 'detect', '1', '2']:
> +        return True
> +
> +    return False
> +
> +def check_dlm_sysconfig():
> +    """Parse /etc/sysconfig/dlm"""
> +    try:
> +        #fp = open('/etc/sysconfig/dlm', 'r')
> +        fp = open('dlm', 'r')
> 
> You don't seem to close the handle, so better to use the context manager.

Yep.

> 
> > +        # Python2 ConfigParser doesn't have cfg.read_string
> +        from cStringIO import StringIO
> +        from ConfigParser import SafeConfigParser
> +        cfg = SafeConfigParser()
> +        cfg.readfp(StringIO(cfgs))
> +
> +    proto = cfg.get('dlm', 'protocol').lower()
> +    if proto in ['sctp', 'detect', '1', '2']:
> +        return True
> +
> +    return False
> +
> +def check_dlm_sysconfig():
> +    """Parse /etc/sysconfig/dlm"""
> +    try:
> +        #fp = open('/etc/sysconfig/dlm', 'r')
> 
> I guess  you mean that you are using dlm vs /etc/sysconfig/dlm?

Yes, sorry for that. Forgot my test files in place.

> 
> > +    import os
> +    for f in files:
> +        try:
> +            if os.stat(f):
> +                logging.debug('File %s is present.' % module)
> +                return True
> +        except OSError:
> +            continue
> +    return False
> +
> +#
> +# Functions for probing for SCTP usage
> +#
> +def check_dlm_cfgfile():
> +    """Parse the config file"""
> +    fname = "test.conf"
> 
> You will need to put the full path spec here, I don't think this file resides with the actor.py (The CWD)

Same here, actually. This should have been /etc/dlm/dlm.conf

> 
> > +
> +    # Basic files from lksctp-tools. This check is enough and checking RPM database
> +    # is an overkill here.
> +    lksctp_files = [ '/usr/lib64/libsctp.so.1',
> +                     '/usr/lib/libsctp.so.1',
> +                     '/usr/bin/sctp_test' ]
> +    if check_any_file_present(lksctp_files):
> +        return True
> +
> +    if is_dlm_using_sctp():
> +        return True
> +
> +    return False
> +
> +def was_sctp_used():
> +    from subprocess import check_output
> 
> Please use [`leapp.libraries.stdlib.call`](https://leapp.readthedocs.io/en/latest/pydoc/leapp.libraries.stdlib.html#leapp.libraries.stdlib.call) instead 

Ok.

> 
> > +    # Basic files from lksctp-tools. This check is enough and checking RPM database
> +    # is an overkill here.
> +    lksctp_files = [ '/usr/lib64/libsctp.so.1',
> +                     '/usr/lib/libsctp.so.1',
> +                     '/usr/bin/sctp_test' ]
> +    if check_any_file_present(lksctp_files):
> +        return True
> +
> +    if is_dlm_using_sctp():
> +        return True
> +
> +    return False
> +
> +def was_sctp_used():
> +    from subprocess import check_output
> +    cmd = ['/bin/sh', '-c', "/usr/bin/journalctl --system -S '1 month ago' | /usr/bin/grep -qw sctp"]
> 
> Please try to avoid /bin/sh -c  here, if you really need a shell script, the best is to add a 'tools' folder within your actor folder and put a script into there, that is executable (a+x flag) and call by name. The path to the tools folder is added to the PATH variable before the actor is executed so you can call it by referring to its name only.

Ok.

> 
> Additionally you tell grep -q which means exit code 0 or 1 this code will raise a CalledProcessError on check_output if it didn't match anything. Which seems according to your code to be not what you wanted. 

Right, it would log the wrong log message then.

> 
> > +
> +    if is_dlm_using_sctp():
> +        return True
> +
> +    return False
> +
> +def was_sctp_used():
> +    from subprocess import check_output
> +    cmd = ['/bin/sh', '-c', "/usr/bin/journalctl --system -S '1 month ago' | /usr/bin/grep -qw sctp"]
> +    try:
> +        output = check_output(cmd)
> +        if len(output):
> +            logging.debug('Found logs regarding SCTP on journal.')
> +            return True
> +        logging.debug('Nothing regarding SCTP was found on journal.')
> +    except Exception as err:
> 
> Please do not Catch exceptions this generic, they preferably should be restricted to the exceptions that should be thrown at that time. Here it could be an `OSError` or a `subprocess.CalledProcessError` due to check_output

Ok

> 
> > +
> +class SCTPConfigRead(Actor):
> +    name = 'sctp_read_status'
> +    description = 'This actor reads SCTP status, if it was/is used.'
> +    consumes = ()
> +    produces = (SCTPConfig,)
> +    tags = (FactsPhaseTag, IPUWorkflowTag)
> +
> +    def process(self):
> +        global logging
> +        logging = self.log
> +
> +        try:
> +            sctpconfig = SCTPConfig()
> +            sctpconfig.wanted = is_sctp_wanted()
> +        except Exception as e:
> 
> Again, too wide exception catching here
> 
> > +class SCTPConfigUpdate(Actor):
> +    name = 'sctp_config_update'
> +    description = 'This actor updates SCTP configuration for RHEL8.'
> +    consumes = (SCTPConfig,)
> +    produces = ()
> +    tags = (ApplicationsPhaseTag, IPUWorkflowTag)
> +
> +    def enable_sctp(self):
> +        from subprocess import check_call
> +        self.log.info('Enabling SCTP.')
> +        try:
> +            # FIXME: Improve error logging
> +            check_call(['/usr/bin/sed', '-i', 's/^\s*blacklist.*sctp/#&/',
> +                        '/etc/modprobe.d/sctp_diag-blacklist.conf',
> +                        '/etc/modprobe.d/sctp-blacklist.conf'])
> +        except Exception as err:
> 
> Too wide exception handling here. 

Ok

> 
> > +
> +    if check_dlm_sysconfig():
> +        logging.info('DLM is configured to use SCTP on sysconfig.')
> +        return True
> +
> +    return False
> +
> +def is_sctp_used():
> +    # If anything is using SCTP, be it for listening on new connections or
> +    # connecting somewhere else, the module will be loaded. Thus, no need to
> +    # also probe on sockets.
> +    if check_loaded_module('sctp'):
> +        return True
> +
> +    # Basic files from lksctp-tools. This check is enough and checking RPM database
> +    # is an overkill here.
> 
> You can also consume the `leapp.models.InstalledRPM` Model and check against them if the lksctp-tools are installed. But this should be ok, if this is safe enough for your test.

It also covers the case on which it was installed without our
rpm. Unfortunatelly it is something quite possible.

Thanks Vinzenz, I'm working in all the fixes and should have a new PR
by the end of the week.

marceloleitner:
I rebased my master and I guess that confused GH.
Please let me know if I should issue a new PR.
I think I have accommodated all your comments with the changes above, despite the confusion.

marceloleitner:
Had to force-push it, but here we go. Should be good now.

marceloleitner:
Btw feel free to squash it. Kept the history just in case it helps to review it.

vinzenz:
It looks good to me now, please rebase and squash your changes to a single commit. Thanks.

vinzenz:
leapp-ci build

repos/system_upgrade/el7toel8/actors/sctpconfigread/actor.py
```diff
@@ -0,0 +1,71 @@
+from leapp.actors import Actor
+from leapp.models import SCTPConfig, SystemFacts
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+
+
+class SCTPConfigRead(Actor):
+    name = 'sctp_read_status'
+    description = 'This actor reads SCTP status, if it was/is used.'
+    consumes = (SystemFacts,)
```
vinzenz: Please use `ActiveKernelModulesFacts` instead of `SystemFacts` now. We have split the SystemFacts model into separate models. Sorry I did miss that before. The CI just showed me that on the failure.

vinzenz:
leapp-ci build

examon:
leapp-ci build

marceloleitner:
Pushed a new commit, with vinzenz changes applied and a new actor to request the install of kernel-modules-extra. This last, though, is not working and I'm not sure why.

pirat89:
I am going to investigate that

repos/system_upgrade/el7toel8/actors/sctpchecks/actor.py
```diff
@@ -0,0 +1,23 @@
...
+    is collected on SCTPConfig message. If yes, it then produces a RpmTransactionTasks
+    requesting to install the package.
+    """
+    name = 'sctp_checks'
+    consumes = (SCTPConfig,)
+    produces = (RpmTransactionTasks, )
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        for sctpconfig in api.consume(SCTPConfig):
```
pirat89:
```suggestion
        for sctpconfig in self.consume(SCTPConfig):
```

You should use `self.*` (actor's) methods here.

repos/system_upgrade/el7toel8/actors/sctpchecks/actor.py
```diff
@@ -0,0 +1,23 @@
...
+            if sctpconfig.wanted:
+                self.produce(RpmTransactionTasks(to_install='kernel-modules-extra'))
```
pirat89:
```suggestion
                self.produce(RpmTransactionTasks(to_install=['kernel-modules-extra']))
```

vinzenz: This should have caused a Validation Error

pirat89: Yep, it causes, that's why I suggest to use list.

pirat89:
@marceloleitner So I found the reason for the issue (I will report issue about that later. It would be great to catch such thing automatically or at least inform about such cases). Currently the problem is that the actor processing the `RpmTransactionTasks` is processed during the `FactsPhase` phase. I will send PR fixing that.

marceloleitner:
I fixed the 2 issues with SCTPChecks you both reported but was still
scratching my head here. Cool, thanks Petr.

repos/system_upgrade/el7toel8/models/sctpconfigmodel.py
```diff
@@ -0,0 +1,6 @@
+from leapp.models import Model, fields
+from leapp.topics import SCTPConfigTopic
+
+class SCTPConfig(Model):
+    topic = SCTPConfigTopic
+    wanted = fields.Boolean(False)
```
vinzenz:
```suggestion
    wanted = fields.Boolean(default=False)
```
Better to specify the parameter name to guarantee it's the right field which is assigned in case of some refactoring or parameter placement changes.

marceloleitner:
On Tue, Feb 26, 2019 at 12:57:48AM -0800, vinzenz wrote:
> Better to specify the parameter name to guarantee it's the right field which is assigned in case of some refactoring or parameter placement changes.

Okay.

repos/system_upgrade/el7toel8/actors/sctpconfigread/libraries/sctpdlm.py
```diff
@@ -0,0 +1,64 @@
...
+    try:
+        with open(fname, 'r') as fp:
+            cfgs = '[dlm]\n' + fp.read()
+    except (OSError, IOError):
+        return False
+
+    cfg = SafeConfigParser()
+    try:
+        cfg.read_string(cfgs)
+    except:
```
vinzenz: too broad exception specification. IIRC it should be sufficient to check for AttributeError

pirat89:
Congratulations! :congratulations:  :fireworks:  :champagne:

PR #36, opened 2019-01-09
Add PCIDevicesScanner and CheckPCIDevices to verify unsupported devices

artmello:


repos/system_upgrade/el7toel8/actors/checkpcidevices/actor.py
```diff
@@ -0,0 +1,26 @@
...
+    def process(self):
+        for data in self.consume(PCIDevices):
+            for device in data.devices:
+                if 'SCSI' in device.cls and 'LSI Logic' in device.vendor:
+                    self.produce(CheckResult(
+                        severity='Error',
+                        result='Fail',
+                        summary='LSI Logic SCSI Controller is not supported',
+                        details='Kernel driver necessary for LSI Logic SCSI Controller (mpt*) '
+                                'is not available at Red Hat Enterprise Linux 8. Since this '
```
shaded-enmity:
```suggestion
                                'is not available in Red Hat Enterprise Linux 8. Since this '
```

repos/system_upgrade/el7toel8/actors/checkpcidevices/actor.py
```diff
@@ -0,0 +1,26 @@
...
+                                'would prevent controller of working properly upgrade process '
```
shaded-enmity:
```suggestion
                                'would prevent controller from working properly, upgrade process '
```

repos/system_upgrade/el7toel8/actors/checkpcidevices/tests/test_checkpcidevices.py
```diff
@@ -0,0 +1,71 @@
...
+def create_pcidevices(items):
+    attrs = [
+        'slot',
+        'cls',
+        'vendor',
+        'name',
+        'subsystem_vendor',
+        'subsystem_name',
+        'rev',
+        'progif']
```
shaded-enmity:
```suggestion
        'progif'
    ]
```

bocekm:
I think this can be a good candidate for QE testing. You can follow the workflow defined by @AloisMahdal - <gitlab>/oamg/qe-docs/blob/wfver/workflows/verification.md

repos/system_upgrade/el7toel8/actors/pcidevicesscanner/libraries/pcidevicesscanner.py
```diff
@@ -0,0 +1,73 @@
...
+    return ret
+
+
+@aslist
+def get_pci_devices():
+    ''' Get all PCI devices from system '''
+    try:
+        items = call(['lspci', '-mm'])
+    except subprocess.CalledProcessError:
+        items = []
```
bocekm: How does the actor behaves when you get to this line? This function would return empty list. How does `actor.produce()` behave when you feed it with `PCIDevices(devices=[])`? It would be great to have a unit test for that. To be able to unit test it, I would move the produce to the `libraries/pcidevicesscanner.py` as well. Similar to https://github.com/oamg/leapp-repository/pull/5/files#diff-7a7cabedb8bdac2014c9c17631ff259aR24.

repos/system_upgrade/el7toel8/actors/checkpcidevices/actor.py
```diff
@@ -0,0 +1,39 @@
...
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckPCIDevices(Actor):
+    name = 'check_pci_devices'
+    description = """
+        Some PCI devices does not have necessary kernel module available in Red Hat Enterprise Linux 8 what would
+        prevent such device from working properly. At the time of writing this actor, LSI Logic SCSI Controllers
+        suffer of such problem and if it's present on target system, upgrade process should not be executed.
+        Test case table:
```
shaded-enmity: I guess that this doesn't really belong here, we should figure out where to put this kind of data and how to keep them in sync with the actual code.

artmello: @shaded-enmity / @bocekm Until we find out the best place for this sort of data, should we keep it here?

artmello:
Thanks everyone for the reviews. With the latest commit tests are updated and the PR shoudl be ready for another round of reviews.

repos/system_upgrade/el7toel8/actors/checkpcidevices/actor.py
bocekm: Yeah, until we agree on where to put it, let's keep it here. We'll know that by the end of next sprint (RHELLEAPP-974) ;)

repos/system_upgrade/el7toel8/actors/pcidevicesscanner/libraries/pcidevicesscanner.py
```diff
@@ -0,0 +1,71 @@
...
+    ''' Get item at index from list or return a default '''
+    try:
+        ret = l[idx]
+    except IndexError:
+        ret = default
+
+    return ret
+
+
+@aslist
```
bocekm: This is more like a question out of curiosity. Why is it better to have a function as a generator and convert the output to list using a decorator, instead of return a list by the function right away?

artmello: For sometime I imagined that PCIDeviceScanner could be just another function inside SystemFacts actor's library. So I kept the code very close to SystemFacts codebase. Not sure the advantages on this case codewise.

shaded-enmity: I like the succinct form of the generators, it reduces the necessary boilerplate and makes the body of the function more readable, i.e.:
```python
def variant_a():
   to_return = []
   for something in alot_of_things:
       to_return.append(operation(something))
   return to_return
```

Versus

```python
@aslist
def variant_b():
   for something in alot_of_things:
       yield operation(something)
```

shaded-enmity: Oh, and why to convert it to a list straight away? Generators, as useful as they might be, they can sometimes bite your back in unexpected ways, consider:

```python
foo = some_generator()
data = {
  'important': [f for f in foo if f.severity > 100],
  'all': list(foo)
}
```

While the `important` field will have it's proper value, the `all` field will be empty since the generator had already been exhausted by that point.

repos/system_upgrade/el7toel8/actors/checkpcidevices/tests/test_checkpcidevices.py
```diff
@@ -0,0 +1,61 @@
+from leapp.models import CheckResult, PCIDevices, PCIDevice
+from leapp.snactor.fixture import current_actor_context
```
bocekm: Remove this line. `current_actor_context` fixture is injected under the hood.

repos/system_upgrade/el7toel8/actors/pcidevicesscanner/tests/test_pcidevicesscanner.py
```diff
@@ -0,0 +1,122 @@
+from leapp.snactor.fixture import current_actor_context, current_actor_libraries
```
bocekm: `current_actor_libraries` is deprecated and `current_actor_context` is injected under the hood. You can safely remove this line.
See example in https://leapp.readthedocs.io/en/latest/unit-testing.html#testing-private-actor-library-unit-tests

repos/system_upgrade/el7toel8/actors/pcidevicesscanner/tests/test_pcidevicesscanner.py
```diff
@@ -0,0 +1,122 @@
...
+from leapp.models import PCIDevices, PCIDevice
+
+
+def test_parse_pci_devices(current_actor_libraries):
+    from leapp.libraries.actor.pcidevicesscanner import parse_pci_devices
```
bocekm: You can import the library in the top of the file. It doesn't need to be imported within a test function anymore.

artmello: At least on my environment moving the import to top of the file is failing. Any idea what cna be wrong?

```
(leapp-devel) [vagrant@localhost pcidevicesscanner]$ pwd
/home/vagrant/work/leapp-repository/repos/system_upgrade/el7toel8/actors/pcidevicesscanner
(leapp-devel) [vagrant@localhost pcidevicesscanner]$ ls -la libraries/
total 8
drwxrwxr-x. 2 vagrant vagrant   63 Jan 17 22:34 .
drwxrwxr-x. 4 vagrant vagrant   85 Jan 17 22:22 ..
-rw-rw-r--. 1 vagrant vagrant 1801 Jan 17 22:32 pcidevicesscanner.py
-rw-rw-r--. 1 vagrant vagrant 3117 Jan 17 22:34 pcidevicesscanner.pyc
(leapp-devel) [vagrant@localhost pcidevicesscanner]$ git diff tests/test_pcidevicesscanner.py
diff --git a/repos/system_upgrade/el7toel8/actors/pcidevicesscanner/tests/test_pcidevicesscanner.py b/repos/system_upgrade/el7toel8/actors/pcidevicesscanner/tests/test_pc
index c9420e1..217f67e 100644
--- a/repos/system_upgrade/el7toel8/actors/pcidevicesscanner/tests/test_pcidevicesscanner.py
+++ b/repos/system_upgrade/el7toel8/actors/pcidevicesscanner/tests/test_pcidevicesscanner.py
@@ -1,5 +1,6 @@
 from leapp.snactor.fixture import current_actor_context, current_actor_libraries
 from leapp.models import PCIDevices, PCIDevice
+from leapp.libraries.actor import pcidevicesscanner
 
 
 def test_parse_pci_devices(current_actor_libraries):
(leapp-devel) [vagrant@localhost pcidevicesscanner]$ cd tests/ && pytest
========================================================================== test session starts ===========================================================================
platform linux2 -- Python 2.7.5, pytest-4.1.1, py-1.7.0, pluggy-0.8.1
rootdir: /home/vagrant/work/leapp-repository/repos/system_upgrade/el7toel8/actors/pcidevicesscanner/tests, inifile:
plugins: leapp-0.4.0
collected 0 items / 1 errors                                                                                                                                             

================================================================================= ERRORS =================================================================================
_______________________________________________________________ ERROR collecting test_pcidevicesscanner.py _______________________________________________________________
ImportError while importing test module '/home/vagrant/work/leapp-repository/repos/system_upgrade/el7toel8/actors/pcidevicesscanner/tests/test_pcidevicesscanner.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/home/vagrant/devel/leapp-devel/lib/python2.7/site-packages/six.py:709: in exec_
    exec("""exec _code_ in _globs_, _locs_""")
test_pcidevicesscanner.py:3: in <module>
    from leapp.libraries.actor import pcidevicesscanner
E   ImportError: cannot import name pcidevicesscanner
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
======================================================================== 1 error in 0.09 seconds =========================================================================
(leapp-devel) [vagrant@localhost tests]$ 
```

artmello: @bocekm Moved the import to the top of the file and tests worked as expected now. So this should be resolved on latest commit :)

PR #37, opened 2019-01-10
Exclude test file names from naming convention

bocekm:
We require using lowercase without underscores for filenames but tests need to be named `test_*.py` or `*_test.py`.


pirat89:
leapp-ci build

PR #40, opened 2019-01-13
Use the right copr repo for downloading pes data

vojtechsokol:


vojtechsokol:
@pirat89 Yes, totally true, I didn't realized that, so closing this PR.

PR #42, opened 2019-01-15
remove debug from cmdline when adding boot entry

jmikovic:


jmikovic:
leapp-ci build

pirat89:
leapp-ci build

PR #44, opened 2019-01-15
Implement usage of RHEL8 DNF/RPM stack

vinzenz:
This patch introduces the ability to run DNF from RHEL8 from a RHEL7
system to perform the upgrade. This will allow us to support all new RPM
features that are not supported by the RHEL7 stack.

Additional changes:
- Facts Phase:
    - Userspace generation 
       - Target RHSM info gathering (switch product certs, retrieve enabled / available repos, SKUs...)
    - Source RHSM info gathering
    - Feeding modules needed for dracut + el8 dependencies to install  for upgrade initramdisk generation
    - SetupTargetRepos moved from ChecksPhase to FactsPhase, check part split as separate actor to ChecksPhase
- Checks Phase:
    - Check for skipped repos
    - Perform transaction check
    - CheckRedHatSubscriptionManagerSKU - Checks for attached SKUs on the current subscription 
- InterimPreparationPhase:
    - InitramDiskGenerator
        - Install packages needed for the initram disk
        - Generate initram disk 

Added libraries:
- Mounting:
    - Handles multiple mounting scenarios:
        - Overlayfs
        - Bind mounts (Mounting directories into other directories)
        - Loop mounts (disk image for XFS)
        - Typed mounts (procfs mount for example)
    - Abstraction for calling processes in an isolated manner
        - systemd-nspawn (as we used to before)
        - chrooted (for package installation when leapp already runs with in systemd-nspawn in the initramdisk)
 - DNF Plugin
    Various tasks that are executed on top of DNF + plugin or not
 - guards
     - Moved code of guards from previous dnfpreupgradetransaction to a common library
 - Overlaygen 
     - Tasks around generating to the overlayfs needed for el7toel8 upgrades
 - RHSM
    - Subscription manager tasks used by multiple actors, for gathering information and modifying things like the set release  etc
 - Utils a list of functions that were used all over the place by the code
    


vojtechsokol:
leapp-ci build

vojtechsokol:
leapp-ci build

repos/system_upgrade/el7toel8/actors/dnfshellrpmupgrade/actor.py
```diff
@@ -44,55 +27,11 @@ def process(self):
...
-        debugsolver = True if os.environ.get('LEAPP_DEBUG', '0') == '1' else False
-
-        shutil.copyfile(
-            '/etc/yum.repos.d/redhat.repo.upgrade',
-            '/etc/yum.repos.d/redhat.repo'
-        )
-
-        # FIXME: that's ugly hack, we should get info which file remove and
-        # + do it more nicely..
-        cmd = ['rm', '-f', '/etc/pki/product/69.pem']
```
vinzenz: @pirat89  Feels strange, this shouldn't be necessary - /etc/pki/product-default/69.pem is installed by the release package and should get removed during the transaction.
So this seems off, that we have to add this for /etc/pki/product/69.pem. What's the reason for this being in product instead of product-default?

If we know that then we can check for this maybe. Anyway I have left this out for this moment until I get an answer.

pirat89: in short: `/etc/pki/product-default` is directory where we store the certificate of the system (RHEL product). IOW, the directory is sort of specific to RHEL systems. OTOH, `/etc/pki/product` is stadard path used by many other products (other RHEL products, 3rd-party products, ...). In case of RHEL, when you register a system, the certificate is copied to `/etc/pki/product` as there should be stored (let's say) "active" certificates. RHSM can store there additional certificates - based on SKUs you use. In case of update, the certificate in `/etc/pki/product-default` is updated, but not the one in `/etc/pki/product/` and the one stored here has higher priority. So you would not find the RHEL 8 system. It is possible that behaviour can be different, but this is one how it should behave and we've got recommendation to remove this one as well and ensure the new one is copied to the directory - that's something what we do not do yet if I am right, and we should.

vinzenz: @pirat89 Ok thanks - Restored

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -0,0 +1,127 @@
...
+
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.actor import constants
+from leapp.libraries.common import dnfplugin, mounting, overlaygen, rhsm, utils
+from leapp.libraries.stdlib import api, run
+from leapp.models import (OSReleaseFacts, RequiredTargetUserspacePackages,
+                          SourceRHSMInfo, TargetUserSpaceInfo)
+
+
+def _logger():
```
fernflower: Same question why do we need a function here

repos/system_upgrade/el7toel8/libraries/overlaygen.py
```diff
@@ -0,0 +1,142 @@
...
+import os
+import shutil
+
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.common import mounting, utils
+from leapp.libraries.stdlib import CalledProcessError, api, run
+from leapp.models import XFSPresence
+
+
+def _logger():
```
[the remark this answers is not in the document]
vinzenz: Because at import time, current_logger is not yet reachable - The actor that imports this library, has not been instantiated

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
vinzenz: Same answer

repos/system_upgrade/el7toel8/actors/checkskippedrepositories/actor.py
```diff
@@ -0,0 +1,44 @@
...
+            repos.update(message.repos)
+            packages.update(message.packages)
+
+        if repos:
+            title = 'Some enabled RPM repositories are unknown to Leapp'
+            summary_data = []
+            summary_data.append('The following repositories with Red Hat-signed packages are unknown to Leapp:')
+            summary_data.extend(['- ' + r for r in repos])
+            summary_data.append('And the following packages installed from those repositories may not be upgraded:')
+            summary_data.extend(['- ' + p for p in packages])
```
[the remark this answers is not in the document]
vinzenz: Right, this was just a cut and paste from a different place, but yeah

repos/system_upgrade/el7toel8/actors/scansubscriptionmanagerinfo/libraries/scanrhsm.py
```diff
@@ -0,0 +1,12 @@
+from leapp.models import SourceRHSMInfo
+from leapp.libraries.common import rhsm
+from leapp.libraries.common.mounting import NotIsolatedActions
```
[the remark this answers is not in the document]
vinzenz: Yeah, will be helpful I guess :D I missed that

repos/system_upgrade/el7toel8/actors/checkskippedrepositories/tests/test_checkskippedrepos.py
```diff
@@ -0,0 +1,70 @@
+from leapp.models import Report, SkippedRepositories
```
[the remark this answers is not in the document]
vinzenz: See, forgot to do that then, yes, this was the plan

repos/system_upgrade/el7toel8/actors/initramdiskgenerator/files/generate-initram.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
vinzenz: /\t gives me nothing in vim

repos/system_upgrade/el7toel8/libraries/utils.py
```diff
@@ -0,0 +1,129 @@
...
+
+def _logger():
+    return api.current_logger()
+
+
+def makedirs(path, mode=0o777, exists_ok=True):
+    mounting._makedirs(path=path, mode=mode, exists_ok=exists_ok)
+
+
+def apply_yum_workaround(context=None):
```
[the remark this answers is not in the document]
vinzenz: The fact that the script that is part of the leapp-repository package was not found - And this shouldn't happen unless the package installation got corrupted.

repos/system_upgrade/el7toel8/actors/initramdiskgenerator/files/generate-initram.sh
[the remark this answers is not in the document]
vinzenz: Trailing spaces removed from some lines

repos/system_upgrade/el7toel8/libraries/overlaygen.py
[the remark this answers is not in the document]
vinzenz: Ok well current_logger itself is reachable, but the instance of the logger it would return is not.
It would default to some fallback logger. I rather could import api.current_logger as _logger

repos/system_upgrade/el7toel8/actors/initramdiskgenerator/libraries/initramgen.py
```diff
@@ -0,0 +1,97 @@
...
+from leapp.libraries.common import dnfplugin, mounting
+from leapp.libraries.stdlib import api
+from leapp.models import BootContent, RequiredUpgradeInitramPackages, TargetUserSpaceInfo, UpgradeDracutModule
+
+INITRAM_GEN_SCRIPT_NAME = 'generate-initram.sh'
+
+
+def _reinstall_leapp_repository_hint():
+    """
+    Convenience function for creating a detail for StopActorExecutionError with a hint to reinstall the
```
pirat89: Keep short title line and body.

repos/system_upgrade/el7toel8/actors/initramdiskgenerator/libraries/initramgen.py
```diff
@@ -0,0 +1,97 @@
...
+        'LEAPP_ADD_DRACUT_MODULES={modules} {cmd}'.format(
+            modules=','.join([mod.name for mod in modules]),
+            cmd=os.path.join('/', INITRAM_GEN_SCRIPT_NAME))
+    ])
+    copy_boot_files(context)
+
+
+def copy_boot_files(context):
+    """
+    Function to copy the generated initram and corresponding kernel to /boot - Additionally produces a BootContent
```
pirat89: Keep shorter title and body.

repos/system_upgrade/el7toel8/libraries/dnfplugin.py
```diff
@@ -0,0 +1,199 @@
...
+DNF_PLUGIN_NAME = 'rhel_upgrade.py'
+DNF_PLUGIN_PATH = os.path.join('/lib/python3.6/site-packages/dnf-plugins', DNF_PLUGIN_NAME)
+DNF_PLUGIN_DATA_NAME = 'dnf-plugin-data.txt'
+DNF_PLUGIN_DATA_PATH = os.path.join('/var/lib/leapp', DNF_PLUGIN_DATA_NAME)
+DNF_PLUGIN_DATA_LOG_PATH = os.path.join('/var/log/leapp', DNF_PLUGIN_DATA_NAME)
+DNF_DEBUG_DATA_PATH = '/var/log/leapp/dnf-debugdata/'
+
+
+def install(target_basedir):
+    """ Installs our plugin to the DNF plugins """
```
pirat89: Use please rather this:
```suggestion
    """
    Installs our plugin to the DNF plugins.
    """
```

repos/system_upgrade/el7toel8/actors/initramdiskgenerator/libraries/initramgen.py
```diff
@@ -0,0 +1,97 @@
...
+    leapp-repository package
+    """
+    return {
+        'hint': 'Try to reinstall the `leapp-repository` package.'
+    }
+
+
+def copy_dracut_modules(context, modules):
+    """
+    Copies our dracut modules into the target userspace.
```
pirat89:
```suggestion
    Copies required dracut modules into the target userspace.
```

repos/system_upgrade/el7toel8/libraries/dnfplugin.py
```diff
@@ -0,0 +1,199 @@
...
+    """
+    # get list of repo IDs of target repositories that should be used for upgrade
+    tasks = next(api.consume(FilteredRpmTransactionTasks), FilteredRpmTransactionTasks())
+
+    data = {
+        'pkgs_info': {
+            'local_rpms': [os.path.join('/installroot', pkg.lstrip('/')) for pkg in tasks.local_rpms],
+            'to_install': [pkg for pkg in tasks.to_install],
+            'to_remove': [pkg for pkg in tasks.to_remove],
+            #            'to_upgrade': [pkg for pkg in tasks.to_upgrade]
```
pirat89: Why this one is commented out? (nowadays, `to_upgrade` is `to_keep`)

repos/system_upgrade/el7toel8/actors/checkrhsmsku/tests/test_rhsmsku.py
```diff
@@ -0,0 +1,28 @@
...
+
+def test_sku_report_has_no_skus(monkeypatch, current_actor_context):
+    with monkeypatch.context() as context:
+        context.setattr(rhsm, 'skip_rhsm', lambda: True)
+        current_actor_context.feed(SourceRHSMInfo(attached_skus=[]))
+        current_actor_context.run()
+        reports = list(current_actor_context.consume(Report))
+        assert reports and len(reports) == 1
+        assert reports[0].severity == 'high'
+        assert reports[0].title == 'The system is not registered or subscribed.'
```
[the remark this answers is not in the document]
vinzenz: Yes they are, and interestingly it doesn't fail

repos/system_upgrade/el7toel8/actors/checkskippedrepositories/actor.py
```diff
@@ -0,0 +1,44 @@
...
+
+        for message in self.consume(SkippedRepositories):
+            repos.update(message.repos)
+            packages.update(message.packages)
+
+        if repos:
+            title = 'Some enabled RPM repositories are unknown to Leapp'
+            summary_data = []
+            summary_data.append('The following repositories with Red Hat-signed packages are unknown to Leapp:')
+            summary_data.extend(['- {}'.format(r) for r in repos])
```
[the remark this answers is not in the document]
vinzenz: Well I didn't implement this, I just kept the format as it was before - I even did the format now. I feel this starts being unrelated to the actual PR. Unification of output should be probably part of a different PR. 
If you check setuptargetrepos report_skipped_repos function you'll see that it came from there.

repos/system_upgrade/el7toel8/actors/checkskippedrepositories/actor.py
```diff
@@ -0,0 +1,44 @@
...
+            summary_data.append('And the following packages installed from those repositories may not be upgraded:')
```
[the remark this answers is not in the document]
vinzenz: Also not changed from the original

repos/system_upgrade/el7toel8/actors/checkskippedrepositories/actor.py
```diff
@@ -0,0 +1,44 @@
...
+            summary_data.extend(['- {}'.format(p) for p in packages])
```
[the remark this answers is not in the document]
vinzenz: Same

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/libraries/modscan.py
```diff
@@ -0,0 +1,54 @@
...
+    'openssh-clients',
+    'plymouth',
+    'rpcbind',
+    'systemd-container',
+    'tar'
+]
+
+
+def _create_dracut_modules():
+    dracut_base_path = api.get_actor_folder_path('dracut')
```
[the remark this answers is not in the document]
vinzenz: files are only bundled with the actor and not really private - They cannot be discovered by other actors but by specifying the fullpath in the message, their content is accessible. I don't think we need to put them globally since there's no sense for making those files/folders discoverable by all actors.

repos/system_upgrade/el7toel8/libraries/dnfplugin.py
vinzenz: There is a pending PR with this and I already had it prepared, to_keep isn't used by the DNF plugin, this was to keep me reminded that this has to be updated in case 226 gets merged before.

https://github.com/oamg/leapp-repository/pull/226

pirat89: Would you put there at least FIXME/TODO msg so in case this will be merged, we will be sure nobody forget about that? Possibly this will be merged before #226

repos/system_upgrade/el7toel8/libraries/mounting.py
```diff
@@ -0,0 +1,384 @@
...
+
+    def makedirs(self, path, mode=0o777, exists_ok=True):
+        """
+        Creates the whole path recursively for any missing part.
+        """
+        _makedirs(path=self.full_path(path), mode=mode, exists_ok=exists_ok)
+
+    def copytree_to(self, src, dst):
+        """
+        Recursively copy an entire directory tree rooted at src. The destination directory,
```
pirat89: please, keep structure:
'''
title one line

..body..
'''

pirat89: (everywhere)

repos/system_upgrade/el7toel8/actors/initramdiskgenerator/libraries/initramgen.py
```diff
@@ -0,0 +1,97 @@
...
+        except shutil.Error as e:
+            api.current_logger().error('Failed to copy dracut module "{name}" from "{source}" to "{target}"'.format(
+                name=module.name, source=module.module_path, target=context.full_path('/dracut')), exc_info=True)
+            raise StopActorExecutionError(
+                message='Failed to install dracut modules required in the initram. Error: {}'.format(e.message)
+            )
+
+
+def _get_dracut_modules():
+    return list(api.consume(UpgradeDracutModule))
```
pirat89: this is wrong. You cannot cosume stuff like that in shared libraries - at least, you cannot do that without checks that the actor is consuming this type of message. Otherwise someone would use the library and nothing will happen because the list of consumed messages will be empty. This is problem on several other places (all consumes in shared libraries)

vinzenz: This particular one is actually in a private library.

pirat89: we should even consider putting some guidance about that into the best practices or somewhere

pirat89: @vinzenz ah. yes, I marked the wrong one. Sorry, but the problem is in two shared libraries.

repos/system_upgrade/el7toel8/libraries/mounting.py
vinzenz: I will fix those in a follow up to not stall this PR any longer

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -0,0 +1,168 @@
...
+            mounts_dir=constants.MOUNTS_DIR,
+            scratch_dir=constants.SCRATCH_DIR,
+            xfs_present=xfs_present) as overlay:
+        with overlay.nspawn() as context:
+            with rhsm.switched_certificate(context, rhsm_info, prod_cert_path) as target_rhsm_info:
+                _logger().debug("Target RHSM Info: SKUs: {skus} Repositories: {repos}".format(
+                    repos=target_rhsm_info.enabled_repos,
+                    skus=rhsm_info.attached_skus if rhsm_info else []
+                ))
+                target_repoids = rhsm.gather_target_repositories(target_rhsm_info)
```
pirat89:
```suggestion
                target_repoids = gather_target_repositories(target_rhsm_info)
```

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -0,0 +1,168 @@
...
+        raise StopActorExecutionError(
+            message='Failed to prepare environment for package download while creating directories.',
+            details={
+                'hint': 'Please ensure that {directory} is empty and modifiable.'.format(
+                    directory=target_userspace)
+            }
+        )
+
+
+def gather_target_repositories(target_repos, target_rhsm_info):
```
pirat89:
```suggestion
def gather_target_repositories(target_rhsm_info):
```

bocekm:
I ran the upgrade without being subscribed and I'm getting a traceback:
```
2019-06-26 13:17:30.903 INFO     PID: 21332 leapp.workflow.FactsCollection: Executing actor target_userspace_creator 
Process Process-143:
    self.run()
  File "/usr/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 63, in _do_run
    target_actor(logger=logger, messaging=messaging, config_model=config_model).run(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/leapp/actors/__init__.py", line 320, in run
    self.process(*args)
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/targetuserspacecreator/actor.py", line 25, in process
    userspacegen.perform()
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py", line 140, in perform
    api.log.warn("Could not receive RHSM information - Is this system registered?")
AttributeError: 'module' object has no attribute 'log'

=========================================================================================================
Actor target_userspace_creator unexpectedly terminated with exit code: 1 - Please check the above details
=========================================================================================================
```

bocekm:
To be exact, I executed it on the latest RHEL 7.7 compose, if that can make any difference in this case.

bocekm:
@vinzenz, where's the https://github.com/oamg/leapp-repository/blob/master/sources/dracut used? It's a symlink to the removed `prepareupgradetransaction` actor.

vinzenz:
It should be obsolete now

Issue #32, opened 2019-01-03
system_facts actor fails when no baseurl is present in repo file


vojtechsokol:
When no `baseurl` is present in repo file (e.g. repo file for epel7 uses `metalink` instead), the `system_facts` fails.

**Reproducer**
```
yum install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm

leapp upgrade

2019-01-03 16:50:32.507 INFO     PID: 4256 leapp: Logging has been initialized
2019-01-03 16:50:32.541 INFO     PID: 4256 leapp.repository.common: A new repository 'common' is initialized at /etc/leapp/repos.d/common
2019-01-03 16:50:32.570 INFO     PID: 4256 leapp.repository.system_upgrade_el7toel8: A new repository 'system_upgrade_el7toel8' is initialized at /etc/leapp/repos.d/system_upgrade/el7toel8
2019-01-03 16:50:32.774 INFO     PID: 4256 leapp.workflow: Starting workflow execution: InplaceUpgrade - ID: 66c51e03-1134-445f-9811-0223be43c173
2019-01-03 16:50:32.796 INFO     PID: 4256 leapp.workflow: Starting phase Facts collection
2019-01-03 16:50:32.820 INFO     PID: 4256 leapp.workflow.Facts collection: Starting stage Before of phase Facts collection
2019-01-03 16:50:32.870 INFO     PID: 4256 leapp.workflow.Facts collection: Starting stage Main of phase Facts collection
2019-01-03 16:50:32.896 INFO     PID: 4256 leapp.workflow.Facts collection: Executing actor rpm_scanner
2019-01-03 16:50:33.452 INFO     PID: 4256 leapp.workflow.Facts collection: Executing actor system_facts
Process Process-34:
Traceback (most recent call last):
  File "/usr/lib64/python2.7/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/lib64/python2.7/multiprocessing/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 59, in _do_run
    target_actor(logger=logger, messaging=messaging).run(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/leapp/actors/__init__.py", line 121, in run
    self.process(*args)
[...]

PR #45, opened 2019-01-15
Make "baseurl" field optional and add "metalink" field

Rezney:
Currently "baseurl" field is mandatory and when we parse the field
commented we try to add it to "additional_fields" string  as "#baseurl"
and thus omitting the wanted value.

We also add "metalink" field which now ends in "additional_fields"
string too.

Without this patch we fails here:

```python  
File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py", line 205, in get_repositories
    yield Repositories(file=repo, data=_parse(repo))
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py", line 27, in inner
    return list(f(*args, **kwargs))
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py", line 198, in _parse
    yield RepositoryData(**prepared)
  File "/usr/lib/python2.7/site-packages/leapp/models/__init__.py", line 86, in __init__
    getattr(defined_fields[field], init_method)(kwargs, field, self)
  File "/usr/lib/python2.7/site-packages/leapp/models/fields/__init__.py", line 110, in from_initialization
    self._validate_model_value(value=source_value, name=name)
  File "/usr/lib/python2.7/site-packages/leapp/models/fields/__init__.py", line 166, in _validate_model_value
    super(BuiltinField, self)._validate_model_value(value, name)
  File "/usr/lib/python2.7/site-packages/leapp/models/fields/__init__.py", line 62, in _validate_model_value
    raise ModelViolationError('The value of "{name}" field is None, but this is not allowed'.format(name=name))
ModelViolationError: The value of "baseurl" field is None, but this is not allowed
```

### DIFF
--- a/repos/system_upgrade/el7toel8/models/systemfacts.py
+++ b/repos/system_upgrade/el7toel8/models/systemfacts.py
@@ -23,7 +23,9 @@ class RepositoryData(Model):
     topic = SystemInfoTopic
 
     name = fields.String()
-    baseurl = fields.String()
+    baseurl = fields.Nullable(fields.String())
+    metalink = fields.Nullable(fields.String())
+    mirrorlist = fields.Nullable(fields.String())
     enabled = fields.Boolean(default=True)
     additional_fields = fields.Nullable(fields.String())
 


repos/system_upgrade/el7toel8/models/systemfacts.py
```diff
@@ -23,7 +23,8 @@ class RepositoryData(Model):
     topic = SystemInfoTopic
 
     name = fields.String()
-    baseurl = fields.String()
+    baseurl = fields.Nullable(fields.String())
+    metalink = fields.Nullable(fields.String())
```
pirat89: I see here problem, that only one of them (`baseurl` or `metalink`) can be set and the other one has to be `None`. In addition, there can be even `mirrorlist` regarding the [documentation](https://dnf.readthedocs.io/en/latest/conf_ref.html#mirrorlist-label). So at least this one is still missing when we are fixing the model. I am going to test behaviour when more then one of the parameters are set.

vojtechsokol:
@Rezney Please mark #32 as resovled when merging this PR.

pirat89:
As @vojtechsokol wrote, it would be better to put line like:
  Resolves: #32 
into the commit message, so the issue will be connected automatically and easier to find just using git command. But it is not requirement for merge.
W8 for results of tests as we have broken CI.

Rezney:
@vojtechsokol @pirat89 makes sense. Actually I did not know about the already opened issue. Firstly I wanted to create a new one (and thus checking for duplicates) but then I just went with the fix as it is a few-liner...

vojtechsokol:
Sorry for the broken CI, typical issues with Jenkins in the cloud :)

vojtechsokol:
leapp-ci build

pirat89:
@vojtechsokol thanks for fixing it :)

pirat89:
I see know, that I have forgotten to request tests! Next time guys :)

PR #48, opened 2019-01-16
Add NetworkManager actors

bengal:
This is my first attempt at writing Leapp actors for NM. Any feedback is appreciated, thanks!


bengal:
I tested a full upgrade using a vagrant box and the NetworkManagerUpdateClientId actor fails with the following error:
```
 2019-01-16 18:50:32.944 INFO     PID: 1 leapp.workflow.Applications: Executing actor network_manager_update_client_id
 Process Process-47:
     sys.stdin.close()
   File "/usr/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 56, in _do_run
     definition.load()
   File "/usr/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 139, in load
     self._module = importer.find_module(name).load_module(name)
   File "/usr/lib64/python2.7/pkgutil.py", line 246, in load_module
     self.file.close()
   File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/networkmanagerupdateclientid/actor.py", line 5, in <module>
     import gi
   File "/usr/lib64/python2.7/site-packages/gi/__init__.py", line 37, in <module>
     from . import _gi
 ImportError: /lib64/libc.so.6: version `GLIBC_2.25' not found (required by /lib64/libmount.so.1)
 =================================================================================================================
 Actor network_manager_update_client_id unexpectedly terminated with exit code: 1 - Please check the above details
 =================================================================================================================
```
I don't have any clue about the missing symbol error, the C library does contain the symbol:
```
# nm /lib64/libc.so.6 | grep GLIBC_2.25
0000000000000000 A GLIBC_2.25
```
Does anybody know what the problem is?

repos/system_upgrade/el7toel8/actors/networkmanagerupdateservice/actor.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def unit_enabled(self, name):
+        try:
+            ret_list = self.call(['systemctl', 'is-enabled', name])
+            enabled = ret_list[0] == 'enabled'
+        except CalledProcessError:
+            enabled = False
+        return enabled
+
+    def call(self, args):
```
vinzenz: we actually have a new generic library call for this now:

`from leapp.libraries.stdlib import call`

We need to do some capturing of executed processes, timings and outputs for logging purposes within our audit. Also for check_call/call cases to redirect the output. But this is still WIP, anyway please switch to the usage of the call function, but keep the catching for the CalledProcessError for now, this has not been changed yet.

repos/system_upgrade/el7toel8/actors/networkmanagerupdateservice/actor.py
```diff
@@ -0,0 +1,56 @@
...
+    tags = (ApplicationsPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        nm_enabled = self.unit_enabled('NetworkManager.service')
+        nmwo_enabled = self.unit_enabled('NetworkManager-wait-online.service')
+        self.log_services_state('initial', nm_enabled, nmwo_enabled)
+
+        if not nm_enabled and nmwo_enabled:
+            self.log.info('Disabling NetworkManager-wait-online.service')
+            self.call(['systemctl', 'disable', 'NetworkManager-wait-online.service'])
```
vinzenz: This might raise, therefore please guard this from escaping - Non caught exceptions are considered a bug. There are only certain allowed exceptions like StopActorExecution and StopActorExecutionError which will gracefully be handled by the framework and handled accordingly.

repos/system_upgrade/el7toel8/actors/networkmanagerupdateservice/actor.py
```diff
@@ -0,0 +1,56 @@
+from leapp.actors import Actor
+from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
+
+import os
+import six
+import subprocess
+from subprocess import CalledProcessError
+
+# On RHEL7 if the NetworkManager service was disabled and
```
vinzenz: Again good description, probably most suitable for the description.

repos/system_upgrade/el7toel8/actors/networkmanagerupdateservice/actor.py
```diff
@@ -0,0 +1,56 @@
+from leapp.actors import Actor
+from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
+
+import os
```
vinzenz: python standard library imports are supposed to be before library imports

repos/system_upgrade/el7toel8/actors/networkmanagerreadconfig/actor.py
```diff
@@ -0,0 +1,42 @@
...
+    produces = (NetworkManagerConfig,)
+    tags = (IPUWorkflowTag, FactsPhaseTag,)
+
+    def process(self):
+        try:
+            nm_config = NetworkManagerConfig()
+            # Use 'NM --print-config' to read the configurationo so
+            # that the main configuration file and other files in
+            # various directories get merged in the right way.
+            subp = subprocess.Popen(('/usr/sbin/NetworkManager', '--print-config'), stdout=subprocess.PIPE)
```
vinzenz: Please use `from leapp.libraries.stdlib import call` instead, no need for this and it's decoded already to utf-8

repos/system_upgrade/el7toel8/actors/networkmanagerreadconfig/actor.py
```diff
@@ -0,0 +1,42 @@
+from leapp.actors import Actor
+from leapp.models import NetworkManagerConfig
+from leapp.tags import IPUWorkflowTag, FactsPhaseTag
+import subprocess
```
vinzenz: please move python standard library imports on top and sorted please.

See: https://www.python.org/dev/peps/pep-0008/#imports

> Imports should be grouped in the following order:
> 
> * Standard library imports.
> * Related third party imports.
> * Local application/library specific imports.
>
> You should put a blank line between each group of imports.

repos/system_upgrade/el7toel8/actors/networkmanagerreadconfig/actor.py
```diff
@@ -0,0 +1,42 @@
...
+        try:
+            nm_config = NetworkManagerConfig()
+            # Use 'NM --print-config' to read the configurationo so
+            # that the main configuration file and other files in
+            # various directories get merged in the right way.
+            subp = subprocess.Popen(('/usr/sbin/NetworkManager', '--print-config'), stdout=subprocess.PIPE)
+            conf, err = subp.communicate()
+            parser = ConfigParser()
+            parser.readfp(io.StringIO(conf.decode('utf-8')))
+        except Exception as e:
```
vinzenz: This is too wide, please try to restrict the exceptions handling first of all around the place where it could happen and secondly as restrictive as possible. e.g around the call there could be OSError, CalledProcessError. 

For the readfp part it should be ParsingError

repos/system_upgrade/el7toel8/actors/networkmanagerupdateclientid/actor.py
```diff
@@ -0,0 +1,58 @@
...
+                        self.log.info('  * client-id : {}'.format(client_id))
+                        if not self.is_hexstring(client_id):
+                            new_client_id = ':'.join(hex(ord(x))[2:] for x in client_id)
+                            self.log.info('  * new client-id : {}'.format(new_client_id))
+                            s_ip4.set_property(NM.SETTING_IP4_CONFIG_DHCP_CLIENT_ID, new_client_id)
+                            if not c.commit_changes(True, None):
+                                self.log.warning('Failed to update client-id for connection {}'.format(c.get_uuid()))
+            break
+
+    def is_hexstring(self, s):
```
vinzenz: methods that don't use self (which is this case) are usually preferably functions or decorated with @staticmethod

repos/system_upgrade/el7toel8/actors/networkmanagerupdateclientid/actor.py
```diff
@@ -0,0 +1,58 @@
+from leapp.actors import Actor
+from leapp.models import NetworkManagerConfig
+from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
+
+import gi
```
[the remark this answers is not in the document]
vinzenz: It's quite difficult to say what's the exact problem here. However this already causes a failure during the discovery phase in your testing. I'd have to check what's going on there. But considering that RHEL8 (where this actor is going to run in the end) system python will be 3.6 and we are running in a 2.7 process, you might have to put the whole thing into a separate python file that you start from this one. 
Let's look at this a bit later and address the other cosmetics first and I will check what's going on with that symbol problem.

repos/system_upgrade/el7toel8/topics/networkmanagerconfigtopic.py
```diff
@@ -0,0 +1,5 @@
+from leapp.topics import Topic
+
+
+class NetworkManagerConfigTopic(Topic):
+    name = 'network_manager_config_topic'
```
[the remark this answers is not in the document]
bengal: Ok. I removed the topic altogether and used the SystemInfoTopic instead.

repos/system_upgrade/el7toel8/actors/networkmanagerreadconfig/actor.py
bengal: ok

repos/system_upgrade/el7toel8/actors/networkmanagerupdateclientid/actor.py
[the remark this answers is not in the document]
bengal: I moved the code to a separate python file, hopefully this will solve the problem (haven't tested yet).

repos/system_upgrade/el7toel8/actors/networkmanagerreadconfig/actor.py
bengal: Ok.

bengal:
> In a general note, we prefer using the library approach for the implementations that means, that we try to move all the implementation in a separate module under the actor's libraries folder and import that module with `from leapp.libraries.actor import mymodulename`
> 
> That allows to write better testable code which brings me to the second part, that we would like to have component and unit tests for the code.

Hi, since all these actors interact with the system, it's not clear to me how to unit-test them. Do you have any suggestions? 

About moving code into a library, actors do very few things so perhaps the only function I could move is the unit_enabled() of NetworkManagerUpdateService. What do you think?

Regarding the other points you raised, I have hopefully addressed them. Thanks for the review.

PR #49, opened 2019-01-22
Add metapackage for dependencies

pirat89:
Currently we are not able to use python dependencies correctly to
    ensure the RPM will not be removed during the upgrade of the system.
    As a workround, put all such system dependencies into the
    leapp-repository-deps metapackage which is supposed to be replaced
    later (another commit(s)) during the upgrade transaction by the
    another metapackage containing dependencies satisfying dependencies
    of target (RHEL 8) system.
    
    To reflect way of the leapp framework how this should be resolved,
    the leapp-repository rpm should create dependency on specific
    capability:
      leapp-repository-dependencies
    
    which has to be provided by such meta package. Any time the
    dependencies are changed, the capability number has to be incremented
    by one. Such change has to be reflected by both (for RHEL7 and RHEL8)
    metapackages.

    Handle leapp & leapp-repository dependencies during upgrade
    
    Add new SPEC file for RHEL 8 metapackes that handle dependencies of
    leapp and leapp-repository packages. These replace original
    metapackages produced on RHEL 7 system.
    
    The Makefile has been modified to build those packages during
    processing of the `source` target and create new source consumed
    and required by original SPEC file. Such packages are bundled inside
    the leapp-repository RPM and they are installed automatically
    during the IPU.
    
    For the building of RHEL 8 metapackages has to be used separate
    rpm repository, otherwise packages are viewable by DNF inside
    repository and do cause installation issue of leapp and
    leapp-repository packages as these metapackages cannot be installed
    on RHEL 7 system.

### DIFF
--- a/packaging/leapp-repository.spec
+++ b/packaging/leapp-repository.spec
@@ -12,12 +12,13 @@ URL:        https://leapp-to.github.io
 Source0:    https://github.com/oamg/leapp-repository/archive/leapp-repository-%{version}.tar.gz
 Source1:    leapp-repository-initrd.tar.gz
 Source2:    leapp-repository-data.tar.gz
+Source3:    deps-pkgs.tar.gz
 BuildArch:  noarch
-Requires:   dnf >= 4
 Requires:   %{name}-data = %{version}-%{release}
-%if 0%{?fedora} || 0%{?rhel} > 7
-Requires:   systemd-container
-%endif
+
+# IMPORTANT: everytime the requirements are changed, increment number by one
+# - same for Provides in deps subpackage
+Requires:   leapp-repository-dependencies = 1
 
 %description
 Repositories for leapp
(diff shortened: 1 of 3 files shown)


packaging/leapp-repository.spec
```diff
@@ -26,11 +23,29 @@ Repositories for leapp
...
+
+##################################################
+# Real requirements for the leapp-repository HERE
+##################################################
+Requires:   dnf >= 4
+%if 0%{?rhel} && 0%{?rhel} == 7
+# Required to gather system facts about SELinux
+Requires:   libselinux-python
+%else ## RHEL 8 dependencies ##
+# Requires:   systemd-container
```
bocekm: Why is this commented out?

packaging/leapp-repository.spec
```diff
@@ -26,11 +23,29 @@ Repositories for leapp
 %package data
 License: Red Hat Enterprise Agreement
 Summary: Package evolution data for leapp
-Requires: %{name} = %{version}-%{release}
+Requires:   %{name} = %{version}-%{release}
```
bocekm: Is this new indentation intentional?

packaging/leapp-repository.spec
```diff
@@ -26,11 +23,29 @@ Repositories for leapp
...
 
 %description data
 Package evolution data for leapp.
 
+%package deps
+Summary:    Meta-package with system dependencies of %{name} package
```
bocekm: I think it would be worth to add a comment here about why do we create this meta package.

bocekm:
How exactly this change makes sure that the python dependencies will not be removed during the upgrade?

packaging/leapp-repository.spec
pirat89: Because we do not need this RPM in our case

packaging/leapp-repository.spec
pirat89: small thing. I don't now.

pirat89:
leapp-ci build

Rezney:
LGTM. Tested.

pirat89:
As we need this and CI completely broken (and it doesn't seem to be fixed soon) and we need to release new version now, going to merge this as admin.

PR #50, opened 2019-01-22
Fix "transaction_workarounds" actor tests

Rezney:
We changed "to_install" field to "local_rpms" field and did not
reflect that in the tests.


repos/system_upgrade/el7toel8/actors/transactionworkarounds/tests/test_transaction_workarounds.py
```diff
@@ -5,6 +5,6 @@
 def test_execution(current_actor_context):
     current_actor_context.run()
     assert current_actor_context.consume(RpmTransactionTasks)
-    assert current_actor_context.consume(RpmTransactionTasks)[0].to_install
+    assert current_actor_context.consume(RpmTransactionTasks)[0].local_rpms
```
vinzenz:
```suggestion
    assert current_actor_context.consume(RpmTransactionTasks)[0].local_rpms
    assert not current_actor_context.consume(RpmTransactionTasks)[0]. to_install
```

Rezney: aah makes sense, forgot that. Fixed...

PR #53, opened 2019-01-24
Add actors for transferring SELinux customizations

vmojzis:
Edit by pstodulk (as nothing was here):

Add actors to scan system for custom SELinux rules and transfer these to RHEL 8 when possible. The process looks like this:
  - get all custom rules during the scan phase
  - remove all custom rules during the preparation phase (inside initrd before rpm transaction]
  - try to reapply custom rules and inform about those that cannot be reapplied

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
@@ -1,0 +1,104 @@
+from leapp.actors import Actor
+from leapp.models import SELinuxModules, SELinuxCustom, CheckResult
+from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
+from leapp.libraries.stdlib import call
+import subprocess
+import os
+from shutil import rmtree
+
+WORKING_DIRECTORY = '/tmp/selinux/'
+
+class SELinuxApplyCustom(Actor):
+    name = 'selinuxapplycustom'
+    description = 'No description has been provided for the selinuxapplycustom actor.'
+    consumes = (SELinuxCustom, SELinuxModules, )
+    produces = (CheckResult, )
+    tags = (ApplicationsPhaseTag, IPUWorkflowTag, )
+
+    def process(self):
+        # cil module files need to be extracted to disk in order to be installed
+        try:
+            # clear working directory
+            rmtree(WORKING_DIRECTORY)
+        except OSError:
+            #expected
+            pass
+        try:
+            os.mkdir(WORKING_DIRECTORY)
+        except OSError:
+            self.log.info("Failed to access working directory! Aborting.")
+            return
+
+        # import custom SElinux modules
+        for semodules in self.consume(SELinuxModules):
+            self.log.info("Processing custom SELinux policy modules. Count: " +
+                str(len(semodules.modules))
+            )
+            for module in semodules.modules:
+                cil_filename = WORKING_DIRECTORY + module.name + ".cil"
+                self.log.info("Installing " + module.name
+                 + " on priority " + str(module.priority) + ".")
+                if module.removed:
+                    self.log.info("The following lines where removed because of incompatibility: ")
+                    self.log.info('\n'.join(module.removed))
+                # write module content to disk
+                try:
+                    with open(cil_filename, 'w') as file:
+                        file.write(module.content)
+                except OSError as e:
+                    self.log.info("Error writing " + cil_filename + " :" + e.strerror)
+                    continue
+
+                try:
+                    semanage = call([
+                        'semodule',
+                        '-X',
+                        str(module.priority),
+                        '-i',
+                        cil_filename]
+                    )
+                except subprocess.CalledProcessError as e:
+                    self.log.info("Error installing module: " + e.strerror)
+                    pass
+                try:
+                    os.remove(cil_filename)
+                except OSError:
+                    self.log.info("Error removing module file")
+        # import SELinux customizations collected by "semanage export"
+        for custom in self.consume(SELinuxCustom):
+            self.log.info('Importing SELinux customizations collected by "semanage export".')
+            semanage_filename = WORKING_DIRECTORY + "semanage"
+            # save SELinux customizations to disk
+            try:
+                with open(semanage_filename, 'w') as file:
+                    file.write('\n'.join(custom.commands))
+            except OSError as e:
+                self.log.info("Error writing SELinux customizations:" + e.strerror)
+            # import customizations
+            try:
+                call(['semanage', 'import', '-f', semanage_filename])
+            except subprocess.CalledProcessError:
+                continue
+            # clean-up
+            try:
+                os.remove(semanage_filename)
+            except OSError:
+                continue
+
+        # clean-up
+        try:
+            os.rmdir("/tmp/selinux")
+        except OSError:
+            pass
+
+        self.log.info("SElinux customizations reapplied successfully.")
+        self.produce(
+           CheckResult(
+               severity='Info',
+               result='Pass',
+               summary='SElinux customizations reapplied successfully.',
+               details='SELinux modules with non-standard priority and other custom settings where reapplied after the upgrade.',
+               solutions=None
+        ))
+
+
--- a/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
@@ -41,7 +41,8 @@ class SELinuxApplyCustom(Actor):
                 cil_filename = os.path.join(WORKING_DIRECTORY, "%s.cil" % module.name)
                 self.log.info("Installing module %s on priority %d.", module.name, module.priority)
                 if module.removed:
-                    self.log.info("The following lines where removed because of incompatibility: \n%s", '\n'.join(module.removed))
+                    self.log.info("The following lines where removed because of incompatibility: \n%s",
+                                  '\n'.join(module.removed))
                 # write module content to disk
                 try:
                     with open(cil_filename, 'w') as cil_file:
@@ -62,14 +63,14 @@ class SELinuxApplyCustom(Actor):
                     self.log.info("Error installing module: %s", str(e))
                     # TODO - save the failed module to /etc/selinux ?
                     # currently it is still left in the old policy store
-                    pass
                 try:
                     os.remove(cil_filename)
                 except OSError as e:
                     self.log.info("Error removing module file: %s", str(e))
         # import SELinux customizations collected by "semanage export"
         for custom in self.consume(SELinuxCustom):
-            self.log.info('Importing the following SELinux customizations collected by "semanage export": \n%s', '\n'.join(custom.commands))
+            self.log.info('Importing the following SELinux customizations collected by "semanage export": \n%s',
+                          '\n'.join(custom.commands))
             semanage_filename = os.path.join(WORKING_DIRECTORY, "semanage")
             # save SELinux customizations to disk
             try:
@@ -103,5 +104,3 @@ class SELinuxApplyCustom(Actor):
 
         # TODO - summarize all changes after LEAPP team rewrites reporting
         # from leapp.reporting import Report
-
-
--- a/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
@@ -9,6 +9,7 @@ from leapp.libraries.stdlib import run, CalledProcessError
 
 WORKING_DIRECTORY = '/tmp/selinux/'
 
+
 class SELinuxApplyCustom(Actor):
     '''
     Re-apply SELinux customizations from RHEL-7 installation
@@ -52,13 +53,13 @@ class SELinuxApplyCustom(Actor):
                     continue
 
                 try:
-                    run([
-                        'semodule',
-                        '-X',
-                        str(module.priority),
-                        '-i',
-                        cil_filename]
-                    )
+                    run(['semodule',
+                         '-X',
+                         str(module.priority),
+                         '-i',
+                         cil_filename
+                         ]
+                        )
                 except CalledProcessError as e:
                     self.log.info("Error installing module: %s", str(e))
                     # TODO - save the failed module to /etc/selinux ?
--- a/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
@@ -1,88 +1,93 @@
-from leapp.actors import Actor
-from leapp.models import SELinuxModules, SELinuxCustom, CheckResult
-from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
-from leapp.libraries.stdlib import call
-import subprocess
 import os
 from shutil import rmtree
+
+from leapp.actors import Actor
+from leapp.models import SELinuxModules, SELinuxCustom, SELinuxRequestRPMs
+from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
+from leapp.libraries.stdlib import run, CalledProcessError
+
 
 WORKING_DIRECTORY = '/tmp/selinux/'
 
 class SELinuxApplyCustom(Actor):
+    '''
+    Re-apply SELinux customizations from RHEL-7 installation
+
+    Re-apply SELinux policy customizations (custom policy modules and changes
+    introduced by semanage). Any changes (due to incompatiblity with RHEL-8
+    SELinux policy) are reported to user.
+    '''
     name = 'selinuxapplycustom'
-    description = 'No description has been provided for the selinuxapplycustom actor.'
-    consumes = (SELinuxCustom, SELinuxModules, )
-    produces = (CheckResult, )
-    tags = (ApplicationsPhaseTag, IPUWorkflowTag, )
+    consumes = (SELinuxCustom, SELinuxModules)
+    produces = ()
+    tags = (ApplicationsPhaseTag, IPUWorkflowTag)
 
     def process(self):
         # cil module files need to be extracted to disk in order to be installed
-        try:
-            # clear working directory
-            rmtree(WORKING_DIRECTORY)
-        except OSError:
-            #expected
-            pass
+
+        # clear working directory
+        rmtree(WORKING_DIRECTORY, ignore_errors=True)
+
         try:
             os.mkdir(WORKING_DIRECTORY)
         except OSError:
-            self.log.info("Failed to access working directory! Aborting.")
+            self.log.info("Failed to create working directory! Aborting.")
             return
 
         # import custom SElinux modules
         for semodules in self.consume(SELinuxModules):
-            self.log.info("Processing custom SELinux policy modules. Count: " +
-                str(len(semodules.modules))
-            )
+            self.log.info("Processing custom SELinux policy modules. Count: %d.", len(semodules.modules))
             for module in semodules.modules:
-                cil_filename = WORKING_DIRECTORY + module.name + ".cil"
-                self.log.info("Installing " + module.name
-                 + " on priority " + str(module.priority) + ".")
+                cil_filename = os.path.join(WORKING_DIRECTORY, "%s.cil" % module.name)
+                self.log.info("Installing module %s on priority %d.", module.name, module.priority)
                 if module.removed:
-                    self.log.info("The following lines where removed because of incompatibility: ")
-                    self.log.info('\n'.join(module.removed))
+                    self.log.info("The following lines where removed because of incompatibility: \n%s", '\n'.join(module.removed))
                 # write module content to disk
                 try:
-                    with open(cil_filename, 'w') as file:
-                        file.write(module.content)
+                    with open(cil_filename, 'w') as cil_file:
+                        cil_file.write(module.content)
                 except OSError as e:
-                    self.log.info("Error writing " + cil_filename + " :" + e.strerror)
+                    self.log.info("Error writing %s : %s", cil_filename, str(e))
                     continue
 
                 try:
-                    semanage = call([
+                    run([
                         'semodule',
                         '-X',
                         str(module.priority),
                         '-i',
                         cil_filename]
                     )
-                except subprocess.CalledProcessError as e:
-                    self.log.info("Error installing module: " + e.strerror)
+                except CalledProcessError as e:
+                    self.log.info("Error installing module: %s", str(e))
+                    # TODO - save the failed module to /etc/selinux ?
+                    # currently it is still left in the old policy store
                     pass
                 try:
                     os.remove(cil_filename)
-                except OSError:
-                    self.log.info("Error removing module file")
+                except OSError as e:
+                    self.log.info("Error removing module file: %s", str(e))
         # import SELinux customizations collected by "semanage export"
         for custom in self.consume(SELinuxCustom):
-            self.log.info('Importing SELinux customizations collected by "semanage export".')
-            semanage_filename = WORKING_DIRECTORY + "semanage"
+            self.log.info('Importing the following SELinux customizations collected by "semanage export": \n%s', '\n'.join(custom.commands))
+            semanage_filename = os.path.join(WORKING_DIRECTORY, "semanage")
             # save SELinux customizations to disk
             try:
-                with open(semanage_filename, 'w') as file:
-                    file.write('\n'.join(custom.commands))
+                with open(semanage_filename, 'w') as s_file:
+                    s_file.write('\n'.join(custom.commands))
             except OSError as e:
-                self.log.info("Error writing SELinux customizations:" + e.strerror)
+                self.log.info("Error writing SELinux customizations: %s", str(e))
             # import customizations
             try:
-                call(['semanage', 'import', '-f', semanage_filename])
-            except subprocess.CalledProcessError:
+                run(['semanage', 'import', '-f', semanage_filename])
+            except CalledProcessError as e:
+                self.log.info("Failed to import SELinux customizations: %s", str(e))
                 continue
             # clean-up
             try:
                 os.remove(semanage_filename)
-            except OSError:
+            except OSError as e:
+                self.log.info("Failed to remove temporary file %s: %s", semanage_filename, str(e))
                 continue
 
         # clean-up
@@ -91,14 +96,12 @@ class SELinuxApplyCustom(Actor):
         except OSError:
             pass
 
-        self.log.info("SElinux customizations reapplied successfully.")
-        self.produce(
-           CheckResult(
-               severity='Info',
-               result='Pass',
-               summary='SElinux customizations reapplied successfully.',
-               details='SELinux modules with non-standard priority and other custom settings where reapplied after the upgrade.',
-               solutions=None
-        ))
+        # TODO - Verify that all RPM packages reqested by selinux actors are installed
+        self.log.info("Verifying selinux-related RPMs requested before upgrade.")
+        for rpms in self.consume(SELinuxRequestRPMs):
+            self.log.info("To keep: %s \n To install: %s", ", ".join(rpms.to_keep), ", ".join(rpms.to_install))
+
+        # TODO - summarize all changes after LEAPP team rewrites reporting
+        # from leapp.reporting import Report
 
 
--- a/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
@@ -1,5 +1,5 @@
 import os
-from shutil import rmtree
+import shutil
 
 from leapp.actors import Actor
 from leapp.models import SELinuxModules, SELinuxCustom, SELinuxRequestRPMs
@@ -27,47 +27,45 @@ class SELinuxApplyCustom(Actor):
         # cil module files need to be extracted to disk in order to be installed
 
         # clear working directory
-        rmtree(WORKING_DIRECTORY, ignore_errors=True)
+        shutil.rmtree(WORKING_DIRECTORY, ignore_errors=True)
 
         try:
             os.mkdir(WORKING_DIRECTORY)
         except OSError:
-            self.log.info("Failed to create working directory! Aborting.")
+            self.log.warning("Failed to create working directory! Aborting.")
             return
 
         # import custom SElinux modules
         for semodules in self.consume(SELinuxModules):
             self.log.info("Processing custom SELinux policy modules. Count: %d.", len(semodules.modules))
             for module in semodules.modules:
-                cil_filename = os.path.join(WORKING_DIRECTORY, "%s.cil" % module.name)
+                cil_filename = os.path.join(WORKING_DIRECTORY, "{}.cil".format(module.name))
                 self.log.info("Installing module %s on priority %d.", module.name, module.priority)
                 if module.removed:
-                    self.log.info("The following lines where removed because of incompatibility: \n%s",
-                                  '\n'.join(module.removed))
+                    self.log.warning("The following lines where removed because of incompatibility: \n%s",
+                                     '\n'.join(module.removed))
                 # write module content to disk
                 try:
                     with open(cil_filename, 'w') as cil_file:
                         cil_file.write(module.content)
                 except OSError as e:
-                    self.log.info("Error writing %s : %s", cil_filename, str(e))
+                    self.log.warning("Error writing %s : %s", cil_filename, str(e))
                     continue
 
                 try:
                     run(['semodule',
-                         '-X',
-                         str(module.priority),
-                         '-i',
-                         cil_filename
+                         '-X', str(module.priority),
+                         '-i', cil_filename
                          ]
                         )
                 except CalledProcessError as e:
-                    self.log.info("Error installing module: %s", str(e))
+                    self.log.warning("Error installing module: %s", str(e.stderr))
                     # TODO - save the failed module to /etc/selinux ?
                     # currently it is still left in the old policy store
                 try:
                     os.remove(cil_filename)
                 except OSError as e:
-                    self.log.info("Error removing module file: %s", str(e))
+                    self.log.warning("Error removing module file: %s", str(e))
         # import SELinux customizations collected by "semanage export"
         for custom in self.consume(SELinuxCustom):
             self.log.info('Importing the following SELinux customizations collected by "semanage export": \n%s',
@@ -78,25 +76,22 @@ class SELinuxApplyCustom(Actor):
                 with open(semanage_filename, 'w') as s_file:
                     s_file.write('\n'.join(custom.commands))
             except OSError as e:
-                self.log.info("Error writing SELinux customizations: %s", str(e))
+                self.log.warning("Error writing SELinux customizations: %s", str(e))
             # import customizations
             try:
                 run(['semanage', 'import', '-f', semanage_filename])
             except CalledProcessError as e:
-                self.log.info("Failed to import SELinux customizations: %s", str(e))
+                self.log.warning("Failed to import SELinux customizations: %s", str(e.stderr))
                 continue
             # clean-up
             try:
                 os.remove(semanage_filename)
             except OSError as e:
-                self.log.info("Failed to remove temporary file %s: %s", semanage_filename, str(e))
+                self.log.warning("Failed to remove temporary file %s: %s", semanage_filename, str(e))
                 continue
 
         # clean-up
-        try:
-            os.rmdir("/tmp/selinux")
-        except OSError:
-            pass
+        shutil.rmtree(WORKING_DIRECTORY, ignore_errors=True)
 
         # TODO - Verify that all RPM packages reqested by selinux actors are installed
         self.log.info("Verifying selinux-related RPMs requested before upgrade.")
--- a/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/tests/component_test.py
+++ b/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/tests/component_test.py
@@ -1,5 +1,7 @@
 import os
 
+import pytest
+
 from leapp.snactor.fixture import current_actor_context
 from leapp.models import SELinuxModule, SELinuxModules, SELinuxCustom, SELinuxFacts, SELinuxRequestRPMs
 from leapp.libraries.stdlib import api, run, CalledProcessError
@@ -15,16 +17,18 @@ test_modules = [
 ]
 
 semanage_commands = [
-['fcontext', '-t', 'ganesha_var_run_t', "'/ganesha(/.*)?'"],
-['fcontext', '-t', 'httpd_sys_content_t', "'/web(/.*)?'"],
-['port', '-t', 'http_port_t', '-p', 'udp', '81']
+    ['fcontext', '-t', 'ganesha_var_run_t', "'/ganesha(/.*)?'"],
+    ['fcontext', '-t', 'httpd_sys_content_t', "'/web(/.*)?'"],
+    ['port', '-t', 'http_port_t', '-p', 'udp', '81']
 ]
+
 
 def findModuleSemodule(semodule_lfull, name, priority):
     for line in semodule_lfull:
         if name in line and priority in line:
             return line
     return None
+
 
 def findSemanageRule(rules, rule):
     for r in rules:
@@ -35,11 +39,13 @@ def findSemanageRule(rules, rule):
             return r
     return None
 
+
+@pytest.mark.skip(reason='Test disabled because it would modify the system')
 def test_SELinuxApplyCustom(current_actor_context):
 
     semodule_list = [SELinuxModule(name=module, priority=int(prio),
                                    content="(allow domain proc_type (file (getattr open read)))", removed=[])
-                                   for (prio, module) in test_modules]
+                     for (prio, module) in test_modules]
 
     commands = [" ".join([c[0], "-a"] + c[1:]) for c in semanage_commands[1:]]
     semanage_removed = [" ".join([semanage_commands[0][0], "-a"] + semanage_commands[0][1:])]
@@ -68,7 +74,10 @@ def test_SELinuxApplyCustom(current_actor_context):
     for command in semanage_commands[1:-1]:
         assert findSemanageRule(semanage_export, command)
 
-def teardown():
+
+# Test disabled because it's setup and teardown would modify the system
+# Remove "_" before re-activation
+def teardown_():
     for priority, module in test_modules:
         try:
             run(["semodule", "-X", priority, "-r", module])
--- a/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/tests/component_test.py
+++ b/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/tests/component_test.py
@@ -1,0 +1,91 @@
+import os
+
+from leapp.snactor.fixture import current_actor_context
+from leapp.models import SELinuxModule, SELinuxModules, SELinuxCustom, SELinuxFacts, SELinuxRequestRPMs
+from leapp.libraries.stdlib import api, run, CalledProcessError
+from leapp.reporting import Report
+
+test_modules = [
+    ["400", "mock1"],
+    ["99", "mock1"],
+    ["200", "mock1"],
+    ["400", "mock2"],
+    ["999", "mock3"],
+    ["400", "permissive_abrt_t"]
+]
+
+semanage_commands = [
+['fcontext', '-t', 'ganesha_var_run_t', "'/ganesha(/.*)?'"],
+['fcontext', '-t', 'httpd_sys_content_t', "'/web(/.*)?'"],
+['port', '-t', 'http_port_t', '-p', 'udp', '81']
+]
+
+def findModuleSemodule(semodule_lfull, name, priority):
+    for line in semodule_lfull:
+        if name in line and priority in line:
+            return line
+    return None
+
+def findSemanageRule(rules, rule):
+    for r in rules:
+        for word in rule:
+            if word not in r:
+                break
+        else:
+            return r
+    return None
+
+def test_SELinuxApplyCustom(current_actor_context):
+
+    semodule_list = [SELinuxModule(name=module, priority=int(prio),
+                                   content="(allow domain proc_type (file (getattr open read)))", removed=[])
+                                   for (prio, module) in test_modules]
+
+    commands = [" ".join([c[0], "-a"] + c[1:]) for c in semanage_commands[1:]]
+    semanage_removed = [" ".join([semanage_commands[0][0], "-a"] + semanage_commands[0][1:])]
+
+    current_actor_context.feed(SELinuxModules(modules=semodule_list))
+    current_actor_context.feed(SELinuxCustom(commands=commands, removed=semanage_removed))
+    current_actor_context.run()
+
+    # check if all given modules and local customizations where removed
+    semodule_lfull = []
+    semanage_export = []
+    try:
+        semodule = run(["semodule", "-lfull"], split=True)
+        semodule_lfull = semodule.get("stdout", "")
+        semanage = run(["semanage", "export"], split=True)
+        semanage_export = semanage.get("stdout", "")
+    except CalledProcessError as e:
+        api.current_logger().warning("Error listing selinux customizations: %s", str(e.stderr))
+        assert False
+
+    # check that all modules installed during test setup where reported
+    for priority, name in test_modules:
+        if priority != "100" and priority != "200":
+            assert findModuleSemodule(semodule_lfull, name, priority)
+    # check that all valid commands where reintroduced to the system
+    for command in semanage_commands[1:-1]:
+        assert findSemanageRule(semanage_export, command)
+
+def teardown():
+    for priority, module in test_modules:
+        try:
+            run(["semodule", "-X", priority, "-r", module])
+        except CalledProcessError as e:
+            # expected if the test fails
+            api.current_logger().warning("Error removing selinux modules after testing: %s", str(e.stderr))
+
+    for command in semanage_commands[1:]:
+        try:
+            run(["semanage", command[0], "-d"] + [x.strip('"\'') for x in command[1:]])
+        except CalledProcessError as e:
+            # expected if the test fails
+            api.current_logger().warning("Error removing selinux customizations after testing: %s", str(e.stderr))
+            continue
+
+    try:
+        run(["semanage", semanage_commands[0][0], "-d"] + semanage_commands[0][1:])
+    except CalledProcessError:
+        # expected
+        pass
--- a/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/tests/component_test.py
+++ b/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/tests/component_test.py
@@ -31,11 +31,11 @@ def _run_cmd(cmd, logmsg="", split=True):
             api.current_logger().warning("%s: %s", logmsg, str(e.stderr))
 
 
-def findModuleSemodule(semodule_lfull, name, priority):
+def find_module_semodule(semodule_lfull, name, priority):
     return next((line for line in semodule_lfull if (name in line and priority in line)), None)
 
 
-def findSemanageRule(rules, rule):
+def find_semanage_rule(rules, rule):
     return next((r for r in rules if all(word in r for word in rule)), None)
 
 
@@ -79,7 +79,7 @@ def test_SELinuxApplyCustom(current_actor_context, destructive_selinux_teardown)
     # check that all reported modules where introduced to the system
     for priority, name in TEST_MODULES:
         if priority != "100" and priority != "200":
-            assert findModuleSemodule(semodule_lfull, name, priority)
+            assert find_module_semodule(semodule_lfull, name, priority)
     # check that all valid commands where introduced to the system
     for command in SEMANAGE_COMMANDS[1:-1]:
-        assert findSemanageRule(semanage_export, command)
+        assert find_semanage_rule(semanage_export, command)
--- a/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/tests/component_test.py
+++ b/repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/tests/component_test.py
@@ -7,7 +7,7 @@ from leapp.models import SELinuxModule, SELinuxModules, SELinuxCustom, SELinuxFa
 from leapp.libraries.stdlib import api, run, CalledProcessError
 from leapp.reporting import Report
 
-test_modules = [
+TEST_MODULES = [
     ["400", "mock1"],
     ["99", "mock1"],
     ["200", "mock1"],
@@ -16,85 +16,70 @@ test_modules = [
     ["400", "permissive_abrt_t"]
 ]
 
-semanage_commands = [
+SEMANAGE_COMMANDS = [
     ['fcontext', '-t', 'ganesha_var_run_t', "'/ganesha(/.*)?'"],
     ['fcontext', '-t', 'httpd_sys_content_t', "'/web(/.*)?'"],
     ['port', '-t', 'http_port_t', '-p', 'udp', '81']
 ]
 
 
+def _run_cmd(cmd, logmsg="", split=True):
+    try:
+        return run(cmd, split=split).get("stdout", "")
+    except CalledProcessError as e:
+        if logmsg:
+            api.current_logger().warning("%s: %s", logmsg, str(e.stderr))
+
+
 def findModuleSemodule(semodule_lfull, name, priority):
-    for line in semodule_lfull:
-        if name in line and priority in line:
-            return line
-    return None
+    return next((line for line in semodule_lfull if (name in line and priority in line)), None)
 
 
 def findSemanageRule(rules, rule):
-    for r in rules:
-        for word in rule:
-            if word not in r:
-                break
-        else:
-            return r
-    return None
+    return next((r for r in rules if all(word in r for word in rule)), None)
 
 
-@pytest.mark.skip(reason='Test disabled because it would modify the system')
-def test_SELinuxApplyCustom(current_actor_context):
+@pytest.fixture(scope="function")
+def destructive_selinux_teardown():
+    # actor introduces changes to the system, therefore only teardown is needed
+    yield
+
+    for priority, module in TEST_MODULES:
+        _run_cmd(["semodule", "-X", priority, "-r", module],
+                 "Error removing module {} after testing".format(module))
+
+    for command in SEMANAGE_COMMANDS[1:]:
+        _run_cmd(["semanage", command[0], "-d"] + [x.strip('"\'') for x in command[1:]],
+                 "Failed to remove SELinux customizations after testing")
+
+    _run_cmd(["semanage", SEMANAGE_COMMANDS[0][0], "-d"] + SEMANAGE_COMMANDS[0][1:],
+             "Failed to remove SELinux customizations after testing")
+
+
+@pytest.mark.skipif(os.getenv("DESTRUCTIVE_TESTING", False) in [False, "0"],
+                    reason='Test disabled by default because it would modify the system')
+def test_SELinuxApplyCustom(current_actor_context, destructive_selinux_teardown):
 
     semodule_list = [SELinuxModule(name=module, priority=int(prio),
                                    content="(allow domain proc_type (file (getattr open read)))", removed=[])
-                     for (prio, module) in test_modules]
+                     for (prio, module) in TEST_MODULES]
 
-    commands = [" ".join([c[0], "-a"] + c[1:]) for c in semanage_commands[1:]]
-    semanage_removed = [" ".join([semanage_commands[0][0], "-a"] + semanage_commands[0][1:])]
+    commands = [" ".join([c[0], "-a"] + c[1:]) for c in SEMANAGE_COMMANDS[1:]]
+    semanage_removed = [" ".join([SEMANAGE_COMMANDS[0][0], "-a"] + SEMANAGE_COMMANDS[0][1:])]
 
     current_actor_context.feed(SELinuxModules(modules=semodule_list))
     current_actor_context.feed(SELinuxCustom(commands=commands, removed=semanage_removed))
     current_actor_context.run()
 
-    # check if all given modules and local customizations where removed
-    semodule_lfull = []
-    semanage_export = []
-    try:
-        semodule = run(["semodule", "-lfull"], split=True)
-        semodule_lfull = semodule.get("stdout", "")
-        semanage = run(["semanage", "export"], split=True)
-        semanage_export = semanage.get("stdout", "")
-    except CalledProcessError as e:
-        api.current_logger().warning("Error listing selinux customizations: %s", str(e.stderr))
-        assert False
+    semodule_lfull = _run_cmd(["semodule", "-lfull"],
+                              "Error listing selinux modules")
+    semanage_export = _run_cmd(["semanage", "export"],
+                               "Error listing selinux customizations")
 
-    # check that all modules installed during test setup where reported
-    for priority, name in test_modules:
+    # check that all reported modules where introduced to the system
+    for priority, name in TEST_MODULES:
         if priority != "100" and priority != "200":
             assert findModuleSemodule(semodule_lfull, name, priority)
-    # check that all valid commands where reintroduced to the system
-    for command in semanage_commands[1:-1]:
+    # check that all valid commands where introduced to the system
+    for command in SEMANAGE_COMMANDS[1:-1]:
         assert findSemanageRule(semanage_export, command)
-
-
-# Test disabled because it's setup and teardown would modify the system
-# Remove "_" before re-activation
-def teardown_():
-    for priority, module in test_modules:
-        try:
-            run(["semodule", "-X", priority, "-r", module])
-        except CalledProcessError as e:
-            # expected if the test fails
-            api.current_logger().warning("Error removing selinux modules after testing: %s", str(e.stderr))
-
-    for command in semanage_commands[1:]:
-        try:
-            run(["semanage", command[0], "-d"] + [x.strip('"\'') for x in command[1:]])
-        except CalledProcessError as e:
-            # expected if the test fails
-            api.current_logger().warning("Error removing selinux customizations after testing: %s", str(e.stderr))
-            continue
-
-    try:
-        run(["semanage", semanage_commands[0][0], "-d"] + semanage_commands[0][1:])
-    except CalledProcessError:
-        # expected
-        pass
--- a/repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
@@ -1,20 +1,11 @@
-import os
-import re
-from shutil import rmtree
-
 from leapp.actors import Actor
 from leapp.models import SELinuxModules, SELinuxModule, SELinuxCustom, SELinuxFacts, SELinuxRequestRPMs, RpmTransactionTasks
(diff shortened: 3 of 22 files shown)


repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
```diff
@@ -0,0 +1,193 @@
...
+                commands=semanage_valid,
+                removed=semanage_removed
+            )
+        )
+
+    def checkModule(self, name):
+        """ Check if given module contains one of removed types.
+            If so, comment out corresponding lines and return them.
+            The function expects a text file "$name" containing cil policy
+            to be present in the current directory.
```
pirat89: Consider following of [PEP 257](https://www.python.org/dev/peps/pep-0257) with exceptions mentioned in contribution guidelines:
- https://github.com/oamg/leapp-repository/blob/master/CONTRIBUTING.md#code-style

It would help to the others and code will be more unified with the rest of the project. Similar for other descriptions in the code.

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
```diff
@@ -0,0 +1,193 @@
+import os
+import re
+from shutil import rmtree
+import subprocess
```
pirat89: the `subprocess` library should be omitted in favor of `leapp.libraries.stdlib` functions. When you find any use that you are not able to do with proposed functions, let me know and we will try to implement that. for each use of subprocess that is necessary, add please `# FIXME` line, so we can discuss it.

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
```diff
@@ -0,0 +1,193 @@
...
+                call(['semodule', '-c', '-X', priority, '-E', name])
+                # check if the module contains invalid types and remove them if so
+                removed = self.checkModule(name + ".cil")
+
+                # get content of the module
+                try:
+                    with open(name + ".cil", 'r') as file:
+                        module_content = file.read()
+                except OSError as e:
+                    self.log.info("Error reading %s : %s", name + ".cil", e.strerror)
```
pirat89: please use `format` when it is possible

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
```diff
@@ -0,0 +1,193 @@
...
+        except OSError:
+            self.log.info("Failed to access working directory! Aborting.")
+            return []
+
+        for (name, priority) in modules:
+            if priority == "200":
+                #TODO - request "name-selinux" to be installed
+                continue
+            if priority == "100":
+                #module from selinux-policy-* package - skipping
```
pirat89:
```suggestion
                # module from selinux-policy-* package - skipping
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
```diff
@@ -0,0 +1,193 @@
...
+            wd = os.getcwd()
+            os.mkdir(WORKING_DIRECTORY)
+            os.chdir(WORKING_DIRECTORY)
+        except OSError:
+            self.log.info("Failed to access working directory! Aborting.")
+            return []
+
+        for (name, priority) in modules:
+            if priority == "200":
+                #TODO - request "name-selinux" to be installed
```
pirat89: Not sure what exactly is meant by this comment.

pirat89:
@vmojzis  Do not forget to rebase branch with master with new update as the recent changes has to be applied to be able to upgrade.

bocekm:
@vmojzis, we are changing license of this repository from AGPL 3.0 to Apache 2.0 under PR https://github.com/oamg/leapp-repository/pull/112. Is it ok with you?

vmojzis:
Updated (see last commit message for deails) :)

bocekm:
@vmojzis, we are changing license of this repository from AGPL 3.0 to Apache 2.0 under PR https://github.com/oamg/leapp-repository/pull/112. Is it ok with you?

vmojzis:
@bocekm , yes, either licence is fine with me.

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+        try:
+            # clear working directory
+            rmtree(WORKING_DIRECTORY)
+        except OSError:
+            # expected
+            pass
+        try:
+            os.mkdir(WORKING_DIRECTORY)
+        except OSError:
+            self.log.info("Failed to access working directory! Aborting.")
```
fernflower: Failed to access or failed to create?

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+            return
+
+        # import custom SElinux modules
+        for semodules in self.consume(SELinuxModules):
+            self.log.info("Processing custom SELinux policy modules. Count: %d." % len(semodules.modules))
```
fernflower: Looks like log message will stay the same throughout the cycle run. Move outside to L38?

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+            for module in semodules.modules:
+                cil_filename = "%s%s.cil" % (WORKING_DIRECTORY, module.name)
```
fernflower: please don't use string formatting with paths, use os.path.join

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+                self.log.info("Installing module %s on priority %d." % (module.name, module.priority))
+                if module.removed:
+                    self.log.info("The following lines where removed because of incompatibility: ")
```
fernflower: Please squash L45-L46 into 1 self.log.info call

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+                # write module content to disk
+                try:
+                    with open(cil_filename, 'w') as file:
+                        file.write(module.content)
+                except OSError as e:
+                    self.log.info("Error writing %s : %s" % (cil_filename, e.strerror))
+                    continue
+
+                try:
+                    semanage = run([
```
fernflower: unused semanage var?

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+                    # currently it is still left in the old policy store
+                    pass
+                try:
+                    os.remove(cil_filename)
+                except OSError:
+                    self.log.info("Error removing module file.")
+        # import SELinux customizations collected by "semanage export"
+        for custom in self.consume(SELinuxCustom):
+            self.log.info('Importing SELinux customizations collected by "semanage export".')
+            semanage_filename = WORKING_DIRECTORY + "semanage"
```
fernflower: os.path.join

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+            # save SELinux customizations to disk
+            try:
+                with open(semanage_filename, 'w') as file:
```
fernflower: shadowed python builtin, rename please

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+                    file.write('\n'.join(custom.commands))
+            except OSError as e:
+                self.log.info("Error writing SELinux customizations: %s" % e.strerror)
+            # import customizations
+            try:
+                run(['semanage', 'import', '-f', semanage_filename])
+            except CalledProcessError:
```
fernflower: Should we log that?

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+                continue
+            # clean-up
+            try:
+                os.remove(semanage_filename)
+            except OSError:
```
fernflower: Should be log that?

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+
+        # clean-up
+        try:
+            os.rmdir("/tmp/selinux")
+        except OSError:
+            pass
+
+        # TODO - Verify that all RPM packages reqested by selinux actors are installed
+        for rpms in self.consume(SELinuxRequestRPMs):
+            self.log.info("Verifying selinux-related RPMs requested before upgrade.")
```
fernflower: Please squash into 1 log call

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
```diff
@@ -0,0 +1,223 @@
...
+from leapp.models import SELinuxModules, SELinuxModule, SELinuxCustom, SELinuxFacts, SELinuxRequestRPMs, RpmTransactionTasks
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+from leapp.libraries.stdlib import run, CalledProcessError
+
+WORKING_DIRECTORY = '/tmp/selinux/'
+
+# types and attributes that where removed between RHEL 7 and 8
+REMOVED_TYPES_=["base_typeattr_15","direct_run_init","gpgdomain","httpd_exec_scripts","httpd_user_script_exec_type","ibendport_type","ibpkey_type","pcmcia_typeattr_2","pcmcia_typeattr_3","pcmcia_typeattr_4","pcmcia_typeattr_5","pcmcia_typeattr_6","pcmcia_typeattr_7","sandbox_caps_domain","sandbox_typeattr_2","sandbox_typeattr_3","sandbox_typeattr_4","server_ptynode","systemctl_domain","user_home_content_type","userhelper_type","cgdcbxd_exec_t","cgdcbxd_t","cgdcbxd_unit_file_t","cgdcbxd_var_run_t","ganesha_use_fusefs","ganesha_exec_t","ganesha_t","ganesha_tmp_t","ganesha_unit_file_t","ganesha_var_log_t","ganesha_var_run_t","ganesha_use_fusefs"]
+# to be used with grep
+REMOVED_TYPES="|".join(REMOVED_TYPES_)
```
fernflower: Should REMOVED_TYPES really be a constant? Afaics it is used once in the code, can it be computed there?

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
```diff
@@ -0,0 +1,223 @@
...
+# to be used with sed
+SED_COMMAND='/' + '\|'.join(REMOVED_TYPES_) + '/s/^/;/g'
+
+# types, attributes and boolean contained in container-selinux
+CONTAINER_TYPES="|".join(["container_connect_any","container_runtime_t","container_runtime_exec_t","spc_t","container_auth_t","container_auth_exec_t","spc_var_run_t","container_var_lib_t","container_home_t","container_config_t","container_lock_t","container_log_t","container_runtime_tmp_t","container_runtime_tmpfs_t","container_var_run_t","container_plugin_var_run_t","container_unit_file_t","container_devpts_t","container_share_t","container_port_t","container_build_t","container_logreader_t","docker_log_t","docker_tmpfs_t","docker_share_t","docker_t","docker_lock_t","docker_home_t","docker_exec_t","docker_unit_file_t","docker_devpts_t","docker_config_t","docker_tmp_t","docker_auth_exec_t","docker_plugin_var_run_t","docker_port_t","docker_auth_t","docker_var_run_t","docker_var_lib_t","container_domain","container_net_domain"])
```
fernflower: Please see comments above about constants usage.
Imho
CONTAINER_TYPES_=["container_connect_any","container_runtime_t","container_runtime_exec_t","spc_t","container_auth_t","container_auth_exec_t","spc_var_run_t","container_var_lib_t","container_home_t","container_config_t","container_lock_t","container_log_t","container_runtime_tmp_t","container_runtime_tmpfs_t","container_var_run_t","container_plugin_var_run_t","container_unit_file_t","container_devpts_t","container_share_t","container_port_t","container_build_t","container_logreader_t","docker_log_t","docker_tmpfs_t","docker_share_t","docker_t","docker_lock_t","docker_home_t","docker_exec_t","docker_unit_file_t","docker_devpts_t","docker_config_t","docker_tmp_t","docker_auth_exec_t","docker_plugin_var_run_t","docker_port_t","docker_auth_t","docker_var_run_t","docker_var_lib_t","container_domain","container_net_domain"]
Other vars should be derived from it, but not defined as constants

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
```diff
@@ -0,0 +1,223 @@
...
+from leapp.libraries.stdlib import run, CalledProcessError
+
+WORKING_DIRECTORY = '/tmp/selinux/'
+
+# types and attributes that where removed between RHEL 7 and 8
+REMOVED_TYPES_=["base_typeattr_15","direct_run_init","gpgdomain","httpd_exec_scripts","httpd_user_script_exec_type","ibendport_type","ibpkey_type","pcmcia_typeattr_2","pcmcia_typeattr_3","pcmcia_typeattr_4","pcmcia_typeattr_5","pcmcia_typeattr_6","pcmcia_typeattr_7","sandbox_caps_domain","sandbox_typeattr_2","sandbox_typeattr_3","sandbox_typeattr_4","server_ptynode","systemctl_domain","user_home_content_type","userhelper_type","cgdcbxd_exec_t","cgdcbxd_t","cgdcbxd_unit_file_t","cgdcbxd_var_run_t","ganesha_use_fusefs","ganesha_exec_t","ganesha_t","ganesha_tmp_t","ganesha_unit_file_t","ganesha_var_log_t","ganesha_var_run_t","ganesha_use_fusefs"]
+# to be used with grep
+REMOVED_TYPES="|".join(REMOVED_TYPES_)
+# to be used with sed
+SED_COMMAND='/' + '\|'.join(REMOVED_TYPES_) + '/s/^/;/g'
```
fernflower: Same thing here, I'd rather not have this as a constant but define in code as a local var

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
```diff
@@ -0,0 +1,223 @@
...
+    produces = (SELinuxModules, SELinuxCustom, SELinuxRequestRPMs, RpmTransactionTasks,)
+    tags = (FactsPhaseTag, IPUWorkflowTag, )
+
+    def process(self):
+        # exit if SELinux is disabled
+        for fact in self.consume(SELinuxFacts):
+            if not fact.enabled:
+                return
+
+        (semodule_list, request_rpms) = self.getSELinuxModules()
```
fernflower: Maybe this will be more readable?
(semodule_list, rpms_to_keep, rpms_to_install) = self.getSELinuxModules()

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
```diff
@@ -0,0 +1,223 @@
...
+            semanage = run(['semanage', 'export'], split=True)
+            for line in semanage.get("stdout", ""):
+                for setype in REMOVED_TYPES_:
+                    if setype in line:
+                        semanage_removed.append(line)
+                        break
+                else:
+                    semanage_valid.append(line)
+
+        except CalledProcessError:
```
fernflower: Should we log that?

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
```diff
@@ -0,0 +1,223 @@
...
+        Check if given module contains one of removed types.
+
+        If so, comment out corresponding lines and return them.
+        The function expects a text file "$name" containing cil policy
+        to be present in the current directory.
+        '''
+        try:
+            removed = run(['grep', '-w', '-E', REMOVED_TYPES, name], split=True)
+            run(['sed', '-i', SED_COMMAND, name])
+            return removed.get("stdout", "")
```
fernflower: Inconsistent return, [] in case of Exception and ""

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
```diff
@@ -0,0 +1,223 @@
...
+        except CalledProcessError:
+            return []
```
fernflower: You really need to return a mutable here? Won't None do?

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
```diff
@@ -0,0 +1,223 @@
...
+        # list of rpms containing policy modules to be installed on RHEL 8
+        retain_rpms = []
+        install_rpms = []
+
+        # modules need to be extracted into cil files
+        # cd to /tmp/selinux and save working directory so that we can return there
+        try:
+            # clear working directory
+            rmtree(WORKING_DIRECTORY)
+        except OSError:
```
fernflower: Use ignore_errors=true instead of try/except

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
```diff
@@ -0,0 +1,223 @@
...
+            #expected
+            pass
+        try:
+            wd = os.getcwd()
+            os.mkdir(WORKING_DIRECTORY)
+            os.chdir(WORKING_DIRECTORY)
+        except OSError:
+            self.log.info("Failed to access working directory! Aborting.")
+            return []
```
fernflower: Incosistent return, [] and tuple of 3.

repos/system_upgrade/el7toel8/models/selinux.py
```diff
@@ -0,0 +1,36 @@
...
+
+    Listed packages provide types that where used in policy
+    customizations (to_install), or the corresponding policy 
+    was installed on RHEL-7 installation with priority 200
+    (to_keep).
+    """
+    topic = TransactionTopic
+    to_keep = fields.List(fields.String(), default=[])
+    to_install = fields.List(fields.String(), default=[])
+
```
fernflower: empty line

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+    name = 'selinuxapplycustom'
+    consumes = (SELinuxCustom, SELinuxModules, )
+    produces = (CheckResult, )
+    tags = (ApplicationsPhaseTag, IPUWorkflowTag, )
+
+    def process(self):
+        # cil module files need to be extracted to disk in order to be installed
+        try:
+            # clear working directory
+            rmtree(WORKING_DIRECTORY)
```
fernflower: Use ignore_errors=true instead of try/except

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
vmojzis: Create... thank you, missed this after refactoring.

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
vmojzis: The "for" loop is in place only in case there are multiple SELinuxModules objects (at this point there should only be one, but I was told this is the proper way to process the output of "consume").

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+class SELinuxApplyCustom(Actor):
+    '''
+    Re-apply SELinux customizations from RHEL-7 installation
+
+    Re-apply SELinux policy customizations (custom policy modules and changes
+    introduced by semanage). Any changes (due to incompatiblity with RHEL-8
+    SELinux policy) are reported to user.
+    '''
+    name = 'selinuxapplycustom'
+    consumes = (SELinuxCustom, SELinuxModules, )
```
vinzenz:
```suggestion
    consumes = (SELinuxCustom, SELinuxModules)
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+    produces = (CheckResult, )
+    tags = (ApplicationsPhaseTag, IPUWorkflowTag, )
```
vinzenz:
```suggestion
    tags = (ApplicationsPhaseTag, IPUWorkflowTag)
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+    '''
+    Re-apply SELinux customizations from RHEL-7 installation
+
+    Re-apply SELinux policy customizations (custom policy modules and changes
+    introduced by semanage). Any changes (due to incompatiblity with RHEL-8
+    SELinux policy) are reported to user.
+    '''
+    name = 'selinuxapplycustom'
+    consumes = (SELinuxCustom, SELinuxModules, )
+    produces = (CheckResult, )
```
vinzenz:
```suggestion
    produces = (CheckResult,)
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
vinzenz: Actually, you don't even know. 
```suggestion
        except OSError as e:
            self.log.info("Skipping to apply custom SELinux modules due to a failure while creating the working directory! Error reason: %s.", str(e))
```

I'd do something like this

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
vinzenz: yeah, @vmojzis is correct about this here. Just I would probably rephrase the log message:
```suggestion
            self.log.info("Processing custom SELinux policy modules message with %d modules." % len(semodules.modules))
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+        except OSError:
+            self.log.info("Failed to access working directory! Aborting.")
+            return
+
+        # import custom SElinux modules
+        for semodules in self.consume(SELinuxModules):
+            self.log.info("Processing custom SELinux policy modules. Count: %d." % len(semodules.modules))
+            for module in semodules.modules:
+                cil_filename = "%s%s.cil" % (WORKING_DIRECTORY, module.name)
+                self.log.info("Installing module %s on priority %d." % (module.name, module.priority))
```
vinzenz: For logging you can avoid applying the arguments immediately and have it done by the logger, in case info wouldn't be logged, it wouldn't perform the string formatting. Premature optimization, I know, but that's how one should use those loggers. Otherwise please prefer .format style string formatting.
```suggestion
                self.log.info("Installing module %s on priority %d.", module.name, module.priority)                
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+                if module.removed:
+                    self.log.info("The following lines where removed because of incompatibility: ")
+                    self.log.info('\n'.join(module.removed))
+                # write module content to disk
+                try:
+                    with open(cil_filename, 'w') as file:
```
vinzenz: please try to avoid using file, as it is a builtin function and you are overwriting it this way. You could just call it cil here too I think.

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+                try:
+                    semanage = run([
+                        'semodule',
+                        '-X',
+                        str(module.priority),
+                        '-i',
+                        cil_filename]
+                    )
+                except CalledProcessError as e:
+                    self.log.info("Error installing module: %s" % e.strerror)
```
vinzenz:
```suggestion
                    self.log.info("Error installing module: %s", e.strerror)
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+                    # TODO - save the failed module to /etc/selinux ?
+                    # currently it is still left in the old policy store
+                    pass
+                try:
+                    os.remove(cil_filename)
+                except OSError:
+                    self.log.info("Error removing module file.")
```
vinzenz: I'd add the exception message too
```suggestion
                    self.log.info("Error removing module file. %s", str(e))
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+                        cil_filename]
+                    )
+                except CalledProcessError as e:
+                    self.log.info("Error installing module: %s" % e.strerror)
+                    # TODO - save the failed module to /etc/selinux ?
+                    # currently it is still left in the old policy store
+                    pass
+                try:
+                    os.remove(cil_filename)
+                except OSError:
```
vinzenz:
```suggestion
                except OSError as e:
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,115 @@
...
+        # import SELinux customizations collected by "semanage export"
+        for custom in self.consume(SELinuxCustom):
+            self.log.info('Importing SELinux customizations collected by "semanage export".')
+            semanage_filename = WORKING_DIRECTORY + "semanage"
+            # save SELinux customizations to disk
+            try:
+                with open(semanage_filename, 'w') as file:
+                    file.write('\n'.join(custom.commands))
+            except OSError as e:
+                self.log.info("Error writing SELinux customizations: %s" % e.strerror)
```
vinzenz:
```suggestion
                self.log.info("Error writing SELinux customizations: %s" % e.strerror)
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
vinzenz: Oh yes I think so, this is pretty relevant IMHO

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
vinzenz: Definitely

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
vmojzis: There should only be one SELinuxRequestRPMs object.

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
vmojzis: Not true. 
run returns:
'stdout': result['stdout'].splitlines()
which is a list

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/actor.py
vmojzis: I'm using the returned value in SELinuxModule where a list is expected.

vmojzis:
Updated (all issues should be resolved) and rebased :)

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,107 @@
...
+            # save SELinux customizations to disk
+            try:
+                with open(semanage_filename, 'w') as s_file:
+                    s_file.write('\n'.join(custom.commands))
+            except OSError as e:
+                self.log.info("Error writing SELinux customizations: %s", str(e))
+            # import customizations
+            try:
+                run(['semanage', 'import', '-f', semanage_filename])
+            except CalledProcessError:
```
vinzenz: Missing:
```suggestion
            except CalledProcessError as e:
```

vmojzis: Good catch, thank you.

vmojzis:
Rebased.

pirat89:
@vmojzis Is it in that case still WIP? Anyway, please close all discussions when you resolved everything. It helps you to check you fixed everything and we do not have to go through it whether it was missed or not.

vmojzis:
Test customizations:

policycoreutils and policycoreutils-python have to be installed
```

semodule --cil -E zebra
semodule -X 400 -i zebra.cil 
semodule -X 99 -i zebra.cil
semodule -X 300 -i zebra.cil  
semodule --cil -E abrt
semodule -X 400 -i abrt.cil

semanage fcontext -a -t httpd_sys_content_t "/web(/.*)?"
semanage port -a -t http_port_t -p udp 81
semanage boolean --modify --on cron_can_relabel
semanage permissive -a abrt_t


To test if the changes where reapplied successfully use:

# semodule -lfull | grep -v 100
400 abrt              cil         
400 permissive_abrt_t cil         
400 zebra             cil         
300 zebra             cil         
099 zebra             cil 

semanage export
boolean -D
login -D
interface -D
user -D
port -D
node -D
fcontext -D
module -D
boolean -m -1 cron_can_relabel
port -a -t http_port_t -p udp 81
fcontext -a -f a -t httpd_sys_content_t '/web(/.*)?'


```

vmojzis:
It's not WIP any more. I resolved all comments and updated the PR message

AloisMahdal:
AFAIK we should switch and keep the system in Permissive post-upgrade, due to robustness reasons. Is this statement still true, @pirat89?

I tried to run downstream suite with this, and after the upgrade system is in Enforcing.

AloisMahdal:
Another question: when are user's SELinux modules supposed to be active --- is it possible to have them active all the time? If not, what is the window?

vinzenz:
> On 6 May 2019, at 19:01, Alois Mahdal <notifications@github.com> wrote:
> 
> AFAIK we should switch and keep the system in Permissive post-upgrade, due to robustness reasons. Is this statement still true, @pirat89?
> 
> I tried to run downstream suite with this, and after the upgrade system is in Enforcing.
> 
I knew it. I was “talking” with Petr about that the other day and I could have sworn we stopped putting it into permissive mode.
And I think we stopped doing it because we didn’t run into any problems.
During the initram phase we are in permissive mode when it was enforcing or permissive already.

AloisMahdal:
Few weeks ago Petr actually recommended to add specific check if system is in Permissive after upgrade.  That check has been passing ever since, I only saw it fail with this PR.

As far as I understand the reasons; this is also to prevent sec issues that arise during the upgrade *or FirstBoot* be fatal.  But actors might do things that would be triggered only by specific config, so unless we have 100% coverage of every actor, not running into any issues shouldn't make us confident to disable it.  (Ie. our only stragegy would be to enable Permissive; no matter what.  Which we're also saying in official docs)

It's not obvious to me if this PR should change anything about that.

vinzenz:
> On 6 May 2019, at 21:00, Alois Mahdal <notifications@github.com> wrote:
> 
> Few weeks ago Petr actually recommended to add specific check if system is in Permissive after upgrade. That check has been passing ever since, I only saw it fail with this PR.
> 
> As far as I understand the reasons; this is also to prevent sec issues that arise during the upgrade or FirstBoot be fatal. But actors might do things that would be triggered only by specific config, so unless we have 100% coverage of every actor, not running into any issues shouldn't make us confident to disable it. (Ie. our only stragegy would be to enable Permissive; no matter what. Which we're also saying in official docs)
> 
> It's not obvious to me if this PR should change anything about that.
> 
No it shouldn’t - it’s rather related to custom rules

vmojzis:
> Another question: when are user's SELinux modules supposed to be active --- is it possible to have them active all the time? If not, what is the window?

All custom modules are removed during the PreparationPhase and are reintroduced during ApplicationsPhase. This is necessary to ensure that the upgrade of selinux-policy works properly.

vojtechsokol:
leapp-ci build

fernflower:
Please rebase, master branch has updated linter checks (wrong-import-order you had because of leapp dependency was treated as 3rd party should not be there anymore)

examon:
leapp-ci build

vmojzis:
leapp-ci build

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/tests/component_test_.py
```diff
@@ -0,0 +1,95 @@
+import os
```
vinzenz: Please rename the file back to component_test.py

import pytest and disable tests by using:

```python
@pytest.mark.skip(reason='Test disabled because it would modify the system')
def test_SELinuxApplyCustom(current_actor_context):
...
```

vmojzis: Thank you. Updated.

vmojzis:
Rebased.

pirat89:
I am going to review this in next two days. Let's focus on the goal merge it in next 7days.

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
```diff
@@ -0,0 +1,200 @@
...
+            retain_rpms.append(name + "-selinux")
+            continue
+        if priority == "100":
+            # module from selinux-policy-* package - skipping
+            continue
+        # extract custom module and save it to SELinuxModule object
+        try:
+            run(['semodule', '-c', '-X', priority, '-E', name])
+            # check if the module contains invalid types and remove them if so
+            removed = checkModule(name + ".cil")
```
pirat89:
```suggestion
            removed = checkModule("%s.cil" % name)
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
```diff
@@ -0,0 +1,200 @@
...
+        os.chdir(WORKING_DIRECTORY)
+    except OSError:
+        api.current_logger().info("Failed to access working directory! Aborting.")
+        return ([], [], [])
+
+    for (name, priority) in modules:
+        if priority == "200":
+            # Module on priority 200 was installed by an RPM
+            # Request $name-selinux to be installed on RHEL8
+            retain_rpms.append(name + "-selinux")
```
pirat89:
```suggestion
            retain_rpms.append("%s-selinux" % name)
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
```diff
@@ -0,0 +1,200 @@
...
+            continue
+        # extract custom module and save it to SELinuxModule object
+        try:
+            run(['semodule', '-c', '-X', priority, '-E', name])
+            # check if the module contains invalid types and remove them if so
+            removed = checkModule(name + ".cil")
+
+            # get content of the module
+            try:
+                with open(name + ".cil", 'r') as cil_file:
```
pirat89:
```suggestion
                with open("%s.cil" % name, 'r') as cil_file:
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
```diff
@@ -0,0 +1,200 @@
...
+        # expected, ignore exception
+        pass
+
+    try:
+        os.chdir(wd)
+    except OSError:
+        pass
+    rmtree(WORKING_DIRECTORY, ignore_errors=True)
+
+    return (semodule_list, retain_rpms, install_rpms)
```
pirat89: I prefer to ensure that retain_rpms and install_rpms do not contain duplicit values regarding the implementation.

```suggestion
    return (semodule_list, list(set(retain_rpms)), list(set(install_rpms)))
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/tests/component_test.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def setup_():
+    for priority, module in test_modules:
+        try:
+            semodule = run(["semodule", "-X", priority, "-i", os.path.join(testmoduledir, module + ".cil")])
+        except CalledProcessError as e:
+            api.current_logger().warning("Error installing mock module: %s", e.stderr)
+            api.current_logger().warning("Error installing mock module: %s, %s", str(e.stderr),
+                                         semodule.get("stderr", "fuck"))
```
pirat89: Rather avoid that word in public project :-P
```suggestion
                                         semodule.get("stderr", "o-oh"))
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
```diff
@@ -0,0 +1,200 @@
...
+            run(['semodule', '-c', '-X', priority, '-E', name])
+            # check if the module contains invalid types and remove them if so
+            removed = checkModule(name + ".cil")
+
+            # get content of the module
+            try:
+                with open(name + ".cil", 'r') as cil_file:
+                    module_content = cil_file.read()
+            except OSError as e:
+                api.current_logger().info("Error reading %s.cil : %s", name, e.get("stderr", ""))
```
pirat89: I guess that this will cause traceback as OEError do not have `get` method.

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
```diff
@@ -0,0 +1,200 @@
...
+                   "container_home_t", "container_config_t", "container_lock_t", "container_log_t",
+                   "container_runtime_tmp_t", "container_runtime_tmpfs_t", "container_var_run_t",
+                   "container_plugin_var_run_t", "container_unit_file_t", "container_devpts_t", "container_share_t",
+                   "container_port_t", "container_build_t", "container_logreader_t", "docker_log_t", "docker_tmpfs_t",
+                   "docker_share_t", "docker_t", "docker_lock_t", "docker_home_t", "docker_exec_t",
+                   "docker_unit_file_t", "docker_devpts_t", "docker_config_t", "docker_tmp_t", "docker_auth_exec_t",
+                   "docker_plugin_var_run_t", "docker_port_t", "docker_auth_t", "docker_var_run_t",
+                   "docker_var_lib_t", "container_domain", "container_net_domain"]
+
+WORKING_DIRECTORY = '/tmp/selinux/'
```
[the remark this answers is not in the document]
vmojzis: No. But the files are in place only for a few seconds.
I asked about a better place to temporarily store files and didn't get an answer.

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/tests/component_test.py
vmojzis: Sorry about that, I was testing how to properly report the error message (got a little frustrated there) and forgot the code there.

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
```diff
@@ -0,0 +1,200 @@
...
+def listSELinuxModules():
+    '''
+    Produce list of SELinux policy modules
+
+    Returns list of tuples (name,priority)
+    '''
+    try:
+        semodule = run(['semodule', '-lfull'], split=True)
+    except CalledProcessError:
+        return []
```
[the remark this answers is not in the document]
vmojzis: Will fix.

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
```diff
@@ -0,0 +1,200 @@
...
+            api.current_logger().info("Module %s could not be extracted!", name)
+            continue
+        # rename the cil module file so that it does not clash
+        # with the same module on different priority
+        try:
+            os.rename(name + ".cil", "%s_%s" % (name, priority))
+        except OSError:
+            # TODO leapp.libraries.stdlib.api.current_logger()
+            # and move the method to a library
+            api.current_logger().info("Failed to rename module file.")
```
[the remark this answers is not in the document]
vmojzis: Yep, will fix.

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
vmojzis: Will fix.

vmojzis:
All, except for the temporary directory, should be fixed now.

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/tests/component_test.py
```diff
@@ -0,0 +1,112 @@
...
+    ['fcontext', '-t', 'httpd_sys_content_t', '"/web(/.*)?"'],
+    ['fcontext', '-t', 'ganesha_var_run_t', '"/ganesha(/.*)?"'],
+    ['port', '-t', 'http_port_t', '-p', 'udp', '81'],
+    ['permissive', 'abrt_t']
+]
+
+testmoduledir = os.path.join(os.getcwd(), "tests/mock_modules/")
+
+
+def setup_():
```
fernflower: I'm really surprised pytest allows setup_/teardown_ for names. Please follow their naming convention and add a module name after _, like https://docs.pytest.org/en/latest/xunit_setup.html

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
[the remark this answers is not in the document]
pirat89: Resolved for now. Set chmod for the directory so only root will be able to access it.

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,102 @@
...
+                self.log.info("Installing module %s on priority %d.", module.name, module.priority)
+                if module.removed:
+                    self.log.warning("The following lines where removed because of incompatibility: \n%s",
+                                     '\n'.join(module.removed))
+                # write module content to disk
+                try:
+                    with open(cil_filename, 'w') as cil_file:
+                        cil_file.write(module.content)
+                except OSError as e:
+                    self.log.warning("Error writing %s : %s", cil_filename, str(e.stderr))
```
pirat89: Does OSError have stderr at all? It's on many places.
```
>>> aaa = OSError('sfsfs')
>>> aaa.stder
AttributeError: 'exceptions.OSError' object has no attribute 'stderr'
```

vinzenz: Only our CalledProcessError does have it

OSError has errno however

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,102 @@
...
+                continue
+            # clean-up
+            try:
+                os.remove(semanage_filename)
+            except OSError as e:
+                self.log.warning("Failed to remove temporary file %s: %s", semanage_filename, str(e))
+                continue
+
+        # clean-up
+        rmtree(WORKING_DIRECTORY, ignore_errors=True)
```
fernflower: please use shutil.rmtree

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,102 @@
...
+
+        # TODO - Verify that all RPM packages reqested by selinux actors are installed
+        self.log.info("Verifying selinux-related RPMs requested before upgrade.")
+        for rpms in self.consume(SELinuxRequestRPMs):
+            self.log.info("To keep: %s \n To install: %s", ", ".join(rpms.to_keep), ", ".join(rpms.to_install))
```
fernflower: Sorry if I'm missing it, but where is the verification that requested rpms have been installed? I'm a bit mislead by the TODO comment - is it part of this patch or a follow up?

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/tests/component_test.py
```diff
@@ -0,0 +1,100 @@
...
+
+def findModuleSemodule(semodule_lfull, name, priority):
+    for line in semodule_lfull:
+        if name in line and priority in line:
+            return line
+    return None
+
+
+def findSemanageRule(rules, rule):
+    for r in rules:
```
fernflower: Same question as about findModuleSemodule

```
def findSemanageRule(rules, rule):
    return next((r for r in rules if all(word in r for word in rule)), None)
```
Another question - is `all(word in r for word in rule)` strong enough to judge that a matching rule is found? If there are more words in r then in rule it will still produce True. Shouldn't we have 2 identical word-sets comparison here?

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/tests/component_test.py
```diff
@@ -0,0 +1,100 @@
+import os
+
+import pytest
+
+from leapp.snactor.fixture import current_actor_context
+from leapp.models import SELinuxModule, SELinuxModules, SELinuxCustom, SELinuxFacts, SELinuxRequestRPMs
+from leapp.libraries.stdlib import api, run, CalledProcessError
+from leapp.reporting import Report
+
+test_modules = [
```
fernflower: I'd prefer TEST_MODULES to make 100% clear that's a global var

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/tests/component_test.py
```diff
@@ -0,0 +1,100 @@
...
+    ["400", "mock1"],
+    ["99", "mock1"],
+    ["200", "mock1"],
+    ["400", "mock2"],
+    ["999", "mock3"],
+    ["400", "permissive_abrt_t"]
+]
+
+semanage_commands = [
```
fernflower: SEMANAGE_COMMANDS

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
```diff
@@ -0,0 +1,102 @@
...
+
+        # TODO - summarize all changes after LEAPP team rewrites reporting
```
fernflower: Same question about reporting, the TODO is misleading.

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/tests/component_test.py
```diff
@@ -0,0 +1,100 @@
...
+    for priority, name in test_modules:
+        if priority != "100" and priority != "200":
+            assert findModuleSemodule(semodule_lfull, name, priority)
+    # check that all valid commands where reintroduced to the system
+    for command in semanage_commands[1:-1]:
+        assert findSemanageRule(semanage_export, command)
+
+
+# Test disabled because it's setup and teardown would modify the system
+# Remove "_" before re-activation
```
fernflower: Hmm does it mean there is no way to run this test automatically even by enabling skipped tests?

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def listSELinuxModules():
+    '''
+    Produce list of SELinux policy modules
+
+    Returns list of tuples (name,priority)
+    '''
+    try:
+        semodule = run(['semodule', '-lfull'], split=True)
```
fernflower: missed `return semodule.get("stdout", [])`?

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
```diff
@@ -0,0 +1,201 @@
...
+                   "container_port_t", "container_build_t", "container_logreader_t", "docker_log_t", "docker_tmpfs_t",
+                   "docker_share_t", "docker_t", "docker_lock_t", "docker_home_t", "docker_exec_t",
+                   "docker_unit_file_t", "docker_devpts_t", "docker_config_t", "docker_tmp_t", "docker_auth_exec_t",
+                   "docker_plugin_var_run_t", "docker_port_t", "docker_auth_t", "docker_var_run_t",
+                   "docker_var_lib_t", "container_domain", "container_net_domain"]
+
+WORKING_DIRECTORY = '/tmp/selinux/'
+
+
+def checkModule(name):
```
fernflower: rename to get_invalid_modules or smth more corresponding to the value returned?
Also please don't use CamelCase for functions.

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
```diff
@@ -0,0 +1,201 @@
...
+    try:
+        removed = run(['grep', '-w', '-E', "|".join(REMOVED_TYPES_), name], split=True)
+        # Add ";" at the beginning of invalid lines (comment them out)
+        run(['sed', '-i', '/{}/s/^/;/g'.format(r'\|'.join(REMOVED_TYPES_)), name])
+        return removed.get("stdout", [])
+    except CalledProcessError:
+        return []
+
+
+def listSELinuxModules():
```
fernflower: list_selinux_modules

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
```diff
@@ -0,0 +1,201 @@
...
+        if not m:
+            # invalid output of "semodule -lfull"
+            api.current_logger().warning('Invalid output of "semodule -lfull": %s', module)
+            continue
+        modules.append((m.group(2), m.group(1)))
+
+    return modules
+
+
+def getSELinuxModules():
```
fernflower: get_selinux_modules

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
```diff
@@ -0,0 +1,201 @@
...
+            # Request $name-selinux to be installed on RHEL8
+            retain_rpms.append(name + "-selinux")
+            continue
+        if priority == "100":
+            # module from selinux-policy-* package - skipping
+            continue
+        # extract custom module and save it to SELinuxModule object
+        try:
+            run(['semodule', '-c', '-X', priority, '-E', name])
+            # check if the module contains invalid types and remove them if so
```
fernflower: No need to form filename 2 times
```
module_file = name + ".cil"
removed = checkModule(module_file)
...
try:
    with open(module_file) as cil_file:
        ...
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/tests/component_test.py
```diff
@@ -0,0 +1,114 @@
...
+
+    for command in semanage_commands:
+        try:
+            run(["semanage", command[0], "-a"] + command[1:])
+        except CalledProcessError as e:
+            api.current_logger().warning("Error applying selinux customizations %s", str(e.stderr))
+            continue
+
+
+def findModule(selinuxmodules, name, priority):
```
fernflower: Code duplication?

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/tests/component_test.py
```diff
@@ -0,0 +1,114 @@
...
+    for module in selinuxmodules.modules:
+        if module.name == name and module.priority == int(priority):
+            return module
+    return None
+
+
+def findSemanageRule(rules, rule):
```
fernflower: Code duplication?

repos/system_upgrade/el7toel8/actors/selinux/selinuxprepare/actor.py
```diff
@@ -0,0 +1,24 @@
...
+    '''
+    Remove selinux policy customizations before updating selinux-policy* packages
+
+    RHEL-7 policy customizations could cause policy package upgrade to fail and therefore
+    need to be removed. Customizations introduced by semanage are removed first,
+    followed by custom policy modules gathered by SELinuxContentScanner.
+    '''
+
+    name = 'selinuxprepare'
+    # TODO change description to doc string - first line is summary, followed by more in-depth description
```
fernflower: Can this TODO be removed?

repos/system_upgrade/el7toel8/actors/selinux/selinuxprepare/tests/component_test.py
```diff
@@ -0,0 +1,101 @@
...
+        semodule_lfull = semodule.get("stdout", "")
+        semanage = run(["semanage", "export"], split=False)
+        semanage_export = semanage.get("stdout", "")
+    except CalledProcessError as e:
+        api.current_logger().warning("Error listing SELinux customizations: %s", str(e.stderr))
+
+testmoduledir = os.path.join(os.getcwd(), "tests/mock_modules/")
+
+
+def setup():
```
fernflower: Not skipped as other tests. Wondering how to do this straight (without blowing up a system during setup/teardown for a skipped test and still being able to run component tests automatically). An open question for the whole leapp team I guess :)

repos/system_upgrade/el7toel8/actors/selinux/selinuxprepare/tests/component_test.py
```diff
@@ -0,0 +1,101 @@
...
+if enabled:
+    try:
+        semodule = run(["semodule", "-lfull"], split=False)
+        semodule_lfull = semodule.get("stdout", "")
+        semanage = run(["semanage", "export"], split=False)
+        semanage_export = semanage.get("stdout", "")
+    except CalledProcessError as e:
+        api.current_logger().warning("Error listing SELinux customizations: %s", str(e.stderr))
+
+testmoduledir = os.path.join(os.getcwd(), "tests/mock_modules/")
```
fernflower: This way tests are not discovered properly, as `ACTOR=SELinuxPrepare make test` will have leapp-repository/ as cwd, not actor-specific directory.
You may try changing it to `testmoduledir = "tests/mock_modules/"` and constructing mock data dir in test_SELinuxPrepare() like
```
testdir = os.path.join(os.getenv('PYTEST_CURRENT_TEST').rsplit('/', 2)[0], testmoduledir)
run(["semodule", "-X", priority, "-i", os.path.join(testdir, module + ".cil")])
```

After that the dir is right but component test starts to fail with AssertionError
```
>   assert semodule_lfull == semodule.get("stdout", "")
E   AssertionError: assert '100 abrt    ... pp         \n' == '300 mock1    ... pp         \n'
E     + 300 mock1             cil         
E       100 abrt              pp         
E       100 accountsd         pp         
E       100 acct              pp         
E       100 afs               pp         
E       100 aiccu             pp         
E       100 aide              pp         ...
E     
E     ...Full output truncated (400 lines hidden), use '-vv' to show
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxprepare/tests/component_test.py
```diff
@@ -68,7 +68,7 @@ def test_SELinuxPrepare(current_actor_context):
         api.current_logger().warning("Error listing SELinux customizations: %s", str(e.stderr))
 
     semodule_list = [SELinuxModule(name=module, priority=int(prio), content="", removed=[])
-                     for (prio, module) in test_modules + [["400", "permissive_abrt_t"]]]
+                     for (prio, module) in test_modules.reverse() + [["400", "permissive_abrt_t"]]]
```
fernflower: that wont' work
- test_modules.reverse() returns None while changing underlying list structure;
- you shouldn't modify global vars

repos/system_upgrade/el7toel8/actors/selinux/selinuxprepare/tests/component_test.py
fernflower: Hmm interesting that after rebase the test actually passes. Will look into it tomorrow.

repos/system_upgrade/el7toel8/actors/selinux/selinuxprepare/tests/component_test.py
```diff
@@ -7,95 +7,86 @@
...
-            try:
-                run(["semanage", command[0], "-a"] + command[1:])
-            except CalledProcessError as e:
-                api.current_logger().warning("Error applying selinux customizations %s", str(e.stderr))
-                continue
-
-
-@pytest.mark.skip(reason='Test disabled because it would modify the system')
+    # NOTE(ivasilev) Maybe there is a more elegant way to conditionally run setup/teardown
+    enabled = os.environ.get("DESTRUCTIVE_TESTING", False)
```
vinzenz: well besides it probably is a bad idea just to check for the existence of an environment variable. If someone sets it to '0', I wouldn't expect this to trigger anyway but here it would.

pirat89:
Processed on system without custom selinux changes. See the output and the err (warning) msg file not found.

```
16070 Jul 12 12:50:08 localhost upgrade[311]: 2019-07-12 14:50:08.435 INFO     PID: 3 leapp.workflow.Applications: Executing actor selinuxapplycustom
16071 Jul 12 12:50:08 localhost upgrade[311]: 2019-07-12 14:50:08.467 INFO     PID: 438 leapp.workflow.Applications.selinuxapplycustom: Processing custom SELinux policy modules. Count: 0.
16072 Jul 12 12:50:08 localhost upgrade[311]: 2019-07-12 14:50:08.481 INFO     PID: 438 leapp.workflow.Applications.selinuxapplycustom: Importing the following SELinux customizations collected by "semanage export":
16073 Jul 12 12:50:08 localhost upgrade[311]: boolean -D                              
16074 Jul 12 12:50:08 localhost upgrade[311]: login -D                                
16075 Jul 12 12:50:08 localhost upgrade[311]: interface -D                            
16076 Jul 12 12:50:08 localhost upgrade[311]: user -D                                 
16077 Jul 12 12:50:08 localhost upgrade[311]: port -D                                 
16078 Jul 12 12:50:08 localhost upgrade[311]: node -D                                 
16079 Jul 12 12:50:08 localhost upgrade[311]: fcontext -D                             
16080 Jul 12 12:50:08 localhost upgrade[311]: module -D                               
16081 Jul 12 12:50:08 localhost upgrade[311]: boolean -m -1 virt_sandbox_use_all_caps 
16082 Jul 12 12:50:08 localhost upgrade[311]: boolean -m -1 virt_use_nfs              
16083 Jul 12 12:50:08 localhost upgrade[311]: module -d ganesha                       
16084 Jul 12 12:50:09 localhost upgrade[311]: 2019-07-12 14:50:09.586 WARNING  PID: 438 leapp.workflow.Applications.selinuxapplycustom: Failed to import SELinux customizations: FileNotFoundError: [Errno 2] No such file or directory
16085 Jul 12 12:50:09 localhost upgrade[311]: 2019-07-12 14:50:09.594 INFO     PID: 438 leapp.workflow.Applications.selinuxapplycustom: Verifying selinux-related RPMs requested before upgrade.
```

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/tests/component_test.py
```diff
@@ -0,0 +1,100 @@
...
+
+semanage_commands = [
+    ['fcontext', '-t', 'ganesha_var_run_t', "'/ganesha(/.*)?'"],
+    ['fcontext', '-t', 'httpd_sys_content_t', "'/web(/.*)?'"],
+    ['port', '-t', 'http_port_t', '-p', 'udp', '81']
+]
+
+
+def findModuleSemodule(semodule_lfull, name, priority):
+    for line in semodule_lfull:
```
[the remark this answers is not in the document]
vmojzis: What do you mean by "code with next"?

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
vmojzis: It's not implemented yet. I don't know how. Back when I wrote the comment I was told it would be somehow possible to check installed RPM's. But at this point I'm fine with "best effort".

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
vmojzis: Will fix

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/tests/component_test.py
vmojzis: Set comparison sounds better, will fix

repos/system_upgrade/el7toel8/actors/selinux/selinuxapplycustom/actor.py
vmojzis: Will fix

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
vmojzis: Nope... L64: for module in semodule.get("stdout", []):

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
vmojzis: It checks one module for lines that cannot be used in rhel 8. The name seems fine to me.

Seriously, you want me to rename all the functions now?

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
vmojzis: Will fix

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/tests/component_test.py
vmojzis: no, it's slightly different

repos/system_upgrade/el7toel8/actors/selinux/selinuxprepare/tests/component_test.py
vmojzis: Sorry, should have been reversed(test_modules)

repos/system_upgrade/el7toel8/actors/selinux/selinuxprepare/tests/component_test.py
vmojzis: Fixed

repos/system_upgrade/el7toel8/actors/selinux/selinuxprepare/tests/component_test.py
vmojzis: It seems that CWD differs in between tests for "make test"... fixed by using path to test file

repos/system_upgrade/el7toel8/actors/selinux/selinuxprepare/tests/component_test.py
vmojzis: Thanks for the changes.

repos/system_upgrade/el7toel8/actors/selinux/selinuxprepare/actor.py
vmojzis: yep, thanks

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/tests/component_test.py
vmojzis: Yes. I was told it's better than trying to set up sharing code between tests because of 1 function.

vmojzis:
All issues (except for "Camel case function names" and minor code duplication) should be resolved now. See "fixup" commits.

pirat89:
Added the PR to milestone for IPU 7.6 -> 8.1.

vmojzis:
Can I fix the last few commits up now? (The commits starting with "fixup" where created this way so that reviewers could easily identify the most recent changes)

pirat89:
@vmojzis Yes, you can

vmojzis:
Camel case in function names removed.

repos/system_upgrade/el7toel8/actors/selinux/selinuxcontentscanner/libraries/library.py
vmojzis: Fixed

vmojzis:
How do I "request review"? I went through all the reviews and they seem to be resolved, but github is still showing "requested changes" next to each reviewer.

vinzenz:
> How do I "request review"? I went through all the reviews and they seem to be resolved, but github is still showing "requested changes" next to each reviewer.

The little icon next to the names: image

vmojzis:
image

fernflower:
Also please rebase against master

vmojzis:
@fernflower : Could the failure be caused by missing package such as policycoreutils or policycoreutils-python (BTW the second one is not installed by "make install-deps"), it works fine for me on fresh vagrant machine?

pirat89:
@vmojzis if you need that rpm installed for tests, shouldbe installed by install-deps.

fernflower:
Looks like test failures are not PR-induced after all, subsequent actor-by-actor and `DESTRUCTIVE_TESTING make test` didn't produce the failures. Framework bug/cleanup race? 
Anyway I'm happy with the way the PR is now.

pirat89:
@vmojzis I don't see these rpms even in the spec file. We should fix that before merge

pirat89:
@vmojzis I don't see any of those in the spec nor in the Makefile. In case there is possibility that these will not be installed on the machine, the spec file and related Makefile have to be modified to reflect that. I will block th emerge for now until it will be clarified. We can set requirements tomorrow as expected, it's easy thing that will not take so much time, if needed.

AloisMahdal:
Nice, this already started breaking our auto-waivers downstream, which is kind of an extra sanity test that it works. :)

PR #55, opened 2019-01-28
Add actor exceptions to best practices doc page

asilveir:


docs/best-practices.md
```diff
@@ -132,3 +132,16 @@ the package if they want to experience as smooth upgrade as possible.
 If you're writing an actor for the RHEL 7 to RHEL 8 in-place upgrade workflow and you really need to have some package
 installed, then the only acceptable packages to depend on are the the ones available in the minimal installation of
 RHEL 7 (packages from the @core group + their dependencies).
+
+## When something goes wrong stop the execution of your actor in the right way
+
+If your actor get unexpected input or result of some operation, you must notify the framework about it. For that use these exceptions with a descriptive message:
```
jmikovic: This is a bit misleading since only Stop...Error exception takes message, StopActorExecution stops without message

docs/best-practices.md
```diff
@@ -132,3 +132,16 @@ the package if they want to experience as smooth upgrade as possible.
...
+- [StopActorExecution](https://leapp.readthedocs.io/en/latest/pydoc/leapp.html#leapp.exceptions.StopActorExecution) - will stop the current actor execution
+- [StopActorExecutionError](https://leapp.readthedocs.io/en/latest/pydoc/leapp.html#leapp.exceptions.StopActorExecutionError) - will stop the current actor execution, but also notify the framework that an error has occurred and can influence the result of the upgrade process.
+
+In case of [StopActorExecutionError](https://leapp.readthedocs.io/en/latest/pydoc/leapp.html#leapp.exceptions.StopActorExecutionError) the execution will stop or not according to the [policy](https://leapp.readthedocs.io/en/latest/pydoc/leapp.workflows.html?highlight=FailPhase#module-leapp.workflows.policies) defined in the workflow, there are three possibilities:
```
jmikovic: This might need more detail because execution of the actor ends every time, policy defines whether (and when) execution of the whole workflow stops

docs/best-practices.md
jmikovic: I'm not sure if 'you must notify' is the right way of putting it, I would rather say something like If you encounter unexpected input or any other error during execution of an actor and want to stop it, use these exceptions... Might be also a good idea to mention that these can be used in libraries and not only in actor functions.

bocekm: I agree. Since this document is Best practices, it shouldn't be forcing anyone to do something. So "must" is too strong a word. Somebody could do error handling in their own way.
If you're gonna mention the libraries, mention both the actor's private library and shared library.

docs/best-practices.md
bocekm: Yeah, I guess being explicit here won't hurt.
```
Framework will act based on the setting of the workflow phase policy [link] in one of the following three ways:
- end the upgrade process right away, or
- end the upgrade process after finishing the current phase, or
- do not end the upgrade process at all and continue with logging the issue only.
```

bocekm: Ah sorry, I found out you mention it down below. Then, just use "execution of the workflow" instead of just "execution".

docs/best-practices.md
```diff
@@ -132,3 +132,16 @@ the package if they want to experience as smooth upgrade as possible.
 If you're writing an actor for the RHEL 7 to RHEL 8 in-place upgrade workflow and you really need to have some package
 installed, then the only acceptable packages to depend on are the the ones available in the minimal installation of
 RHEL 7 (packages from the @core group + their dependencies).
+
+## When something goes wrong stop the execution of your actor in the right way
+
+If your actor get unexpected input or result of some operation, you must notify the framework about it. For that use these exceptions with a descriptive message:
+- [StopActorExecution](https://leapp.readthedocs.io/en/latest/pydoc/leapp.html#leapp.exceptions.StopActorExecution) - will stop the current actor execution
+- [StopActorExecutionError](https://leapp.readthedocs.io/en/latest/pydoc/leapp.html#leapp.exceptions.StopActorExecutionError) - will stop the current actor execution, but also notify the framework that an error has occurred and can influence the result of the upgrade process.
```
bocekm: Do not mention "upgrade" here (and down below). The best practices should be not only for the upgrade process, but for any other (future) workflow as well.
So, change "upgrade process" to "workflow execution", or something like that.

docs/best-practices.md
```diff
@@ -132,3 +132,16 @@ the package if they want to experience as smooth upgrade as possible.
 If you're writing an actor for the RHEL 7 to RHEL 8 in-place upgrade workflow and you really need to have some package
 installed, then the only acceptable packages to depend on are the the ones available in the minimal installation of
 RHEL 7 (packages from the @core group + their dependencies).
+
+## When something goes wrong stop the execution of your actor in the right way
+
+If your actor get unexpected input or result of some operation, you must notify the framework about it. For that use these exceptions with a descriptive message:
+- [StopActorExecution](https://leapp.readthedocs.io/en/latest/pydoc/leapp.html#leapp.exceptions.StopActorExecution) - will stop the current actor execution
```
bocekm: I would change "will stop the current actor execution" to "raising this exception is the same as returning None by the main `process()` method in the `actor.py`.

So, it is more of a convenient way to stop the actor, instead for example digging up from within deep call stack of actor's library functions using return values, for example.

docs/best-practices.md
```diff
@@ -132,3 +132,16 @@ the package if they want to experience as smooth upgrade as possible.
 If you're writing an actor for the RHEL 7 to RHEL 8 in-place upgrade workflow and you really need to have some package
 installed, then the only acceptable packages to depend on are the the ones available in the minimal installation of
 RHEL 7 (packages from the @core group + their dependencies).
+
+## When something goes wrong stop the execution of your actor in the right way
```
bocekm: "Use convenience exceptions to stop the actor's execution"

I think the "when" part doesn't necessarily be in the headline. That can be found in the details below.

jmikovic: +1

docs/best-practices.md
vinzenz: > Somebody could do error handling in their own way.
Which I would discourage if it's not necessary for some reason. Having a coherent codebase is better than too much variation.

But otherwise I agree with the discussion here.

PR #56, opened 2019-01-29
selinux unit tests

radomil:
Tests related to #27


radomil:
Tests fails due ImportError: No module named selinux

requirements.txt
```diff
@@ -5,4 +5,5 @@ pytest==3.6.4
 pytest-ordering==0.5
 funcsigs==1.0.2
 mock==2.0.0
+selinux
```
vinzenz: NO this is not ok, what you want is the Makefile with 'install-deps' target. Like this: https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/actors/systemfacts/Makefile

repos/system_upgrade/el7toel8/actors/systemfacts/tests/test_systemfacts_selinux.py
```diff
@@ -0,0 +1,50 @@
+import selinux
+from leapp.libraries.actor.systemfacts import get_selinux_status
```
vinzenz: There should be an empty line between

requirements.txt
radomil: @vinzenz , I put libselinux-python to Makefile, still that hasn't help to the import.

radomil:
@vinzenz , I've added the last test I wished to have in place. It's ready.

PR #58, opened 2019-01-31
mount overlayfs into diskimage to avoid problems with ftype=0

jmikovic:


repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/libraries/preparetransaction.py
```diff
@@ -163,3 +163,34 @@ def mount_dnf_cache(overlayfs_info):
...
 def umount_dnf_cache(overlayfs_info):
     return check_cmd_call(['/bin/umount', os.path.join(overlayfs_info.merged, 'var', 'cache', 'dnf')])
+
+
+def create_disk_image(path):
+    diskimage_path = os.path.join(path, 'diskimage')
+    mounts_path = os.path.join(path, 'mounts')
+
+    if os.path.isdir(mounts_path):
+        check_cmd_call(['/bin/umount', '-fl', mounts_path])
```
Rezney: You could give "remove_disk_image" function an "ignore_errors" parameter and then call it here.

Rezney: I know it is not related to this PR but do not we have already some stlib stuff for calling external programs?

vinzenz: the function check_cmd_call should have been refactored in #13 to use stdlib.call but it wasn't so yeah this needs to be done in a different PR

vinzenz:
@Rezney If we always go for it, we don't need to check for all the cases where this will be necessary. So I guess yeah always is probably better just to ensure it will work all the time and we'd find issues with this approach too.

vinzenz:
@jmikovic Please rebase and solve the conflicts

jmikovic:
leapp-ci build

PR #59, opened 2019-01-31
WIP: CheckNfs actor

examon:
Actor that checks if there is nfs on the system and inhibits upgrade if there is.
Currently, it checks:
- fstab
- mount
- systemd-mount
I don't know if this is sufficient. Are there any other places where is should look for nfs?

Also, I'm not really sure what to put as summary/details/solution for CheckResult.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checknfs/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checknfs/actor.py
@@ -1,0 +1,36 @@
+from leapp.actors import Actor
+from leapp.models import StorageInfo, CheckResult
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckNfs(Actor):
+    name = "check_nfs"
+    description = "Check if NFS filesystem is in use. If yes, inhibit the upgrade process."
+    consumes = (StorageInfo,)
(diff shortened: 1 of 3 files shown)


repos/system_upgrade/el7toel8/actors/checknfs/actor.py
```diff
@@ -0,0 +1,36 @@
...
+
+class CheckNfs(Actor):
+    name = "check_nfs"
+    description = "Check if NFS filesystem is in use. If yes, inhibit the upgrade process."
+    consumes = (StorageInfo,)
+    produces = (CheckResult,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag,)
+
+    def error(self):
+        self.produce(CheckResult(
```
jmikovic: Use Inhibitor model for this

repos/system_upgrade/el7toel8/actors/checknfs/actor.py
```diff
@@ -0,0 +1,36 @@
...
+            severity="Error",
+            result="Fail",
+            summary="NFS is not supported.",
```
vinzenz: And please try to make this all more friendly (I know it's WIP but still)

PR #64, opened 2019-02-05
CheckNfs actor

examon:
Checks for nfs entry in the fstab, mount and systemd-mount and inhibits upgrade if there is nfs present.

- now using Inhibitor instead of CheckResults

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checknfs/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checknfs/actor.py
@@ -35,7 +35,7 @@ class CheckNfs(Actor):
                     details += "- Currently mounted NFS shares\n"
                     break
 
-            # Check selinux-mount
+            # Check systemd-mount
             for systemdmount in storage.systemdmount:
                 if systemdmount.fs_type == "nfs":
                     nfs_found = True
--- a/repos/system_upgrade/el7toel8/actors/checknfs/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checknfs/actor.py
@@ -1,15 +1,15 @@
-"""
-Check if NFS filesystem is in use. If yes, inhibit the upgrade process.
-
-Actor looks for NFS in the following sources: /ets/fstab, mount and systemd-mount.
-If there is NFS in any of the mentioned sources, actors inhibits the upgrade.
-"""
 from leapp.actors import Actor
 from leapp.models import StorageInfo, Inhibitor
 from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
 
 
 class CheckNfs(Actor):
+    """
+    Check if NFS filesystem is in use. If yes, inhibit the upgrade process.
+
+    Actor looks for NFS in the following sources: /ets/fstab, mount and systemd-mount.
+    If there is NFS in any of the mentioned sources, actors inhibits the upgrade.
+    """
     name = "check_nfs"
     consumes = (StorageInfo,)
     produces = (Inhibitor,)
--- a/repos/system_upgrade/el7toel8/actors/checknfs/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checknfs/actor.py
@@ -44,7 +44,7 @@ class CheckNfs(Actor):
 
         if nfs_found:
             self.produce(Inhibitor(
-                summary="Unsupported NFS usage found.",
+                summary="Use of NFS detected. Upgrade can't proceed.",
                 details=details,
-                solutions="Please consider not using NFS."))
+                solutions="Disable NFS temporarily for the upgrade if possible."))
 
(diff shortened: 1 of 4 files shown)


repos/system_upgrade/el7toel8/actors/checknfs/actor.py
```diff
@@ -0,0 +1,34 @@
...
+        for storage in self.consume(StorageInfo):
+            # Check fstab
+            for fstab in storage.fstab:
+                if fstab.fs_vfstype == "nfs":
+                    self.inhibit("Discovered NFS entry in the fstab.")
+
+            # Check mount
+            for mount in storage.mount:
+                if mount.tp == "nfs":
+                    self.inhibit("Discovered NFS entry in the mount.")
```
vinzenz: Discovered mounted NFS share

examon:
leapp-ci build

vinzenz:
Now add some tests for verifying the input data and you're good to go I think

repos/system_upgrade/el7toel8/actors/checknfs/actor.py
```diff
@@ -0,0 +1,45 @@
+from leapp.actors import Actor
+from leapp.models import StorageInfo, Inhibitor
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckNfs(Actor):
+    name = "check_nfs"
+    description = "Check if NFS filesystem is in use. If yes, inhibit the upgrade process."
```
vinzenz: Please use a docstring now for the description - First line a short summary, empty line and  a body with a bit more detail.

repos/system_upgrade/el7toel8/actors/checknfs/actor.py
```diff
@@ -0,0 +1,50 @@
...
+                if systemdmount.fs_type == "nfs":
+                    nfs_found = True
+                    details += "- One or more configured NFS mounts in systemd-mount\n"
+                    break
+
+        if nfs_found:
+            self.produce(Inhibitor(
+                summary="Unsupported NFS usage found.",
+                details=details,
+                solutions="Please consider not using NFS."))
```
bocekm: This message sounds a bit funny to me. Like "Please consider not going to work" when you visit a therapist because of stress from work.
What about "Disable NFS temporarily for the upgrade if possible."?

repos/system_upgrade/el7toel8/actors/checknfs/actor.py
```diff
@@ -0,0 +1,50 @@
...
+            # Check systemd-mount
+            for systemdmount in storage.systemdmount:
+                if systemdmount.fs_type == "nfs":
+                    nfs_found = True
+                    details += "- One or more configured NFS mounts in systemd-mount\n"
+                    break
+
+        if nfs_found:
+            self.produce(Inhibitor(
+                summary="Unsupported NFS usage found.",
```
bocekm: I would change this sentence because as a user I see:
`2019-02-19 18:12:32.653994 [ERROR] Actor: verify_check_results Message: Error: Fail: Unsupported NFS usage found.` And I would be bewildered a bit by that.
What about: "Use of NFS detected. Upgrade can't proceed."

I would add something like "See report for details." But this is actually something that the VerifyCheckResults should be printing.

repos/system_upgrade/el7toel8/models/storageinfo.py
```diff
@@ -82,6 +82,16 @@ class LvdisplayEntry(Model):
     cpy_sync = fields.String()
     convert = fields.String()
 
+class SystemdMountEntry(Model):
```
bocekm: Leave two newlines above and after the class definition (PEP E302).

bocekm:
Squash please.

marceloleitner:
This breaks the tests using vagrant, from https://gitlab.cee.redhat.com/leapp/oamg-rhel7-vagrant.git, as it mounts a nfs share at /vagrant.

examon:
@marceloleitner Yeah, this is unfortunate. We will try to get storage team to write proper actor that deals with nfs.
In the meanwhile, I think you can work around this issue by fully removing `checknfs` actor from `repos/system_upgrade/el7toel8/actors`.

bocekm:
@vinzenz, I guess now this is the first real use case where a leapp option like --disable-actors would come handy.
We're now in closer contact with the Storage team, so we can bring this up. It would be great if they come up with a solution to allow the use of NFS.

vinzenz:
@bocekm Ok will add it
Well the problem with NFS is that we won't have it in the initram phase because we do not bring up the network. I am not sure if that helps.

bocekm:
@vinzenz, cool.
By allowing NFS I meant something more like having a smarter solution, for example disabling the NFS temporarily instead of inhibiting the upgrade.

pirat89:
That's the problem. We cannot disable NFS temporarily. But we would rather create possibility to define which NFS partitions can be ignored by the actor (in case they are not in the fstab - in such case `mount -a` fails and not sure what will happen exactly.

bocekm:
> We cannot disable NFS temporarily.
I disagree with this statement. It is technically possible to prevent from mounting NFS before the upgrade and get it to the original mount state after the upgrade.

@shaded-enmity had a valid argument why we shouldn't disable it by default - users can have a database stored on an NFS share and the database package during upgrade may have scriptlets that would upgrade the database structure. So preventing the NFS should not be automatic but intentional (by answering some question for example).

PR #66, opened 2019-02-06
Make sure /boot is mounted before removing boot entry

Rezney:
We need to make sure /boot is mounted before trying to remove the boot entry.


Rezney:
@vinzenz tests are passing now...

PR #67, opened 2019-02-07
Add ntp migration actors

mlichvar:
Add CheckNtp actor to check if the ntpd and/or ntpdate services are
enabled, active and their configuration files exist. Also, save the
configuration files in a string containing a base64-encoded compressed
tarball to avoid having to deal with .rpmsave files when the packages
are removed in the rpm upgrade.

Add MigrateNtp actor to enable the chronyd service, restore the
ntp/ntpdate configuration files, and run the included ntp2chrony
migration script. Produce a warning with a suggestion to check the
generated /etc/chrony.conf if some lines in ntp.conf were ignored in the
migration.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkntp/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkntp/actor.py
@@ -5,8 +5,11 @@ from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
 
 
 class CheckNtp(Actor):
+    """
+    Check if ntp and/or ntpdate configuration needs to be migrated.
+    """
+
     name = 'check_ntp'
-    description = 'Check if ntp and/or ntpdate configuration needs to be migrated.'
     consumes = (InstalledRedHatSignedRPM,)
     produces = (Report, NtpMigrationDecision)
     tags = (ChecksPhaseTag, IPUWorkflowTag)
--- a/repos/system_upgrade/el7toel8/actors/checkntp/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkntp/actor.py
@@ -3,6 +3,7 @@ from leapp.libraries.actor.library import check_ntp
 from leapp.models import Report, InstalledRedHatSignedRPM, NtpMigrationDecision
 from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
 
+
 class CheckNtp(Actor):
     name = 'check_ntp'
     description = 'Check if ntp and/or ntpdate configuration needs to be migrated.'
@@ -16,6 +17,6 @@ class CheckNtp(Actor):
         signed_rpms = self.consume(InstalledRedHatSignedRPM)
         for rpm_pkgs in signed_rpms:
             for pkg in rpm_pkgs.items:
-	        installed_packages.add(pkg.name)
+                installed_packages.add(pkg.name)
 
         self.produce(check_ntp(installed_packages))
--- a/repos/system_upgrade/el7toel8/actors/checkntp/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkntp/actor.py
@@ -1,0 +1,21 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.library import check_ntp
+from leapp.models import Report, InstalledRedHatSignedRPM, NtpMigrationDecision
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+class CheckNtp(Actor):
+    name = 'check_ntp'
+    description = 'Check if ntp and/or ntpdate configuration needs to be migrated.'
+    consumes = (InstalledRedHatSignedRPM,)
+    produces = (Report, NtpMigrationDecision)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        installed_packages = set()
+
+        signed_rpms = self.consume(InstalledRedHatSignedRPM)
+        for rpm_pkgs in signed_rpms:
+            for pkg in rpm_pkgs.items:
+	        installed_packages.add(pkg.name)
+
+        self.produce(check_ntp(installed_packages))
--- a/repos/system_upgrade/el7toel8/actors/checkntp/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/checkntp/libraries/library.py
@@ -1,11 +1,10 @@
 import base64
 import io
 import os
-from subprocess import CalledProcessError
 import tarfile
 
 from leapp.libraries.common import reporting
-from leapp.libraries.stdlib import api, run
+from leapp.libraries.stdlib import CalledProcessError, api, run
 from leapp.models import NtpMigrationDecision
 
 
--- a/repos/system_upgrade/el7toel8/actors/checkntp/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/checkntp/libraries/library.py
@@ -1,0 +1,65 @@
+from leapp.libraries.common import reporting
+from leapp.libraries.stdlib import api, run
+from leapp.models import NtpMigrationDecision
+
+from subprocess import CalledProcessError
+
+import base64, io, os, tarfile
+
+# Check if a service is active and enabled
+def check_service(name):
+    for state in ['active', 'enabled']:
+        try:
+            run(['systemctl', 'is-{}'.format(state), name])
+            api.current_logger().debug('{} is {}'.format(name, state))
+        except CalledProcessError:
+            api.current_logger().debug('{} is not {}'.format(name, state))
+            return False
+
+    return True
+
+# Check if a file exists
+def is_file(name):
+    return os.path.isfile(name)
+
+# Get a base64-encoded gzipped tarball of specified files
+def get_tgz64(filenames):
+    stream = io.BytesIO()
+    tar = tarfile.open(fileobj=stream, mode='w:gz')
+    for filename in filenames:
+        if os.path.isfile(filename):
+            tar.add(filename)
+    tar.close()
+
+    return base64.b64encode(stream.getvalue())
+
+# Check services from the ntp packages for migration
+def check_ntp(installed_packages):
+    service_data = [('ntpd', 'ntp', '/etc/ntp.conf'),
+                    ('ntpdate', 'ntpdate', '/etc/ntp/step-tickers'),
+                    ('ntp-wait', 'ntp-perl', None)]
+
+    migrate_services = []
+    migrate_configs = []
+    for service, package, main_config in service_data:
+        if package in installed_packages and \
+                check_service('{}.service'.format(service)) and \
+                (not main_config or is_file(main_config)):
+            migrate_services.append(service)
+            if main_config:
+                migrate_configs.append(service)
+
+    if len(migrate_configs):
+        reporting.report_generic(
+             title='{} configuration will be migrated'.format(' and '.join(migrate_configs)),
+             summary='{} service(s) detected to be enabled and active'.format(', '.join(migrate_services)),
+             severity='low')
+        # Save configuration files that will be renamed in the upgrade
+        config_tgz64 = get_tgz64(['/etc/ntp.conf', '/etc/ntp/keys',
+                                  '/etc/ntp/crypto/pw', '/etc/ntp/step-tickers'])
+    else:
+        api.current_logger().info('ntpd/ntpdate configuration will not be migrated')
+        migrate_services = []
+        config_tgz64 = ''
+
+    return NtpMigrationDecision(migrate_services=migrate_services, config_tgz64=config_tgz64)
--- a/repos/system_upgrade/el7toel8/actors/checkntp/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/checkntp/libraries/library.py
@@ -1,10 +1,13 @@
+import base64
+import io
+import os
+from subprocess import CalledProcessError
+import tarfile
+
 from leapp.libraries.common import reporting
 from leapp.libraries.stdlib import api, run
 from leapp.models import NtpMigrationDecision
 
-from subprocess import CalledProcessError
-
-import base64, io, os, tarfile
 
 # Check if a service is active and enabled
 def check_service(name):
@@ -18,9 +21,11 @@ def check_service(name):
 
     return True
 
+
 # Check if a file exists
 def is_file(name):
     return os.path.isfile(name)
+
 
 # Get a base64-encoded gzipped tarball of specified files
 def get_tgz64(filenames):
@@ -32,6 +37,7 @@ def get_tgz64(filenames):
     tar.close()
 
     return base64.b64encode(stream.getvalue())
+
 
 # Check services from the ntp packages for migration
 def check_ntp(installed_packages):
--- a/repos/system_upgrade/el7toel8/actors/migratentp/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/migratentp/actor.py
@@ -5,8 +5,11 @@ from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
 
 
 class MigrateNtp(Actor):
+    """
+    Migrate ntp and/or ntpdate configuration to chrony.
+    """
+
     name = 'migrate_ntp'
-    description = 'Migrate ntp and/or ntpdate configuration to chrony.'
     consumes = (NtpMigrationDecision,)
     produces = (Report,)
     tags = (ApplicationsPhaseTag, IPUWorkflowTag)
--- a/repos/system_upgrade/el7toel8/actors/migratentp/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/migratentp/actor.py
@@ -1,0 +1,15 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.library import migrate_ntp
+from leapp.models import Report, NtpMigrationDecision
+from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
+
+class MigrateNtp(Actor):
+    name = 'migrate_ntp'
+    description = 'Migrate ntp and/or ntpdate configuration to chrony.'
+    consumes = (NtpMigrationDecision,)
+    produces = (Report,)
+    tags = (ApplicationsPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        for decision in self.consume(NtpMigrationDecision):
+            migrate_ntp(decision.migrate_services, decision.config_tgz64)
--- a/repos/system_upgrade/el7toel8/actors/migratentp/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/migratentp/actor.py
@@ -3,6 +3,7 @@ from leapp.libraries.actor.library import migrate_ntp
 from leapp.models import Report, NtpMigrationDecision
 from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
 
+
 class MigrateNtp(Actor):
     name = 'migrate_ntp'
     description = 'Migrate ntp and/or ntpdate configuration to chrony.'
--- a/repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
+++ b/repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
@@ -1,0 +1,671 @@
+#!/usr/bin/python
+#
+# Convert ntp configuration to chrony
+#
+# Copyright (C) 2018-2019  Miroslav Lichvar <mlichvar@redhat.com>
+#
+# Permission is hereby granted, free of charge, to any person obtaining
+# a copy of this software and associated documentation files (the
+# "Software"), to deal in the Software without restriction, including
+# without limitation the rights to use, copy, modify, merge, publish,
+# distribute, sublicense, and/or sell copies of the Software, and to
+# permit persons to whom the Software is furnished to do so, subject to
+# the following conditions:
+#
+# The above copyright notice and this permission notice shall be included
+# in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+# IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+# CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+# TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+# SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+
+import argparse
+import ipaddress
+import logging
+import os
+import os.path
+import re
+import subprocess
+import sys
+
+# python2 compatibility hacks
+if sys.version_info[0] < 3:
+    from io import open
+    reload(sys)
+    sys.setdefaultencoding("utf-8")
+
+class NtpConfiguration(object):
+    def __init__(self, root_dir, ntp_conf, step_tickers):
+        self.root_dir = root_dir if root_dir != "/" else ""
+        self.ntp_conf_path = ntp_conf
+        self.step_tickers_path = step_tickers
+
+        # Read and write files using an 8-bit transparent encoding
+        self.file_encoding = "latin-1"
+        self.enabled_services = set()
+        self.step_tickers = []
+        self.time_sources = []
+        self.fudges = {}
+        self.restrictions = {
+                # Built-in defaults
+                ipaddress.ip_network(u"0.0.0.0/0"): set(),
+                ipaddress.ip_network(u"::/0"): set(),
+        }
+        self.keyfile = ""
+        self.keys = []
+        self.trusted_keys = []
+        self.driftfile = ""
+        self.statistics = []
+        self.leapfile = ""
+        self.tos_options = {}
+        self.ignored_directives = set()
+        self.ignored_lines = []
+
+        #self.detect_enabled_services()
+        self.parse_step_tickers()
+        self.parse_ntp_conf()
+
+    def detect_enabled_services(self):
+        for service in ["ntpdate", "ntpd", "ntp-wait"]:
+            if os.path.islink("{}/etc/systemd/system/multi-user.target.wants/{}.service"
+                    .format(self.root_dir, service)):
+                self.enabled_services.add(service)
+        logging.info("Enabled services found in /etc/systemd/system: %s",
+                     " ".join(self.enabled_services))
+
+    def parse_step_tickers(self):
+        if not self.step_tickers_path:
+            return
+
+        path = os.path.join(self.root_dir, self.step_tickers_path)
+        if not os.path.isfile(path):
+            logging.info("Missing %s", path)
+            return
+
+        with open(path, encoding=self.file_encoding) as f:
+            for line in f:
+                line = line[:line.find('#')]
+
+                words = line.split()
+
+                if not words:
+                    continue
+
+                self.step_tickers.extend(words)
+
+    def parse_ntp_conf(self, path=None):
+        if path is None:
+            path = os.path.join(self.root_dir, self.ntp_conf_path)
+
+        with open(path, encoding=self.file_encoding) as f:
+            logging.info("Reading %s", path)
+
+            for line in f:
+                line = line[:line.find('#')]
+
+                words = line.split()
+
+                if not words:
+                    continue
+
+                if not self.parse_directive(words):
+                    self.ignored_lines.append(line)
+
+    def parse_directive(self, words):
+        name = words.pop(0)
+        if name.startswith("logconfig"):
+            name = "logconfig"
+
+        if words:
+            if name in ["server", "peer", "pool"]:
+                return self.parse_source(name, words)
+            elif name == "fudge":
+                return self.parse_fudge(words)
+            elif name == "restrict":
+                return self.parse_restrict(words)
+            elif name == "tos":
+                return self.parse_tos(words)
+            elif name == "includefile":
+                return self.parse_includefile(words)
+            elif name == "keys":
+                return self.parse_keys(words)
+            elif name == "trustedkey":
+                return self.parse_trustedkey(words)
+            elif name == "driftfile":
+                self.driftfile = words[0]
+            elif name == "statistics":
+                self.statistics = words
+            elif name == "leapfile":
+                self.leapfile = words[0]
+            else:
+                self.ignored_directives.add(name)
+                return False
+        else:
+            self.ignored_directives.add(name)
+            return False
+
+        return True
+
+    def parse_source(self, source_type, words):
+        ipv4_only = False
+        ipv6_only = False
+        source = {
+                "type": source_type,
+                "options": []
+        }
+
+        if words[0] == "-4":
+            ipv4_only = True
+            words.pop(0)
+        elif words[0] == "-6":
+            ipv6_only = True
+            words.pop(0)
+
+        if not words:
+            return False
+
+        source["address"] = words.pop(0)
+
+        # Check if -4/-6 corresponds to the address and ignore hostnames
+        if ipv4_only or ipv6_only:
+            try:
+                version = ipaddress.ip_address(source["address"]).version
+                if (ipv4_only and version != 4) or (ipv6_only and version != 6):
+                    return False
+            except ValueError:
+                return False
+
+        if source["address"].startswith("127.127."):
+            if not source["address"].startswith("127.127.1."):
+                # Ignore non-LOCAL refclocks
+                return False
+
+        while words:
+            if len(words) >= 2 and words[0] in ["minpoll", "maxpoll", "version", "key"]:
+                source["options"].append((words[0], words[1]))
+                words = words[2:]
+            elif words[0] in ["burst", "iburst", "noselect", "prefer", "true", "xleave"]:
+                source["options"].append((words[0],))
+                words.pop(0)
+            else:
+                return False
+
+        self.time_sources.append(source)
+        return True
+
+    def parse_fudge(self, words):
+        address = words.pop(0)
+        options = {}
+
+        while words:
+            if len(words) >= 2 and words[0] in ["stratum"]:
+                if not words[1].isdigit():
+                    return False
+                options[words[0]] = int(words[1])
+                words = words[2:]
+            elif len(words) >= 2:
+                words = words[2:]
+            else:
+                return False
+
+        self.fudges[address] = options
+        return True
+
+    def parse_restrict(self, words):
+        ipv4_only = False
+        ipv6_only = False
+        flags = set()
+        mask = ""
+
+        if words[0] == "-4":
+            ipv4_only = True
+            words.pop(0)
+        elif words[0] == "-6":
+            ipv6_only = True
+            words.pop(0)
+
+        if not words:
+            return False
+
+        address = words.pop(0)
+
+        while words:
+            if len(words) >= 2 and words[0] == "mask":
+                mask = words[1]
+                words = words[2:]
+            else:
+                if words[0] not in ["kod", "nomodify", "notrap", "nopeer", "noquery",
+                                    "limited", "ignore", "noserve"]:
+                    return False
+                flags.add(words[0])
+                words.pop(0)
+
+        # Convert to IP network(s), ignoring restrictions with hostnames
+        networks = []
+        if address == "default" and not mask:
+            if not ipv6_only:
+                networks.append(ipaddress.ip_network(u"0.0.0.0/0"))
+            if not ipv4_only:
+                networks.append(ipaddress.ip_network(u"::/0"))
+        else:
+            try:
+                if mask:
+                    networks.append(ipaddress.ip_network(u"{}/{}".format(address, mask)))
+                else:
+                    networks.append(ipaddress.ip_network(address))
+            except ValueError:
+                return False
+
+            if (ipv4_only and networks[-1].version != 4) or \
+                    (ipv6_only and networks[-1].version != 6):
+                return False
+
+        for network in networks:
+            self.restrictions[network] = flags
+
+        return True
+
+    def parse_tos(self, words):
+        options = {}
+        while words:
+            if len(words) >= 2 and words[0] in ["minsane", "orphan"]:
+                if not words[1].isdigit():
+                    return False
+                options[words[0]] = int(words[1])
+                words = words[2:]
+            elif len(words) >= 2 and words[0] in ["maxdist"]:
+                # Check if it is a float value
+                if not words[1].replace('.', '', 1).isdigit():
+                    return False
+                options[words[0]] = float(words[1])
+                words = words[2:]
+            else:
+                return False
+
+        self.tos_options.update(options)
+
+        return True
+
+    def parse_includefile(self, words):
+        path = os.path.join(self.root_dir, words[0])
+        if not os.path.isfile(path):
+            return False
+
+        self.parse_ntp_conf(path)
+        return True
+
+    def parse_keys(self, words):
+        keyfile = words[0]
+        path = os.path.join(self.root_dir, keyfile)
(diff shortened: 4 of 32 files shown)


pirat89:
leapp-ci build

vojtechsokol:
leapp-ci build

repos/system_upgrade/el7toel8/actors/migratentp/actor.py
```diff
@@ -0,0 +1,76 @@
...
+            else:
+                # Use a minimal secure ntp.conf if ntpd was disabled
+                ntp_conf = '/etc/ntp.conf.nosources'
+                with open(ntp_conf, 'w') as f:
+                    f.write('# This file was created to migrate ntpdate configuration to chrony\n')
+                    f.write('# without ntp configuration (ntpd service was disabled)\n')
+                    f.write('driftfile /var/lib/ntp/drift\n')
+                    f.write('restrict default ignore nomodify notrap nopeer noquery\n')
+
+            if 'ntpdate' in decision.migrate_services:
```
fernflower: nit: oneliner will be easier to read

step_tickers = '/etc/ntp/step-tickers' if 'ntpdate' in decision.migrate_services else ''

repos/system_upgrade/el7toel8/actors/checkntp/actor.py
```diff
@@ -0,0 +1,74 @@
...
+    description = 'Check if ntp and/or ntpdate configuration needs to be migrated.'
+    consumes = (InstalledRedHatSignedRPM,)
+    produces = (CheckResult, NtpMigrationDecision)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def check_service(self, name):
+        active = False
+        enabled = False
+        try:
+            ret = call(['systemctl', 'is-active', name])
```
fernflower: unused var?

repos/system_upgrade/el7toel8/actors/checkntp/actor.py
```diff
@@ -0,0 +1,74 @@
...
+            active = True
+            self.log.debug('{} is active'.format(name))
+        except CalledProcessError:
+            self.log.debug('{} is not active'.format(name))
+
+        try:
+            ret = call(['systemctl', 'is-enabled', name])
```
fernflower: Can 2 check procedures be merged into one? Or reuse code?

repos/system_upgrade/el7toel8/actors/migratentp/actor.py
```diff
@@ -0,0 +1,76 @@
...
+            p = Popen(['/bin/sh', '-c', 'base64 -d | tar xz -C /'], stdin=PIPE, stdout=None, stderr=None)
+            p.communicate(input=decision.config_tgz64.encode())
+
+            if 'ntpd' in decision.migrate_services:
+                ntp_conf = '/etc/ntp.conf'
+            else:
+                # Use a minimal secure ntp.conf if ntpd was disabled
+                ntp_conf = '/etc/ntp.conf.nosources'
+                with open(ntp_conf, 'w') as f:
+                    f.write('# This file was created to migrate ntpdate configuration to chrony\n')
```
fernflower: Maybe move minimal config definiton contents to some variable instead of writing it here line per line?

repos/system_upgrade/el7toel8/actors/migratentp/actor.py
```diff
@@ -0,0 +1,76 @@
...
+            try:
+                ntp_configuration = ntp2chrony.NtpConfiguration('/', ntp_conf, step_tickers, 0)
+                ntp_configuration.write_chrony_configuration('/etc/chrony.conf', '/etc/chrony.keys',
+                                                             False, True)
+            except Exception as e:
+                raise StopActorExecutionError('ntp2chrony failed: {}'.format(e))
+
+            # Report ignored lines in ntp.conf, except 'disable monitor' from
+            # the default ntp.conf
+            for ignored_line in ntp_configuration.ignored_lines:
```
fernflower: Can't get the logic here. IIUC the first line that is ignored but isn't 'disable_monitor' will break the loop (not processing any other lines) and go straight to L72, while L66 will be reached if and only if ntp_configuration.ignored_lines == ['disable monitor'].
Is this intended? In that case - can you please rewrite it to make it more obvious?

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,664 @@
...
+        self.parse_step_tickers()
+        self.parse_ntp_conf()
+
+    def detect_enabled_services(self):
+        for service in ["ntpdate", "ntpd", "ntp-wait"]:
+            if os.path.islink("{}/etc/systemd/system/multi-user.target.wants/{}.service"
+                    .format(self.root_dir, service)):
+                self.enabled_services.add(service)
+        if self.verbose > 0:
+            print("Enabled services found in /etc/systemd/system: " +
```
fernflower: please you string formatting

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,664 @@
...
+                    " ".join(self.enabled_services))
+
+    def parse_step_tickers(self):
+        if not self.step_tickers_path:
+            return
+
+        path = self.root_dir + self.step_tickers_path
+        if not os.path.isfile(path):
+            if self.verbose > 0:
+                print("Missing " + path)
```
fernflower: Use logging instead of prints?

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,664 @@
...
+            return
+
+        with open(path, encoding="latin-1") as f:
```
fernflower: Is this the default encoding for every single case? Imho worth adding an encoding param to the fucntion and not having magic constants

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,664 @@
...
+
+                words = line.split()
+
+                if not words:
+                    continue
+
+                self.step_tickers.extend(words)
+
+    def parse_ntp_conf(self, path=None):
+        if path is None:
```
fernflower: nit: path = self.root_dir + self.ntp_conf_path if not path else path

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,664 @@
...
+
+    def parse_directive(self, words):
+        name = words.pop(0)
+        if name.startswith("logconfig"):
+            name = "logconfig"
+
+        if words:
+            if name in ["server", "peer", "pool"]:
+                return self.parse_source(name, words)
+            elif name == "fudge":
```
fernflower: nit: elif -> if for all cases with return

Actually you can even have an action map like
action_map = {"tos": self.parse_tos,
                      "includefile": self.parse_includefile}
and call the majority of methods as action_map[key](words)

But it's just a suggestion of course

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,664 @@
...
+
+        for tos in self.tos_options:
+            if tos[0] == "maxdist":
+                maxdistance = float(tos[1])
+            elif tos[0] == "minsane":
+                minsources = int(tos[1])
+            elif tos[0] == "orphan":
+                orphan_stratum = int(tos[1])
+            else:
+                assert False
```
fernflower: assert False?

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,664 @@
...
+        maxdistance = 0.0
+        minsources = 1
+        orphan_stratum = 0
+        logs = []
+
+        for source in self.time_sources:
+            address = source["address"]
+            if address.startswith("127.127.1."):
+                if address in self.fudges and "stratum" in self.fudges[address]:
+                    local_stratum = int(self.fudges[address]["stratum"])
```
fernflower: Can this raise a ValueError during convertion?

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,664 @@
...
+
+            for line in self.ignored_lines:
+                # Remove sensitive information
+                line = re.sub(r"\s+pw\s+\S+", " pw XXX", line.rstrip())
+                conf += "# " + line + "\n"
+            conf += "\n"
+
+        conf += self.get_chrony_conf_sources()
+
+        conf += "# Record the rate at which the system clock gains/losses time.\n"
```
fernflower: Again imho worth considering having the base of config in some variable.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,664 @@
...
+
+        return keys
+
+    def write_file(self, path, mode, content, backup):
+        path = self.root_dir + path
+        if backup and os.path.isfile(path):
+            os.rename(path, path + ".old")
+
+        with open(os.open(path, os.O_CREAT | os.O_WRONLY | os.O_EXCL, mode), "w",
+                  encoding="latin-1") as f:
```
fernflower: please do not hardcode encoding

repos/system_upgrade/el7toel8/actors/migratentp/actor.py
```diff
@@ -0,0 +1,76 @@
...
+            # Enable the chronyd service (the chrony package should be
+            # installed after rpm upgrade)
+            try:
+                call(['systemctl', 'enable', 'chronyd.service'])
+            except CalledProcessError:
+                raise StopActorExecutionError('Could not enable chronyd service')
+
+            # Unpack archive with configuration files
+            #call(['/bin/sh', '-c', 'base64 -d | tar xv'], stdin=decision.config_tgz64)
+            p = Popen(['/bin/sh', '-c', 'base64 -d | tar xz -C /'], stdin=PIPE, stdout=None, stderr=None)
```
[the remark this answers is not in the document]
pirat89: anyway, use of the `subprocess` module and Popen especially is something we do not want to have in the code in future. But agree with @fernflower, python modules

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
pirat89: In this case, we are talking about script that is used probably somewhere else as well and tested already. I would keep it as it is as it can follow even different practices in other projects. I think this is something what we do not have mentioned anywhere even best practices - how we will look at the third party code. Of course, we can keep it as it is code going just into the project, but... I feel that we should think that yet.

fernflower:
leapp-ci build

vojtechsokol:
leapp-ci build

vojtechsokol:
Tests are failing because of missing import: `ImportError: No module named ipaddress`.
Status is not updated to `failure` due to a bug in our CI.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,664 @@
...
+            elif name == "leapfile":
+                self.leapfile = words[0]
+            else:
+                self.ignored_directives.add(name)
+                return False
+        else:
+            self.ignored_directives.add(name)
+            return False
+
+        return True
```
fernflower: Maybe convert 'return True\False' to 'return a_dict_of_parsed_values'? This will make it instantly more obvious which exactly params were processed and make writing tests for actual conversion a bit easier, as this will allow section-by-section convertion testing

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,664 @@
...
+            return False
+
+        source["address"] = words.pop(0)
+
+        # Check if -4/-6 corresponds to the address and ignore hostnames
+        if ipv4_only or ipv6_only:
+            try:
+                version = ipaddress.ip_address(source["address"]).version
+                if (ipv4_only and version != 4) or (ipv6_only and version != 6):
+                    return False
```
fernflower: same thing here, please see comment for changing return False to return a_dict_of_processed_values

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,664 @@
...
+        self.trusted_keys = []
+        self.driftfile = ""
+        self.statistics = []
+        self.leapfile = ""
+        self.tos_options = []
+        self.ignored_directives = set()
+        self.ignored_lines = []
+
+        #self.detect_enabled_services()
+        self.parse_step_tickers()
```
fernflower: I'd move calls to logic outside of __init__, intoduce addional parse() class-method and call it specifically from main(). 
I also feel that some methods can be made _private.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: Should be fixed in HEAD.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: Makes sense. Should be fixed in HEAD.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: We don't know the encoding, so files are read and written using the same (8-bit transparent) encoding. I've added a variable for the encoding to make it more clear.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: This looks like a matter of preference.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: That would work, but it would need to be applied to the whole parser and I'm not sure if it's worth the trouble. I prefer consistency.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: That would fail is there was a unexpected tos (i.e. bug). Anyway, I've reworked this for the other issue with int() failing.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: Yes, good point. It should be now fixed.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: There are many strings in the code like this. I don't think it makes sense to put only some in variables.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: Same as above.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: How would you distinguish between an error in the parsing and nothing parsed?

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: That would mean parsing could be performed twice or not at all? That doesn't make much sense to me. The class represents a single instance of ntp configuration.

pirat89:
@mlichvar Do not forget to rebase branch with master with new update as the recent changes has to be applied to be able to upgrade.

repos/system_upgrade/el7toel8/actors/checkntp/actor.py
mlichvar: This was simplified in the latest commit.

repos/system_upgrade/el7toel8/actors/checkntp/actor.py
mlichvar: Fixed.

repos/system_upgrade/el7toel8/actors/migratentp/actor.py
mlichvar: Ok. Fixed.

repos/system_upgrade/el7toel8/actors/migratentp/actor.py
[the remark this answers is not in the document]
mlichvar: Ok. Rewritten to use the tarfile, io, base64 modules.

fernflower:
leapp-ci build

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,670 @@
...
+            password = key[2]
+
+            if type in ["m", "M"]:
+                type = "MD5"
+            elif type not in ["MD5", "SHA1", "SHA256", "SHA384", "SHA512"]:
+                continue
+
+            prefix = "ASCII" if len(password) <= 20 else "HEX"
+
+            for first, last in self.trusted_keys:
```
fernflower: iiuc L600-L605 are the same as
trusted = any(id in range(first, last+1) for (first, last) in self.trusted_keys)

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,670 @@
...
+
+    def get_chrony_keys(self):
+        if not self.keyfile:
+            return ""
+
+        keys = "# This file was converted from {}.\n".format(self.keyfile)
+        keys += "\n"
+
+        for key in self.keys:
+            id = key[0]
```
fernflower: We don't have it in the guidelines yet but please don't use builtins (id, type) as vars

key_id, key_type, password = key

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,670 @@
...
+
+    def write_file(self, path, mode, content, backup):
+        path = self.root_dir + path
+        if backup and os.path.isfile(path):
+            os.rename(path, path + ".old")
+
+        with open(os.open(path, os.O_CREAT | os.O_WRONLY | os.O_EXCL, mode), "w",
+                  encoding=self.file_encoding) as f:
+            logging.info("Writing %s", path)
+            f.write(u"" + content)
```
fernflower: f.write(unicode(content))?

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,670 @@
...
+        convert_servers = set()
+        for zone, pool in pools.items():
+            pool.sort()
+            if [number for number, source in pool] != ["0", "1", "2", "3"]:
+                continue
+            for i in range(1, len(pool)):
+                if pool[0][1]["options"] != pool[i][1]["options"]:
+                    break
+            else:
+                assert len(pool) > 2
```
fernflower: Unfortunately asserts are against leapp guidelines https://github.com/oamg/leapp-guidelines/blob/master/python-coding-guidelines.md

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: Similar, but faster for wide ranges and maybe slightly more readable.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: Makes sense. I'll fix that.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: Is there a difference?

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: The guideline just says asserts shouldn't mutate (`len(pool) > 2` certainly doesn't).

bocekm:
@mlichvar, we are changing license of this repository from AGPL 3.0 to Apache 2.0 under PR https://github.com/oamg/leapp-repository/pull/112. Is it ok with you?

mlichvar:
> @mlichvar, we are changing license of this repository from AGPL 3.0 to Apache 2.0 under PR #112. Is it ok with you?

Fine with me. I guess I need to relicense the ntp2chrony script?

pirat89:
@mlichvar Regarding the [Fedora Licensing](https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing) page, ASL 2.0 is not compatible with GPLv2.

bocekm:
@mlichvar, yes please. Apache 2.0 is not compatible with GPL 2.0. Either change the license to a compatible one or remove the license notice completely if you agree with Apache 2.0 for the script.

I think you can also remove the `#!/usr/bin/python` and `if __name__ == "__main__":
    main()`.

mlichvar:
Relicensing only the leapp copy would mean it couldn't get any later improvements from upstream (made by other people). I think I'll change the license to MIT, so it stays compatible with GPLv2.

bocekm:
And where's the upstream copy of the script?

mlichvar:
It's here: https://github.com/mlichvar/ntp2chrony/blob/master/ntp2chrony.py

bocekm:
Ok, yeah. That will work. Changing it to MIT on both places, here and in the upstream.
By the way, in the upstream [you changed](https://github.com/mlichvar/ntp2chrony/commit/dcf7a05070ebdf7bbe1d59f3a53c68e8795c3d5f) the license to [2-Clause BSD](https://opensource.org/licenses/BSD-2-Clause), not [MIT](https://opensource.org/licenses/MIT).

mlichvar:
Oops. I'll fix that. Thanks for checking.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
fernflower: Is it a bug in time source parsing with include file https://github.com/mlichvar/ntp2chrony/blob/master/tests/data/chronyconfs/3_chrony.conf#L4? Several lines appear twice in generated config.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,676 @@
...
+    def get_processed_time_sources(self):
+        # Convert {0,1,2,3}.*pool.ntp.org servers to 2.*pool.ntp.org pools
+
+        pools = {}
+        for source in self.time_sources:
+            if source["type"] != "server":
+                continue
+            if not source["address"].endswith(".pool.ntp.org"):
+                continue
+            m = re.match("^([0123])(\\.\\w+)?\\.pool\\.ntp\\.org$", source["address"])
```
fernflower: This will add '.' to the zone name, is it intended?

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,676 @@
...
+            zone = m.group(2)
+            if zone not in pools:
+                pools[zone] = []
+            pools[zone].append((number, source))
+
+        remove_servers = set()
+        convert_servers = set()
+        for zone, pool in pools.items():
+            pool.sort()
+            if [number for number, source in pool] != ["0", "1", "2", "3"]:
```
fernflower: If there are not exactly 4 servers (0, 1, 2, 3) the check on L376-L380 won't happen. Is it intended?

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,676 @@
...
+            number = m.group(1)
+            zone = m.group(2)
+            if zone not in pools:
+                pools[zone] = []
+            pools[zone].append((number, source))
+
+        remove_servers = set()
+        convert_servers = set()
+        for zone, pool in pools.items():
+            pool.sort()
```
fernflower: Do you mean lexicographic check here, or are you missing int conversion for pool[0]?

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,676 @@
...
+                                    "noquery" not in self.restrictions[n] and
+                                    n != ipaddress.ip_network(u"127.0.0.1/32") and
+                                    n != ipaddress.ip_network(u"::1/128"),
+                                  self.restrictions.keys())
+
+        ip_versions = set()
+        conf = ""
+        for network in sorted(allowed_networks, key=lambda n: (n.version, n)):
+            ip_versions.add(network.version)
+            if network.num_addresses > 1:
```
fernflower: nit: conf += "cmdallow {}\n".format(network if network.num_addresses > 1 else network.network_address)

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
fernflower: As in unnecessary string concatenation with forceful unicode convertion vs direct unicode convertion

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,676 @@
...
+            self.write_file(chrony_keys_path, 0o640, chrony_keys, backup)
+
+    def get_processed_time_sources(self):
+        # Convert {0,1,2,3}.*pool.ntp.org servers to 2.*pool.ntp.org pools
+
+        pools = {}
+        for source in self.time_sources:
+            if source["type"] != "server":
+                continue
+            if not source["address"].endswith(".pool.ntp.org"):
```
fernflower: Unnecessary check, will be covered by regex at L361

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
fernflower: L370-L396 look overcomplicated. 
Both convert_ and remove_ sets are unnecessary, you don't need convert_set and the cycle L388-L394 at all as you can change the object in place. len(pool) is 4 all the time, this cycle L376-L380 can also be substituted for an easily readable remove_set update.

mlichvar: Should be fixed now.

repos/system_upgrade/el7toel8/actors/migratentp/actor.py
```diff
@@ -0,0 +1,92 @@
...
+
+class MigrateNtp(Actor):
+    name = 'migrate_ntp'
+    description = 'Migrate ntp and/or ntpdate configuration to chrony.'
+    consumes = (NtpMigrationDecision,)
+    produces = (Report,)
+    tags = (ApplicationsPhaseTag, IPUWorkflowTag)
+
+    def extract_tgz64(self, s):
+        import base64, io, tarfile
```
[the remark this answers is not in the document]
mlichvar: Fixed.

repos/system_upgrade/el7toel8/actors/checkntp/actor.py
```diff
@@ -0,0 +1,82 @@
...
+                self.log.debug('{} is {}'.format(name, state))
+            except CalledProcessError:
+                self.log.debug('{} is not {}'.format(name, state))
+                return False
+
+        return True
+
+    def get_tgz64(self, filenames):
+        # Return a base64-encoded gzipped tarball
+        import base64, io, tarfile
```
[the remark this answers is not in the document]
mlichvar: Fixed.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,669 @@
...
+                words = words[2:]
+            else:
+                return False
+
+        self.tos_options.update(options)
+
+        return True
+
+    def parse_includefile(self, words):
+        path = self.root_dir + words[0]
```
fernflower: Please use os.path.join for all path building operations

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
fernflower: Please use os.path.join for all path building operations

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,669 @@
...
+                    .format(self.root_dir, service)):
+                self.enabled_services.add(service)
+        logging.info("Enabled services found in /etc/systemd/system: %s",
+                     " ".join(self.enabled_services))
+
+    def parse_step_tickers(self):
+        if not self.step_tickers_path:
+            return
+
+        path = self.root_dir + self.step_tickers_path
```
fernflower: Please use os.path.join for all path building operations

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,669 @@
...
+        path = self.root_dir + words[0]
+        if not os.path.isfile(path):
+            return False
+
+        self.parse_ntp_conf(path)
+        return True
+
+    def parse_keys(self, words):
+        keyfile = words[0]
+        path = self.root_dir + keyfile
```
fernflower: Please use os.path.join for all path building operations

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,669 @@
...
+            conf += "\n"
+
+        conf += "# Specify time sources.\n"
+
+        for source in self.get_processed_time_sources():
+            address = source["address"]
+            if address.startswith("127.127."):
+                if address.startswith("127.127.1."):
+                    continue
+                assert False
```
fernflower: Sorry, not getting it - what is this assert used for? What special case are you addressing by it?

shaded-enmity: +1, what would make more sense to me:
```
if not address.startswith('127.127.1.'):
   raise Exception('Some meaningful message as to why this is not correct')
```

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,669 @@
...
+                for option in source["options"]:
+                    if option[0] in ["minpoll", "maxpoll", "version", "key",
+                                     "iburst", "noselect", "prefer", "xleave"]:
+                        conf += " {}".format(" ".join(option))
+                    elif option[0] == "burst":
+                        conf += " presend 6"
+                    elif option[0] == "true":
+                        conf += " trust"
+                    else:
+                        assert False
```
shaded-enmity: Ditto here, please raise an exception with a meaningful message.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: asserts catch bugs in the code. In a bug-free code they do nothing. In this case there are two places in the code that need to be consistent with each other. The parser and the config generator. If the parser was modified to handle another refclock, but the config generator wasn't, the assert would trigger.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: I don't think a different exception would be appropriate here. I have added a comment to make it more clear what the assertion is supposed to catch.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: Fixed.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
shaded-enmity: @mlichvar This is handling user provided data, there may not be a bug in the code, but the data can be ill-formed, and you NEVER use asserts for that. Please also read the guidance about "In-your-face error handling", the point here is that these things MAY bubble up to users, and hence the very clear distinction between:
```
AssertionError: False
```
And
```
NtpMigrationError: Configuration in /some/path contains multiple refclocks
```

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
shaded-enmity: We try to provide all the necessary context to the users to fix the actual problem, why am I supposed to read code when the actual problem is in my data?

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: But that doesn't apply to this case. No user provided data can trigger that assert. The parser ignores all refclocks except 127.127.1.0, so they are listed in the generated file as ignored directives. The assert will trigger only if the parser is updated to handle other refclocks, but the generator is not, i.e. there is a bug in the logic of the code.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
mlichvar: Users will not see this. Please have a look at the actual code.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
shaded-enmity: Thanks for the clarification @mlichvar - in this case I'd suggest removing the two checks altogether and flesh them out into a unit test, because that's what it is, we will not be testing if someone made an incompatible change to he parser when running on a production system, we need to test for that before we release.

As a side note, about asserts, since I'm the one who wrote that section of the guidelines, the "mutating assert" scenario is spelled out merely as a worst case scenario, but the more fundamental takeaway is that relying on assert is simply making assumptions as to how people are executing our code. We cannot make these assumptions.

mlichvar: Right, there should be tests with non-LOCAL refclocks. And those may or may not trigger the assert depending on what bug the code actually has. Checking the output is important, but it may not catch everything.

mlichvar:
Added unit tests.

repos/system_upgrade/el7toel8/actors/migratentp/tests/unit_test.py
```diff
@@ -0,0 +1,95 @@
+from leapp.libraries.actor import library
+from leapp.libraries.common import reporting
+
+class report_generic_mocked(object):
```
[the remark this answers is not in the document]
mlichvar: I can rename the classes, but I've copied this from other tests that are already included in leapp-repository, so it would be nice to have some consistency here.

repos/system_upgrade/el7toel8/actors/migratentp/tests/unit_test.py
```diff
@@ -0,0 +1,95 @@
...
+    def __init__(self):
+        self.called = 0
+
+    def __call__(self, **report_fields):
```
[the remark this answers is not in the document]
mlichvar: No, I'm not familiar with the mock library. I've just followed the leapp recommendations for writing tests.

repos/system_upgrade/el7toel8/actors/migratentp/libraries/ntp2chrony.py
```diff
@@ -0,0 +1,671 @@
...
+
+        conf += "# Specify time sources.\n"
+
+        for source in self.get_processed_time_sources():
+            address = source["address"]
+            if address.startswith("127.127."):
+                if address.startswith("127.127.1."):
+                    continue
+                # No other refclocks are expected from the parser
+                assert False
```
shaded-enmity: @mlichvar I can still see the `assert` here (and below) and want to reiterate that the conditions that make the code reach such code paths should be rather handled by proper unit test, if we get here, it's a code error and we need to catch that early. If you need help with the tests, let me know.

mlichvar: There are tests that would trigger both asserts in the upstream repository:
https://github.com/mlichvar/ntp2chrony/blob/master/tests/data/ntpconfs/6_ntp.conf
https://github.com/mlichvar/ntp2chrony/blob/master/tests/data/ntpconfs/7_ntp.conf

mlichvar:
It looks like the code is catching a wrong CalledProcessError, after leapp has switched to its own? Should I fix the first commit in the series?

vinzenz:
@mlichvar You are right that's the reason - You can also just add another patch on top of it, as we have to squash this PR anyway.
And please when you are on it - Transform please the description fields of the Actors to class docstrings instead and try to avoid using double quotes for non docstrings.

Thanks.

fernflower:
Yeah, that's the issue. Good news is that after that is fixed the remaining part seems to be ok.
I used a different approach, by switching to a quiet mode with checked=False you implemented, but either way it is ok.
```
def check_service(name):
    for state in ['active', 'enabled']:
        res = run(['systemctl', 'is-{}'.format(state), name], checked=False)
        res_state = res['stdout'].strip()
        api.current_logger().debug('{} is {}'.format(name, res_state))
        if res_state != state:
            return False

    return True
```
You can add a commit on top or fixup the change into first commit - it doesn't matter imho as I suppose all commits will be squashed upon merge.

mlichvar:
Ok, I fixed the imports to use the right exception (with checked=False the function would always return True) and fixed the actor descriptions to use docstrings. If you want to convert ntp2chrony.py from quotes to double quote, please feel free.

Thanks for your help with this PR.

repos/system_upgrade/el7toel8/actors/checkntp/libraries/library.py
```diff
@@ -0,0 +1,70 @@
...
+            api.current_logger().debug('{} is {}'.format(name, state))
+        except CalledProcessError:
+            api.current_logger().debug('{} is not {}'.format(name, state))
+            return False
+
+    return True
+
+
+# Check if a file exists
+def is_file(name):
```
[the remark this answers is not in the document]
mlichvar: The function is mocked in the unit tests.

PR #69, opened 2019-02-12
Rework the NetworkManagerUpdateClientId actor

bengal:
Move the external tool code into a library and convert it to use D-Bus
instead of the GObject introspection library. The main reason for this
is to avoid the large dependency chain needed by python-gobject. Also,
add the python-dbus dependency.


bengal:
I could only test this locally using snactor. When I copy my actor code to the vagrant box and launch an upgrade, after the reboot leapp fails because the python-dbus dependency in RHEL8 is missing. Is there a way I can test a full upgrade with my code and with new dependencies?

pirat89:
@bengal Hi Beniamino, yes it is. However in your case it looks on some CI issue so I will have to prepare builds manually. I will send you an email with intructions in several minutes.

pirat89:
leapp-ci build

pirat89:
From the CI:
```python
17:13:28 17:13:28 Traceback (most recent call last):
17:13:28 17:13:28   File "/usr/lib64/python2.7/multiprocessing/process.py", line 258, in _bootstrap
17:13:28 17:13:28     self.run()
17:13:28 17:13:28   File "/usr/lib64/python2.7/multiprocessing/process.py", line 114, in run
17:13:28 17:13:28     self._target(*self._args, **self._kwargs)
17:13:28 17:13:28   File "/tmp/repo/tut/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 26, in inspect_actor
17:13:28 17:13:28     definition.load()
17:13:28 17:13:28   File "/tmp/repo/tut/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 135, in load
17:13:28 17:13:28     with self.injected_context():
17:13:28 17:13:28   File "/usr/lib64/python2.7/contextlib.py", line 17, in __enter__
17:13:28 17:13:28     return self.gen.next()
17:13:28 17:13:28   File "/tmp/repo/tut/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 245, in injected_context
17:13:28 17:13:28     map(lambda x: os.path.join(self._repo_dir, self.directory, x), self.libraries))
17:13:28 17:13:28   File "/tmp/repo/tut/lib/python2.7/site-packages/leapp/repository/loader.py", line 19, in library_loader
17:13:28 17:13:28     imported = importer.find_module(name).load_module(name)
17:13:28 17:13:28   File "/usr/lib64/python2.7/pkgutil.py", line 246, in load_module
17:13:28 17:13:28     mod = imp.load_module(fullname, self.file, self.filename, self.etc)
17:13:28 17:13:28   File "/tmp/repo/repos/system_upgrade/el7toel8/actors/networkmanagerupdateclientid/libraries/networkmanagerupdateclientid.py", line 1, in <module>
17:13:28 17:13:28     import dbus
17:13:28 17:13:28 ImportError: No module named dbus
```
Regarding the PR changes in spec file, probably you need to require such package for tests as well. You can do that by Makefile inside the actor directory with content like this:
```make
install-deps:
    yum install python-dbus
```
See e.g. [this one](https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/actors/setpermissiveselinux/Makefile).

packaging/leapp-el7toel8-deps.spec
```diff
@@ -18,11 +18,12 @@ URL:        https://leapp-to.github.io
 ##################################################
 %package -n %{lrdname}
 Summary:    Meta-package with system dependencies for leapp repository
-Provides:   leapp-repository-dependencies = 2
+Provides:   leapp-repository-dependencies = 3
```
pirat89: You correctly updated this capability, however, you need to update this one in the 2nd spec file as well. See `leapp-repository.spec`, line 33

repos/system_upgrade/el7toel8/actors/networkmanagerupdateclientid/libraries/networkmanagerupdateclientid.py
```diff
@@ -0,0 +1,66 @@
+import dbus
```
vinzenz: Considering that this in the ApplicationsPhase and I can't see a python2-dbus I wonder if will run into some trouble with this.

packaging/leapp-el7toel8-deps.spec
```diff
@@ -18,11 +18,12 @@ URL:        https://leapp-to.github.io
...
 Obsoletes:  leapp-repository-deps
 
 Requires:   dnf >= 4
 Requires:   pciutils
+Requires:   dbus-python
```
vinzenz: no such thing on el8. will this will work?

repos/system_upgrade/el7toel8/actors/networkmanagerupdateclientid/libraries/networkmanagerupdateclientid.py
bengal: Understood, I couldn't test it due to other issues I had with the upgrade.

So, everything that runs in ApplicationsPhase uses python2, right?

The actor needs to talk with NM to update connection profiles. The option available are:

 1. use D-Bus: we need python2-dbus, which seems not available in RHEL8

 2. use libnm through GObject-introspection: we need python2-gobject not available on RHEL8

 3. use nmcli commands: this is the worst option because it's inefficient and requires additional string	parsing

As I understand it, we're left with 3), right? Or do you think in some ways we could use one of the other two?

pirat89: @bengal We will probably switch into the python3 after the UpgradePhase is done. So you would be able to (ok , you will have to in case we will do that) use Python3 libraries in actor for  `ApplicationsPhase`. But in that case, you need to wait for implementation. Not sure when it would be resolved.

bengal: Okay, then I'll wait the switch to python 3 for the upgrade phase.

vojtechsokol:
leapp-ci build

bengal:
Closing for now, will reopen when needed.

PR #70, opened 2019-02-12
Append Leapp upgrade.log file instead of backup it

asilveir:
The previous behavior was to save only the last log file, which causes log loss if it goes into a looping process.

Now the log file will be appended.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/removeoldlog/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/removeoldlog/actor.py
@@ -1,0 +1,18 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.library import remove_log
+from leapp.tags import IPUWorkflowTag, FactsPhaseTag
+
+
+class RemoveOldLog(Actor):
+    """
+    Removes the old log from the previous Leapp run
+    to ensure that you have only valid and updated logs for debugging.
+    """
+
+    name = 'remove_old_log'
+    consumes = ()
+    produces = ()
+    tags = (IPUWorkflowTag, FactsPhaseTag)
+
+    def process(self):
+        remove_log()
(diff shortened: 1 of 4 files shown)


repos/system_upgrade/el7toel8/actors/removeoldlog/actor.py
```diff
@@ -0,0 +1,18 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.library import remove_log
+from leapp.tags import IPUWorkflowTag, FactsPhaseTag
+
+
+class RemoveOldLog(Actor):
+    """
+    Removes the old log from the previous Leapp run
```
vinzenz: 1st line should be a summary and details with one empty line in between should follow.

repos/system_upgrade/el7toel8/actors/removeoldlog/actor.py
```diff
@@ -0,0 +1,18 @@
...
+    to ensure that you have only valid and updated logs for debugging.
+    """
+
+    name = 'remove_old_log'
+    consumes = ()
+    produces = ()
+    tags = (IPUWorkflowTag, FactsPhaseTag)
```
vinzenz: I am still not sure if we should do it in the FactsPhase or in the InitramPreparationPhase

@pirat89 what do you think?

pirat89: The `InterimPreparationPhaseTag` - yep.

pirat89:
@asilveir Is it still WIP or is it prepared for review (& merge).

asilveir:
> @asilveir Is it still WIP or is it prepared for review (& merge).

It's done, I will remove WIP tag.

repos/system_upgrade/el7toel8/actors/removeoldlog/libraries/library.py
```diff
@@ -0,0 +1,17 @@
...
+    if os.path.isfile(filepath):
+        api.current_logger().info('already exists a log file from a previous Leapp run, it will be removed')
+        remove_file(filepath)
+
+
+def remove_file(filepath):
+    try:
+        os.remove(filepath)
+    except OSError as err:
+        api.current_logger().error('Could not remove {0}: {1}.'.format(filepath, err))
```
pirat89: I would prefer maybe warning instead of error in this case, but it's ok.

PR #71, opened 2019-02-12
Copy yum v3 configuration data to yum v4 /etc/dnf directory

radomil:
Yum package from RHEL7 is upgradable to RHEL8 without having to remove the yum package during the upgrade transaction
    The solution is proposed by YUM/DNF developers.
Using chroot
 1) Create /etc/dnf directory
 2) Copy yum configuration to the /etc/dnf directory

[Addressing https://github.com/oamg/leapp-repository/issues/19]

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
@@ -194,6 +194,9 @@ class PrepareUpgradeTransaction(Actor):
             json.dump(plugin_data, data)
             data.flush()
 
+        # secure yum upgradability (workaround)
+        preparetransaction.yum2dnf_configuration(overlayfs_info)
+
         # FIXME: fails on insufficient permissions
         cmd = ['--', '/bin/bash', '-c', ' '.join(dnf_command + ['/var/lib/leapp/dnf-plugin-data.txt'])]
         _unused, error = preparetransaction.guard_container_call(
@@ -283,7 +286,11 @@ class PrepareUpgradeTransaction(Actor):
 
             error = preparetransaction.copy_file_to_container(ofs_info, dnfplugin_spath, dnfplugin_dpath)
             if error:
-                return error
+                preparetransaction.produce_error(error, summary='Cannot copy rhel_upgrade.py to container.')
+                return
+        else:
+            preparetransaction.produce_error(error)
+            return
 
         error = self.dnf_plugin_rpm_download(ofs_info)
 
--- a/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/libraries/preparetransaction.py
+++ b/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/libraries/preparetransaction.py
@@ -11,7 +11,6 @@ from leapp.libraries.stdlib import api
 from leapp.libraries.stdlib import call, run
 from leapp.libraries.stdlib.call import STDOUT
 
-
 OverlayfsInfo = namedtuple('OverlayfsInfo', ['upper', 'work', 'merged'])
 
 
@@ -282,3 +281,26 @@ def create_disk_image(path):
 def remove_disk_image(path):
     guard_call(['/bin/umount', '-fl', os.path.join(path, 'mounts')])
     shutil.rmtree(path, ignore_errors=True)
+
+
+def yum2dnf_configuration(overlayfs_info):
+    """
+    Secure yum upgradability.
+
+     (workaround)
+
+    :return:
+    """
+    cmds = [
+        ['mkdir', '-p', '/etc/dnf'],
+        ['cp', '-af', '/etc/yum/*', '/etc/dnf/'],
+        ['rm', '-rf', '/etc/yum/pluginconf.d', '/etc/yum/protected.d', '/etc/yum/vars'],
+        ['ln', '-s', '/etc/dnf/plugins/', '/etc/yum/pluginconf.d'],
+        ['ln', '-s', '/etc/dnf/protected.d/', '/etc/yum/protected.d'],
+        ['ln', '-s', '/etc/dnf/vars/', '/etc/yum/vars']
+    ]
+
+    for cmd in cmds:
+        _unused, error = guard_container_call(overlayfs_info, cmd)
+        if error:
+            produce_warning(error, summary='Yum upgradability may be affected.')
--- a/repos/system_upgrade/el7toel8/actors/yumupgradability/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/yumupgradability/actor.py
@@ -1,0 +1,25 @@
+from leapp.actors import Actor
+from leapp.tags import IPUWorkflowTag, ExperimentalTag, PreparationPhaseTag
+from leapp.libraries.actor.yumupgradability import secure_yum_upgradability
+
+
+class YumUpgradability(Actor):
+    """
+    Copy yum v3 configuration to yum v4 /etc/dnf directory
+
+    Steps
+        1) Create /etc/dnf directory
+        2) Copy yum configuration to the /etc/dnf directory
+        3) Make symlinks under /etc/yum
+    """
+
+    name = 'yum_upgradability'
+    consumes = ()
+    produces = ()
+    tags = (IPUWorkflowTag, PreparationPhaseTag)
+
+    def process(self):
+        """
+        Process yum upgradability actor
+        """
+        secure_yum_upgradability()
--- a/repos/system_upgrade/el7toel8/actors/yumupgradability/libraries/yumupgradability.py
+++ b/repos/system_upgrade/el7toel8/actors/yumupgradability/libraries/yumupgradability.py
@@ -1,0 +1,43 @@
+from leapp.libraries.stdlib import api
+from leapp.libraries.stdlib import CalledProcessError
+from leapp.libraries.stdlib import run
+
+CMDS = [
+    ['mkdir', '-p', '/etc/dnf'],
+    ['cp', '-af', '/etc/yum/*', '/etc/dnf/'],
+    ['rm', '-rf', '/etc/yum/pluginconf.d', '/etc/yum/protected.d', '/etc/yum/vars'],
+    ['ln', '-s', '/etc/dnf/plugins/', '/etc/yum/pluginconf.d'],
+    ['ln', '-s', '/etc/dnf/protected.d/', '/etc/yum/protected.d'],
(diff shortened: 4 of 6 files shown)


repos/system_upgrade/el7toel8/actors/yumupgradability/actor.py
```diff
@@ -0,0 +1,43 @@
+from leapp.actors import Actor
+from leapp.models import Inhibitor
+from leapp.tags import IPUWorkflowTag, ExperimentalTag, InitRamStartPhaseTag
+from leapp.libraries.stdlib import call
+
+
+class YumUpgradability(Actor):
+    """
+    An actor that takes care of yum upgradability.
```
vinzenz: Please be more specific what it odes. Basically what you describe down there in secure_yum_upgradability would be more suitable here

repos/system_upgrade/el7toel8/actors/yumupgradability/actor.py
```diff
@@ -0,0 +1,43 @@
...
+        :return: bool
+        """
+        # Step 1
+        call(['mkdir', '/etc/dnf'])
+        # Step 2
+        call(['cp', '-a', '/etc/yum/*', '/etc/dnf/'])
+        # Step 3
+        call(['rm', '/etc/yum/*', '-rf'])
+
+        return True
```
vinzenz: Why do you return True here?

repos/system_upgrade/el7toel8/actors/yumupgradability/actor.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def process(self):
+        """
+        Process yum upgradability actor
+
+        """
+        if not self.secure_yum_upgradability():
```
vinzenz: You need to handle the Exceptions that can come through call OSError, CalledProcessError

repos/system_upgrade/el7toel8/actors/yumupgradability/actor.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class YumUpgradability(Actor):
+    """
+    An actor that takes care of yum upgradability.
+    """
+
+    name = 'yum_upgradability'
+    consumes = ()
+    produces = (Inhibitor,)
```
pirat89: Inhibitor doesn't make sense.
```suggestion
```

repos/system_upgrade/el7toel8/actors/yumupgradability/actor.py
```diff
@@ -0,0 +1,43 @@
...
+
+        This is done in 3 steps
+            1) Create /etc/dnf directory
+            2) Copy yum data to the /etc/dnf directory
+            3) Remove all yum data from the /etc/yum directory
+
+        :return: bool
+        """
+        # Step 1
+        call(['mkdir', '/etc/dnf'])
```
pirat89:
```suggestion
        call(['mkdir', ''-p, '/etc/dnf'])
```
But I prefer more using `os.mkdir` function for this step.

repos/system_upgrade/el7toel8/actors/yumupgradability/actor.py
pirat89:
```suggestion
        return
```

repos/system_upgrade/el7toel8/actors/yumupgradability/actor.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    An actor that takes care of yum upgradability.
+    """
+
+    name = 'yum_upgradability'
+    consumes = ()
+    produces = (Inhibitor,)
+    tags = (IPUWorkflowTag, InitRamStartPhaseTag, ExperimentalTag)
+
+    def secure_yum_upgradability(self):
```
pirat89: As we discussed with @vinzenz, the best would be move of this function into the tool - means, create sh script that will do these 3 operations. Additionally, it should be tool usable by any actor in the repository, so it should be under `repos/system_upgrade/el7toel8/tools` directory (@vinzenz  correct me in case I missed something)

repos/system_upgrade/el7toel8/actors/yumupgradability/actor.py
pirat89: If I am right, there is LeappCommandError (or something like that) that should be used here, right?

repos/system_upgrade/el7toel8/actors/yumupgradability/actor.py
```diff
@@ -0,0 +1,43 @@
...
+            3) Remove all yum data from the /etc/yum directory
+
+        :return: bool
+        """
+        # Step 1
+        call(['mkdir', '/etc/dnf'])
+        # Step 2
+        call(['cp', '-a', '/etc/yum/*', '/etc/dnf/'])
+        # Step 3
+        call(['rm', '/etc/yum/*', '-rf'])
```
pirat89: This particular step will have additional consquences, but which are minimalized enough in case it will be moved into the  phase mentioned above, so I could live with that).

repos/system_upgrade/el7toel8/actors/yumupgradability/actor.py
```diff
@@ -0,0 +1,39 @@
...
+        2) Copy yum data to the /etc/dnf directory
+        3) Make symlinks under /etc/yum
+    """
+
+    name = 'yum_upgradability'
+    consumes = ()
+    produces = ()
+    tags = (IPUWorkflowTag, PreparationPhaseTag, ExperimentalTag)
+
+    script_path = '../../tools/copy_yum_v3_configuration.sh'
```
pirat89: @vinzenz There is API function that returns correct path, isn't?

radomil: Removed the relative path. As this should be in the executable path.

repos/system_upgrade/el7toel8/actors/yumupgradability/actor.py
radomil: @pirat89 I've received @vinzenz 's confirmation for `repos/system_upgrade/el7toel8/tools`.

asilveir:
LGTM

radomil:
@pirat89 , may I ask you to check?

> Still there is not applied change inside the container to be able to calculate transaction with the same settings (otherwise this will not resolve the problem as test of the transaction will fail (I guess). But still the point is same, for calculation of the RPM transaction, the env has to be set up proplerly.

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
```diff
@@ -264,6 +264,20 @@ def process(self):
...
+        cmd = [syu_dpath + 'copy_yum_v3_configuration.sh']
+        _unused, error = preparetransaction.guard_container_call(
+            ofs_info,
+            cmd,
+            guards=(preparetransaction.connection_guard(),)
+        )
+        if error:
+            error.summary = 'Could not secure yum upgradability.'
+            return error
+
```
pirat89: Only thing I can see is broken upgrade, even without properly produced  err message. There is not produced any error message... - and the `process` method should not return anything....

pirat89:
Additional note: of course, regarding the previous review, tests are missing or they are wrong - otherwise this would be caught by CI automatically. So changes should be covered by tests that will proove the solution is working.

vojtechsokol:
leapp-ci build

vojtechsokol:
e2e tests

repos/system_upgrade/el7toel8/actors/yumupgradability/libraries/yumupgradability.py
```diff
@@ -0,0 +1,43 @@
...
+
+    :return: Yellow Flag flag
+    """
+
+    yellow_flag = False
+    for cmd in CMDS:
+        if run_cmd(cmd):
+            yellow_flag = True
+
+    return yellow_flag
```
jmikovic: what's the point of this if you are not using the return value anywhere?

repos/system_upgrade/el7toel8/actors/yumupgradability/tests/test_yumupgradability.py
```diff
@@ -0,0 +1,17 @@
...
+    Test Secure Yum upgradability library function
+
+    :return: Indicates Pass/Fail - whether tests succeeded
+    """
+
+    monkeypatch.setattr(yumupgradability, "CMDS", [['whoami'], ['ls']])
+    monkeypatch.setattr(yumupgradability, 'run_cmd', call)
+    expected_yellow_flag = False
+    actual_yellow_flag = yumupgradability.secure_yum_upgradability()
+    assert actual_yellow_flag == expected_yellow_flag
```
jmikovic: could be written directly as `assert yumupgradability.secure_yum_upgradability() == False`

I would expect some tests where the function returns the flag

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/libraries/preparetransaction.py
```diff
@@ -282,3 +281,26 @@ def create_disk_image(path):
...
+        ['rm', '-rf', '/etc/yum/pluginconf.d', '/etc/yum/protected.d', '/etc/yum/vars'],
+        ['ln', '-s', '/etc/dnf/plugins/', '/etc/yum/pluginconf.d'],
+        ['ln', '-s', '/etc/dnf/protected.d/', '/etc/yum/protected.d'],
+        ['ln', '-s', '/etc/dnf/vars/', '/etc/yum/vars']
+    ]
+
+    for cmd in cmds:
+        _unused, error = guard_container_call(overlayfs_info, cmd)
+        if error:
+            produce_warning(error, summary='Yum upgradability may be affected.')
```
jmikovic: was this discussed somewhere, that these errors should be only reported and upgrade should go on? If so this should at least be reported to the user, not only logged

repos/system_upgrade/el7toel8/actors/yumupgradability/libraries/yumupgradability.py
radomil: I wanted a self-explanatory variable which bears a return value that is used in unit tests, at the moment.

repos/system_upgrade/el7toel8/actors/yumupgradability/tests/test_yumupgradability.py
radomil: You're right. I just wanted self-explanatory variables which refer to the return value of the tested function.

pirat89:
Will be resolved in separate PR.

PR #73, opened 2019-02-13
Move the filterrpmtransactionevents actor to ChecksPhase

pirat89:
The actor was originally in the FactsPhase phase. However, this is confusing as
 - it is supposed to consume RpmTransactionTasks msgs that itself suggest by name that these are the specific one that user should produce when they want to influence RPM transaction
 - the name itelf is check_rpm_transaction_events

*NOTE:* If you do not like this fix, we should discuss additional changes like names of related models and name of the actor as current state is definitely wrong, but I am not telling that this is the solution we have to apply. Feel free to write your opinion about this guys.


pirat89:
The related issue: #78

PR #74, opened 2019-02-14
precompile python files to avoid left over pyc files

jmikovic:
Python files outside /usr/lib(64)/pythonX.Y/ are not byte-compiled
automatically which causes pyc files to be left over after update
or removal of leapp-repository. These left over files are then
loaded by leapp and can potentially cause bugs.
https://fedoraproject.org/wiki/Changes/No_more_automagic_Python_bytecompilation

### DIFF
--- a/packaging/leapp-repository.spec
+++ b/packaging/leapp-repository.spec
@@ -2,23 +2,35 @@
 %global repositorydir %{leapp_datadir}/repositories
 %global custom_repositorydir %{leapp_datadir}/custom-repositories
 
-Name:       leapp-repository
-Version:    0.5.0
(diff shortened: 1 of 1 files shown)


pirat89:
@jmikovic I spent some time to get info why this is [needed](https://fedoraproject.org/wiki/Changes/No_more_automagic_Python_bytecompilation). Would you add additional info (e.g. the link) into the commit message so it will be clear why we are doing that (maybe makes sense to put some small info (e.g. link again) into the specfile as well. That this is only about python files out of the `/usr/lib(64)?/pythonX.Y/` path.

PR #75, opened 2019-02-14
Increase diskimage size

Rezney:
In recent testing just encountered a scenario where our overlayfs2 disk image size was not large enough and got:

```Error Summary
-------------
Disk Requirements:
   At least 75MB more space needed on the / filesystem.
```

This is just a temporary workaround and we should think about
more robust solution and how to count total needed size for all
packages installation.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/libraries/preparetransaction.py
+++ b/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/libraries/preparetransaction.py
@@ -246,7 +246,16 @@ def create_disk_image(path):
             summary='Error while trying to create destination path on the host system.',
             details=str(e))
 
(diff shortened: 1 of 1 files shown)


Rezney:
After discussion with @vinzenz realized that we are using bind mount. In this case we need to find out why this happens and increasing disk image size helps.

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/libraries/preparetransaction.py
```diff
@@ -246,7 +246,16 @@ def create_disk_image(path):
             summary='Error while trying to create destination path on the host system.',
             details=str(e))
 
-    cmds = [['/bin/dd', 'if=/dev/zero', 'of={}'.format(diskimage_path), 'bs=1M', 'count=1500'],
+    disk_size = os.getenv('LEAPP_OVL_SIZE', default='2048')
+
```
radomil: some trailing whitespace, they say..

PR #77, opened 2019-02-14
Add actor responsible for repository mapping

artmello:


artmello:
Right now, repomap.csv is empty while we discuss internally about it contents.

artmello:
After discussion with @pirat89, we added the info that was previous hardcoded at SetupTargetRepos actor to repomap.csv. This will guarantee that upgrade process will not be broken by now.

repos/system_upgrade/el7toel8/actors/repositoriesmapping/actor.py
```diff
@@ -0,0 +1,36 @@
...
+
+        with open(self.get_file_path('repomap.csv')) as f:
+            data = csv.reader(f)
+            for row in data:
+                if len(row) != 6:
+                    continue
+                
+                from_uid, to_uid, from_minor_version, to_minor_version, arches, repo_type = row
+
+                repos_maps.repositories.append(RepositoryMap(from_uid=from_uid,
```
bocekm: If I'm not mistaken (and I tried to google it), it is not repository **UID** but **ID**. I changed it in the source sheet. UID from what I know stands for user ID. Please change it to _id_ wherever you use _uid_.

repos/system_upgrade/el7toel8/models/repositoriesmap.py
```diff
@@ -0,0 +1,18 @@
...
+
+class RepositoryMap(Model):
+    topic = TransactionTopic
+
+    from_uid = fields.String()
+    to_uid = fields.String()
+    from_minor_version = fields.String()
+    to_minor_version = fields.String()
+    arches = fields.String()
+    repo_type = fields.String()
```
bocekm: I would limit it to "rpm, srpm, debuginfo".
```
repo_type = fields.StringEnum(choices=['rpm', 'srpm', 'debuginfo'])
```

repos/system_upgrade/el7toel8/models/repositoriesmap.py
```diff
@@ -0,0 +1,18 @@
+from leapp.models import Model, fields
+from leapp.topics import TransactionTopic 
+
+class RepositoryMap(Model):
```
bocekm: Provide a docstring here, e.g. "Mapping of RHEL 7 repository ID to RHEL 8 repository to determine which RHEL 8 repository to enable for the upgrade."

repos/system_upgrade/el7toel8/actors/setuptargetrepos/actor.py
```diff
@@ -30,9 +32,21 @@ def process(self):
         for repo in self.consume(CustomTargetRepository):
```
bocekm: Remove the TODO 6 lines above.

repos/system_upgrade/el7toel8/actors/repositoriesmapping/actor.py
pirat89: @bocekm I guess in the meaning here it is unique ID. I will try to find that again, but I believe the UID it is used here. In same source it is called even "label" and not ID.

pirat89: It's possible that I proposed UID as I found in the past references like: unique repository ID (where UID is more accurate from my point of view). However, in other source is used just "repo ID" (e.g. `yum repolist all`). So not sure what to use, but in case we want to use something else then UID, it should be changed everywhere - see `targetrepositories.py`. So the same thing will not use various names. So you can
- rename it to something else - then use same naming everywhere
- keep it as it is now and create new PR to change this everywhere at once later
- you can rename it here and create new PR for already existing code

..whatever.

artmello: I am not marking this one as resolved, but I followed Michal suggestion and renamed UID to ID for this PR. I did it mostly since "repo ID" is used inside yum. But, as Petr mentioned, there are many places where UID is used. Let me know if I should keep ID for this PR and rename everything else in a follow up PR or something else.

bocekm: Yeah, keep ID in this PR and create a new PR to change it on the other places. Thanks.

repos/system_upgrade/el7toel8/models/repositoriesfacts.py
```diff
@@ -5,6 +5,7 @@
 class RepositoryData(Model):
     topic = SystemFactsTopic
 
+    uid = fields.String()
```
bocekm: Change the uid here to repoid.

repos/system_upgrade/el7toel8/actors/setuptargetrepos/tests/test_setuptargetrepos.py
```diff
@@ -0,0 +1,44 @@
+from leapp.models import CustomTargetRepository, TargetRepositories, RepositoryData, \
+    RepositoryFile, RepositoriesFacts, RepositoryMap, RepositoriesMap
+
+
+def test_minimal_execution(current_actor_context):
+    current_actor_context.run()
+
+
+def test_custom_repos(current_actor_context):
+    custom = CustomTargetRepository(uid='rhel-7-server-rpms',
```
bocekm: I know this is just a test and the content is not that important. But in case somebody wants to get inspiration from tests about the use of some functionality, as I sometimes do, I'd change the RHEL 7 to RHEL 8 repos. Because I was a bit confused as to why to enable RHEL 7 repository during the upgrade to RHEL 8.

bocekm:
Squash the commits please.

PR #79, opened 2019-02-19
Add actor and functions for detecting and fixing syntax error in grub…

jmikovic:
… configuration

There are images with syntax error in GRUB_CMDLINE_LINUX value of grub config,
which causes issues with booting, rhel8 kernel grub record and other issues.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
@@ -51,4 +51,4 @@ def fix_grub_config_error(conf_file):
 
     config = config.replace(original_value, new_value)
 
-    return write_to_file(conf_file, config)
+    write_to_file(conf_file, config)
--- a/repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
@@ -1,4 +1,5 @@
 import os
+import re
 
 from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.stdlib import api, call
@@ -30,3 +31,24 @@ def get_boot_file_paths():
                                       details={'details': 'Did not receive a message about the leapp-provided'
                                                           'kernel and initramfs'})
     return boot_content.kernel_path, boot_content.initram_path
+
+
+def write_to_file(filename, content):
+    with open(filename, 'w') as f:
+        f.write(content)
+
+
+def fix_grub_config_error(conf_file):
+    with open(conf_file, 'r') as f:
+        config = f.read()
+
+    # move misplaced '"' to the end
+    pattern = r'GRUB_CMDLINE_LINUX=.+?(?=GRUB|\Z)'
+    original_value = re.search(pattern, config, re.DOTALL).group()
+    parsed_value = original_value.split('"')
+    new_value = '{KEY}"{VALUE}"{END}'.format(KEY=parsed_value[0], VALUE=''.join(parsed_value[1:]).rstrip(),
+                                             END=original_value[-1])
+
+    config = config.replace(original_value, new_value)
+
+    return write_to_file(conf_file, config)
--- a/repos/system_upgrade/el7toel8/actors/addupgradebootentry/tests/unit_test.py
+++ b/repos/system_upgrade/el7toel8/actors/addupgradebootentry/tests/unit_test.py
@@ -10,6 +10,11 @@ class call_mocked(object):
         self.args = args
 
 
+class write_to_file_mocked(object):
+    def __call__(self, filename, content):
+        self.content = content
+
+
 def test_add_boot_entry(monkeypatch):
     def get_boot_file_paths_mocked():
         return '/abc', '/def'
@@ -44,10 +49,8 @@ def test_get_boot_file_paths(monkeypatch):
 
 
 def test_fix_grub_config_error(monkeypatch):
-    def write_to_file_mocked(filename, content):
-        return content
-    monkeypatch.setattr(library, 'write_to_file', write_to_file_mocked)
-    fixed = library.fix_grub_config_error('files/grub_test.wrong')
+    monkeypatch.setattr(library, 'write_to_file', write_to_file_mocked())
+    library.fix_grub_config_error('files/grub_test.wrong')
 
     with open('files/grub_test.fixed') as f:
-        assert fixed == f.read()
+        assert library.write_to_file.content == f.read()
--- a/repos/system_upgrade/el7toel8/actors/addupgradebootentry/tests/unit_test.py
+++ b/repos/system_upgrade/el7toel8/actors/addupgradebootentry/tests/unit_test.py
@@ -1,7 +1,6 @@
 import pytest
 
 from leapp.exceptions import StopActorExecutionError
-from leapp.libraries import stdlib
 from leapp.libraries.actor import library
 from leapp.models import BootContent
 
@@ -42,3 +41,13 @@ def test_get_boot_file_paths(monkeypatch):
 
     with pytest.raises(StopActorExecutionError):
         library.get_boot_file_paths()
+
+
+def test_fix_grub_config_error(monkeypatch):
+    def write_to_file_mocked(filename, content):
+        return content
+    monkeypatch.setattr(library, 'write_to_file', write_to_file_mocked)
+    fixed = library.fix_grub_config_error('files/grub_test.wrong')
+
+    with open('files/grub_test.fixed') as f:
+        assert fixed == f.read()
--- a/repos/system_upgrade/el7toel8/actors/detectgrubconfigerror/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/detectgrubconfigerror/actor.py
@@ -1,0 +1,29 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.scanner import detect_config_error
+from leapp.models import CheckResult, GrubConfigError
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class DetectGrubConfigError(Actor):
+    """
+    Check grub configuration for syntax error in GRUB_CMDLINE_LINUX value.
+    """
+
+    name = 'detect_grub_config_error'
+    consumes = ()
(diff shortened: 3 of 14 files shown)


repos/system_upgrade/el7toel8/actors/detectgrubconfigerror/actor.py
```diff
@@ -0,0 +1,29 @@
...
+    def process(self):
+        error_detected = detect_config_error('/etc/default/grub')
+        if error_detected:
+            self.produce(CheckResult(
+                severity='Info',
+                result='Fixed',
+                summary='Syntax error detected in grub configuration.',
+                details='Syntax error was detected in GRUB_CMDLINE_LINUX value of grub configuration. '
+                        'This error is causing booting and other issues. '
+                        'Error is autamatically fixed by fix_grub_config_error actor.'
```
pirat89:
```suggestion
                        'Error is automatically fixed by fix_grub_config_error actor.'
```

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
```diff
@@ -4,14 +4,14 @@
 from leapp.actors import Actor
 from leapp.libraries.actor import preparetransaction
 from leapp.models import FilteredRpmTransactionTasks, OSReleaseFacts, TargetRepositories
-from leapp.models import UsedTargetRepositories, UsedTargetRepository
+from leapp.models import UsedTargetRepositories, UsedTargetRepository, GrubConfigError
```
pirat89: Wrong place to fix the grub :) Discussed in person.

repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
```diff
@@ -30,3 +30,21 @@ def get_boot_file_paths():
                                       details={'details': 'Did not receive a message about the leapp-provided'
                                                           'kernel and initramfs'})
     return boot_content.kernel_path, boot_content.initram_path
+
+
+def fix_grub_config_error():
+    lines = []
```
bocekm: You don't need to initialize it when you set it later by `lines = f.readlines()`.

repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
```diff
@@ -30,3 +30,21 @@ def get_boot_file_paths():
...
+    conf_file = '/etc/default/grub'
+    with open(conf_file, 'r') as f:
+        lines = f.readlines()
+        for i, line in enumerate(lines):
+            if not line.startswith('GRUB_CMDLINE_LINUX'):
```
bocekm: This won't work in all cases. Per https://www.gnu.org/software/grub/manual/grub/html_node/Simple-configuration.html the content of `/etc/default/grub` "must be valid POSIX shell input" which is to be possible to source by a shell script.
Thus, the valid content of the file could be also just a one-liner:
```
GRUB_DEFAULT=saved GRUB_DISABLE_SUBMENU=true GRUB_TERMINAL_OUTPUT="console" GRUB_CMDLINE_LINUX="vconsole.font=latarcyrheb-sun16 $([ -x /usr/sbin/rhcrashkernel-param ] && /usr/sbin/rhcrashkernel-param || :) vconsole.keymap=us rhgb quiet" GRUB_DISABLE_RECOVERY="true"
```
Source: https://docs.fedoraproject.org/en-US/Fedora/23/html/Multiboot_Guide/GRUB-configuration.html

jmikovic:
@bocekm I probably misunderstood 'it will just be a sequence of ‘KEY=value’ lines' as one pair per line, 
but I think updating regex to take this into account shouldn't be a problem

jmikovic:
@bocekm @pirat89 I made the changes so that it works with single-line as well as multi-line config

repos/system_upgrade/el7toel8/actors/detectgrubconfigerror/libraries/scanner.py
```diff
@@ -0,0 +1,14 @@
...
+def detect_config_error(conf_file):
+    '''
+    Checks grub configuration for syntax error in GRUB_CMDLINE_LINUX value.
+
+    :return: Function returns True if error was detected, otherwise False.
+    '''
+    with open(conf_file, 'r') as f:
+        config = f.read()
+
+    pattern = 'GRUB_CMDLINE_LINUX="[^"]+"\s(?=[^A-Z]+)'
```
bocekm:
```suggestion
    pattern = r'GRUB_CMDLINE_LINUX="[^"]+"(?!(\s*$)|(\s+GRUB))'
```
Your regex incorrectly marked the following as **having** syntax error:
```
GRUB_TERMINAL_OUTPUT="console"
GRUB_CMDLINE_LINUX="console=tty0 crashkernel=auto console=ttyS0,115200n8 no_timer_check net.ifnames=0" 
GRUB_DISABLE_RECOVERY="true"
```
Or, the regex incorrectly marked the following as **not** having syntax error:
```
GRUB_TERMINAL_OUTPUT="console" GRUB_CMDLINE_LINUX="vconsole.font=latarcyrheb-sun16 vconsole.keymap=us "rhgb quiet GRUB_DISABLE_RECOVERY="true"
```
Please add these two files to the tests.

repos/system_upgrade/el7toel8/actors/detectgrubconfigerror/actor.py
```diff
@@ -0,0 +1,29 @@
...
+    def process(self):
+        error_detected = detect_config_error('/etc/default/grub')
+        if error_detected:
+            self.produce(CheckResult(
+                severity='Info',
+                result='Fixed',
+                summary='Syntax error detected in grub configuration.',
+                details='Syntax error was detected in GRUB_CMDLINE_LINUX value of grub configuration. '
+                        'This error is causing booting and other issues. '
+                        'Error is automatically fixed by fix_grub_config_error actor.'
```
bocekm: There's not such an actor. What about moving production of the message to actually the place where it is fixed?

repos/system_upgrade/el7toel8/actors/detectgrubconfigerror/libraries/scanner.py
```diff
@@ -0,0 +1,14 @@
...
+def detect_config_error(conf_file):
+    '''
+    Checks grub configuration for syntax error in GRUB_CMDLINE_LINUX value.
+
+    :return: Function returns True if error was detected, otherwise False.
+    '''
+    with open(conf_file, 'r') as f:
+        config = f.read()
+
+    pattern = 'GRUB_CMDLINE_LINUX="[^"]+"(?!(\s*$)|(\s+GRUB))'
```
bocekm:
```suggestion
    pattern = r'GRUB_CMDLINE_LINUX="[^"]+"(?!(\s*$)|(\s+GRUB))'
```
Without the _r_ marking the string as regex, Python tries to escape whatever is after _\\_, however \s is not a valid escape sequence ([flake8 W605](https://lintlyci.github.io/Flake8Rules/rules/W605.html)).

repos/system_upgrade/el7toel8/actors/detectgrubconfigerror/actor.py
```diff
@@ -0,0 +1,29 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.scanner import detect_config_error
+from leapp.models import CheckResult, GrubConfigError
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class DetectGrubConfigError(Actor):
+    """
+    Checks grub configuration for syntax error in GRUB_CMDLINE_LINUX value.
```
bocekm:
```suggestion
    Check grub configuration for syntax error in GRUB_CMDLINE_LINUX value.
```
Use command/imperative style, see [PEP257](https://www.python.org/dev/peps/pep-0257/):

> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

The same in the other docstrings.

repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
```diff
@@ -30,3 +31,21 @@ def get_boot_file_paths():
...
     return boot_content.kernel_path, boot_content.initram_path
+
+
+def fix_grub_config_error():
+    conf_file = '/etc/default/grub'
+    with open(conf_file, 'r') as f:
+        config = f.read()
+
+    # move misplaced '"' to the end
+    pattern = 'GRUB_CMDLINE_LINUX=.+(?=GRUB)'
```
bocekm: This won't work in case `GRUB_CMDLINE_LINUX` is the last envvar in the file.

bocekm: You can add a new test file for example with this content:
```
GRUB_TERMINAL_OUTPUT="console" GRUB_CMDLINE_LINUX="vconsole.font=latarcyrheb-sun16 vconsole.keymap=us "rhgb quiet
```
Also, it would be great to have tests that try to fix the test files with the syntax error.

repos/system_upgrade/el7toel8/actors/addupgradebootentry/tests/unit_test.py
```diff
@@ -42,3 +43,9 @@ def consume_no_message_mocked(*models):
 
     with pytest.raises(StopActorExecutionError):
         library.get_boot_file_paths()
+
+
+def test_fix_grub_config_error():
+    library.fix_grub_config_error('files/grub_test.wrong')
```
bocekm: Crrently the test changes the test file. Ideally, unit tests should not be changing any file when executed. One solution is to monkeypatch file `write()` function.
It would be safer to use `monkeypatch.context()` [1] to make sure you don't break _pytest_ internals by monkeypatching `write()`.
[1] https://docs.pytest.org/en/latest/reference.html#_pytest.monkeypatch.MonkeyPatch.context

bocekm: Or, probably the easiest would be to create a file write wrapper and monkeypatch that:
```python
def write_to_file(filepath, content):
    with open(filepatch, 'w') as f:
        f.write(content)
```

PR #81, opened 2019-02-19
Documentation for metapackages dependencies

asilveir:
This document aims to inform how to proceed when is necessary to add new dependencies to the repositories.

### DIFF
--- a/docs/dependencies.md
+++ b/docs/dependencies.md
@@ -1,0 +1,56 @@
+# How to deal with dependencies in leapp-repository
+
+First, read this
+[document](https://github.com/oamg/leapp/blob/master/docs/source/dependencies.md)
+to better understand the difficulties related to package dependencies in the
+Leapp project.
+
+When talking about RHEL 7 to RHEL 8 upgrade, the goal is to cover dependencies
+of all Leapp project-related packages, including the leapp-repository packages,
+for both RHEL 7 and RHEL 8. Since the situation with dependencies of the leapp
+packages is similar to the situation with the leapp-repository dependencies,
+this document focuses on the leapp-repository specifics only.
+
+Currently there are two SPEC files for leapp-repository:
+
+- The
+[leapp-repository.spec](https://github.com/oamg/leapp-repository/blob/master/packaging/leapp-repository.spec)
+file is used to build leapp-repository packages and their dependency
+metapackage _leapp-repository-deps_ **for RHEL 7**.
+- The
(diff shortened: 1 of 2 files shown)


vojtechsokol:
When opening PR with changes only to documentation, please add label `documentation`. This way tests and copr build will be skipped, as they are not needed.

asilveir:
> When opening PR with changes only to documentation, please add label `documentation`. This way tests and copr build will be skipped, as they are not needed.

Thank you @vojtechsokol for the tip.
Next time I will do that.

docs/dependencies.md
```diff
@@ -0,0 +1,72 @@
+
+# How to deal with dependencies
+
+For work with dependencies in Leapp repositories, you need to understand the
+dependencies types and why Leapp has a special way to deal with dependencies.
```
pirat89: I would reword this as `dependencies types` is something what people will think: I understand RPM (or in general: package) dependencies. Something like:
> problem we have with dependencies itself. Read this....

docs/dependencies.md
```diff
@@ -0,0 +1,72 @@
...
+
+Read this
+[Leapp documentation](https://github.com/oamg/leapp/blob/master/docs/source/dependencies.md)
+to understand it.
+
+## What to do when dependencies needs to be changed
+
+You can add dependencies for all repositories or only for one repository
```
pirat89: Nope. There is nothing like that. Both things you described are now basically for all repositories. When we will add additional repository, we will have to rethink completely how to deal with packaging. Probably packaging will have to be hidden under repositories itself. The thing is whether you are resolving dependencies for target system or src system (system where you install  the leapp).

asilveir:
> Update the _"Do not introduce new dependencies"_ section in https://github.com/oamg/leapp-repository/blob/master/docs/best-practices.md.

@bocekm I added a "for more details" in the footer I think is enough if you don't, please let me know.

bocekm:
> @bocekm I added a "for more details" in the footer I think is enough if you don't, please let me know.

Yeah, it is. Thanks.
Is this still WIP or ready for review?

bocekm:
PR created: https://github.com/asilveir/leapp-repository/pull/3.

asilveir:
> It looks like you replaced the `dependencies.md` file by the one for best practices.



> It looks like you replaced the `dependencies.md` file by the one for best practices.

Ops, fixed.

PR #86, opened 2019-02-22
Report refactoring

Rezney:
Repository actors refactoring related to https://github.com/oamg/leapp/pull/445


Rezney:
Should be ready for review. 

Inhibitor actor tests needs further refactoring. Will consult with @bocekm.

Verify result actor and "/tmp/leapp-report.txt" will probably need further refactoring too but this can be done in other PR.

repos/system_upgrade/el7toel8/actors/scheduleselinuxrelabeling/actor.py
```diff
@@ -14,7 +15,7 @@ class ScheduleSeLinuxRelabeling(Actor):
 
     name = 'schedule_se_linux_relabelling'
     consumes = (SelinuxRelabelDecision,)
-    produces = (FinalReport,)
+    produces = ()
```
jmikovic: missing Report

repos/system_upgrade/el7toel8/actors/scheduleselinuxrelabeling/actor.py
```diff
@@ -23,10 +24,10 @@ def process(self):
                 try:
                     with open('/.autorelabel', 'w'):
                         pass
+                    report_generic(
+                        severity='low',
+                        title='SElinux scheduled for relabelling',
+                        summary='/.autorelabel file touched on root in order to schedule SElinux relabelling',
+                    )
                 except OSError as e:
-                    self.produce(FinalReport(
```
jmikovic: why are we no longer reporting error here and only logging it?

vinzenz: Yeah this should be still reported, in case something went wrong.

repos/system_upgrade/el7toel8/actors/setpermissiveselinux/actor.py
```diff
@@ -16,17 +16,12 @@ class SetPermissiveSelinux(Actor):
...
-    produces = (FinalReport,)
+    produces = ()
     tags = (FinalizationPhaseTag, IPUWorkflowTag)
 
     def process(self):
         for decision in self.consume(SelinuxPermissiveDecision):
             if decision.set_permissive:
                 success, err_msg = selinux_set_permissive()
                 if not success:
-                    self.produce(FinalReport(
```
vinzenz: Same here, why is this removed? The whole point was that we can do the post install reports too. No?

repos/system_upgrade/el7toel8/actors/verifycheckresults/actor.py
```diff
@@ -13,24 +13,26 @@ class VerifyCheckResults(Actor):
...
             self.report_error('Report Error: ' + error)
         else:
             self.log.info('Generated report at ' + report_file)
 
-        if errors:
-            for e in errors:
-                self.report_error('%s: %s: %s' % (e.severity, e.result, e.summary))
+        if inhibitors:
+            for e in inhibitors:
+                self.report_error('%s: %s: %s' % (e.title, e.severity, e.detail['summary']))
```
vinzenz: I might be wrong, but I think `e.detail['summary']` is not working as you'd expect it. If you didn't decode the json string.

repos/system_upgrade/el7toel8/actors/scheduleselinuxrelabeling/actor.py
pirat89: @jmikovic  because originally there was not actor that is consuming that message and it was idea to do something like that, when we have reporting.

vinzenz: @pirat89 Considering the way how it is now going to be implemented, it's even more encouraging now to NOT remove it, since all reports now can be collected.

repos/system_upgrade/el7toel8/actors/inhibitwhenluks/actor.py
```diff
@@ -12,17 +13,19 @@ class InhibitWhenLuks(Actor):
...
             for blk in storage_info.lsblk:
                 if blk.tp == 'crypt':
-                    self.produce(Inhibitor(
-                        summary='LUKS encrypted partition detected',
-                        details='Upgrading system with encrypted partitions is not supported',
-                        solutions='If the encrypted partition is not system one and the system '
-                                  'is not depending on it, you can remove/blacklist it from '
-                                  'the system'))
+                    report_with_remediation(
+                        severity='high',
```
pirat89: Does it make sense to set the "severity" when it is `inhibitor`? Just reading that - would be severity anything else than the highest one when we inhibit that?

repos/system_upgrade/el7toel8/actors/inhibitwhenluks/tests/test_inhibitwhenluks.py
```diff
@@ -9,7 +10,7 @@ def test_actor_with_luks(current_actor_context):
 
     current_actor_context.feed(StorageInfo(lsblk=with_luks))
     current_actor_context.run()
-    assert current_actor_context.consume(Inhibitor)
+    assert current_actor_context.consume(Report)
```
pirat89: I am worried to check just that "Report" message can be consumed. It should be checked, the upgrade is inhibited (check flags) as ;Report; != `Inhibitor`

repos/system_upgrade/el7toel8/actors/inhibitwhenluks/tests/test_inhibitwhenluks.py
```diff
@@ -19,4 +20,4 @@ def test_actor_without_luks(current_actor_context):
 
     current_actor_context.feed(StorageInfo(lsblk=without_luks))
     current_actor_context.run()
-    assert not current_actor_context.consume(Inhibitor)
+    assert not current_actor_context.consume(Report)
```
pirat89: again - Report != Inhibitor. Check flags.

repos/system_upgrade/el7toel8/actors/firewalldisable/actor.py
```diff
@@ -85,10 +81,8 @@ def process(self):
...
-            self.produce(
-                Inhibitor(
-                    summary='No message to consume',
-                    details='No decision message to consume.',
-                    solutions=None
-                    ))
+            report_generic(
+                severity='low',
+                title='No message to consume',
+                summary='No decision message to consume.')
```
pirat89: Inhibitor changed to nothing?  This is fishy; - yep, previous code looks fishy too.

repos/system_upgrade/el7toel8/actors/verifycheckresults/actor.py
Rezney: Did not tested it yet but you are probably right. Shame we do not have flag between WIP and untested :)

repos/system_upgrade/el7toel8/actors/inhibitwhenluks/actor.py
Rezney: Default severity is currently "medium" and in the framework there is a check if there is inhibitor flag and severity != high, raise an error. But this is not firm, we just got to this after discussion with @shaded-enmity. I am in favour of setting it explicitly to high.

repos/system_upgrade/el7toel8/actors/inhibitwhenluks/tests/test_inhibitwhenluks.py
Rezney: It is stated in this PR description as a known issue.

repos/system_upgrade/el7toel8/actors/inhibitwhenluks/tests/test_inhibitwhenluks.py
Rezney: It is stated in this PR description as a known issue.

repos/system_upgrade/el7toel8/actors/firewalldisable/actor.py
Rezney: Forgot the flag. Will fix it.

Rezney: Forgot the flag. Will fix it.

repos/system_upgrade/el7toel8/actors/setpermissiveselinux/actor.py
Rezney: It is nothing against the final report. However taking a look back into it this shoulds be something to keep reporting indeed. Will get it back. Applies also for the previous one.

repos/system_upgrade/el7toel8/actors/scheduleselinuxrelabeling/actor.py
Rezney: Indeed...

repos/system_upgrade/el7toel8/actors/verifycheckresults/actor.py
vinzenz: @Rezney untested by author = WIP to me

Rezney: Fair enough. Basically I just wanted to move with this a little while the framework PR is still being evaluated so we can identify possible caveats...

repos/system_upgrade/el7toel8/actors/scheduleselinuxrelabeling/actor.py
Rezney: Does adding inhibitor too make sense? As we found out we cannot log in to the system if it is not relabeled? Same for the other message...

vinzenz: @Rezney well it is not an inhibitor anymore by the definition we are using it for. Because this won’t block anything.
Now it’s a high impact problem, but what if there is another problem and this not even is executed, then that would be a case where there wouldn’t be that “inhibitor”

Tl;Dr no idea, @pirat89 what do you think?

pirat89: @vinzenz @Rezney tl;dr it should be high impact without inhibitor flag. Additionally, it is not inhibitor when we are switching into the Permissive mode - so user can do relabeling manually. You should report what user should do (required action after upgrade - as we discussed, these flags are missing yet, but can be added later; just put here something like FIXME so we will find that additional flags should be added).

Inhibitor's should be used "only" before reporting phase (discussable but as I understand inhibitor, that is message consumed by reporting actor before upgrade, it doesn't make sense to use inhibitors from that point after the reporting phase. What do you want to inhibit when even upgrade transaction has been performed already? All checks should be performed during the ChecksPhase - and that's place where you should use inhibitors. After that, we are talking much more about errors as you usually expect that things will work, you are not doing additional checks whether you can do upgrade - excluding networking that needs to be checked additionally from the initramdisk with new kernel).

pirat89:
leapp-ci build

pirat89:
leapp-ci build

repos/system_upgrade/el7toel8/actors/checkpcidevices/actor.py
```diff
@@ -0,0 +1,46 @@
+from leapp.actors import Actor
+from leapp.models import PCIDevices
+from leapp.reporting import Report
+from leapp.libraries.common.reporting import report_with_remediation
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckPCIDevices(Actor):
```
artmello: Is this actor supposed to be resurrected?

vinzenz: nope

vojtechsokol:
leapp-ci build

vojtechsokol:
leapp-ci build

repos/system_upgrade/el7toel8/actors/checkpcidevices/actor.py
Rezney: Thanks for catching this...

Rezney:
@pirat89 @vinzenz @jmikovic @artmello 
Thank you very much guys for your time on this PR. It should hopefully be ok now.

repos/system_upgrade/el7toel8/actors/removeresumeservice/actor.py
```diff
@@ -29,9 +30,9 @@ def process(self):
...
                     raise
 
-        self.produce(FinalReport(
-            severity='Info',
-            result='Pass',
-            summary='"{}" service deleted'.format(service_name),
-            details='"{}" was taking care of resuming upgrade process '
-                    'after the first reboot'.format(service_name)))
+        report_generic(
+            title='"{}" service deleted.'.format(service_name),
```
pirat89:
```suggestion
            title='"{}" service deleted'.format(service_name),
```

repos/system_upgrade/el7toel8/actors/scheduleselinuxrelabeling/actor.py
```diff
@@ -23,10 +25,17 @@ def process(self):
...
-                        severity='Error',
-                        result='Fail',
-                        summary='Could not schedule SElinux for relabelling',
-                        details=e,))
-                    self.log.critical('Could not schedule SElinux for relabelling: %s' % e)
+                    # FIXME: add an "action required" flag later
+                    report_with_remediation(
+                        title='Could not schedule SElinux for relabelling.',
```
pirat89:
```suggestion
                        title='Could not schedule SElinux for relabelling',
```

repos/system_upgrade/el7toel8/actors/removeresumeservice/actor.py
Rezney: $#&^%*&^(*&)(*

repos/system_upgrade/el7toel8/actors/scheduleselinuxrelabeling/actor.py
Rezney: $#&^%&^(&)(*

PR #89, opened 2019-02-27
Show dnf command output

asilveir:
Some changes to show DNF output

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
@@ -198,7 +198,7 @@ class PrepareUpgradeTransaction(Actor):
         cmd = ['--', '/bin/bash', '-c', ' '.join(dnf_command + ['/var/lib/leapp/dnf-plugin-data.txt'])]
         _unused, error = preparetransaction.guard_container_call(
             overlayfs_info, cmd,
-            guards=(preparetransaction.connection_guard(), preparetransaction.space_guard()))
+            guards=(preparetransaction.connection_guard(), preparetransaction.space_guard()), print_output=True)
 
         if os.environ.get('LEAPP_DEBUG', '0') == '1':
             # We want the debug data available where we would expect it usually.
--- a/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/libraries/preparetransaction.py
+++ b/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/libraries/preparetransaction.py
@@ -1,13 +1,15 @@
 import os
 import shutil
 import subprocess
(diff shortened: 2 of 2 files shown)


vinzenz:
I think it'd be worth it to look into adding a parameter into [leapp.libraries.stdlib.run](https://github.com/oamg/leapp/blob/master/leapp/libraries/stdlib/__init__.py#L121) to support this out of the box and use this then.

asilveir:
> I think it'd be worth it to look into adding a parameter into [leapp.libraries.stdlib.run](https://github.com/oamg/leapp/blob/master/leapp/libraries/stdlib/__init__.py#L121) to support this out of the box and use this then.
> 
> The logging handler then can be then configured to handle the output.

Thank you Vinny, I think it's gotten better now.

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
```diff
@@ -1,5 +1,6 @@
 import os
 import json
+import sys
```
pirat89: where is the import used?

pirat89:
@asilveirI I guess that this depends on PR https://github.com/oamg/leapp/pull/453, right?

repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/libraries/preparetransaction.py
```diff
@@ -52,9 +53,19 @@ def permission_guard():
     # FIXME: Not implemented yet. Is it even useful?
     raise NotImplementedError
 
+def _logging_handler(fd_info, buffer):
```
pirat89: Why are you implementing the function? I know, it has "_" underscore, but it's completely same function as in the framework. I would think about the rename (e.g. default_handler) in the framework so there will be something default that could be used. (what do you think guys? @vinzenz, @shaded-enmity, @fernflower)

shaded-enmity: +1 on providing something like a `standard_io_handler` for these purposes

jmikovic: can this be done in follow-up PR?
I would like this to be merged ASAP as it is quite difficult to debug errors in transaction

Rezney:
As per @jmikovic comment, I would also prefer this to be done sooner then later.

asilveir:
> missing sys import in https://github.com/oamg/leapp-repository/blob/10cdaac215fb5ff4694fc9cc43a4e7176ddcfb58/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/libraries/preparetransaction.py

Done

PR #90, opened 2019-02-28
Check if XFS is in use and influence PrepareUpgradeTransaction

artmello:
Based on the information of XFS being in use without a ftype, create or not disk image inside PrepareUpgradeTransaction actor

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkxfs/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/checkxfs/libraries/library.py
@@ -1,0 +1,66 @@
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.stdlib import api, call
+from leapp.models import StorageInfo, XFSPresence
+
+
+def check_xfs_fstab(data):
+    mountpoints = set()
+    for entry in data:
+        if entry.fs_vfstype == "xfs":
+            mountpoints.add(entry.fs_file)
+
+    return mountpoints
+
+
+def check_xfs_mount(data):
+    mountpoints = set()
+    for entry in data:
+        if entry.tp == "xfs":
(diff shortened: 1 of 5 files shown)


repos/system_upgrade/el7toel8/actors/checkxfs/libraries/library.py
```diff
@@ -0,0 +1,66 @@
...
+
+    mountpoints = fstab_data | mount_data | systemdmount_data
+    
+    xfs_presence = XFSPresence()
+    # By now, we only care for XFS without ftype in use for /var
+    for mp in ('/var', '/'):
+        if mp in mountpoints:
+            xfs_presence.present = True
+            xfs_presence.without_ftype = is_xfs_without_ftype(mp)
+            break
```
jmikovic: couldn't this be a problem if both /var and / use xfs but /var have ftype=1 and / ftype=0, as diskimage wouldn't be created in that case

pirat89:
@jmikovic: sorry. I closed your comment by mistake. Github is stucked by ... whatever and I cannot open it, show it, ... just anything right now.

repos/system_upgrade/el7toel8/actors/checkxfs/libraries/library.py
artmello: @jmikovic thanks for reviewing. I tried to change the approach now to notify other actors if either / or /var has XFS without ftype. Let me know if this handles your concern.

artmello:
@pirat89 / @jmikovic Is the "closed comment" the one about /var with ftype and / without? Or something else that needs to be handled?

jmikovic:
@artmello it was the one about /var, I just reopened it

PR #91, opened 2019-03-01
Prevent showing our "gpg-pubkey" pkg as unsigned

Rezney:
Our package "gpg-pubkey" is not signed and we are currently suggesting our users in the RedHatSignedRpmCheck actor to remove it even as it is our pkg.

This will also be needed if we would like to inhibit the upgrade with unsigned packages.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/redhatsignedrpmscanner/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/redhatsignedrpmscanner/actor.py
@@ -30,7 +30,9 @@ class RedHatSignedRpmScanner(Actor):
 
         for rpm_pkgs in self.consume(InstalledRPM):
             for pkg in rpm_pkgs.items:
-                if any(key in pkg.pgpsig for key in RH_SIGS):
+                # "gpg-pubkey" is not signed as it would require another package to verify its signature
+                if any(key in pkg.pgpsig for key in RH_SIGS) or \
+                        (pkg.name == 'gpg-pubkey' and pkg.packager.startswith('Red Hat, Inc.')):
                     signed_pkgs.items.append(pkg)
                     continue
 
--- a/repos/system_upgrade/el7toel8/actors/rpmscanner/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/rpmscanner/actor.py
@@ -22,18 +22,20 @@ class RpmScanner(Actor):
             '/bin/rpm',
             '-qa',
             '--queryformat',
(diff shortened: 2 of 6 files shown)


vinzenz:
@Rezney you forgot to update the tests

repos/system_upgrade/el7toel8/actors/rpmscanner/actor.py
```diff
@@ -22,18 +22,20 @@ def process(self):
             '/bin/rpm',
             '-qa',
             '--queryformat',
-            r'%{NAME}|%{VERSION}|%{RELEASE}|%|EPOCH?{%{EPOCH}}:{}||%|ARCH?{%{ARCH}}:{}||%|DSAHEADER?{%{DSAHEADER:pgpsig}}:{%|RSAHEADER?{%{RSAHEADER:pgpsig}}:{(none)}|}|\n'
+            r'%{NAME}|%{VERSION}|%{RELEASE}|%|EPOCH?{%{EPOCH}}:{(none)}||%|PACKAGER?{%{PACKAGER}}:{(none)}||%|'
+            'ARCH?{%{ARCH}}:{}||%|DSAHEADER?{%{DSAHEADER:pgpsig}}:{%|RSAHEADER?{%{RSAHEADER:pgpsig}}:{(none)}|}|\n'
```
bocekm: For the sake of consistency, add the r prefix also to this string.

repos/system_upgrade/el7toel8/actors/redhatsignedrpmscanner/actor.py
```diff
@@ -30,7 +30,8 @@ def process(self):
 
         for rpm_pkgs in self.consume(InstalledRPM):
             for pkg in rpm_pkgs.items:
-                if any(key in pkg.pgpsig for key in RH_SIGS):
+                if any(key in pkg.pgpsig for key in RH_SIGS) or \
+                    (pkg.name == 'gpg-pubkey' and pkg.packager.startswith('Red Hat, Inc.')):
```
bocekm: Indent by additional 4 spaces ([flake8 E128](https://lintlyci.github.io/Flake8Rules/rules/E128.html)).

bocekm: Also add a comment to this line as the reason may not by obvious to everybody.

Rezney:
@bocekm Thanks for the review. Will take a look at it on Friday.

Rezney:
aah, yep also tests...

Rezney:
@bocekm should be ok now

PR #95, opened 2019-03-06
SKIP: Copy yum configuration to the container

radomil:


pirat89:
This PR is not supposed to be merged... After @radomil finish his work, problem will be resolved in PR #71 .

PR #96, opened 2019-03-08
Bind mount udev database during upgrade

Rezney:
During the upgrade transaction many scriptlets call various commands which require access to udev database. When database is not accessible we are hitting all kind of issues like timeouting and the upgrade takes too much time.


Rezney:
Errg took some other branches in...fixing...

PR #97, opened 2019-03-08
Add actor to inhibit upgrade if broute used in firewalld

erig0:
The broute table is not available in RHEL-8. As such we can't do an
upgrade of firewalld's rules if they're attempting to use the broute
table.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/firewalldcheckebtablesbroute/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/firewalldcheckebtablesbroute/actor.py
@@ -1,0 +1,36 @@
+from leapp.actors import Actor
+from leapp.models import Inhibitor
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+from leapp.libraries.actor import private
+
+import os
+import xml.etree.ElementTree as ElementTree
+
+
+class FirewalldCheckEbtablesBroute(Actor):
+    """
+    In RHEL-8 the ebtables table broute is not available. If the user is using
+    firewalld direct rules that utilize this table then we need to inhibit the
+    upgrade.
+    """
+
+    name = 'firewalld_check_ebtables_broute'
+    consumes = ()
+    produces = (Inhibitor,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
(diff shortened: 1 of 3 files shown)


repos/system_upgrade/el7toel8/actors/firewalldcheckebtablesbroute/actor.py
```diff
@@ -0,0 +1,36 @@
...
+    produces = (Inhibitor,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        path = '/etc/firewalld/direct.xml'
+        if not os.path.exists(path):
+            return
+
+        tree = ElementTree.parse(path)
+        root = tree.getroot()
```
pirat89: you cannot interact with system in `ChecksPhase`. See the [IPU workflow](https://github.com/oamg/leapp-repository/blob/a72e1c0d311dfbcb0121db79305b9a0e48474dc4/repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py#L18) file with description of all phases. When you want to read a file from the system, you need to create actor in scanner phase and produce message using a model. So the file will be read only once and other actors will use the data.

And to inhibit the upgrade, you need an actor running in `ChecksPhase` which will consume those data.

pirat89:
@erig0 just regarding your additional PRs related to Firewalld, what about to create one scanner actor that will produce all relevant information you need about firewalld? I am ok to keep all PRs separate now and merge all firewalld scanner actors into the one after all PRs will be merged.

erig0:
@pirat89, I'll move to a single scanner to gather all firewalld information. I'll close all my current PRs. I'll combine them and open a new PR after I make all the changes.

erig0:
> what about to create one scanner actor that will produce all relevant information you need about firewalld?

@pirat89, if I consume the facts in the `ChecksPhaseTag` I won't be able to also consume it in the `ApplicationsPhaseTag`. I need to alter on disk configuration in the applications phase. So I need to produce two different models. Is this accurate?

vinzenz:
@erig0 No, consume won't remove messages. Messages are never removed. Also it is possible that your actor is started with no messages available, this needs to be handled by the actor too.

PR #98, opened 2019-03-08
Add actors for firewalld ipset types

erig0:
firewalld's nftables backend does not yet support all "ipset" types.
Support is missing in nftables (sets with concatenations and intervals).
The nftables backend is the preferred backend so we must catch
configurations that use the unsupported ipset types.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/firewalldcheckipsettypes/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/firewalldcheckipsettypes/actor.py
@@ -1,0 +1,46 @@
+from leapp.actors import Actor
+from leapp.models import Inhibitor
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+from leapp.libraries.actor import private
+
+import os
+import xml.etree.ElementTree as ElementTree
+
+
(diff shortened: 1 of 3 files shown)


repos/system_upgrade/el7toel8/actors/firewalldcheckipsettypes/actor.py
```diff
@@ -0,0 +1,46 @@
...
+                        solutions='Remove ipsets of type {} from firewalld.'.format(ipset.attrib['type'])
+                        ))
+
+    def process(self):
+        directory = '/etc/firewalld/ipsets'
+        for file in os.listdir(directory):
+            if not file.endswith('.xml'):
+                continue
+
+            self.process_ipset(file[:-4], os.path.join(directory, file))
```
pirat89: Same as for PR #97: you cannot interact with system in `ChecksPhase`. Think about that like: Actors in this phase would be processed on completely different machine and you need to work with data produced by something (scanner actors) that is running on the real machine.

erig0:
Closed as per https://github.com/oamg/leapp-repository/pull/97#issuecomment-473375140.

PR #100, opened 2019-03-12
SKIP RHSM (devel)

pirat89:
For the testing purposes we need to skip use of subscription-manager
and go on just with custom repositories. This is now possible with
the patch when proper custom repositories are set. The solution will
be reworked in future.

To skip the subscription-manager, set
```
LEAPP_DEVEL_SKIP_RHSM=1
```

Keep in mind that when the option is used, user (devel/tester) is reponsible that proper repositories are delivered. There are skipped several checks regarding the installation repositories (see `CustomTargetRepository` model; the repo file has to exist already on the system)

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
@@ -36,6 +36,18 @@ class PrepareUpgradeTransaction(Actor):
                 return True
         return False
 
+    def get_rhsm_system_release(self):
+        """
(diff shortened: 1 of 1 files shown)


jmikovic:
I'm missing some documentation explaining how to make use of the skip, how to setup custom repos for upgrade.

jmikovic:
please rebase so I can test (ATM leapp is being removed during upgrade)

pirat89:
> I'm missing some documentation explaining how to make use of the skip, how to setup custom repos for upgrade.

Commit message :) I will update description of PR

pirat89:
rebased

PR #102, opened 2019-03-14
Support upgrade to authselect

pbrezina:
Authconfig was replaced with authselect in RHEL8. The conversion logic
that is implemented in those actors goes like this:

1. Detect possible authselect profile by looking up modules in PAM
   or by checking that daemon is enabled.
   - pam_sss -> sssd
   - pam_winbind -> winbind
   - ypbind enabled -> nis

   If more then one module/daemon is detected that we will keep the
   configuration intact. No authselect profile can be applied.

2. Detect authselect profile features by looking up modules in PAM
   or nsswitch.conf.
   - pam_faillock => with-faillock
   - pam_fprintd => with-fingerprint
   - pam_access => with-pamaccess
   - pam_mkhomedir => with-mkhomedir
   - pam_oddjob_mkhomedir => with-mkhomedir

3. Detect if pam_krb5 or pam_pkcs11 is used.
   These are no longer present in RHEL-8 and must be removed from PAM
   configuration, otherwise it may lock out the system.

4. Check if there are any unknown PAM modules.
   If there are used PAM modules not used in authselect (such as pam_ldap),
   we must keep the configuration intact.

5. Check if authconfig was used to create current configuration.
   If yes, we can automatically convert the configuration to authselect.
   If no, we need admin's confirmation.

   - Check that /etc/sysconfig/authconfig exists.
   - Check that PAM configuration uses authconfig files.
   - Check that PAM configuration was not touch after sysconfig file
     was created.

I am not a python developer. Feel free to suggest better code.


pbrezina:
Thank you for your review. All things should be addressed.

pbrezina:
I'm not sure how leapp is run on RHEL-8 machine, but I suppose that root login must work for it. Therefore it would be better to call `AuthselectApply::remove_old_pam_modules` on RHEL-7 machine before the upgrade. How do I achieve that?

vinzenz:
> I'm not sure how leapp is run on RHEL-8 machine, but I suppose that root login must work for it. Therefore it would be better to call `AuthselectApply::remove_old_pam_modules` on RHEL-7 machine before the upgrade. How do I achieve that?

Well if it really has to be, the 'PreparationPhase' is the place to do it then. This is the place where we do apply workarounds to ensure the success of the transaction. So you'd have to have a separate actor for this, in this case.

pbrezina:
> > I'm not sure how leapp is run on RHEL-8 machine, but I suppose that root login must work for it. Therefore it would be better to call `AuthselectApply::remove_old_pam_modules` on RHEL-7 machine before the upgrade. How do I achieve that?
> 
> Well if it really has to be, the 'PreparationPhase' is the place to do it then. This is the place where we do apply workarounds to ensure the success of the transaction. So you'd have to have a separate actor for this, in this case.
> 
> Edit: However if you can get around by just scanning for the modules to remove before the transaction, and actually doing the moving in the ApplicationPhase (after the RPM upgrade has been applied) then I could sleep better ;-)

It depends on the execution. The problem is that `pam_pkcs11` and `pam_krb5` module are not present in RHEL8 anymore.  If they are left in the PAM configuration and are required or requisite (which is common) it will block the authentication. So if we can do RPM upgrade and fix PAM without rebooting we can do it in this order. However once RPM upgrade removes these modules from the system, you can no longer login until these module are removed from PAM. So it pretty much depends on leapp internals, so you tell me what is the best place to do it?

pbrezina:
I split this functionality into separate model and actors (RemovedPAMModules*), the phase tag can be changed easily.

pbrezina:
I updated the patches. I renamed RemovedOldPAMModules to RemoveOldPAMModules and now it will ask the admin if the modules should be disabled automatically. If not the upgrade process should be interrupted.

bocekm:
@pbrezina, we are changing license of this repository from AGPL 3.0 to Apache 2.0 under PR https://github.com/oamg/leapp-repository/pull/112. Is it ok with you?

pbrezina:
@vinzenz  Hi, do you have any other requirements?

pirat89:
@pbrezina Hi, @vinzenz is on PTO now so we will take a look at it soon, someone.

repos/system_upgrade/el7toel8/actors/authselectapply/actor.py
```diff
@@ -0,0 +1,65 @@
...
+            model.profile,
+            ' '.join(model.features)
+        )
+
+        try:
+            call(self.command.split(' '))
+        except Exception as err:
+            self.failure(err)
+            return
+        except:
```
fernflower: Sorry, not getting it - what purpose has the second except block at L39?

repos/system_upgrade/el7toel8/actors/authselectapply/actor.py
```diff
@@ -0,0 +1,65 @@
...
+    name = 'authselect_apply'
+    consumes = (Authselect, AuthselectDecision,)
+    produces = (Report,)
+    tags = (IPUWorkflowTag, ApplicationsPhaseTag)
+
+    def process(self):
+        model = next(self.consume(Authselect))
+        decision = next(self.consume(AuthselectDecision))
+
+        if not decision.confirmed:
```
fernflower: Please merge L23-L27 under one condition

repos/system_upgrade/el7toel8/actors/authselectcheck/actor.py
```diff
@@ -0,0 +1,163 @@
...
+            )
+        )
+
+        self.produce(
+            Report(
+                severity='Info',
+                result='Not Applicable',
+                summary=self.summary_keep,
+                details=self.details_none,
+                solutions=None
```
fernflower: Aren't solutions None by default?

repos/system_upgrade/el7toel8/actors/authselectscanner/actor.py
```diff
@@ -0,0 +1,222 @@
...
+            if not os.path.islink(name):
+                return
+
+            if os.readlink(name) != target:
+                return
+
+        # Check that all file were not modified after
+        # /etc/sysconfig/authconfig was created.
+        mtime = os.path.getmtime(sysconfig)
+        for file in links.values():
```
fernflower: overridden python file built-in at L218, rename?

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
```diff
@@ -0,0 +1,105 @@
...
+        call(
+            ["/usr/bin/systemctl", "is-enabled", service + ".service"],
+        )
+    except (OSError, CalledProcessError):
+        return False
+
+    return True
+
+
+class Authconfig(object):
```
fernflower: Correct me if I'm wrong, please
It look like classes Authconfig and DConf do the same config parsing (if the option is enabled, the only difference is the actual value - yes vs true - of the truth value). Can the code be generalized and reused then?

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
```diff
@@ -0,0 +1,105 @@
+import errno
+import re
+
+from leapp.libraries.stdlib import call
+from subprocess import CalledProcessError
+
+
+def read_file(file):
```
fernflower: If the purpose is to read a single file and return '' if there is None - maybe remove the try/catch and check file presence with other methods?

repos/system_upgrade/el7toel8/actors/authselectapply/actor.py
pbrezina: None. I removed it.

repos/system_upgrade/el7toel8/actors/authselectcheck/actor.py
pbrezina: Yes. I removed the line.

repos/system_upgrade/el7toel8/actors/authselectscanner/actor.py
pbrezina: Renamed to `f`.

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
pbrezina: Yes, this is the purpose. What would be a better way to do it?

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
pbrezina: Done.

pbrezina:
No there are  no unit tests. How would you like it to be tested?
I addressed other comments. But now I'm getting:
```
  File "/shared/code/repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py", line 27, in is_service_enabled
    call(["/usr/bin/systemctl", "is-enabled", service + ".service"])
TypeError: 'module' object is not callable
```

I'm not sure why, it does not seem to be related to the changes I made. As I wrote earlier, I'm not python developer. Any help is appreciated.

vinzenz:
@pbrezina you need to rebase and use run instead. The call function was removed

pbrezina:
Rebased. Thanks.

vojtechsokol:
leapp-ci build

pbrezina:
I pushed a commit called `wip`. The commit switch from producing `Report` to the new `report_generic` and `report_with_remediation`. It also moves stuff out from `AuthselectScanner` actor to standalone  library so it can be unit tested. However, I get:

```
[root@leapp-20190415120054 el7toel8]# snactor --debug --verbose run --print-output AuthselectScanner
Process Process-83:
    self.run()
  File "/usr/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 59, in _do_run
    target_actor(logger=logger, messaging=messaging).run(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/leapp/actors/__init__.py", line 299, in run
    self.process(*args)
  File "/shared/code/repos/system_upgrade/el7toel8/actors/authselectscanner/actor.py", line 83, in process
    ac = Authconfig(read_file('/etc/sysconfig/authconfig'))
  File "/shared/code/repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py", line 67, in __init__
    super(Authconfig, self).__init__('[authconfig]\n' + config)
TypeError: super(type, obj): obj must be an instance or subtype of type
```

The code works well without leapp. Do you know this error?

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
fernflower: I'd substitute L16-L23 with smth like
```
if not os.path.exists(config) or not os.path.isfile(config):
    return ""
with open(config) as f:
    return f.read()
```
Please don't shadow file builtin, use another name like config/filepath/etc.
Also imho there is no point in raising an exception at L23.

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
```diff
@@ -0,0 +1,273 @@
...
+        else:
+            raise 
+
+
+def is_service_enabled(service):
+    """
+        Return true if @service is enabled with systemd, false otherwise.
+    """
+    try:
+        run(["/usr/bin/systemctl", "is-enabled", service + ".service"])
```
fernflower: Please use string formatting instead of concatenation `{}.service.format(service)`

repos/system_upgrade/el7toel8/actors/removeoldpammodulescheck/actor.py
```diff
@@ -0,0 +1,116 @@
...
+                self.produce_report(module)
+            else:
+                self.produce(Inhibitor(
+                    summary='Upgrade process was interrupted because there are '
+                            'some unsupported PAM modules in configuration.',
+                    details='SA user chose to interrupt the upgrade.',
+                    solutions='Remove pam_pkcs11 and pam_krb5 from PAM '
+                              'configuration and replace them with SSSD.'
+                ))
+            pass
```
fernflower: not sure pass is needed here

repos/system_upgrade/el7toel8/libraries/pam.py
```diff
@@ -0,0 +1,54 @@
...
+    def has_unknown_module(self, known_modules):
+        for module in self.modules:
+            if module not in known_modules:
+                print(module)
+                return True
+
+        return False
+    
+    @staticmethod
+    def read_file(file):
```
fernflower: Not sure you really need to handle those exceptions, see comment above about read_file

repos/system_upgrade/el7toel8/libraries/pam.py
```diff
@@ -0,0 +1,54 @@
...
+        except OSError as err:
+            if err.errno == errno.ENOENT:
+                return ""
+            else:
+                raise 
+
+    @staticmethod
+    def from_system_configuration():
+        config = ""
+        for file in PAM.files:
```
fernflower: Please don't shadow file built-in.

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
```diff
@@ -0,0 +1,273 @@
...
+        self.config = ConfigParser.ConfigParser()
+        self.config.readfp(StringIO.StringIO(textwrap.dedent(content)))
+
+    def get_string(self, section, option):
+        if not self.config.has_option(section, option):
+            return None
+
+        return self.config.get(section, option)
+
+    def get_bool(self, options):
```
vinzenz: missing section
```suggestion
    def get_bool(self, section, options):
```

repos/system_upgrade/el7toel8/actors/removeoldpammodulescheck/actor.py
vinzenz: It's actually useless

repos/system_upgrade/el7toel8/actors/authselectscanner/actor.py
```diff
@@ -0,0 +1,93 @@
+import os
+import re
+
+from leapp.actors import Actor
+from leapp.libraries.actor.library import *
```
vinzenz: Please do not use star imports

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
```diff
@@ -0,0 +1,273 @@
...
+        return super(Authconfig, self).get_bool('authconfig', option)
+
+
+class DConf(ConfigFile):
+    """
+        Parse dconf configuration.
+    """
+
+    def __init__(self, config):
```
vinzenz: This is not needed

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
```diff
@@ -0,0 +1,273 @@
...
+        super(DConf, self).__init__(config)
```
vinzenz:
```suggestion
```

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
```diff
@@ -0,0 +1,273 @@
+import ConfigParser
+import errno
+import re
```
vinzenz:
```suggestion
import os
import re
```

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
```diff
@@ -0,0 +1,273 @@
...
+        
+        self.profile = None
+        self.features = []
+        self.confirm = True
+
+    def process(self):
+        # Detect possible authselect configuration
+        self.profile = self.step_detect_profile()
+        self.features += self.step_detect_features()
+        self.features += self.step_detect_sssd_features(profile)
```
vinzenz:
```suggestion
        self.features += self.step_detect_sssd_features(self.profile)
```

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
```diff
@@ -0,0 +1,273 @@
...
+        self.features += self.step_detect_winbind_features(profile)
```
vinzenz:
```suggestion
        self.features += self.step_detect_winbind_features(self.profile)
```

vinzenz:
> I pushed a commit called `wip`. The commit switch from producing `Report` to the new `report_generic` and `report_with_remediation`. It also moves stuff out from `AuthselectScanner` actor to standalone library so it can be unit tested. However, I get:
> 
> ```
> [root@leapp-20190415120054 el7toel8]# snactor --debug --verbose run --print-output AuthselectScanner
> Process Process-83:
> Traceback (most recent call last):
>   File "/usr/lib64/python2.7/multiprocessing/process.py", line 258, in _bootstrap
>     self.run()
>   File "/usr/lib64/python2.7/multiprocessing/process.py", line 114, in run
>     self._target(*self._args, **self._kwargs)
>   File "/usr/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 59, in _do_run
>     target_actor(logger=logger, messaging=messaging).run(*args, **kwargs)
>   File "/usr/lib/python2.7/site-packages/leapp/actors/__init__.py", line 299, in run
>     self.process(*args)
>   File "/shared/code/repos/system_upgrade/el7toel8/actors/authselectscanner/actor.py", line 83, in process
>     ac = Authconfig(read_file('/etc/sysconfig/authconfig'))
>   File "/shared/code/repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py", line 67, in __init__
>     super(Authconfig, self).__init__('[authconfig]\n' + config)
> TypeError: super(type, obj): obj must be an instance or subtype of type
> ```
> 
> The code works well without leapp. Do you know this error?

So with all the changes applied that I gave as comments I had no problems running this on a centos:7 container + fedora:29 container using the latest version of leapp - So whatever it was, it might have been a fluke.

pbrezina:
I fixed all linter̈́s errors and added unit tests to PAM library and RemoveOldPAMModulesScanner/Apply and AuthselectScanner actors. Other actors should be in my opinon tested through downstream integration tests, let me know if you require unit tests for it and if so how it should look like.

There are two issues:
* `test_AuthselectScannerLibrary_process_*` fails in travis-ci, I do not know why. It works locally.
* I do not know how to run tests for PAM library.

Thank you.

pbrezina:
Travis CI pass now. Thanks @vinzenz.

repos/system_upgrade/el7toel8/actors/authselectapply/actor.py
```diff
@@ -0,0 +1,53 @@
...
+    tags = (IPUWorkflowTag, ApplicationsPhaseTag)
+
+    def process(self):
+        model = next(self.consume(Authselect))
+        decision = next(self.consume(AuthselectDecision))
+
+        if not decision.confirmed or model.profile is None:
+            return
+
+        self.command = 'authselect select {0} {1} --force'.format(
```
vinzenz:
```suggestion
        command = ['authselect', 'select', '--force', model.profile] + model.features]
```

repos/system_upgrade/el7toel8/actors/authselectapply/actor.py
```diff
@@ -0,0 +1,53 @@
...
+            model.profile,
+            ' '.join(model.features)
+        )
+
+        try:
+            run(self.command.split(' '))
```
vinzenz:
```suggestion
            run(command)
```

repos/system_upgrade/el7toel8/actors/authselectapply/actor.py
```diff
@@ -0,0 +1,53 @@
...
+        except CalledProcessError as err:
+            self.failure(err)
+            return
+
+        self.success()
```
vinzenz:
```suggestion
        self.success(command)
```

repos/system_upgrade/el7toel8/actors/authselectapply/actor.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def failure(self, err):
+        report_generic(
+            title='Authselect call failed.',
+            summary=str(err)
+        )
+
+    def success(self):
```
vinzenz:
```suggestion
    def success(self, command):
```

repos/system_upgrade/el7toel8/actors/authselectapply/actor.py
```diff
@@ -0,0 +1,53 @@
...
+        report_generic(
+            title='System was converted to authselect.',
+            summary='System was converted to authselect with the '
+                    'following call: %s' % self.command
```
vinzenz:
```suggestion
                    'following call: {}'.format(command)
```

vinzenz: Please prefer using .format formatting over % formatting

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def __init__(self, content):
+        self.config = ConfigParser.ConfigParser()
+        self.config.readfp(StringIO.StringIO(textwrap.dedent(content)))
+
+    def get_string(self, section, option):
+        if not self.config.has_option(section, option):
+            return None
+
+        str = self.config.get(section, option)
```
vinzenz: str is a builtin for the string type, this is a dangerous name - considering you don't do anything else than stripping just do this:
```suggestion
        return self.config.get(section, option).strip('"\'')
```

repos/system_upgrade/el7toel8/actors/removeoldpammodulesapply/actor.py
```diff
@@ -0,0 +1,34 @@
...
+    def process(self):
+        model = next(self.consume(RemovedPAMModules))
+        files = {}
+
+        # Read files
+        for path in PAM.files:
+            files[path] = read_file(path)
+
+        # Comment modules
+        for file, content in files.items():
```
vinzenz: just call it pam_file
```suggestion
        for pam_file, content in files.items():
```

repos/system_upgrade/el7toel8/actors/removeoldpammodulesapply/actor.py
```diff
@@ -0,0 +1,34 @@
...
+            files[file] = comment_modules(model.modules, content)
```
vinzenz:
```suggestion
            files[pam_file] = comment_modules(model.modules, content)
```

repos/system_upgrade/el7toel8/actors/removeoldpammodulesapply/actor.py
vinzenz: just call it pam_file
```suggestion
        for pam_file, content in files.items():
```

repos/system_upgrade/el7toel8/actors/removeoldpammodulesapply/actor.py
```diff
@@ -0,0 +1,34 @@
...
+    name = 'removed_pam_modules_apply'
+    consumes = (RemovedPAMModules,)
+    produces = ()
+    tags = (IPUWorkflowTag, PreparationPhaseTag)
+
+    def process(self):
+        model = next(self.consume(RemovedPAMModules))
+        files = {}
+
+        # Read files
```
vinzenz: Is there actually any particular need for 3 loops over the same content?

The whole `process` method looks to me it could be reduced to this:
```python
for model in self.consume(RemovedPAMModules):
    for path in PAM.files:
        content = read_file(path)
        with open(path, 'w') as f:
            f.write(comment_modules(model.modules, content))
    break
```

pbrezina:
Thank you. Updated patches were pushed.

vinzenz:
@shaded-enmity In this PR are used user questions - Do you have a better suggestion for this as we try to avoid them for now?

pbrezina:
The reason why it uses dialogs is that we want to convert as many systems to authselect as possible but PAM configuration can be quite complex and we can not expect every use case. The conversion is done automatically if authconfig configuration is detected. But we require confirmation otherwise to ensure that we do not change behaviour of current PAM stack.

shaded-enmity:
@vinzenz @pbrezina  Let's keep the user questions in place for now. One thing I'd like to ask, though, is to break down the actual code path for the user question into a callable script, so that it can be used for remediations as well.

Jakuje:
We will need to write a leapp check `pam_tally2`, which got removed in RHEL8 and can break authentications after update. For this, I would need more elaborate PAM library, that could check all the files under `/etc/pam.d/` and report what all modules are being used.

I would like to avoid writing my own PAM library and different set of models for quite much the same as is done here so I would like to discuss whether it would be possible to extend this library and the leapp models to support this (either before merging if there is still something blocking the merge or after if it should unnecessarily delay the merge).

I can imagine there would be a model `PamConfiguration`, which would contain a list of `PamService` models with information what files under `/etc/pam.d/` does it come from and what are the pam modules used in that service. The consumers could chose what files they are interested in (here only the limited subset managed by authselect) and handle them, giving a bit more freedom than the current implementation of `RemovedPamModules` model.

Please, let me know whether something like this would make sense for you, whether is it something I should start looking into.

pbrezina:
@shaded-enmity I'm not sure I follow. Can you explain it more?

@Jakuje I would like to avoid further functional changes within this PR so it gets merged. But after that feel free to extend/change it. It makes sense. I chose only this subset of files because it is the supported way to configure authentication for common services, but it would not hurt to check other files as well.

Jakuje:
@pbrezina Thank you. I will have a look into that next week, separately from this PR now.

shaded-enmity:
@pbrezina Apologies for the noise, I just noticed that the autheselect is already applied as standalone command
```
command = ['authselect', 'select', '--force', model.profile] + model.features
```
The idea here is to be able to execute these remediations at scale at some point, if we can boil down these fixes to simple & self-contained scripts, we can collect and execute these using Ansible at as many machines as we like.

pbrezina:
@shaded-enmity  Ok, so do you request any changes from me or can this PR be accepted? Thank you.

repos/system_upgrade/el7toel8/actors/authselectapply/actor.py
```diff
@@ -0,0 +1,50 @@
...
+
+        if not decision.confirmed or model.profile is None:
+            return
+
+        command = ['authselect', 'select', '--force', model.profile] + model.features
+
+        try:
+            run(command)
+        except CalledProcessError as err:
+            self.failure(err)
```
fernflower: nit: as you are not reusing code in self.failure and self.success I'd rather not have 2 separate methods but call report_generic functions directly

repos/system_upgrade/el7toel8/actors/authselectcheck/actor.py
```diff
@@ -0,0 +1,140 @@
...
+            return
+
+        # Authselect profile is available but we require confirmation.
+        confirmed = self.get_confirmation(model, command)
+        self.produce_suggested_configuration(model, confirmed, command)
+
+    def get_confirmation(self, model, command):
+        dialog = self.dialogs[0]
+
+        dialog.components[0].label = '{0}\n{1}\n'.format(
```
fernflower: nit: imho it's cleaner, but up to you
```
dialog.components[0].label += "\n{}\n".format(command)
```

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
```diff
@@ -0,0 +1,267 @@
...
+import os
+import re
+import StringIO
+import textwrap
+
+from leapp.libraries.stdlib import run, CalledProcessError
+from leapp.models import Authselect
+
+
+def read_file(config):
```
fernflower: I see identical function in libraries/pam.py
Can this code be shared?

repos/system_upgrade/el7toel8/actors/authselectscanner/tests/unit_test.py
```diff
@@ -0,0 +1,624 @@
+import textwrap
```
fernflower: Wrong import order. Wondering why our tests didn't catch it

fernflower:
Update on live upgrade: the vm didn't survive reboot, looks like it is PR-related.
`Inspection of actor in actors/authselectscanner failed` and `"ModuleNotFoundError": No module named ConfigParser`

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
```diff
@@ -0,0 +1,267 @@
+import ConfigParser
```
vinzenz:
```suggestion
from six.moves import configparser
```

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
```diff
@@ -0,0 +1,267 @@
...
+    return True
+
+
+class ConfigFile(object):
+    """
+        Base class for config parsers.
+    """
+
+    def __init__(self, content):
+        self.config = ConfigParser.ConfigParser()
```
vinzenz:
```suggestion
        self.config = configparser.ConfigParser()
```

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
vinzenz: I'd rather leave it as is TBH and leave room for improvement in other ways later.

repos/system_upgrade/el7toel8/actors/authselectapply/actor.py
vinzenz: But I won't let this prevent a merge now

repos/system_upgrade/el7toel8/actors/authselectscanner/tests/unit_test.py
vinzenz: Missing rebase?

vinzenz:
Please fix the issues with configparser which is necessary, the code needs to be python 2 and python 3 compatible.

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
pbrezina: Leaving this as is.

pbrezina:
Fixed.

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
```diff
@@ -0,0 +1,268 @@
+import os
+import re
+import StringIO
```
vinzenz:
```suggestion
from six import StringIO
```

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
```diff
@@ -0,0 +1,268 @@
...
+
+
+class ConfigFile(object):
+    """
+        Base class for config parsers.
+    """
+
+    def __init__(self, content):
+        self.config = configparser.ConfigParser()
+        self.config.readfp(StringIO.StringIO(textwrap.dedent(content)))
```
vinzenz:
```suggestion
        self.config.readfp(StringIO(textwrap.dedent(content)))
```

repos/system_upgrade/el7toel8/actors/authselectscanner/libraries/library.py
```diff
@@ -0,0 +1,268 @@
+import os
+import re
+import StringIO
+import textwrap
+
+from six.moves import configparser
```
vinzenz:
```suggestion
from six import StringIO
from six.moves import configparser
```

pbrezina:
Thank you.  I squashed your changes in.

repos/system_upgrade/el7toel8/actors/removeoldpammodulescheck/actor.py
```diff
@@ -0,0 +1,101 @@
...
+    produces = (Report,)
+    tags = (IPUWorkflowTag, ChecksPhaseTag)
+    dialogs = (
+        Dialog(
+            scope='remove_old_pam_modules_check',
+            reason='Confirmation',
+            components=(
+                BooleanComponent(
+                    key='confirm',
+                    label='Disable pam_pkcs11 module in PAM configuration? '
```
fernflower: Probably you should put information about 'yes' option here, as currently the message shows option for 'no' only. Might be confusing as the operator doesn't know what to put ('y' or 'yes')

pirat89: currently, it is not possible to use dialogs as interactive upgrades are no go now.

pirat89: let's discuss that next week, maybe I am missing something. But would like to stay clear: When upgrade is not possible to do without interactivity, it's no-go for many users. When there will be dialogs, there has to be way to skip any interactions and use defaults for example (and that should be default)

pbrezina: I was under the impression that it some answer file and unattended upgrade is already supported by the leapp tool. If it is not the case and it won't be, I agree that running the upgrade in non-interactive way has precedence. However, in this case customers must have the option to opt out from the authselect actors and it should be thoroughly documented.

pbrezina: > Probably you should put information about 'yes' option here, as currently the message shows option for 'no' only. Might be confusing as the operator doesn't know what to put ('y' or 'yes')

This should be handled by the leapp framework. I'm just using BooleanComponent with default=False which is represented by leapp as `Question? [no]` instead of 'Question? [yes/no] [default: no]' or something similar.

fernflower:
Live upgrade passed. Dialogs are to be discussed but general concern - upgrade has to be non-interactive by default, especially as RemoveOldPAMModules actor is hit in 100% cases on defautl rhel7 image

pbrezina:
Done. Thank you.

fernflower:
On the other hand - it's probably not important to have experimental-related info in commit message. Merging.

PR #103, opened 2019-03-15
Add jinja2 pkg requirement

pirat89:
Fix dependencies regarding the PR oamg/leapp#445


pirat89:
@vinzenz Vinnie, look at it please and merge it. So we will not have to w8 for fix of #86

PR #104, opened 2019-03-17
Get rid of the leapp-repository-data subpackage and data file

pirat89:
We do not plan to use this subpackage anymore, neither provide the data file in rpm.

Note: **Do not hurry with merging** in case you approve the changes. It would be better to coordinate work with @artmello so we will have prepared infrastructure.

Note: Added Obsoletes & Provides is temporary to ensure the subpackage is removed from the system after the upgrade. It would happen that someone would use pretty old data from the past and I prefer to get "issue" about missing data than spend a time on traceback because of the old data format (e.g.).

~Note: Currently the rpm will contain the "empty" pes-event.json file. It is expected as the removal of the file from the repository will be part of following PR.~ (relevant PR has been merged already)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,4 @@
 __PKGNAME=$${_PKGNAME:-leapp-repository}
-__DATA_ORIG_PKGNAME=$${_DATA_ORIG_PKGNAME:-leapp-pes-data}
 PKGNAME=leapp-repository
 DEPS_PKGNAME=leapp-el7toel8-deps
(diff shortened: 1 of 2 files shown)


jmikovic:
please rebase and fix conflicts

pirat89:
@jmikovic done

pirat89:
@jmikovic done

jmikovic:
leapp-ci build

PR #105, opened 2019-03-18
Introduction of initial sos plugin for leapp

vinzenz:
In order to allow an easier collection of data needed for debugging,
we will add a plugin for the sos tool to include our data to be
collected. That would include the upgrade.log, the leapp.db and
potential debugdata from dnf generated during the upgrade.

### DIFF
--- a/sources/sos-report/leapp.py
+++ b/sources/sos-report/leapp.py
@@ -1,0 +1,19 @@
+from sos.plugins import Plugin, RedHatPlugin
+
+
+class Leapp(Plugin, RedHatPlugin):
+    """
+    This plugin is used to gather all information necessary for reporting
+    problems in regards to the leapp application.
+    """
+
+    plugin_name = 'leapp'
+    packages = ('leapp', 'leapp-repository')
+
+    def setup(self):
+        self.add_copy_spec([
+            '/var/lib/leapp/leapp.db',
+            '/tmp/download-debugdata',
+            '/var/log/upgrade.log',
+            '/tmp/leapp-report.txt'
+        ])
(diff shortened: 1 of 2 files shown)


jmikovic:
`plugin leapp does not install, skipping: 'str' object is not callable` when running sosreport

sources/sos-report/leapp.py
```diff
@@ -0,0 +1,18 @@
+from sos.plugins import Plugin, RedHatPlugin
+
+
+class Leapp(Plugin, RedHatPlugin):
+    """
+    This plugin is used to gather all information necessary for reporting
+    problems in regards to the leapp application.
+    """
+
+    name = 'leapp'
```
jmikovic:
```suggestion
    plugin_name = 'leapp'
```

jmikovic:
Shouldn't we collect /tmp/leapp-report.txt as well?
Plugin is not installed after upgrade, because sosreport uses py3, need to copy the plugin to python3 sosreport during upgrade

sources/sos-report/leapp.py
vinzenz: https://github.com/sosreport/sos/wiki/How-to-Write-a-Plugin they don't say 'plugin_name' for what ever reason.

vinzenz:
> Shouldn't we collect /tmp/leapp-report.txt as well?
> Plugin is not installed after upgrade because sosreport uses py3 in rhel8, need to copy the plugin to python3 sosreport during upgrade

Yeah, i know that it isn't there in Python3, probably using an actor might be probably the easiest solution, however I am not sure if that's a good idea though :S

sources/sos-report/leapp.py
jmikovic: last edit is 2015, so I guess they changed it and didn't update docs, existing plugins use `plugin_name`

vinzenz:
leapp-ci build

vinzenz:
leapp-ci build

PR #107, opened 2019-03-21
Inhibit upgrade if PES data or repomap not found.

asilveir:


repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -135,13 +139,26 @@ def process_events(events):
         api.produce(RepositoriesSetupTasks(to_enable=list(to_enable_repos)))
 
 
+def verify_pes_data(path):
+    """Verify if PES data is available if not inhibit upgrade"""
+    if path is None or not os.path.isfile(path):
+        title = 'PES data is invalid (/etc/leapp/files/pes-data.json)'
```
jmikovic:
```suggestion
        title = 'Missing PES data ({})'.format(path)
```

Invalid doesn't seem right if you only check existence of the file

repos/system_upgrade/el7toel8/actors/repositoriesmapping/actor.py
```diff
@@ -15,17 +18,25 @@ class RepositoriesMapping(Actor):
...
     tags = (IPUWorkflowTag, ChecksPhaseTag)
 
-    
     def process(self):
+        file_path = '/etc/leapp/files/repomap.csv'
         repos_maps = RepositoriesMap()
 
-        with open(self.get_file_path('repomap.csv')) as f:
+        if not os.path.isfile(file_path):
+            title = 'Repositories data file is invalid (/etc/leapp/files/repomap.csv)'
```
jmikovic: same as above

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -135,13 +139,26 @@ def process_events(events):
         api.produce(RepositoriesSetupTasks(to_enable=list(to_enable_repos)))
 
 
+def verify_pes_data(path):
```
jmikovic: am I missing something or is this function not used anywhere,
I guess you wanted to use it in parse_file?

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -135,13 +139,26 @@ def process_events(events):
...
-        raise StopActorExecutionError(
-            message='Found invalid or outdated PES data.',
-            details={'details': error, 'hint': 'Consult docs to retrieve updated data.'})
+    except (ValueError, KeyError) as value_error:
+        title = 'PES data is invalid (/etc/leapp/files/pes-data.json)'
+        summary = 'Read documentation at: https://access.redhat.com/articles/3664871 for more information ' \
+            'about how to retrieve the files'
+        error = str(value_error)
+        reporting.report_generic(title=title, summary=summary, severity='high', flags=['inhibitor'])
+        raise StopActorExecutionError(message=title, details={'hint': summary, 'details': error})
```
jmikovic: duplicate of duplicate_pes_data func

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -135,13 +139,26 @@ def process_events(events):
...
 
 
+def verify_pes_data(path):
+    """Verify if PES data is available if not inhibit upgrade"""
+    if path is None or not os.path.isfile(path):
+        title = 'PES data is invalid (/etc/leapp/files/pes-data.json)'
+        summary = 'Read documentation at: https://access.redhat.com/articles/3664871 for more information ' \
+            'about how to retrieve the files'
+        reporting.report_generic(title=title, summary=summary, severity='high', flags=['inhibitor'])
+        raise StopActorExecutionError(message=title, details={'details': summary})
```
artmello: Up to you, but I would raise this StopActorExecutionError either here or after the "try/except" at "scan_events". Not in both. Maybe you can call "verify_pes_data" before "parse_file" inside "scan_events" and if a problem is found raise an exception inside "verify_pes_data" that will be catch inside "scan_events".

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -135,13 +139,26 @@ def process_events(events):
...
 def scan_events(path):
     """ Scan JSON file containing PES events """
     try:
         events = parse_file(path)
-    except (ValueError, KeyError) as error:
-        raise StopActorExecutionError(
-            message='Found invalid or outdated PES data.',
-            details={'details': error, 'hint': 'Consult docs to retrieve updated data.'})
+    except (ValueError, KeyError) as value_error:
+        title = 'PES data is invalid (/etc/leapp/files/pes-data.json)'
```
artmello: Use "path" variable instead of "/etc/leapp/...", that way if the path changes in future we can fix it in a single place

repos/system_upgrade/el7toel8/actors/repositoriesmapping/actor.py
```diff
@@ -15,17 +18,25 @@ class RepositoriesMapping(Actor):
...
+            summary = 'Read documentation at: https://access.redhat.com/articles/3664871 for more information ' \
+                'about how to retrieve the files'
+            reporting.report_generic(title=title, summary=summary, severity='high', flags=['inhibitor'])
+            raise StopActorExecutionError(message=title, details={'details': summary})
+
+        with open(file_path) as f:
+            print(f.readlines())
             data = csv.reader(f)
```
artmello: Would be good to have some check for repomap.csv file content. Maybe check if csv.reader was able to read the file and notify user if some problem happens.

repos/system_upgrade/el7toel8/actors/peseventsscanner/actor.py
```diff
@@ -14,9 +15,8 @@ class PesEventsScanner(Actor):
...
     name = 'pes_events_scanner'
     consumes = (InstalledRedHatSignedRPM,)
-    produces = (RpmTransactionTasks, RepositoriesSetupTasks,)
+    produces = (RpmTransactionTasks, RepositoriesSetupTasks, Report,)
     tags = (IPUWorkflowTag, FactsPhaseTag)
 
-
     def process(self):
-        scan_events(self.get_file_path('pes-events.json'))
+        scan_events('/etc/leapp/files/pes-data.json')
```
artmello:
```suggestion
        scan_events('/etc/leapp/files/pes-events.json')
```

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -139,9 +143,12 @@ def scan_events(path):
     """ Scan JSON file containing PES events """
     try:
         events = parse_file(path)
-    except (ValueError, KeyError) as error:
-        raise StopActorExecutionError(
-            message='Found invalid or outdated PES data.',
-            details={'details': error, 'hint': 'Consult docs to retrieve updated data.'})
+    except (ValueError, KeyError) as value_error:
+        title = 'Missing PES data ({})'.format(path)
```
artmello: I think "Missing" is not the right word here. You can reach this with an invalid data as much as for missing. We will explain better for the user what the error is in details. With that I would go with something like "Missing/Invalid PES data file", or anything else you think it is better.

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -139,9 +143,12 @@ def scan_events(path):
     """ Scan JSON file containing PES events """
     try:
         events = parse_file(path)
-    except (ValueError, KeyError) as error:
-        raise StopActorExecutionError(
-            message='Found invalid or outdated PES data.',
-            details={'details': error, 'hint': 'Consult docs to retrieve updated data.'})
+    except (ValueError, KeyError) as value_error:
```
artmello: Sorry for being pick, but I think "value_error" is a wrong name for this value since it can be a key error also. What about "error" or "err"

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -139,9 +143,12 @@ def scan_events(path):
...
+        title = 'Missing PES data ({})'.format(path)
+        summary = 'Read documentation at: https://access.redhat.com/articles/3664871 for more information ' \
+            'about how to retrieve the files'
+        error = str(value_error)
```
artmello: Since we are using this only once, I would vote for dropping it and using "str(value_error)" directly in place

repos/system_upgrade/el7toel8/actors/peseventsscanner/tests/unit_test.py
```diff
@@ -210,3 +221,21 @@ def consume_message_mocked(*models):
 
     with pytest.raises(StopActorExecutionError):
         scan_events('files/tests/sample02.json')
+    assert reporting.report_generic.called == 1
+    assert 'inhibitor' in reporting.report_generic.report_fields['flags']
+
+    with pytest.raises(StopActorExecutionError):
+        scan_events('files/tests/sample03.json')
+    assert reporting.report_generic.called == 2
```
artmello: What do you think about cleaning "called" count, before checking for it again? Looking only for the test I become curious why first called is 1 and second is 2, since, in the end, the two calls for scan_events are not connected in fact.

repos/system_upgrade/el7toel8/actors/repositoriesmapping/actor.py
```diff
@@ -12,26 +16,41 @@ class RepositoriesMapping(Actor):
 
     name = 'repository_mapping'
     consumes = ()
-    produces = (RepositoriesMap,)
+    produces = (RepositoriesMap, Report,)
     tags = (IPUWorkflowTag, ChecksPhaseTag)
+    file_path = '/etc/leapp/files/repomap.csv'
```
artmello: I think this can be a variable inside "process", maybe being exposed for all actor is unnecessary.

repos/system_upgrade/el7toel8/actors/repositoriesmapping/actor.py
```diff
@@ -12,26 +16,41 @@ class RepositoriesMapping(Actor):
...
+                                                                 to_id=to_id,
+                                                                 from_minor_version=from_minor_version,
+                                                                 to_minor_version=to_minor_version,
+                                                                 arch=arch,
+                                                                 repo_type=repo_type))
+            if len(repos_maps.repositories) > 0:
+                self.produce(repos_maps)
+            else:
+                self.inhibit_upgrade('No repositories found in map file ({})'.format(self.file_path))
+        except StopIteration:
```
artmello: If I understood correctly, we are trying to handle StopIteration that can be raised by "csv.reader" only. If that is the case, could we move this except to be closer to "csv.reader" call? If we are only checking for exceptions raised for reader, I would advocate that we keep only reader inside the try/except.

repos/system_upgrade/el7toel8/actors/repositoriesmapping/files/tests/sample01.csv
```diff
@@ -0,0 +1,5 @@
+header 1,header 2,header 3,header 4,header 5,header 6
```
artmello: If you don't mind, could you add some sample data more descriptive? Like, real title for columns. Content should be mocked (i.e. "SampleRepo01" or something), but with more descriptive data this can be used as some sort of documentation to someone looking to the actor and trying to understand what is expected.

repos/system_upgrade/el7toel8/actors/repositoriesmapping/files/tests/sample04.csv
```diff
@@ -0,0 +1,5 @@
+header 1,header 2,header 3,header 4,header 5,header 6
```
artmello: Same point as the other sample file. Up to you :)

repos/system_upgrade/el7toel8/actors/repositoriesmapping/libraries/library.py
```diff
@@ -0,0 +1,50 @@
...
+    reporting.report_generic(title=title, summary=summary, severity='high', flags=['inhibitor'])
+    raise StopActorExecutionError(message=title, details={'details': summary})
+
+
+def scan_repositories(path):
+    if not os.path.isfile(path):
+        inhibit_upgrade('Repositories map file not found ({})'.format(path))
+
+    with open(path) as f:
+        if len(f.readlines()) == 0:
```
artmello: If the idea is to check if file is not empty you could use "os.path.getsize(path)". That would be more descriptive of your intention and no need to call an "open"

repos/system_upgrade/el7toel8/actors/repositoriesmapping/tests/unit_test.py
```diff
@@ -0,0 +1,81 @@
...
+class report_generic_mocked(object):
+    def __init__(self):
+        self.called = 0
+
+    def __call__(self, **report_fields):
+        self.called += 1
+        self.report_fields = report_fields
+
+
+def test_scan_valid_file(monkeypatch):
```
artmello: Thanks a lot for adding tests. Looks pretty good to me :)

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -140,8 +144,10 @@ def scan_events(path):
...
         events = parse_file(path)
     except (ValueError, KeyError) as error:
-        raise StopActorExecutionError(
-            message='Found invalid or outdated PES data.',
-            details={'details': error, 'hint': 'Consult docs to retrieve updated data.'})
+        title = 'Missing/Invalid PES data file ({})'.format(path)
+        summary = 'Read documentation at: https://access.redhat.com/articles/3664871 for more information ' \
+            'about how to retrieve the files'
+        reporting.report_generic(title=title, summary=summary, severity='high', flags=['inhibitor'])
+        raise StopActorExecutionError(message=title, details={'hint': summary, 'details': str(error)})
```
pirat89: This doesn't make sense. You cannot report when you tell that execution should be stopped (so the executor taking care of report will not be procesesd). Only the first one or the second one.

asilveir: @pirat89 I opened another PR to solve this https://github.com/oamg/leapp-repository/pull/109

ilatypov:
How much interweb does leapp assume one needs to search before figuring the role of the missing 2 files?  There are stale references to sample contents of these files, and without a Leapp Guru I doubt I would be able to find the answer.

shaded-enmity:
hi @ilatypov our goal is to make the process simple and easy. Based on your message there is a room for improvement - however we need to balance information density and focus. Not everyone is interested in the WHY do I need these files, but rather where do I get them so that the message goes away. 
With that being said, do you think that explaining the purpose of those files in the linked article is sufficient?

ilatypov:
I guess it is for those who have a subscription?  The `--no-rhsm` option seems redundant, then.

pirat89:
> I guess it is for those who have a subscription? The --no-rhsm option seems redundant, then.

@ilatypov No. Thesetwo files are needed always (in case of the repomap, it's valid mostly just in combination with RHSM, but currently still required as there is no benefit to require just one of those file...) - otherwise the most time you will end up with broken DNF transaction (means, you will not be able to do upgrade) or in case you are lucky, you would find the upgraded system contains different RPMs than you expected (e.g. you could realize that python related rpms have not been replaced by python3 rpms, or that you are missing some functions in some specific rpms, because it was split into own subpackages, ....). But in most cases, you dnf will report conflicts and upgrade will not be possible.

The `--no-rhsm` option is for those, who would like to use own mirrors of target repositories (e.g. on own ftp server, ...). But still it is expected that repositories will contain the up-to-date content that is available through CDN. You can try in a testing VM distrosync between RHEL 7 and RHEL 8.

ilatypov:
I was able to find the leapp tar balls but the next run showed the "inhibitor" issue of missing the local copies of the RedHat repositories.  I don't know if the ISO image from the free developer subscription can be turned into repositories.  I am limited in permission to overwrite the VM image of RedHat 7 from Azure, but I guess it's easier to apply for that permission than to try an upgrade in-place.

```
Risk Factor: high (inhibitor)
Title: Using RHSM has been skipped but no custom repositories have been delivered.
Summary: Leapp is run in the mode when the Red Hat Subscription Manager is not used (the --no-rhsm option or the LEAPP_NO_RHSM=1 environment variable has been s
et) so leapp is not able to obtain YUM/DNF repositories with the content for the target system in the standard way. The content has to be delivered by the user
manually.
Remediation: [hint] Create the repository file according to instructions in the referred document on the following path with all repositories that should be used during the upgrade: "/etc/leapp/files/leapp_upgrade_repositories.repo".
```

vinzenz:
@ilatypov If it is just about getting a RHEL8 system, I am not sure why you need to 'overwrite' the image - Can't you just boot from a RHEL8 iso and perform the installation instead of an IPU.
Even though this is off topic - Just it should make your life easier if you don't really need to make an upgrade but a fresh installation is good enough.

ilatypov:
I'll see if Azure Portal allows uploading an ISO image, I guess not.

bocekm:
We have the possibility of upgrades of RHEL on Azure on our roadmap.

Before it's a reality though, you can mount the RHEL 8 ISO, serve it through a web server, and then use the http URL as a baseurl of a repository you define. It's described in https://access.redhat.com/articles/4977891#repos-known-issues.

Then you'll use this custom repository this way: `leapp preupgrade --no-rhsm --enablerepo rhel8_iso`.

PR #109, opened 2019-03-26
Remove stoperror exception from inhibitors

asilveir:
Changes to address review points from @pirat89


jmikovic:
easier solution would be to replace StopActorExecutionError exception with StopActorExecution ;)

vinzenz:
> easier solution would be to replace StopActorExecutionError exception with StopActorExecution ;)

I agree

asilveir:
> easier solution would be to replace StopActorExecutionError exception with StopActorExecution ;)

Done! Thank you @jmikovic

PR #111, opened 2019-03-27
removed systemd-mount usage

examon:
systemd-mount is not needed. When searching for nfs/xfs, it should be enough to look into mount/fstab. Also, systemd-mount may not be installed on the system by default.

- removed systemd-mount usage from checknfs and checkxfc actors and tests
- removed systemd-mount collector from storagescanner actor
- removed SystemdMountEntry model


vinzenz:
Instead of removal I prefer to make it optional in case it's not installed - The data provided by it, might be relevant for future actors - Also we should not modify models and just remove fields, because that breaks backwards compatibility - Even if not maybe now, but we have to keep now have an eye out on this!

examon:
I'm closing this. It is no longer relevant. We've dealt with the issue in the #127

PR #115, opened 2019-03-29
Handle the yum configuration files

### LINKED ISSUE #19: Directories under /etc/yum/* need to make way for ../dnf symlinks after upgrade.
[original report](https://github.com/oamg/leapp/issues/399)  by @cwyamz
> preforming an upgrade from 7.6 to 8 using leapp the user is required to dnf install yum due to yum being in the to_remove file but the transaction fails:
Error unpacking rpm package yum-4.0.9-1.el8.noarch
"error: unpacking of archive failed on file /etc/yum/pluginconf.d: cpio: File from package already exists as a directory in system"
>
> The work around is to rename /etc/yum to /etc/yum.bak and retry.
More info here: https://bugzilla.redhat.com/show_bug.cgi?id=1626736#c9
>
> libdnf-0.22.3-1.el8.x86_64
> dnf-4.0.9-1.el8.noarch
> leapp-0.4.0-1.el7_5.noarch

EDIT: See #29 for additional info.

pirat89:
RPM cannot handle replace of directories by symlinks by default
without the %pretrans scriptlet. As yum package is packaged wrong,
we have to workround that by actors so we will be able to keep
installed yum package.

Current solution handle the yum configuration files to replace
relevant directories by symlinks as expected on target system. This
change is done on two places:
- inside the container when RPM transactions is calculated and rpms
  are downloaded (the preupgradeupgradetransaction actor)
  - system is unaffected by this change
- before the upgrade itself (PreparationPhase) in the prepareyumconfig
  actor.

The important code handling that migration is stored in the tools
directory of the el7toel8 repository. See the handleyumconfig tool.

Fixes #19

### DIFF
--- a/etc/leapp/transaction/to_remove
+++ b/etc/leapp/transaction/to_remove
@@ -1,2 +1,1 @@
-yum
 redhat-rpm-config
--- a/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py
@@ -181,6 +181,12 @@ class PrepareUpgradeTransaction(Actor):
 
         debugsolver = True if os.environ.get('LEAPP_DEBUG', '0') == '1' else False
 
+        yum_script_path = self.get_tool_path('handleyumconfig')
+        cmd = ['--', '/bin/bash', '-c', yum_script_path]
+        _unused, error = preparetransaction.guard_container_call(overlayfs_info, cmd)
+        if error:
+            return error
+
         error = preparetransaction.mount_dnf_cache(overlayfs_info)
         if error:
             return error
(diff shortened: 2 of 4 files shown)


jmikovic:
```bash
[root@leapp-20190402100721 ~]# tree -l /etc/yum
/etc/yum
├── pluginconf.d -> ../dnf/plugins
│   ├── copr.conf
│   ├── copr.d
│   ├── debuginfo-install.conf
│   ├── langpacks.conf.rpmsave
│   ├── product-id.conf
│   ├── rhnplugin.conf.rpmsave
│   ├── spacewalk.conf
│   └── subscription-manager.conf
├── protected.d -> ../dnf/protected.d
│   ├── dnf.conf
│   ├── sudo.conf
│   └── systemd.conf
└── vars -> ../dnf/vars
```

there are still .rpmsave files left over, but that shouldn't cause any problems

other than that, looks good

PR #119, opened 2019-04-02
Add a warning message if found packages unsigned by Red Hat in system

asilveir:
The purpose of this PR is to alert the user to the non-signed Red Hat packages found in the system and to alert them that they can be removed in the upgrade process

In addition to the report this message will be printed on the console:

```
7 packages have not been signed by Red Hat and may be removed in the upgrade process
- leapp
- leapp-deps
- leapp-repository
- leapp-repository-deps
- leapp-repository-sos-plugin
- python2-leapp
- snactor
```


examon:
Tested this and it looks good.

PR #121, opened 2019-04-02
WIP: Delete unused actors

vojtechsokol:
NetIfaceScanner (and related models and topics) is removed as its code is unusable (parsing of journal etc.) and it's not even used in the IPU workflow.


vojtechsokol:
Duplicate of #122.

PR #122, opened 2019-04-02
Remove unused and unwanted actors

pirat89:
These three actors have never been used and we found those actors
are not needed even because:
- we want to keep firewalls enabled after the upgrade (additional
  work is needed yet)
- netifacescanner is broken and it is not written correctly. Guys,
  that would like to use such info, reported this structure as
  not usable or hard to use and they prepared own solutions. So
  we are going to remove the original scanner


pirat89:
@vojtechsokol good catch. Thanks. I will modify the doc. However, I am thinking the `NetworkInfoTopic` topic is something we will use soon. I will think about that yet. not sure whether make sense to remove it, on the other hand - in case we will not use it, nobody will realize we have mess in the code.

pirat89:
@vojtechsokol The topic is removed as well :)

vojtechsokol:
e2e tests

bocekm:
> additional work is needed yet
@pirat89, what did you mean by that?

PR #125, opened 2019-04-03
Change the way of warning about skipped packages

asilveir:
Previous we have always this output in the console:

```
2 packages will not be installed due to blacklisted repositories:
- ustr
- python3-pyxattr
```

Now this information will be present in _verbose/debug mode_ and also in report.


asilveir:
> Tests are using 'notify_skipped_packages' and therefore failing, please update those as well

Fixed, Thanks

PR #128, opened 2019-04-04
Add output to console in verbose mode about unsigned pkgs

asilveir:
Print in the console in **verbose mode** the information about packages that can be removed because they are not signed by Red Hat.

Right now this information is only present in the report but to make it more like  `PesEventsScanner` actor this change is needed.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/redhatsignedrpmcheck/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/redhatsignedrpmcheck/libraries/library.py
@@ -3,6 +3,7 @@ import os
 from leapp.exceptions import StopActorExecution
 from leapp.libraries.common import reporting
 from leapp.libraries.stdlib import api
+from leapp.libraries.stdlib.config import is_verbose
 from leapp.models import InstalledUnsignedRPM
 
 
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/el7toel8/actors/redhatsignedrpmcheck/libraries/library.py
```diff
@@ -30,6 +31,8 @@ def generate_report(packages):
         remediation=remediation,
         severity='high',
     )
+    if is_verbose():
+        api.show_message(summary)
```
jmikovic: what about using log.info here, since it would do the same but also give us info about where was this message produced
other option would be using some prefix or something to make this more visible to user, since show_message doesn't have any

asilveir: I tried some experiments with log and adding prefix.

But I think the default behaviour is good enough look:

```
2019-04-08 17:20:23.795 INFO     PID: 20311 leapp.workflow.Checks: Executing actor red_hat_signed_rpm_check
The following packages have not been signed by Red Hat and may be removed in the upgrade process:
- leapp
- leapp-deps
- leapp-repository
- leapp-repository-deps
- leapp-repository-sos-plugin
- python2-leapp
- snactor
```

Is your concern about which Actor you printed on the output?

PR #132, opened 2019-04-08
Py3 after upgrade

pirat89:
Start leapp with Python3 after RPMUpgradePhase
    
    We found several issues using Python2 after the RPM transaction, e.g.
    missing python libraries which are no longer provided for Python2 or
    mixture of loaded libraries from RHEL 7 and RHEL 8 when run leapp
    in one neverending process.
    
    To resolve this, we will stop the IPUWorkflow after the RPM
    transaction (RPMUpgradePhase) and restart it with Python3.
    Additionally, initramdisk is modified as well to start the new
    leapp script with python3 as expected.
    
    To stop the processing of the workflow, use the is_checkpoint flag
    which has been provided recently. Current solution is hacky as we
    are creating directory with files inside under the /root/ - but
    this is necessary hack now. We will try to come with better solution.
    
    Added RPM dependency for the python3 package and bumped capability.

**IMPORTANT**: You have to use the new initrd. (means, create that for yourself of follow the email with testing instructions - just internal, sorry).

### DIFF
--- a/packaging/leapp-el7toel8-deps.spec
+++ b/packaging/leapp-el7toel8-deps.spec
@@ -18,11 +18,12 @@ URL:        https://oamg.github.io/leapp/
 ##################################################
 %package -n %{lrdname}
 Summary:    Meta-package with system dependencies for leapp repository
-Provides:   leapp-repository-dependencies = 2
+Provides:   leapp-repository-dependencies = 3
 Obsoletes:  leapp-repository-deps
 
 Requires:   dnf >= 4
 Requires:   pciutils
+Requires:   python3
 
 %description -n %{lrdname}
 %{summary}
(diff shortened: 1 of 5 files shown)


vinzenz:
You want to use /usr/bin/platform-python instead of python3 otherwise you have to explicitly depend on python 3 and I think we can handle to depend on the platform python

pirat89:
@vinzenz we cannot use this one - I spoke with guys from Python. We really want to depend on Python3. I will add dependencies later, but you are right, that the spec file should be updated! Thanks for reminder.

pirat89:
SPEC files are updated. Additionally I updated the resumeservice actore to run in python3 as well. But we should think about better possible solution later.

PR #139, opened 2019-04-09
Add chrony check actor

mlichvar:
Check if chrony is installed and using the default chrony.conf. Warn
that the new default config in RHEL8 uses the leapsectz directive, which
does not work with leap smearing NTP servers.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkchrony/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkchrony/actor.py
@@ -1,0 +1,29 @@
+from leapp.actors import Actor
+from leapp.models import InstalledRedHatSignedRPM
+from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+from leapp.libraries.actor.library import check_chrony
+
+
+class CheckChrony(Actor):
+    """
+    Check for incompatible changes in chrony configuration.
+
+    Warn that the default chrony configuration in RHEL8 uses the leapsectz
+    directive.
+    """
+
+    name = 'check_chrony'
+    consumes = (InstalledRedHatSignedRPM,)
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        installed_packages = set()
+
+        signed_rpms = self.consume(InstalledRedHatSignedRPM)
+        for rpm_pkgs in signed_rpms:
+            for pkg in rpm_pkgs.items:
+                installed_packages.add(pkg.name)
+
+        check_chrony('chrony' in installed_packages)
--- a/repos/system_upgrade/el7toel8/actors/checkchrony/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/checkchrony/libraries/library.py
@@ -1,0 +1,28 @@
+from leapp.libraries.common import reporting
+from leapp.libraries.stdlib import api, run
+
+# Check if the chrony config file was not modified since installation
+def is_config_default():
+    try:
+        result = run(['rpm', '-V', '--nomtime', 'chrony'], checked=False)
+        return '/etc/chrony.conf' not in result['stdout']
+    except OSError as e:
+        api.current_logger().warn("rpm verification failed: %s", str(e))
+        return True
+
+# Report potential issues in chrony configuration
+def check_chrony(chrony_installed):
+    if not chrony_installed:
+        api.current_logger().info('chrony package is not installed')
+        return
+
+    if is_config_default():
+        reporting.report_generic(
+                title='chrony using default configuration',
+                summary='default chrony configuration in RHEL8 uses leapsectz directive, which cannot be used with leap smearing NTP servers, and uses a single pool directive instead of four server directives',
+                severity='medium')
+    else:
+        reporting.report_generic(
+                title='chrony using non-default configuration',
+                summary='chrony behavior will not change in RHEL8',
+                severity='low')
--- a/repos/system_upgrade/el7toel8/actors/checkchrony/tests/unit_test.py
+++ b/repos/system_upgrade/el7toel8/actors/checkchrony/tests/unit_test.py
@@ -1,0 +1,43 @@
+from leapp.libraries.actor import library
+from leapp.libraries.common import reporting
+
+
+class report_generic_mocked(object):
+    def __init__(self):
+        self.called = 0
+
+    def __call__(self, **report_fields):
+        self.called += 1
+        self.report_fields = report_fields
+
+
+def test_uninstalled(monkeypatch):
+    for config_default in (False, True):
+        monkeypatch.setattr(reporting, 'report_generic', report_generic_mocked())
+        monkeypatch.setattr(library, 'is_config_default', lambda: config_default)
+
+        library.check_chrony(False)
+
+        assert reporting.report_generic.called == 0
+
+
+def test_installed_defconf(monkeypatch):
+    monkeypatch.setattr(reporting, 'report_generic', report_generic_mocked())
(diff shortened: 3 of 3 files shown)


repos/system_upgrade/el7toel8/actors/checkchrony/actor.py
```diff
@@ -0,0 +1,47 @@
+from leapp.actors import Actor
+from leapp.models import InstalledRedHatSignedRPM
+from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+from leapp.libraries.common.reporting import report_generic
+from leapp.libraries.stdlib import CalledProcessError, run
+
+
+class CheckChrony(Actor):
```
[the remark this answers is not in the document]
mlichvar: Done.

repos/system_upgrade/el7toel8/actors/checkchrony/actor.py
```diff
@@ -0,0 +1,47 @@
...
+    produces = (Report)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        installed_packages = set()
+
+        signed_rpms = self.consume(InstalledRedHatSignedRPM)
+        for rpm_pkgs in signed_rpms:
+            for pkg in rpm_pkgs.items:
+	        installed_packages.add(pkg.name)
```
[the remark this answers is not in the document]
mlichvar: Fixed.

repos/system_upgrade/el7toel8/actors/checkchrony/actor.py
```diff
@@ -0,0 +1,53 @@
...
+        try:
+            run(['/bin/sh', '-c', 'rpm -V chrony | grep chrony.conf'])
+            default_chrony_conf = False
+        except (OSError, CalledProcessError):
+            default_chrony_conf = True
+
+        if default_chrony_conf:
+            report_generic(
+                    title='chrony using default configuration file',
+                    summary='default chrony configuration in RHEL8 uses leapsectz directive, which cannot be used with leap smearing NTP servers, and uses a single pool directive instead of four server directives',
```
fernflower: s/leap/leapp

vinzenz: @fernflower nope, this refers to leap as in leap seconds ;-)

repos/system_upgrade/el7toel8/actors/checkchrony/actor.py
```diff
@@ -0,0 +1,53 @@
...
+
+    name = 'check_chrony'
+    consumes = (InstalledRedHatSignedRPM,)
+    produces = (Report)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        installed_packages = set()
+
+        signed_rpms = self.consume(InstalledRedHatSignedRPM)
```
fernflower: Isn't common practice with the actors to use next(self.consumes) instead of directly referencing class?

repos/system_upgrade/el7toel8/actors/checkchrony/actor.py
fernflower: Oh my, I'll go brew some more coffee, thanks :coffee:

repos/system_upgrade/el7toel8/actors/checkchrony/actor.py
vinzenz: @fernflower you are confusing something
next(self.consume(...)) maybe is used mostly where they want at most one message.

Here the usage is perfectly valid. They get an iterable (at the moment this is a generator) with all messages that are of type InstalledRedHatSignedRPM

Then they iterate over them below (see for loop)

repos/system_upgrade/el7toel8/actors/checkchrony/actor.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    Check for incompatible changes in chrony configuration.
+
+    Warn that the default chrony configuration in RHEL8 uses the leapsectz
+    directive.
+    """
+
+    name = 'check_chrony'
+    consumes = (InstalledRedHatSignedRPM,)
+    produces = (Report)
```
vinzenz: missing trailing comma
```suggestion
    produces = (Report,)
```

repos/system_upgrade/el7toel8/actors/checkchrony/actor.py
```diff
@@ -0,0 +1,53 @@
...
+    def process(self):
+        installed_packages = set()
+
+        signed_rpms = self.consume(InstalledRedHatSignedRPM)
+        for rpm_pkgs in signed_rpms:
+            for pkg in rpm_pkgs.items:
+                installed_packages.add(pkg.name)
+
+        if 'chrony' not in installed_packages:
+            report_generic(
```
vinzenz: Hmm I wonder what's the purpose/value of this report.
Imagine someone would want to read the report and is told about all the packages that are not installed. 

Couldn't this be solved by a log message instead?

repos/system_upgrade/el7toel8/actors/checkchrony/actor.py
```diff
@@ -0,0 +1,53 @@
...
+                    summary='chrony package was not found in the rpm database',
+                    severity='low')
+            return
+
+        # Check if using default chrony.conf
+        try:
+            run(['/bin/sh', '-c', 'rpm -V chrony | grep chrony.conf'])
+            default_chrony_conf = False
+        except (OSError, CalledProcessError):
+            default_chrony_conf = True
```
vinzenz: Worth logging an error here in case of OSError, since a OSError shouldn't happen here.

Also I am wondering if it wouldn't be better to do something along these lines:

```python
default_chrony_conf = False
try:
    result = run(['rpm', '-V', 'chrony'], split=False)
    default_chrony_conf = 'chrony.conf' in result['stdout']
except (OSError, CalledProcessError) as e:
    self.log.warn("Failed to call chrony rpm verification failed: %s", str(e))
``` 
**Note** not checked or tested, just an idea

The reason for this is, that I am not a big fan of shell execution like this, and I think it's good enough here to check the output with python, using grep just complicates finding the source of programming errors.

repos/system_upgrade/el7toel8/actors/checkchrony/actor.py
```diff
@@ -0,0 +1,53 @@
...
+
+        if default_chrony_conf:
+            report_generic(
+                    title='chrony using default configuration file',
+                    summary='default chrony configuration in RHEL8 uses leapsectz directive, which cannot be used with leap smearing NTP servers, and uses a single pool directive instead of four server directives',
+                    severity='medium')
+        else:
+            report_generic(
```
vinzenz: Once again, is it worth reporting this fact?

vinzenz:
Please also add tests for the code above.

repos/system_upgrade/el7toel8/actors/checkchrony/actor.py
mlichvar: This would be nicer, but there is an issue that rpm -V returns with a non-zero exit code when the config files are modified and the run() method doesn't seem to have an option to ignore that. Should I try to extend that function first?

repos/system_upgrade/el7toel8/actors/checkchrony/actor.py
mlichvar: Ok. Makes sense.

repos/system_upgrade/el7toel8/actors/checkchrony/actor.py
mlichvar: This should be fixed now.

mlichvar:
Added unit tests.

AloisMahdal:
This was probably merged too soon; causes traceback: #195

PR #141, opened 2019-04-09
Do not try to install what's already installed

bocekm:
We used to see messages like the following one during transaction calculation:
```
Package nss-tools-3.36.0-7.1.el7_6.x86_64 is already installed.
Package glibc-2.17-260.el7.x86_64 is already installed.
```
That's because we were giving dnf the instruction to install even those pkgs, that were already installed.

Sometimes this caused conflicts resulting in transaction calculation failure:
```
Error:
       Problem 1: cannot install the best candidate for the job
        - conflicting requests
       Problem 2: package velocity-1.7-10.el7.noarch requires hsqldb, but none of the providers can be installed
        - conflicting requests
```
That happened when we mandated installation of already installed package (velocity), but from whatever reason it was not available in RHEL 8 repos, but mandated removal of the RHEL 7 version dependency (hsqldb).

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/peseventsscanner/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/peseventsscanner/actor.py
@@ -1,7 +1,8 @@
 from leapp.actors import Actor
-from leapp.libraries.actor.library import scan_events
-from leapp.models import InstalledRedHatSignedRPM, RepositoriesBlacklisted, RepositoriesSetupTasks, \
-RpmTransactionTasks
+from leapp.libraries.actor.library import pes_events_scanner
+from leapp.models import (InstalledRedHatSignedRPM, PESRpmTransactionTasks,
+                          RepositoriesBlacklisted, RepositoriesSetupTasks,
+                          RpmTransactionTasks)
 from leapp.reporting import Report
 from leapp.tags import FactsPhaseTag, IPUWorkflowTag
 
@@ -15,9 +16,9 @@ class PesEventsScanner(Actor):
     """
 
     name = 'pes_events_scanner'
-    consumes = (InstalledRedHatSignedRPM, RepositoriesBlacklisted,)
-    produces = (RpmTransactionTasks, RepositoriesSetupTasks, Report,)
+    consumes = (InstalledRedHatSignedRPM, RpmTransactionTasks, RepositoriesBlacklisted)
+    produces = (PESRpmTransactionTasks, RepositoriesSetupTasks, Report,)
     tags = (IPUWorkflowTag, FactsPhaseTag)
 
     def process(self):
-        scan_events('/etc/leapp/files/pes-events.json')
+        pes_events_scanner('/etc/leapp/files/pes-events.json')
--- a/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
@@ -2,51 +2,97 @@ import json
 import os
 from collections import namedtuple
 
-from leapp.exceptions import StopActorExecution
+from leapp.exceptions import StopActorExecution, StopActorExecutionError
 from leapp.libraries.common import reporting
 from leapp.libraries.stdlib import api
 from leapp.libraries.stdlib.config import is_verbose
-from leapp.models import InstalledRedHatSignedRPM, RepositoriesBlacklisted, RepositoriesSetupTasks,\
-    RpmTransactionTasks
-
+from leapp.models import (InstalledRedHatSignedRPM, PESRpmTransactionTasks,
+                          RepositoriesSetupTasks, RpmTransactionTasks, RepositoriesBlacklisted)
 
 # FIXME: this mapping is not complete and will need to be manually updated frequently
 REPOSITORIES_MAPPING = {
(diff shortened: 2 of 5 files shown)


bocekm:
Tested on our vagrant box (with all base + base optional pkgs) + updated _to_remove_ file from https://github.com/oamg/leapp-repository/pull/150. No issues.

repos/system_upgrade/el7toel8/actors/peseventsscanner/actor.py
```diff
@@ -15,9 +16,9 @@ class PesEventsScanner(Actor):
...
 
     name = 'pes_events_scanner'
-    consumes = (InstalledRedHatSignedRPM, RepositoriesBlacklisted,)
-    produces = (RpmTransactionTasks, RepositoriesSetupTasks, Report,)
+    consumes = (InstalledRedHatSignedRPM, RpmTransactionTasks, RepositoriesBlacklisted)
+    produces = (PESRpmTransactionTasks, RepositoriesSetupTasks, Report,)
     tags = (IPUWorkflowTag, FactsPhaseTag)
 
     def process(self):
-        scan_events('/etc/leapp/files/pes-events.json')
```
Rezney: pes_events_scanner('/etc/leapp/files/pes-events.json') is more self-descriptive

bocekm: Changed.

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -118,46 +244,55 @@ def map_repositories(to_install):
...
+            to_install.pop(pes_based_pkg_to_install)
+            pes_based_pkgs_not_to_be_installed.append(pes_based_pkg_to_install)
+    for pes_based_pkg_to_remove in to_remove.keys():
+        if pes_based_pkg_to_remove in (transaction_configuration.to_install + transaction_configuration.to_keep):
+            to_remove.pop(pes_based_pkg_to_remove)
+            pes_based_pkgs_not_to_be_removed.append(pes_based_pkg_to_remove)
+
+    if pes_based_pkgs_not_to_be_installed:
+        api.current_logger().debug('The following packages will not be installed '
+                                   'because of the transaction configuration files in /etc/leapp/transaction/:'
```
Rezney: Please add whole path to the config. E.g. /etc/leapp/transaction/to_remove

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -118,46 +244,55 @@ def map_repositories(to_install):
...
+                                   '\n- ' + '\n- '.join(pes_based_pkgs_not_to_be_installed))
+    if pes_based_pkgs_not_to_be_removed:
+        api.current_logger().debug('The following packages will not be removed '
+                                   'because of the transaction configuration files in /etc/leapp/transaction/:'
```
Rezney: same here...

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -56,41 +102,113 @@ def parse_file(path):
...
 
-def filter_events(events):
-    """ Filter PES events """
-    installed_pkgs = set()
-    for pkgs in api.consume(InstalledRedHatSignedRPM):
-        installed_pkgs.update([pkg.name for pkg in pkgs.items])
+def parse_entry(entry):
+    """Parse PES event data"""
+    action = parse_action(entry['action'])
+    in_pkgs = parse_packageset(entry.get('in_packageset', {}) or {})
```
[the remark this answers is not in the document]
bocekm: Let me see. I'm not the author.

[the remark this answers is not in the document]
vinzenz: Well not if the entry could be none, you only get the default value if the key does not exist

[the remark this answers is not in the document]
bocekm: 1. PES currently does not produce events without 'in_packageset' or 'out_packageset'. If there are no "out" packages, PES gives you `'out_packageset': null` in JSON translating to  `'out_packageset': None` in Python. Meaning that `entry.get('out_packageset', {})` gives you `None` and not `{}`. And we don't want to pass `None` to the `parse_packageset()`.
1. Simplification: I changed it to `in_pkgs = parse_packageset(entry.get('in_packageset') or {})` because {} does the same job as the _get()_'s default `None` when using it with the shortcircuiting `or` operator.
   Note: We could also use `in_pkgs = parse_packageset(entry['out_packageset'] or {})` because the PES events currently always contain the `out_packageset` key, but you know, maybe it won't be like that in future and then it would crash with KeyError.

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
bocekm: Done

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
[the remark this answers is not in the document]
Rezney: Ahh I see, thanks for the clarification guys...

PR #148, opened 2019-04-12
Fix PES repository ID mapping to use new repositories for GA

artmello:


repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -12,8 +12,8 @@
 REPOSITORIES_BLACKLIST = ('rhel8-buildroot', 'rhel8-crb')
 # FIXME: this mapping is not complete and will need to be manually updated frequently
 REPOSITORIES_MAPPING = {
-    'rhel8-appstream': 'rhel-8-for-x86_64-appstream-htb-rpms',
-    'rhel8-baseos': 'rhel-8-for-x86_64-baseos-htb-rpms'}
+    'rhel8-appstream': 'rhel-8-for-x86_64-appstream-rpms',
```
vojtechsokol: Is there a reason why the repomapping is hardcoded?

vojtechsokol: Is this PR to be merged before GA? Attempt to upgrade fails in `prepare_upgrade_transaction` actor with following error:
```
Cannot find any required target repository. The list of available required repositories is empty.
```

artmello: About why it is hardcoded, PES stores repositories IDs in a different format that the one will be used by RHSM. That makes sense since they don't need to always update the database for HTB/GA and so on. In the future, I think PES data will come with some sort of mapping and this will not be necessary to be done by Leapp any more. For a mid term solution, we will move this mapping data to repomap.csv and update the actors (I am doing this right now in a different PR).

This PR needs to be landed before GA, since we should use GA repositories and not HTB. About the error you are facing, I am not totally sure. It can be related with how you are testing it or some other issue. Few days ago we updated both leapp-repository (adding new product certification file for GA) and Vagrant box (disabling HTB SKU and updating repomap.csv to point to new repositories). Please, confirm you have these new data since they will be necessary to access the repositories, If the problem persist we would need to investigate.

artmello: I forgot to mention, the mid term solution (moving data to repomap.csv) may or may not land with GA. It depends on when the final build will be done and when the fix will be done/reviewed/merged. That is why this fix needs to land for sure (ofc, if it works :))

bocekm: I have tested this and it works. This is not that crucial because base on this mapping the repos mentioned in PES are enabled. But, the BaseOS and AppStream repos are enabled either way in the repositoriesmapping actor.

bocekm: And to the error Vojta has. I agree with Arthur - we have updated the RHSM product certs from HTB to GA. And for example our internal Satellite does not have the GA repos available.

Issue #130, opened 2019-04-05
kernel-rpm-macros fails transaction


bocekm:
**Actual behavior**
We have redhat-rpm-config in /etc/leapp/transaction/to_remove meaning that it is always removed. But this package is required by kernel-rpm-macros, so the transaction fails if kernel-rpm-macros is installed:
```
Error: 
 Problem 1: cannot install the best candidate for the job
  - conflicting requests
 Problem 2: package kernel-rpm-macros-116-1.el8.noarch requires redhat-rpm-config >= 13, but none of the providers can be installed
  - cannot install the best candidate for the job
  - conflicting requests
```

**To Reproduce**
Steps to reproduce the behavior
1. install kernel-rpm-macros
1. run leapp upgrade --debug
1. See error

**Expected behavior**
Transaction does not fail with kernel-rpm-macros installed. It can be done by adding it to the  to_remove file as well.

bocekm:
It'd be better to rather remove the redhat-rpm-config from the to_remove.
```
 Problem 11: package python2-devel-2.7.15-21.module+el8+2540+b19c9b35.x86_64 requires redhat-rpm-config, but none of the providers can be installed
  - package python2-numpy-f2py-1:1.14.2-9.module+el8+2540+b19c9b35.x86_64 requires python2-devel, but none of the providers can be installed
  - package redhat-rpm-config-116-1.el8.noarch requires perl-srpm-macros, but none of the providers can be installed
  - redhat-rpm-config-9.1.0-87.el7.noarch does not belong to a distupgrade repository
  - conflicting requests
```

pirat89:
This would be possible only with the https://github.com/oamg/leapp-repository/pull/44 PR, otherwise installation fails.

PR #150, opened 2019-04-12
Remove packages depending on redhat-rpm-config + tpm2-abrmd

bocekm:
- We are unable to install redhat-rpm-config because it uses rich deps.
- All packages that depend on it will fail the transaction.
- It's better to remove them instead of transaction failure.
- We've identified a second package that uses rpm rich deps - tpm2-abrmd. So we're removing it too.

Resolves https://github.com/oamg/leapp-repository/issues/130


bocekm:
Added tpm2-abrmd to the list as it uses rpm rich deps too.

PR #153, opened 2019-04-14
persistentnetnamesdisable: introduce new actor

msekletar:
Actor will disable udev's persistent network naming feature when we
detect that we are upgrading system which has only single eth0
interface. When multiple ethX interfaces are detected we abort the
upgrade and point the user to relevant documentation on how he can
resolve the issue.


pirat89:
Waiting for build of new initrd images from PR #158 (will be built after the merge or PR #158 into the master)

pirat89:
Left the world in :fire: :-) +  :eyeglasses:

PR #154, opened 2019-04-15
Add ha cluster check actor

idevat:
Check if HA Cluster is in use. If yes, inhibit the upgrade process.


jmikovic:
please rebase

PR #158, opened 2019-04-16
dracut initrd: add sync commands and fix mounting from fstab in container after upgrade phase

pirat89:
We found some issues that are *maybe* related to cache. Use the sync
command on several places in initrd to ensure we are not hit by that
on some machines.

ADDITIONALLY!!!!
There are errors made because the partitions from FSTAB are not
mounted. It should be save to call "mount -a" after the upgrade
as already proceed earlier and we require to mount additional
partitions to be able to process the rest of workflow.

### DIFF
--- a/sources/dracut/85sys-upgrade-redhat/do-upgrade.sh
+++ b/sources/dracut/85sys-upgrade-redhat/do-upgrade.sh
@@ -51,7 +51,14 @@ do_upgrade() {
         # run leapp to proceed phases after the upgrade with Python3
         #PY_LEAPP_PATH=/usr/lib/python2.7/site-packages/leapp/
         #$NEWROOT/bin/systemd-nspawn $nspawn_opts -D $NEWROOT -E PYTHONPATH="${PYTHONPATH}:${PY_LEAPP_PATH}" /usr/bin/python3 $LEAPPBIN upgrade --resume $args
-        $NEWROOT/bin/systemd-nspawn $nspawn_opts -D $NEWROOT /usr/bin/python3 $LEAPP3_BIN upgrade --resume $args
+
+        # NOTE:
+        # mount everything from FSTAB before run of the leapp as mount inside
+        # the container is not persistent and we need to have mounted /boot
+        # all FSTAB partitions. As mount was working before, hopefully will
+        # work now as well. Later this should be probably modified as we will
+        # need to handle more stuff around storage at all.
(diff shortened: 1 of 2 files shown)


sources/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -39,6 +39,7 @@ do_upgrade() {
     nspawn_opts="--capability=all --bind=/sys --bind=/dev --bind=/proc --bind=/run/udev --keep-unit --register=no"
     $NEWROOT/bin/systemd-nspawn $nspawn_opts -D $NEWROOT $LEAPPBIN upgrade --resume $args
     rv=$?
+    sync
```
bocekm: Add at least some comment to this so people won't be wondering why it's here when looking at it in a year from now.

sources/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -48,7 +51,7 @@ do_upgrade() {
         # run leapp to proceed phases after the upgrade with Python3
         #PY_LEAPP_PATH=/usr/lib/python2.7/site-packages/leapp/
         #$NEWROOT/bin/systemd-nspawn $nspawn_opts -D $NEWROOT -E PYTHONPATH="${PYTHONPATH}:${PY_LEAPP_PATH}" /usr/bin/python3 $LEAPPBIN upgrade --resume $args
-        $NEWROOT/bin/systemd-nspawn $nspawn_opts -D $NEWROOT /usr/bin/python3 $LEAPP3_BIN upgrade --resume $args
+        $NEWROOT/bin/systemd-nspawn $nspawn_opts -D $NEWROOT /usr/bin/bash -c "mount -a; /usr/bin/python3 $LEAPP3_BIN upgrade --resume $args"
```
vojtechsokol: Is it safe to chang the mount and pytho3 command like that? I.e. what will happend if `mount -a` fails?

pirat89: see the commit message. if it fails we are done as without mounted partitions the upgrade will fail in majority cases as well

vinzenz: Duh omg right this is the problem I totally forgot that part

PR #165, opened 2019-04-17
Fix inhibition when ethX + other ifaces exist

pirat89:
Currently the inhibition doesn't work at all because the Report
module is not included in the tuple of `Produces` and original
condition was wrong. Now it will inhibit upgrade as expected when
ethX network interface exists and there are other network interfaces.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/persistentnetnamesdisable/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/persistentnetnamesdisable/actor.py
@@ -2,7 +2,8 @@ import re
 
 from leapp.actors import Actor
 from leapp.libraries.common import reporting
-from leapp.models import PersistentNetNamesFacts, KernelCmdlineArg
+from leapp.models import KernelCmdlineArg, PersistentNetNamesFacts
+from leapp.reporting import Report
 from leapp.tags import FactsPhaseTag, IPUWorkflowTag
 
 
@@ -13,7 +14,7 @@ class PersistentNetNamesDisable(Actor):
 
     name = 'persistentnetnamesdisable'
     consumes = (PersistentNetNamesFacts,)
-    produces = (KernelCmdlineArg,)
+    produces = (KernelCmdlineArg, Report)
     tags = (FactsPhaseTag, IPUWorkflowTag)
 
     def ethX_count(self, interfaces):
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/el7toel8/actors/persistentnetnamesdisable/actor.py
```diff
@@ -14,7 +15,7 @@ class PersistentNetNamesDisable(Actor):
 
     name = 'persistentnetnamesdisable'
     consumes = (PersistentNetNamesFacts,)
-    produces = (KernelCmdlineArg,)
+    produces = (KernelCmdlineArg, Report,)
```
vinzenz:
```suggestion
    produces = (KernelCmdlineArg, Report)
```

repos/system_upgrade/el7toel8/actors/persistentnetnamesdisable/actor.py
```diff
@@ -1,10 +1,11 @@
 import re
 
-from leapp.libraries.common import reporting
 
 from leapp.actors import Actor
 from leapp.models import PersistentNetNamesFacts, KernelCmdlineArg
 from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+from leapp.reporting import Report
+from leapp.libraries.common import reporting
```
artmello: Sort of nit picking, but do you mind sorting the leapp imports alphabetically?

repos/system_upgrade/el7toel8/actors/persistentnetnamesdisable/actor.py
```diff
@@ -38,7 +39,7 @@ def process(self):
 
         if self.single_eth0(interfaces):
             self.disable_persistent_naming()
-        elif self.ethX_count(interfaces) > 1:
+        elif self.ethX_count(interfaces) > 0 and len(interfaces) > 1:
```
artmello: Again, pretty minor and no blocking, but checking the length of interfaces is cheaper than searching for ethX, so I would change the order of checks.

repos/system_upgrade/el7toel8/actors/persistentnetnamesdisable/actor.py
pirat89: I can.

repos/system_upgrade/el7toel8/actors/persistentnetnamesdisable/actor.py
pirat89: Good point. Just from my point of view, this looks more readable. But I will swap them

pirat89:
leapp-ci build

pirat89:
leapp-ci build

PR #167, opened 2019-04-18
dracut: Ensure the $NEWROOT is set and use default otherwise

pirat89:
dracut: Ensure the $NEWROOT is set and use default otherwise
    
    We hit *somtimes* on some machines problem that we are not even able
    to mount the rootfs. In such cases, there is just err message like
           mount: bad usage
           Try 'mount --help' for more information.
    
    It's weird, some modules checks the  $NEWROOT variable, some not.
    So just from this point to ensure whether this would be the source
    of the issue, putting here the initialisation when NEWROOT is not
    defined.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
+++ b/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
@@ -11,6 +11,8 @@ export LEAPPBIN=/usr/bin/leapp
 export LEAPPHOME=/root/tmp_leapp_py3
 export LEAPP3_BIN=$LEAPPHOME/leapp3
 
+export NEWROOT=${NEWROOT:-"/sysroot"}
+
 do_upgrade() {
     local args="" rv=0
     #FIXME: set here params we would like to possible use...


AloisMahdal:
> from this point to ensure whether this would be the source of the issue

I don't see how defaulting gets you closer to know source of the issue.  From what I've seen the missing value seems like a symptom, not a cause of the issue.

Also if the path is of any importance, defaulting does not seem safe to me.  Either you need the variable because someone needs to tell you where the root is, or it's always `/sysroot` (but then we wouldn't have variable for that, would we?).

pirat89:
@AloisMahdal  @vinzenz  Guys, even the creation of the dir make sense and setting the default for $NEWROOT as well. It is used in various dracut modules - exactly as I am doing that - and it is not done just one time like that, but several times. You can spend time to find people who understand why it is written like that for last 10+ years. Maybe Harald would know why dracut modules are written in that way. I have no clue now.

AloisMahdal:
I don't mean to block this, just please understand that setting the default based on "how other people do it" seems like cargo-cult programming.  Maybe good enough for hotfix, but we should be suspicious about the module and find out how to write it properly.  (I'm pretty sure someone somewhere understands how the modules are supposed to operate.)

pirat89:
@vinzenz I found in the end that creation of the directory doesn't make sense really, as we want to remount some stuff and in case rootfs is not mounted, it crashes anyway. So removing the line.

PR #168, opened 2019-04-18
[WIP]: Improve report messages

bocekm:
Related: OAMG-888

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
@@ -251,13 +251,16 @@ def filter_out_pkgs_in_blacklisted_repos(to_install):
     # FIXME The to_install contains just a limited subset of packages - those that are *not* currently installed and
     # are to be installed. But we should also warn about the packages that *are* installed.
     blacklisted_pkgs = set()
+    blacklisted_repos = get_repositories_blacklisted()
     for pkg, repo in to_install.items():
-        if repo in get_repositories_blacklisted():
+        if repo in blacklisted_repos:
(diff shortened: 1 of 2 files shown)


vinzenz:
Need a rebase and merge

vojtechsokol:
leapp-ci build

bocekm:
I still need to improve it a bit.

ZhukovGreen:
closed in favour of https://github.com/oamg/leapp-repository/pull/519

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -251,13 +251,16 @@ def filter_out_pkgs_in_blacklisted_repos(to_install):
     # FIXME The to_install contains just a limited subset of packages - those that are *not* currently installed and
     # are to be installed. But we should also warn about the packages that *are* installed.
     blacklisted_pkgs = set()
+    blacklisted_repos = get_repositories_blacklisted()
```
bocekm, line 254: @ZhukovGreen, I've done this change to not call the `get_repositories_blacklisted()` in every pass of the for loop. And there might be a lot of passes.
I know that gcc is able to do such optimization under the hood. I don't think that Python can do that as well.

ZhukovGreen: @bocekm Ok, I will consider this;)

ZhukovGreen: @bocekm This makes sense. Moreover, in the upstream, this change was added in the beginning of the year by Vinny

PR #170, opened 2019-04-19
using directory listing of /sys/module instead of lsmod fixes #46

### LINKED ISSUE #46: Replace the call to lsmod by a dir listing routine
The call to lsmod at https://github.com/oamg/leapp-repository/blob/88a3a26cd8bc860f5128a5f22313ad09b402d4a4/repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py#L86
can be replaced by a directory listing, pointing to /sys/module/

One good aspect of this change is that listing /sys/module will also give you visibility
over modules that are compiled as built-in (that's why it has more entries than shown in lsmod).

Seenivasanseeni:
As Issue #46 needed, This PR uses such directory listing. 
Possible: Since This directory listing can be done with the simple os.listdir(), should this PR be modified with such call ?

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
+++ b/repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
@@ -77,13 +77,10 @@ def get_system_groups_status():
 
 @aslist
 def _get_active_kernel_modules(logger):
-    lines = run(['ls',"-d","/sys/module"], split=True)['stdout'] # can be implemented by os.lisdir(\'/sys/module\')
-    for l in lines[1:]:
-        name = l.split(' ')[0]
+    lines = [module for module in os.listdir("/sys/module") if os.path.isdir("/sys/module/{}".format(module))]
 
-        # Read parameters of the given module as exposed by the
-        # `/sys` VFS, if there are no parameters exposed we just
-        # take the name of the module
+    for l in lines:
+        name = l
         base_path = '/sys/module/{module}'.format(module=name)
         parameters_path = os.path.join(base_path, 'parameters')
         if not os.path.exists(parameters_path):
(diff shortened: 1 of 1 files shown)


Seenivasanseeni:
I have added the change requested. Can you review @pirat89 ?

repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -77,13 +77,10 @@ def get_system_groups_status():
...
-    lines = run(['lsmod'], split=True)['stdout']
-    for l in lines[1:]:
-        name = l.split(' ')[0]
+    lines = [module for module in os.listdir("/sys/module") if os.path.isdir("/sys/module/{}".format(module))]
 
-        # Read parameters of the given module as exposed by the
-        # `/sys` VFS, if there are no parameters exposed we just
-        # take the name of the module
+    for l in lines:
+        name = l
```
pirat89: - We should always use `os.path.join` instead of format string as was used originally already. We missed that earlier as I see.
- nitpick - when it is possible use for strings same literal as it is used in the rest of code / file. In this case prefer apostrophe instead of quotes
- It is not good idea to use variable names like `l`. One of problems is, that it can be simply interchanged with `I` or `1` (depends on used font). Additionally, it is confusing to get information what the variable represent. In this case, the rename on the another line gives you hint that you should do that on one line of code.
- You do not need to use `lines` name anymore as now it is not output of subcommand, but you already provide information what modules exist on the system.

What about something like:
```python
    modules = [module for module in os.listdir('/sys/module') if os.path.isdir(os.path.join('/sys/module/', module))]
    for module in modules:
        base_path = os.path.join('/sys/module/', module)
```

pirat89:
leapp-ci build

pirat89:
leapp-ci build

repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
bocekm: @Seenivasanseeni, could you please take a look?

bocekm:
Hi @Seenivasanseeni, could you please update the actor based on the comments? Thanks.

Seenivasanseeni:
Since the branch has moved further, I will close this and create a new pr with all the changes.

PR #172, opened 2019-04-24
Check kernel drivers actor

examon:
- file `files/removed_modules.txt` contains list of kernel modules that are in the RHEL7 but not in RHEL8
- actor consumes `ActiveKernelModulesFacts` and checks against `files/removed_modules.txt` if any module present is missing in the RHEL8 (and inhibits the upgrade if it is)


jmikovic:
while the code looks fine it seems to inhibit every time since even on clean VM or vagrant box there are always some of the removed modules

shaded-enmity:
List of modules from the Vagrant box that are no longer supported:

```
ablk_helper
crct10dif_common
cryptd
floppy
gf128mul
glue_helper
iosf_mbi
pata_acpi
virtio
virtio_pci
virtio_ring
```

Obtained via:
```bash
comm -12 <(lsmod | cut -d' ' -f1 | sort -u) removed_modules.txt
```

Most of these look benign however I'm a little nervous about the `virtio` ones, especially since the beta docs mention them (https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8-beta/html/configuring_and_managing_virtualization/virtualization-in-rhel-8-an-overview_configuring-and-managing-virtualization) - we need to get some clarity here.

examon:
leapp-ci build

examon:
leapp-ci build

examon:
> while the code looks fine it seems to inhibit every time since even on clean VM or vagrant box there are always some of the removed modules

With the latest changes, this is not longer the case. Actor does not inhibit upgrade on the clean installed vm via vagrant.

repos/system_upgrade/el7toel8/actors/checkkernelmodules/actor.py
```diff
@@ -0,0 +1,58 @@
...
+                for active_module in fact.kernel_modules:
+                    if active_module.filename in removed_modules \
+                        and active_module.filename not in whitelisted_modules:
+                        modules_to_report.add(active_module.filename)
+
+            if modules_to_report:
+                title = "Detected unavailable kernel modules. Upgrade can't proceed."
+                summary = "The following kernel modules are not available on" \
+                          " the RHEL 8 system: {}".format(modules_to_report)
+                remediation = "Disable detected unavailable kernel modules in" \
```
vinzenz: We prefer single quote and parenthesis for multi line strings instead of \ joined strings

repos/system_upgrade/el7toel8/actors/checkkernelmodules/actor.py
```diff
@@ -0,0 +1,58 @@
...
+            removed_modules = [x.strip() for x in removed.readlines()]
+            modules_to_report = set()
+
+            # FIXME:
+            # These modules are from the files/removed_modules.txt.
+            # However, they are present on the clean RHEL7.6 installation
+            # in virtual machine via vagrant. They are whitelisted because
+            # we do not want to inhibit upgrade from clean installation.
+            whitelisted_modules = [
+                "ablk_helper", "crct10dif_common", "cryptd", "floppy",
```
vinzenz: " -> '

repos/system_upgrade/el7toel8/actors/checkkernelmodules/actor.py
```diff
@@ -0,0 +1,58 @@
...
+        with open("files/removed_modules.txt", "r") as removed:
+            removed_modules = [x.strip() for x in removed.readlines()]
+            modules_to_report = set()
+
+            # FIXME:
+            # These modules are from the files/removed_modules.txt.
+            # However, they are present on the clean RHEL7.6 installation
+            # in virtual machine via vagrant. They are whitelisted because
+            # we do not want to inhibit upgrade from clean installation.
+            whitelisted_modules = [
```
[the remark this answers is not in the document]
examon: > TBH I wonder if we don't want to make a whitelist message instead, to allow mitigation by an actor

We could do that. What do you think @shaded-enmity ?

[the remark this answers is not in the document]
shaded-enmity: Yeah that is a very good point, we can get rid of the `FIXME:` entirely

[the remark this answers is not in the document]
examon: @vinzenz @shaded-enmity So, I added the `WhitelistKernelModules` actor that sends the `WhitelistedKernelModules` message to the `CheckKernelModules` actor. `CheckKernelModules` does the inhibition based on received the message.

repos/system_upgrade/el7toel8/actors/kernel/checkkernelmodules/actor.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def process(self):
+        with open("files/removed_modules.txt", "r") as removed:
+            removed_modules = [x.strip() for x in removed.readlines()]
+            whitelisted_modules = set()
+            modules_to_report = set()
+
+            for fact in self.consume(WhitelistedKernelModules):
+                for module in fact.whitelisted_modules:
+                    whitelisted_modules.add(module)
```
vinzenz:
```suggestion
```

repos/system_upgrade/el7toel8/actors/kernel/checkkernelmodules/actor.py
```diff
@@ -0,0 +1,52 @@
...
+
+            for fact in self.consume(ActiveKernelModulesFacts):
+                for active_module in fact.kernel_modules:
+                    if active_module.filename in removed_modules and \
```
vinzenz: Instead of a complex check like this I could imagine this:
```python
            for fact in self.consume(ActiveKernelModulesFacts):
                for active_module in fact.kernel_modules:
                    if activate_module.filename in whitelisted_modules:
                        continue
                    if active_module.filename in removed_modules:
                        modules_to_report.add(active_module.filename)
```

But this is up to you

examon: I like this. It is cleaner.

fernflower: (Just thinking aloud)
If we had create_lookup moved out of libraries/rpms.py to a more general place (leapp libraries? some general library in leapp-repository) you could theoretically form a lookup and use set difference
```
# currently it's here so importing from rpms
from leapp.libraries.common.rpms import create_lookup
...
active_modules = create_lookup(ActiveKernelModulesFacts, 'kernel_modules', 'filename')
modules_to_report = active_modules - (set(removed_modules) - set(whitelisted_modules))
```

vinzenz:
@examon 
LGTM - But I don't have time for testing right now, please find a different reviewer for that.

repos/system_upgrade/el7toel8/actors/kernel/checkkernelmodules/files/removed_modules.txt
```diff
@@ -0,0 +1,254 @@
+3w-9xxx
```
bocekm: I'd like to see a comment line describing where did we get the list. We got it by comparing RHEL 7 and RHEL 8 modules. But what version/compose of these systems?

repos/system_upgrade/el7toel8/actors/kernel/checkkernelmodules/actor.py
```diff
@@ -0,0 +1,52 @@
...
+
+            for fact in self.consume(ActiveKernelModulesFacts):
+                for active_module in fact.kernel_modules:
+                    if active_module.filename in whitelisted_modules:
+                        continue
+                    if active_module.filename in removed_modules:
+                        modules_to_report.add(active_module.filename)
+
+            if modules_to_report:
+                title = 'Detected unavailable kernel modules. Upgrade cannot proceed.'
```
bocekm:
```suggestion
                title = 'Detected loaded kernel drivers which have been removed in RHEL 8. Upgrade cannot proceed.'
```

repos/system_upgrade/el7toel8/actors/kernel/checkkernelmodules/actor.py
```diff
@@ -0,0 +1,52 @@
...
+                summary = ('The following kernel modules are not available on '
```
bocekm: "Support for the following currently loaded RHEL 7 device drivers has been removed in RHEL 8: <list> See https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/considerations_in_adopting_rhel_8/index#removed-device-drivers_hardware-enablement for details."

repos/system_upgrade/el7toel8/actors/kernel/checkkernelmodules/actor.py
```diff
@@ -0,0 +1,52 @@
...
+                           'the RHEL 8 system: \n     - {}'.format('\n     - '.join(modules_to_report)))
+                remediation = ('Disable detected unavailable kernel modules in '
```
bocekm: Tell the user how to disable the kernel drivers.

repos/system_upgrade/el7toel8/actors/kernel/checkkernelmodules/actor.py
```diff
@@ -0,0 +1,52 @@
+from leapp.actors import Actor
+from leapp.libraries.common.reporting import report_with_remediation
+from leapp.models import ActiveKernelModulesFacts, WhitelistedKernelModules
```
bocekm: It would be nice to rename the ActiveKernelModulesFacts to LoadedKernelModulesFacts.

repos/system_upgrade/el7toel8/actors/kernel/checkkernelmodules/actor.py
```diff
@@ -0,0 +1,52 @@
...
+
+            for fact in self.consume(WhitelistedKernelModules):
+                whitelisted_modules.update(fact.whitelisted_modules)
+
+            for fact in self.consume(ActiveKernelModulesFacts):
+                for active_module in fact.kernel_modules:
+                    if active_module.filename in whitelisted_modules:
+                        continue
+                    if active_module.filename in removed_modules:
+                        modules_to_report.add(active_module.filename)
```
bocekm: Before inhibiting, check if the driver is used by any device: `udevadm info -e` and look for `DRIVER=`.

examon:
I going to wait for Pavel to compile new list of kernel drivers before making changes.

examon:
leapp-ci build

repos/system_upgrade/el7toel8/actors/kernel/checkkernelmodules/actor.py
examon: We can do this in the separate PR. This will require changes across multiple actors.

repos/system_upgrade/el7toel8/actors/kernel/checkkerneldrivers/actor.py
```diff
@@ -0,0 +1,82 @@
...
+        with open('files/removed_drivers.txt', 'r') as removed:
+            removed_drivers = []
+            whitelisted_modules = set()
+            collected_drivers = set()
+            drivers_to_report = set()
+
+            # Extracting kernel drivers from the files/removed_drivers.txt.
+            for line in removed.readlines():
+                token = line.strip()
+                if token.startswith('#') or len(token) == 0:
```
vinzenz:
```suggestion
                if not token or token.startswith('#'):
```

repos/system_upgrade/el7toel8/actors/kernel/checkkerneldrivers/actor.py
```diff
@@ -0,0 +1,82 @@
...
+            for fact in self.consume(ActiveKernelModulesFacts):
+                for active_module in fact.kernel_modules:
+                    if active_module.filename in whitelisted_modules:
+                        continue
+                    if active_module.filename in removed_drivers:
+                        collected_drivers.add(active_module.filename)
+
+            # Going over collected drivers and considering for reporting only
+            # those drivers that are currently used by some device.
+            udevadm_out = run('udevadm info -e'.split(), split=True)['stdout']
```
vinzenz: Please move this out into a separate actor in the facts collection phase
And please do not use 'cmd string whatever'.split() due to laziness.

repos/system_upgrade/el7toel8/actors/kernel/checkkerneldrivers/actor.py
```diff
@@ -0,0 +1,84 @@
...
+            # - currently being used by some device
+            if drivers_to_report:
+                title = ('Detected loaded kernel drivers which have been removed '
+                         'in RHEL 8. Upgrade cannot proceed.')
+                URL = ('https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/'
+                       'considerations_in_adopting_rhel_8/index#removed-device-drivers_hardware-enablement')
+                summary = ('Support for the following currently loaded RHEL 7 '
+                           'device drivers has been removed in RHEL 8: \n     - {}'
+                           '\nPlease see {} for details.'.format('\n     - '.join(drivers_to_report), URL))
+                remediation = ('Please disable detected kernel drivers in '
```
vinzenz: @Rezney shouldn't this be now some kind of command? Or is this ok too?

Rezney: As discussed with @shaded-enmity we will be obsoleting/refactoring this. But I understand you concern. This should have probably be differentiated something like "remediation_command" vs "remediation_hint".

We have something similar even here:

https://github.com/oamg/leapp-repository/blob/38af7da7dd3355b573284ebbda825e333f74e1c4/repos/system_upgrade/el7toel8/actors/checkbrltty/actor.py#L26

So this should probably not block this PR...

PR #174, opened 2019-04-26
Add SSSDCheck actor

pbrezina:
Check SSSD configuration for changes in RHEL8 and report them.

These changes are:
- id_provider=local is no longer supported and will be ignored
- ldap_groups_use_matching_rule_in_chain was removed and will be ignored
- ldap_initgroups_use_matching_rule_in_chain was removed and will be ignored
- ldap_sudo_include_regexp changed default from true to false

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/sssdcheck/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/sssdcheck/actor.py
@@ -1,0 +1,64 @@
+from leapp.actors import Actor
+from leapp.libraries.common.reporting import report_generic, report_with_remediation
+from leapp.models import SSSDConfig
+from leapp.reporting import Report
+from leapp.tags import IPUWorkflowTag, ChecksPhaseTag
+
+
+class SSSDCheck(Actor):
+    '''
+    Check SSSD configuration for changes in RHEL8 and report them.
+
+    These changes are:
+    - id_provider=local is no longer supported and will be ignored
+    - ldap_groups_use_matching_rule_in_chain was removed and will be ignored
+    - ldap_initgroups_use_matching_rule_in_chain was removed and will be ignored
+    - ldap_sudo_include_regexp changed default from true to false
+    '''
+
+    name = 'sssd_check'
+    consumes = (SSSDConfig,)
+    produces = (Report,)
+    tags = (IPUWorkflowTag, ChecksPhaseTag)
+
+    def process(self):
+        model = next(self.consume(SSSDConfig), None)
+        if not model:
+            return
+
+        for domain in model.domains:
+            if 'local_provider' in domain.options:
+                self.reportLocalProvider(domain)
+
+            if 'groups_chain' in domain.options:
+                self.reportRemovedOption(domain, 'ldap_groups_use_matching_rule_in_chain')
+
+            if 'initgroups_chain' in domain.options:
+                self.reportRemovedOption(domain, 'ldap_initgroups_use_matching_rule_in_chain')
+
+            if 'sudo_regexp' in domain.options:
+                self.reportSudoRegexp(domain)
+
+    def reportLocalProvider(self, domain):
+        report_generic(
+            title='SSSD Domain "%s": local provider is no longer '
+                  'supported and the domain will be ignored.' % domain,
+            summary='Local provider is no longer supported.'
+        )
+
+    def reportRemovedOption(self, domain, option):
+        report_generic(
+            title='SSSD Domain "%s": option %s has no longer '
+                  'any effect' % (domain, option),
+            summary='Option %s was removed and it will be ignored.' % option
+        )
+
+    def reportSudoRegexp(self, domain):
+        report_with_remediation(
+            title='SSSD Domain "%s": sudo rules containing wildcards '
+                  'will stop working.' % domain,
+            summary='Default value of ldap_sudo_include_regexp changed '
+                    'from true to false for performance reason.',
+            remediation='If you use sudo rules with wildcards, set this option '
+                        'to true explicitly.'
+        )
--- a/repos/system_upgrade/el7toel8/actors/sssdfacts/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/sssdfacts/actor.py
@@ -1,0 +1,34 @@
+from six.moves import configparser
+
+from leapp.actors import Actor
+from leapp.libraries.actor.library import SSSDFactsLibrary
+from leapp.models import SSSDConfig
+from leapp.tags import IPUWorkflowTag, FactsPhaseTag
+
+
+class SSSDFacts(Actor):
+    '''
+    Check SSSD configuration for changes in RHEL8 and report them in model.
+
+    These changes are:
+    - id_provider=local is no longer supported and will be ignored
+    - ldap_groups_use_matching_rule_in_chain was removed and will be ignored
+    - ldap_initgroups_use_matching_rule_in_chain was removed and will be ignored
+    - ldap_sudo_include_regexp changed default from true to false
+    '''
+
+    name = 'sssd_facts'
+    consumes = ()
+    produces = (SSSDConfig,)
+    tags = (IPUWorkflowTag, FactsPhaseTag)
+
+    def process(self):
+        try:
+            config = configparser.RawConfigParser()
+            config.read('/etc/sssd/sssd.conf')
+        except configparser.Error:
+            # SSSD is not configured properly. Nothing to do.
+            self.log.warn('SSSD configuration unreadable.')
+            return
+
+        self.produce(SSSDFactsLibrary(config).process())
--- a/repos/system_upgrade/el7toel8/actors/sssdfacts/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/sssdfacts/libraries/library.py
@@ -1,0 +1,105 @@
+from leapp.models import SSSDConfig, SSSDDomainConfig
+
+
+class SSSDFactsLibrary(object):
+    """
+    Helper library from SSSDFacts actor to allow unit testing.
+    """
+    def __init__(self, config):
+        self.config = config
+
+    def process(self):
+        """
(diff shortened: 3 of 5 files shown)


repos/system_upgrade/el7toel8/actors/sssdcheck/actor.py
```diff
@@ -0,0 +1,98 @@
...
+    
+    name = 'sssd_check'
+    consumes = ()
+    produces = (Report,)
+    tags = (IPUWorkflowTag, ChecksPhaseTag)
+    
+    def process(self):
+        try:
+            self.config = ConfigParser.RawConfigParser()
+            self.config.read('/etc/sssd/sssd.conf')
```
pirat89: it is forbidden to access the system in this phase.

repos/system_upgrade/el7toel8/actors/sssdcheck/actor.py
```diff
@@ -0,0 +1,98 @@
...
+        if not self.config.has_option(section, 'ldap_sudo_include_regexp'):
+            return
+        
+        report_generic(
+            title='SSSD Domain "%s": sudo rules containing wildcards ' 
+                  'will stop working.' % domain,
+            summary='Default value of ldap_sudo_include_regexp changed '
+                    'from true to false for performance reason. If you use '
+                    'sudo rules with wildcards, set this option to true '
+                    'explicitly.'
```
pirat89: I am thinking whether would be better to use `report_with_remediation` function instead of this. Where the content would be modified (or split) between summary and remedaiation remediation like: remediation="If you use sudo...true explicitly".

Do you plan or have you thought about possibility to fix it automatically in future? Would you detect such rules? In case it is not possible (or it is hard) to detect such rules, maybe you would just think later about use of user dialogs, to ask user what should happened and by default set the option to true (not implemented yet, in case you are curious, make sense to create comment line like `# TODO: ANSWER_FILE` . Actually, I am curious why we do not set the value to true now explicitly? How big issue it can be for user in case they use sudo rules with wildcards? Would be the upgrade fatal for them in such cases? If so, I suggest to create even an actor with the `ApplicationsPhaseTag`  tag that will set the value to `true` explicitly and document that planned change.

repos/system_upgrade/el7toel8/actors/sssdcheck/actor.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+        
+        # Produce report
+
+    def get_domain_section(self, domain):
+        return 'domain/' + domain
+
+    def get_provider(self, domain, provider, fallbacks=[]):
```
pirat89: Please, you rather `list()` here.

```suggestion
    def get_provider(self, domain, provider, fallbacks=list()):
```

repos/system_upgrade/el7toel8/actors/sssdcheck/actor.py
pbrezina: The original plan was to ask the administrator and act upon his/her choice. But unfortunately, I found a bug in SSSDConfig python API (which manipulates SSSD configuration) and this bug prevented me from using it to write the sudo option. I can not tell when this bug will be solved.

The one library I found for manipulating ini files removes comments from the config which is unacceptable for SSSD. Because ldap_sudo_include_regexp is rarely if ever used corner case, we decided that just reporting it is sufficient so I did not work on any custom parser.

pbrezina:
Thank you for review, I made required changes. I still need to add tests.

pbrezina:
I started looking into writing unit tests and I have a question about unit testing the actor itself.

Why is it not possible to unit test actors? I can imagine that `current_actor_context` would get the ability to return the actor's instance in addition to `run()` method. Then you can unit test all actor's methods and even `process()` if it is written in a smart way. So what is the reason why this has not been implemented?

vinzenz:
@pbrezina current_actor_context does not return the instance of an actor it merely allows to run the actor. Actors are run in a child process for isolation so they cannot interfere with the state of the application that runs the actors.
This has been a design decision to make the execution more robust, this way we can control much better the survival of the application even if an actor process crashes.

pbrezina:
Thank you for explanation. This makes sense for production, I am not yet convinced that it is necessary for testing and it seems to be quite strange. But I do not know the internals, so let's go ahead with the PR.

I updated the patches. `SSSDFacts` actor functionality was moved to `SSSDFactsLibrary` and this is covered by unit tests. The actor it self just produces whatever the library `SSSDFactsLibrary.process` returns.

I also updated `SSSDCheck` actor (I did not commit the changes in previous iteration). The actor is quite simple and moving the process function to a standalone library would just duplicate the code -- it is just `if X: report X; if Y: report Y`. Do you require any test for it?

I wanted to write component test for it but it fails even for the simplest case:
```python
from leapp.models import SSSDConfig, SSSDDomainConfig

def test_actor_execution(current_actor_context):
    current_actor_context.feed(SSSDConfig(domains=[]))
    current_actor_context.run()
```

```python
ERROR at setup of test_actor_execution 
pytest.fixture functions cannot use ``yield``. Instead write and return an inner function/generator and let the consumer call and iterate over it.:

    @pytest.fixture(scope='module')
    def loaded_leapp_repository(request):
        """
        This fixture will ensure that the repository for the current test run is loaded with all its links etc.
    
        This enables running actors and using models, tags, topics, workflows etc.
    
        Additionally loaded_leapp_repository gives you access to a :py:class:`leapp.repository.manager.RepositoryManager`
        instance.
    
        :Example:
    
        .. code-block:: python
    
            from leapp.snactor.fixture import loaded_leapp_repository
            from leapp.models import ExampleModel, ProcessedExampleModel
    
            def my_repository_library_test(loaded_leapp_repository):
                from leapp.libraries.common import global
                e = ExampleModel(value='Some string')
                result = global.process_function(e)
                assert type(result) is ProcessedExampleModel
    
        """
        repository_path = find_repository_basedir(request.module.__file__)
        os.environ['LEAPP_CONFIG'] = os.path.join(repository_path, '.leapp', 'leapp.conf')
        os.environ['LEAPP_HOSTNAME'] = socket.getfqdn()
        context = str(uuid.uuid4())
        with get_connection(None):
            Execution(context=context, kind='snactor-test-run', configuration='').store()
            os.environ["LEAPP_EXECUTION_ID"] = context
    
            manager = getattr(request.session, 'leapp_repository', None)
            if not manager:
                manager = find_and_scan_repositories(repository_path, include_locals=True)
                manager.load(resolve=True)
            yield manager
/usr/lib/python2.7/site-packages/leapp/snactor/fixture.py:126
```

vojtechsokol:
leapp-ci build

pbrezina:
Travis CI now pass.

repos/system_upgrade/el7toel8/actors/sssdfacts/actor.py
```diff
@@ -0,0 +1,33 @@
+import ConfigParser
```
vinzenz:
```suggestion
from six.moves import ConfigParser
```

repos/system_upgrade/el7toel8/actors/sssdfacts/actor.py
```diff
@@ -0,0 +1,33 @@
...
+    '''
+
+    name = 'sssd_facts'
+    consumes = ()
+    produces = (SSSDConfig,)
+    tags = (IPUWorkflowTag, FactsPhaseTag)
+
+    def process(self):
+        try:
+            config = ConfigParser.RawConfigParser()
```
vinzenz:
```suggestion
            config = configparser.RawConfigParser()
```

repos/system_upgrade/el7toel8/actors/sssdfacts/actor.py
```diff
@@ -0,0 +1,33 @@
...
+            config.read('/etc/sssd/sssd.conf')
+        except ConfigParser.Error:
```
vinzenz:
```suggestion
        except configparser.Error:
            self.log.warn('SSSD configuration unreadable - Falling back to empty configuration.')
```

repos/system_upgrade/el7toel8/actors/sssdfacts/actor.py
```diff
@@ -0,0 +1,33 @@
...
+            # SSSD is not configured properly. Nothing to do.
+            config = ConfigParser.RawConfigParser()
```
vinzenz:
```suggestion
            config = configparser.RawConfigParser()
```

repos/system_upgrade/el7toel8/actors/sssdfacts/actor.py
```diff
@@ -0,0 +1,33 @@
...
+
+        self.produce(SSSDFactsLibrary(config).process())
```
vinzenz: Does it actually make sense at all to produce this message in the end if you fail to read the SSSD config?

repos/system_upgrade/el7toel8/actors/sssdcheck/actor.py
```diff
@@ -0,0 +1,62 @@
...
+    - ldap_sudo_include_regexp changed default from true to false
+    '''
+
+    name = 'sssd_check'
+    consumes = (SSSDConfig,)
+    produces = (Report,)
+    tags = (IPUWorkflowTag, ChecksPhaseTag)
+
+    def process(self):
+        model = next(self.consume(SSSDConfig))
```
vinzenz: What, if there's no message SSSDConfig?
```suggestion
        model = next(self.consume(SSSDConfig), SSSDConfig(domains=[]))
```

repos/system_upgrade/el7toel8/actors/sssdfacts/actor.py
pbrezina: If this case is handled correctly in SSSDCheck actor then no. Fixed.

repos/system_upgrade/el7toel8/actors/sssdcheck/actor.py
pbrezina: Thank you. I used the latter one.

PR #175, opened 2019-04-26
selinux: fix getting facts

pirat89:
selinux: fix getting facts
    
    The selinux.selinux_getenforcemode function returns list
      [ret, int] - where int is -1 when disabled in config file
                                 0 when permissive
                                 1 when enforcing
                 - ret value is return value of underneath function(s).
    Originally the actor expected just int instead of list. Take just
    the second value now.
    
    FIXME: Ignoring for now the first value


Rezney:
Please add commentary:

# selinux_getenforcemode returns a list where 1st value is return code and 2nd value is status

Rezney:
Tested. SElinux is now permissive as expected and FirstBoot phase started again.

PR #182, opened 2019-04-27
CUPS-filters migration actors

zdohnal:
Hi!

I created two actors for cups-filters migration - cupsfilterscheck and cupsfiltersmigrate. The code is currently 'working prototype', tested only manually for now.
I will move all 'non-actor' methods to outside module and create unit tests after my PTO, would you mind reviewing the code in meanwhile?

Thank you in advance!

Zdenek

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/actor.py
@@ -1,0 +1,29 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import library
+from leapp.models import InstalledRedHatSignedRPM
+from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
+
+
+class CupsfiltersMigrate(Actor):
+    """
+    Actor for migrating package cups-filters.
+
+    Migrating cups-filters package means adding two directives into
+    /etc/cups/cups-browsed.conf - LocalQueueNamingRemoteCUPS and
+    CreateIPPPrinterQueues.
+
+    LocalQueueNamingRemoteCUPS directive indicates what will be used as a name
+    for local print queue creation - the default is DNS-SD ID of advertised
+    print queue now, it was the name of remote print queue in the past.
+
+    CreateIPPPrinterQueues directive serves for telling cups-browsed to create
+    local print queues for all available IPP printers.
+    """
+
+    name = 'cupsfilters_migrate'
+    consumes = (InstalledRedHatSignedRPM,)
+    produces = ()
+    tags = (ApplicationsPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        library.update_cups_browsed()
--- a/repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/libraries/library.py
@@ -1,0 +1,116 @@
+from leapp.libraries.common.rpms import has_package
+from leapp.libraries.stdlib import api
+from leapp.models import InstalledRedHatSignedRPM
+
+# rpm : the default config file
+BROWSED_CONFIG = '/etc/cups/cups-browsed.conf'
+
+
+# The list of macros that should be set to get the behavior
+# from previous RHEL
+NEW_MACROS = [
+    ('LocalQueueNamingRemoteCUPS', 'RemoteName'),
+    ('CreateIPPPrinterQueues', 'All')
+]
+
+
+def _macro_exists(path, macro):
+    """
+    Check if macro is in the file.
+
+    :param str path: string representing the full path of the config file
+    :param str macro: new directive to be added
+    :return boolean res: macro does/does not exist in the file
+    """
+    with open(path, 'r') as f:
+        lines = f.readlines()
+
+    for line in lines:
+        if line.lstrip().startswith(macro):
+            return True
+    return False
+
+
+def _append_string(path, content):
+    """
+    Append string at the end of file.
+
+    :param str path: string representing the full path of file
+    :param str content: preformatted string to be added
+    """
+    with open(path, 'a') as f:
+        f.write(content)
+
+
+def update_config(path, check_function=_macro_exists,
+                  append_function=_append_string):
+    """
+    Insert expected content into the file on the path if it is not
+    in the file already.
+
+    :param str path: string representing the full path of the config file
+    :param func check_function: function to be used to check if string is in the file
+    :param func append_function: function to be used to append string
+    """
+
+    macros = []
+    for macro in NEW_MACROS:
+        if not check_function(path, macro[0]):
+            macros.append(' '.join(macro))
+
+    if not macros:
+        return
+
+    fmt_input = "\n{comment_line}\n{content}\n".format(comment_line='# content added by Leapp',
+                                                       content='\n'.join(macros))
+
+    try:
+        append_function(path, fmt_input)
+    except IOError:
+        raise IOError('Error during writing to file: {}.'.format(path))
+
+
+def _check_package(pkg):
+    """
+    Checks if a package is installed and signed
+
+    :param str pkg: name of package
+    """
+    return has_package(InstalledRedHatSignedRPM, pkg)
+
+
+def update_cups_browsed(debug_log=api.current_logger().debug,
+                        error_log=api.current_logger().error,
+                        is_installed=_check_package,
+                        append_function=_append_string,
+                        check_function=_macro_exists):
+    """
+    Update cups-browsed configuration file
+
+    :param func debug_log: function for debug logging
+    :param func error_log: function for error logging
+    :param func is_installed: checks if the package is installed
+    :param func append_function: appends string into file
+    :param func check_function: checks if macro is in the file
+    """
+
+    error_list = []
+
+    if not is_installed('cups-filters'):
+        return
+
+    debug_log('Updating cups-browsed configuration file {}.'
+              .format(BROWSED_CONFIG))
+
+    try:
+        update_config(BROWSED_CONFIG,
+                      check_function,
+                      append_function)
+    except (OSError, IOError) as error:
+        error_list.append((BROWSED_CONFIG, error))
+    if error_list:
+        error_log('The files below have not been modified '
+                  '(error message included):' +
+                  ''.join(['\n    - {}: {}'.format(err[0], err[1])
+                          for err in error_list]))
+        return
--- a/repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/tests/test_update_config.py
+++ b/repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/tests/test_update_config.py
@@ -1,0 +1,101 @@
+import pytest
+
+from leapp.libraries.actor.library import NEW_MACROS
+from leapp.libraries.actor.library import update_config
+
+
+def _gen_append_str(list_out=None):
+    """
+    Just helper function to generate string expected to be added for an input (see testdata) for testing.
+
+    :param list list_out: None, [0], [1], [0,1] - no more expected vals,
+                          which represents what macros should be appended
+                          in output
+    """
+    if not list_out:
+        return ''
+    _out_list = ('LocalQueueNamingRemoteCUPS RemoteName', 'CreateIPPPrinterQueues All')
+    output = ['# content added by Leapp']
+    for i in list_out:
+        output.append(_out_list[i])
+    # ensure the extra NL is before the string and the empty NL is in the end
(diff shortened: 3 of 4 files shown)


repos/system_upgrade/el7toel8/actors/cupsfilterscheck/actor.py
```diff
@@ -0,0 +1,49 @@
+from leapp.actors import Actor
+from leapp.libraries.common.reporting import report_generic
+from leapp.models import Report, InstalledRedHatSignedRPM, CupsFiltersModel
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+from subprocess import Popen, PIPE
```
artmello: I have the feeling that this import is not being used any more. Just for information, Leapp stdlib provides a run() that can be used to execute commands from the system., so subprocess is usually not necessary.

repos/system_upgrade/el7toel8/actors/cupsfilterscheck/actor.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+
+    name = 'cupsfilters_check'
+    consumes = (InstalledRedHatSignedRPM,)
+    produces = (Report, CupsFiltersModel)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+      installed_cupsfilters = False
+      migrateable = False
```
artmello: If I understood correctly both variables, installed_cupsfilters and migrateable, have always the same values. Either both True or both False. I think we could simplify this by only using one of them.

repos/system_upgrade/el7toel8/actors/cupsfilterscheck/actor.py
```diff
@@ -0,0 +1,49 @@
...
+        for pkg in rpm_pkgs.items:
+          if pkg.name == "cups-filters":
+            installed_cupsfilters = True
+
+      if installed_cupsfilters is False:
+        report_generic(
+                        title='cups-filters will not be migrated',
+                        summary='cups-filters is not installed',
+                        severity='low'
+                      )
```
artmello: There is nothing wrong codewise with the call to report_generic. But, reports will be read by sys admins/developers after the upgrade process. I, personally, think we should keep it as concisely as possible to avoid readers skipping it and missing valuable information. Since there is no action that user should take if cups-filters is not installed, what do you think we change this for a debug message or even remove it at all?

repos/system_upgrade/el7toel8/actors/cupsfilterscheck/actor.py
```diff
@@ -0,0 +1,49 @@
...
+        # Always migrateable when installed, because post install script
+        # modifies the configuration file
+        migrateable = True
+
+      if migrateable is True:
+        report_generic(
+                        title='Package cups-filters will be migrated',
+                        summary='The package is installed and configuration file is modified by user.',
+                        severity='low'
+                      )
```
artmello: ditto as last message about report_generic vs log.

repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/actor.py
```diff
@@ -0,0 +1,86 @@
...
+config = 'cups-browsed.conf'
+
+"""
+Directives to be added
+"""
+
+directives = [
+              'LocalQueueNamingRemoteCUPS RemoteName', 
+              'CreateIPPPrinterQueues All'
+             ]
```
artmello: What do you think if we move CONFDIR, config and directives to inside the functions they are used. They seem to be used one single time and being close to where they are used helps on debuging, imo.

repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/actor.py
```diff
@@ -0,0 +1,86 @@
...
+      """
+      insert_strings_into_file
+
+      parameters:
+      - stringed_list - list of strings
+      - filename - string
+
+      returns boolean
+      """
+      path = CONFDIR + filename
```
artmello: Could we check if the file path exists? I am aware this should only be executed if the package was found in the system, but I think we are better safe than sorry here.

repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/actor.py
```diff
@@ -0,0 +1,86 @@
...
+      with open(path, 'a') as opened_conf_file:
+        opened_conf_file.write('\n')
+        for string in stringed_list:
+          opened_conf_file.write(string)
+          opened_conf_file.write('\n')
+        return True
+      return False
```
artmello: I am afraid this return will never be reach. If you are able to open the config file, the processing will happen and you will return True. If the open fails, an exception will be raised and since it is not treated the process will stop. Probably checking if the file is writable before and returning False otherwise can have the desired behaviour.

repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/actor.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def process(self):
+      for i in self.consume(CupsFiltersModel):
+        migrateable = i.migrateable
+
+      done = False
+
+      if migrateable is True:
```
artmello: Not necessary to use "is True". "if migrateable:" is enough to check the condition.

repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/actor.py
```diff
@@ -0,0 +1,86 @@
...
+        done = self.insert_strings_into_file(directives, config)
+        if done is True:
```
artmello: ditto for unnecessary "is True".

repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/actor.py
```diff
@@ -0,0 +1,86 @@
...
+          report_generic(
+                          title='Package cups-filters was migrated',
+                          summary='Directives were succesfully added.',
```
artmello: Could we describe more what directives were added and why? Should a sysadmin be aware of some new behaviour on the system with the new directives enabled?

repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/actor.py
```diff
@@ -0,0 +1,86 @@
...
+                          severity='medium'
+                        )
+        else:
+          report_generic(
+                          title='Package cups-filters could not be migrated',
+                          summary='Directives could not be add due error.',
+                          severity='medium'
+                        )
```
artmello: What kind of problems the system will have if the directives were not added? There is something that sysadmin can do? Should we continue the upgrade process without such directives correctly applied?

repos/system_upgrade/el7toel8/models/cupsfiltersmodel.py
```diff
@@ -0,0 +1,7 @@
+from leapp.models import Model, fields
+from leapp.topics import SystemInfoTopic
+
+
+class CupsFiltersModel(Model):
```
artmello: Could we have a docstring documentation to this model describing its purpose?

zdohnal:
Hi everyone,

I'm still working on the actor for now, I'll let you know here in the ticket when it will be ready for new review.

repos/system_upgrade/el7toel8/actors/cupsfilterscheck/actor.py
vinzenz:
```suggestion
```

repos/system_upgrade/el7toel8/actors/cupsfilterscheck/actor.py
vinzenz: Please remove this report, as it is not useful to anyone. If you want, feel free to log it via `self.log.info`

repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/actor.py
```diff
@@ -0,0 +1,86 @@
+from leapp.actors import Actor
+from leapp.libraries.common.reporting import report_generic
+from leapp.models import Report, CupsFiltersModel
+from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
+
+"""
+cups-browsed configuration directory
+"""
+
```
vinzenz: Doc strings for variables should follow the declaration not lead

repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/actor.py
```diff
@@ -0,0 +1,86 @@
...
+CONFDIR = '/etc/cups/'
+
+"""
+Configuration file name
+"""
+
+config = 'cups-browsed.conf'
```
vinzenz: globals should be uppercase

repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/actor.py
vinzenz:
```suggestion
      path = os.path.join(CONFDIR, filename)
```

repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/actor.py
```diff
@@ -0,0 +1,86 @@
...
+      insert_strings_into_file
+
+      parameters:
+      - stringed_list - list of strings
+      - filename - string
+
+      returns boolean
+      """
+      path = CONFDIR + filename
+      with open(path, 'a') as opened_conf_file:
```
vinzenz:
```python
        opened_conf_file.write('\n{}\n'.format('\n'.join(stringed_list))
```
has the same effect as the whole code below

vinzenz:
@zdohnal Also rebase the PR then, because there are now better CI checks in-place which give you faster results

zdohnal:
Ready for review.

zdohnal:
All previous notes are outdated, closing them.

zdohnal:
Ready for review.

pirat89:
@zdohnal putting on the top of my todolist for tomorrow.

zdohnal:
Ready for review.

pirat89:
in my todo for today.

repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/libraries/library.py
```diff
@@ -0,0 +1,115 @@
...
+from leapp.models import InstalledRedHatSignedRPM
+
+# rpm : the default config file
+BROWSED_CONFIG = '/etc/cups/cups-browsed.conf'
+
+
+# list of macros that should be set
+NEW_MACROS = [
+    'LocalQueueNamingRemoteCUPS RemoteName',
+    'CreateIPPPrinterQueues All'
```
pirat89: As discussed, it should not be issue as those macros are not on RHEL 7 system. However it will be better to check really just macros in such case and not string 'Macro Value'. So the list could be defined e.g. like that:
```diff
- NEW_MACROS = [
-    'LocalQueueNamingRemoteCUPS RemoteName',
-    'CreateIPPPrinterQueues All'
+ NEW_MACROS = [
+    ('LocalQueueNamingRemoteCUPS', 'RemoteName'),
+    ('CreateIPPPrinterQueues', 'All')
```

@zdohnal  Let's sync on monday.

zdohnal: Just to be precise - the strings need to be checked in the case when configuration file is not modified by user before, so dnf transaction went well during upgrade and there is no .rpmnew file.

zdohnal:
Rebased to current master.

repos/system_upgrade/el7toel8/actors/cupsfiltersmigrate/tests/test_update_config.py
```diff
@@ -0,0 +1,64 @@
...
+        if not f.exists(path, macro):
+            macros.append(macro)
+
+    fmt_input = ''
+    if macros:
+        fmt_input = "\n{comment_line}\n{content}\n".format(comment_line='# content added by Leapp',
+                                                           content='\n'.join(macros))
+    update_config(path, f.exists, f.append)
+
+    assert f.content == content + fmt_input
```
pirat89: as discussed, the test doesn't proove functionality of the tested function, as it is used the same code to prepare expected output as the tested function.

zdohnal:
Commited changes without testing and squashing commits.

zdohnal:
Squashed and tested, ready for review.

pirat89:
LGTM. Regarding the previous testing before the latest changes and covering by unit tests, I believe this is working as expected.

PR #183, opened 2019-04-29
Add powertop actor

yarda:


repos/system_upgrade/el7toel8/actors/powertop/actor.py
```diff
@@ -0,0 +1,32 @@
...
+
+    name = 'powertop'
+    consumes = (InstalledRedHatSignedRPM,)
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        for fact in self.consume(InstalledRedHatSignedRPM):
+            for rpm in fact.items:
+                print(rpm)
```
artmello: Is this "print" really necessary? If so, could we use a debug message from our log system?

yarda: Ah, sorry, I forgot it there :) It was used for debugging.

PR #186, opened 2019-05-02
Add tcp_wrappers and vsftpd actors

olysonek:
This pull request adds a couple of actors and some helpers for testing.

The `TcpWrappersConfigRead` actor reads the tcp_wrappers configuration (/etc/hosts.{allow,deny}) and stores some data about it in the `TcpWrappersFacts` model. An instance of that model can then be passed to the `config_applies_to_daemon` library function which can be used in other actors to determine, if the tcp_wrappers configuration applies to the specified daemon.

The `VsftpdConfigRead` actor gathers some data about the vsftpd configuration (but only if vsftpd is installed) and  stores it in the `VsftpdFacts` model.

The `VsftpdConfigCheck` actor consumes `TcpWrappersFacts` and `VsftpdFacts`. The actor determines, whether the vsftpd configuration is supported in RHEL-8. If it is not, it inhibits the upgrade. Namely, it checks whether vsftpd was using tcp_wrappers, which is not supported in RHEL-8.

The `VsftpdConfigUpdate` actor consumes `VsftpdFacts` and modifies the vsftpd config files so that they still work in RHEL-8, and so that the effective configuration is as close to the original as possible. See the comit message of the last commit for more details.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/tcpwrappersconfigread/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/tcpwrappersconfigread/libraries/library.py
@@ -1,0 +1,54 @@
+import errno
+import re
+
+from leapp.libraries.stdlib import api
+from leapp.models import DaemonList, TcpWrappersFacts
+
+CONFIG_FILES = ['/etc/hosts.allow', '/etc/hosts.deny']
+
+
+def _read_file(path):
+    with open(path, 'r') as f:
+        return f.read()
+
+
+def _get_lines(content):
+    content = re.sub(r'\\\n', '', content)
+    return content.split('\n')
+
+
+def _is_comment(line):
+    return len(line) == 0 or line.isspace() or line.startswith('#')
+
+
+def _get_daemon_list_in_line(line):
+    daemon_list = line.split(':', 1)[0]
+    daemon_list = re.split(',| |\t', daemon_list)
+    daemon_list = [word for word in daemon_list if len(word) > 0]
+    return DaemonList(value=daemon_list)
+
+
+def _get_daemon_lists_in_file(path, read_func=_read_file):
+    ret = []
+    try:
+        content = read_func(path)
+    except IOError as e:
+        if e.errno != errno.ENOENT:
+            api.current_logger().warning('Failed to read %s: %s' % (path, e))
+        return ret
+    lines = [line for line in _get_lines(content) if not _is_comment(line)]
+    for line in lines:
+        ret.append(_get_daemon_list_in_line(line))
+    return ret
+
+
+def _get_daemon_lists(read_func=_read_file):
+    daemon_lists = []
+    for path in CONFIG_FILES:
+        daemon_lists.extend(_get_daemon_lists_in_file(path, read_func=read_func))
+    return daemon_lists
+
+
+def get_tcp_wrappers_facts(read_func=_read_file):
+    daemon_lists = _get_daemon_lists(read_func=read_func)
+    return TcpWrappersFacts(daemon_lists=daemon_lists)
--- a/repos/system_upgrade/el7toel8/actors/vsftpdconfigcheck/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/vsftpdconfigcheck/libraries/library.py
@@ -1,0 +1,20 @@
+from leapp.libraries.common.reporting import report_with_links
+from leapp.libraries.common.tcpwrappersutils import config_applies_to_daemon
+
+
+def check_config_supported(tcpwrap_facts, vsftpd_facts):
+    bad_configs = [config.path for config in vsftpd_facts.configs if config.tcp_wrappers]
+    if len(bad_configs) > 0 and config_applies_to_daemon(tcpwrap_facts, 'vsftpd'):
+        list_separator_fmt = '\n    - '
+        report_with_links(title='Unsupported vsftpd configuration',
+                          summary=('tcp_wrappers support has been removed in RHEL-8. '
+                                   'Some configuration files set the tcp_wrappers option to true and '
+                                   'there is some vsftpd-related configuration in /etc/hosts.deny '
+                                   'or /etc/hosts.allow. Please migrate it manually. '
+                                   'The list of problematic configuration files:{}{}'
+                                  ).format(list_separator_fmt,
+                                           list_separator_fmt.join(bad_configs)),
+                          links=[{'title': 'Replacing TCP Wrappers in RHEL 8',
+                                  'href': 'https://access.redhat.com/solutions/3906701'}],
+                          severity='high',
+                          flags=['inhibitor'])
--- a/repos/system_upgrade/el7toel8/actors/vsftpdconfigcheck/tests/test_library.py
+++ b/repos/system_upgrade/el7toel8/actors/vsftpdconfigcheck/tests/test_library.py
@@ -1,0 +1,81 @@
+from leapp.models import DaemonList, TcpWrappersFacts, VsftpdConfig, VsftpdFacts
+from leapp.reporting import Report
+from leapp.snactor.fixture import current_actor_context
+
+
+def test_actor_with_unsupported_tcpwrap_and_vsftpd_config(current_actor_context):
+    config1 = VsftpdConfig(path='/etc/vsftpd/foo.conf', tcp_wrappers=False)
+    config2 = VsftpdConfig(path='/etc/vsftpd/bar.conf', tcp_wrappers=True)
+    vsftpd_facts = VsftpdFacts(configs=[config1, config2])
+    daemon_list = DaemonList(value=['vsftpd'])
+    tcpwrap_facts = TcpWrappersFacts(daemon_lists=[daemon_list])
+
+    current_actor_context.feed(vsftpd_facts)
+    current_actor_context.feed(tcpwrap_facts)
+    current_actor_context.run()
+    report = current_actor_context.consume(Report)[0]
+
+    assert 'inhibitor' in report.flags
+    assert 'foo.conf' not in report.detail['summary']
+    assert 'bar.conf' in report.detail['summary']
+
+
+def test_actor_with_unsupported_tcpwrap_multiple_unsupported_vsftpd_configs(current_actor_context):
+    config1 = VsftpdConfig(path='/etc/vsftpd/foo.conf', tcp_wrappers=True)
+    config2 = VsftpdConfig(path='/etc/vsftpd/bar.conf', tcp_wrappers=False)
+    config3 = VsftpdConfig(path='/etc/vsftpd/goo.conf', tcp_wrappers=True)
+    vsftpd_facts = VsftpdFacts(configs=[config1, config2, config3])
+    daemon_list = DaemonList(value=['vsftpd'])
+    tcpwrap_facts = TcpWrappersFacts(daemon_lists=[daemon_list])
+
+    current_actor_context.feed(vsftpd_facts)
+    current_actor_context.feed(tcpwrap_facts)
+    current_actor_context.run()
+    report = current_actor_context.consume(Report)[0]
+
+    assert 'inhibitor' in report.flags
+    assert 'foo.conf' in report.detail['summary']
+    assert 'bar.conf' not in report.detail['summary']
+    assert 'goo.conf' in report.detail['summary']
+
+
+def test_actor_with_unsupported_tcpwrap_config(current_actor_context):
+    config1 = VsftpdConfig(path='/etc/vsftpd/foo.conf', tcp_wrappers=False)
+    config2 = VsftpdConfig(path='/etc/vsftpd/bar.conf', tcp_wrappers=None)
+    vsftpd_facts = VsftpdFacts(configs=[config1, config2])
+    daemon_list = DaemonList(value=['vsftpd'])
+    tcpwrap_facts = TcpWrappersFacts(daemon_lists=[daemon_list])
+
+    current_actor_context.feed(vsftpd_facts)
+    current_actor_context.feed(tcpwrap_facts)
+    current_actor_context.run()
+
+    assert not current_actor_context.consume(Report)
+
+
+def test_actor_with_unsupported_vsftpd_config(current_actor_context):
+    config1 = VsftpdConfig(path='/etc/vsftpd/foo.conf', tcp_wrappers=False)
+    config2 = VsftpdConfig(path='/etc/vsftpd/bar.conf', tcp_wrappers=True)
+    vsftpd_facts = VsftpdFacts(configs=[config1, config2])
+    daemon_list = DaemonList(value=['all', 'except', 'vsftpd'])
+    tcpwrap_facts = TcpWrappersFacts(daemon_lists=[daemon_list])
+
+    current_actor_context.feed(vsftpd_facts)
+    current_actor_context.feed(tcpwrap_facts)
+    current_actor_context.run()
+
+    assert not current_actor_context.consume(Report)
+
+
+def test_actor_with_supported_tcpwrap_and_vsftpd_config(current_actor_context):
+    config1 = VsftpdConfig(path='/etc/vsftpd/foo.conf', tcp_wrappers=False)
+    config2 = VsftpdConfig(path='/etc/vsftpd/bar.conf', tcp_wrappers=False)
+    vsftpd_facts = VsftpdFacts(configs=[config1, config2])
+    daemon_list = DaemonList(value=['all', 'except', 'vsftpd'])
+    tcpwrap_facts = TcpWrappersFacts(daemon_lists=[daemon_list])
+
+    current_actor_context.feed(vsftpd_facts)
+    current_actor_context.feed(tcpwrap_facts)
+    current_actor_context.run()
+
+    assert not current_actor_context.consume(Report)
--- a/repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/library.py
@@ -1,0 +1,58 @@
+import errno
+import os
+
+from leapp.libraries.actor import config_parser
+from leapp.libraries.common.vsftpdutils import get_config_contents, get_default_config_hash, \
+                                               read_file, \
+                                               VSFTPD_CONFIG_DIR, \
+                                               STRICT_SSL_READ_EOF, TCP_WRAPPERS
+from leapp.libraries.stdlib import api
+from leapp.models import VsftpdConfig, VsftpdFacts
+
+
+def _parse_config(path, content):
+    try:
+        parser = config_parser.VsftpdConfigParser(content)
+        return parser.parsed_config
+    except config_parser.ParsingError:
+        api.current_logger().info('File %s does not look like vsftpd configuration, skipping.'
+                                  % path)
+        return None
+
+
+def _get_parsed_configs(read_func=read_file, listdir=os.listdir):
+    res = []
+    try:
+        for fname in listdir(VSFTPD_CONFIG_DIR):
+            path = os.path.join(VSFTPD_CONFIG_DIR, fname)
+            if not path.endswith('.conf'):
+                continue
+            content = get_config_contents(path, read_func=read_func)
+            if content is None:
+                continue
+            parsed = _parse_config(path, content)
+            if parsed is not None:
+                res.append((path, parsed))
+    except OSError as e:
+        if e.errno != errno.ENOENT:
+            api.current_logger().warning('Failed to read vsftpd configuration directory: %s'
+                                         % e)
+    return res
+
+
+def get_vsftpd_facts(read_func=read_file, listdir=os.listdir):
+    config_hash = get_default_config_hash(read_func=read_func)
+    configs = _get_parsed_configs(read_func=read_func, listdir=listdir)
+    res_configs = []
+    for path, config in configs:
+        res_configs.append(VsftpdConfig(path=path,
+                                        strict_ssl_read_eof=config.get(STRICT_SSL_READ_EOF),
+                                        tcp_wrappers=config.get(TCP_WRAPPERS)))
+    return VsftpdFacts(default_config_hash=config_hash, configs=res_configs)
+
+
+def is_processable(installed_rpm_facts):
+    for pkg in installed_rpm_facts.items:
+        if pkg.name == 'vsftpd':
+            return True
+    return False
--- a/repos/system_upgrade/el7toel8/actors/vsftpdconfigupdate/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/vsftpdconfigupdate/actor.py
@@ -1,0 +1,33 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import library
+from leapp.models import VsftpdFacts
+from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
+
+
+class VsftpdConfigUpdate(Actor):
+    """
+    Modifies vsftpd configuration files on the target RHEL-8 system so that the effective
+    configuration is the same, where possible. This means doing two things:
+    1. Reverting the default configuration file (/etc/vsftpd/vsftpd.conf) to its state
+       before the upgrade (where it makes sense), if the configuration file was being used
+       with its default content (i.e., unmodified) on the source system (the configuration
+       file gets replaced with a new version during the RPM upgrade in this case).
+       The anonymous_enable option falls in this category.
+    2. Adding 'option=old_effective_value' to configuration files for options whose default
+       value has changed, if the option is not explicitly specified in the configuration file.
+       The strict_ssl_read_eof option falls in this category.
+    3. Disabling options that cannot be enabled, otherwise vsftpd wouldn't work.
+       The tcp_wrappers option falls in this category.
+    """
+
+    name = 'vsftpd_config_update'
+    consumes = (VsftpdFacts,)
+    produces = ()
+    tags = (ApplicationsPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        try:
+            vsftpd_facts = next(self.consume(VsftpdFacts))
+        except StopIteration:
+            return
+        library.migrate_configs(vsftpd_facts)
--- a/repos/system_upgrade/el7toel8/libraries/tcpwrappersutils.py
+++ b/repos/system_upgrade/el7toel8/libraries/tcpwrappersutils.py
@@ -1,0 +1,83 @@
+import re
+
+
+def _build_regex(pattern):
+    regex = '^'
+    part_beginning = 0
+    while part_beginning < len(pattern):
+        ix1 = pattern.find('*', part_beginning)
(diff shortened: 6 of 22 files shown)


repos/system_upgrade/el7toel8/libraries/tests/test_vsftpdutils.py
```diff
@@ -0,0 +1,82 @@
...
+
+    res = get_config_contents(path, read_func=f.read_file)
+
+    assert not f.error
+    assert res is None
+
+
+def test_getting_inaccessible_config_gives_None():
+    path = 'my_file'
+    f = MockFile(path, to_raise=IOError(errno.EACCES))
```
fernflower: Nit: 
While having read_file function (leapp.libraries.common.vsftpdutils.read_file) implemented separately makes the read file functionality easily testable, having it passed as a callback into other methods like get_config_contents and get_default_config_hash imho is overcomplicating the api (honestly I can't think of a second way to read the file).
Have you considered using mock's context manager to control output of open() function instead of doing all the work yourself? 

Something like 
```
with mock.patch('builtins.open', side_effect=IOError) as open_mock:
    res = get_config_contents(path, read_func=f.read_file)
```

It's an open question, I might not have the full picture as I've definitely spent a lot less time with this patch than you had.

repos/system_upgrade/el7toel8/actors/tcpwrappersconfigread/libraries/library.py
```diff
@@ -0,0 +1,67 @@
...
+
+CONFIG_FILES = ['/etc/hosts.allow', '/etc/hosts.deny']
+
+
+def read_file(path):
+    with open(path, 'r') as f:
+        return f.read()
+
+
+def get_lines(content):
```
fernflower: Wondering if this will do the trick

```
re.sub(r'//[\S\t\n]*\n', '', content)
return content.split('\n')
```

olysonek: You mean this, right? Good thinking, I'll change it.
```
content = re.sub(r'\\\n', '', content)
return content.split('\n')
```

repos/system_upgrade/el7toel8/libraries/tcpwrappersutils.py
```diff
@@ -0,0 +1,81 @@
...
+    '''
+    Takes a TcpWrappersFacts representation of the tcp_wrappers configuration and
+    returns True if the config applies to the specified daemon. Otherwise returns False.
+
+    This information is intended to be used in the Checks phase to check whether there is
+    any tcp_wrappers configuration that the user needs to migrate manually and whether we
+    should inhibit the upgrade, so that the upgraded system is not insecure.
+    '''
+    for daemon_list in facts.daemon_lists:
+        value = daemon_list.value
```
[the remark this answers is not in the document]
olysonek: > L77-L78 can be squashed
> 
> value = [item.lower() for item in daemon_list.value]

Done.

> Nit: iiuc L76-L81 of config_applies_to_daemon are equivalent to
> 
> return any((_daemon_list_matches_daemon([item.lower() for item in daemon_list.value], daemon, 0) for daemon_list in facts.daemon_lists))

The old variant is more readable to me :).

repos/system_upgrade/el7toel8/libraries/tests/test_vsftpdutils.py
```diff
@@ -0,0 +1,82 @@
...
+        self.path = path
+        self.content = content
+        self.to_raise = to_raise
+        self.error = False
+
+    def read_file(self, path):
+        if path != self.path:
+            self.error = True
+            raise ValueError
+        if self.to_raise is None:
```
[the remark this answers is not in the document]
olysonek: OK, done.

repos/system_upgrade/el7toel8/libraries/tests/test_vsftpdutils.py
olysonek: Personally, I like the idea of injecting dependencies, which interact with the external system. It seems cleaner than making the test reliant on the way the implementation reads the files (another way to open the file would be `os.open`). But I do agree that my implementation of dependency injection is not ideal and makes it look a bit messy. I haven't found a better way so far.

Anyway, I did try mocking `builtins.open`, but couldn't get it to work. It seemed to me that apart from the function under test, something else was also using `open`, so mocking it broke that "something". But I may be wrong.

repos/system_upgrade/el7toel8/libraries/vsftpdutils.py
```diff
@@ -0,0 +1,32 @@
...
+        return f.read()
+
+
+def get_config_contents(path, read_func=read_file):
+    try:
+        return read_func(path)
+    except IOError as e:
+        if e.errno != errno.ENOENT:
+            api.current_logger().warn('Failed to read vsftpd configuration file: %s' % e)
+        return None
```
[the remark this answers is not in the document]
olysonek: I'm not sure what you mean. Can you elaborate?

repos/system_upgrade/el7toel8/libraries/vsftpdutils.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def get_default_config_hash(read_func=read_file):
+    content = get_config_contents(VSFTPD_DEFAULT_CONFIG_PATH, read_func=read_func)
+    if content is None:
```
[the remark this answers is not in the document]
olysonek: That would change the behaviour slightly, because `bool("") == False`. I can drop `None` from  `return None` though, if you like.

repos/system_upgrade/el7toel8/libraries/tests/test_vsftpdutils.py
pirat89: @olysonek: `builtins.open` is kind of problematic. I thought it is just issue when you want to mock stuff and runnning whole actor in tests. But nowadays we have the `api` library, it is possible that there is hidden use of `open` as well in background.

repos/system_upgrade/el7toel8/models/vsftpdfacts.py
```diff
@@ -0,0 +1,17 @@
+from leapp.models import Model, fields
+from leapp.topics import SystemInfoTopic
+
+
+class VsftpdConfig(Model):
+    topic = SystemInfoTopic
+
+    path = fields.String()
+    strict_ssl_read_eof = fields.Nullable(fields.Boolean())
+    tcp_wrappers = fields.Nullable(fields.Boolean())
```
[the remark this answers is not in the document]
olysonek: If `tcp_wrappers is None`, it means that the "tcp_wrappers" option was left unspecified in the configuration file, so vsftpd will use the default value for the option. We could assign the default value to the attribute, if the option is unspecified in the config file, however I think it's better the way I designed it - this way the config reader is more generic, because reading the config, and deciding what the default value is, is separated.

repos/system_upgrade/el7toel8/actors/vsftpdconfigcheck/tests/test_library.py
```diff
@@ -0,0 +1,59 @@
...
+    current_actor_context.feed(vsftpd_facts)
+    current_actor_context.feed(tcpwrap_facts)
+    current_actor_context.run()
+
+    assert 'inhibitor' in current_actor_context.consume(Report)[0].flags
+
+
+def test_actor_with_unsupported_tcpwrap_config(current_actor_context):
+    config1 = VsftpdConfig(path='/etc/vsftpd/foo.conf', tcp_wrappers=False)
+    config2 = VsftpdConfig(path='/etc/vsftpd/bar.conf', tcp_wrappers=None)
```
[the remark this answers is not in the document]
olysonek: It does, it means the option is unspecified in the configuration file, so the default value was being used. The default value in RHEL-7 is "YES", so if the option in the config file is unspecified, tcp_wrappers support is enabled. In this test case, we test that the actor produces the warning report in such a case.

repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/library.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def get_vsftpd_facts(read_func=read_file, listdir=os.listdir):
+    config_hash = get_default_config_hash(read_func=read_func)
+    configs = get_parsed_configs(read_func=read_func, listdir=listdir)
+    res_configs = []
+    for path, config in configs:
+        res_configs.append(VsftpdConfig(path=path,
+                                        strict_ssl_read_eof=config.get(STRICT_SSL_READ_EOF),
+                                        tcp_wrappers=config.get(TCP_WRAPPERS)))
```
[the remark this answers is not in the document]
olysonek: We could do that, but the default value would have to be `True`. However, the actor is more generic, if we put the code that decides what the default is, in the config check and config update actors.

repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/library.py
```diff
@@ -0,0 +1,45 @@
...
+    return res
+
+
+def get_vsftpd_facts(read_func=read_file, listdir=os.listdir):
+    config_hash = get_default_config_hash(read_func=read_func)
+    configs = get_parsed_configs(read_func=read_func, listdir=listdir)
+    res_configs = []
+    for path, config in configs:
+        res_configs.append(VsftpdConfig(path=path,
+                                        strict_ssl_read_eof=config.get(STRICT_SSL_READ_EOF),
```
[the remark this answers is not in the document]
olysonek: Yes, it was.

repos/system_upgrade/el7toel8/actors/vsftpdconfigcheck/tests/test_library.py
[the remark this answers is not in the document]
olysonek: Waaaait a minute. I think the test is wrong and there's a bug...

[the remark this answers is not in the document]
olysonek: No, it's all good. I was just wrong about what the default value is. The default is "NO".

repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/library.py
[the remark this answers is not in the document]
olysonek: > We could do that, but the default value would have to be `True`

I take that back. `False` is correct, as that is the default value of the option.

repos/system_upgrade/el7toel8/actors/vsftpdconfigcheck/libraries/library.py
```diff
@@ -0,0 +1,19 @@
...
+        if config.tcp_wrappers:
+            bad_configs.append(config.path)
+    if len(bad_configs) > 0 and config_applies_to_daemon(tcpwrap_facts, 'vsftpd'):
+        report_generic(title='Unsupported vsftpd configuration',
+                       summary=('tcp_wrappers support has been removed in RHEL-8. '
+                                'Configuration files %s set the tcp_wrappers option to true and '
+                                'there is some vsftpd-related configuration in /etc/hosts.deny or '
+                                '/etc/hosts.allow. Please migrate it manually (see '
+                                'https://access.redhat.com/solutions/3906701.')
+                               % ". ".join(bad_configs),
```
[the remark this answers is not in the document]
olysonek: Applied, with some minor modifications. Thanks!

repos/system_upgrade/el7toel8/libraries/vsftpdutils.py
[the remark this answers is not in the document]
fernflower: I was saying that the function will return None in any case, with or without explicit return None at L23.
But as you wish, it's not important thing.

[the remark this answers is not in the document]
olysonek: Well there's two cases:
1. if `read_func` succeeds, i.e. does not raise an exception, `get_config_contents` will return the file contents (it cannot return None in this case)
2. if `read_func` fails, i.e. raises an exception, it means the file could not be read (for whatever reason) and `get_config_contents` returns None

I think all is good here.

repos/system_upgrade/el7toel8/actors/vsftpdconfigcheck/libraries/library.py
```diff
@@ -0,0 +1,19 @@
...
+from leapp.libraries.common.tcpwrappersutils import config_applies_to_daemon
+
+
+def check_config_supported(tcpwrap_facts, vsftpd_facts):
+    bad_configs = []
+    for config in vsftpd_facts.configs:
+        if config.tcp_wrappers:
+            bad_configs.append(config.path)
+    if len(bad_configs) > 0 and config_applies_to_daemon(tcpwrap_facts, 'vsftpd'):
+        report_generic(title='Unsupported vsftpd configuration',
```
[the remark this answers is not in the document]
olysonek: Done, thanks.

repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/library.py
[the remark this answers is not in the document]
pirat89: Discussed in person. The `None` makes sense.

repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/library.py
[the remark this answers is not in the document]
pirat89: Discussed in person. The `None` makes sense.

repos/system_upgrade/el7toel8/libraries/vsftpdutils.py
```diff
@@ -0,0 +1,32 @@
...
+STRICT_SSL_READ_EOF = 'strict_ssl_read_eof'
+TCP_WRAPPERS = 'tcp_wrappers'
+
+
+def read_file(path):
+    with open(path, 'r') as f:
+        return f.read()
+
+
+def get_config_contents(path, read_func=read_file):
```
pirat89:
```suggestion
def _get_config_contents(path, read_func=read_file):
```

repos/system_upgrade/el7toel8/libraries/vsftpdutils.py
```diff
@@ -0,0 +1,32 @@
...
+
+from leapp.libraries.stdlib import api
+
+VSFTPD_CONFIG_DIR = '/etc/vsftpd'
+VSFTPD_DEFAULT_CONFIG_PATH = '/etc/vsftpd/vsftpd.conf'
+STRICT_SSL_READ_EOF = 'strict_ssl_read_eof'
+TCP_WRAPPERS = 'tcp_wrappers'
+
+
+def read_file(path):
```
pirat89:
```suggestion
def _read_file(path):
```

repos/system_upgrade/el7toel8/models/vsftpdfacts.py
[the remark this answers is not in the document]
pirat89: Discussed in person. Nullable makes sense. will be documented.

pirat89:
@olysonek This is how you can document fields in models: https://github.com/oamg/leapp/blob/1f842237752860d395e184b4934f8c3ff98dff57/leapp/workflows/__init__.py#L51

It should be the same as we use for classes.

repos/system_upgrade/el7toel8/actors/tcpwrappersconfigread/libraries/library.py
```diff
@@ -0,0 +1,57 @@
...
+
+def is_comment(line):
+    return len(line) == 0 or line.isspace() or line.startswith('#')
+
+
+def get_daemon_list_in_line(line):
+    daemon_list = line.split(':', 1)[0]
+    daemon_list = daemon_list.replace(',', ' ')
+    daemon_list = daemon_list.replace('\t', ' ')
+    daemon_list = daemon_list.split(' ')
```
fernflower: nit: wondering if re.split(',| |\t', daemon_list) is more readable

repos/system_upgrade/el7toel8/actors/tcpwrappersconfigread/libraries/library.py
```diff
@@ -0,0 +1,57 @@
...
+def get_daemon_lists_in_file(path, read_func=read_file):
+    ret = []
+    try:
+        content = read_func(path)
+    except IOError as e:
+        if e.errno != errno.ENOENT:
+            api.current_logger().warning('Failed to read %s: %s' % (path, e))
+        return ret
+    lines = get_lines(content)
+    lines = [line for line in lines if not is_comment(line)]
```
fernflower: nit: Squash L41-L45?
```
return [get_daemon_list_in_line(line) for line in get_lines(content) if not is_comment(line)]
```

repos/system_upgrade/el7toel8/actors/vsftpdconfigcheck/libraries/library.py
```diff
@@ -0,0 +1,23 @@
+from leapp.libraries.common.reporting import report_with_links
+from leapp.libraries.common.tcpwrappersutils import config_applies_to_daemon
+
+
+def check_config_supported(tcpwrap_facts, vsftpd_facts):
+    bad_configs = []
```
fernflower: nit: I'd rather use list comprehension here as well instead of L6-L9
```bad_configs = [config.path for config in vsftpd_facts.configs if config.tcp_wrappers]```

File: repos/system_upgrade/el7toel8/libraries/tcpwrappersutils.py [hunk 18]
```diff
@@ -0,0 +1,80 @@
...
+        ix1 = pattern.find('*', part_beginning)
+        ix2 = pattern.find('?', part_beginning)
+        ix1 = len(pattern) if ix1 < 0 else ix1
+        ix2 = len(pattern) if ix2 < 0 else ix2
+        part_end = min(ix1, ix2)
+
+        regex += re.escape(pattern[part_beginning:part_end])
+
+        if part_end < len(pattern):
+            if pattern[part_end] == '*':
```
fernflower: nit: Squash L17-L20 to ternary operator?
```regex += '.*' if pattern[part_end] == '*' else '.'```

File: repos/system_upgrade/el7toel8/actors/tcpwrappersconfigread/libraries/library.py (hunk 15 above)
Comment (olysonek) [reply repos/system_upgrade/el7toel8/actors/tcpwrappersconfigread/libraries/library.py]: Nice, thanks! I didn't realize that exists :).

Comment (olysonek) [reply repos/system_upgrade/el7toel8/actors/tcpwrappersconfigread/libraries/library.py]: Done. I've also extended the test coverage a bit.

File: repos/system_upgrade/el7toel8/actors/tcpwrappersconfigread/libraries/library.py (hunk 16 above)
Comment (olysonek) [reply repos/system_upgrade/el7toel8/actors/tcpwrappersconfigread/libraries/library.py]: OK. I've learned to instinctively avoid complex list comprehensions, that's why you're seeing this :). But this is just fine. Done.

File: repos/system_upgrade/el7toel8/actors/vsftpdconfigcheck/libraries/library.py (hunk 17 above)
Comment (olysonek) [reply repos/system_upgrade/el7toel8/actors/vsftpdconfigcheck/libraries/library.py]: Absolutely, thanks. Done. I've also extended the test coverage a bit.

File: repos/system_upgrade/el7toel8/libraries/tcpwrappersutils.py (hunk 18 above)
Comment (olysonek) [reply repos/system_upgrade/el7toel8/libraries/tcpwrappersutils.py]: I find my version more readable.

File: repos/system_upgrade/el7toel8/actors/vsftpdconfigupdate/actor.py [hunk 19]
```diff
@@ -0,0 +1,26 @@
...
+class VsftpdConfigUpdate(Actor):
+    """
+    Modifies vsftpd configuration files on the target RHEL-8 system so that the effective
+    configuration is the same, where possible. This means doing two things:
+    1. Reverting the default configuration file (/etc/vsftpd/vsftpd.conf) to its state
+       before the upgrade (where it makes sense), if the configuration file was being used
+       with its default content (i.e., unmodified) on the source system (the configuration
+       file gets replaced with a new version during the RPM upgrade in this case).
+    2. Adding 'option=old_effective_value' to configuration files for options whose default
+       default value has changed, if the option is not explicitly specified in the configuration
```
pirat89: double *default*
```suggestion
       value has changed, if the option is not explicitly specified in the configuration
```

repos/system_upgrade/el7toel8/libraries/vsftpdutils.py
olysonek: This turned out to be a public function :). I'll document it. I've renamed the other private functions though.

pirat89:
Missing a check whether the RPM is installed. Upgrade fails on system without vsftpd:
```
    self.run()
  File "/usr/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 59, in _do_run
    target_actor(logger=logger, messaging=messaging).run(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/leapp/actors/__init__.py", line 299, in run
    self.process(*args)
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/vsftpdconfigread/actor.py", line 18, in process
    self.produce(library.get_vsftpd_facts())
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/library.py", line 39, in get_vsftpd_facts
    configs = _get_parsed_configs(read_func=read_func, listdir=listdir)
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/library.py", line 24, in _get_parsed_configs
    for fname in listdir(VSFTPD_CONFIG_DIR):
OSError: [Errno 2] No such file or directory: '/etc/vsftpd'
```

repos/system_upgrade/el7toel8/libraries/vsftpdutils.py
olysonek: I've documented it.

repos/system_upgrade/el7toel8/libraries/vsftpdutils.py
olysonek: Actually, this is public API as well :). I've documented it.

repos/system_upgrade/el7toel8/actors/vsftpdconfigupdate/actor.py
olysonek: Good catch, thanks. Fixed.

olysonek:
> Missing a check whether the RPM is installed. Upgrade fails on system without vsftpd:

Thanks, and sorry about that. I've fixed the traceback and added a test for it. I'll add the RPM check too.

repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/library.py
```diff
@@ -0,0 +1,51 @@
...
+            path = os.path.join(VSFTPD_CONFIG_DIR, fname)
+            if not path.endswith('.conf'):
+                continue
+            content = get_config_contents(path, read_func=read_func)
+            if content is None:
+                continue
+            parsed = _parse_config(path, content)
+            if parsed is not None:
+                res.append((path, parsed))
+    except IOError as e:
```
pirat89: OSError: [Errno 2] No such file or directory: '/etc/vsftpd'

```suggestion
    except OSError as e:
```

pirat89:
@olysonek I will wait for the RPM check. anyway, when you add that, please rebase against current master branch. From my light testing when `vsftpd` is installed, everything seems ok and working regarding described behaviour.

repos/system_upgrade/el7toel8/actors/vsftpdconfigread/libraries/library.py
olysonek: Thanks again. Hopefully fixed now. I also noticed that I wasn't creating IOError and OSError in the tests correctly (the errno attribute wasn't being set), so I fixed that. And rebased against origin/master.

olysonek: Notice that there's a newly added first commit.

olysonek:
@fernflower @pirat89 Thanks for the thorough review and testing!

Jakuje:
I am still wondering whether there should not be some kind of global check that could inhibit the update whenever there is _anything_ useful in tcp_wrappers files, rather than writing actors per daemon to check this. As far as I know, most of the other daemons really do not need special treating since they do not need another configuration option to enable tcp wrappers like in vsftpd.

For example, I would certainly like to have openssh check, since that is really common, but it would be a very simple one using the library provided by this PR.

Please, let me know about your thoughts what would be preferred.

I would personally like to avoid writing some whitelist of daemons (even though I have some list from fedora [1]) that would be tried since there might be some customer application that used to use the tcp_wrappers and they still think they depend on it even after the rebuild (with a good configure check disabling the support in rebuild) the no longer do.

[1] https://fedoraproject.org/wiki/Changes/Deprecate_TCP_wrappers

olysonek:
> I would personally like to avoid writing some whitelist of daemons (even though I have some list from fedora [1]) that would be tried since there might be some customer application that used to use the tcp_wrappers and they still think they depend on it even after the rebuild (with a good configure check disabling the support in rebuild) the no longer do.

That is a good point. We may need to rethink this.

The reason I originally decided to do the check directly in the vsftpd actor is so that the upgrade process is as smooth as possible. I.e. do the reporting and upgrade inhibiting only when really necessary.

For example, let's say there's some tcp_wrappers configuration that applies to vsftpd, but
a) tcp_wrappers support is turned off in the vsftpd configuration, or
b) vsftpd is not even installed (anymore)

In either of these cases, it seems unnecessary to nag the user about unsupported configuration. So I check these conditions in the vsftpd actors.

But given the point you made above about customer applications depending on tcp_wrappers, I'm starting to think a global check might be a better solution after all.

Jakuje:
On the other hand, I am quite not sure what assurances we can provide for such custom applications. The tcp wrappers package is gone from RHEL8 so we might even just do a list of applications that used tcp_wrappers in RHEL7 and do this in one separate actor (probably also with check whether the separate package is installed to avoid too much noise).

Please, let me know your thoughts or whether you will have time to check that or I should have a look into that.

pirat89:
@olysonek There is chance that users are using custom / 3rd-party applications that depends on tcp_wrappers itself. From that point, not sure whether we should inhibit the upgrade or just inform users about that. Probably better would be to wait for implementation of the answerfile - so it will be possible to inhibit upgrade by default until user do changes or keep users to decide whether this is OK or not.

olysonek:
@Jakuje Please look into it, I'm afraid I won't have the time. I'm happy to do changes to the vsftpd actors however, if that becomes necessary.

@pirat89 The VsftpdConfigCheck actor currently inhibits the upgrade and tells the user to migrate the tcp_wrappers configuration (all that is really needed is to remove the tcp_wrappers configuration). It does this so that the upgraded system is not insecure. Do you think it should not inhibit the upgrade? I mean, we can change it to use answerfile once it becomes available, however until then, personally I'd leave it as it is.

PR #189, opened 2019-05-07
Added acpid actor

yarda:


yarda:
Renamed to checkacpid

PR #191, opened 2019-05-07
Added irssi actor

yarda:

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkirssi/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkirssi/actor.py
@@ -1,0 +1,30 @@
+from leapp.actors import Actor
+from leapp.models import InstalledRedHatSignedRPM
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/el7toel8/actors/irssi/actor.py
```diff
@@ -0,0 +1,30 @@
+from leapp.actors import Actor
+from leapp.models import InstalledRedHatSignedRPM
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+from leapp.reporting import Report
+from leapp.libraries.common.reporting import report_with_remediation
+
+
+class Irssi(Actor):
```
artmello: Could we have a more descriptive name for this actor? Maybe something like "CheckIrssiInstalled"?

yarda: I renamed all to checkirssi (including the filename).

yarda: If you need different names, please tell me what to rename to what.

PR #193, opened 2019-05-13
Add dhclient actor

landgraf:


repos/system_upgrade/el7toel8/actors/dhclientreport/actor.py
```diff
@@ -0,0 +1,33 @@
...
+    def process(self):
+        if is_dhclient_installed():
+            title = 'dhclient command line options changed'
+            summary = 'dhclient command line options have been changed ' \
+                      'since RHEL 7.7 release.'
+            remediation = 'If you have custom script which '\
+                          'involves dhclient please check dhclient(8) man page for ' \
+                          '-I/-C options'
+            severity = 'low'
+            report_with_remediation(title = title, severity = severity, remediation = remediation, summary = summary)
```
fernflower: I think this would trigger E251 Unexpected spaces around keyword

Please remove spaces around keyword params of report_with_remediation

repos/system_upgrade/el7toel8/actors/dhclientreport/actor.py
```diff
@@ -0,0 +1,33 @@
...
+        for fact in self.consume(InstalledRedHatSignedRPM):
+            for rpm in fact.items:
+                if rpm.name == 'dhclient':
+                    return True
+        return False
+    
+    def process(self):
+        if is_dhclient_installed():
+            title = 'dhclient command line options changed'
+            summary = 'dhclient command line options have been changed ' \
```
fernflower: Nit: I'd use multiline strings instead \, something like 
summary = ("dhclient command line options have been changed "
                   "since Rhel 7.7 release")

repos/system_upgrade/el7toel8/actors/dhclientreport/actor.py
landgraf: Done. Pep8 is happy now

repos/system_upgrade/el7toel8/actors/dhclientreport/actor.py
landgraf: I started with this format actually but checked other actors and chose style they use.

vojtechsokol:
leapp-ci build

repos/system_upgrade/el7toel8/actors/dhclientreport/actor.py
```diff
@@ -0,0 +1,37 @@
...
+    Check if dhclient is installed
+    and report backward incompatible change in CLI options
+    """
+
+    name = "dhclient"
+    consumes = (InstalledRedHatSignedRPM,)
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def is_dhclient_installed():
```
vinzenz: Missing self argument
```suggestion
    def is_dhclient_installed(self):
```

repos/system_upgrade/el7toel8/actors/dhclientreport/actor.py
```diff
@@ -0,0 +1,37 @@
...
+        for fact in self.consume(InstalledRedHatSignedRPM):
+            for rpm in fact.items:
+                if rpm.name == 'dhclient':
+                    return True
+        return False
+
+    def process(self):
+        if is_dhclient_installed():
```
vinzenz:
```suggestion
        if self.is_dhclient_installed():
```

repos/system_upgrade/el7toel8/actors/dhclientreport/actor.py
```diff
@@ -0,0 +1,37 @@
...
+            title = 'dhclient command line options changed'
+            summary = 'dhclient command line options have been changed ' \
+                      'since RHEL 7.7 release.'
+            remediation = 'If you have custom script which involves dhclient '\
```
vinzenz: Please do not use backslash line continuity
```python
remediation = ('If you have custom script which involves dhclient '
               'please check dhclient(8) man page for '
               '-I/-C options')
```

pirat89:
@landgraf Hi, do I read it well that this actor is relevant only for upgrades from RHEL 7.6 to RHEL 8.x? IOW, that for upgrade from RHEL 7.7 and newer is not relevant?

repos/system_upgrade/el7toel8/actors/dhclientreport/actor.py
```diff
@@ -0,0 +1,36 @@
...
+    and report backward incompatible change in CLI options.
+    """
+
+    name = "dhclient"
+    consumes = (InstalledRedHatSignedRPM,)
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        if not has_package(InstalledRedHatSignedRPM, 'dhclient'):
```
pirat89: @landgraf In case this is related just for RHEL 7.6:
```suggestion
        if not has_package(InstalledRedHatSignedRPM, 'dhclient') and version.matches_source_version(['7.6']):
```

and add import:
```
from leapp.libraries.common.config import version
```

pirat89: but in such case, you would need modify even tests and move the code to the library, to be able to mock the config object.

landgraf: @pirat89  Thank you for the comment. It's not just for 7.6 but for 7.6+

pirat89:
As discussed offline, the PR is supposed to be dropped already. Closing it, but feel free to reopen if needed.

PR #194, opened 2019-05-13
Add nmap-frontend actor

landgraf:


landgraf:
Deprecation/removal of nmap-frontend package will be handled separately. Closing for now

PR #199, opened 2019-05-17
rpmscanner: move get_package_repository_data to library

pirat89:
Because of `import yum` upgrades started to be broken as when we
start leapp under Python3 (after the rpmupgrade phase) leapp crashes
during the scanning of the repository because the yum module for
Python3 doesn't exist.

Moving the functionality into the actor's library as it is not
necessary to parse the code when it is not used (the function is
used just during the FactsCollectionPhase phase.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/rpmscanner/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/rpmscanner/actor.py
@@ -1,8 +1,6 @@
-import yum
-
 from leapp.actors import Actor
+from leapp.libraries.actors import get_package_repository_data
 from leapp.libraries.common.rpms import get_installed_rpms
-from leapp.libraries.stdlib import run
 from leapp.models import InstalledRPM, RPM
 from leapp.tags import IPUWorkflowTag, FactsPhaseTag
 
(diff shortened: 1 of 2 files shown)


vinzenz:
Missing library :-)

jmikovic:
leapp-ci build

vojtechsokol:
leapp-ci build

repos/system_upgrade/el7toel8/actors/rpmscanner/actor.py
```diff
@@ -1,8 +1,6 @@
-import yum
-
 from leapp.actors import Actor
+from leapp.libraries.actors import get_package_repository_data
```
vojtechsokol:
```suggestion
from leapp.libraries.actor.library import get_package_repository_data
```

pirat89:
. o O (How many errors it is possible to do in one line...)

repos/system_upgrade/el7toel8/actors/rpmscanner/libraries/library.py
```diff
@@ -0,0 +1,12 @@
+
+def get_package_repository_data(self):
```
vojtechsokol: copy&paste error? :)
```suggestion
def get_package_repository_data():
```

repos/system_upgrade/el7toel8/actors/rpmscanner/libraries/library.py
```diff
@@ -0,0 +1,12 @@
...
+    """ Return dictionary mapping package name with repository from which it was installed """
+    # import has to be inside the function to obey troubles with non-existing
```
vojtechsokol:
```suggestion
    # import has to be inside the function to evade troubles with non-existing
```

jmikovic: I would say 'avoid' would be the best word here :D

pirat89:
I have time today, so I am going to even test it before I will send new version guys :D

/me is happy to see that people are testing the code when they are doing reviews! :champagne:

pirat89:
Now it should work. @vojtechsokol just curious, when CI writes me that everything is green and ok, where is the issue? I will try it again. But I think it points to broken CI

pirat89:
I merged this already to fix the master. However, the question around tests is not closed.

PR #202, opened 2019-05-21
Add LEAPP_DEVEL_DM_DISABLE_UDEV env var

Rezney:
When, during dnf upgrade transaction, some package calls
device_mapper which requires udev to properly release a cookie
set by device mapper, the transaction gets stuck as udev on the
host cannot see namespaced IPC inside our container.

We workaround this by setting "DM_DISABLE_UDEV=1" env var.

### DIFF
--- a/repos/system_upgrade/el7toel8/libraries/config/__init__.py
+++ b/repos/system_upgrade/el7toel8/libraries/config/__init__.py
@@ -1,0 +1,14 @@
+from leapp.libraries.stdlib import api
+
+
+def get_env(name, default=None):
+    """ Return Leapp environment variable value if matched by name """
+    for var in api.current_actor().configuration.leapp_env_vars:
+        if var.name == name:
+            return var.value
+    return default
+
+
+def get_all_envs():
+    """ Return all Leapp environment variables (both name and value) """
+    return api.current_actor().configuration.leapp_env_vars
--- a/repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/libraries/library.py
@@ -8,13 +8,23 @@ CURRENT_TARGET_VERSION = '8.1'
 ENV_IGNORE = ('LEAPP_CURRENT_PHASE', 'LEAPP_CURRENT_ACTOR', 'LEAPP_VERBOSE',
               'LEAPP_DEBUG')
 
+ENV_MAPPING = {'LEAPP_DEVEL_DM_DISABLE_UDEV': 'DM_DISABLE_UDEV'}
+
 
 def get_env_vars():
     """
-    Gather LEAPP_DEVEL environment variables and provide them as messages to be
+    Gather LEAPP_DEVEL environment variables and respective mappings to provide them as messages to be
     available after reboot.
     """
-    return [EnvVar(name=k, value=v) for (k, v) in os.environ.items() if k.startswith('LEAPP_') and k not in ENV_IGNORE]
+    env_vars = []
+    leapp_vars = {k: v for (k, v) in os.environ.items() if k.startswith('LEAPP_') and k not in ENV_IGNORE}
+    for k, v in leapp_vars.items():
+        if k in ENV_MAPPING:
+            env_vars.append(EnvVar(name=ENV_MAPPING.get(k), value=v))
+            continue
+        env_vars.append(EnvVar(name=k, value=v))
+
+    return env_vars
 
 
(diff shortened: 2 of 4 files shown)


Rezney:
Need to find out a way to pass the env var into initramdisk...

pirat89:
Please, update commit message and description. You skipped the LEAPP_DEVEL_ prefix for the var name :)

Rezney:
> Please, update commit message and description. You skipped the LEAPP_DEVEL_ prefix for the var name :)

I wanted to mention the original env var we are using, but you are right I should mention both. Anyway it is not working atm...

Rezney:
@vinzenz yep, already relalized that :) We were thinking about using the soon-to-be-added configuration actor for this. Or do you have other idea?

Rezney:
@vinzenz could you please take a look at this when you have some free cycles? thx

repos/system_upgrade/el7toel8/libraries/dnfplugin.py
```diff
@@ -106,7 +106,8 @@ def _transaction(context, stage, target_repoids, tasks, test=False):
                     stage,
                     DNF_PLUGIN_DATA_PATH
                 ],
-                callback_raw=utils.logging_handler
+                callback_raw=utils.logging_handler,
+                env=cmd_env_vars or {}
```
vinzenz: IIRC the or is irrelevant None is fine
I think that’s the default value on run

repos/system_upgrade/el7toel8/libraries/config/__init__.py
```diff
@@ -0,0 +1,10 @@
+from leapp.libraries.stdlib import api
+
+
+def get_env_var_value(name):
```
vinzenz: Why not just get_env with an or value and default value None as fallback?

repos/system_upgrade/el7toel8/libraries/dnfplugin.py
```diff
@@ -89,7 +89,7 @@ def backup_debug_data(context):
             api.current_logger().warn('Failed to copy debugdata. Message: {}'.format(e.message), exc_info=True)
 
 
-def _transaction(context, stage, target_repoids, tasks, test=False):
+def _transaction(context, stage, target_repoids, tasks, cmd_env_vars=None, test=False):
```
vinzenz: Why not just env or at least cmd_env

Rezney: Fixed the namings. It was indeed too long.

repos/system_upgrade/el7toel8/libraries/config/__init__.py
Rezney: like this?

repos/system_upgrade/el7toel8/libraries/config/__init__.py
```diff
@@ -0,0 +1,10 @@
+from leapp.libraries.stdlib import api
+
+
+def get_env(name, default=None):
+    """ Return Leapp environment variable value if matched by name """
+    env_vars = api.current_actor().configuration.leapp_env_vars
```
vinzenz: I think this temporary isn't necessary

repos/system_upgrade/el7toel8/libraries/config/__init__.py
```diff
@@ -0,0 +1,10 @@
...
+    for var in env_vars:
```
vinzenz:
```suggestion
    for var in api.current_actor().configuration.leapp_env_vars:
```

Rezney: why not, done...

Rezney:
rebased

Rezney:
@jmikovic prosim, prosim

Rezney:
@jmikovic so you can leave this for the moment. It may not be working as expected as for now...

repos/system_upgrade/el7toel8/libraries/dnfplugin.py
```diff
@@ -106,7 +106,8 @@ def _transaction(context, stage, target_repoids, tasks, test=False):
                     stage,
                     DNF_PLUGIN_DATA_PATH
                 ],
-                callback_raw=utils.logging_handler
+                callback_raw=utils.logging_handler,
+                env=cmd_env
```
vinzenz: The problem why this doesn't work is that this is a container execution and environment variables aren't forwarded. The solution to this is that we add additional support for it to the mounting.py in the libraries

drehak:
LGTM.

PR #206, opened 2019-05-23
Do not traceback when dmi subsystem not available

vojtechsokol:
When running on ppc64le or s390x machines, Biosdevname actor tracebacks, because those systems lacks DMI tables.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/biosdevname/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/biosdevname/actor.py
@@ -1,11 +1,10 @@
-import pyudev
 import re
+
+import pyudev
 
 from leapp.actors import Actor
 from leapp.models import PersistentNetNamesFacts, KernelCmdlineArg
 from leapp.tags import FactsPhaseTag, IPUWorkflowTag
-from leapp.libraries.stdlib import run, CalledProcessError
-from leapp.exceptions import StopActorExecutionError
 
 
 class Biosdevname(Actor):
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/el7toel8/actors/biosdevname/actor.py
```diff
@@ -29,9 +29,12 @@ def is_vendor_dell(self):
         context = pyudev.Context()
 
         # There should be only one dmi/id device
-        dmi = pyudev.Enumerator(context).match_subsystem('dmi')
-        dev = list(filter(lambda d: d.sys_name == 'id', dmi))[0]
-        vendor = dev.attributes.get('sys_vendor')
+        dmi = pyudev.Enumerator(context).match_subsystem('dmi').match_sys_name('id')
+        try:
```
vinzenz:
```
if not dmi:
    return False
```
?

vojtechsokol:
@vinzenz @pirat89 Could you please advise if it makes sense to return `False` from function `is_vendor_dell` when dmi info cannot be retrieved? This happens on on ppc64le and s390x systems, that lacks DMI tables. Specifically does it make sense to check if vendor is Dell for those systems? And if the retrieval should be done even then, is there a better way than parsing the ouput of `lshw` command? This should work on all systems:
```python
lshw -quiet -json | jq '.vendor'
```

repos/system_upgrade/el7toel8/actors/biosdevname/actor.py
vojtechsokol: Unfortunately this doesn't work. I tried to figure out some more elegant way than using try/except block, but I failed. This also doesn't work.

vinzenz: Well I was writing it on a mobile.
I rather expect something like this to work:

```python
dmi = pyudev.Enumerator(context).match_subsystem('dmi').match_sys_name('id')
dev = next(dmi, None)
if not dev:
    return None
vendor = dev.attributes.get('sys_vendor')
return re.search('Dell.*', str(vendor)) is not None
```

vojtechsokol: `dev = next(dmi, None)` won't work either, it would have to be like this:
`dev = next(dmi.__iter__(), None)`, which is really ugly. But the for loop should work, that's the first thing I should have tried...

PR #207, opened 2019-05-23
[WIP][DO NOT MERGE] Makefile: fix building of packages

pirat89:
Previous solution worked just for copr-cli v1.78 and lower. Adding
the helper script that process data obtained from the COPR server
and provide expected content for scripts in Makefile, so building
will be possible with various versions of copr-cli.

Additionally, the helper script provides simple library that helps
with management of COPR builds in repository, which can be useful
in case someone would like to manage builds in COPR repository.

TODO: this is not final solution, but it is almost done. I will finish it in couple of hours.


vinzenz:
@pirat89 Not knowing what all you are going to do in that python script, but well check this approach:
https://github.com/oamg/leapp-repository/pull/210/files#diff-380150e85aade1d136d8ae49f710eb8e

pirat89:
@vinzenz yes, this can be the solution as well :) I just started to write library instead of simple script just because it started to be fun. Sorry for delay

pirat89:
@vinzenz By that, I mean, this PR can be closed as the lib would be probably better to use as a script in snippet repo (or I can just store it under my repository). I think that your approach is better in the concept of this repository to keep it simple. Think about to create separate PR in case you do not want to merge the whole PR now. I think that there is just one possible problem, which is really corner case - it would happen that the top succeeded ID will be the different from the one used for the running leapp-repository build. As a machine A would do the thing faster than the machine B so you would get content for different PR. But that's really corner case and can be resolved just by checking the release of packages (just additional 3 lines).

pirat89:
Closing. It will be fixed by https://github.com/oamg/leapp-repository/pull/212

PR #211, opened 2019-05-24
lint: introduce linter check with pylint

fernflower:
Added a minimal pylint config with everything disabled,
so that removing any line from disabled checks will
cause pylint falures on existing codebase.
The following commits will remove checks from
.pylintrc step by step with the following necessary code
refactor to make linter happy again.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,63 @@
+[MESSAGES CONTROL]
+disable=
+# "F" Fatal errors that prevent further processing
+ import-error,
+# "I" Informational noise
+# "E" Error for important programming issues (likely bugs)
+ no-member,
+ no-name-in-module,
+ raising-bad-type,
+ undefined-variable,
+# "W" Warnings for stylistic problems or minor programming issues
+ anomalous-backslash-in-string,
+ arguments-differ,
+ attribute-defined-outside-init,
+ bad-indentation,
+ broad-except,
+ cell-var-from-loop,
+ deprecated-lambda,
+ fixme,
+ logging-not-lazy,
+ lost-exception,
+ no-init,
+ pointless-string-statement,
+ protected-access,
+ redefined-builtin,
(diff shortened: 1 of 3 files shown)


.pylintrc
```diff
@@ -0,0 +1,63 @@
...
+ too-few-public-methods,
+ too-many-branches,
+ too-many-locals,
+ too-many-statements,
+# new for python3 version of pylint
+ useless-object-inheritance
+
+[FORMAT]
+# Maximum number of characters on a single line.
+max-line-length=200
```
pirat89: We agreed time ago to keep `max-line-length` to 119. See the *.flake8* file
```suggestion
max-line-length=119
```

fernflower: Done

vojtechsokol:
Please rebase to master, CI issues should be fixed by now.

fernflower:
@vojtechsokol I couldn't find an easy way to utilize pytest to do flake8/linter checks for leapp-repository as it stumbles upon import errors. Probably it can be done by tuning our utils/tun_pytest.py helper but honestly I wouldn't complicate things and go with pure pylint for linting and no flake8 check.

vojtechsokol:
np, I won't block this PR because of this. However there seems to be still an issue in the copr build - I joust quickyi skimmed through the logs and there seems to be a problem with name of the branch, specificaly slash seems to be the problem.

vinzenz:
Yeah for the rpms slashes are a no no in the name
Since the branch name is part of the rpm for PRs

fernflower:
Opened a https://github.com/oamg/leapp-repository/pull/214 . Won't delete old branch for some time, not sure if deleting it won't cause the loss of all comments due to PR auto closing.

pirat89:
@fernflower comments should not be lost, they should be just 'rolled' and you can unroll them. If I am not missing something (it's possible too).

PR #212, opened 2019-05-27
Makefile: Fix copr-cli incompatibility

vinzenz:
This patch introduces a script to retrieve the latest build id from
COPR via the v3 client API.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,9 @@ DEPS_PKGNAME=leapp-el7toel8-deps
 VERSION=`grep -m1 "^Version:" packaging/$(PKGNAME).spec | grep -om1 "[0-9].[0-9.]**"`
 DEPS_VERSION=`grep -m1 "^Version:" packaging/$(DEPS_PKGNAME).spec | grep -om1 "[0-9].[0-9.]**"`
 
+# needed only in case the Python2 should be used
+_PYTHON_INTERPRETER=$$(_PYTHON_INTERPRETER)
+
 # by default use values you can see below, but in case the COPR_* var is defined
 # use it instead of the default
 _COPR_REPO=$${COPR_REPO:-leapp}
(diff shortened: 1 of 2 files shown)


vinzenz:
leapp-ci build

utils/get_latest_copr_build
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python
```
fernflower: Just curious - did you intentionally leave out .py extension from utils/get_latest_copr_build ?

vinzenz: in this case yeah, because I don't really like to have tools with .py extension - I find it awkward to call anything that has still .py in its name

pirat89: we should not use `/usr/bin/env python` and we should use `/usr/bin/python3` nowadays, but anyway, as this is just helping script, I think it is ok.

pirat89: heh. Ok, just now I found that the library is just for Python3. So this should be changed as well (F29 still exists with Python2 as default ;-))

pirat89:
In my case it's broken.
```
COPR_REPO=${COPR_REPO_TMP:-leapp-tmp} COPR_PACKAGE=leapp-el7toel8-deps ./utils/get_latest_copr_build --id
  File "./utils/get_latest_copr_build", line 18, in <module>
    ownername, projectname = os.getenv('COPR_REPO', '').split('/', 1)
ValueError: not enough values to unpack (expected 2, got 1)
make: *** [Makefile:85: source] Error 1
```

The COPR_REPO contains: 'leapp-tmp'
Possibly you ran `make` with additional agurments? In default this is not expected usually. I am working on PR on this PR.

pirat89:
leapp-ci build

Makefile
```diff
@@ -4,6 +4,9 @@ DEPS_PKGNAME=leapp-el7toel8-deps
 VERSION=`grep -m1 "^Version:" packaging/$(PKGNAME).spec | grep -om1 "[0-9].[0-9.]**"`
 DEPS_VERSION=`grep -m1 "^Version:" packaging/$(DEPS_PKGNAME).spec | grep -om1 "[0-9].[0-9.]**"`
 
+# needed only in case the Python2 should be used
+_PYTHON_INTERPRETER=$$(_PYTHON_INTERPRETER)
```
pirat89: the brackets... 
```suggestion
_PYTHON_INTERPRETER=$${_PYTHON_INTERPRETER}
```

vinzenz:
leapp-ci build

PR #213, opened 2019-05-28
WIP: Unify rpm installation source

Rezney:
Put local/remote rpm paths into the same bucket with
to_install pkg names and decide install method in the
plugin.

This also fixes/enables including remote/local rpms
coming from /etc/transaction files

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/dnfshellrpmupgrade/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/dnfshellrpmupgrade/actor.py
@@ -72,7 +72,6 @@ class DnfShellRpmUpgrade(Actor):
         plugin_data = {
             'pkgs_info':
                 {
-                    'local_rpms': [pkg for pkg in data.local_rpms],
                     'to_install': [pkg for pkg in data.to_install],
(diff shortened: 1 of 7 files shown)


Rezney:
Having issues with dnf plugin not using cache with remote packages. Need to investigate further.

shaded-enmity:
@Rezney is it a problem though? I think that the only supported use case for this is CI anyway, where I guess this is not much of a problem. How about we just add a comment?

Rezney:
@shaded-enmity It is a problem cause we fail in the initramdisk when trying to fetch the package without networking as it is not cached. We can definitely workaround it somehow but it should work:

  https://dnf.readthedocs.io/en/latest/api_base.html#dnf.Base.add_remote_rpms says: "Does the download to a temporary files for each path if path is a remote URL"

Rezney:
Setting `keepcache=True` helps to keep remote_packages cached but the cache stays there even after successful transaction (obviously)...

Rezney:
This was mainly for CI use case but as we would have to change the plugin behaviour, we will abandon this solution.

PR #219, opened 2019-05-30
testing: Tweaked testing to allow one actor test

vinzenz:
This patch allows to run tests without linting via
`make test_no_lint`

Additionally this patch introduces necessary changes to allow to specify
one actor whose tests are going to be run and none else.

This is very convenient when writing actors and not having to execute
the whole repository of tests every time.

`make test ACTOR=MyActorClassName`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,10 @@ PKGNAME=leapp-repository
 DEPS_PKGNAME=leapp-el7toel8-deps
 VERSION=`grep -m1 "^Version:" packaging/$(PKGNAME).spec | grep -om1 "[0-9].[0-9.]**"`
 DEPS_VERSION=`grep -m1 "^Version:" packaging/$(DEPS_PKGNAME).spec | grep -om1 "[0-9].[0-9.]**"`
-
+ACTOR_PATH=repos
+ifdef ACTOR
+	ACTOR_PATH=`python utils/actor_path.py $(ACTOR)`
+endif
 # needed only in case the Python2 should be used
 _USE_PYTHON_INTERPRETER=$${_PYTHON_INTERPRETER}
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -148,13 +148,15 @@ install-deps:
 	pip install --upgrade -r requirements.txt
 	python utils/install_actor_deps.py --actor=$(ACTOR)
 
-test:	lint
+test_no_lint:
 	. tut/bin/activate; \
 	python utils/run_pytest.py --actor=$(ACTOR) --report=$(REPORT)
 
+test: lint test_no_lint
```
fernflower: That's really weird but 
`lint test_no_lint` gives me only linter output while `test_no_lint lint` runs both the unittests for actor and the linter for everything as expected.

vinzenz: While on travis it works o.O

vojtechsokol: Works for me on rhel7 in VM...

vojtechsokol:
It would also be nice to propagate the selection of chosen actor to the `lint` target. Because `make test ACTOR=MyActorClassName` runs the test only for the selected actor but it lints the whole actor repository.

vinzenz:
Well for now one can use at least test_no_lint to avoid it and call the pylint manually.
Running manually the tests for an actor was more often a problem.

vinzenz:
> It would also be nice to propagate the selection of chosen actor to the `lint` target. Because `make test ACTOR=MyActorClassName` runs the test only for the selected actor but it lints the whole actor repository.

Well anyway added

Makefile
fernflower: If it's not a big deal, can we have `test_no_lint lint` order? )

vinzenz: Then you lint after the tests?

fernflower: Is there a problem? To receive +1 from actual ci we would still need to have a happy linter and 100% passing tests regardless of the execution order

vinzenz: I just didn't want to change the current behaviour

But instead of a 'it doesn't work for me, so change the order' it would be a good thing to know, __why__ it doesn't work for you

PR #201, opened 2019-05-21
[Duplicate] [bad branch name] upgrade library: added has_package check

fernflower:
Added method to the signed/unsigned rpms models to
check for the given package presence.
This should reduce a bit code duplication in any other actors
that check that the given package is present in the system.

PR #225, opened 2019-05-31
upgrade library: added has_package check 

fernflower:
Duplicate of https://github.com/oamg/leapp-repository/pull/201 but with proper branch name.
Has been rebased and has additional test coverage for create_lookup functionality.


repos/system_upgrade/el7toel8/libraries/rpms.py
```diff
@@ -16,3 +17,37 @@ def get_installed_rpms():
...
+    :param field: model field, its value will be taken for lookup data
+    :param key: property of the field's data that will be used to build a resulting set
+    :param context: context of the execution
+    """
+    data = getattr(next((m for m in context.consume(model)), model()), field)
+    try:
+        return {getattr(obj, key) for obj in data} if data else {}
+    except TypeError:
+        # data is not iterable, not lookup can be built
+        stdlib.api.current_logger().error(
```
shaded-enmity: Shouldn't we use context here as well? i.e.
`stdlib.api.current_logger()` → `context.current_logger()` ?

vinzenz: @shaded-enmity no not really because context can be an actor and that doesn’t have current_logger

PR #218, opened 2019-05-29
Handle supplementary repository

bocekm:
This change will make sure that any PES event pointing to the RHEL 8 Supplementary repo will get processed. Process:
- there is a RH-signed pkg installed from RHEL 7 Supp repo
- there is a PES event saying that this pkg is "Present" in RHEL 8 Supp repo
- based on the above, leapp will:
  - enable RHEL 8 Supp repo for the upgrade transaction (using the dnf upgrade plugin)
  - upgrade the package in initramfs
  - enable RHEL 8 Supp repo in the first boot phase on the upgraded RHEL 8 so that it's available to the user (using subscription-mananger)

With this PR Leapp will start processing also Present, Deprecated and Moved PES events. For example virtio-win has a Present event in PES, because it was on RHEL 7 Supplementary repo and on RHEL 8 it's also in Supplementary (in 8.0, not 8.1).

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1621775

PR #228, opened 2019-06-03
Fix finding copr builds

bocekm:
- use str.startwith instead of str.rstrip as rstrip looks for chars, not for a string
  - the previous code did not find the build for https://github.com/oamg/leapp-repository/pull/218
- also print last 10 found copr builds only with --debug
  - this was incorrectly the other way around before: the script was printing it only without --debug

### DIFF
--- a/utils/get_latest_copr_build
+++ b/utils/get_latest_copr_build
@@ -6,25 +6,58 @@ import sys
 
 import copr.v3
 
-CONFIG_PATH = os.path.expandvars(os.getenv('_COPR_CONFIG', '~/.config/copr'))
-client = copr.v3.Client(copr.v3.config_from_file(path=CONFIG_PATH))
 
-# In case we are not able to retrieve from env vars, use defaults from the
-# configuration files.
-ownername = os.getenv('COPR_OWNER')
-projectname = os.getenv('COPR_PROJECT')
-packagename = os.getenv('COPR_PACKAGE')
-pkgrelease = os.getenv('PKG_RELEASE')
+def exit_fail():
+    sys.stderr.write('Passed (or default) environment variables:\n')
+    for var, value in ENV_VARS.items():
+        sys.stderr.write('  {}: {}\n'.format(var, value))
+    sys.exit(1)
+
+
+def get_owner_from_config():
+    # We have just the name of project inside COPR_REPO value which is expected when the user
+    # is the owner of the project. So set the user as owner from the config file.
+    return client.config['username']
+
+
+ENV_VARS = {
+    '_COPR_CONFIG': '~/.config/copr',  # Copr config file. Get it through https://<copr instance>/api/.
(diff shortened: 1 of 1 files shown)


jmikovic:
copr build job fails:

```
17:38:41 17:38:41 Traceback (most recent call last):
17:38:41 17:38:41   File "./utils/get_latest_copr_build", line 45, in <module>
17:38:41 17:38:41     _release = re.findall(r"^[0-9.]*?-(.*)\.el8.*?$", build)
17:38:41 17:38:41   File "/usr/lib64/python2.7/re.py", line 177, in findall
17:38:41 17:38:41     return _compile(pattern, flags).findall(string)
17:38:41 17:38:41 TypeError: expected string or buffer```

File: utils/get_latest_copr_build [hunk 1]
```diff
@@ -35,20 +36,16 @@ builds = client.build_proxy.get_list(
...
 for build in builds:
     # Version in COPR contains VERSION-RELEASE string. We need just the
-    # release. Additionally we do not care about *elX* suffix. So remove it
-    # when compare.
-    _release = build['source_package']['version'].split("-")[-1]
-    # FIXME: instead of rsplit(.el8) remove whole elx.* as it can contain
-    # later e.g. .el8_0 instead of .el8
-    if _release.rstrip(".el8") == pkgrelease:
+    # release. Additionally we do not care about *el8* suffix.
+    _release = re.findall(r"^[0-9.]*?-(.*)\.el8.*?$", build)
```
vinzenz: I don't see the reason here for this regex I would do it similar like it was before:

```python
_release = build['source_package']['version'].split('_')[-1]
if _release.startswith(pkgrelease):
```

And then it doesn't matter what the dist part is

bocekm: Ok, that makes sense and the code is shorter. Thanks.

pirat89:
Hmm.. it looks like there is a problem with API v3 for Python2. I tested this with Python3.

pirat89:
leapp-ci build

vinzenz:
@pirat89 possibly the  lib from pypi needs to be added for compatibility

bocekm:
If you mean the error `copr.v3.exceptions.CoprRequestException: Request wasn't successful, there is probably a bug in the API code.`, I got it too when I was testing the `get_latest_copr_build` - in case I ran just `python3 utils/get_latest_copr_build` without setting COPR_REPO, COPR_OWNER or COPR_PROJECT env vars. I'll add some additional checks there.

pirat89:
@bocekm  maybe put there print of those values when --debug is set

pirat89:
aand maybe in case of the crash,  print automatically that stuff too...

bocekm:
@pirat89, @vinzenz, requested changes done.

bocekm:
@vinzenz, please check the current version. I've implemented your comment.

PR #229, opened 2019-06-03
biosdevname: make doc string accurate

msekletar:
We enable biosdevname only if system is a Dell box and used biosdevname previously on all physical network interfaces.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/biosdevname/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/biosdevname/actor.py
@@ -9,7 +9,7 @@ from leapp.tags import FactsPhaseTag, IPUWorkflowTag
 
 class Biosdevname(Actor):
     """
-    Enable biosdevname on RHEL8 if all interfaces on RHEL-7 used biosdevname naming scheme and if machine vendor is DELL
+    Enable biosdevname on RHEL8 if all interfaces on RHEL7 use biosdevname naming scheme and if machine vendor is DELL
     """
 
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/el7toel8/actors/biosdevname/actor.py
```diff
@@ -9,7 +9,7 @@
 
 class Biosdevname(Actor):
     """
-    Enable biosdevname on RHEL8 if all interfaces on RHEL-7 used biosdevname naming scheme or if machine vendor is DELL
+    Enable biosdevname on RHEL8 if all interfaces on RHEL-7 used biosdevname naming scheme and if machine vendor is DELL
```
examon: > [C0301 line-too-long] Line too long (120/119) File: repos/system_upgrade/el7toel8/actors/biosdevname/actor.py, line 12

;-(

vojtechsokol: Using present tense in both parts of the statement fixes the 'problem' :)
```suggestion
    Enable biosdevname on RHEL8 if all interfaces on RHEL-7 use biosdevname naming scheme and if machine vendor is DELL
```

vinzenz: @vojtechsokol unfortunately that's not correct because it would have to be: 

if all interfaces on RHEL-7 use **the** biosdevname naming scheme

PR #232, opened 2019-06-05
lint: enable some more checks

fernflower:
Enabled:
- attribute-defined-outside-init,
- deprecated-lambda,
- logging-not-lazy,
- redefined-builtin,
- undefined-loop-variable [possibly uncovered a bug in SetupTargetRepos actor]

Besides some duplicated code from tests has been moved to
leapp.library.common.testutils module, other tests that were
using report_generic_mocked and produce_mocked refactored.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkosrelease/tests/test_checkosrelease.py
+++ b/repos/system_upgrade/el7toel8/actors/checkosrelease/tests/test_checkosrelease.py
@@ -6,7 +6,7 @@ from leapp.snactor.fixture import current_actor_context
 def create_osrelease(release_id, version_id=None):
     version_id = version_id or '42'
     return OSReleaseFacts(
-        id=release_id,
+        release_id=release_id,
         name='test',
         pretty_name='test {}'.format(version_id),
         version='Some Test {}'.format(version_id),
--- a/repos/system_upgrade/el7toel8/actors/checkosrelease/tests/test_checkosrelease.py
+++ b/repos/system_upgrade/el7toel8/actors/checkosrelease/tests/test_checkosrelease.py
@@ -3,10 +3,10 @@ from leapp.reporting import Report
 from leapp.snactor.fixture import current_actor_context
 
 
-def create_osrelease(id, version_id=None):
+def create_osrelease(release_id, version_id=None):
     version_id = version_id or '42'
     return OSReleaseFacts(
-        id=id,
+        id=release_id,
         name='test',
         pretty_name='test {}'.format(version_id),
         version='Some Test {}'.format(version_id),
@@ -17,7 +17,7 @@ def create_osrelease(id, version_id=None):
 
 
 def test_not_supported_id(current_actor_context):
-    current_actor_context.feed(create_osrelease(id='not_supported_id'))
+    current_actor_context.feed(create_osrelease(release_id='not_supported_id'))
     current_actor_context.run()
     assert current_actor_context.consume(Report)
     assert current_actor_context.consume(Report)[0].title == 'Unsupported OS id'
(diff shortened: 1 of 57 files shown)


vojtechsokol:
LGTM.

Fix for #233 is here #234 (hopefully).

Also I encountered other candidates for deduplication:
- function [get_rhel8_kernel_version](https://github.com/oamg/leapp-repository/search?q=get_rhel8_kernel_version&unscoped_q=get_rhel8_kernel_version)
- function [_logging_handler](https://github.com/oamg/leapp-repository/search?q=_logging_handler&unscoped_q=_logging_handler)

Do you want to dedup those in this PR?

fernflower:
@vojtechsokol I also thought it was a copy-paste error at first, but turns out it has to be solved differently as used_repo has completely different structure. I think I'll skip this erroneous check in the current patch but will refactor duplicated code you found, thanks!

fernflower:
@vojtechsokol tbh I'm having second thoughts about refactoring of duplicated code - there are only 2 occurencies so probably I'd do that in a following patch together with enabling duplicate-code pylint check.
I think I might have hit another bug/strange behavior while trying to enable inconsistent-return-statements with preupgradetransaction actor https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/actor.py#L339 

It either returns None in case ok/error OR returns an error message in one specific error case. This behavior looks strange to me, but I'm out of context a bit and can't find unit tests for the actor.

pirat89:
@fernflower The `process` function in actor can return always only `None`. This is mistake - probably copy&paster error. I missed that during review probably even when it seems faimilar to me. There should be something like:

```diff
             dnfplugin_dpath = '/lib/python2.7/site-packages/dnf-plugins'
 
             error = preparetransaction.copy_file_to_container(ofs_info, dnfplugin_spath, dnfplugin_dpath)
-            if error:
-                return error
 
-        error = self.dnf_plugin_rpm_download(ofs_info)
+        if not error:
+            error = self.dnf_plugin_rpm_download(ofs_info)
         if not error:
             error = preparetransaction.copy_file_from_container(
                 ofs_info,

```

Do not take care that's is ugly solution as @vinzenz completely rewrites the actor with his RHEL 8 RPM stack PR.

Do you want to resolve it as part of this PR instead of different one?

fernflower:
@pirat89 Thanks a lot, it makes sense now. I'll fix this as part of this PR then and enable skipped inconsistent-return-statements check.

repos/system_upgrade/el7toel8/actors/checkosrelease/tests/test_checkosrelease.py
```diff
@@ -3,10 +3,10 @@
 from leapp.snactor.fixture import current_actor_context
 
 
-def create_osrelease(id, version_id=None):
+def create_osrelease(release_id, version_id=None):
     version_id = version_id or '42'
     return OSReleaseFacts(
-        id=id,
+        id=release_id,
```
pirat89: note: the `id` in OSReleaseFacts will be renamed as well.

repos/system_upgrade/el7toel8/libraries/persistentnetnames.py
```diff
@@ -11,7 +11,7 @@ def physical_interfaces():
     Returns a list of pyudev.Device objects for all physical network interfaces
     """
     enumerator = pyudev.Enumerator(udev_context).match_subsystem('net')
-    return filter(lambda d: not d.device_path.startswith('/devices/virtual/'), enumerator)
+    return (d for d in enumerator if not d.device_path.startswith('/devices/virtual/'))
```
pirat89: List or tuple? "*Returns a list of...*"

fernflower: Yeah, messed up here. Should be a list, thanks!

pirat89:
LGTM. Waiting for tests and builds just to run try it to ensure we didn't miss something :)

repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -2,23 +2,19 @@
...
 import os
 import pwd
 import re
-import logging
 
-try:
-    import ConfigParser as configparser
-except ImportError:
-    import configparser
+from six.moves.configparser import ConfigParser as configparser
```
pirat89:
```suggestion
from import six.moves import configparser
```

PR #234, opened 2019-06-05
Fix processing of UsedRepositories

### LINKED ISSUE #233: setuptargetrepos: copy paste error?
See https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/actors/setuptargetrepos/actor.py#L83

which uses the `repo_file` var but in the loop it is not defined anywhere. It is used in different loop around the 60 line. That doesn't seems to be intention.

vojtechsokol:
Fixes #233

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/setuptargetrepos/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/setuptargetrepos/actor.py
@@ -80,7 +80,7 @@ class SetupTargetRepos(Actor):
         for used_repos in self.consume(UsedRepositories):
             for used_repo in used_repos.repositories:
                 used[used_repo.repository] = used_repo.packages
-                for repo in repo_file.data:
+                for repo in used_repo.data:
                     enabled_repos.add(repo.repoid)
 
         skipped_repos = skipped_repos.intersection(used.keys())


pirat89:
@vojtechsokol fix commit msg

pirat89:
@fernflower Hi, you wrote on IRC that *it's more than a copy-paste error* and from the fix I don't see anything else than copy paste error. Would you specify whether something is missing here or whether it is ok?

repos/system_upgrade/el7toel8/actors/setuptargetrepos/actor.py
```diff
@@ -80,7 +80,7 @@ def process(self):
         for used_repos in self.consume(UsedRepositories):
             for used_repo in used_repos.repositories:
                 used[used_repo.repository] = used_repo.packages
-                for repo in repo_file.data:
+                for repo in used_repo.data:
```
fernflower: My first attempt to fixing it was exactly like yours but I'm not sure it will work this way :(
`repo_file` is of type [RepositoryFile](https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/models/repositoriesfacts.py#L17) and has a data field (a list of RepositoryData objects).
While `used_repo` is of type [UsedRepository](https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/models/usedrepositories.py#L5) which doesn't have data field at all.

vinzenz:
```suggestion
```

repos/system_upgrade/el7toel8/actors/setuptargetrepos/actor.py
```diff
@@ -80,7 +80,7 @@ def process(self):
...
                     enabled_repos.add(repo.repoid)
```
vinzenz:
```suggestion
                enabled_repos.add(used_repo.repository)
```

vinzenz: While this would fix the issue, I am not sure that this is actually correct - Because who says that a used repository is actually enabled?

bocekm:
Closing in favor of https://github.com/oamg/leapp-repository/pull/301

PR #235, opened 2019-06-06
Use rhel8 systemd-nspawn

Rezney:
In order to use new "--resolv-conf=" and "--timezone=" parameters
we need systemd-nspawn at least of version 239. Currently we are
using v219.

The above mentioned parameters are needed in order to avoid
removing "/etc/localtime" symlink and "/etc/resolv.conf" file
during the upgrade.


pirat89:
@msekletar What do you thing about that?

msekletar:
I think it should be fine. Using single version of systemd-nspawn could save us some surprises exactly like this one. 

Of course this change will require proper testing.

Rezney:
rebased...

PR #236, opened 2019-06-06
WIP: Display message with path to generated report

drehak:
Related to https://github.com/oamg/leapp/pull/515.

After `leapp preupgrade` successfully ends, there's no info about where the report was generated (unless running in debug/verbose mode). This PR attempts to change that. Since the report path is known by the actor only, the message is generated from inside the actor.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/verifycheckresults/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/verifycheckresults/actor.py
@@ -1,4 +1,5 @@
 from leapp.actors import Actor
+from leapp.dialogs import RawMessageDialog
 from leapp.libraries.actor import report
 from leapp.reporting import Report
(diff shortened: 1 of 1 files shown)


pirat89:
I suggest to close this PR as the functionality should be provided by the leapp utility itself. Same like in case of the "ERRORS" message. 
```
                        ERRORS

2019-06-06 18:16:40.200865 [ERROR] Actor: verify_check_results Message: Possible problems with remote login using root account
2019-06-06 18:16:40.213454 [ERROR] Actor: verify_check_results Message: Ending process due to errors found during checks, see /var/log/leapp/leapp-report.txt for detailed report.

                     END OF ERRORS
```

@vinzenz What do you think? 

Additionally the "log paths" and path to the generated report should be provided by the leapp itself.

PR #237, opened 2019-06-07
Avoid possible misinterpratation of grubby --title parameter

jmikovic:
Value of --title wasn't in quotes which could result in it being split and cause a failure of grubby

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,3 +104,6 @@ ENV/
 # All kinds of vim stuff
 **/*~
 *.sw[a-z]
+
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -104,3 +104,6 @@ ENV/
 # All kinds of vim stuff
 **/*~
 *.sw[a-z]
+
+# visual studio code configuration
+.vscode
```
vinzenz: This should be actually a different commit, not talking about a different PR ... ;-)
But yes this is good to have

PR #241, opened 2019-06-10
Add debug-shell service to dracut

Rezney:
With debug-shell.service running we get another console
on tty9 which can help with debugging currently running
processes.

Unfortunately atm activating the service from kernel
cmdline using "systemd.debug-shell=1" does not seem
to work.

As a workaround we can drop to dracut emergency shell
using "rd.break=upgrade", then issue:

\# systemctl start debug-shell

then we can start the upgrade process manually and
switch to debug-shell to investigate \o/

One small step in the code, one giant Leapp for debugging.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/files/generate-initram.sh
+++ b/repos/system_upgrade/el7toel8/actors/prepareupgradetransaction/files/generate-initram.sh
@@ -42,6 +42,7 @@ build() {
         --confdir /var/empty \
         --force \
         --add sys-upgrade \
+        --install "/usr/lib/systemd/system/debug-shell.service /usr/lib/systemd/system-generators/systemd-debug-generator" \
         --no-hostonly \
         --nolvmconf \
         --nomdadmconf \


Rezney:
@msekletar 

any idea why would I get smaller initramdisk size with "debug-shell" included?

```
58909774 Jun 12 08:12 initramfs-upgrade.x86_64.img --without-debug-shell
58801033 Jun 12 08:27 initramfs-upgrade.x86_64.img --with-debug-shell 
```

Would it be safer to use "--install-optional" instead?

vinzenz:
TBH I rather would make this part of our dracut module instead of adding it to the command line here

vinzenz:
> @msekletar
> 
> any idea why would I get smaller initramdisk size with "debug-shell" included?
> 
> ```
> 58909774 Jun 12 08:12 initramfs-upgrade.x86_64.img --without-debug-shell
> 58801033 Jun 12 08:27 initramfs-upgrade.x86_64.img --with-debug-shell 
> ```
> 
> Would it be safer to use "--install-optional" instead?

Install optional only will not fail if the file was not found, why you end up with a smaller size, is rather puzzling to me however

Rezney:
@vinzenz 
I meant "--install-optional" as a second question (not related to the size) but I see I screwed at structuring the message.

Ok, understandable. I will close this PR. I already mentioned this PR in the ticket which should take care of improving debugging.

vinzenz:
@Rezney I know you meant it as second question - And I don't think you screwed up there at all. It's just fine.
Closing the PR wasn't necessary though

msekletar:
> any idea why would I get smaller initramdisk size with "debug-shell" included?

No idea. Initrd is just a compressed cpio archive (by default). Please inspect it manually to figure out why there is a difference. If the cpio content is the expected one, i.e. they are the same except debug-shell, then I the only thing that can make a difference is a compression. Maybe compression algorithm was bit more effective and was able to produce smaller archive.

Issue #242, opened 2019-06-11
upgrade initramfs entry cannot be found by grubby


AloisMahdal:
**Actual behavior**
After running `leapp --upgrade`, `grubby --info=DEFAULT` cannot find the newly added kernel, exiting with 1 and printing `grubby: kernel not found`.

**To Reproduce**

    leapp --upgrade
    grubby --info=DEFAULT

**Expected behavior**
Newly added entry

**System information (please complete the following information):**

RHEL-7.6 and _leapp-repository-0.7.0-1.201906101721Z.8ebef3f.master.el7_6.noarch_

****

**Additional context**

Likely caused by #238 plus a likely bug in grubby.  #238 actually added quotes around the entry title text, which results in a technically valid name but this could have broken subsequent grubby call.

I'm not sure if the upgrade is affected; it could be just the grubby bug.

bocekm:
Or another possible cause: EFI+BIOS https://bugzilla.redhat.com/show_bug.cgi?id=1718161#c5

AloisMahdal:
Well it's not an EFI system...

PR #243, opened 2019-06-11
Split grubby args and avoid need to quote

AloisMahdal:
The --foo=bar syntax can be confusing at times, so let's avoid it.

Fixes: #242

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
@@ -12,12 +12,12 @@ def add_boot_entry():
     kernel_dst_path, initram_dst_path = get_boot_file_paths()
     run([
         '/usr/sbin/grubby',
-        '--add-kernel={0}'.format(kernel_dst_path),
-        '--initrd={0}'.format(initram_dst_path),
-        '--title="RHEL Upgrade Initramfs"',
+        '--add-kernel', '{0}'.format(kernel_dst_path),
+        '--initrd', '{0}'.format(initram_dst_path),
+        '--title', 'RHEL Upgrade Initramfs',
         '--copy-default',
         '--make-default',
-        '--args="{DEBUG} enforcing=0 rd.plymouth=0 plymouth.enable=0"'.format(DEBUG=debug)
+        '--args', '{DEBUG} enforcing=0 rd.plymouth=0 plymouth.enable=0'.format(DEBUG=debug)
     ])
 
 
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
```diff
@@ -12,12 +12,12 @@ def add_boot_entry():
     kernel_dst_path, initram_dst_path = get_boot_file_paths()
     run([
         '/usr/sbin/grubby',
-        '--add-kernel={0}'.format(kernel_dst_path),
-        '--initrd={0}'.format(initram_dst_path),
-        '--title="RHEL Upgrade Initramfs"',
+        '--add-kernel', '{0}'.format(kernel_dst_path),
+        '--initrd', '{0}'.format(initram_dst_path),
+        '--title', 'RHEL Upgrade Initramfs',
```
```suggestion
        '--title', '"RHEL Upgrade Initramfs"',
```
I would leave there the quotation marks, the same as you would use them in a terminal.
Neither stdlib _run_ or __call_ tell in description how the processing of the list works, so I'd better stay defensive.
CCing @shaded-enmity.

AloisMahdal: That's not defensive; you're just passing the text *with* quotes.

I'm not sure about `run` and `_call` but standard library passes the args right to the forked process without involving shell at all.

shaded-enmity: @bocekm @AloisMahdal The list of arguments is passed verbatim to `os.exec*` function.

shaded-enmity: I.e. 
```
os.exec('/bin/echo', ['/bin/echo', '"FOO"'])
"FOO" 
```

whereas

```
os.exec('/bin/echo', ['/bin/echo', 'FOO'])
FOO
```

So it depends on what Grubby expects.

AloisMahdal:
Upon consulting in person, I've also added commit to make the unit test less confusing.

repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
AloisMahdal: Yep, that's what I expected.  (It's the safest way to call commands anyway...)

So quotes/spaces/any-old-junk does not really influence "safety".

AloisMahdal:
Already tested with `leapp-repository-0.7.0-0.201906111421Z.1d63bb7.grubby_split_args.PR243.el7_6.noarch`.  That does not contain the later commit but the later commit is only relevant to unit tests, so that one will be tested by CI.

AloisMahdal:
> ..so it will be tested by CI.

and tested it was, and fail it did.  I made a typo; force-pushing.. .:)

PR #244, opened 2019-06-11
lint: check imports order with flake8 plugin

fernflower:
Pylint's wrong-import-order was not configurable
enough to treat leapp as an application and not a
3rd party dependency.
This patch addressed the issue by utilizing
flake8-import-order plugin that can be customized.
The alphabetical ordering within group will be
checked as well so some refactoring will be done to
align with the changes.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -6,3 +6,8 @@ max-line-length = 119
     # 'leapp.snactor.fixture.current_actor_libraries' imported but unused
 per-file-ignores =
     repos/system_upgrade/el7toel8/actors/*/tests/*.py:F811,F401
+application-import-names = leapp
+# check only for correct group type and abc sorting without group
+import-order-style = pep8
+# default values + ignore strict alphabetical sorting within one module import
+ignore = E121,E123,E126,E133,E241,E242,E226,W503,W504,W505,I101
--- a/.pylintrc
+++ b/.pylintrc
@@ -25,6 +25,7 @@ disable=
  bad-continuation,
  invalid-name,
  missing-docstring,
+ wrong-import-order,
 # "R" Refactor recommendations
  duplicate-code,
  no-self-use,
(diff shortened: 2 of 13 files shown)


setup.cfg
```diff
@@ -6,3 +6,8 @@ max-line-length = 119
     # 'leapp.snactor.fixture.current_actor_libraries' imported but unused
 per-file-ignores =
     repos/system_upgrade/el7toel8/actors/*/tests/*.py:F811,F401
+application-import-names = leapp
+# check only for correct group type and abc sorting without group
+import-order-style = pep8
+# default values + ignore strict alphabetical sorting within one module import
+ignore = E121,E123,E126,E226,E24,E704,W503,W504,I101
```
vojtechsokol: I wonder why are these checks ignored - by default ignored checks also contains E133 and W505 but lacks I101.

Also I would prefer not to use E24, but to explicitly write down E241 adn E242, in case some other E24* check is added to flake8 in the future.

vojtechsokol: Ignore my comment about check I101, I have mistaken it for E101.

fernflower: Hmm the original defaults I took were from flake8 --help
``` 
 --ignore=errors       Comma-separated list of errors and warnings to ignore
                        (or skip). For example, ``--ignore=E4,E51,W234``.
                        (Default: E121,E123,E126,E226,E24,E704,W503,W504)
```
You are saying E133 and W505 are there as well?

fernflower: On the other hand - it's nice having them here.

vojtechsokol: Sorry for yet another confusion, the default configuration that ignores also E133 and W505 is for [pycodestyle](https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes), not for flake8.

PR #245, opened 2019-06-13
TcpWrappersCheck: New actor checking all daemons possibly affected by removal of the tcp wrappers package

Jakuje:
Based on the #186 and followup discussion, I created the list of packages and their keywords/daemons that were previously using tcp wrappers and that were in RHEL 7 and that are affected by the update.

I am actually not sure if all the values are correct for some of the packages so double-checking and review would be appreciated. Unfortunately some of the applications really have free-form of what is the service name in the tcp_wrappers configuration (glad we are getting rid of that) so they can be caugh only with the ALL keywords, that should match in all of the tools.

For the completeness there are also mentioned the packages that are already covered by other wrappers, which have also some more specific checks that this one general.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/tcpwrapperscheck/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/tcpwrapperscheck/actor.py
@@ -1,0 +1,66 @@
+from leapp.actors import Actor
+from leapp.models import Report, TcpWrappersFacts, InstalledRedHatSignedRPM
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+from leapp.libraries.stdlib import api
+from leapp.libraries.actor.library import config_affects_daemons
+from leapp.libraries.common.rpms import create_lookup
+from leapp.libraries.common.reporting import report_with_links
+
+
+DAEMONS = [
+    ("audit", ["auditd"]),
+    ("bacula", ["bacula"]),
+    ("conserver", ["conserver"]),
+    ("cyrus-imapd", ["imap", "pop3", "imaps", "pop3s", "sieve", "lmtp"]),
+    ("dovecot", ["imap", "pop3", "imaps", "pop3s", "sieve", "lmtp"]),
+    ("nfs-utils", ["mountd", "statd"]),
+    ("openssh-server", ["sshd"]),
+    ("proftpd", ["proftpd"]),
+    ("quota", ["rquotad"]),
+    ("rpcbind", ["rpcbind"]),
+    # ("sendmail", ["sendmail"]),  # Handled in CheckSendmail Actor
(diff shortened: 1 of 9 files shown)


repos/system_upgrade/el7toel8/actors/tcpwrapperscheck/actor.py
```diff
@@ -0,0 +1,64 @@
...
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        tcp_wrappers_facts = next(self.consume(TcpWrappersFacts))
+
+        found_packages = set()
+        found_daemons = set()
+        for (package, keywords) in DAEMONS:
+            if has_package(InstalledRedHatSignedRPM, package):
```
fernflower: Not sure if has_package is the right choice here - it will recreate the installed_packages lookup for each package in a cycle.
I'd probably go with manually creating the lookup and using it to check for package presence in a cycle, smth like
```
from leapp.libraries.common.rpms import create_lookup
...

def process(self):
...
installed_packages  = create_lookup(InstalledRedHatSignedRPM, field='items', key='name')
for (package, keywords) in DAEMONS:
    if package in installed_packages:
        ...
```

repos/system_upgrade/el7toel8/actors/tcpwrapperscheck/actor.py
```diff
@@ -0,0 +1,64 @@
...
+    might make the service more accessible than expected.
+    """
+
+    name = 'tcp_wrappers_check'
+    consumes = (TcpWrappersFacts, InstalledRedHatSignedRPM,)
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        tcp_wrappers_facts = next(self.consume(TcpWrappersFacts))
```
asilveir: Please add an alert to the user if there are more messages than expected.
Something like [this](https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/actors/redhatsignedrpmcheck/libraries/library.py#L41)

Jakuje: Thank you. I modified this also in the OpenSSH checks that I submitted some time before to do that.

Jakuje:
Simple unit test is written with the current version.

Jakuje:
I do not see why the leapp-ci failed.

asilveir:
> I do not see why the leapp-ci failed.

Thanks again, I'll take a look at your new code.
About leapp-ci is falling for all PR's it's not your fault :)

asilveir:
leapp-ci build

PR #246, opened 2019-06-13
Add additional tests for RHEL8 userspace PR

vinzenz:
PR https://github.com/oamg/leapp-repository/pull/44 did not add tests for everything. However to keep the sanity due to the size of the PR I added another PR with just added tests.


vinzenz:
merged with #44

PR #248, opened 2019-06-15
[WIP] Add shared function that blocks actor when requirements are not met

vojtechsokol:
Add `require_arch` and `require_src_minor_version` shared library functions that can be used to stop actor's execution (without stopping a workflow) if requirement on minor version of source system or architecture is not met.

`get_os_release_info` function from `OSReleaseCollector` actor is split into two functions:
- Part that parses the data from provided file is moved into the shared library as `get_os_release_info`. This function is used in the `require_src_minor_version` to check the source's system minor version.
- Part that converts the output of the former function to an `OSReleaseFacts` model. This function remains in `OSReleaseCollector` actor's library as `get_os_release_model`.

### DIFF
--- a/repos/system_upgrade/el7toel8/libraries/requirearchver.py
+++ b/repos/system_upgrade/el7toel8/libraries/requirearchver.py
@@ -1,0 +1,130 @@
+import operator
+import platform
+
+from leapp.exceptions import StopActorExecution
+from leapp.libraries.common import reporting
+from leapp.libraries.stdlib import api
+
+
+def get_os_release_info(path):
+    """
+    Retrieve data about System OS release from provided file.
+
+    :rtype: dictionary or None
+    """
+    try:
+        with open(path) as f:
+            data = dict(l.strip().split('=', 1) for l in f.readlines() if '=' in l)
+            return {key: value.strip('"') for key, value in data.items()}
+    except IOError as e:
+        reporting.report_generic(
+            title='Error while collecting system OS facts',
+            summary=str(e),
+            severity='high',
+            flags=['inhibitor'])
+        return None
+
+
+def _log_and_raise(reason, required, detected):
+    """Log that requirements were not met and stop actor's execution by raising a StopActorExecution exception."""
+    msg = 'Actor skipped, {reason} requirement not met, required: {required}, detected: {detected}'.format(
+        reason=reason, required=required, detected=detected)
+    api.current_logger().info(msg)
+    raise StopActorExecution
+
+
+def _log_passed(reason, required, detected):
+    """Log that requirements were fulfilled."""
+    msg = 'Requirement for {reason} fulfilled, required: {required}, detected: {detected}'.format(
+        reason=reason, required=required, detected=detected)
+    api.current_logger().debug(msg)
+
+
+def require_arch(required):
+    """
+    Stop actor's execution if requirement on architecture is not met.
+
+    :param required: list or tuple of vrchitectures
+    :type required: list, tuple
+    """
+    if not isinstance(required, (list, tuple)):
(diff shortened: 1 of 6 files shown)


pirat89:
@vojtechsokol commit msg & nor PR description are telling what's going on.

repos/system_upgrade/el7toel8/libraries/requirearchver.py
```diff
@@ -0,0 +1,130 @@
...
+    msg = 'Requirement for {reason} fulfilled, required: {required}, detected: {detected}'.format(
+        reason=reason, required=required, detected=detected)
+    api.current_logger().debug(msg)
+
+
+def require_arch(required):
+    """
+    Stop actor's execution if requirement on architecture is not met.
+
+    :param required: list or tuple of vrchitectures
```
Rezney: Just a typo > "architectures"

repos/system_upgrade/el7toel8/libraries/requirearchver.py
```diff
@@ -0,0 +1,130 @@
...
+    Stop actor's execution if requirement on minor version of source system is not met.
+
+    :param destination: 'source' or 'target'
+    :type destination: string
+    :param required: list or tuple of versions - integers or strings in the form '<integer>['+'|'>'|'-'|'<']'
+    :type required: list, tuple
+    :param detected: Detected version
+    :type detected: int
+    """
+    detected = int(get_os_release_info('/etc/os-release')['VERSION_ID'].split('.')[1])
```
pirat89: You have to use data from the actor config (PR is waiting for review and finishing) or you have to use OSReleaseFacts as temporary workround. You cannot access the file in all phases.

repos/system_upgrade/el7toel8/libraries/requirearchver.py
```diff
@@ -0,0 +1,130 @@
...
+#     _require_minor_version('target', required, detected)
+
+
+def require_src_minor_version(required):
+    """
+    Stop actor's execution if requirement on minor version of source system is not met.
+
+    :param destination: 'source' or 'target'
+    :type destination: string
+    :param required: list or tuple of versions - integers or strings in the form '<integer>['+'|'>'|'-'|'<']'
```
pirat89: Is it what we talked about? I am not against this proposal, just If I remember well, we discussed semantic notation as it is used in SPEC format: `>, <, >=, <=`. But I am ok with that. .. Hmm. Additionally, from what you choosed, the `<` and `>` could have different meaning as `5>` means usually that everything *lower then five*. Because you expect that the whole equation is like `5 > minor_version`.

repos/system_upgrade/el7toel8/libraries/requirearchver.py
```diff
@@ -0,0 +1,130 @@
...
+#     :param detected: Detected version
+#     :type detected: int
+#     """
+#     detected = 0
+#     _require_minor_version('target', required, detected)
+
+
+def require_src_minor_version(required):
+    """
+    Stop actor's execution if requirement on minor version of source system is not met.
```
pirat89: I don't like this function. It is not doing what we expect to deliver. First, you target minor version only. But real code should operated with Major.Minor. So instead of 5, 6, 7... You should wirk with 7.5, 7.6, 7.7... 8.1, 8.3, 9.5....

Additionally, the logic is wrong. You are not stopping actor's execution. This is completely different story. It is expected to deliver functions that returns `True` or `False`. Because you need to process the actor maybe, but just don't want to process specific part of code or do various actions based on what upgrade path is processed.

vojtechsokol:
Closing in favor of #262.

PR #249, opened 2019-06-20
Add IPU workflow config actor

Rezney:
Also starts gathering LEAPP_DEVEL env vars.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/actor.py
@@ -1,6 +1,9 @@
+import platform
+
 from leapp.actors import Actor
-from leapp.libraries.actor.library import get_env_vars
+from leapp.libraries.actor import library
 from leapp.models import IPUConfig
+from leapp.reporting import Report
 from leapp.tags import IPUWorkflowTag
 
 
@@ -11,8 +14,11 @@ class IPUWorkflowConfig(Actor):
 
     name = 'ipu_workflow_config'
     consumes = ()
-    produces = (IPUConfig,)
+    produces = (IPUConfig, Report)
     tags = (IPUWorkflowTag,)
 
     def process(self):
-        self.produce(IPUConfig(leapp_env_vars=get_env_vars()))
+        self.produce(IPUConfig(
+            leapp_env_vars=library.get_env_vars(),
+            os_release=library.get_os_release('/etc/os-release'),
+            architecture=platform.machine()))
--- a/repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/actor.py
@@ -1,0 +1,18 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.library import get_env_vars
+from leapp.models import IPUConfig
+from leapp.tags import IPUWorkflowTag
+
+
+class IPUWorkflowConfig(Actor):
+    """
+    IPU workflow config actor
+    """
+
+    name = 'ipu_workflow_config'
+    consumes = ()
+    produces = (IPUConfig,)
+    tags = (IPUWorkflowTag,)
+
+    def process(self):
+        self.produce(IPUConfig(leapp_env_vars=get_env_vars()))
--- a/repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/libraries/library.py
@@ -1,6 +1,7 @@
 import os
 
-from leapp.models import EnvVar
+from leapp.libraries.common import reporting
+from leapp.models import EnvVar, OSRelease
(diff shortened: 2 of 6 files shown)


repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/libraries/library.py
```diff
@@ -0,0 +1,10 @@
+import os
+
+
+def get_env_vars():
+    """
+    Gather LEAPP_DEVEL environment variables and provide them as messages to be
+    available after reboot.
+    """
+    dev_vars = [(k, v) for (k, v) in os.environ.items() if k.startswith('LEAPP_DEVEL')]
```
vinzenz:
```suggestion
    return [(k, v) for (k, v) in os.environ.items() if k.startswith('LEAPP_DEVEL')]
```

repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/libraries/library.py
```diff
@@ -0,0 +1,10 @@
...
+    return dev_vars
```
vinzenz:
```suggestion
```

repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/actor.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+
+    name = 'ipu_workflow_config'
+    consumes = ()
+    produces = (IPUConfig,)
+    tags = (IPUWorkflowTag,)
+
+    def process(self):
+        dev_env_vars = get_env_vars()
+        self.produce(IPUConfig(leapp_env_vars=dev_env_vars))
```
vinzenz:
```suggestion
        self.produce(IPUConfig(leapp_env_vars=get_env_vars()))
```

repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/actor.py
```diff
@@ -0,0 +1,19 @@
...
+    IPU workflow config actor
+    """
+
+    name = 'ipu_workflow_config'
+    consumes = ()
+    produces = (IPUConfig,)
+    tags = (IPUWorkflowTag,)
+
+    def process(self):
+        dev_env_vars = get_env_vars()
```
vinzenz:
```suggestion
```

vinzenz:
Somehow I'd go with all LEAPP_ environment variables, not only LEAPP_DEVEL,

repos/system_upgrade/el7toel8/models/ipuconfig.py
```diff
@@ -0,0 +1,11 @@
...
+from leapp.topics import SystemInfoTopic
+
+
+class IPUConfig(Model):
+    """
+    IPU workflow configuration model
+    """
+    topic = SystemInfoTopic
+
+    leapp_env_vars = fields.List(fields.JSON(), default=[])
```
vinzenz: Why a list of JSON? This doesn't seem right. There are two things I can imagine:

`fields.List(fields.Model(EnvVar))`
or just `fields.JSON()`

Where EnvVar would be:

```python
class EnvVar(Model):
     topic = ...
     name = fields.String()
     value = fields.String()
```

vinzenz: and to be straight, I prefer the List(EnvVar) approach

Rezney:
@vinzenz I appreciate your checking this, however I just put this together fast (without much thinking) for testing.

I think it is good that people are (pre)checking also WIP PRs, so should we introduce something like "WIP-ignore"?

vinzenz:
I didn't do a real 'review' I commented on things I saw :-) - This helps you on not spending your time on things that you might have to redo later on anyway. ;-)

Rezney:
Ok, so we were both thinking about saving other side's time :) I will create a copr build manually next time then.

pirat89:
Depends on Leapp PR: https://github.com/oamg/leapp/pull/505

Rezney:
@vinzenz should I add ignore for "LEAPP_CURRENT_PHASE" and "LEAPP_CURRENT_ACTOR" or we do not care? Just asking it is minor thing.

vinzenz:
@Rezney i guess they should be ignored there are some more we would have to add if necessary- just check what’s there for now

Rezney:
@vinzenz There are these:

```
LEAPP_COMMON_TOOLS
LEAPP_CURRENT_ACTOR
LEAPP_HOSTNAME
LEAPP_VERBOSE
LEAPP_COMMON_FILES
LEAPP_EXECUTION_ID
LEAPP_CURRENT_PHASE
LEAPP_DEBUG
```

so we are ruling out "LEAPP_CURRENT_PHASE" and "LEAPP_CURRENT_ACTOR" for now.

Rezney:
Added also a test...

Rezney:
@vojtechsokol thanks for PR to this PR. I left a comment there.

Rezney:
@vojtechsokol 's PR merged into this one...

PR #252, opened 2019-06-26
packaging: get_latest_copr_build: tmp fix pagination

pirat89:
There is bug in copr (or copr cli module) for pagination. Currently
sorting doesn't work so sometimes we cannot find the build we want.
Removing the limitation temporary to ensure we always get the build,
until the issue in copr will be fixed.


vinzenz:
Let's wait for the build

PR #254, opened 2019-06-26
Add source/target version info to configuration actor

vojtechsokol:


repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/actor.py
```diff
@@ -18,7 +18,11 @@ class IPUWorkflowConfig(Actor):
     tags = (IPUWorkflowTag,)
 
     def process(self):
+        target_version = '8.0'
```
jmikovic: I don't like this hard-coded version here.
If we ever want to support more than 1 target version this will be a problem.

pirat89: I expect that this will be changed in future. Currently it doesn't matter as this is blackbox for other actors and it works for us now.

pirat89: To be more clear, there will be always just one target version during the runtime which will be specified probably by leapp tool itself.

PR #255, opened 2019-06-26
Ignore also "LEAPP_DEBUG" and "LEAPP_VERBOSE" env vars

Rezney:
We want "LEAPP_DEBUG and LEAPP_VERBOSE" env vars to be checked
by our "is_debug()" and "is_verbose()" functions.


Rezney:
@pirat89 here you go...

Issue #239, opened 2019-06-10
network_manager_update: the tool script fails


pirat89:
**Actual behavior**

The network_manager_update actor is failing:
```
1961|log-message|2019-06-10T13:21:23.877040Z|0ebe6e50-0257-4743-88c0-c8512c1e5678|121||{"level": "INFO", "message": "2019-06-10 13:21:23.877 INFO     PID: 1335 leapp.workflow.FirstBoot: Executing actor network_manager_update_connections "}
1962|log-message|2019-06-10T13:21:23.986432Z|0ebe6e50-0257-4743-88c0-c8512c1e5678|123||{"level": "DEBUG", "message": "2019-06-10 13:21:23.986 DEBUG    PID: 2199 leapp.workflow.FirstBoot.network_manager_update_connections: External command is started: [/usr/bin/python3 tools/nm-update-client-ids.py]"}
1963|process-start|2019-06-10T13:21:23.992039Z|0ebe6e50-0257-4743-88c0-c8512c1e5678|123||{"id": "22a6979b-43b3-4c62-a120-346c8c8cf97b", "parameters": ["/usr/bin/python3", "tools/nm-update-client-ids.py"], "env": null}
1964|process-end|2019-06-10T13:21:24.284408Z|0ebe6e50-0257-4743-88c0-c8512c1e5678|123||22a6979b-43b3-4c62-a120-346c8c8cf97b
1965|process-result|2019-06-10T13:21:24.286618Z|0ebe6e50-0257-4743-88c0-c8512c1e5678|123||{"id": "22a6979b-43b3-4c62-a120-346c8c8cf97b", "parameters": ["/usr/bin/python3", "tools/nm-update-client-ids.py"], "result": {"signal": 0, "exit_code": 1, "pid": 2204, "stdout": "", "stderr": "Traceback (most recent call last):\n  File \"tools/nm-update-client-ids.py\", line 20, in <module>\n    client = NM.Client.new(None)\nGLib.Error: g-io-error-quark: Could not connect: No such file or directory (1)\n"}, "env": null}
1966|log-message|2019-06-10T13:21:24.288657Z|0ebe6e50-0257-4743-88c0-c8512c1e5678|123||{"level": "DEBUG", "message": "2019-06-10 13:21:24.288 DEBUG    PID: 2199 leapp.workflow.FirstBoot.network_manager_update_connections: External command is finished: [/usr/bin/python3 tools/nm-update-client-ids.py]"}
[...]

PR #261, opened 2019-06-27
createresumeservice: add deps to prevent race condition

pirat89:
createresumeservice: add deps to prevent race condition (#239)
    
    Actor taking care about NetworkManager fails from time to time during
    the FirstBoot phase probably because of race condition during the
    boot of the system. To prevent this problems, exec the resume service
    after the network-online target. Additionally to be really sure we
    do not have hidden troubles, put there dbus.service as well.
    
    Fixes: #239

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/createresumeservice/files/leapp_resume.service
+++ b/repos/system_upgrade/el7toel8/actors/createresumeservice/files/leapp_resume.service
@@ -2,6 +2,9 @@
 Description=Temporary Leapp service which resumes execution after reboot
 After=default.target
 DefaultDependencies=no
+After=dbus.service
+After=network-online.target
+Wants=network-online.target
 
 [Service]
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/el7toel8/actors/createresumeservice/files/leapp_resume.service
```diff
@@ -2,6 +2,8 @@
 Description=Temporary Leapp service which resumes execution after reboot
 After=default.target
 DefaultDependencies=no
+After=NetworkManager.service
+After=dbus.service
```
vinzenz:
```suggestion
After=dbus.service
After=network-online.target
Wants=network-online.target
```

pirat89: sovled

PR #264, opened 2019-07-01
Add support for linting and testing of repository libraries

vojtechsokol:
This PR allows for testing and linting of repository libraries. When environment variable `TEST_LIBS` is used for Makefile targets `lint`, `test` and `test_no_lint`, linting and tests are carried out for repository libraries. When no environment variable is specified (neither `ACTOR` nor `TEST_LIBS`), linting and testing is carried out for the whole repo:

- `make test` -> whole repo
- `TEST_LIBS=y make test` -> repository libraries only
- `ACTOR=someactor make test` -> specified actor only
- `ACTOR=someactor TEST_LIBS=y make test` -> specified actor and repository libraries

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,19 @@ DEPS_PKGNAME=leapp-el7toel8-deps
 VERSION=`grep -m1 "^Version:" packaging/$(PKGNAME).spec | grep -om1 "[0-9].[0-9.]**"`
 DEPS_VERSION=`grep -m1 "^Version:" packaging/$(DEPS_PKGNAME).spec | grep -om1 "[0-9].[0-9.]**"`
 ACTOR_PATH=repos
+LIBRARY_PATH=`python utils/library_path.py`
 ifdef ACTOR
 	ACTOR_PATH=`python utils/actor_path.py $(ACTOR)`
+	TEST_ACTOR_ARG=--actor=$(ACTOR)
+	ifneq ($(TEST_LIBS),y)
+		LIBRARY_PATH=
+	endif
+endif
+ifeq ($(TEST_LIBS),y)
+	TEST_LIBRARIES_ARG=--libraries
+	ifndef ACTOR
+		ACTOR_PATH=
+	endif
 endif
 # needed only in case the Python2 should be used
 _USE_PYTHON_INTERPRETER=$${_PYTHON_INTERPRETER}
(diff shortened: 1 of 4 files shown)


conftest.py
```diff
@@ -1,26 +1,40 @@
...
+        session.actor_context.__enter__()
+    elif library_path:
+        repo = find_and_scan_repositories(find_repository_basedir(library_path), include_locals=True)
+        repo.load()
+        # TODO: How to change directory to subpackage?
+        #       Consider that library_path directory contains package 'subpackage'
+        #       with tests in 'subpackage/tests' and that certain test needs
+        #       mocked file 'subpackage/tests/files/dummy'
+        #       When directory is changed to library_path, providing test with
+        #       the dummy file is only possible vit using 'full' path to the dummy
```
vinzenz: I have no idea what you are saying here. I am trying to guess what it should have been:

```python
        # TODO: How to change the directory to a subpackage?
        #       Assuming that the library_path directory contains a package 'subpackage'
        #       with tests in 'subpackage/tests' and that certain tests need the mocked file
        #       'subpackage/tests/files/dummy'
        #       When the directory is changed to library_path, providing the test with the dummy
        #       file, is only possible by using the 'full' path to the dummy file -
        #       'subpackage/tests/files/dummy' instead of the 'relative' path 'tests/files/dummy'
```

vinzenz: If this is what you meant, the answer is, that the test file should take care of injection of such things:
``` python
os.path.join(os.path.dirname(__file__), 'files', 'dummy')
```
And if it is about making it like it would be a repo/files then it would have to set / extend the environment variable `LEAPP_COMMON_FILES`

```python
# override an existing file in the repo with a mocked one:
override_path = os.path.join(os.path.dirname(__file__), 'files')
os.environ['LEAPP_COMMON_FILES'] =  ':'.join([override_path] + os.getenv('LEAPP_COMMON_FILES', '').split(':'))
```

Makefile
```diff
@@ -4,8 +4,19 @@ DEPS_PKGNAME=leapp-el7toel8-deps
...
 ACTOR_PATH=repos
+LIBRARY_PATH=`python utils/library_path.py $(ACTOR)`
 ifdef ACTOR
 	ACTOR_PATH=`python utils/actor_path.py $(ACTOR)`
+	TEST_ACTOR_ARG=--actor=$(ACTOR)
+	ifndef TEST_LIBS
+		LIBRARY_PATH=
+	endif
+endif
+ifdef TEST_LIBS
```
jmikovic: I think it would be better if we had an expected value for TEST_LIBS because now you can write `TEST_LIBS=false make test` and it will only test libraries, which is confusing

utils/library_path.py
```diff
@@ -0,0 +1,15 @@
+import logging
```
vinzenz:
```suggestion
from __future__ import print_function

import logging
```

conftest.py
vojtechsokol: I updated the RTD section of repository layout with this info, see https://github.com/oamg/leapp/pull/532.

PR #265, opened 2019-07-02
dnfplugin: Make repos also obey gpgcheck = False

vinzenz:
In case of upgrades, we usually set the gpgcheck to false. However the
repositories have their own values which need to be explicitly
overridden by the plugin too, if we don't want the gpgcheck to be
performed.
This patch ensures, that the gpgcheck value is set to false explicitly
if we are instructed to set the gpgcheck=False through the plugin
config.

### DIFF
--- a/repos/system_upgrade/el7toel8/files/rhel_upgrade.py
+++ b/repos/system_upgrade/el7toel8/files/rhel_upgrade.py
@@ -62,6 +62,8 @@ class RhelUpgradeCommand(dnf.cli.Command):
         for repo in self.base.repos.all():
             if repo.id in enabled_repos:
                 repo.skip_if_unavailable = False
+                if not self.base.conf.gpgcheck:
+                    repo.gpgcheck = False
                 repo.enable()
 
     def run(self):


repos/system_upgrade/el7toel8/files/rhel_upgrade.py
```diff
@@ -62,6 +62,8 @@ def configure(self):
         for repo in self.base.repos.all():
             if repo.id in enabled_repos:
                 repo.skip_if_unavailable = False
+                if not self.base.conf.gpgcheck:
+                    repo.gpgcheck = False
```
bocekm: Why not simply `repo.gpgcheck = self.base.conf.gpgcheck`?

vinzenz: Because I don't want it to be always the same, I just want to disable it, if it is enabled. And I want to make it explicit in the code.

bocekm: Ok. So if the dnf plugin data tell to *do* gpg checking, we'll leave it up to the setup of each repo. Which is the default behaviour of dnf I guess. Fair point.

bocekm:
Let's wait for the verification from @AloisMahdal and then it's good to be merged.

AloisMahdal:
Tested with *leapp-repository-0.7.0-0.201907021018Z.98e771c.fix_gpgcheck_errors.PR265.el7_6* and passed.

PR #267, opened 2019-07-02
Do not call set_release if no release value

vinzenz:
In case of skipping RHSM (LEAPP_DEVEL_SKIP_RHSM) no target release is
set and we should not try to set it.


pirat89:
Technically, it's morning.

PR #268, opened 2019-07-03
pylint: Extend to test for Python3 compatibility

vinzenz:
Previously we only verified with Python2 compatibility, however our
codebase needs to be running also on Python 3, therefore we will enable
the compatibility checks for Python 3.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,3 +1,6 @@
+[MASTER]
+py3k=
+
 [MESSAGES CONTROL]
 disable=
 # "F" Fatal errors that prevent further processing
@@ -8,6 +11,7 @@ disable=
  no-name-in-module,
  raising-bad-type,
 # "W" Warnings for stylistic problems or minor programming issues
+ no-absolute-import,
  arguments-differ,
  cell-var-from-loop,
  fixme,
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -178,7 +178,7 @@ lint:
 	SEARCH_PATH={$(ACTOR_PATH),$(LIBRARY_PATH)} && \
 	echo "Using search path '$${SEARCH_PATH}'" && \
 	echo "--- Running pylint ---" && \
-	bash -c "[[ ! -z $${SEARCH_PATH} ]] && find $${SEARCH_PATH} -name '*.py' | sort -u | xargs pylint" && \
+	bash -c "[[ ! -z $${SEARCH_PATH} ]] && find $${SEARCH_PATH} -name '*.py' | sort -u | xargs pylint --py3k --disable=no-absolute-import" && \
```
fernflower: Maybe add no-absolute-import to .pylintrc?

vinzenz: Done, even py3k can be in pylintrc :-)

fernflower:
Cool, `import ConfigParser` issued 
```************* Module abc
[W1648 bad-python3-import] Module moved in Python 3 File: repos/system_upgrade/el7toel8/abc.py, line 1, in ```

PR #269, opened 2019-07-03
Warn about removed pam_tally2 module

Jakuje:
The RHEL 8 does not come with `pam_tally2` module anymore and if the pam stacks of some services used to be configured with this module, they can very simply prevent login on the upgraded systems.

This patch set is using the PAM library from #102 (not merged at the time of writing) to generate `PamConfiguration` messages, which are then consumed by another actor, verifying that none of the services depend on this removed pam module. It is written in a way that it can check for more modules, but we did not identify any other module (that would not be handled by #102) that was removed.

I will also have to figure out the final wording of the reports before merging, whether the removal is documented somewhere and whether we have some migration steps.

~~**Edit:** I cherry-picked part of the #102 which my pull request is using to be able to move on, since the original PR is blocked on the dialogs.~~

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1713645

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkremovedpammodules/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkremovedpammodules/actor.py
@@ -1,0 +1,60 @@
+from leapp.actors import Actor
+from leapp.exceptions import StopActorExecutionError
+from leapp.models import Report, PamConfiguration
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+from leapp.libraries.stdlib import api
+from leapp.libraries.common.reporting import report_with_remediation
+
+
+class CheckRemovedPamModules(Actor):
+    """
+    Check for modules that are not available in RHEL 8 anymore
+
+    At this moment, we check only for pam_tally2. Few more modules
+    are alredy covered in RemoveOldPAMModulesApply actor
+    """
+
+    name = 'removed_pam_modules'
+    consumes = (PamConfiguration, )
+    produces = (Report, )
+    tags = (ChecksPhaseTag, IPUWorkflowTag, )
+
+    def process(self):
+        messages = self.consume(PamConfiguration)
+        config = next(messages, None)
+        if list(messages):
+            api.current_logger().warning('Unexpectedly received more than one PamConfiguration message.')
+        if not config:
+            raise StopActorExecutionError(
+                'Could not check pam configuration', details={'details': 'No PamConfiguration facts found.'}
+            )
+
+        # This list contain tupples of removed modules and their recommended replacements
+        removed_modules = [
+            ('pam_tally2', 'pam_faillock'),
+            ]
+        found_services = set()
+        found_modules = set()
+        replacements = set()
+        for service in config.services:
+            for module in removed_modules:
+                removed = module[0]
+                replacement = module[1]
+                if removed in service.modules:
+                    found_services.add(service.service)
+                    found_modules.add(removed)
+                    replacements.add(replacement)
+
+        if found_modules:
+            report_with_remediation(
+                title='The {} pam module(s) no longer available'.format(', '.join(found_modules)),
+                summary='The services {} using PAM are configured to '
+                        'use {} module(s), which is no longer available '
+                        'in Red Hat Enterprise Linux 8.'.format(
+                        ', '.join(found_services), ', '.join(found_modules)),
+                remediation='If you depend on its functionality, it is '
+                            'recommended to migrate to {}. Otherwise '
+                            'please remove the pam module(s) from all the files '
+                            'under /etc/pam.d/.'.format(', '.join(replacements)),
+                severity='high',
+                flags=['inhibitor'])
--- a/repos/system_upgrade/el7toel8/actors/pammodulesscanner/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/pammodulesscanner/actor.py
@@ -1,0 +1,43 @@
+import os
+
+from leapp.actors import Actor
+from leapp.models import PamConfiguration, PamService
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+from leapp.libraries.stdlib import api
+from leapp.libraries.common.pam import PAM
+
+
+class PamModulesScanner(Actor):
+    """
+    Scan the pam directory for services and modules used in them
+
+    This produces a PAMConfiguration message containing the whole
+    list of configured PAM services and what modules they contain.
+    """
+
+    name = 'pam_modules_scanner'
+    consumes = ()
+    produces = (PamConfiguration, )
+    tags = (FactsPhaseTag, IPUWorkflowTag, )
+
+    def process(self):
+        conf = []
+        path = "/etc/pam.d/"
+        for f in os.listdir(path):
+            pam_file = os.path.join(path, f)
+            # Ignore symlinks (usually handled by authconfig)
+            if not os.path.isfile(pam_file) or os.path.islink(pam_file):
+                continue
+
+            # Use the existing PAM library to parse the files, but unpack it to our model
+            try:
+                content = PAM.read_file(pam_file)
+                modules = PAM(content)
+                service = PamService(service=f, modules=modules.modules)
+                conf.append(service)
+            except OSError as err:
+                # if leapp can not read that file it will probably not be important
+                api.current_logger().warning('Failed to read file {}: {}'.format(pam_file, err.strerror))
+
+        pam = PamConfiguration(services=conf)
+        self.produce(pam)
--- a/repos/system_upgrade/el7toel8/actors/tcpwrapperscheck/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/tcpwrapperscheck/actor.py
@@ -1,4 +1,5 @@
 from leapp.actors import Actor
+from leapp.exceptions import StopActorExecutionError
 from leapp.models import Report, TcpWrappersFacts, InstalledRedHatSignedRPM
 from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
 from leapp.libraries.stdlib import api
@@ -44,9 +45,13 @@ class TcpWrappersCheck(Actor):
     def process(self):
         # Consume a single TCP Wrappers message
         tcp_wrappers_messages = self.consume(TcpWrappersFacts)
-        tcp_wrappers_facts = next(tcp_wrappers_messages)
+        tcp_wrappers_facts = next(tcp_wrappers_messages, None)
         if list(tcp_wrappers_messages):
             api.current_logger().warning('Unexpectedly received more than one TcpWrappersFacts message.')
+        if not tcp_wrappers_facts:
+            raise StopActorExecutionError(
+                'Could not check tcp wrappers configuration', details={'details': 'No TcpWrappersFacts found.'}
+            )
 
         # Convert installed packages message to list
         packages = create_lookup(InstalledRedHatSignedRPM, field='items', key='name')
--- a/repos/system_upgrade/el7toel8/actors/opensshalgorithmscheck/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/opensshalgorithmscheck/actor.py
@@ -1,4 +1,5 @@
 from leapp.actors import Actor
+from leapp.exceptions import StopActorExecutionError
 from leapp.models import Report, OpenSshConfig
 from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
 from leapp.libraries.stdlib import api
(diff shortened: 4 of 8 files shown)


repos/system_upgrade/el7toel8/actors/pammodulesscanner/actor.py
```diff
@@ -0,0 +1,37 @@
+from leapp.actors import Actor
+from leapp.models import PamConfiguration, PamService
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+from leapp.libraries.common.pam import PAM
+
+from os import listdir
+from os.path import isfile, islink, join
```
vinzenz: please prefer to import only os
```suggestion
import os
```

repos/system_upgrade/el7toel8/actors/pammodulesscanner/actor.py
```diff
@@ -0,0 +1,37 @@
+from leapp.actors import Actor
+from leapp.models import PamConfiguration, PamService
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+from leapp.libraries.common.pam import PAM
+
+from os import listdir
```
vinzenz:
```suggestion
```

repos/system_upgrade/el7toel8/actors/pammodulesscanner/actor.py
```diff
@@ -0,0 +1,37 @@
...
+
+    name = 'pam_modules_scanner'
+    consumes = ()
+    produces = (PamConfiguration, )
+    tags = (FactsPhaseTag, IPUWorkflowTag, )
+
+    def process(self):
+        conf = []
+        path = "/etc/pam.d/"
+        for f in listdir(path):
```
vinzenz:
```suggestion
        for f in os.listdir(path):
```

repos/system_upgrade/el7toel8/actors/pammodulesscanner/actor.py
```diff
@@ -0,0 +1,37 @@
...
+            file = join(path, f)
```
vinzenz:
```suggestion
            svc_file = os.path.join(path, f)
```

repos/system_upgrade/el7toel8/actors/pammodulesscanner/actor.py
```diff
@@ -0,0 +1,37 @@
...
+            # Ignore symlinks (usually handled by authconfig)
+            if not isfile(file) or islink(file):
```
vinzenz:
```suggestion
            if not os.path.isfile(svc_file) or os.path.islink(svc_file):
```

repos/system_upgrade/el7toel8/actors/pammodulesscanner/actor.py
```diff
@@ -0,0 +1,37 @@
...
+                continue
+
+            # Use the existing PAM library to parse the files, but unpack it to our model
+            content = PAM.read_file(file)
```
vinzenz:
```suggestion
            content = PAM.read_file(svc_file)
```

Jakuje:
@vinzenz Thank you for the review. I addressed your current comments. The CI is still expected to fail and I will check what unification can be done after the #102 will be final and merged.

Jakuje:
@vinzenz it looks like the #102 is stalled and it still needs some questions to be answered about the interactivity of the update so I pulled the library part of it so we can at least test this one. For the Travis CI failure, I am not particularly sure where does the failure come from since I do not see this name anywhere in the code:
```
--- Linting ... ---
  File "utils/library_path.py", line 13, in <module>
    repos.load()
  File "/payload/tut/lib/python2.7/site-packages/leapp/repository/manager.py", line 103, in load
    resolve_model_references()
  File "/payload/tut/lib/python2.7/site-packages/leapp/models/__init__.py", line 192, in resolve_model_references
    reference.resolve()
  File "/payload/tut/lib/python2.7/site-packages/leapp/models/__init__.py", line 179, in resolve
    raise ModelDefinitionError('Undefined Model "{}"'.format(cls._referenced))
leapp.exceptions.ModelDefinitionError: Undefined Model "RemovedPAMModulesReference"
make: *** [lint] Error 1```
There is no

File: repos/system_upgrade/el7toel8/actors/removedpammodules/actor.py [hunk 7]
```diff
@@ -0,0 +1,48 @@
+from leapp.actors import Actor
+from leapp.models import Report, PamConfiguration
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+from leapp.libraries.stdlib import api
+from leapp.libraries.common.reporting import report_with_remediation
+
+
+class RemovedPamModules(Actor):
```
vinzenz: Please call this actor CheckRemovedPamModules

vinzenz:
> @vinzenz it looks like the #102 is stalled and it still needs some questions to be answered about the interactivity of the update so I pulled the library part of it so we can at least test this one. For the Travis CI failure, I am not particularly sure where does the failure come from since I do not see this name anywhere in the code:
> 
> ```
> --- Linting ... ---
> Traceback (most recent call last):
>   File "utils/library_path.py", line 13, in <module>
>     repos.load()
>   File "/payload/tut/lib/python2.7/site-packages/leapp/repository/manager.py", line 103, in load
>     resolve_model_references()
>   File "/payload/tut/lib/python2.7/site-packages/leapp/models/__init__.py", line 192, in resolve_model_references
>     reference.resolve()
>   File "/payload/tut/lib/python2.7/site-packages/leapp/models/__init__.py", line 179, in resolve
>     raise ModelDefinitionError('Undefined Model "{}"'.format(cls._referenced))
> leapp.exceptions.ModelDefinitionError: Undefined Model "RemovedPAMModulesReference"
> make: *** [lint] Error 1```
> ```

You probably forgot to commit the model, because I cannot see a RemovePAMModules model in your PR :-)

repos/system_upgrade/el7toel8/actors/removedpammodules/actor.py
vinzenz: And please also update the path to be called checkremovedpammodules

Jakuje: Will do.

Jakuje:
I actually left there one of the actors from the original commit with the library which should not be there.

Now, it looks better. I also updated the text in the report so this should be quite final version, I think

vinzenz:
Sorry for leaving you hanging in there, I wanted to see what we can do - We will make the existing actors with the dialogs Experimental for now, and the library will go in. As the original PR 102 was hanging in there for so long, we decided to give it a priority and i'd like you to rebase on top of master once it is in. Sorry for the inconvenience and thanks for your cooperation!

Jakuje:
Rebased on top of current master after merging of #102.

Jakuje:
@vinzenz Ping. Any update here? Is there something needed from me now?

repos/system_upgrade/el7toel8/actors/checkremovedpammodules/actor.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+
+    name = 'removed_pam_modules'
+    consumes = (PamConfiguration, )
+    produces = (Report, )
+    tags = (ChecksPhaseTag, IPUWorkflowTag, )
+
+    def process(self):
+        messages = self.consume(PamConfiguration)
+        config = next(messages)
```
vinzenz: you need to handle StopIteration by next(messages, None) and then ofc checking whether you got a message or not. (if not config:...)

otherwise the PR is fine - Sorry for the delay though

Jakuje: Seems like this issue is in all the actors that I wrote so far and also in many others. In this case, assuming PAM is mandatory in RHEL, we can require to have exactly one message of this type.

I fixed this one in "fixup" commit to be squashed later. I will have a look into the other occurrences to increase the robustness of the actors.

Jakuje: @vinzenz ping?

Jakuje:
ping. Any update here? Is there something else needed for this PR?

bocekm:
Hi @Jakuje, I'll take a look at your PR tomorrow.

repos/system_upgrade/el7toel8/actors/checkremovedpammodules/actor.py
```diff
@@ -0,0 +1,60 @@
...
+    tags = (ChecksPhaseTag, IPUWorkflowTag, )
+
+    def process(self):
+        messages = self.consume(PamConfiguration)
+        config = next(messages, None)
+        if list(messages):
+            api.current_logger().warning('Unexpectedly received more than one PamConfiguration message.')
+        if not config:
+            raise StopActorExecutionError(
+                'Could check pam configuration', details={'details': 'No PamConfiguration facts found.'}
```
bocekm:
```suggestion
                'Couldn't check pam configuration', details={'details': 'No PamConfiguration facts found.'}
```

repos/system_upgrade/el7toel8/actors/checkremovedpammodules/actor.py
```diff
@@ -0,0 +1,60 @@
...
+                if removed in service.modules:
+                    found_services.add(service.service)
+                    found_modules.add(removed)
+                    replacements.add(replacement)
+
+        if found_modules:
+            report_with_remediation(
+                title='The {} pam module(s) no longer available'.format(', '.join(found_modules)),
+                summary='The services {} using PAM are configured to '
+                        'use {} module(s), which is no longer available '
```
bocekm:
```suggestion
                        'use {} module(s), which is/are no longer available '
```

repos/system_upgrade/el7toel8/actors/tcpwrapperscheck/actor.py
```diff
@@ -44,9 +45,13 @@ class TcpWrappersCheck(Actor):
     def process(self):
         # Consume a single TCP Wrappers message
         tcp_wrappers_messages = self.consume(TcpWrappersFacts)
-        tcp_wrappers_facts = next(tcp_wrappers_messages)
+        tcp_wrappers_facts = next(tcp_wrappers_messages, None)
         if list(tcp_wrappers_messages):
             api.current_logger().warning('Unexpectedly received more than one TcpWrappersFacts message.')
+        if not tcp_wrappers_facts:
+            raise StopActorExecutionError(
+                'Could check tcp wrappers configuration', details={'details': 'No TcpWrappersFacts found.'}
```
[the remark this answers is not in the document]
Jakuje: The tcp_wrappers files are provided by the setup package and they should be available on every installation of RHEL7. The tcp_wrappers scanner generates this message every time.

OpenSSH does not have to be installed on all the systems, but I think the OpenSSH config scanner is made the way it returns the object even if the file is not in place and openssh not installed. I think this message does not trigger any blocking leapp reports. 

The pam is part of the system authentication so I think it is not possible to remove it and not use it. The pam scanner always returns some messages listing the pam modules used, which always make sense to scan.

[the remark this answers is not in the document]
bocekm: If all these three things are mandatory to be set (PAM and OpenSSH configs, TCP wrappers)  on every RHEL 7, then the upgrade should be stopped in the Checks phase, not the FactsCollection phase. And in the checks phase should create a report message for the user together with some remediation recommendation (how to proceed to be able to continue with the upgrade).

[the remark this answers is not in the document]
Jakuje: Indeed, if there is simple way to test it, it would make sense to verify that automatically that the messages are sent even if some of the files and packages are missing.

bocekm:
Also please rebase to the latest master so that the builds created are installable. They are not now.

Jakuje:
For the unit tests, this PR adds two actors (and few bugfixes around as I am learning around):
 * `PamModulesScanner`: Trivial actor using already tested `PAM` library (just packing the output to leapp message)
 * `CheckRemovedPamModules`: Trivial check for the specific module in the leapp message lists.

If you insist on writing unit tests for this, I can do that, but it will not be just now.

I rebased on current master just now.

bocekm:
Regarding the tests, I do not insist. It's up to you, even though it would be great if you write them. Otherwise we can write some basic unit tests, as we know this PR has been opened quite long and you are probably eager to have it merged.

repos/system_upgrade/el7toel8/actors/tcpwrapperscheck/actor.py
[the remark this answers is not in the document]
Jakuje: Not the files, but the messages are "mandatory" -- will always be sent by the leapp scanners. You can have working system without openssh and without tcp_wrappers files (but probably not without pam configuration). But that does not have anything to do with the upgrades, neither it should stop the update. The above are just sanity checks as pointed out in previous reviews by @vinzenz .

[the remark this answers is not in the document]
bocekm: I see, something like internal error.

bocekm:
Successfully tested with snactor. Wll trigger end to end tests execution by the next comment.

bocekm:
e2e tests

bocekm:
The e2e failure is an internal problem of our CI but I've verified it's actually passing. So I'm merging the PR.

PR #270, opened 2019-07-05
Add ip6tables to firewall facts

holser:
ip6tables is part of iptables so it should be treated respectfully.


repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -267,4 +267,5 @@ def _get_firewall_status(service_name):
     return FirewallsFacts(
         firewalld=_get_firewall_status('firewalld'),
         iptables=_get_firewall_status('iptables'),
+        ip6tables=_get_firewall_status('ip6tables'),
```
vinzenz: You will have to update the model to make this work

holser: Thanks a lot. Totally forgot about model

Issue #263, opened 2019-06-29
Actor prepare_upgrade_transaction unexpectedly terminated


bmarjan:
```
[root@linux ~]# leapp upgrade
Process Process-62:
Traceback (most recent call last):
  File "/usr/lib64/python2.7/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/lib64/python2.7/multiprocessing/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 59, in _do_run
    target_actor(logger=logger, messaging=messaging).run(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/leapp/actors/__init__.py", line 299, in run
    self.process(*args)
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/storagescanner/actor.py", line 21, in process
    self.produce(library.get_storage_info())
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py", line 200, in get_storage_info
    fstab=_get_fstab_info('/etc/fstab'),
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py", line 15, in inner
    return list(f(*args, **kwargs))
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py", line 90, in _get_fstab_info
    fs_spec, fs_file, fs_vfstype, fs_mntops, fs_freq, fs_passno = entry.split()
ValueError: need more than 4 values to unpack

================================================================================================
Actor storage_scanner unexpectedly terminated with exit code: 1 - Please check the above details
================================================================================================
```

To Reproduce
Steps to reproduce the behavior

    install 'leapp'
    run 'leapp upgrade'
    See error

Expected behavior
leapp upgrade successfully

System information (please complete the following information):
```
[root@linux ~]# cat /etc/system-release
Red Hat Enterprise Linux Workstation release 7.6 (Maipo)
```

EDIT(pirat89): fixed formatting to make it readable

pirat89:
@bmarjan Please add the rest of required information that were mentioned in the bug template:
  - version of packages (or used commit in case you applied something manually)
  - missing setup of the system

[...]

PR #271, opened 2019-07-05
Add support for fstab entries that do not have optional fields

asilveir:
- the last two fstab fields are optional and may be missing.
- the default values were added based on the fstab man page.
- update unit tests.
- add warning for comments at end of valid lines.

This PR aims to fix issue #263

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py
@@ -5,6 +5,7 @@ import functools
 
 from leapp.models import StorageInfo, PartitionEntry, FstabEntry, MountEntry, LsblkEntry, \
     PvsEntry, VgsEntry, LvdisplayEntry, SystemdMountEntry
+from leapp.libraries.common import reporting
 from leapp.libraries.stdlib import api
 
 
@@ -79,7 +80,7 @@ def _get_fstab_info(fstab_path):
     ''' Collect storage info from /etc/fstab file '''
     if _is_file_readable(fstab_path):
         with open(fstab_path, 'r') as fstab:
-            for entry in fstab:
+            for line, entry in enumerate(fstab, 1):
                 if entry.startswith('#'):
                     continue
 
@@ -87,7 +88,43 @@ def _get_fstab_info(fstab_path):
                 if not entry:
                     continue
 
-                fs_spec, fs_file, fs_vfstype, fs_mntops, fs_freq, fs_passno = entry.split()
+                entries = entry.split()
+
+                if len(entries) == 4:
+                    entries.append('0')
+
+                if len(entries) == 5:
+                    entries.append('0')
+
+                if len(entries) != 6:
+                    if any(value.startswith('#') for value in entries):
+                        remediation = (
+                            'Comments in the /etc/fstab file must be at the beginning of the line, your file has a'
+                            ' comment at the end of the line at line {}, please edit and fix this, for further'
+                            ' information read fstab man page (man 5 fstab).'.format(line)
+                        )
+                    else:
+                        remediation = (
+                            'The /etc/fstab file must have at least 4 values and at most 6 per line, your file on the'
+                            ' line: {} have {} values, please edit and fix this, for further information read'
+                            ' fstab man page (man 5 fstab). '.format(line, len(entries))
+                        )
+                    summary = (
+                        'The fstab configuration file seems to be invalid. You need to fix it to be able to proceed'
+                        ' with the upgrade process.'
+                    )
+                    reporting.report_with_remediation(
+                        title='Problems with parsing data in /etc/fstab',
+                        summary=summary,
+                        remediation=remediation,
+                        severity='high',
+                        flags=['inhibitor']
+                    )
+
+                    api.current_logger().error(summary)
+                    break
+
+                fs_spec, fs_file, fs_vfstype, fs_mntops, fs_freq, fs_passno = entries
                 yield FstabEntry(
                     fs_spec=fs_spec,
                     fs_file=fs_file,
--- a/repos/system_upgrade/el7toel8/actors/storagescanner/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/storagescanner/actor.py
@@ -1,6 +1,7 @@
 from leapp.actors import Actor
 from leapp.libraries.actor import library
 from leapp.models import StorageInfo
+from leapp.reporting import Report
 from leapp.tags import IPUWorkflowTag, FactsPhaseTag
 
 
@@ -14,8 +15,8 @@ class StorageScanner(Actor):
 
     name = 'storage_scanner'
     consumes = ()
-    produces = (StorageInfo,)
-    tags = (IPUWorkflowTag, FactsPhaseTag,)
+    produces = (Report, StorageInfo)
+    tags = (IPUWorkflowTag, FactsPhaseTag)
 
     def process(self):
         self.produce(library.get_storage_info())
--- a/repos/system_upgrade/el7toel8/actors/storagescanner/tests/files/fstab
+++ b/repos/system_upgrade/el7toel8/actors/storagescanner/tests/files/fstab
@@ -7,5 +7,8 @@
 # See man pages fstab(5), findfs(8), mount(8) and/or blkid(8) for more info
 #
 /dev/mapper/rhel_ibm--p8--kvm--03--guest--02-root /                       xfs     defaults        0 0
-UUID=0a5215ef-1fb4-4b1b-8860-be4baa9e624c /boot                   xfs     defaults        0 0
+UUID=0a5215ef-1fb4-4b1b-8860-be4baa9e624c /boot                   xfs     defaults        0 1
+UUID=acf9f525-3691-429f-96d7-3f8530227062 /var                   xfs     defaults
+# Testing comment at the middle of valid lines
+UUID=d74186c9-21d5-4549-ae26-91ca9ed36f56 /tmp                   ext4     defaults,nodev,nosuid,noexec        1
 /dev/mapper/rhel_ibm--p8--kvm--03--guest--02-swap swap                    swap    defaults        0 0
--- a/repos/system_upgrade/el7toel8/actors/storagescanner/tests/files/invalid_fstab
+++ b/repos/system_upgrade/el7toel8/actors/storagescanner/tests/files/invalid_fstab
@@ -1,0 +1,11 @@
+
+#
+# /etc/fstab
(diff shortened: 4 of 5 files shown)


repos/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py
```diff
@@ -87,14 +87,23 @@ def _get_fstab_info(fstab_path):
...
-                    fs_passno=fs_passno)
+                entries = entry.split()
+
+                if len(entries) == 4:
+                    entries.append('0')
+
+                if len(entries) == 5:
+                    entries.append('0')
+
+                if len(entries) == 6:
```
pirat89: The logic should be different. If you end up with anything else then 6 entries, you should:
- log error & raise stopactorexecutionerror
- or log error and  inhibit upgrade with a msg that fstab seems to be wrong and user should check it. Would be even nice to print the line (or anything that would point what line is problematic at all) so msg will not be confusing so much and user could fix it (or we will be able to investigate it better)

pirat89:
@asilveir pls add reference to the issue #263 in the commit message and PR description so it will be closed automatically with merge.

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py
examon: I agree with @pirat89 . It would be nice to log error and inhibit upgrade in case there is a line in fstab with anything else then 6 entries.

asilveir:
I noticed that we are not treating comments at the end of the line I hope I have fixed this too, please @examon and @pirat89 review this too.

I already added a test case for this scenario.

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py
```diff
@@ -79,22 +80,32 @@ def _get_fstab_info(fstab_path):
...
         with open(fstab_path, 'r') as fstab:
-            for entry in fstab:
+            keys = ['fs_spec', 'fs_file', 'fs_vfstype', 'fs_mntops', 'fs_freq', 'fs_passno']
+            for line, entry in enumerate(fstab, 1):
                 if entry.startswith('#'):
                     continue
 
+                if '#' in entry:
+                    # Ignore comments at end of lines.
+                    entry = entry.split('#')[0]
```
pirat89: This seems wrong, as from the documentation:
> Lines starting with '#' are comments.

Regarding the tests, fstab can even contain "#" comments in the end of line (even when this is not documented and it is not recommended to do). So in case this occurs, I would log warning or something. Keep in mind, that current parsing we are doing is not 100% right, as even the hashtag carachter can be used as part of value (e.g. name or password) and do not have to mean it is comment. So I would be careful here and rather than split, I would log error and require remove of the comment from the line or at least log warning.

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py
```diff
@@ -79,22 +80,32 @@ def _get_fstab_info(fstab_path):
...
-                    fs_freq=fs_freq,
-                    fs_passno=fs_passno)
+                entries = entry.split()
+
+                if len(entries) not in [4, 5, 6]:
+                    summary = "Please fix your /etc/fstab file on the line: {}, it must have" \
+                        "at least 4 values and at most 6 this line has {}".format(line, len(entries))
+                    reporting.report_generic(
+                        title='Invalid Fstab', summary=summary, severity='high', flags=['inhibitor']
+                    )
```
pirat89: You should use remediation report. The summary you wrote is actually more like remediation and it is not telling what is wrong. Summary should be about the error that occur (received X fields, expected...) and remediation instruction what user should do. So, in example I would point users to `man 5 fstab` and explain we expect format documented in the man page with comments on separate lines (ideally).

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py
```diff
@@ -79,22 +80,41 @@ def _get_fstab_info(fstab_path):
...
-            for entry in fstab:
+            keys = ['fs_spec', 'fs_file', 'fs_vfstype', 'fs_mntops', 'fs_freq', 'fs_passno']
+            for line, entry in enumerate(fstab, 1):
                 if entry.startswith('#'):
                     continue
 
+                if '#' in entry:
+                    message = "Your /etc/fstab has '#' comment in line {} please verify" \
+                        "if it's correct.".format(line)
+                    api.current_logger().warning(message)
```
jmikovic: You should also strip the comment from the line else you can get inhibitor for valid fstab when it has comment at the end or not get it when there is comment somewhere in the middle but some options are missing

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py
```diff
@@ -79,22 +80,41 @@ def _get_fstab_info(fstab_path):
...
     if _is_file_readable(fstab_path):
         with open(fstab_path, 'r') as fstab:
-            for entry in fstab:
+            keys = ['fs_spec', 'fs_file', 'fs_vfstype', 'fs_mntops', 'fs_freq', 'fs_passno']
+            for line, entry in enumerate(fstab, 1):
                 if entry.startswith('#'):
                     continue
 
+                if '#' in entry:
+                    message = "Your /etc/fstab has '#' comment in line {} please verify" \
```
vinzenz: Please do not use \ for string continuation rather use parenthesis. Logging format strings should usually prefer to use %s.
Strings are supposed to use single quotes not double quotes. 
```suggestion
                    api.current_logger().warning(
                        'Detected in /etc/fstab a "#" comment in line %s please verify if it is correct.', line)
```

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py
```diff
@@ -79,22 +80,41 @@ def _get_fstab_info(fstab_path):
...
-                    fs_spec=fs_spec,
-                    fs_file=fs_file,
-                    fs_vfstype=fs_vfstype,
-                    fs_mntops=fs_mntops,
-                    fs_freq=fs_freq,
-                    fs_passno=fs_passno)
+                entries = entry.split()
+
+                if len(entries) not in [4, 5, 6]:
+                    summary = "The fstab configuration file is invalid, you need to fix this to avoid" \
```
vinzenz: Again double quotes and \ for line continuation

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py
```diff
@@ -79,22 +80,41 @@ def _get_fstab_info(fstab_path):
...
+                        "problems in the upgrade process."
+                    remediation = "The file /etc/fstab has a problem on the line: {}, it must have" \
```
vinzenz: Same here

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py
asilveir: I did it before but @pirat89  has commented about it, the '#' is a valid character when you have a password in options for example.

Now we are warning the user when there is the '#'

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py
```diff
@@ -79,22 +80,46 @@ def _get_fstab_info(fstab_path):
...
         with open(fstab_path, 'r') as fstab:
-            for entry in fstab:
+            keys = ['fs_spec', 'fs_file', 'fs_vfstype', 'fs_mntops', 'fs_freq', 'fs_passno']
+            for line, entry in enumerate(fstab, 1):
                 if entry.startswith('#'):
                     continue
 
+                if '#' in entry:
+                    api.current_logger().warning(
+                        'Detected in /etc/fstab a "#" comment in line %s please verify if it is correct.',
```
pirat89:
```suggestion
                        'Detected a "#" inline comment in the /etc/fstab file which could lead into unexpected behaviour: "%s"',
```

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py
```diff
@@ -79,22 +80,46 @@ def _get_fstab_info(fstab_path):
...
+                        line
+                    )
+
                 entry = entry.strip()
                 if not entry:
                     continue
```
pirat89: That's something that can be done safelly in the beginning of the loop. Whitespaces are not something that would bother us even in case of comments.

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py
```diff
@@ -79,22 +80,46 @@ def _get_fstab_info(fstab_path):
...
-                yield FstabEntry(
-                    fs_spec=fs_spec,
-                    fs_file=fs_file,
-                    fs_vfstype=fs_vfstype,
-                    fs_mntops=fs_mntops,
-                    fs_freq=fs_freq,
-                    fs_passno=fs_passno)
+                entries = entry.split()
+
+                if len(entries) not in [4, 5, 6]:
```
fernflower: `if len(entries) in range(4, 7)` is more self-descriptive imho

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py
```diff
@@ -79,22 +80,46 @@ def _get_fstab_info(fstab_path):
...
+                        remediation=remediation,
+                        severity='high',
+                        flags=['inhibitor']
+                    )
+
+                    api.current_logger().error(summary)
+                    continue
+
+                data = dict(zip(keys, entries))
+                yield FstabEntry(**data)
```
fernflower: I have 2 concerns here - kwargs are against our guidelines unless absolutely necessary and data isn't validated anywhere (on the other side - it has never been)

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/library.py
```diff
@@ -79,22 +80,46 @@ def _get_fstab_info(fstab_path):
     ''' Collect storage info from /etc/fstab file '''
     if _is_file_readable(fstab_path):
         with open(fstab_path, 'r') as fstab:
-            for entry in fstab:
+            keys = ['fs_spec', 'fs_file', 'fs_vfstype', 'fs_mntops', 'fs_freq', 'fs_passno']
+            for line, entry in enumerate(fstab, 1):
                 if entry.startswith('#'):
                     continue
 
+                if '#' in entry:
```
fernflower: This will generate a lot of false positives. Either remove this warning and check completely, or check that '#' is the first character of a word.

PR #275, opened 2019-07-09
to_remove: Remove all files that have been added to richdeps

vinzenz:
Previously we did not support upgrades of RPMs where the RHEL8
counterpart was using RichDependencies.
Since we merged the RHEL8 userspace this feature should now be available
to us and therefore we are starting with a clean slate here.
Left are only packages that are in the list due to other reasons than
richdependencies.


etc/leapp/transaction/to_remove
```diff
@@ -1,51 +1,2 @@
-redhat-rpm-config
-python2-devel
```
pirat89: If I am right, this one was added because of conflicting obsoletes/provides, not because of rich-deps - or at least the issue with provides/obsoletes has been discovered later. Maybe it was a different problem, but rather check that python2-devel is not problematic from this point.

More context:
- python-devel on RHEL 7 provides python2-devel so we cannot install python2-devel rpm because:
```
 # dnf install python2-devel
   Package python-devel-2.7.5-76.el7.x86_64 is already installed. 
```

vinzenz: I went by git blame which said otherwise

vinzenz:
https://github.com/oamg/leapp-repository/pull/276 is a more complete solution

PR #276, opened 2019-07-10
Clean "to_remove" transaction file

Rezney:
After we started using RHEL8 rpm userspace, we should not hit
richdeps issues anymore. As a result we are removing:

"tpm2-abrmd"
"redhat-rpm-config"

And packages that depend on "redhat-rpm-config".

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/rpmtransactionconfigtaskscollector/libraries/scanner.py
+++ b/repos/system_upgrade/el7toel8/actors/rpmtransactionconfigtaskscollector/libraries/scanner.py
@@ -8,7 +8,10 @@ def load_tasks_file(path, logger):
     if os.path.isfile(path):
         try:
             with open(path, 'r') as f:
-                return list(set([entry.strip() for entry in f.read().split() if entry.strip()]))
+                return list(
+                    {entry.strip() for entry in f.read().split('\n') if entry.strip() and
+                        not entry.strip().startswith('#')}
+                )
         except IOError as e:
             logger.warn('Failed to open %s to load additional transaction data. Error: %s', path, str(e))
     return []
--- a/etc/leapp/transaction/to_install
+++ b/etc/leapp/transaction/to_install
@@ -1,0 +1,1 @@
+### List of packages (each on new line) to be added to the upgrade transaction
--- a/etc/leapp/transaction/to_keep
+++ b/etc/leapp/transaction/to_keep
@@ -1,3 +1,5 @@
+### List of packages (each on new line) to be kept in the upgrade transaction
+
 leapp
 python2-leapp
 python3-leapp
--- a/etc/leapp/transaction/to_remove
+++ b/etc/leapp/transaction/to_remove
@@ -1,51 +1,14 @@
-redhat-rpm-config
+### List of packages (each on new line) to be removed from the upgrade transaction
+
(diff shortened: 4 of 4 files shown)


repos/system_upgrade/el7toel8/actors/rpmtransactionconfigtaskscollector/libraries/scanner.py
```diff
@@ -8,7 +8,9 @@ def load_tasks_file(path, logger):
     if os.path.isfile(path):
         try:
             with open(path, 'r') as f:
-                return list(set([entry.strip() for entry in f.read().split() if entry.strip()]))
+                return list(
```
vojtechsokol: Just a nitpick - what if line starts with ` #` (space + hash)? It should be treated as a comment but it ends up in the list.

AloisMahdal:
All our tests passed with

    leapp-0.7.0-1.201907100534Z.0f1dcf1.master.el7_6.noarch
    leapp-repository-0.7.0-0.201907100724Z.ad02e82.clean_to_remove.PR276.el7_6.noarch

repos/system_upgrade/el7toel8/actors/rpmtransactionconfigtaskscollector/libraries/scanner.py
pirat89: We can require there will not be space :)

repos/system_upgrade/el7toel8/actors/rpmtransactionconfigtaskscollector/libraries/scanner.py
```diff
@@ -8,7 +8,9 @@ def load_tasks_file(path, logger):
...
+                    {entry.strip() for entry in f.read().split('\n') if entry.strip() and not entry.startswith('#')}
```
vojtechsokol:
```suggestion
                    {entry.strip() for entry in f.read().split('\n') if entry.strip() and not entry.strip().startswith('#')}
```

Rezney: Implemented. Thanks...

Rezney:
> All our tests passed with
> 
> ```
> leapp-0.7.0-1.201907100534Z.0f1dcf1.master.el7_6.noarch
> leapp-repository-0.7.0-0.201907100724Z.ad02e82.clean_to_remove.PR276.el7_6.noarch
> ```

How many kernels did you have on the machine? @vinzenz hit a transaction error probably related to kernel-devel with more then two kernels installed. I was not able to reproduce it so far, but will try it again with clean env.

Rezney:
Will fix the long line later...

AloisMahdal:
> How many kernels did you have on the machine? @vinzenz hit a transaction error probably related to kernel-devel with more then two kernels installed. I was not able to reproduce it so far, but will try it again with clean env.

Only one.  I tried this multi-kernel scenario but it turns out it's tricky to get multiple kernels in our downstream testing environment; we normally only have access to one compose.

pirat89:
@Rezney pls rebase it again master.

Rezney:
rebased

Rezney:
> EDIT: @Rezney pls rebase it again, master.
> ###### sed/again,/against/
> 
> smirk_cat :octocat:

kree

PR #279, opened 2019-07-11
Enable crb <=> the Optional repo is enabled

pirat89:
Blacklist the CRB repository (as before) iff the Optional repository is not enabled. Otherwise keep the rest on other actors that will handle it as other repositories. In case that the CRB repository is used during the upgrade (can be only when it is not blacklisted), enable the repo on the target system as well during the `FirstBoot` phase.

*Note*: Planned todo in future: the actor is working with this one repo now, but would be nice to make it more robust to be used for any repository task in future. See the TODO msg in code.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/enablecrbrepo/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/enablecrbrepo/actor.py
@@ -1,0 +1,21 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import library
+from leapp.models import UsedTargetRepositories
+from leapp.tags import FirstBootPhaseTag, IPUWorkflowTag
+
+
+class EnableCRBrepo(Actor):
+    """
+    Enable CRB repository when used during the upgrade.
+    """
+    # TODO: replace this actor by more robust one which will enable repositories
+    # based on msg; be careful about the solution as there is similar thing
+    # for the upgrade process already - see the RepositoriesSetupTasks model
+
+    name = 'enable_crb_repo'
+    consumes = (UsedTargetRepositories,)
+    produces = ()
+    tags = (IPUWorkflowTag, FirstBootPhaseTag)
+
+    def process(self):
+        library.process()
--- a/repos/system_upgrade/el7toel8/actors/enablecrbrepo/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/enablecrbrepo/libraries/library.py
@@ -1,0 +1,22 @@
+from leapp.libraries.stdlib import api, CalledProcessError, run
+from leapp.models import UsedTargetRepositories
+
+
+CRB_REPOID = 'codeready-builder-for-rhel-8-x86_64-rpms'
+
+
+def _is_crb_used():
+    # the UsedTargetRepositories has to be set always, by design of IPU
+    used_repos = next(api.consume(UsedTargetRepositories), None)
+    for repo in used_repos.repos:
+        if repo.repoid == CRB_REPOID:
+            return True
+    return False
+
+
+def process():
+    if _is_crb_used():
+        try:
+            run(['subscription-manager', 'repos', '--enable', CRB_REPOID])
+        except CalledProcessError as e:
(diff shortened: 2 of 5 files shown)


pirat89:
Btw, probably would be better to inform even that when Optional channel is enabled, CRB will be enabled as well and in such case upgrade is not supported (or any error related somehow to the content in those repos)

repos/system_upgrade/el7toel8/actors/enablecrbrepo/actor.py
```diff
@@ -0,0 +1,21 @@
...
+from leapp.models import UsedTargetRepositories
+from leapp.tags import FirstBootPhaseTag, IPUWorkflowTag
+
+
+class EnableCRBrepo(Actor):
+    """
+    Enable CRB repository when used during the upgrade.
+    """
+    # TODO: replace this actor by more robust one which will enable repositories
+    # based on msg; be careful about the solution as there is similar think
```
vojtechsokol: typo
```suggestion
    # based on msg; be careful about the solution as there is similar thing
```

repos/system_upgrade/el7toel8/actors/enablecrbrepo/libraries/library.py
```diff
@@ -0,0 +1,22 @@
+from leapp.libraries.stdlib import api, CalledProcessError, run
+from leapp.models import UsedTargetRepositories
+
+
+CRB_REPOID = 'codeready-builder-for-rhel-8-x86_64-rpms'
```
vojtechsokol: What about moving this to config actor?

repos/system_upgrade/el7toel8/actors/enablecrbrepo/libraries/library.py
```diff
@@ -0,0 +1,22 @@
...
+            return True
+    return False
+
+
+def process():
+    if _is_crb_used():
+        try:
+            run(['subscription-manager', 'repos', '--enable', CRB_REPOID])
+        except CalledProcessError as e:
+            api.current_logger().error('Cannot enable CRB repository: %s', str(e))
```
vojtechsokol:
```suggestion
            api.current_logger().error('Cannot enable CRB repository: {}'.format(str(e)))
```

repos/system_upgrade/el7toel8/actors/repositoriesblacklist/actor.py
```diff
@@ -9,13 +9,21 @@ class RepositoriesBlacklist(Actor):
     """
 
     name = 'repositories_blacklist'
-    consumes = ()
+    consumes = (RepositoriesFacts,)
     produces = (RepositoriesBlacklisted,)
     tags = (IPUWorkflowTag, FactsPhaseTag)
 
+    def _is_repo_enabled(self, repoid):
```
jmikovic: shouldn't this function be moved to library?

repos/system_upgrade/el7toel8/actors/enablecrbrepo/libraries/library.py
pirat89: Nope as this is not anything related to the config

repos/system_upgrade/el7toel8/actors/enablecrbrepo/libraries/library.py
pirat89: regarding discussion, this is allowed for logging (see various other actors)

repos/system_upgrade/el7toel8/actors/enablecrbrepo/actor.py
pirat89: fixed

repos/system_upgrade/el7toel8/actors/repositoriesblacklist/actor.py
pirat89: Not necessary. The actor can be tested as whole component. I am thinking more about shared library to work with repositories completely but this is for following PR.

PR #283, opened 2019-07-15
Skip comments in repository mapping file

vojtechsokol:


Rezney:
Looks good to me. 

As per https://stackoverflow.com/questions/1961006/can-a-csv-file-have-a-comment it looks like there is no standard about how comments should be handled in csv so we might add one comment to our repomap file to indicate that it is indeed "#" in our case?

repos/system_upgrade/el7toel8/actors/repositoriesmapping/libraries/library.py
```diff
@@ -15,6 +15,13 @@ def inhibit_upgrade(title):
     raise StopActorExecution()
 
 
+def strip_comments(file_):
+    for line in file_:
+        data = line.split('#')[0].strip()
```
Rezney: What about if we ignore a line starting with "#" and communicate that only new line comments are valid?

vojtechsokol:
@Rezney There are comments in the repository mapping file that is updated for alt arches and.

repos/system_upgrade/el7toel8/actors/repositoriesmapping/libraries/library.py
```diff
@@ -28,10 +28,15 @@ def scan_repositories(path):
         next(data)  # skip header
 
         for row in data:
-            if len(row) != 6:
+            # skip empty lines and comments
+            if not row or row[0].startswith('#'):
```
Rezney: I guess we will not catch empty lines if they begin with white-space, however not sure whether we care that much as those will be caught later as invalid record...

vojtechsokol: That is on purpose, I think it is better to keep the parsing as simple as possible, which servers two purposes: 1) less complex code, 2) motivation to keep the repomapping file clean.

Rezney: fair enough :)

PR #286, opened 2019-07-18
WIP: Remove python-tools before transaction

AloisMahdal:
Unfortunately I don't have valid reference URL for this.  During
downstream testing, we've found that presence of this package prevents
leapp from creating a valid transaction.

At some point @pirat89 mentioned that this was known already (@bocekm
should also know) but I haven't found Bugzilla or Github issue.

### DIFF
--- a/etc/leapp/transaction/to_remove
+++ b/etc/leapp/transaction/to_remove
@@ -12,3 +12,5 @@ python2-virtualenv
 squid
 # python-docs - https://bugzilla.redhat.com/show_bug.cgi?id=1703605
 python-docs
+
+python-tools


AloisMahdal:
e2e tests

etc/leapp/transaction/to_remove
```diff
@@ -12,3 +12,5 @@ python2-virtualenv
 squid
 # python-docs - https://bugzilla.redhat.com/show_bug.cgi?id=1703605
 python-docs
+
```
vinzenz: Please add a comment here as of why we remove python-tools, that's helpful when trying to resolve problems

AloisMahdal: The thing is, I don't actually know.  I know the ref should be here, I was expecting @bocekm or @pirat89 to chip in.

I probably should have marked the PR as WIP....

pirat89: @AloisMahdal not sure what is still unclear on this approach...

@vinzenz The same reason as `python2-devel` in the same list.

pirat89:
```
# rpm -q --provides python-tools
python-tools = 2.7.5-76.el7
python-tools(x86-64) = 2.7.5-76.el7
python2-tools = 2.7.5
[root@localhost ~]# rpm -q --obsoletes python-toolspython2-tools

```

pirat89:
Closing this one in favour of PR #290  as we need to resolve the problem right now.

etc/leapp/transaction/to_remove
AloisMahdal: @pirat89 I was responding to Vinny pointing out missing **reference**.  Whether I understand the mechanics of how the transaction is broken is irrelevant.

Frankly, I even think I made that pretty clear even in the OP.  You seem to be missing this constantly. Understanding some technical aspects issue is one thing, referencing a decision transparently (ie. linking to BZ/gh) is another one.  My goal is the second one.

PR #287, opened 2019-07-18
Move documentation to Read the Docs.

asilveir:
This documentation below is centralized in RTD:

- Best practices
- Dependencies
- Howto Inhibit Upgrade
- Tests


jmikovic:
I don't think we need 4 files with the same content here... It would be better to have just one file explaining that the docs have been moved and maybe listing what user can find at RtD

examon:
I agree with Jozef. It will be much cleaner with one file. :)

jmikovic:
@asilveir did you forgot to add the new file? :)

asilveir:
> @asilveir did you forget to add the new file? :)

Ops, sorry :(

jmikovic:
I would put it as README there, so it renders in github

PR #288, opened 2019-07-19
Mount also "/boot/efi" if system is EFI

Rezney:
Currently, when removing our upgrade boot entry, we try to
mount only "/boot", but when the system is EFI, we fails
as "grubby" see only "/boot" as try to update "/boot/grub/grub.cfg"
instead of "/boot/efi/EFI/redhat/grub.cfg".

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/removeupgradebootentry/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/removeupgradebootentry/libraries/library.py
@@ -1,17 +1,28 @@
 from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.stdlib import CalledProcessError, api, run
-from leapp.models import BootContent
+from leapp.models import BootContent, FirmwareFacts
 
 
 def remove_boot_entry():
     # we need to make sure /boot is mounted before trying to remove the boot entry
-    try:
-        run([
-            '/bin/mount', '/boot'
-        ])
-    except CalledProcessError:
-        # /boot has been most likely already mounted
-        pass
+
+    facts_msg = api.consume(FirmwareFacts)
+    facts = next(facts_msg, None)
+    if not facts:
+        raise StopActorExecutionError('Could not identify system firmware',
+                                      details={'details': 'Actor did not receive FirmwareFacts message.'})
+
+    mount_points_per_firmware = {
+        'bios': ['/boot'],
+        'efi': ['/boot', '/boot/efi']
+    }
+
+    for mp in mount_points_per_firmware[facts.firmware]:
+        try:
+            run(['/bin/mount', mp])
+        except CalledProcessError:
+            # partitions have been most likely already mounted
+            pass
     kernel_filepath = get_upgrade_kernel_filepath()
     run([
         '/usr/sbin/grubby',
@@ -30,7 +41,7 @@ def get_upgrade_kernel_filepath():
     if list(boot_content_msgs):
         api.current_logger().warning('Unexpectedly received more than one BootContent message.')
     if not boot_content:
-        raise StopActorExecutionError('Could not create a GRUB boot entry for the upgrade initramfs.',
+        raise StopActorExecutionError('Could not remove GRUB boot entry for the upgrade initramfs.',
                                       details={'details': 'Did not receive a message about the leapp-provided'
                                                           ' kernel and initramfs.'})
     return boot_content.kernel_path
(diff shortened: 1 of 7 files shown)


repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -269,3 +269,7 @@ def _get_firewall_status(service_name):
         iptables=_get_firewall_status('iptables'),
         ip6tables=_get_firewall_status('ip6tables'),
     )
+
```
asilveir: lint is complaining that needs another blank line here

repos/system_upgrade/el7toel8/actors/removeupgradebootentry/libraries/library.py
```diff
@@ -1,14 +1,24 @@
...
+    mount_points_per_firmware = {
+        'bios': ['/boot'],
+        'efi': ['/boot', '/boot/efi']
+    }
+
     try:
-        run([
-            '/bin/mount', '/boot'
-        ])
+        run(['/bin/mount'] + mount_points_per_firmware[facts.firmware])
```
vinzenz: I think this should be rather a for loop, no? Otherwise you try to mount /boot to /boot/efi if I am not mistaken

repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -13,7 +13,7 @@
 from leapp.libraries.stdlib import CalledProcessError, api, run
 from leapp.models import SysctlVariablesFacts, SysctlVariable, ActiveKernelModulesFacts, ActiveKernelModule, \
     KernelModuleParameter, UsersFacts, User, GroupsFacts, Group, RepositoriesFacts, RepositoryFile, RepositoryData, \
-    SELinuxFacts, fields, FirewallStatus, FirewallsFacts
+    SELinuxFacts, fields, FirewallStatus, FirewallsFacts, SystemFacts
```
vinzenz: Did you forget to commit this Model?

Rezney: yep, fixed.

repos/system_upgrade/el7toel8/actors/removeupgradebootentry/libraries/library.py
Rezney: You are not, fixed.

repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
Rezney: fixed

repos/system_upgrade/el7toel8/actors/removeupgradebootentry/libraries/library.py
```diff
@@ -1,16 +1,27 @@
...
+
     try:
-        run([
-            '/bin/mount', '/boot'
-        ])
+        for mp in mount_points_per_firmware[facts.firmware]:
+            run(['/bin/mount'] + mp)
     except CalledProcessError:
-        # /boot has been most likely already mounted
+        # partitions have been most likely already mounted
```
vinzenz: I don't think we should do it like this, but have the exception being caught in the loop, just in case one of the given mount points is not a mount point.

repos/system_upgrade/el7toel8/libraries/dnfplugin.py
```diff
@@ -168,8 +168,17 @@ def perform_transaction_install(target_userspace_info, used_repos, tasks):
         '/proc:/installroot/proc',
         '/run/udev:/installroot/run/udev'
     ]
-    if os.path.ismount('/boot'):
-        bind_mounts.append('/boot:/installroot/boot')
+    # the order has to persist
+    mount_points_maps = (
```
vinzenz:
```python
    mount_points_maps = {
        '/boot': '/boot:/installroot/boot',
        '/boot/efi': '/boot/efi:/installroot/boot/efi'
    }
```

repos/system_upgrade/el7toel8/actors/removeupgradebootentry/libraries/library.py
```diff
@@ -1,17 +1,28 @@
...
+                                      details={'details': 'Actor did not receive SystemFacts message.'})
+
+    mount_points_per_firmware = {
+        'bios': ['/boot'],
+        'efi': ['/boot', '/boot/efi']
+    }
+
+    for mp in mount_points_per_firmware[facts.firmware]:
+        try:
+            run(['/bin/mount'] + [mp])
```
vinzenz:
```suggestion
            run(['/bin/mount', mp])
```

bocekm:
Tested on UEFI RHEL 7.6 EUS and it upgraded fine.

Rezney:
@asilveir anything still needed as per you request?
@pirat89 do not remember what all was in the "additional stuff" as per you request comment. Could please take a look?

pirat89:
@Rezney I do not remember already what was it and I am little in time pressure, so I dissmiss my review.

PR #292, opened 2019-07-25
Allow to override the target version for devel

vinzenz:
This PR introduces to use the LEAPP_DEVEL_TARGET_VERSION environment
variable which allows to override the targeted version.
This value will then be used to set a target version for RHSM.

As requested by @AloisMahdal

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/actor.py
@@ -18,7 +18,7 @@ class IPUWorkflowConfig(Actor):
     tags = (IPUWorkflowTag,)
 
     def process(self):
-        target_version = '8.0'
+        target_version = library.get_target_version()
         os_release = library.get_os_release('/etc/os-release')
         self.produce(IPUConfig(
             leapp_env_vars=library.get_env_vars(),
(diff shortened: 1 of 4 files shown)


repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/actor.py
```diff
@@ -18,7 +18,7 @@ class IPUWorkflowConfig(Actor):
     tags = (IPUWorkflowTag,)
 
     def process(self):
-        target_version = '8.0'
+        target_version = libary.get_target_version()
```
asilveir: little typo here

AloisMahdal:
Will this also affect the {{--set}} we (IIRC) do in initramdisk or later?

pirat89:
> Will this also affect the RHSM `--set` we (IIRC) do in initramdisk or later?

Yes, it will.

repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/libraries/library.py
```diff
@@ -37,3 +38,7 @@ def get_os_release(path):
             severity='high',
             flags=['inhibitor'])
         return None
+
+
+def get_target_version():
+    return os.getenv('LEAPP_DEVEL_TARGET_RELEASE', None) or CURRENT_TARGET_VERSION
```
fernflower: Isn't it the same as `os.getenv('LEAPP_DEVEL_TARGET_RELEASE', CURRENT_TARGET_VERSION)`?

vinzenz: No it's not the same. If it's an empty environment variable CURRENT_TARGET_VERSION is used in your version its not using it.

vinzenz: Ofc, I could go even more paranoid and check for the stripped string, but I didn't want to go that far

PR #293, opened 2019-07-25
RHSM: Retry refresh and release write actions on failure

vinzenz:
Previously it happened that Satellite subscribed machines were failing
when trying to perform a task too fast. This patch introduces a way to
rate limit these tasks and have them retry on failures after sleeping
for some time.

Retry count with this patch is set to 5
Sleep time with this patch is set to 5 seconds

### DIFF
--- a/repos/system_upgrade/el7toel8/libraries/rhsm.py
+++ b/repos/system_upgrade/el7toel8/libraries/rhsm.py
@@ -2,15 +2,44 @@ import contextlib
 import functools
 import os
 import re
+import time
 
 from leapp.exceptions import StopActorExecutionError
-from leapp.libraries.stdlib import api, CalledProcessError
+from leapp.libraries.stdlib import CalledProcessError, api
 from leapp.models import TargetRHSMInfo
-
 
 _RE_REPO_UID = re.compile(r'Repo ID:\s*([^\s]+)')
 _RE_RELEASE = re.compile(r'Release:\s*([^\s]+)')
 _RE_SKU_CONSUMED = re.compile(r'SKU:\s*([^\s]+)')
+_RETRY_TIMES = 5
+_RETRY_SLEEP = 5
+
+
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/el7toel8/libraries/rhsm.py
```diff
@@ -11,6 +11,35 @@
...
+                    if times < attempts:
+                        api.current_logger().warning(
+                            'Attempt %d of %d to perform %s failed. Maximum number of retries have been reached.',
+                            attempts, times, f.__name__)
+                        raise
+                    if sleep:
+                        api.current_logger().info(
+                            'Attempt %d of %d to perform %s failed - Retrying after %s seconds',
+                            attempts, times, f.__name__, str(sleep))
+                        time.sleep(sleep)
```
asilveir: missing import of time

pirat89:
For small testing to simulate troubles with subscription-manager, you can use something like this:

- move /usr/sbin/subscription-manager to /usr/sbin/subscription-manager
- create new executable script like:

```bash
#!/bin/bash

[ ! -e /root/retry ] && echo 0 > /root/retry
retry=$(cat /root/retry)

echo "$@" | grep "release" | grep -q -- "--set"

if [ $? -eq 0 ]; then
    echo >&2 "RHSM:: RELEASE SET!! $retry"
    [ $retry -gt 6 ] && {
      echo >&2 "RHSM: but continue"
      /usr/sbin/subscription-manager2 $@
      exit $?
    }
    echo $(( retry + 1 )) > /root/retry
    sleep 2
    echo >&2 "KILLLLING"
    sleep 2
    exit 1
else
    echo >&2 "RHSM:: continue"
    /usr/sbin/subscription-manager2 $@
    exit $?
fi
```

- first run can be simulated complete crash
- second run can be simulated problem + working rhsm

PR #298, opened 2019-08-07
Install product certificates

vojtechsokol:
Previously the installation of product certificate was hardcoded and only
correct for x86_64 architecture.

This patch introduces installation of the correct product certificate based on
variant, architecture and target version of the system provided by a IPUConfig
message.

Also removal of the old product certificate is not hardcoded any more and is
based on a SourceRHSMInfo message.


repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -47,29 +50,45 @@ def _prep_repository_access(context, target_userspace):
...
     Retrieves the required / used product certificate for RHSM.
     """
-    sys_var = get_os_variant()
-    var_prodcert = {'server': '479.pem'}
-    if sys_var not in var_prodcert:
-        raise StopActorExecutionError(
-            message=('Failed to to retrieve Product Cert file.'
-                     'Product cert file not available for System Variant \'{}\'.'.format(sys_var))
-        )
+    config = next(api.consume(IPUConfig), None)
```
pirat89: This is wrong. You do not need to consume IPUConfig - that's not how the configuration works.

Rezney: to expand on this. You could do: 

`-
api.current_actor().configuration.<model_field_you_want>`

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -47,29 +50,45 @@ def _prep_repository_access(context, target_userspace):
...
+    architecture = config.architecture
+    tgt_version = config.version.target
+    # TODO: so far only base channel is available in rhel8
+    tgt_channel = 'base'
+    certs_folder = api.get_common_folder_path(PROD_CERTS_FOLDER)
+
+    prod_certs = {
+        'server': {
+            'x86_64': {
+                'base': '479.pem',
```
pirat89: Instead of `base` I think we should rather use `standard` - that's the proper wording here.

vojtechsokol: As per https://access.redhat.com/articles/rhel-eus#c4, `base` is the right term to use here.

bocekm:
@AloisMahdal, could you please schedule e2e tests on all arches for this PR?

AloisMahdal:
@bocekm I did, but due to outage we're extremely short on s390x, so I wouldn't hold my breath for these. I'll check back in few hours.

bocekm:
@vojtechsokol, please rebase (due to the leapp framework reporting refactoring).

PR #301, opened 2019-08-09
setuptargetrepos actor code cleanup

bocekm:
To changelog:
Remove RepositoriesSetupTasks which was kinda duplicate to the CustomTargetRepository.

Resolves https://github.com/oamg/leapp-repository/issues/233.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
@@ -6,8 +6,9 @@ from leapp.exceptions import StopActorExecution, StopActorExecutionError
 from leapp import reporting
 from leapp.libraries.stdlib import api
 from leapp.libraries.stdlib.config import is_verbose
-from leapp.models import (InstalledRedHatSignedRPM, PESRpmTransactionTasks,
-                          RepositoriesSetupTasks, RpmTransactionTasks, RepositoriesBlacklisted)
+from leapp.models import (CustomTargetRepositories, CustomTargetRepository,
+                          InstalledRedHatSignedRPM, PESRpmTransactionTasks,
+                          RepositoriesBlacklisted, RpmTransactionTasks)
 
 # FIXME: this mapping is not complete and will need to be manually updated frequently
 REPOSITORIES_MAPPING = {
@@ -406,4 +407,4 @@ def produce_messages(tasks):
                                            to_remove=to_remove_pkgs))
 
     if to_enable_repos:
-        api.produce(RepositoriesSetupTasks(to_enable=to_enable_repos))
+        api.produce(CustomTargetRepositories(repos=[CustomTargetRepository(repoid=repo) for repo in to_enable_repos]))
--- a/repos/system_upgrade/el7toel8/actors/setuptargetrepos/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/setuptargetrepos/actor.py
@@ -1,24 +1,23 @@
 import platform
 
 from leapp.actors import Actor
-from leapp.models import (CustomTargetRepository, RepositoriesBlacklisted, RepositoriesFacts, RepositoriesMap,
-                          RepositoriesSetupTasks, RHELTargetRepository, SkippedRepositories, TargetRepositories,
-                          UsedRepositories)
+from leapp.models import (CustomTargetRepositories, RepositoriesBlacklisted,
+                          RepositoriesFacts, RepositoriesMap,
+                          RHELTargetRepository, SkippedRepositories,
+                          TargetRepositories, UsedRepositories)
 from leapp.tags import FactsPhaseTag, IPUWorkflowTag
 
 
 class SetupTargetRepos(Actor):
     """
-    Produces list of repositories that should be available to be used by Upgrade process.
+    Produces list of repositories that should be available to be used by the upgrade process.
 
-    Based on current set of Red Hat Enterprise Linux repositories, produces the list of target
-    repositories. Additionaly process request to use custom repositories during the upgrade
-    transaction.
+    Based on the enabled RHEL 7 repositories, it produces a list of RHEL 8 repositories to be enabled for the upgrade.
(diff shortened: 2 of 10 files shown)


repos/system_upgrade/el7toel8/actors/setuptargetrepos/actor.py
```diff
@@ -1,91 +1,109 @@
...
+
+    def get_used_repos(self):
+        """Get those enabled repositories from which at least one RH-signed package is currently installed."""
+        for used_repos_msg in self.consume(UsedRepositories):
+            self.used_repos += used_repos_msg.repositories
+
+    def get_pkgs_installed_from_not_mapped_repos(self, not_mapped_repos):
+        pkgs = set()
+        for repo in not_mapped_repos:
+            pkgs.update(self.used_repos[repo])
```
asilveir: I got this error:
``
    pkgs.update(self.used_repos[repo])
TypeError: list indices must be integers, not unicode
``

bocekm: Thanks! Fixed.

repos/system_upgrade/el7toel8/actors/setuptargetrepos/actor.py
```diff
@@ -1,91 +1,110 @@
...
+
+        if not_mapped_repos:
+            pkgs = self.get_pkgs_installed_from_not_mapped_repos(not_mapped_repos)
+            self.produce(SkippedRepositories(repos=list(not_mapped_repos), packages=list(pkgs)))
+
+    def get_used_repos(self):
+        """Get those enabled repositories from which at least one RH-signed package is currently installed."""
+        for used_repos_msg in self.consume(UsedRepositories):
+            for used_repo in used_repos_msg.repositories:
+                self.used_repos[used_repo.repository] = used_repo.packages
```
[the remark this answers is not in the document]
bocekm: Sorry about that, I should have tested it properly. I improved one unit test to catch such an issue.

bocekm:
leapp-ci build

bocekm:
leapp-ci build

bocekm:
@asilveir, ready for re-review.

bocekm:
leapp-ci build

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -127,20 +127,19 @@ def gather_target_repositories(target_rhsm_info):
                 }
             )
 
-    target_repoids = []
+    target_repoids = set()
```
bocekm: This is to avoid duplicate repoids being added to the list.

bocekm:
Rebased and conflicts resolved.

asilveir:
Tested and working.

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
pirat89: just for info, usually is better to work with list and than create set. Here it is not so big deal, but in case you modfy set manytimes and set is going to be big,  you will reach performance issues.

repos/system_upgrade/el7toel8/models/repositoriessetuptasks.py
```diff
@@ -6,6 +6,7 @@ class RepositoriesSetupTasks(Model):
     """
     Information about repositories that must be managed in order to complete upgrade process.
 
+    !!! THIS MODEL IS OBSOLETE. Use for example CustomTargetRepositories instead. !!!
```
pirat89: I do not agree with the idea this model is kind of obsoleted. I can imagine in future that based on checks, there will be case a specific repository needs to be enabled / disabled. I am fun of   *Tasks* suffix as this is kind of self explaining name we could use across many problems we would like to handle on one place. But the approach is questionable. Let's see what will be the way in future.

repos/system_upgrade/el7toel8/actors/setuptargetrepos/actor.py
```diff
@@ -1,91 +1,109 @@
...
+    custom_repos = []
+    enabled_repos = set()
+    mapping_of_repositories = []  # list of RepositoryMap
+    mapped_rhel7_repoids = set()  # enabled RHEL 7 repos that have been successfully mapped to the RHEL 8 ones
+    target_repos = []
+    used_repos = {}
 
-        enabled_repos = set()
+    def process(self):
+        self.get_custom_repos()
```
pirat89: Actually, it returns now rhel repositories. not custom

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -406,4 +407,4 @@ def produce_messages(tasks):
                                            to_remove=to_remove_pkgs))
 
     if to_enable_repos:
-        api.produce(RepositoriesSetupTasks(to_enable=to_enable_repos))
+        api.produce(CustomTargetRepositories(repos=[CustomTargetRepository(repoid=repo) for repo in to_enable_repos]))
```
pirat89: that's the thing. You are presenting RH repos as custom which is wrong. You are changing here the meaning of CustomTargetRepositories modesl at all.

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
pirat89: You are right. I tried that with more complex structures and direct use of set is more efficient than hack mentioned above.  Maybe it was wrong all the time, maybe it was correct when set was not built-in yet in python and you had to use the `sets` module (no longer exists in Py3, still can be found in Py2). Anyway direct use of the `set` is faster today.

Interesting is, that Python 2.7 is even faster than Python3.7 (in both cases, list and test as well) around 3.5s (tried several times). I expected that Python3 will be more efficient.

PR #303, opened 2019-08-14
DEVEL_ONLY: Allow accepting unsigned pkgs

Rezney:


pirat89:
@Rezney : I think the name "accept_unsigned" could be misleading a little. Someone would point out that they want to do something with unsigned RPMs (install durign the transaction for example) and they could think they need to use the envar. Instead of that, what about `LEAPP_DEVEL_RPMS_ALL_SIGNED`

Rezney:
@pirat89  "LEAPP_DEVEL_RPMS_ALL_SIGNED" sounds fair

repos/system_upgrade/el7toel8/actors/redhatsignedrpmscanner/tests/test_redhatsignedrpmscanner.py
```diff
@@ -104,7 +124,7 @@ def test_has_package(current_actor_context):
     ]
 
     current_actor_context.feed(InstalledRPM(items=installed_rpm))
-    current_actor_context.run()
+    current_actor_context.run(config=mock_configs.CONFIG)
```
jmikovic:
```suggestion
    current_actor_context.run(config_model=mock_configs.CONFIG)
```

jmikovic:
leapp-ci build

repos/system_upgrade/el7toel8/actors/redhatsignedrpmscanner/tests/test_redhatsignedrpmscanner.py
Rezney: Fixed, thanks.

repos/system_upgrade/el7toel8/libraries/config/mock_configs.py
```diff
@@ -0,0 +1,33 @@
+from leapp.models import IPUConfig, EnvVar, OSRelease, Version
+
```
pirat89: Maybe little too late, but couldn't be better to put a discaimer into the file? I knot that there is *mock_* prefix in the name, but to ensure that people will not use it in the standard code

bocekm: I'm about to merge #340 in a while.

PR #308, opened 2019-08-15
Add tests for BiosDevName actor.

asilveir:
- Refactored to library
- Added tests

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/biosdevname/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/biosdevname/libraries/library.py
@@ -1,0 +1,57 @@
+import re
+
+import pyudev
+
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.stdlib import api
+from leapp.models import KernelCmdlineArg, PersistentNetNamesFacts
+
+
+def is_biosdevname_disabled():
+    with open('/proc/cmdline') as cmdline:
+        if 'biosdevname=0' in cmdline.read():
+            return True
+
+    return False
+
+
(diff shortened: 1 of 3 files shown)


repos/system_upgrade/el7toel8/actors/biosdevname/libraries/library.py
```diff
@@ -0,0 +1,57 @@
...
+def check_biosdevname():
+    if is_biosdevname_disabled():
+        return
+
+    net_names_facts_messages = api.consume(PersistentNetNamesFacts)
+    net_names_facts = next(net_names_facts_messages, None)
+    if not net_names_facts:
+        raise StopActorExecutionError(
+            'Could not read interfaces names',
+            details={'details': 'No PersistentNetNamesFacts facts found.'},
```
jmikovic:
```suggestion
            details={'details': 'No PersistentNetNamesFacts message found.'},
```

repos/system_upgrade/el7toel8/actors/biosdevname/tests/test_biosdevname.py
```diff
@@ -0,0 +1,128 @@
...
+import pytest
+import pyudev
+
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.actor import library
+from leapp.libraries.stdlib import api
+from leapp.models import Interface, PCIAddress
+
+
+class logger_mocked(object):
```
jmikovic:
```suggestion
class LoggerMocked(object):
```

I noticed that we are using this class in multiple tests now, so we should probably move it to library but that's a task for different PR.

Issue #177, opened 2019-04-26
Do not use PES events with input packages defined in to_remove


bocekm:
**Actual behavior**
We have /etc/leapp/transaction/to_remove file which is used to list packages to be removed during the upgrade. Even though such package is passed to dnf as to be removed, if this package is present in any Renamed/Replaced/Split/Merged PES event as "input" package, such event is parsed and the event's "output" packages are added incorrectly to the dnf transaction.

Example: _python-docs_ is in PES as split to _python2-docs_ and _python3-docs_.

**To Reproduce**

1. Install python-docs
3. Run `leapp upgrade --debug`
4. See error:
```
Package python-docs-2.7.5-2.el7.noarch is already installed.
Problem: cannot install the best candidate for the job
  - conflicting requests
```

**Expected behavior**
dnf is not told to install python2-docs, python3-docs.

**Attachments**
[dnf-plugin-data.txt](https://github.com/oamg/leapp-repository/files/3122739/dnf-plugin-data.txt) - mentions:

- _python-devel_ in `to_remove`
- _python2-docs_, _python3-docs_ in `to_install`

Issue #124, opened 2019-04-03
snactor discovery crashed because of missing repository


landgraf:
**Actual behavior**
I try to discover supported entities from the el7toel8 repository and it's failed because of missing common repository. Linking these repos together doesn't help.

**To Reproduce**
Clone repo
cd leapp-repository/repos/system_upgrade/el7toel8
snactor discovery
<traceback_see_below>
snactor link --path ../../common
snactor discovery
<traceback_see_below>

**Expected behavior**
list of entities

**System information (please complete the following information):**
fedora rawhide + virtualenv leapp
snactor version 0.5.0
leapp-repository commit 8d722afcc23c9402a0e6a33b8fb47d0f6e93dac9

$ snactor discover
Traceback (most recent call last):
  File "/usr/bin/snactor", line 9, in <module>
    load_entry_point('leapp==0.5.0', 'console_scripts', 'snactor')()
  File "/usr/lib/python2.7/site-packages/leapp/snactor/__init__.py", line 67, in main
    cli.command.execute(version=_('snactor version {}').format(VERSION))
  File "/usr/lib/python2.7/site-packages/leapp/utils/clicmd.py", line 80, in execute
    args.func(args)
  File "/usr/lib/python2.7/site-packages/leapp/utils/clicmd.py", line 102, in called
    self.target(args)
  File "/usr/lib/python2.7/site-packages/leapp/utils/repository.py", line 20, in checker
    return f(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/leapp/snactor/commands/discover.py", line 105, in cli
    repository = find_and_scan_repositories(base_dir, include_locals=True)
  File "/usr/lib/python2.7/site-packages/leapp/repository/scan.py", line 57, in find_and_scan_repositories
    _resolve_repository_links(manager=manager, include_locals=include_locals)
  File "/usr/lib/python2.7/site-packages/leapp/repository/scan.py", line 39, in _resolve_repository_links
    raise RepositoryConfigurationError('Missing repositories detected: {}'.format(', '.join(missing)))
leapp.exceptions.RepositoryConfigurationError: Missing repositories detected: efcf9016-f2d1-4609-9329-a298e6587b3c

bocekm:
CCing @vinzenz.

Honza0297:
I am experiencing the same issue.
Not only with  `snactor discovery`, but also with `snactor run  actors/(does not matter)/actor.py`


#### System info
[...]

PR #310, opened 2019-08-19
Add actor checking KDE and GNOME

Honza0297:
A simple script detecting installed KDE desktop environment and its apps,
written in leapp framework. When started, the script will check whether
KDE is installed together with GNOME desktop to inform whether we can
inhibit the upgrade process. When both are installed, we need to inform
the user that KDE will be removed and GNOME will be used instead. If only
KDE is installed, we want to inhibit the upgrade process otherwise the
user will end up without a desktop. Also if both are installed,
but KDE is not used as default desktop, we further check most common
KDE apps to inform the user that those will be removed in case the user is using them.

Possible values in message CanUpgrade:

* 0: we can safely continue with the upgrade process
* 1: we cannot continue with the upgrade process
* 2: we can continue with the upgrade process, but KDE desktop will be removed
together with KDE apps in favor of GNOME desktop

Original script in bash written by jgrulich [0].
[0] https://gitlab.cee.redhat.com/jgrulich/rhel-upgrade

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkkdeapps/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/checkkdeapps/libraries/library.py
@@ -1,0 +1,25 @@
+from leapp.libraries.stdlib import api
+from leapp.libraries.common.rpms import has_package
+from leapp.models import InstalledRPM
+
+
+def get_kde_apps_info():
+    installed = list()
+    base_kde_apps = ("kde-baseapps",
+                     "okular",
+                     "ark",
+                     "kdepim",
+                     "konsole",
+                     "gwenview",
+                     "kdenetwork",
+                     "kate",
+                     "kwrite")
+
+    api.current_logger().info("  Detecting installed KDE apps  ")
+    api.current_logger().info("================================")
+    for app in [application for application in base_kde_apps if has_package(InstalledRPM, application)]:
+        api.current_logger().info("Application {0} is installed.".format(app))
+        installed.append(app)
+    api.current_logger().info("----------------------------------")
+
+    return installed
--- a/repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
@@ -1,0 +1,28 @@
+"""
+Actor to check if KDE and/or GNOME are installed
+Author: Jan Beran
+Email: jaberan@redhat.com
+"""
+
+from leapp.actors import Actor
+from leapp.tags import IPUWorkflowTag, ChecksPhaseTag
+from leapp.reporting import Report
+from leapp.libraries.actor.library import check_kde_gnome
+
+
+class CheckKdeGnome(Actor):
+    """
+    Actor will check whether KDE is installed together with GNOME desktop to inform whether we can
+    inhibit the upgrade process. When both are installed, we need to inform the user that KDE will
+    be removed and GNOME will be used instead. If only KDE is installed, we want to inhibit
+    the upgrade process otherwise the user will end up without a desktop.
+    Also if both are installed, but KDE is not used as default desktop, we further check most common
+    KDE apps to inform the user that those will be removed in case the user is using them.
+    """
+    name = 'check_kde_gnome'
+    consumes = ()
+    produces = (Report,)
+    tags = (IPUWorkflowTag, ChecksPhaseTag)
+
+    def process(self):
+        check_kde_gnome()
--- a/repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
@@ -5,9 +5,11 @@ Email: jaberan@redhat.com
 """
 
 from leapp.actors import Actor
-from leapp.tags import IPUWorkflowTag, ChecksPhaseTag
 from leapp.reporting import Report
+from leapp.tags import IPUWorkflowTag, ChecksPhaseTag
 from leapp.libraries.actor.library import check_kde_gnome
+from leapp.models import (InstalledRPM, InstalledDesktopsFacts,
+                          InstalledKdeAppsFacts)
 
 
 class CheckKdeGnome(Actor):
@@ -18,11 +20,9 @@ class CheckKdeGnome(Actor):
     inhibit the upgrade process. When both are installed, we need to inform the user that KDE will
     be removed and GNOME will be used instead. If only KDE is installed, we want to inhibit
     the upgrade process otherwise the user will end up without a desktop.
-    Also if both are installed, but KDE is not used as default desktop, we further check most common
-    KDE apps to inform the user that those will be removed in case the user is using them.
     """
     name = 'check_kde_gnome'
-    consumes = ()
+    consumes = (InstalledRPM, InstalledDesktopsFacts, InstalledKdeAppsFacts)
     produces = (Report,)
     tags = (IPUWorkflowTag, ChecksPhaseTag)
 
--- a/repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
@@ -12,6 +12,8 @@ from leapp.libraries.actor.library import check_kde_gnome
 
 class CheckKdeGnome(Actor):
     """
+    Checks whether KDE is installed
+
     Actor will check whether KDE is installed together with GNOME desktop to inform whether we can
     inhibit the upgrade process. When both are installed, we need to inform the user that KDE will
     be removed and GNOME will be used instead. If only KDE is installed, we want to inhibit
--- a/repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
@@ -20,6 +20,8 @@ class CheckKdeGnome(Actor):
     inhibit the upgrade process. When both are installed, we need to inform the user that KDE will
     be removed and GNOME will be used instead. If only KDE is installed, we want to inhibit
     the upgrade process otherwise the user will end up without a desktop.
+    Note: The Package Evolution Service data makes sure the KDE-related packages are removed in the
+    dnf upgrade transaction.
     """
     name = 'check_kde_gnome'
     consumes = (InstalledRPM, InstalledDesktopsFacts, InstalledKdeAppsFacts)
--- a/repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
@@ -113,7 +113,7 @@ def check_kde_gnome():
                     reporting.Severity(reporting.Severity.MEDIUM),
                     reporting.Tags([
                         reporting.Tags.UPGRADE_PROCESS
-                    ])
+                    ])])
             else:
                 api.current_logger().info("GNOME used as default session. Continuing with the upgrade.")
         api.current_logger().info("----------------------------------")
--- a/repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
@@ -1,74 +1,12 @@
-import os
-import subprocess
-import time
-
 from leapp import reporting
 from leapp.libraries.stdlib import api
-
-ONE_MONTH = 2592000  # Number of seconds in one month
-
-
-def is_executable(path):
-    """
-    Checks if path exists, if it is file and if is executable.
-    """
-    return os.path.exists(path) and os.path.isfile(path) and os.access(path, os.X_OK)
-
-
-def get_xsession(path):
-    """
-    Gets current XSession.
-
-    If there is more than one definition of xsession for some reason,
-    function returns last definition. Return empty string if no XSession found in file given
-    by path (any reason including bad path etc.)
-    """
-    default_xsession = ""
-    if not isinstance(path, str):
-        return default_xsession
-    if not (os.path.exists(path) and os.path.isfile(path)):  # Bad path - in container for example
-        return default_xsession
-    with open(path, "r") as f:
-        for line in f.readlines():
-            if "xsession" in line.lower():
-                default_xsession = line.split("=")[1].lower()
-    return default_xsession
-
-
-def check_app_in_use(app):
-    """
-    Method return True if application was used in last month, False in other cases.
-    """
-    path = "{0}/.kde/share/config/{1}rc".format(os.environ.get("HOME"), app)
-    if os.path.isfile(path):
-        last_modified = os.stat(path).st_mtime
-        # Application is considered actively used, if it has been used in last month.
-        return last_modified >= int(time.time() - ONE_MONTH)
-    return False
-
-
-def is_installed(app):
-    """
-    Wrapper for "rpm -q <app>" command
-
-    Return value: True if application is found,
-    False in other cases.
-    Output of rpm command is not supressed.
-    """
-    return True if subprocess.call(["rpm", "-q", app]) == 0 else False
+from leapp.models import InstalledDesktopsFacts, InstalledKdeAppsFacts
 
 
 def check_kde_gnome():
-    apps_in_use = 0
-    api.current_logger().info("  Detecting desktop environments  ")
-    api.current_logger().info("==================================")
-
-    # Detect installed desktops by their startup files
-    kde_desktop_installed = is_executable("/usr/bin/startkde")
-    gnome_desktop_installed = is_executable("/usr/bin/gnome-session")
-    api.current_logger().info("* KDE installed: {0}".format(kde_desktop_installed))
-    api.current_logger().info("* Gnome installed: {0}".format(gnome_desktop_installed))
-    api.current_logger().info("----------------------------------")
+    desktopFacts = next(api.consume(InstalledDesktopsFacts))
+    kde_desktop_installed = desktopFacts.kde_installed
+    gnome_desktop_installed = desktopFacts.gnome_installed
 
     # No desktop installed, we don't even care about apps as they are most likely not used or even installed
     if not kde_desktop_installed and not gnome_desktop_installed:
@@ -83,7 +21,7 @@ def check_kde_gnome():
             # We cannot continue with the upgrade process
             reporting.create_report([
                 reporting.Title("Cannot upgrade because there is no other desktop than KDE installed."),
-                reporting.Summary("With only KDE installed, there would be no other desktop after upgrade."),
+                reporting.Summary("With only KDE installed, there would be no other desktop env. after upgrade."),
                 reporting.Severity(reporting.Severity.HIGH),
                 reporting.Tags([
                     reporting.Tags.UPGRADE_PROCESS
@@ -95,66 +33,35 @@ def check_kde_gnome():
                     hint="Install GNOME desktop to be able to upgrade.")
                 ])
             return
-
-        # Assume GNOME is installed in this state
-
-        user = os.environ.get("USER")
-        default_xsession = get_xsession("/var/lib/AccountsService/users/{0}".format(user))
-        if not default_xsession:
-            api.current_logger().warn("Unable to detect default session.")
         else:
-            if "plasma" in default_xsession:  # using in because there can be some white spaces
-                api.current_logger().info("KDE used as default session.")
-                api.current_logger().info("Upgrade can be performed, but KDE desktop will"
-                                          " be removed in favor of GNOME")
-                reporting.create_report([
-                    reporting.Title("Upgrade can be performed, but KDE will be uninstalled."),
-                    reporting.Summary("KDE has to be uninstalled in favor of GNOME to perform the upgrade."),
-                    reporting.Severity(reporting.Severity.MEDIUM),
-                    reporting.Tags([
-                        reporting.Tags.UPGRADE_PROCESS
-                    ])])
-            else:
-                api.current_logger().info("GNOME used as default session. Continuing with the upgrade.")
+            # Assume both GNOME and KDE are installed in this state
+            api.current_logger().info("Upgrade can be performed, but KDE desktop will"
+                                      " be removed in favor of GNOME")
+            reporting.create_report([
+                reporting.Title("Upgrade can be performed, but KDE will be uninstalled."),
+                reporting.Summary("KDE has to be uninstalled in favor of GNOME to perform the upgrade."),
+                reporting.Severity(reporting.Severity.MEDIUM),
+                reporting.Tags([
+                    reporting.Tags.UPGRADE_PROCESS
+                ])])
         api.current_logger().info("----------------------------------")
 
-    # At this state we can assume that KDE desktop as such is not installed or used and we just need to
-    # detect whether any KDE/Qt app is actively used to inform user that the application will be removed
-    # during the upgrade process
+    # At this state we just need to detect whether any KDE/Qt app is installed to inform user
+    # that the application will be removed during the upgrade process. No matter if KDE is installed
+    # or not.
 
-    base_kde_apps = ("kde-baseapps",
-                     "okular",
-                     "ark",
-                     "kdepim",
-                     "konsole",
-                     "gwenview",
-                     "kdenetwork",
-                     "kate", "kwrite")
-    api.current_logger().info("  Detecting installed KDE apps  ")
-    api.current_logger().info("================================")
-
-    for app in base_kde_apps:
-        if is_installed(app):
-            if check_app_in_use(app):
-                api.current_logger().info("Application {0} is actively used".format(app))
-                apps_in_use += 1
-        api.current_logger().info("* {0} {1} installed.".format(app, "is" if is_installed(app) else "is not"))
-
-    api.current_logger().info("----------------------------------")
-
-    if apps_in_use > 0:
-        api.current_logger().info("KDE apps in use detected.")
+    KDEAppsFacts = next(api.consume(InstalledKdeAppsFacts))
+    apps_in_use = KDEAppsFacts.installed_apps
+    if apps_in_use:
+        # upgrade can be performed, but user will loose KDE apps
+        api.current_logger().info("Installed KDE/Qt apps detected.")
         reporting.create_report([
-            reporting.Title("Upgrade can be performed, but KDE apps will be uninstalled."),
-            reporting.Summary("KDE apps will be removed to perform the upgrade."),
+            reporting.Title("Upgrade can be performed, but KDE/Qt apps will be uninstalled."),
+            reporting.Summary("KDE/Qt apps will be removed to perform the upgrade."),
             reporting.Severity(reporting.Severity.MEDIUM),
             reporting.Tags([
                 reporting.Tags.UPGRADE_PROCESS
-            ]),
-            reporting.Remediation(
-                hint="KDE apps has to be removed, no other solution is possible.")
-            ])
-        # upgrade can be performed, but user will loose KDE desktop in favor of GNOME desktop
+            ])])
     else:
         api.current_logger().info("No KDE app in use detected.")
-    # upgrade can be performed
+        # upgrade can be performed
--- a/repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
@@ -1,0 +1,158 @@
+import os
+import subprocess
+import time
+
+from leapp import reporting
+from leapp.libraries.stdlib import api
+
+ONE_MONTH = 2592000  # Number of seconds in one month
+
+
+def is_executable(path):
+    """
+    Checks if path exists, if it is file and if is executable.
+    """
+    return os.path.exists(path) and os.path.isfile(path) and os.access(path, os.X_OK)
+
+
+def get_xsession(path):
+    """
+    Gets current XSession. If there is more than one definition of xsession for some reason,
+     function returns last definition. Return empty string if no XSession found in file given
+     by path (any reason including bad path etc.)
+    """
+    default_xsession = ""
+    if not isinstance(path, str):
+        return default_xsession
+    if not (os.path.exists(path) and os.path.isfile(path)):  # Bad path - in container for example
+        return default_xsession
+    with open(path, "r") as f:
+        for line in f.readlines():
+            if "xsession" in line.lower():
+                default_xsession = line.split("=")[1].lower()
+    return default_xsession
+
+
+def check_app_in_use(app):
+    """
+    Method return True if application was used in last month, False in other cases.
+    """
+    path = "{0}/.kde/share/config/{1}rc".format(os.environ.get("HOME"), app)
+    if os.path.isfile(path):
+        last_modified = os.stat(path).st_mtime
+        # Application is considered actively used, if it has been used in last month.
+        return last_modified >= int(time.time() - ONE_MONTH)
+    return False
+
+
+def is_installed(app):
+    """
+    Wrapper for "rpm -q <app>" command, returns True if application is found,
+    False in other cases. Output of rpm command is not supressed.
+    """
+    return True if subprocess.call(["rpm", "-q", app]) == 0 else False
+
+
+def check_kde_gnome():
+    apps_in_use = 0
+    api.current_logger().info("  Detecting desktop environments  ")
+    api.current_logger().info("==================================")
+
+    # Detect installed desktops by their startup files
+    kde_desktop_installed = is_executable("/usr/bin/startkde")
+    gnome_desktop_installed = is_executable("/usr/bin/gnome-session")
+    api.current_logger().info("* KDE installed: {0}".format(kde_desktop_installed))
+    api.current_logger().info("* Gnome installed: {0}".format(gnome_desktop_installed))
+    api.current_logger().info("----------------------------------")
+
+    # No desktop installed, we don't even care about apps as they are most likely not used or even installed
+    if not kde_desktop_installed and not gnome_desktop_installed:
+        api.current_logger().info("No desktop installed. Continuing with the upgrade.")
+        return
+
+    if kde_desktop_installed:
+        api.current_logger().info("KDE desktop is installed. Checking what we can do about it.")
+        if not gnome_desktop_installed:
+            api.current_logger().error("Cannot perform the upgrade because there is\
+             no other desktop than KDE installed.")
+            # We cannot continue with the upgrade process
+            reporting.create_report([
+                reporting.Title("Cannot upgrade because there is no other desktop than KDE installed."),
+                reporting.Summary("With only KDE installed, there would be no other desktop after upgrade."),
+                reporting.Severity(reporting.Severity.HIGH),
+                reporting.Tags([
+                    reporting.Tags.UPGRADE_PROCESS
+                ]),
+                reporting.Flags([
+                    reporting.Flags.INHIBITOR
+                ]),
+                reporting.Remediation(
+                    hint="Install GNOME desktop to be able to upgrade.")
+                ])
+            return
+
+        # Assume GNOME is installed in this state
+
+        user = os.environ.get("USER")
+        default_xsession = get_xsession("/var/lib/AccountsService/users/{0}".format(user))
+        if not default_xsession:
+            api.current_logger().warn("Unable to detect default session.")
+        else:
+            if "plasma" in default_xsession:  # using in because there can be some white spaces
+                api.current_logger().info("KDE used as default session.")
+                api.current_logger().info("Upgrade can be performed, but KDE desktop will\
+                 be removed in favor of GNOME")
+                reporting.create_report([
+                    reporting.Title("Upgrade can be performed, but KDE will be uninstalled."),
+                    reporting.Summary("KDE has to be uninstalled in favor of GNOME to perform the upgrade."),
+                    reporting.Severity(reporting.Severity.MEDIUM),
+                    reporting.Tags([
+                        reporting.Tags.UPGRADE_PROCESS
+                    ]),
+                    reporting.Remediation(
+                        hint="KDE has to be removed, no other solution is possible.")
+                    ])
+            else:
+                api.current_logger().info("GNOME used as default session. Continuing with the upgrade.")
+        api.current_logger().info("----------------------------------")
+
+    # At this state we can assume that KDE desktop as such is not installed or used and we just need to
+    # detect whether any KDE/Qt app is actively used to inform user that the application will be removed
+    # during the upgrade process
+
+    base_kde_apps = ("kde-baseapps",
+                     "okular",
+                     "ark",
+                     "kdepim",
+                     "konsole",
+                     "gwenview",
+                     "kdenetwork",
+                     "kate", "kwrite")
+    api.current_logger().info("  Detecting installed KDE apps  ")
+    api.current_logger().info("================================")
+
+    for app in base_kde_apps:
+        if is_installed(app):
+            if check_app_in_use(app):
+                api.current_logger().info("Application {0} is actively used".format(app))
+                apps_in_use += 1
+        api.current_logger().info("* {0} {1} installed.".format(app, "is" if is_installed(app) else "is not"))
+
+    api.current_logger().info("----------------------------------")
+
+    if apps_in_use > 0:
+        api.current_logger().info("KDE apps in use detected.")
+        reporting.create_report([
+            reporting.Title("Upgrade can be performed, but KDE apps will be uninstalled."),
+            reporting.Summary("KDE apps will be removed to perform the upgrade."),
+            reporting.Severity(reporting.Severity.MEDIUM),
+            reporting.Tags([
+                reporting.Tags.UPGRADE_PROCESS
+            ]),
+            reporting.Remediation(
+                hint="KDE apps has to be removed, no other solution is possible.")
+            ])
+        # upgrade can be performed, but user will loose KDE desktop in favor of GNOME desktop
+    else:
+        api.current_logger().info("No KDE app in use detected.")
+    # upgrade can be performed
--- a/repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
@@ -17,9 +17,11 @@ def is_executable(path):
 
 def get_xsession(path):
     """
-    Gets current XSession. If there is more than one definition of xsession for some reason,
-     function returns last definition. Return empty string if no XSession found in file given
-     by path (any reason including bad path etc.)
+    Gets current XSession.
+
+    If there is more than one definition of xsession for some reason,
+    function returns last definition. Return empty string if no XSession found in file given
+    by path (any reason including bad path etc.)
     """
     default_xsession = ""
     if not isinstance(path, str):
@@ -47,8 +49,11 @@ def check_app_in_use(app):
 
 def is_installed(app):
     """
-    Wrapper for "rpm -q <app>" command, returns True if application is found,
-    False in other cases. Output of rpm command is not supressed.
+    Wrapper for "rpm -q <app>" command
+
+    Return value: True if application is found,
+    False in other cases.
+    Output of rpm command is not supressed.
     """
     return True if subprocess.call(["rpm", "-q", app]) == 0 else False
 
@@ -73,8 +78,8 @@ def check_kde_gnome():
     if kde_desktop_installed:
         api.current_logger().info("KDE desktop is installed. Checking what we can do about it.")
         if not gnome_desktop_installed:
-            api.current_logger().error("Cannot perform the upgrade because there is\
-             no other desktop than KDE installed.")
+            api.current_logger().error("Cannot perform the upgrade because there is"
+                                       " no other desktop than KDE installed.")
             # We cannot continue with the upgrade process
             reporting.create_report([
                 reporting.Title("Cannot upgrade because there is no other desktop than KDE installed."),
@@ -100,17 +105,14 @@ def check_kde_gnome():
         else:
             if "plasma" in default_xsession:  # using in because there can be some white spaces
                 api.current_logger().info("KDE used as default session.")
-                api.current_logger().info("Upgrade can be performed, but KDE desktop will\
-                 be removed in favor of GNOME")
+                api.current_logger().info("Upgrade can be performed, but KDE desktop will"
+                                          " be removed in favor of GNOME")
                 reporting.create_report([
                     reporting.Title("Upgrade can be performed, but KDE will be uninstalled."),
                     reporting.Summary("KDE has to be uninstalled in favor of GNOME to perform the upgrade."),
                     reporting.Severity(reporting.Severity.MEDIUM),
                     reporting.Tags([
                         reporting.Tags.UPGRADE_PROCESS
-                    ]),
-                    reporting.Remediation(
-                        hint="KDE has to be removed, no other solution is possible.")
                     ])
             else:
                 api.current_logger().info("GNOME used as default session. Continuing with the upgrade.")
--- a/repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
@@ -21,7 +21,10 @@ def check_kde_gnome():
             # We cannot continue with the upgrade process
             reporting.create_report([
                 reporting.Title("Cannot upgrade because there is no other desktop than KDE installed."),
-                reporting.Summary("With only KDE installed, there would be no other desktop env. after upgrade."),
+                reporting.Summary("The KDE desktop environment is not available on RHEL 8. "
+                                  "The KDE-related packages will be uninstalled during the upgrade and because "
+                                  "the only currently installed desktop environment is KDE, there will be no "
+                                  "other desktop environment after upgrade."),
                 reporting.Severity(reporting.Severity.HIGH),
                 reporting.Tags([
                     reporting.Tags.UPGRADE_PROCESS
@@ -30,20 +33,22 @@ def check_kde_gnome():
                     reporting.Flags.INHIBITOR
                 ]),
                 reporting.Remediation(
-                    hint="Install GNOME desktop to be able to upgrade.")
+                    hint="Install GNOME desktop to be able to upgrade.",
+                    commands=[['yum', '-y', 'groupinstall', '"Server with GUI"']])
                 ])
             return
-        else:
-            # Assume both GNOME and KDE are installed in this state
-            api.current_logger().info("Upgrade can be performed, but KDE desktop will"
-                                      " be removed in favor of GNOME")
-            reporting.create_report([
-                reporting.Title("Upgrade can be performed, but KDE will be uninstalled."),
-                reporting.Summary("KDE has to be uninstalled in favor of GNOME to perform the upgrade."),
-                reporting.Severity(reporting.Severity.MEDIUM),
-                reporting.Tags([
-                    reporting.Tags.UPGRADE_PROCESS
-                ])])
+
+        # Assume both GNOME and KDE are installed in this state
+        api.current_logger().info("Upgrade can be performed, but KDE desktop will"
+                                  " be removed in favor of GNOME")
+        reporting.create_report([
+            reporting.Title("Upgrade can be performed, but KDE will be uninstalled."),
+            reporting.Summary("The KDE desktop environment is not available on RHEL 8. KDE will be uninstalled "
+                              "in favor of GNOME during the upgrade."),
+            reporting.Severity(reporting.Severity.MEDIUM),
+            reporting.Tags([
+                reporting.Tags.UPGRADE_PROCESS
+            ])])
         api.current_logger().info("----------------------------------")
 
     # At this state we just need to detect whether any KDE/Qt app is installed to inform user
@@ -51,13 +56,14 @@ def check_kde_gnome():
     # or not.
 
     KDEAppsFacts = next(api.consume(InstalledKdeAppsFacts))
-    apps_in_use = KDEAppsFacts.installed_apps
-    if apps_in_use:
+    if KDEAppsFacts.installed_apps:
         # upgrade can be performed, but user will loose KDE apps
         api.current_logger().info("Installed KDE/Qt apps detected.")
         reporting.create_report([
             reporting.Title("Upgrade can be performed, but KDE/Qt apps will be uninstalled."),
-            reporting.Summary("KDE/Qt apps will be removed to perform the upgrade."),
+            reporting.Summary("The KDE desktop environment is not available on RHEL 8. "
+                              "All the KDE/Qt apps will be removed during the upgrade, including but not limited "
+                              "to:\n- {0}".format("\n- ".join(KDEAppsFacts.installed_apps))),
             reporting.Severity(reporting.Severity.MEDIUM),
             reporting.Tags([
                 reporting.Tags.UPGRADE_PROCESS
--- a/repos/system_upgrade/el7toel8/actors/checkkdegnome/tests/unit_test.py
+++ b/repos/system_upgrade/el7toel8/actors/checkkdegnome/tests/unit_test.py
@@ -1,0 +1,175 @@
+from leapp.libraries.actor import library
+from leapp.libraries.common.testutils import create_report_mocked
+from leapp import reporting
+
+
+def is_executable_only_gnome(path):
+    return path == "/usr/bin/gnome-session"
+
+
+def is_executable_only_kde(path):
+    return path == "/usr/bin/startkde"
+
+
+def check_app_in_use_mocked(app):
+    #  Only one app in use
+    return app == "okular"
+
+
+def get_xsession_gnome(path):
+    return "gnome"
+
+
+def get_xsession_kde(path):
+    return "plasma"
+
+
+def test_no_desktop(monkeypatch):
+    """
+    Scenario: There is no desktop.
+    Expected behavior: No report
+    """
+    #  Desktop presence is found by trying to execute gnome-session or startkde
+    monkeypatch.setattr(library, "is_executable", lambda x: False)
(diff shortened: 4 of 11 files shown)


Honza0297:
Thank you for fast answer and explanation what is wrong!
I'll fix it in next few hours.

Honza0297:
All tests passed now. Force-push is used because Travis fixes were split into 5 commits as errors were occurring, so they were squashed into one commit for readability.

vojtechsokol:
@Honza0297 You can run `make lint`, `make test` or `make test_no_lint` (check output of `make` for help on possible targets) to check that everything goes well before you push.

Honza0297:
@vojtechsokol Good point, thanks! It can be handy in future.

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
```diff
@@ -0,0 +1,137 @@
...
+    inhibit the upgrade process. When both are installed, we need to inform the user that KDE will be
+    removed and GNOME will be used instead. If only KDE is installed, we want to inhibit
+    the upgrade process otherwise the user will end up without a desktop.
+    Also if both are installed, but KDE is not used as default desktop, we further check most common
+    KDE apps to inform the user that those will be removed in case the user is using them.
+    """
+    name = 'check_kde_gnome'
+    consumes = ()
+    produces = (CanUpgrade,)
+    tags = (RhelUpgradeTag,)
```
bocekm: The tags are as of now used to tell Leapp framework in which workflow and phase of the workflow the actor should be executed. 
Since this actor is for upgrades, use `IPUWorkflowTag`. IPU stands for in-place upgrade.
Since this is a check actor, use `ChecksPhaseTag` here and remove the new RhelUpgradeTag with its file please.
For more details, see https://leapp.readthedocs.io/en/latest/actor-rhel7-to-rhel8.html#including-an-actor-in-the-rhel-7-to-8-upgrade-process.

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
```diff
@@ -0,0 +1,137 @@
...
+        if not kde_desktop_installed and not gnome_desktop_installed:
+            self.log.info("INFO: No desktop installed. Continuing with the upgrade.")
+            self.produce(CanUpgrade(canUpgrade=0))
+            return
+
+        if kde_desktop_installed:
+            self.log.info("INFO: KDE desktop is installed. Checking what we can do about it.")
+            if not gnome_desktop_installed:
+                self.log.info("ERROR: Cannot perform the upgrade because there is no other desktop installed.")
+                self.produce(CanUpgrade(canUpgrade=1))
```
bocekm: Please do not create a message which nobody consumes. You can remove canUpgrade message model with the topic RhelUpgradeInfoTopic as well.
Instead of that, use `create_report` function provided by the leapp framework. To not allow the upgrade, use `reporting.Flags.INHIBITOR`. See example here: https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/actors/checkfirewalld/actor.py#L35.

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
```diff
@@ -0,0 +1,137 @@
...
+
+        self.log.info("----------------------------------")
+
+        if apps_in_use > 0:
+            self.log.info("INFO: KDE apps in use detected.")
+            self.produce(CanUpgrade(canUpgrade=2))
+            # upgrade can be performed, but user will loose KDE desktop in favor of GNOME desktop
+        else:
+            self.log.info("INFO: No KDE app in use detected.")
+            self.produce(CanUpgrade(canUpgrade=0))
```
bocekm: There's no need to send any message if the actor's checks passed with no action to the user. The log.info is sufficient.

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
```diff
@@ -0,0 +1,137 @@
...
+            if self.is_installed(app) == _ret_OK:  # return code of "rpm -q app" command
+                self.log.info("* {0} is installed: True".format(app))
+                self.check_app_in_use(app)
+            else:
+                self.log.info("* {0} is installed: False".format(app))
+
+        self.log.info("----------------------------------")
+
+        if apps_in_use > 0:
+            self.log.info("INFO: KDE apps in use detected.")
```
bocekm: No need to prepend the log message with the log level like "INFO: ". That's being added by the logging handler automatically.

bocekm:
@Honza0297, please talk to your subsystem QEs that we'll need an automated end to end (e2e) test for this actor. They can develop that by using our internal testing framework developed by @AloisMahdal, so please point them to him or his manager amarecek for any questions related to e2e tests.

Also, we encourage all actor developers to write unit tests: https://leapp.readthedocs.io/en/latest/unit-testing.html

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
```diff
@@ -0,0 +1,137 @@
...
+
+            user = os.environ.get("USER")
+            default_xsession = self.get_xsession("/var/lib/AccountsService/users/{0}".format(user))
+            if not default_xsession:
+                self.log.info("WARN: Unable to detect default session.")
+            else:
+                if "plasma" in default_xsession:  # using in because there can be some white spaces
+                    self.log.info("INFO: KDE used as default session.")
+                    self.log.info("INFO: Upgrade can be performed, but KDE desktop will be removed in favor of GNOME")
+                    self.produce(CanUpgrade(canUpgrade=2))
```
Honza0297: @bocekm What is the best way how to handle the state when upgrade can be performed, but KDE desktop will be removed? Just log it (and maybe set the log level to "warning") or give some stronger feedback? Thank you for the answer!

bocekm: I'd create an informative-only report message (again, using framework function `create_report`) with low or medium severity (`reporting.Severity(reporting.Severity.MEDIUM)`). With no remediation (`reporting.Remediation`) as this has no solution.

bocekm: Basically, expect that customers will not read the log messages in 99% of cases. Whatever you want the customer to know needs to land in the pre-upgrade report which we generate in both plaintext and HTML version.

Honza0297: Ok, I will do that this way. Thanks!

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
```diff
@@ -0,0 +1,167 @@
...
+                    reporting.Tags.UPGRADE_PROCESS
+                ]),
+                reporting.Remediation(
+                    hint="KDE apps has to be removed, no other solution is possible.")
+                ])
+            # upgrade can be performed, but user will loose KDE desktop in favor of GNOME desktop
+        else:
+            self.log.info("No KDE app in use detected.")
+            # upgrade can be performed
+        return
```
fernflower: Unnecessary return

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
```diff
@@ -0,0 +1,167 @@
...
+                         "kdepim",
+                         "konsole",
+                         "gwenview",
+                         "kdenetwork",
+                         "kate", "kwrite")
+        self.log.info("  Detecting installed KDE apps  ")
+        self.log.info("================================")
+
+        for app in base_kde_apps:
+            if self.is_installed(app) == _ret_OK:  # return code of "rpm -q app" command
```
fernflower: Having '0' inplace instead of _ret_OK makes more sense imho. Also I'd rather move the whole logic of check into is_installed method and have it return True/False and not an exit code.

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
```diff
@@ -0,0 +1,167 @@
...
+                self.log.info("* {0} is installed: True".format(app))
+                self.check_app_in_use(app)
+            else:
+                self.log.info("* {0} is installed: False".format(app))
```
fernflower: Can be moved outside of the block if is_installed returns a boolean

self.log.info("* {} {} installed".format(app, 'is' if self.is_installed(app) else 'is not'))

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
```diff
@@ -0,0 +1,167 @@
...
+                if "xsession" in line.lower():
+                    default_xsession = line.split("=")[1].lower()
+        return default_xsession
+
+    def check_app_in_use(self, app):
+        """
+        Method return True if application was used in last month, False in other cases.
+        """
+        path = "{0}/.kde/share/config/{1}rc".format(os.environ.get("HOME"), app)
+        if os.path.exists(path) and os.path.isfile(path):
```
fernflower: os.path.isfile will check for path existance iirc

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
```diff
@@ -0,0 +1,167 @@
...
+            last_modified = os.stat(path).st_mtime
+            if last_modified >= int(time.time() - 2592000):
```
fernflower: Sorry, not getting it, what is 2592000? Please don't use magic numbers inside code unless they are self speaking like 0 return code

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
```diff
@@ -0,0 +1,167 @@
...
+                self.log.info("Application {0} is actively used".format(app))
```
fernflower: Nit: I'd move the logging out of check methods (have it in the main processing one only) and rewrite as
`return last_modified >= int(time.time() - 2592000)` (just please no magic numbers)

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
```diff
@@ -0,0 +1,167 @@
...
+            return
+
+        if kde_desktop_installed:
+            self.log.info("KDE desktop is installed. Checking what we can do about it.")
+            if not gnome_desktop_installed:
+                self.log.error("Cannot perform the upgrade because there is no other desktop than KDE installed.")
+                # We cannot continue with the upgrade process
+                create_report([
+                    reporting.Title("Cannot upgrade because there is no other desktop than KDE installed."),
+                    reporting.Summary("With only KDE installed, there would be no other desktop  after upgrade."),
```
fernflower: Nit: 2 whitespaces

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
```diff
@@ -0,0 +1,167 @@
...
+        for app in base_kde_apps:
+            if self.is_installed(app) == _ret_OK:  # return code of "rpm -q app" command
+                self.log.info("* {0} is installed: True".format(app))
+                self.check_app_in_use(app)
+            else:
+                self.log.info("* {0} is installed: False".format(app))
+
+        self.log.info("----------------------------------")
+
+        if apps_in_use > 0:
```
fernflower: Sorry for being blind, where is the apps_in_use incrementation?

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
Honza0297: The magic is that I consider application as actively used if it has been used during last month. 
So `int(time.time() - 2592000)` means "more or less before one month". I'll change it to variable with appropriate name.

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
Honza0297: Lost in transcription from bash. Thanks!!!

Honza0297:
@bocekm  @fernflower  Almost all of your recommendations are included, excluding tests - actually I am not sure how to test it meaningfully without actually installing KDE + apps, GNOME and its combinations. As for e2e test, because I am my own SST QE it will take a while. :) 

Also,  I cannot run my actor locally, because I have similar issue to this: [https://github.com/oamg/leapp-repository/issues/124](url)

examon:
leapp-ci build

Honza0297:
@examon Are there any logs? I cannot run it because of the issue I mentioned above ([https://github.com/oamg/leapp-repository/issues/124](https://github.com/oamg/leapp-repository/issues/124)) so I have no clue what is wrong.

examon:
leapp-ci build

bocekm:
@Honza0297, could you please ping me on IRC (#leapp, nick mbocek) and we can discuss details around the testing, or even meet if you're in Brno.

bocekm:
@Honza0297, by unit tests I meant testing either functions or methods by executing them with fabricated inputs and observe if the outputs are as expected.
The unit tests should ideally not rely on anything being installed except python and its libraries like pytest, and they should definitely not alter the system on which you execute the them.
> actually installing KDE + apps, GNOME and its combinations"

.. that should actually the the aim of the end to end test.

We use [monkeypatch](https://docs.pytest.org/en/latest/monkeypatch.html) from the pytest testing framework to limit the test to the functionality of the function/method under test only.

To be able to properly unit test the code, I'd start with:
- for the reasons written [here](https://leapp.readthedocs.io/en/latest/unit-testing.html#testing-actors-that-modify-the-os), move all the code from `process()` to the actor's library to be able to monkeypatch methods (like `open()`, `os.stat()` and others).
- split the `process()` method into multiple smaller methods

You can read more about python unit testing for example here: https://realpython.com/python-testing/#unit-tests-vs-integration-tests

Honza0297:
@bocekm  Sorry for not responding, I was not in the office. 
Thanks for the explanation, end-to-end test should be done by OAMG-QE (thanks them!), unit tests will be made during this week.

bocekm:
Please remove the `repos/system_upgrade/el7toel8/.leapp/leapp.db` from the PR.

bocekm:
It's already fixed bug: https://github.com/oamg/leapp-repository/pull/311. Please rebase.

Honza0297:
@bocekm Rebase completed. I also squashed all my commits into one for better commit history.

vojtechsokol:
leapp-ci build

vojtechsokol:
e2e tests

repos/system_upgrade/el7toel8/actors/checkkdegnome/tests/unit_test.py
```diff
@@ -0,0 +1,175 @@
...
+    monkeypatch.setattr(library, "check_app_in_use", check_app_in_use_mocked)
+    monkeypatch.setattr(reporting, "create_report", create_report_mocked())
+
+    library.check_kde_gnome()
+
+    assert reporting.create_report.called == 1
+    #  Do not even touch the string! It is ugly, but because of max line lenght it has to
+    #  be written on two lines.
+    assert reporting.create_report.report_fields["title"] == "Upgrade can be\
+ performed, but KDE apps will be uninstalled."
```
pirat89: Pleases, do not use line wrap like this. Use rather like this:
```python
title_field = "Upgrade can be performed, but KDE apps will be uninstalled."
assert reporting.create_report.report_fields["title"] == title_field
```

This is on more places so change it please everywhere.

(just quick review; I will finish it next week;)

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,158 @@
...
+
+        user = os.environ.get("USER")
+        default_xsession = get_xsession("/var/lib/AccountsService/users/{0}".format(user))
+        if not default_xsession:
+            api.current_logger().warn("Unable to detect default session.")
+        else:
+            if "plasma" in default_xsession:  # using in because there can be some white spaces
+                api.current_logger().info("KDE used as default session.")
+                api.current_logger().info("Upgrade can be performed, but KDE desktop will\
+                 be removed in favor of GNOME")
```
pirat89: Again, instead of wrapping like that:
```python
api.current_logger().info("Upgrade can be performed, but KDE desktop will"
                                        "be removed in favor of GNOME")
```

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class CheckKdeGnome(Actor):
+    """
+    Actor will check whether KDE is installed together with GNOME desktop to inform whether we can
+    inhibit the upgrade process. When both are installed, we need to inform the user that KDE will
+    be removed and GNOME will be used instead. If only KDE is installed, we want to inhibit
+    the upgrade process otherwise the user will end up without a desktop.
+    Also if both are installed, but KDE is not used as default desktop, we further check most common
+    KDE apps to inform the user that those will be removed in case the user is using them.
```
pirat89: please keep the format
```
One line title

Description on multiple lines when
needed
```

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,158 @@
...
+    Checks if path exists, if it is file and if is executable.
+    """
+    return os.path.exists(path) and os.path.isfile(path) and os.access(path, os.X_OK)
+
+
+def get_xsession(path):
+    """
+    Gets current XSession. If there is more than one definition of xsession for some reason,
+     function returns last definition. Return empty string if no XSession found in file given
+     by path (any reason including bad path etc.)
```
pirat89: The format again. Do not add addiotional empty space on other lines, I mean, do not do this
```
Something
 something
```

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,158 @@
...
+                    reporting.Flags.INHIBITOR
+                ]),
+                reporting.Remediation(
+                    hint="Install GNOME desktop to be able to upgrade.")
+                ])
+            return
+
+        # Assume GNOME is installed in this state
+
+        user = os.environ.get("USER")
```
pirat89: TODO: @pirat89 : check ipuconfig

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,158 @@
...
+                reporting.create_report([
+                    reporting.Title("Upgrade can be performed, but KDE will be uninstalled."),
+                    reporting.Summary("KDE has to be uninstalled in favor of GNOME to perform the upgrade."),
+                    reporting.Severity(reporting.Severity.MEDIUM),
+                    reporting.Tags([
+                        reporting.Tags.UPGRADE_PROCESS
+                    ]),
+                    reporting.Remediation(
+                        hint="KDE has to be removed, no other solution is possible.")
+                    ])
```
pirat89: It looks little confusing to use `hint`, regarding the fact we will (/should) remove it automatically during the upgrade and user is not required to remove it prior the upgrade. If RH has any documentation about that change, you could add URL into the report, so user can read more details about that.

repos/system_upgrade/el7toel8/actors/checkkdegnome/tests/unit_test.py
Honza0297: No doubt this is much more better. Changed.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
Honza0297: Changed everywhere.

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
Honza0297: Done.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
Honza0297: This is strange, I don't know how the extra space got there. Fixed, of course.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
Honza0297: As I looked into older conversations, it seems that there is no need to use hint or anything else. So I just removed the hint. Is it acceptable?

Honza0297:
@pirat89 Thanks for the review! I fixed all the points you mentioned - is it OK now?

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,158 @@
...
+        return last_modified >= int(time.time() - ONE_MONTH)
+    return False
+
+
+def is_installed(app):
+    """
+    Wrapper for "rpm -q <app>" command, returns True if application is found,
+    False in other cases. Output of rpm command is not supressed.
+    """
+    return True if subprocess.call(["rpm", "-q", app]) == 0 else False
```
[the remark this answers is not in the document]
fernflower: I also have an impression that you shouldn't use subprocess library unless the leapp context.call one doesn't provide the requested feature. @pirat89 @vinzenz ?

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
fernflower: @Honza0297 imho you sure can drop hint as the error message is self-explanatory. Up to you

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,158 @@
...
+                     "ark",
+                     "kdepim",
+                     "konsole",
+                     "gwenview",
+                     "kdenetwork",
+                     "kate", "kwrite")
+    api.current_logger().info("  Detecting installed KDE apps  ")
+    api.current_logger().info("================================")
+
+    for app in base_kde_apps:
```
fernflower: Nit: 
`for app in [app for app in base_kde_apps if is_installed(app) and is_in_use(app)]`

instead of cycles at L135-L136

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,158 @@
...
+    if not (os.path.exists(path) and os.path.isfile(path)):  # Bad path - in container for example
+        return default_xsession
+    with open(path, "r") as f:
+        for line in f.readlines():
+            if "xsession" in line.lower():
+                default_xsession = line.split("=")[1].lower()
+    return default_xsession
+
+
+def check_app_in_use(app):
```
fernflower: I'd rename to is_app_in_use to exaggerate that it returns boolean

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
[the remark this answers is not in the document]
pirat89: There is obviously no need for subprocess nor our call here as the information is provided inside  `InstalledRedHatSignedRPM` (or `InstalledRPM`) and we have implemented the `has_package` function.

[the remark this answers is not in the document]
pirat89: @Honza0297 To be completely clear, actors in the ChecksPhase phase shouldn't "work with system". That means effectively that you should not use `rpm`, usually it means that you should not need to use subprocess / call in generic. Think about that as you should be able to do the check on completely different machine, just using data you already have provided from the previous phase. In your case, see e.g. `checkgrep` actor, using the `has_package` function.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
pirat89: Same issue as above. You should not rely on os.environ in the checksphase. From this point, we already provide set of `LEAPP_DEVEL` env vars in the Actor.configuration object. Not sure, whether we should extend this to get all variables or whether will be better to create actor that will provide information about the rest of variables. @vinzenz @fernflower @oamg/developers  - WDYT guys? I believe that definitely we should not create new actor that provide information just about one, two, specific envvars, but to have something really more generic. Not sure what we should propose as solution here.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,158 @@
...
+def get_xsession(path):
+    """
+    Gets current XSession. If there is more than one definition of xsession for some reason,
+     function returns last definition. Return empty string if no XSession found in file given
+     by path (any reason including bad path etc.)
+    """
+    default_xsession = ""
+    if not isinstance(path, str):
+        return default_xsession
+    if not (os.path.exists(path) and os.path.isfile(path)):  # Bad path - in container for example
```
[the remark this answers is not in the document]
pirat89: This should not be even called during the checksphase. You cannot check whether specific files exists on system, neither open them during the phase. When you need these data, please create new actor that will scan the system and produce message with all data you need in the check actor.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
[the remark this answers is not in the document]
Honza0297: @pirat89 As to the new actor, it should be run in `FactsCollectionPhase`, it means with `FactsPhaseTag`, right?

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
Honza0297: In this case, also HOME envar does not make any sense.  I will try to make a workaround.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,67 @@
...
+                reporting.Tags.UPGRADE_PROCESS
+            ])])
+        api.current_logger().info("----------------------------------")
+
+    # At this state we just need to detect whether any KDE/Qt app is installed to inform user
+    # that the application will be removed during the upgrade process. No matter if KDE is installed
+    # or not.
+
+    KDEAppsFacts = next(api.consume(InstalledKdeAppsFacts))
+    apps_in_use = KDEAppsFacts.installed_apps
```
bocekm: Is installed the same as in use?

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,67 @@
...
+                reporting.Remediation(
+                    hint="Install GNOME desktop to be able to upgrade.")
+                ])
+            return
+
+        # Assume both GNOME and KDE are installed in this state
+        api.current_logger().info("Upgrade can be performed, but KDE desktop will"
+                                  " be removed in favor of GNOME")
+        reporting.create_report([
+            reporting.Title("Upgrade can be performed, but KDE will be uninstalled."),
```
bocekm: Does the KDE uninstallation happen by itself without the need for leapp to initiate it?

repos/system_upgrade/el7toel8/actors/checkkdegnome/tests/unit_test.py
```diff
@@ -0,0 +1,97 @@
...
+def test_KDE_desktop_KDE_apps(current_actor_context):
+    """
+    "Inhibitor" flag in report expected.
+    """
+    current_actor_context.feed(KDE_desktop_env)
+    current_actor_context.feed(some_KDE_apps)
+    current_actor_context.run()
+    message = current_actor_context.consume(Report)[0]
+    assert "inhibitor" in message.report["flags"]
+    # assert [True for message in messages if "inhibitor" in message.report["flags"]]
```
bocekm: You can remove this dead code.

bocekm:
@Honza0297, the PR looks good. Unfortunately we did a change that requires all PRs to rebase. So please rebase yours as well.
@AloisMahdal, how does it look like with a QE coverage?

pirat89:
@Honza0297 Let's sync this tuesday. I apologize for late response.

bocekm:
Per @AloisMahdal, the QE tests are in works, but currently blocked by something. We won't wait with merging this PR for these tests.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
Honza0297: Yes, it was only a relic of previous version. Removed.

repos/system_upgrade/el7toel8/actors/checkkdegnome/tests/unit_test.py
Honza0297: Removed.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
Honza0297: No, it is assumed leapp will initiate it.

Honza0297:
@bocekm  Rebase completed, nitpicks fixed.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
bocekm: In that case, I would say that it should be part of this PR to tell leapp that the KDE pkgs are to be removed.
You would do that by producing `RpmTransactionTasks(to_remove=['kdepkg1', 'kdepkg2'])` in this actor.

@pirat89, @drehak, I fear that what I suggest above could cause a conflict in the transaction similar to https://github.com/oamg/leapp-repository/issues/177. But since the PESEventsScanner runs in the Facts phase, I see no way to deal with that. WDYT?

bocekm: As @pirat89 mentions, the PES data should actually take care of the removal, so disregard my point. I'll test that.

repos/system_upgrade/el7toel8/actors/checkkdeapps/libraries/library.py
```diff
@@ -0,0 +1,25 @@
...
+    installed = list()
+    base_kde_apps = ("kde-baseapps",
+                     "okular",
+                     "ark",
+                     "kdepim",
+                     "konsole",
+                     "gwenview",
+                     "kdenetwork",
+                     "kate",
+                     "kwrite")
```
bocekm: @Honza0297, is it OK if the upgrade passes just fine with message in the report when I have `kate-part` and `kde-runtime` installed?

bocekm:
With the following installed I'm not getting any inhibitor:
```
leapp-repository-0.9.0-0.202002170935Z.b7ebde8.PR_master.PR310.el7_6.noarch
leapp-repository-deps-0.9.0-0.202002170935Z.b7ebde8.PR_master.PR310.el7_6.noarch
kde-baseapps-4.10.5-6.el7.x86_64
```
No gnome apps installed. I'll test it further.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,66 @@
...
+                reporting.Summary("With only KDE installed, there would be no other desktop env. after upgrade."),
+                reporting.Severity(reporting.Severity.HIGH),
+                reporting.Tags([
+                    reporting.Tags.UPGRADE_PROCESS
+                ]),
+                reporting.Flags([
+                    reporting.Flags.INHIBITOR
+                ]),
+                reporting.Remediation(
+                    hint="Install GNOME desktop to be able to upgrade.")
```
bocekm:
```suggestion
                    hint="Install GNOME desktop to be able to upgrade.",
                    commands=[['yum', '-y', 'groupinstall', '"Server with GUI"']])
```

repos/system_upgrade/el7toel8/actors/getinstalleddesktops/libraries/library.py
```diff
@@ -0,0 +1,17 @@
...
+from leapp.libraries.common.rpms import has_package
+from leapp.models import InstalledRPM
+
+
+def get_installed_desktops():
+    api.current_logger().info("  Detecting desktop environments  ")
+    api.current_logger().info("==================================")
+
+    # Detect installed desktops by one of the base rpm packages
+    kde_desktop_installed = has_package(InstalledRPM, "plasma-workspace")
```
bocekm: Where can I get this package? It's not in the RHEL 7 base repo, and also not in Optional or Extras.

Also, it's not mentioned in the Package Evolution Service.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,66 @@
...
+    # that the application will be removed during the upgrade process. No matter if KDE is installed
+    # or not.
+
+    KDEAppsFacts = next(api.consume(InstalledKdeAppsFacts))
+    if KDEAppsFacts.installed_apps:
+        # upgrade can be performed, but user will loose KDE apps
+        api.current_logger().info("Installed KDE/Qt apps detected.")
+        reporting.create_report([
+            reporting.Title("Upgrade can be performed, but KDE/Qt apps will be uninstalled."),
+            reporting.Summary("KDE/Qt apps will be removed to perform the upgrade."),
```
bocekm:
```suggestion
            reporting.Summary("The KDE desktop environment is not available on RHEL 8. All the KDE/Qt apps will be removed during the upgrade, including but not limited to:\n- {0}".format("\n- ".join(KDEAppsFacts.installed_apps))),
```
I think the current message is not sufficient. It's just too generic:
```
Risk Factor: medium
Title: Upgrade can be performed, but KDE/Qt apps will be uninstalled.
Summary: KDE/Qt apps will be removed to perform the upgrade.
```
Please apply the coding standards to the proposed change (max 120 chars one line).

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,66 @@
...
+                reporting.Tags.UPGRADE_PROCESS
+            ])])
+        api.current_logger().info("----------------------------------")
+
+    # At this state we just need to detect whether any KDE/Qt app is installed to inform user
+    # that the application will be removed during the upgrade process. No matter if KDE is installed
+    # or not.
+
+    KDEAppsFacts = next(api.consume(InstalledKdeAppsFacts))
+    if KDEAppsFacts.installed_apps:
```
bocekm: This condition should be evaluated no matter whether GNOME is installed or not. Currently, if GNOME is not installed, this code is not reached.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,66 @@
...
+                    hint="Install GNOME desktop to be able to upgrade.")
+                ])
+            return
+
+        # Assume both GNOME and KDE are installed in this state
+        api.current_logger().info("Upgrade can be performed, but KDE desktop will"
+                                  " be removed in favor of GNOME")
+        reporting.create_report([
+            reporting.Title("Upgrade can be performed, but KDE will be uninstalled."),
+            reporting.Summary("KDE has to be uninstalled in favor of GNOME to perform the upgrade."),
```
bocekm:
```suggestion
            reporting.Summary("The KDE desktop environment is not available on RHEL 8. KDE will be uninstalled in favor of GNOME during the upgrade."),
```
Please apply the coding standards to the proposed change (max 120 chars one line).

repos/system_upgrade/el7toel8/actors/checkkdegnome/actor.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class CheckKdeGnome(Actor):
+    """
+    Checks whether KDE is installed
+
+    Actor will check whether KDE is installed together with GNOME desktop to inform whether we can
+    inhibit the upgrade process. When both are installed, we need to inform the user that KDE will
+    be removed and GNOME will be used instead. If only KDE is installed, we want to inhibit
+    the upgrade process otherwise the user will end up without a desktop.
```
bocekm:
```suggestion
    the upgrade process otherwise the user will end up without a desktop.
    Note: The Package Evolution Service data makes sure the KDE-related pkgs are removed in the dnf upgrade transaction.
```

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,66 @@
...
+
+    if kde_desktop_installed:
+        api.current_logger().info("KDE desktop is installed. Checking what we can do about it.")
+        if not gnome_desktop_installed:
+            api.current_logger().error("Cannot perform the upgrade because there is"
+                                       " no other desktop than KDE installed.")
+            # We cannot continue with the upgrade process
+            reporting.create_report([
+                reporting.Title("Cannot upgrade because there is no other desktop than KDE installed."),
+                reporting.Summary("With only KDE installed, there would be no other desktop env. after upgrade."),
```
bocekm:
```suggestion
                reporting.Summary("The KDE desktop environment is not available on RHEL 8. The KDE-related packages will be uninstalled during the upgrade and because the only currently installed desktop environment is KDE, there will be no other desktop environment after upgrade."),
```
Please apply the coding standards to the proposed change (max 120 chars one line).

repos/system_upgrade/el7toel8/actors/getinstalleddesktops/libraries/library.py
bocekm: I was trying two commands for KDE installation: `# yum groupinstall 'X Window System' 'KDE'` and `yum groupinstall kde-desktop`, the latter advised here https://access.redhat.com/solutions/371643. Both of them gave me the same result - installing the below set of package - **not including** the `plasma-workspace` one.
<details><summary>List of packages in the KDE pkg group:</summary>

```
==============================================================================================================================================================================================================================================
 Package                                                             Arch                                     Version                                                          Repository                                                Size
==============================================================================================================================================================================================================================================
Installing for group install "KDE":
 akonadi                                                             x86_64                                   1.9.2-4.el7                                                      rhel-7-server-eus-rpms                                   725 k
 akonadi-mysql                                                       x86_64                                   1.9.2-4.el7                                                      rhel-7-server-eus-rpms                                    16 k
 ark                                                                 x86_64                                   4.10.5-4.el7                                                     rhel-7-server-eus-rpms                                   277 k
 bluedevil                                                           x86_64                                   2.1-1.el7                                                        rhel-7-server-eus-rpms                                   378 k
 colord-kde                                                          x86_64                                   0.3.0-3.el7                                                      rhel-7-server-eus-rpms                                   131 k
 gwenview                                                            x86_64                                   4.10.5-5.el7                                                     rhel-7-server-eus-rpms                                   1.9 M
 kamera                                                              x86_64                                   4.10.5-5.el7                                                     rhel-7-server-eus-rpms                                    63 k
 kcalc                                                               x86_64                                   4.10.5-4.el7                                                     rhel-7-server-eus-rpms                                   175 k
 kcharselect                                                         x86_64                                   4.10.5-3.el7                                                     rhel-7-server-eus-rpms                                    93 k
 kcm-gtk                                                             x86_64                                   0.5.3-14.el7                                                     rhel-7-server-eus-rpms                                    54 k
 kcm_touchpad                                                        x86_64                                   0.3.1-11.el7                                                     rhel-7-server-eus-rpms                                    47 k
 kcolorchooser                                                       x86_64                                   4.10.5-3.el7                                                     rhel-7-server-eus-rpms                                    15 k
 kde-base-artwork                                                    noarch                                   4.10.5-2.el7                                                     rhel-7-server-eus-rpms                                   7.1 M
 kde-plasma-networkmanagement                                        x86_64                                   1:0.9.0.9-7.el7                                                  rhel-7-server-eus-rpms                                   538 k
 kde-print-manager                                                   x86_64                                   4.10.5-4.el7                                                     rhel-7-server-eus-rpms                                   272 k
 kde-settings-pulseaudio                                             noarch                                   19-23.8.el7                                                      rhel-7-server-eus-rpms                                    15 k
 kde-workspace                                                       x86_64                                   4.11.19-12.el7                                                   rhel-7-server-eus-rpms                                   9.6 M
 kdeaccessibility                                                    noarch                                   1:4.10.5-2.el7                                                   rhel-7-server-eus-rpms                                    14 k
 kdeadmin                                                            x86_64                                   7:4.10.5-4.el7                                                   rhel-7-server-eus-rpms                                   865 k
 kdegraphics-strigi-analyzer                                         x86_64                                   4.10.5-3.el7                                                     rhel-7-server-eus-rpms                                    28 k
 kdegraphics-thumbnailers                                            x86_64                                   4.10.5-3.el7                                                     rhel-7-server-eus-rpms                                    50 k
 kdenetwork-kdnssd                                                   x86_64                                   7:4.10.5-8.el7_0                                                 rhel-7-server-eus-rpms                                    39 k
 kdenetwork-kget                                                     x86_64                                   7:4.10.5-8.el7_0                                                 rhel-7-server-eus-rpms                                   1.3 M
 kdenetwork-krfb                                                     x86_64                                   7:4.10.5-8.el7_0                                                 rhel-7-server-eus-rpms                                   369 k
 kdepim                                                              x86_64                                   7:4.10.5-7.el7                                                   rhel-7-server-eus-rpms                                   5.6 M
 kdeplasma-addons                                                    x86_64                                   4.10.5-5.el7                                                     rhel-7-server-eus-rpms                                   2.4 M
 kdeutils-minimal                                                    noarch                                   6:4.10.5-3.el7                                                   rhel-7-server-eus-rpms                                    16 k
 kgpg                                                                x86_64                                   4.10.5-4.el7                                                     rhel-7-server-eus-rpms                                   962 k
 kmix                                                                x86_64                                   4.10.5-4.el7                                                     rhel-7-server-eus-rpms                                   574 k
 konsole                                                             x86_64                                   4.10.5-4.el7                                                     rhel-7-server-eus-rpms                                    78 k
 kruler                                                              x86_64                                   4.10.5-3.el7                                                     rhel-7-server-eus-rpms                                   134 k
 ksaneplugin                                                         x86_64                                   4.10.5-3.el7                                                     rhel-7-server-eus-rpms                                    26 k
 ksnapshot                                                           x86_64                                   4.10.5-3.el7                                                     rhel-7-server-eus-rpms                                   309 k
 ksshaskpass                                                         x86_64                                   0.5.3-7.el7                                                      rhel-7-server-eus-rpms                                    21 k
 ksysguard                                                           x86_64                                   4.11.19-12.el7                                                   rhel-7-server-eus-rpms                                   278 k
 kwallet                                                             x86_64                                   4.10.5-3.el7                                                     rhel-7-server-eus-rpms                                   351 k
 kwrite                                                              x86_64                                   4.10.5-6.el7                                                     rhel-7-server-eus-rpms                                    78 k
 okular                                                              x86_64                                   4.10.5-4.el7                                                     rhel-7-server-eus-rpms                                   413 k
 oxygen-gtk                                                          noarch                                   1.2.0-6.el7                                                      rhel-7-server-eus-rpms                                   5.0 k
 plasma-scriptengine-python                                          x86_64                                   4.11.19-12.el7                                                   rhel-7-server-eus-rpms                                    86 k
 sweeper                                                             x86_64                                   4.10.5-4.el7                                                     rhel-7-server-eus-rpms                                    97 k
 system-config-date                                                  noarch                                   1.10.6-3.el7                                                     rhel-7-server-eus-rpms                                   591 k
 xsettings-kde                                                       x86_64                                   0.12.3-7.el7                                                     rhel-7-server-eus-rpms                                    27 k
 xterm                                                               x86_64                                   295-3.el7                                                        rhel-7-server-eus-rpms                                   455 k
Installing:
 pinentry-qt                                                         x86_64                                   0.8.1-17.el7                                                     rhel-7-server-eus-rpms                                    66 k
Installing for dependencies:
 LibRaw                                                              x86_64                                   0.14.8-5.el7.20120830git98d925                                   rhel-7-server-eus-rpms                                   250 k
 PyQt4                                                               x86_64                                   4.10.1-13.el7                                                    rhel-7-server-eus-rpms                                   2.9 M
 ark-libs                                                            x86_64                                   4.10.5-4.el7                                                     rhel-7-server-eus-rpms                                   108 k
 boost-program-options                                               x86_64                                   1.53.0-27.el7                                                    rhel-7-server-eus-rpms                                   156 k
 compat-poppler022                                                   x86_64                                   0.22.5-4.el7                                                     rhel-7-server-eus-rpms                                   722 k
 compat-poppler022-qt                                                x86_64                                   0.22.5-4.el7                                                     rhel-7-server-eus-rpms                                   150 k
 cpp                                                                 x86_64                                   4.8.5-37.el7_6                                                   rhel-7-server-eus-rpms                                   6.0 M
 grantlee                                                            x86_64                                   0.3.0-4.el7                                                      rhel-7-server-eus-rpms                                   312 k
 gwenview-libs                                                       x86_64                                   4.10.5-5.el7                                                     rhel-7-server-eus-rpms                                   346 k
 kaccessible                                                         x86_64                                   4.10.5-3.el7                                                     rhel-7-server-eus-rpms                                    40 k
 kaccessible-libs                                                    x86_64                                   4.10.5-3.el7                                                     rhel-7-server-eus-rpms                                    20 k
 kcm_colors                                                          x86_64                                   4.11.19-12.el7                                                   rhel-7-server-eus-rpms                                   146 k
 kde-plasma-networkmanagement-libs                                   x86_64                                   1:0.9.0.9-7.el7                                                  rhel-7-server-eus-rpms                                   1.0 M
 kde-settings-ksplash                                                noarch                                   19-23.8.el7                                                      rhel-7-server-eus-rpms                                    16 k
 kde-settings-plasma                                                 noarch                                   19-23.8.el7                                                      rhel-7-server-eus-rpms                                    16 k
 kde-style-oxygen                                                    x86_64                                   4.11.19-12.el7                                                   rhel-7-server-eus-rpms                                   405 k
 kde-workspace-libs                                                  x86_64                                   4.11.19-12.el7                                                   rhel-7-server-eus-rpms                                   579 k
 kdenetwork-common                                                   noarch                                   7:4.10.5-8.el7_0                                                 rhel-7-server-eus-rpms                                    10 k
 kdenetwork-kget-libs                                                x86_64                                   7:4.10.5-8.el7_0                                                 rhel-7-server-eus-rpms                                   281 k
 kdenetwork-krfb-libs                                                x86_64                                   7:4.10.5-8.el7_0                                                 rhel-7-server-eus-rpms                                    16 k
 kdepim-libs                                                         x86_64                                   7:4.10.5-7.el7                                                   rhel-7-server-eus-rpms                                   3.2 M
 kdepim-runtime                                                      x86_64                                   1:4.10.5-3.el7                                                   rhel-7-server-eus-rpms                                   2.8 M
 kdepim-runtime-libs                                                 x86_64                                   1:4.10.5-3.el7                                                   rhel-7-server-eus-rpms                                   150 k
 kdepimlibs-akonadi                                                  x86_64                                   4.10.5-4.el7                                                     rhel-7-server-eus-rpms                                   1.6 M
 kdeplasma-addons-libs                                               x86_64                                   4.10.5-5.el7                                                     rhel-7-server-eus-rpms                                   369 k
 kdf                                                                 x86_64                                   4.10.5-3.el7                                                     rhel-7-server-eus-rpms                                   246 k
 kgreeter-plugins                                                    x86_64                                   4.11.19-12.el7                                                   rhel-7-server-eus-rpms                                    95 k
 khotkeys                                                            x86_64                                   4.11.19-12.el7                                                   rhel-7-server-eus-rpms                                   304 k
 khotkeys-libs                                                       x86_64                                   4.11.19-12.el7                                                   rhel-7-server-eus-rpms                                   150 k
 kinfocenter                                                         x86_64                                   4.11.19-12.el7                                                   rhel-7-server-eus-rpms                                   591 k
 kmag                                                                x86_64                                   4.10.5-3.el7                                                     rhel-7-server-eus-rpms                                   121 k
 kmenuedit                                                           x86_64                                   4.11.19-12.el7                                                   rhel-7-server-eus-rpms                                   361 k
 konsole-part                                                        x86_64                                   4.10.5-4.el7                                                     rhel-7-server-eus-rpms                                   393 k
 kscreen                                                             x86_64                                   1:1.0.1-9.el7                                                    rhel-7-server-eus-rpms                                   163 k
 ksysguard-libs                                                      x86_64                                   4.11.19-12.el7                                                   rhel-7-server-eus-rpms                                   240 k
 ksysguardd                                                          x86_64                                   4.11.19-12.el7                                                   rhel-7-server-eus-rpms                                   117 k
 ktimer                                                              x86_64                                   4.10.5-3.el7                                                     rhel-7-server-eus-rpms                                   162 k
 kwin                                                                x86_64                                   4.11.19-12.el7                                                   rhel-7-server-eus-rpms                                   2.3 M
 kwin-libs                                                           x86_64                                   4.11.19-12.el7                                                   rhel-7-server-eus-rpms                                   224 k
 libXaw                                                              x86_64                                   1.0.13-4.el7                                                     rhel-7-server-eus-rpms                                   192 k
 libbluedevil                                                        x86_64                                   2.1-1.el7                                                        rhel-7-server-eus-rpms                                    59 k
 libkdcraw                                                           x86_64                                   4.10.5-5.el7                                                     rhel-7-server-eus-rpms                                   120 k
 libkexiv2                                                           x86_64                                   4.10.5-4.el7                                                     rhel-7-server-eus-rpms                                   257 k
 libkipi                                                             x86_64                                   4.10.5-3.el7                                                     rhel-7-server-eus-rpms                                    99 k
 libksane                                                            x86_64                                   4.10.5-4.el7                                                     rhel-7-server-eus-rpms                                   104 k
 libkscreen                                                          x86_64                                   1:1.0.1-8.el7                                                    rhel-7-server-eus-rpms                                    75 k
 libkworkspace                                                       x86_64                                   4.11.19-12.el7                                                   rhel-7-server-eus-rpms                                   111 k
 libvncserver                                                        x86_64                                   0.9.9-13.el7_6                                                   rhel-7-server-eus-rpms                                   234 k
 lzo-minilzo                                                         x86_64                                   2.06-8.el7                                                       rhel-7-server-eus-rpms                                    15 k
 mariadb                                                             x86_64                                   1:5.5.60-1.el7_5                                                 rhel-7-server-eus-rpms                                   8.6 M
 mariadb-server                                                      x86_64                                   1:5.5.60-1.el7_5                                                 rhel-7-server-eus-rpms                                    11 M
 mesa-libGLU                                                         x86_64                                   9.0.0-4.el7                                                      rhel-7-server-eus-rpms                                   196 k
 okular-libs                                                         x86_64                                   4.10.5-4.el7                                                     rhel-7-server-eus-rpms                                   229 k
 okular-part                                                         x86_64                                   4.10.5-4.el7                                                     rhel-7-server-eus-rpms                                   839 k
 oxygen-gtk2                                                         x86_64                                   1.3.4-3.el7                                                      rhel-7-server-eus-rpms                                   431 k
 perl-DBD-MySQL                                                      x86_64                                   4.023-6.el7                                                      rhel-7-server-eus-rpms                                   140 k
 polkit-kde                                                          x86_64                                   0.99.1-4.20130311git.el7                                         rhel-7-server-eus-rpms                                    71 k
 pykde4                                                              x86_64                                   4.10.5-4.el7                                                     rhel-7-server-eus-rpms                                   4.2 M
 qimageblitz                                                         x86_64                                   0.0.6-7.el7                                                      rhel-7-server-eus-rpms                                    56 k
 qjson                                                               x86_64                                   0.8.1-4.el7                                                      rhel-7-server-eus-rpms                                    72 k
 qt-mysql                                                            x86_64                                   1:4.8.7-3.el7_6                                                  rhel-7-server-eus-rpms                                    32 k
 qt3                                                                 x86_64                                   3.3.8b-51.el7                                                    rhel-7-server-eus-rpms                                   3.5 M
 sip                                                                 x86_64                                   4.14.6-4.el7                                                     rhel-7-server-eus-rpms                                   122 k
 system-config-date-docs                                             noarch                                   1.0.11-4.el7                                                     rhel-7-server-eus-rpms                                   527 k
```

</details>

repos/system_upgrade/el7toel8/actors/checkkdeapps/libraries/library.py
Honza0297: There is a question what is the "base" application. The idea is to check only these apps, because they should be the "most used" ones.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
Honza0297: It is intended to behave like this. Because without GNOME installed, there is no need to check for apps because there is no desktop or there is only KDE, which means the upgrade cannot be performed. 

Or the goal is to reveal all potential errors at once?

repos/system_upgrade/el7toel8/actors/getinstalleddesktops/libraries/library.py
Honza0297: Oh, I see the problem. I checked the packages for Fedora, not for RHEL. In Fedora, there really is package named 
`plasma-workspace`, but in RHEL it should be `kde-workspace` I think. I will check the differences and use the right package for RHEL.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
bocekm: - GNOME and KDE installed: warn about the installed KDE apps
- GNOME and KDE not installed: do not warn the installed KDE apps
- GNOME installed but KDE not installed:  warn about the installed KDE apps

The contrast between the second and third bullet seemed to me a bit weird. But I guess that the reason for it is because you can run the KDE apps also in the GNOME desktop env, right?

Honza0297: Yes, exactly.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,73 @@
...
+
+    if kde_desktop_installed:
+        api.current_logger().info("KDE desktop is installed. Checking what we can do about it.")
+        if not gnome_desktop_installed:
+            api.current_logger().error("Cannot perform the upgrade because there is"
+                                       " no other desktop than KDE installed.")
+            # We cannot continue with the upgrade process
+            reporting.create_report([
+                reporting.Title("Cannot upgrade because there is no other desktop than KDE installed."),
+                reporting.Summary("The KDE desktop environment is not available on RHEL 8. "
```
bocekm:
```suggestion
                reporting.Summary("Because the KDE desktop environment is not available on RHEL 8, all the KDE-related packages would be removed during the upgrade. There would be no desktop environment installed after the upgrade."
```

bocekm:
Let's wait for the results of the QE test @kokesak has updated and run today.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
```diff
@@ -0,0 +1,73 @@
...
+        return
+
+    if kde_desktop_installed:
+        api.current_logger().info("KDE desktop is installed. Checking what we can do about it.")
+        if not gnome_desktop_installed:
+            api.current_logger().error("Cannot perform the upgrade because there is"
+                                       " no other desktop than KDE installed.")
+            # We cannot continue with the upgrade process
+            reporting.create_report([
+                reporting.Title("Cannot upgrade because there is no other desktop than KDE installed."),
```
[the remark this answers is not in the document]
lspackova: @bocekm  How about:
The installed KDE environment is unavailable on RHEL 8. Install GNOME to be able to upgrade the desktop environment.

repos/system_upgrade/el7toel8/actors/checkkdegnome/libraries/library.py
lspackova: @bocekm 
The KDE desktop environment is unvailable on RHEL 8, therefore all KDE-related packages need to be removed during the upgrade. As a result, no desktop environment would be installed after the upgrade.

I miss the GNOME part here. However, I don't have a good picture of when and among what other information the messages above are shown. So adjust as needed so that similar information is not duplicated in one place.

bocekm: Gnome is mentioned in the remediation a few lines below. The remediation message I suggest is: "Remove KDE (at least the `kde-workspace` package) or install the GNOME desktop environment to be able to upgrade."

lspackova: I've seen that and the text looks good to me. So for as long as the customer always gets the remediation hint, too, I'm fine with the wording.

bocekm:
All tests passed. I'll update the strings per the above comments in a separate PR.

repos/system_upgrade/el7toel8/actors/getinstalleddesktops/libraries/library.py
```diff
@@ -0,0 +1,17 @@
...
+from leapp.models import InstalledRPM
+
+
+def get_installed_desktops():
+    api.current_logger().info("  Detecting desktop environments  ")
+    api.current_logger().info("==================================")
+
+    # Detect installed desktops by one of the base rpm packages
+    kde_desktop_installed = has_package(InstalledRPM, "kde-workspace")
+    gnome_desktop_installed = has_package(InstalledRPM, "gnome-session")
```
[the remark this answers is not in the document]
Honza0297: This is not what it should do. Moreover, I cannot reproduce it. Could you provide additional info please (Specific RHEL version, installed packages etc.)?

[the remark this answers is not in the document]
bocekm: @Honza0297, it's because `gdm` (The GNOME Display Manager) is the **mandatory** package of the `kde-desktop` package group. 
For a proof, just look at the RHEL 7.x Server repo metadata, e.g.
```
<<URL to the RHEL 7.7 repo>>/Server/x86_64/os/repodata/6d132b085489dcfe4a0ee2b4c38d9578c38ab83cebe4df87e3a409b48e8fee57-comps-Server.x86_64.xml
```

PR #311, opened 2019-08-19
Make Report.to_dict a method, not a property

fernflower:
In spite of changes in leapp PR546 this will
make more sense.


shaded-enmity:
leapp-ci build

PR #318, opened 2019-08-22
Add RelatedResource fields to reports

shaded-enmity:
Report entries can have various `RelatedResources`, currently
we recognize the following 6 types of resources:
- file
- package
- directory
- repository
- kernel-driver
- pam

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
@@ -88,7 +88,8 @@ def get_events(pes_events_filepath):
             reporting.Summary(summary),
             reporting.Severity(reporting.Severity.HIGH),
             reporting.Tags([reporting.Tags.SANITY]),
-            reporting.Flags([reporting.Flags.INHIBITOR])
+            reporting.Flags([reporting.Flags.INHIBITOR]),
+            reporting.RelatedResource('file', pes_events_filepath)
         ])
         raise StopActorExecution()
 
@@ -294,7 +295,7 @@ def report_skipped_packages(message, packages):
         reporting.Summary(summary),
         reporting.Severity(reporting.Severity.HIGH),
         reporting.Tags([reporting.Tags.REPOSITORY]),
-    ])
+    ] + [reporting.RelatedResource('package', p) for p in packages])
     if is_verbose():
         api.show_message(summary)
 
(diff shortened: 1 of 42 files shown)


vojtechsokol:
leapp-ci build

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -294,7 +295,7 @@ def report_skipped_packages(message, packages):
         reporting.Summary(summary),
         reporting.Severity(reporting.Severity.HIGH),
         reporting.Tags([reporting.Tags.REPOSITORY]),
-    ])
+    ] + [reporting.RelatedResource('package', p) for p in packages])
```
jmikovic: TypeError: Values "scheme" and "identifier" of "RelatedResource" must be a string

shaded-enmity:
@jmikovic  Sorry, requires: https://github.com/oamg/leapp/pull/549

repos/system_upgrade/el7toel8/actors/vsftpdconfigcheck/libraries/library.py
```diff
@@ -26,5 +26,9 @@ def check_config_supported(tcpwrap_facts, vsftpd_facts):
...
                 title='Replacing TCP Wrappers in RHEL 8',
                 url='https://access.redhat.com/solutions/3906701'
-            )
-        ])
+            ),
+            reporting.RelatedResource('file', '/etc/hosts.allow'),
+            reporting.RelatedResource('file', '/etc/hosts.deny'),
+            reporting.RelatedResource('package', 'tcp_wrappers'),
+            reporting.RelatedResource('package', 'vsftpd'),
+        ] + [reporting.RelatedResource('file', bc) for bc in bad_configs])
```
fernflower: Are you sure we need to refer to bad_configs here?

fernflower:
leapp-ci build

repos/system_upgrade/el7toel8/actors/vsftpdconfigcheck/libraries/library.py
shaded-enmity: Hmm, it seems that `bad_configs` is basically a set of `/etc/hosts.{deny,allow}`, in that case I'd rather drop 30+31 and actually keep `bad_configs` so that we report only what is affected, what do you think?

fernflower: Yeah, probably that's better

PR #319, opened 2019-08-26
Fix traceback in dnfupgradetransaction if RHSM is skipped

vinzenz:
Previously some change introduced removing certificates as specified by
the SourceRHSMInfo message.
However this message might not be available if RHSM is skipped.
In that case a None object will be returned by the next() call.
While trying to access the existing_product_certificates member of the
message, which actually is a None object, an AttributeError exception
happens.

This patch fixes this issue.

Fixes Rabbit: OAMG-515


pirat89:
LGTM but do not have time for proper testing now. So keeping in that for Vojta.

AloisMahdal:
Veriffied with

leapp-repository-0.8.1-0.201908261348Z.4360417.rabbit_515_fix.PR319.el7_6.noarch

and there's no traceback; upgrade to 8.1 finished normally.

PR #320, opened 2019-08-26
Make some checks before run targetuserspacecreator dnf operations

### LINKED ISSUE #300: Incorrect error when missing repomap.csv or not being on 7.6
When `/etc/leapp/files/repomap.csv` is missing, the upgrade is not inhibited by the _repositoriesmapping_ actor with the message ["_Repositories map file not found_"](https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/actors/repositoriesmapping/libraries/library.py#L20).
Instead I'm getting:
```
2019-08-09 17:08:03.78  DEBUG    PID: 28670 leapp.workflow.FactsCollection.target_userspace_creator: Target RHSM Info: SKUs: [u'ES0113909'] Repositories: [u'rhel-8-for-x86_64-baseos-rpms', u'rhel-8-for-x86_64-appstream-rpms']
2019-08-09 17:08:03.430 DEBUG    PID: 28670 leapp.workflow.FactsCollection.target_userspace_creator: Gathered target repositories:
...
2019-08-09 17:08:27.251042 [ERROR] Actor: target_userspace_creator Message: There are no enabled target repositories for the upgrade process to proceed.
Detail: {u'hint': u'Ensure your system is correctly registered with the subscription manager and that your current subscription is entitled to install the requested target version 8.0'}
```

The _repositoriesmapping_ needs to run before the _targetuserspacecreator_. @vinzenz, don't you know of a better solution than sending a message between these just to create a dependency link?

asilveir:
- Fix bug if release not supported or repomap file not found.
- Fix side effect bug in dnftransactioncheck actor.
- This solution also prevents two lengthy DNF operations from being
  performed before displaying the expected error.

 Resolves #300

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkosrelease/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkosrelease/actor.py
@@ -1,5 +1,6 @@
 from leapp.actors import Actor
 from leapp.libraries.actor.library import check_os_version, skip_check
+from leapp.libraries.common.config import version
 from leapp.models import OSReleaseFacts
 from leapp.reporting import Report
 from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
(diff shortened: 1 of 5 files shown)


jmikovic:
leapp-ci build

PR #321, opened 2019-08-27
Add a comment to a shady Makefile line

bocekm:
This line seems to be useless - to move rpms from where they are to the same location - but it actually moves them from subfolders.


pirat89:
leapp-ci build

PR #322, opened 2019-08-28
initramdisk generator: respect sys architecture

pirat89:
Previously the x86_64 architecture was hardcoded. Now, it should work on other architectures as well (excluding problems with requirements of modules that does not exists - so actor still crashes
in such case in specific cases).


jmikovic:
leapp-ci build

pirat89:
leapp-ci build

PR #327, opened 2019-09-05
Makefile: use release 100 for master branch

pirat89:
We had some issues with rpms in downstream when people installed
mix of upstream and downstream rpms by default because of higher
release there. After some discussions we will rather use release
100 instead of 1 for upstream builds by default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,8 +33,8 @@ MASTER_BRANCH=master
 
 # In case the PR or MR is defined or in case build is not comming from the
 # MATER_BRANCH branch, N_REL=0; (so build is not update of the approved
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -34,7 +34,7 @@ MASTER_BRANCH=master
 # In case the PR or MR is defined or in case build is not comming from the
 # MATER_BRANCH branch, N_REL=0; (so build is not update of the approved
 # upstream solution). For upstream builds N_REL=1;
```
vinzenz:
```suggestion
# upstream solution). For upstream builds N_REL=100;
```

pirat89: Thanks for catch  :)

PR #331, opened 2019-09-09
xfs: Provide information per mountpoint if XFS is present or not

vinzenz:
Previously we did not report if XFS is present on a certain partition.
This was sufficient because did at most need only one overlayfs view.
However since we do want to be able to mount multiple partitions with
overlayfs and bindmount them together, we try to optimize the amount of
space needed for the scratch files, by not creating disk images for
partitions that do not require it.
This patch introduces the update for the information that is passed on.
A followup patch will take care of the actual mounting for the
additional partitions.

Addititonally this patch moves the actor from the checks phase to the
fact collection phase. It does not produce any checks or inhibitors. The
move is necessary as the information is already required in the facts
collection phase.

Fixes: #334


Rezney:
@vinzenz do you think it is suitable to take care of https://github.com/oamg/leapp-repository/issues/334 here?

vinzenz:
Yes can be taken care of here. Good catch

repos/system_upgrade/el7toel8/actors/xfsinfoscanner/actor.py
```diff
@@ -0,0 +1,23 @@
...
+from leapp.models import StorageInfo, XFSPresence
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+
+
+class XFSInfoScanner(Actor):
+    """
+    This actor scans all mounted partitions for XFS information
+
+    The actor will check each mounted partition reported in the StorageInfo message if it is a XFS partition and if
+    the partition is XFS, if it using ftype = 0. The actor will produce a message with the findings.
```
Rezney:
```suggestion
    the partition is XFS one using ftype = 0. The actor will produce a message with the findings.
```

repos/system_upgrade/el7toel8/models/xfspresence.py
```diff
@@ -3,7 +3,14 @@
 
 
 class XFSPresence(Model):
+    """ A generic message indicating if XFS was found and if any parition has ftype = 0 or not """
     topic = SystemInfoTopic
 
     present = fields.Boolean(default=False)
+    """ XFS is used on this partition """
```
Rezney: This model seems to be inconsistent to me.

> XFS is used on this partition

which partition?

> List of mountpoints that have ftype=0

Previously, it mentioned one partition but now we are talking about list of mount points

Maybe I am just misunderstanding it, could you please elaborate on this model?

Rezney:
@examon when can we expect CI to work?

PR #333, opened 2019-09-09
WIP: Enable upgrade on more architectures.

asilveir:
- The upgrade process will be not inhibit in 'x86_64', 'aarch64' and 'ppc64le'
 - Using logic from library.
 - Added unit tests.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checksystemarch/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/checksystemarch/libraries/library.py
@@ -1,0 +1,30 @@
+from leapp import reporting
+from leapp.libraries.common.config import architecture
+from leapp.libraries.stdlib import api
+
+
+def check_architecture():
+    """Check if given architecture is supported by upgrade process"""
+    if not architecture.matches_architecture(*architecture.ARCH_ACCEPTED):
+        inhibit_upgrade()
+
+
+def inhibit_upgrade():
+    """Generate an upgrade inhibitor"""
+    reporting.create_report(
+        [
+            reporting.Title(
+                'Unsupported architecture [{}]'.format(
+                    api.current_actor().configuration.architecture
+                )
+            ),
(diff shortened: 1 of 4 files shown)


repos/system_upgrade/el7toel8/libraries/config/architecture.py
```diff
@@ -6,8 +6,8 @@
 ARCH_X86_64 = 'x86_64'
 ARCH_ARM64 = 'aarch64'
 ARCH_PPC64LE = 'ppc64le'
-ARCH_S390X = 's390x'
-ARCH_ACCEPTED = (ARCH_X86_64, ARCH_ARM64, ARCH_PPC64LE, ARCH_S390X)
+# ARCH_S390X = 's390x'
```
vinzenz: What's the reason for removing the constant here now? and also from the accepted arches. Somehow I don't understand why this change was applied now. I thought we are working on making it work. Some reasoning in the PR and commit message for why to disable something that has been enabled before would be great.

pirat89: `ARCH_ACCEPTED` should be kept unchanged. It doesn't have anything common with the set of supported arches. Instead, should be provided rather new constant: `ARCH_SUPPORTED`.

repos/system_upgrade/el7toel8/actors/checksystemarch/libraries/library.py
```diff
@@ -0,0 +1,30 @@
...
+    if not architecture.matches_architecture(*architecture.ARCH_ACCEPTED):
+        inhibit_upgrade()
+
+
+def inhibit_upgrade():
+    """Generate an upgrade inhibitor"""
+    reporting.create_report(
+        [
+            reporting.Title(
+                'Unsupported architecture [{}]'.format(
```
pirat89: Title should be static. Not dynamic.

PR #336, opened 2019-09-11
 CheckKernelDrivers: List of blacklisted drivers mirrors the official RH docs

examon:
I've updated the list of blacklisted drivers so it mirros the official RH docs [0].
Currently, we have a lot of drivers in the blacklist that are not present in the official list of removed drivers [0] (for example: at24). Our current list was compiled some time ago by Pavel. I've asked bunch of people to comment on this issue and they told me that the official list should be correct. The reason why our current list has more drivers is complex (some drivers were baked into the kernel, some we renamed/deleted/etc.).

From now on, I would stick only to the official list and not our "custom" list.

[0] https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/considerations_in_adopting_rhel_8/index#removed-device-drivers_hardware-enablement


repos/system_upgrade/el7toel8/actors/kernel/checkkerneldrivers/files/removed_drivers.txt
```diff
@@ -116,88 +78,22 @@ pata_serverworks
...
 pata_via
-pcc-cpufreq
-pch_dma
-pch_gbe
-pch_phub
-pci-hyperv
-pcnet32
 pdc_adma
-pm80xx
+pm80xx(pm8001)
```
pirat89: @examon  This seems incorrect.
```suggestion
pm80xx
```

With previous string, you would not be able to find pm80xx.

PR #338, opened 2019-09-12
Read repositories mapping from repomap file

asilveir:
- Removed hardcode repositories mapping.
- Updated unit tests.

**This PR requires updated repomap file**


asilveir:
leapp-ci build

pirat89:
@asilveir please rebase it with the today's masters.

PR #340, opened 2019-09-15
mock_configs: add disclaimer about testing purposes

pirat89:
Not so required to have. But I think it's more safe to add there something like that. Whad do you think?


bocekm:
leapp-ci build

PR #341, opened 2019-09-15
packaging: drop dnf from deps on RHEL-7 system

pirat89:
Since we are using DNF from RHEL-8 userspace, we haven't need to
depend on DNF anymore on RHEL-7 system. Keeping the dependency on
RHEL-8 even when I am not sure whether we need it.


vinzenz:
It is still required for the initial bootstrap of the el8 dnf

pirat89:
Aah. Thanks for remind! I completely missed that. So closing the PR.

PR #342, opened 2019-09-16
Fix a broken assertion in a test

olysonek:
While researching how to unit test that remediation is correctly set, I noticed that
an assertion in a test was incorrectly rewritten in commit 548a3b83909c4. The
assertion is not looking for the expected data where it should. To
test the actual data, we need to reach far deeper into the object hierarchy.

As it is now, the assertion effectively checks that the list [False]
is True (which it is, because a non-zero length list is True). So the
assertion passes even though we don't want it to.


pirat89:
leapp-ci build

bocekm:
leapp-ci build

bocekm:
Thanks. I'll probably remove the remediation soon™ under https://github.com/oamg/leapp-repository/pull/168/files though.

PR #343, opened 2019-09-17
commonleappdracutmodules: require the biosdevname for intel only

pirat89:
The biosdevname rpm is built only for X86_64 architecture. Require
this one to be installed for preparation of upgrade initrd only when
the system is intel arch.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/libraries/modscan.py
+++ b/repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/libraries/modscan.py
@@ -1,11 +1,11 @@
 import os
 import re
 
+from leapp.libraries.common.config import architecture
 from leapp.libraries.stdlib import api
 from leapp.models import RequiredUpgradeInitramPackages, UpgradeDracutModule
 
 _REQUIRED_PACKAGES = [
-    'biosdevname',
     'binutils',
     'cifs-utils',
     'device-mapper-multipath',
@@ -46,7 +46,10 @@ def _create_dracut_modules():
 
 
 def _create_initram_packages():
-    return RequiredUpgradeInitramPackages(packages=_REQUIRED_PACKAGES)
+    required_pkgs = _REQUIRED_PACKAGES
+    if architecture.matches_architecture(architecture.ARCH_X86_64):
+        required_pkgs.append('biosdevname')
+    return RequiredUpgradeInitramPackages(packages=required_pkgs)
 
 
 def process():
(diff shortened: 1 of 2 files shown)


pirat89:
(Untested. I am keeping the solution check just on unit tests and reviewers)

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/tests/test_modscan.py
```diff
@@ -41,8 +43,17 @@ def test_created_modules(monkeypatch):
         assert not required_modules
 
 
-def test_required_packages():
-    assert set(modscan._REQUIRED_PACKAGES) == set(modscan._create_initram_packages().packages)
+def test_required_packages(monkeypatch):
+    class CurrentActorMocked(object):
```
bocekm: The test fails on the other test - `test_required_packages`.
Fixed file: https://hastebin.com/gifijanimu.py
Also there's no need for using the monkeypatch context. It's necessary only in [certain cases](https://docs.pytest.org/en/latest/reference.html#_pytest.monkeypatch.MonkeyPatch.context) which is I think not this case.

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/libraries/modscan.py
```diff
@@ -46,7 +47,10 @@ def _create_dracut_modules():
 
 
 def _create_initram_packages():
-    return RequiredUpgradeInitramPackages(packages=_REQUIRED_PACKAGES)
+    required_pkgs = _REQUIRED_PACKAGES
+    if architecture.matches_architecture(architecture.ARCH_X86_64):
+        required_pkgs.append('biosdevname')
```
bocekm: Didn't you want to remove the biosdevname from the _REQUIRED_PACKAGES?

pirat89:
@bocekm Oh man... Yes! I will update tests first to catch that and then I will fix it.

pirat89:
@bocekm Fixed. See the commits. Please, when you approve it, squash them into the one.

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/libraries/modscan.py
pirat89: @bocekm Oh man... Yes! I will update tests first to catch that and then I will fix it.

pirat89:
Mering. In case of error, blame me :-)

PR #345, opened 2019-09-18
Do not remove what shall be installed

bocekm:
Do not remove what is supposed to be kept or installed.

PES events may give us conflicting requests - to both install/keep and remove a pkg.
Example:

    PES event 1: sip-devel SPLIT INTO python3-sip-devel, sip
    PES event 2: sip SPLIT INTO python3-pyqt5-sip, python3-sip

Without this change, sip would reside in both 'to_keep' and 'to_remove', causing a dnf conflict

Fixes OAMG-513.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
@@ -234,6 +234,7 @@ def process_events(events, installed_pkgs):
     map_repositories(tasks['to_install'])
     map_repositories(tasks['to_keep'])
     filter_out_pkgs_in_blacklisted_repos(tasks['to_install'])
+    resolve_conflicting_requests(tasks)
 
     return tasks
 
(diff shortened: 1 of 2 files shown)


pirat89:
@bocekm this looks to me more like issue in data itself. Are those entries really correct? Anyway, you are right we need to handle it, doesn't matter how much this is valid.

bocekm:
@pirat89, the PES data are valid. See the description of the events:
> Renamed subpackage to reflect used python version. The 'sip' binary has been moved from the devel package to the sip package, as it is not specific to the Python version.

> Python 2->3 rebuild and rename. The RHEL 8 CRB 'sip' package now contains the sip binary used by python*-sip-devel.

pirat89:
@bocekm I see now. It's little confusing but you are right. The code and unit test seems good. I will try to test it.

AloisMahdal:
Can you please rebase to include #364 (sblim-wbemcli `to_remove` merged today)?  (I need both applied to verify either.)

bocekm:
Rebased.

PR #347, opened 2019-09-19
Switch target release from 8.0 to 8.1

bocekm:
Make sure the dnf calls respect this target version.

Since CDN does not contain 8.1 content yet, after this PR is merged, we'll need to test with custom repos only, using the LEAPP_DEVEL_SKIP_RHSM=1.


asilveir:
In my test upgrade works but: 
```console
[root@leapp-20190925152242 ~]# cat /etc/redhat-release 
Red Hat Enterprise Linux release 8.0 (Ootpa)
```

pirat89:
@asilveir but that's expected when you use devel way of upgrades :) and still you cannot use the std way for 8.1. There are currently no checks of the content as in std way the problem is resolved by design.

PR #348, opened 2019-09-19
Fix bug when adding remote rpms

Rezney:
Currently, when adding remote rpms, we added them only to the
sack as rpm objects, however never to the transaction itself.

It was working for us only because the rpms we added to the
sack was brought to the transaction by another package.


AloisMahdal:
Verified with full downstream test suite; everything passed:

    leapp-0.8.1-100.201909191013Z.85c9791.master.el7_6
    leapp-repository-0.8.1-0.201909191510Z.b53ec99.fix_add_remote_rpms.PR348.el7_6

jmikovic:
leapp-ci build

PR #350, opened 2019-09-23
Check s390x

pirat89:
Add two new actors to get information about CPU (in this case, we are interested just about the machine field for now) and check whether the machine's CPU is supported on RHEL 8. Currently we need this now just for the s390x architecture.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/scancpu/libraries/cpu.py
+++ b/repos/system_upgrade/el7toel8/actors/scancpu/libraries/cpu.py
@@ -1,0 +1,24 @@
+import re
+
+from leapp.libraries.stdlib import api
+from leapp.models import CPUInfo
+
+RE_MACHINE_TYPE = re.compile(r'^processor.*\smachine\s*=\s*([0-9]+)')
+
+
+def _get_cpuinfo():
+    """Return lines from /proc/cpuinfo."""
+    # Expecting the file exists on earch system, skipping any check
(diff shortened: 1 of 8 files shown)


pirat89:
leapp-ci build

pirat89:
leapp-ci build

repos/system_upgrade/el7toel8/actors/scancpu/libraries/cpu.py
```diff
@@ -0,0 +1,24 @@
...
+
+def process():
+    cpuinfo = CPUInfo()
+
+    machine_types = [RE_MACHINE_TYPE.findall(line) for line in _get_cpuinfo()]
+    machine_types = [i[0] for i in machine_types if i]
+    if machine_types:
+        # machine type should be same for all found cpus
+        cpuinfo.machine_type = int(machine_types[0])
+    api.produce(cpuinfo)
```
asilveir:
```suggestion
        api.produce(cpuinfo)
```
I think maybe it is better to produce cpuinfo only when you have the _machine_type_

pirat89: Nope. It shoulw actualy produce the message every time. Later it should be extended when needed, but the purpose is to provide info about CPU. Just now it is limited to provide only the machine info. In your case, it could happen that another actor will raise exception when you find s390x machine that will not provide the machine type info. Not sure whether it is possible, but assume it is.

PR #352, opened 2019-09-23
Enable/Fix s390x

pirat89:
Enable upgrade on s390x machines + skip scanning of GRUB2 config on the s390x arch (there is not GRUB2, but ZIPL).

@vinzenz: Currently we do not handle ZIPL, so I am ok to remove the commit modifyinf `ARCH_SUPPORTED` tuple in `architecture`. More important is here the skip of GRUB scanner.


pirat89:
leapp-ci build

pirat89:
leapp-ci build

pirat89:
@asilveir  This should affect only s390x arch.

vinzenz:
Tested as base of my commits and works as expected.

PR #349, opened 2019-09-23
Fix not printing check transaction stderr

Rezney:

PR #353, opened 2019-09-24
Remove transaction check actor

asilveir:
- This actor just gives almost the same error than the upgrade transaction actor.
- Removed to save some upgrade execution time.


vinzenz:
The point of this actor is actually to include those problems in the report eventually. There is no point of removing the actor.

Rezney:
I could imagine that we would skip running the check with `leapp upgrade` but it is essential for `leapp-preupgrade`.

We should definitely not remove it...

Rezney:
We discussed this with @pirat89 and we will need this also in `leapp upgrade` cause we do not differentiate `preupgrade` and `upgrade` in terms of reporting.

bocekm:
Improved solution: https://github.com/oamg/leapp-repository/pull/349

PR #355, opened 2019-09-25
Add unit test to actor SCTP checks.

asilveir:
- just a quick unit test to check out the two possible scenarios.


repos/system_upgrade/el7toel8/actors/sctpchecks/tests/component_test.py
```diff
@@ -0,0 +1,15 @@
+from leapp.actors import Actor
+from leapp.models import SCTPConfig, RpmTransactionTasks
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+def test_scpt_wanted(current_actor_context):
+    current_actor_context.feed(SCTPConfig(wanted=True))
+    current_actor_context.run()
+    assert current_actor_context.consume(RpmTransactionTasks)
```
pirat89: Please, check even whether the produced msg contains expected data and add one more check for the case there is not input msg

pirat89:
leapp-ci build

PR #356, opened 2019-09-26
Change the title of the boot entry

pirat89:
The previous title contained white spaces:
 - `RHEL Upgrade Initramfs`

Whitespaces are not valid characters in title in case of the ZIPL
bootloader. To be consistent everywhere, replaces spaces by dashes:
- `RHEL-Upgrade-Initramfs`


pirat89:
leapp-ci build

pirat89:
I am merging it. The jenkins CI is completely down and we do not have time to check what's wrong with that every day.

PR #358, opened 2019-09-27
[WIP]: s390x: Fixes around initramdisk for s390x

vinzenz:
WIP

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/initramdiskgenerator/libraries/initramgen.py
+++ b/repos/system_upgrade/el7toel8/actors/initramdiskgenerator/libraries/initramgen.py
@@ -68,9 +68,11 @@ def generate_initram_disk(context):
     copy_dracut_modules(context, modules)
     context.copy_to(generator_script, os.path.join('/', INITRAM_GEN_SCRIPT_NAME))
     install_initram_deps(context)
+    if os.path.exists('/etc/dasd.conf'):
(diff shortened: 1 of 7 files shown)


repos/system_upgrade/el7toel8/actors/initramdiskgenerator/libraries/initramgen.py
```diff
@@ -68,9 +68,11 @@ def generate_initram_disk(context):
     copy_dracut_modules(context, modules)
     context.copy_to(generator_script, os.path.join('/', INITRAM_GEN_SCRIPT_NAME))
     install_initram_deps(context)
+    if os.path.exists('/etc/dasd.conf'):
```
pirat89: Isn't that required on every s390x? In such case, non-existing file could be an error on that arch.
```suggestion
    if architecture.matches_architecture(architecture.ARCH_S390X):
```

+ import of `architecture`. But this is ok as well.

vinzenz: I just didnt add checks just yet

pirat89: I know. I just wanted to point that in case it could help later.

pirat89:
It will be resolved in PR #373

PR #363, opened 2019-10-01
Workflow change

pirat89:
We realized we cannot properly gather all information and check it just with one scan and check phase when we need to get and process data from target system (e.g. create target userspace). In some cases we cannot even get those information because current HW is not supported on target system, so all such actions fails and we cannot prevent such actors effectively to no fail.
    
To resolve that, add new 'Target*' phases for actors working with with target userspace or data about calculated RPM transaction and move related actors to thodr phases.

Additionally, in case of inhibition from the original checks phase, skip these two phases and continue with reporting phase (requires leapp PR: https://github.com/oamg/leapp/pull/566). This is done by special extra actor in the checks phase.

Note: currently just two
Note2: I am opened to discussion whether to add just one new phase instead of two, but keep in mind we want to do more stuff in future around that. (check transaction output, ...)

Note3: **Failed tests are expected until the merge of the leapp PR.**

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkskipphase/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkskipphase/actor.py
@@ -1,0 +1,35 @@
+from leapp.actors import Actor
+from leapp.messaging.commands import SkipPhasesUntilCommand
+from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckSkipPhase(Actor):
+    """
+    Skip all the subsequent phases until the report phase.
+
+    The phases that follow after the Checks phase work with the target (RHEL 8)
+    user space - stuff around preparing and checking the rpm transaction.
+    We do not want to process those phases in case of inhibition - e.g. for
+    a specific HW unsupported by the target system we cannot do anything - we
+    can just see some unclear errors in such case. So we want to instead skip
+    to the Reports phase to provide clear report to user without confusing
+    errors.
+
+    The actor is processed after all actors in the phase (that provides Report
+    messages) are processed.
+    """
+
+    name = 'check_skip_phase'
+    consumes = (Report,)
+    produces = ()
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        results = list(self.consume(Report))
+        inhibitors = [msg for msg in results if 'inhibitor' in msg.report.get('flags', [])]
+        if inhibitors:
+            self.log.info("An upgrade inhibitor detected. Skipping to the Report phase.")
+            # until_phase='targettransactioncheck'  === phase after this phase will be processed
+            # === the Reports phase
+            self._messaging.command(SkipPhasesUntilCommand(until_phase='targettransactioncheck'))
--- a/repos/system_upgrade/el7toel8/tags/facts.py
+++ b/repos/system_upgrade/el7toel8/tags/facts.py
@@ -3,3 +3,7 @@ from leapp.tags import Tag
 
 class FactsPhaseTag(Tag):
     name = 'facts'
+
+
+class TargetTransactionFactsPhaseTag(Tag):
+    name = 'target_transaction_facts'
--- a/repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
+++ b/repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
@@ -1,51 +1,85 @@
-from leapp.workflows import Workflow
-from leapp.workflows.phases import Phase
-from leapp.workflows.flags import Flags
-from leapp.workflows.tagfilters import TagFilter
-from leapp.workflows.policies import Policies
+from leapp import tags
 from leapp.models import IPUConfig
-from leapp.tags import IPUWorkflowTag, FactsPhaseTag, ChecksPhaseTag, \
-    DownloadPhaseTag, InterimPreparationPhaseTag, InitRamStartPhaseTag, \
-    LateTestsPhaseTag, PreparationPhaseTag, RPMUpgradePhaseTag, ApplicationsPhaseTag, ThirdPartyApplicationsPhaseTag, \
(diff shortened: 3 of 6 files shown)


repos/system_upgrade/el7toel8/tags/facts.py
```diff
@@ -3,3 +3,6 @@
 
 class FactsPhaseTag(Tag):
     name = 'facts'
+
+class TargetChecksPhaseTag(tag):
```
vinzenz:
```suggestion
class TargetFactsPhaseTag(tag):
```

repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
```diff
@@ -1,51 +1,85 @@
...
+        flags = Flags()
+
+    class TargetCheckPhase(Phase):
+        """
+        Checks upgradability regarding the information gathered about the target system.
+
+        Check whether expected repositories and rpms are available, what rpms are planned
+        to install, remove, ...
+
+        IOW, checks related to rpm transaction mainly.
```
shaded-enmity: This is kinda confusing. Can we just call it `TargetTransactionCheckPhase` or `TargetRPMCheckPhase` so that we're a little bit more explicit about what functionality we expect to be implemented there?

pirat89: I am thinking which of those two could be better. But both are better than the current one.

shaded-enmity: My favorite is `TargetTransactionCheckPhase`, it seems most explicit about what is happening in that phase.

repos/system_upgrade/el7toel8/actors/checkskipphase/actor.py
```diff
@@ -0,0 +1,31 @@
...
+    Skip following phases untill the report phase.
+
+    Following phases works with target user space - stuff around preparing and
+    checking the rpm transaction. We do not want to process those phases in case
+    of inhibition - e.g. for aspecific HW unsupported by the target system we
+    cannot do anything - we can just see some unclear errors in such case. So
+    we want to instead skip to the report phase to provide clear report to user
+    withou confusing errors.
+
+    The actor is processed as the last actor in the checks phase and it is the
```
shaded-enmity: How does this work? What decides that this actor will really be executed last?

pirat89: It's based on leapp framework create execution order based on consumed messages. As the actor consumes the `Report` messages (and do not produce any), it will be resolved as the last one (otherwise I'd call it bug of leapp)

repos/system_upgrade/el7toel8/workflows/inplace_upgrade.py
pirat89: So be it.

repos/system_upgrade/el7toel8/actors/checkskipphase/actor.py
```diff
@@ -0,0 +1,31 @@
+from leapp.actors import Actor
+from leapp.messaging.commands import SkipPhasesUntilCommand
+from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+class CheckSkipPhase(Actor):
+    """
+    Skip following phases untill the report phase.
```
bocekm:
```suggestion
    Skip all the subsequent phases until the report phase.
```

repos/system_upgrade/el7toel8/actors/checkskipphase/actor.py
```diff
@@ -0,0 +1,31 @@
...
+
+    Following phases works with target user space - stuff around preparing and
```
bocekm:
```suggestion
    The phases that follow after the Checks phase work with the target (RHEL8) user space - stuff around preparing and
```

repos/system_upgrade/el7toel8/actors/checkskipphase/actor.py
```diff
@@ -0,0 +1,31 @@
...
+
+    name = 'check_skip_phase'
+    consumes = (Report,)
+    produces = ()
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        results = list(self.consume(Report))
+        for error in [msg for msg in results if 'inhibitor' in msg.report.get('flags', [])]:
+            self.log.info("An upgrade inhibitor detected. Skipping to the report phase.")
```
bocekm:
```suggestion
            self.log.info("An upgrade inhibitor detected. Skipping to the Reports phase.")
```

repos/system_upgrade/el7toel8/actors/checkskipphase/actor.py
```diff
@@ -0,0 +1,31 @@
...
+
+class CheckSkipPhase(Actor):
+    """
+    Skip following phases untill the report phase.
+
+    Following phases works with target user space - stuff around preparing and
+    checking the rpm transaction. We do not want to process those phases in case
+    of inhibition - e.g. for aspecific HW unsupported by the target system we
+    cannot do anything - we can just see some unclear errors in such case. So
+    we want to instead skip to the report phase to provide clear report to user
```
bocekm:
```suggestion
    we want to instead skip to the Reports phase to provide clear report to user
```

repos/system_upgrade/el7toel8/actors/checkskipphase/actor.py
pirat89: fixed

repos/system_upgrade/el7toel8/actors/checkskipphase/actor.py
pirat89: fixed

Rezney:
leapp-ci build

pirat89:
leapp-ci build

pirat89:
@vinzenz @bocekm: should be fixed.

Rezney:
Tested. Works fine.

Issue #365, opened 2019-10-04
repository messes with import hooks


pemensik:
**Actual behavior**
Repository messes badly with import paths. It forces scripts inside git repository to use import leapp. It it is not fully configured (because Fedora 29 is not fully supported), even scripts that do not import any part of leapp modules fail.

Stop that. It complicates debugging of library parts on host machine. Moving the same file outside repository seems lame.

**To Reproduce**
Steps to reproduce the behavior

1. git clone https://github.com/oamg/leapp-repository.git
2. cd leapp-repository
3. (touch empty.py && pytest empty.py; rm empty.py)
```
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/usr/lib/python3.7/site-packages/leapp/models/__init__.py", line 177, in resolve
INTERNALERROR>     cls._resolved = globals()[cls._referenced]
INTERNALERROR> KeyError: 'ReportReference'
INTERNALERROR> 
INTERNALERROR> During handling of the above exception, another exception occurred:
INTERNALERROR> 
INTERNALERROR> Traceback (most recent call last):
INTERNALERROR>   File "/usr/lib/python3.7/site-packages/_pytest/main.py", line 176, in wrap_session
INTERNALERROR>     config.hook.pytest_sessionstart(session=session)
INTERNALERROR>   File "/usr/lib/python3.7/site-packages/pluggy/__init__.py", line 617, in __call__
INTERNALERROR>     return self._hookexec(self, self._nonwrappers + self._wrappers, kwargs)
INTERNALERROR>   File "/usr/lib/python3.7/site-packages/pluggy/__init__.py", line 222, in _hookexec
INTERNALERROR>     return self._inner_hookexec(hook, methods, kwargs)
INTERNALERROR>   File "/usr/lib/python3.7/site-packages/pluggy/__init__.py", line 216, in <lambda>
INTERNALERROR>     firstresult=hook.spec_opts.get('firstresult'),
INTERNALERROR>   File "/usr/lib/python3.7/site-packages/pluggy/callers.py", line 201, in _multicall
INTERNALERROR>     return outcome.get_result()
INTERNALERROR>   File "/usr/lib/python3.7/site-packages/pluggy/callers.py", line 76, in get_result
INTERNALERROR>     raise ex[1].with_traceback(ex[2])
INTERNALERROR>   File "/usr/lib/python3.7/site-packages/pluggy/callers.py", line 180, in _multicall
[...]

PR #367, opened 2019-10-08
Import leapp only when environment is set

pemensik:
Allows using pytest even on not configured system for small parts not
referencing leapp libraries. Would work the same way when environment is
set, but omits fatal failures of pytest in case leapp cannot be imported.

Related to issue #365. Would help local testing of leapp unrelated parts. If environment in not set, scan is not imported.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,16 +1,27 @@
 import os
 
-from leapp.utils.repository import find_repository_basedir
-from leapp.repository.scan import find_and_scan_repositories
+def load_repo(path):
+    """
+    Load repository on demand.
 
+    Do not require paths initialized if no environment is set.
+    Allows some parts to be tested without working leapp installation.
+    """
+    from leapp.utils.repository import find_repository_basedir
+    from leapp.repository.scan import find_and_scan_repositories
+
+    repo = find_and_scan_repositories(find_repository_basedir(path), include_locals=True)
+    repo.load()
+    return repo
+        
 
 def pytest_sessionstart(session):
+
     actor_path = os.environ.get('LEAPP_TESTED_ACTOR', None)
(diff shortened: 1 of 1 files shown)


pirat89:
I would rather do quite opposite thing rather. Do not import leapp when envar is set. Because usually almost everything need leapp.

pemensik:
Usually it does. But it should have explicitly specified at least one dependency. It should depend on relation stated explicitly. Not on feature hidden somewhere in tests configuration. Current code does not call load() unless one of variables is set. Is any magic done inside imports done on first two lines? It there is nothing really special, that import is not needed for anything. If it does, I think it should be moved somewhere expected (and documented).

I think environment variables are not best way to configure py.test. I think `py.test target/file.py` should just work. On both leapp or any random test file. It would be more convenient to run py.test just from library directory or as `py.test argument`, instead of some environment variable. Where it has to be, ok. I think it does not have to be there. Therefore it should not be imported. Instead, special initialization should be done on import of leapp.libraries or leapp.actor, if required.

pirat89:
I see. Ok. It makes sense. For me approved.

conftest.py
```diff
@@ -1,16 +1,26 @@
 import os
 
-from leapp.utils.repository import find_repository_basedir
-from leapp.repository.scan import find_and_scan_repositories
+def load_repo(path):
+    """ Load repository on demand.
 
+        Do not require paths initialized if no environment is set.
+        Allows some parts to be tested without working leapp installation.
+    """
```
pirat89: Just use this  format pls:
```
    """
    Load repository on demand.

    Do not require paths initialized if no environment is set.
    Allows some parts to be tested without working leapp installation.
    """
```

PR #359, opened 2019-09-30
Blacklisting enabling crb

fellipeh:
The  purpose of these PR is to change hardcoded blacklisting of optional repos to dynamic one, using RepositoriesMap data and information about available RH repositories, so the CRB repositories are blacklisted dynamically based on gathered information and data

PR #368, opened 2019-10-09
Let the repomap variable names explain themselves

bocekm:
It was not clear what `to_id` or `to_name` means and what's the difference - both are in fact repo names - one is the CDN repoid and the other is repo name as used in the PES.


pirat89:
I see the point. But I don't like this naming either. Let's discuss it later whether we will come up with something.

repos/system_upgrade/el7toel8/models/repositoriesmap.py
```diff
@@ -12,11 +12,11 @@ class RepositoryMap(Model):
 
     topic = TransactionTopic
 
-    from_id = fields.String()
-    to_id = fields.String()
-    to_name = fields.String()
-    from_minor_version = fields.String()
-    to_minor_version = fields.String()
+    from_repo_cdn = fields.String()  # RHEL 7 repoid as present in the Red Hat CDN
```
pirat89:
```suggestion
    from_repo_cdn = fields.String()
    """RHEL 7 repoid as present in the Red Hat CDN"""
    
```

pirat89:
The https://github.com/oamg/leapp-repository/pull/359 PR has been merged. You need to apply changes in `repositoriesblacklisted` as well.

repos/system_upgrade/el7toel8/models/repositoriesmap.py
bocekm: Done.

bocekm:
`repositoriesblacklisted` actor changed too.
Changed:
`from_id` -> `from_repoid`
`to_id` -> `to_repoid`
`to_name` -> `to_pes_repo`

PR #354, opened 2019-09-25
Revert "Clean the code of setuptargetrepos (#301)"

pirat89:
This reverts commit 47188e3b3475597f5fcea5fdecaf31dd36ab2060.

The commit brings several issues and I do not have power to fix it
now correctly. I am reverting so it can be fixed properly in future
without blocking the others now.

PR #371, opened 2019-10-10
[WIP] Clean the code of setuptargetrepos

### LINKED ISSUE #233: setuptargetrepos: copy paste error?

bocekm:
Resolves #233.

A new PR after the original one https://github.com/oamg/leapp-repository/pull/301 was merged and then reverted https://github.com/oamg/leapp-repository/pull/354 later as it caused that using custom repos stopped working.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
@@ -6,8 +6,9 @@ from leapp.exceptions import StopActorExecution, StopActorExecutionError
 from leapp import reporting
 from leapp.libraries.stdlib import api
 from leapp.libraries.stdlib.config import is_verbose
-from leapp.models import (InstalledRedHatSignedRPM, PESRpmTransactionTasks,
-                          RepositoriesSetupTasks, RpmTransactionTasks, RepositoriesBlacklisted)
+from leapp.models import (CustomTargetRepositories, CustomTargetRepository,
+                          InstalledRedHatSignedRPM, PESRpmTransactionTasks,
+                          RepositoriesBlacklisted, RpmTransactionTasks)
 
 # FIXME: this mapping is not complete and will need to be manually updated frequently
 REPOSITORIES_MAPPING = {
@@ -406,4 +407,4 @@ def produce_messages(tasks):
                                            to_remove=to_remove_pkgs))
 
     if to_enable_repos:
-        api.produce(RepositoriesSetupTasks(to_enable=to_enable_repos))
+        api.produce(CustomTargetRepositories(repos=[CustomTargetRepository(repoid=repo) for repo in to_enable_repos]))
--- a/repos/system_upgrade/el7toel8/actors/setuptargetrepos/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/setuptargetrepos/actor.py
@@ -1,24 +1,23 @@
 import platform
 
 from leapp.actors import Actor
-from leapp.models import (CustomTargetRepository, RepositoriesBlacklisted, RepositoriesFacts, RepositoriesMap,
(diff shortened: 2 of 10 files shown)


repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -406,4 +407,4 @@ def produce_messages(tasks):
                                            to_remove=to_remove_pkgs))
 
     if to_enable_repos:
-        api.produce(RepositoriesSetupTasks(to_enable=to_enable_repos))
+        api.produce(CustomTargetRepositories(repos=[CustomTargetRepository(repoid=repo) for repo in to_enable_repos]))
```
bocekm: From @pirat89:

> that's the thing. You are presenting RH repos as custom which is wrong. You are changing here the meaning of CustomTargetRepositories model at all. (this is not the only issue in the PR as original msgs CustomTargetRepository messages are not processed.)

repos/system_upgrade/el7toel8/actors/setuptargetrepos/actor.py
```diff
@@ -1,91 +1,109 @@
...
+    custom_repos = []
+    enabled_repos = set()
+    mapping_of_repositories = []  # list of RepositoryMap
+    mapped_rhel7_repoids = set()  # enabled RHEL 7 repos that have been successfully mapped to the RHEL 8 ones
+    target_repos = []
+    used_repos = {}
 
-        enabled_repos = set()
+    def process(self):
+        self.get_custom_repos()
```
bocekm: From @pirat89:

> Actually, it returns now rhel repositories. not custom:
> ```python
> p self.custom_repos
> [<leapp.models.targetrepositories.CustomTargetRepository object at 0x7f115f1ed3d0>, <leapp.models.targetrepositories.CustomTargetRepository object at 0x7f115f1ed410>, <leapp.models.targetrepositories.CustomTargetRepository object at 0x7f115f1ed450>]
> (Pdb) p self.custom_repos[0].name
> None
> (Pdb) p self.custom_repos[0].baseurl
> None
> (Pdb) p self.custom_repos[0].enabled
> True
> (Pdb) p self.custom_repos[0].repoid
> u'rhel-8-for-x86_64-baseos-rpms'
> (Pdb) p self.custom_repos[1].repoid
> u'codeready-builder-for-rhel-8-x86_64-rpms'
> (Pdb) p self.custom_repos[2].repoid
> u'rhel-8-for-x86_64-appstream-rpms'
> ...
> ```
> 
> Expected only two customrepos (produced by custom actor) with `APPSTREAM` and `BASEOS` names.

repos/system_upgrade/el7toel8/models/repositoriessetuptasks.py
```diff
@@ -6,6 +6,7 @@ class RepositoriesSetupTasks(Model):
     """
     Information about repositories that must be managed in order to complete upgrade process.
 
+    !!! THIS MODEL IS OBSOLETE. Use for example CustomTargetRepositories instead. !!!
```
bocekm: From @pirat89:

> I do not like to see the model is kind of obsoleted. I can imagine in future that based on checks, there will be case a specific repository needs to be enabled / disabled (talking about RHEL repos).
> 
> I am fun of Tasks suffix as this is kind of self explaining name we could use across many problems we would like to handle on one place. But the approach is questionable. Let's see what will be the way in future.

vojtechsokol:
leapp-ci build

Rezney:
This will be covered by future complete refactoring (part was already refactor).

PR #360, opened 2019-09-30
OpenLDAP - migrate NSS to OpenSSL format

kenoh:
OpenLDAP in el7 used to use NSS as a crypto backend. However, this
changed in el7.5 (with a backwards compaibility support) to OpenSSL
which uses different format for storing certificates. With el8 there is
no more compatibility support for NSS database, therefore there is need
to convert old-style configuration to the new one. In el8 this affects
ldap.conf configuration file but also sssd.conf which uses libldap of
OpenLDAP project with its own configuration needing correction, too.

PR #373, opened 2019-10-14
s390x fixes

pirat89:
For the s390x architecture, we need to call zipl explicitly because of the grubby issue - grubby is not calling zipl when update the bootloader configuration. Add the missing zipl calls on s390x architecture.

Additoinally, to be able to create initrd & bootloader entries for the new kernel, the ZIPL configuration has to be converted to BLS before the upgrade transaction is processed. For this purpose the target userspace is reused (the one that is created before to create the initrd & prepare the rpm transaction).

The last update includes adding the `/etc/dasd.conf` file to the very same target userspace and install it inside the dracut image explicitly to be able to work with storage on s390x arch - where DASD is required to be used. Maybe we will need to copy more files in future (maybe on other architectures as well). This will be investigated later. Currently this is solving the only known problem.

EDIT: Additionally, ensure that we do not create dulicit boot entries, otherwise zipl fails. Should be resolved by the last commit. Now the boot entries reffering to the upgrade initrd should be rmeoved before we add the upgrade entry.

Ref: PR #358 

EDIT2: I had to fix several additional issues but now it is prepared for final review and merge, BUT there are still some limitations and issues (other arches are affected as well with various impacts.

We are not able to handle removal of the old kernel in case it is needed during the upgrade. I mean the situation, when user has installed max limit number of kernels before upgrade (by default, the limit is 3, and if I am right, for upgrade it is always 3 know, as we ignore user setup of yum&dnf  guess). There are several problems:
1. The first is inside s390utils on RHEL 8.1. because of BLS detection - we are breaking it by our kernel-workaround package. The problem is fixed for adding of the new kernels, but it is not fixed for removal of old ones (that's only s390x related).
1. Another one is inside our kernel-workaround rpm that provides `/usr/sbin/new-kernel-pkg`, which is not doing enything, just exists. So kernel is not correctly removed when the scriptlet is processed. This is common for all arches.

EDIT3: Created issue for problem with networking on some s390x machines https://github.com/oamg/leapp-repository/issues/383

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
@@ -3,7 +3,7 @@ import re
 
 from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.common.config import architecture
-from leapp.libraries.stdlib import api, run
+from leapp.libraries.stdlib import api, run, CalledProcessError
 from leapp.models import BootContent
 
 
@@ -11,21 +11,41 @@ def add_boot_entry():
     debug = 'debug' if os.getenv('LEAPP_DEBUG', '0') == '1' else ''
 
     kernel_dst_path, initram_dst_path = get_boot_file_paths()
+    try:
+        _remove_old_upgrade_boot_entry(kernel_dst_path)
+        run([
+            '/usr/sbin/grubby',
+            '--add-kernel', '{0}'.format(kernel_dst_path),
+            '--initrd', '{0}'.format(initram_dst_path),
+            '--title', 'RHEL-Upgrade-Initramfs',
+            '--copy-default',
+            '--make-default',
+            '--args', '{DEBUG} enforcing=0 rd.plymouth=0 plymouth.enable=0'.format(DEBUG=debug)
+        ])
+
+        if architecture.matches_architecture(architecture.ARCH_S390X):
+            # on s390x we need to call zipl explicitly because of issue in grubby,
+            # otherwise the new boot entry will not be set as default
+            # See https://bugzilla.redhat.com/show_bug.cgi?id=1764306
+            run(['/usr/sbin/zipl'])
+    except CalledProcessError as e:
+        raise StopActorExecutionError(
+           'Cannot configure bootloader.',
+           details={'details': '{}: {}'.format(str(e), e.stderr)}
+        )
+
+
+def _remove_old_upgrade_boot_entry(kernel_dst_path):
+    """
+    Remove entry referring to the upgrade kernel.
+
+    We have to ensure there are no duplicit boot entries. Main reason is crash
+    of zipl when duplicit entries exist.
+    """
     run([
         '/usr/sbin/grubby',
-        '--add-kernel', '{0}'.format(kernel_dst_path),
-        '--initrd', '{0}'.format(initram_dst_path),
-        '--title', 'RHEL-Upgrade-Initramfs',
-        '--copy-default',
-        '--make-default',
-        '--args', '{DEBUG} enforcing=0 rd.plymouth=0 plymouth.enable=0'.format(DEBUG=debug)
+        '--remove-kernel', '{0}'.format(kernel_dst_path)
     ])
-
-    if architecture.matches_architecture(architecture.ARCH_S390X):
-        # on s390x we need to call zipl explicitly because of issue in grubby,
-        # otherwise the new boot entry will not be set as default
-        # See https://bugzilla.redhat.com/show_bug.cgi?id=1764306
-        run(['/usr/sbin/zipl'])
 
 
 def get_boot_file_paths():
--- a/repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
@@ -2,6 +2,7 @@ import os
 import re
 
 from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.common.config import architecture
 from leapp.libraries.stdlib import api, run
 from leapp.models import BootContent
 
@@ -19,6 +20,12 @@ def add_boot_entry():
         '--make-default',
         '--args', '{DEBUG} enforcing=0 rd.plymouth=0 plymouth.enable=0'.format(DEBUG=debug)
     ])
+
+    if architecture.matches_architecture(architecture.ARCH_S390X):
+        # on s390x we need to call zipl explicitly because of issue in grubby,
+        # otherwise the new boot entry will not be set as default
+        # See https://bugzilla.redhat.com/show_bug.cgi?id=1764306
+        run(['/usr/sbin/zipl'])
 
 
 def get_boot_file_paths():
--- a/repos/system_upgrade/el7toel8/actors/forcedefaultboottotargetkernelversion/libraries/forcedefaultboot.py
+++ b/repos/system_upgrade/el7toel8/actors/forcedefaultboottotargetkernelversion/libraries/forcedefaultboot.py
@@ -2,6 +2,7 @@ import os
 from collections import namedtuple
 
 from leapp.libraries import stdlib
+from leapp.libraries.common.config import architecture
 from leapp.libraries.stdlib import api
 from leapp.models import InstalledTargetKernelVersion
 
@@ -40,6 +41,11 @@ def update_default_kernel(kernel_info):
     else:
         try:
             stdlib.run(['grubby', '--set-default', kernel_info.kernel_path])
+            if architecture.matches_architecture(architecture.ARCH_S390X):
+                # on s390x we need to call zipl explicitly because of issue in grubby,
+                # otherwise the new boot entry will not be set as default
+                # See https://bugzilla.redhat.com/show_bug.cgi?id=1764306
+                stdlib.run(['/usr/sbin/zipl'])
         except (OSError, stdlib.CalledProcessError):
             api.current_logger().error('Failed to set default kernel to: %s', kernel_info.kernel_path, exc_info=True)
 
--- a/repos/system_upgrade/el7toel8/actors/forcedefaultboottotargetkernelversion/tests/test_forcedefaultboot.py
+++ b/repos/system_upgrade/el7toel8/actors/forcedefaultboottotargetkernelversion/tests/test_forcedefaultboot.py
@@ -5,12 +5,14 @@ import pytest
 
 from leapp.libraries import stdlib
 from leapp.libraries.actor import forcedefaultboot
+from leapp.libraries.common.config import architecture
 from leapp.libraries.stdlib import api
 from leapp.models import InstalledTargetKernelVersion
 
 Expected = namedtuple(
     'Expected', (
-     'grubby_setdefault'
+     'grubby_setdefault',
+     'zipl_called'
     )
 )
 
@@ -20,7 +22,8 @@ Case = namedtuple(
      'kernel_exists',
      'entry_default',
      'entry_exists',
-     'message_available'
+     'message_available',
+     'arch_s390x'
      )
 )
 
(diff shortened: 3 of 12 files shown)


pirat89:
I am going to test the solution yet on real update.

pirat89:
So It's broken. I guess it's about dasd. Not sure. I will end up when I want to load initrd.

pirat89:
I am going to test it yet, but it looks it could work.

pirat89:
There is still the problem with the booting to the initrd. I see that the `/etc/dasd.conf` is not added into the intird. That's because dracut will do that only when initrd is built with the `--hostonly` option. However, I am not sure how reliable it is, as we are building the initrd inside the container (differet rootfs), probably without access to udev I guess as well (which storage team reported to us as problem from their POV, between other things).

Additionally, even when I set the option instead of `--no-hostonly`, system did not boot into the initrd. Now with problem with mapper. I found that in such case, many utilities (lvm related) are not present in the initrd.

Regarding the check I did, it looks it should be safe to just add the file into the initrd when it is needed. I will try this stuff first instead of the `--hostonly` mode. (@oamg/developers , @msekletar)

repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
```diff
@@ -2,22 +2,48 @@
...
+            '--args', '{DEBUG} enforcing=0 rd.plymouth=0 plymouth.enable=0'.format(DEBUG=debug)
+        ])
+
+        if architecture.matches_architecture(architecture.ARCH_S390X):
+            # on s390x we need to call zipl explicitly because of issue in grubby,
+            # otherwise the new boot entry will not be set as default
+            run(['/usr/sbin/zipl'])
+    except CalledProcessError as e:
+        raise StopActorExecutionError(
+           'annot configure bootloader.',
```
```suggestion
           'Cannot configure bootloader.',
```

repos/system_upgrade/el7toel8/actors/forcedefaultboottotargetkernelversion/libraries/forcedefaultboot.py
```diff
@@ -40,6 +41,10 @@ def update_default_kernel(kernel_info):
     else:
         try:
             stdlib.run(['grubby', '--set-default', kernel_info.kernel_path])
+            if architecture.matches_architecture(architecture.ARCH_S390X):
+                # on s390x we need to call zipl explicitly because of issue in grubby,
```
bocekm: If there's a BZ for the grubby issue, link it here.

repos/system_upgrade/el7toel8/actors/forcedefaultboottotargetkernelversion/tests/test_forcedefaultboot.py
```diff
@@ -109,12 +143,45 @@ def impl(path):
     return impl
 
 
+class mocked_logger(object):
```
bocekm: I don't think you need to monkeypatch logger when you actually don't use it (there's no assert related to it).

repos/system_upgrade/el7toel8/actors/ziplconverttobls/actor.py
```diff
@@ -0,0 +1,53 @@
...
+from leapp.models import TargetUserSpaceInfo
+from leapp.tags import IPUWorkflowTag, PreparationPhaseTag
+
+
+class ZiplConvertToBLSCFG(Actor):
+    """
+    Convert the zipl boot loader configuration to the the boot loader specification on s390x systems.
+    """
+
+    name = 'zipl_convert_to_blscfg'
```
bocekm: Could you please change the name of the folder to match the actor name, i.e. ziplconverttoblscfg?

repos/system_upgrade/el7toel8/actors/ziplconverttobls/actor.py
```diff
@@ -0,0 +1,53 @@
...
+            self.log_error(
+                'TargetUserSpaceInfo is missing. Cannot execute zipl-switch-to-blscfg'
+                ' to convert the zipl configuration to BLS.'
+            )
+            return
+
+        with mounting.BindMount(source='/boot', target=os.path.join(userspace.path, 'host-boot')):
+            userspace_zipl_conf = os.path.join(userspace.path, 'etc', 'zipl.conf')
+            if os.path.exists(userspace_zipl_conf):
+                os.remove(os.path.join(userspace.path, 'etc', 'zipl.conf'))
```
bocekm:
```suggestion
                os.remove(userspace_zipl_conf)
```

repos/system_upgrade/el7toel8/actors/forcedefaultboottotargetkernelversion/tests/test_forcedefaultboot.py
pirat89: Actually you need because of mocked actor.

repos/system_upgrade/el7toel8/actors/ziplconverttobls/actor.py
pirat89: yep

repos/system_upgrade/el7toel8/actors/forcedefaultboottotargetkernelversion/libraries/forcedefaultboot.py
pirat89: I have been thinking about that when I wrote the comment, but it could be confusing because the bugzilla is about something different, and the problem is just mentioned there. So not sure how useful it would be for people:
  https://bugzilla.redhat.com/show_bug.cgi?id=1755573

pirat89:
update: the last try seems that is more sucessful. But still it failed.

pirat89:
Yeees!!!! :fireworks:  :champagne: .

pirat89:
Ready for testing and review.

asilveir:
Tested on intel and no side effects, congrats!

repos/system_upgrade/el7toel8/actors/forcedefaultboottotargetkernelversion/libraries/forcedefaultboot.py
bocekm: I think it's worth adding it there. Because if it's not, we wouldn't easily know whether the issue is fixed or not in order to decide whether we can remove this workaround.

repos/system_upgrade/el7toel8/actors/initramdiskgenerator/libraries/initramgen.py
```diff
@@ -68,11 +91,14 @@ def generate_initram_disk(context):
     copy_dracut_modules(context, modules)
     context.copy_to(generator_script, os.path.join('/', INITRAM_GEN_SCRIPT_NAME))
     install_initram_deps(context)
+    install_files = _install_files(context)
     context.call([
         '/bin/sh', '-c',
```
bocekm: FYI, based on a discussion with @AloisMahdal, this style of calling shell script is not that safe. I created an issue to deal with that later: https://github.com/oamg/leapp-repository/issues/376

repos/system_upgrade/el7toel8/actors/forcedefaultboottotargetkernelversion/libraries/forcedefaultboot.py
pirat89: @bocekm My point is, that if the BZ issue is fixed, we cannot remove it because the BZ is about different problem. but ok. I will create new one, but I am almost sure that issue will never be fixed.

repos/system_upgrade/el7toel8/actors/initramdiskgenerator/libraries/initramgen.py
```diff
@@ -68,11 +91,14 @@ def generate_initram_disk(context):
...
-        'LEAPP_ADD_DRACUT_MODULES={modules} LEAPP_KERNEL_ARCH={arch} {cmd}'.format(
+        'LEAPP_ADD_DRACUT_MODULES="{modules}" LEAPP_KERNEL_ARCH={arch} '
+        'LEAPP_DRACUT_INSTALL_FILES={files} {cmd}'.format(
```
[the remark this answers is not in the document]
pirat89: Good catch!

repos/system_upgrade/el7toel8/actors/initramdiskgenerator/libraries/initramgen.py
pirat89: Agree that this is not ideal solution. We should fix it later in different PR as this one is starting to be pretty big.

bocekm: Sure, let's fix it later in a separate PR. That was also the reason why I created the issue.

pirat89:
Currently we discovered that conversion to BLS config is broken. We are working on the fix. I did additional fix in the actor, but probably we will need fix in s390utils as well.

pirat89:
It should be fixed now, but it's not final version. Good for testing, but when disaster happens...

pirat89:
Prepare for final testing/review. There are some issues & known limitations which will be resolved in separate PRs. I will update description to share it.

repos/system_upgrade/el7toel8/actors/ziplconverttoblscfg/actor.py
```diff
@@ -0,0 +1,66 @@
...
+        if not architecture.matches_architecture(architecture.ARCH_S390X):
+            return
+        userspace = next(self.consume(TargetUserSpaceInfo), None)
+        if not userspace:
+            # actually this should not happen, but in such case, we want to still
+            # rather continue even if we boot into the old kernel, but in such
+            # case, people will have to do manual actions.
+            # NOTE: it is really just hypothetical, so no other actions are planned
+            # # now.
+            # FIXME: we do not want to continue anymore
```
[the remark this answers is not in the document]
pirat89: fixed

repos/system_upgrade/el7toel8/actors/forcedefaultboottotargetkernelversion/libraries/forcedefaultboot.py
pirat89: Done. See: See: https://bugzilla.redhat.com/show_bug.cgi?id=1764306

repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
```diff
@@ -2,22 +2,49 @@
...
+from leapp.libraries.stdlib import api, run, CalledProcessError
 from leapp.models import BootContent
 
 
 def add_boot_entry():
     debug = 'debug' if os.getenv('LEAPP_DEBUG', '0') == '1' else ''
 
     kernel_dst_path, initram_dst_path = get_boot_file_paths()
+    try:
+        _remove_old_upgrade_boot_entry(kernel_dst_path)
```
vinzenz: won't grubby fail if kernel_dst_path doesn't exist?

pirat89: It's possible, but in such case we have completely different problem because this should not happen - when the actor is processed, the path should exists, otherwise there is an error in actor that prepares that is supposed to prepare it or stop the upgrade. That's same problem as in case we create the new entry in the next command.

In my testing, this currently works for the first run, even for another one.

vinzenz: I meant that there is no bootloader entry yet for that kernel, not that the file does not exist.

IOW: You are trying to remove something that isn't there yet.

pirat89: And that's what is not doing any problem:
> In my testing, this currently works for the first run, even for another one.

So the boot entry is not there yet when I run that for the first time on completely new machine. It is question whether it is ok, but you can put any value (even nonexisting file) as the argument for `--remove-kernel` and it ends with exit code: `0`.

bocekm:
I'm getting this on max 7.6 installation (from base repo) with the `5659f72` pkg:
```
    self.run()
  File "/usr/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 63, in _do_run
    target_actor(logger=logger, messaging=messaging, config_model=config_model).run(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/leapp/actors/__init__.py", line 320, in run
    self.process(*args)
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/peseventsscanner/actor.py", line 23, in process
    pes_events_scanner('/etc/leapp/files/pes-events.json')
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py", line 34, in pes_events_scanner
    tasks = process_events(arch_events, installed_pkgs)
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py", line 336, in process_events
    p=package, r=current['to_install'][package], v=verbs[key]))
KeyError: u'openldap-servers'
```
Regarding `openldap-servers` there's this story:
https://github.com/oamg/leapp-repository/pull/360#issuecomment-542366944

pirat89:
Just squashed commits, without change. Merging.

PR #374, opened 2019-10-15
Show dnf verbose info

bocekm:
When LEAPP_DEBUG or LEAPP_VERBOSE env vars or --debug or --verbose cli options are used.
It may help with resolving dnf transaction oddities.


asilveir:
Tested and working as expected

PR #375, opened 2019-10-17
persistentnetnames: drop useless sorting

### LINKED ISSUE #317: persistentnetnamesconfig ifaces sorting
`sorted` does not sort in place but returns a sorted list.

@msekletar Is sorting really necessary? It looks like nothing in there is really order-dependent.

https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/actor.py#L52

msekletar:
Fixes #317


pirat89:
I have deja vu. Didn't we merged this already once?

PR #379, opened 2019-10-22
Add check for installed multiple kernels

pirat89:
We discovered recently that removal of old kernels is not handled
correctly during the IPU. In case the maximum number of kernels
are installed, the oldest one is automatically uninstalled during
the rpm upgrade transaction.

This could lead into unbootable state on some machines, and the
bootloader configuration can be corrupted after the removal as well.
This is problem especially on s390x machines.

To prevent any related troubles during the IPU, inhibit the IPU
on s390x unless just one kernel is installed, until the issue will
be fixed correctly.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/kernel/checkinstalledkernels/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/kernel/checkinstalledkernels/actor.py
@@ -1,0 +1,32 @@
+from leapp.actors import Actor
+from leapp.models import InstalledRedHatSignedRPM
+from leapp.tags import IPUWorkflowTag, ChecksPhaseTag
+from leapp.reporting import Report
+from leapp.libraries.actor import library
+
+
+class CheckInstalledKernels(Actor):
+    """
+    Inhibit IPU (in-place upgrade) when multiple kernels are installed.
+
+    We are not able to upgrade correctly when any kernel is expected to be
+    uninstalled during the rpm upgrade transaction now. This is especially
+    problematic on s390x architecture.
+
+    We discovered recently that removal of old kernels is not handled
+    correctly during the IPU. In case the maximum number of kernels
+    are installed, the oldest one is automatically uninstalled during
+    the rpm upgrade transaction.
+
+    To prevent any related troubles during the IPU, inhibit the IPU
+    on s390x unless just one kernel is installed, until the issue will
+    be fixed correctly.
+    """
+
+    name = 'check_installed_kernels'
+    consumes = (InstalledRedHatSignedRPM,)
+    produces = (Report,)
+    tags = (IPUWorkflowTag, ChecksPhaseTag)
+
+    def process(self):
+        library.process()
--- a/repos/system_upgrade/el7toel8/actors/kernel/checkinstalledkernels/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/kernel/checkinstalledkernels/libraries/library.py
@@ -1,0 +1,40 @@
+from leapp import reporting
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.common.config import architecture
+from leapp.libraries.stdlib import api
+from leapp.models import InstalledRedHatSignedRPM
+
+
+def _get_kernel_rpms():
+    rpms = next(api.consume(InstalledRedHatSignedRPM), InstalledRedHatSignedRPM())
+    return [pkg for pkg in rpms.items if pkg.name == 'kernel']
+
+
+def process():
+    if not architecture.matches_architecture(architecture.ARCH_S390X):
+        return
+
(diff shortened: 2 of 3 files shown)


repos/system_upgrade/el7toel8/actors/kernel/checkinstalledkernels/actor.py
```diff
@@ -0,0 +1,53 @@
...
+                reporting.Severity(reporting.Severity.HIGH),
+                reporting.Tags([reporting.Tags.KERNEL, reporting.Tags.BOOT]),
+                reporting.Flags([reporting.Flags.INHIBITOR]),
+                reporting.Remediation(hint=remediation),
+                reporting.RelatedResource('package', 'kernel')
+            ])
+
+    def _get_kernel_rpms(self):
+        rpms = next(self.consume(InstalledRedHatSignedRPM), InstalledRedHatSignedRPM())
+        return [pkg for pkg in rpms if pkg.name == 'kernel']
```
asilveir:
```suggestion
        return [pkg for pkg in rpms.items if pkg.name == 'kernel']
```

pirat89:
@asilveir I changed that a little. Now it should be applied only on s390x, based on Michals check. For testing, I suggest to remove the condition that refers on specific architecture. But I will probably modify the code again because arch-specific actors are not possible to test without library. And I want to add tests.

pirat89:
I have no idea why the test is failing. the produce should not be called anywhere. The create_report is mocked O_o

repos/system_upgrade/el7toel8/actors/kernel/checkinstalledkernels/libraries/library.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def process():
+    if not architecture.matches_architecture(architecture.ARCH_S390X):
+        return
+
+    pkgs = _get_kernel_rpms()
+    if not pkgs:
+        # Hypothatical, user is not allowed to install any kernel that is not signed by RH
+        # In case we would like to be carefuly, we could check whether there are not other
```
bocekm:
```suggestion
        # In case we would like to be cautious, we could check whether there are no other
```

repos/system_upgrade/el7toel8/actors/kernel/checkinstalledkernels/libraries/library.py
```diff
@@ -0,0 +1,41 @@
...
+        # kernels installed as well.
+        api.current_logger().log.error('Cannot find any installed kernel signed by Red Hat.')
+        raise StopActorExecutionError('Cannot find any installed kernel signed by Red Hat.')
+    if len(pkgs) > 1:
+        # It's temporary solution, so no need to try automatize everything.
+        title = 'Detected installed multiple kernels.'
```
bocekm:
```suggestion
        title = 'Multiple kernels installed'
```

repos/system_upgrade/el7toel8/actors/kernel/checkinstalledkernels/libraries/library.py
```diff
@@ -0,0 +1,41 @@
...
+        summary = ('The upgrade process is not handled well in case multiple kernels'
```
bocekm:
```suggestion
        summary = ('The upgrade process does not handle well the case when multiple kernels are installed on s390x. There is a severe risk of the bootloader configuration getting corrupted during the upgrade.'
```

repos/system_upgrade/el7toel8/actors/kernel/checkinstalledkernels/libraries/library.py
```diff
@@ -0,0 +1,41 @@
...
+                   ' are installed. In such case, there is serious risk the bootloader'
+                   ' configuration will be corrupted during the upgrade.')
+        remediation = ('Boot into the most up-to-date kernel and remove all older'
+                       ' kernels installed on the machine before the in-place'
```
bocekm:
```suggestion
                       ' kernels installed on the machine before running Leapp again.'
```

repos/system_upgrade/el7toel8/actors/kernel/checkinstalledkernels/actor.py
```diff
@@ -0,0 +1,32 @@
+from leapp.actors import Actor
+from leapp.models import InstalledRedHatSignedRPM
+from leapp.tags import IPUWorkflowTag, ChecksPhaseTag
+from leapp.reporting import Report
+from leapp.libraries.actor import library
+
+
+class CheckInstalledKernels(Actor):
+    """
+    Inhibit IPU when multiple kernels are installed.
```
bocekm:
```suggestion
    Inhibit IPU (in-place upgrade) when multiple kernels are installed.
```

PR #385, opened 2019-10-25
Fix boot order on EFI systems

shaded-enmity:
We need to make sure that we boot into the expected partition after every restart

### DIFF
--- a/repos/system_upgrade/el7toel8/libraries/efi_reboot_fix.py
+++ b/repos/system_upgrade/el7toel8/libraries/efi_reboot_fix.py
@@ -1,0 +1,44 @@
+from re import compile as regexp
+
+from leapp.libraries.stdlib import run
+
+
+_current_boot_matcher = regexp(r'BootCurrent: (?P<boot_current>(\d*))')
+_next_boot_matcher = regexp(r'BootNext: (?P<boot_next>(\d*))')
+
+
+def get_current_boot_match(string):
+    match = _current_boot_matcher.match(string)
+    if not match:
+        return None
+    captured_groups = match.groupdict()
+    return captured_groups['boot_current']
+
+
+def get_next_boot_match(string):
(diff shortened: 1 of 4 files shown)


shaded-enmity:
ping @AloisMahdal

repos/system_upgrade/el7toel8/libraries/efi_reboot_fix.py
```diff
@@ -0,0 +1,44 @@
...
+    if not match:
+        return None
+    captured_groups = match.groupdict()
+    return captured_groups['boot_next']
+
+
+def maybe_emit_updated_boot_entry():
+    efi_info = run(['/sbin/efibootmgr'], checked=False, split=True)
+    if efi_info['exit_code'] != 0:
+        # Not an efi system
```
pirat89: @shaded-enmity At least a debug msg here could be here, as what in case that something is broken (a config?) a that's why the cmd fails? Additionally, what in (BFU) case you install the rpm with /sbin/efibootmgr but use still BIOS? - I haven't tried that; It's possible it's ok.

shaded-enmity: On BIOS systems this returns -2 and "EFI variables not supported blah blah" and agreed that this could be made more robust, i.e. if this system is really UEFI and this fails, inhibit. Will think about failure modes once we have a clean run.

AloisMahdal:
With

    leapp-repository-0.9.0-0.201911011438Z.03211ae.efi_boot_order_fix.PR385.el7_6

out of 10 aarch+efi machines, all passed.  Thanks!

repos/system_upgrade/el7toel8/libraries/efi_reboot_fix.py
shaded-enmity: Resolved with latest comit

PR #386, opened 2019-10-29
overlay: Do not try to make an overlay over /boot/efi

vinzenz:
/boot/efi happens to be a vfat partition due to how UEFI is designed.
Overlayfs cannot be used on vfat partitions. Therefore we will just
skip the overlay creation for /boot/efi.


AloisMahdal:
Ran on UEFI aarch64 with

    leapp-0.9.0-100.201910242054Z.a83c46f.master.el7_6
    leapp-repository-0.9.0-0.201910301105Z.83956e6.no_overlay_over_boot_efi.PR386.el7_6

and system got upgraded; no tracebacks.

PR #393, opened 2019-10-31
Initram: Initial multipath support

vinzenz:
This patch introduces changes required to support multipath. It's not given
that all things will work out of the box, it's possible that additional
files need to be mapped, copied or bind mounted into the container
during initram creation. The bare minimum for the workings of multipath
are added.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/initramdiskgenerator/libraries/initramgen.py
+++ b/repos/system_upgrade/el7toel8/actors/initramdiskgenerator/libraries/initramgen.py
@@ -77,6 +77,17 @@ def _install_files(context):
     return []
 
 
+def install_multipath_files(context):
+    # Include multipath related files (according to module-setup of multipath)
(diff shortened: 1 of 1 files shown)


Rezney:
@vinzenz please rebase. This PR does not contain the "nsenter" patch and the upgrade is getting stuck during the rpm upgrade phase.

Rezney:
e2e tests

AloisMahdal:
I'm looking at the int/upg failures right now; my guess it's known issue OAMG-726; in that case re-run would be enough

AloisMahdal:
Bad news: it's not the same; it's OAMG-727.  Good news: I know where the problem is so I'm fixing it now.

AloisMahdal:
Looks fixed. so I've rescheduled all.

Rezney:
@AloisMahdal thanks for the re-run

PR #396, opened 2019-11-01
s390x: add actor to inhibit when rd.znet in cmdline

pirat89:
Some s390x machines have broken networking after the upgrade
on RHEL 8. Currently we are not able to tell why and which machines
have this problem, but one of possible workrounds could be remove
of the rd.znet parameter from the kernel cmdline.

Problem is that such change can break networking as well, e.g.
in case a VLAN is set. Inhibit upgrade when rd.znet is detected
on kernel cmdline.

In case any VLAN is not detected, add hint to remove the rd.znet
parameter from the cmdline.

The actor is without tests because of the lack of time and it is
temporary solution until we fix it.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkznet/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkznet/actor.py
@@ -1,0 +1,54 @@
+
+from leapp import reporting
+from leapp.actors import Actor
+from leapp.libraries.actor import library
+from leapp.libraries.common.config import architecture
+from leapp.reporting import Report
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+
+
+class CheckZnet(Actor):
+    """
+    Inhibit upgrade on the s390x architecture when rd.znet is specified on kernel cmdline.
+
+    Some s390x machines have broken networking after the upgrade on RHEL 8.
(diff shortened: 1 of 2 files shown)


examon:
leapp-ci build

examon:
leapp-ci build

repos/system_upgrade/el7toel8/actors/checkznet/actor.py
```diff
@@ -0,0 +1,56 @@
...
+                reporting.Title('Detected the rd.znet parameter in kernel cmdline'),
+                reporting.Severity(reporting.Severity.HIGH),
+                reporting.Tags(([reporting.Tags.SANITY])),
+                reporting.Flags(([reporting.Flags.INHIBITOR])),
+                reporting.Summary(
+                    'We have discovered problems with network on some s390x'
+                    ' machines when the rd.znet parameter is specified on the kernel'
+                    ' cmdline. Currently we do not know what is the root of those'
+                    ' problems. To preempt another problems, the in-place upgrade'
+                    ' process is inhibited for now.'
```
shaded-enmity: Let's adjust the messaging:

```
Upgrade on s390x machines with rd.znet kernel parameter is not supported and the upgrade has been inhibited.
```

PR #397, opened 2019-11-05
Do not add to "to_install" what is already installed

Rezney:
This will prevent potential transaction failures when user tries to install packages that are already installed (these should be upgraded).

It is also needed for gating testing where we add packages to "/etc/leapp/transaction/to_install" without knowing whether a package is or is not installed on the system.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/rpmtransactionconfigtaskscollector/libraries/scanner.py
+++ b/repos/system_upgrade/el7toel8/actors/rpmtransactionconfigtaskscollector/libraries/scanner.py
@@ -1,6 +1,7 @@
 import os.path
 
-from leapp.models import RpmTransactionTasks
+from leapp.models import RpmTransactionTasks, InstalledRedHatSignedRPM
+from leapp.libraries.stdlib import api
 
 
 def load_tasks_file(path, logger):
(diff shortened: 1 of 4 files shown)


Rezney:
@bocekm please

repos/system_upgrade/el7toel8/actors/rpmtransactionconfigtaskscollector/libraries/scanner.py
```diff
@@ -19,7 +20,11 @@ def load_tasks_file(path, logger):
 
 def load_tasks(base_dir, logger):
     # Loads configuration files to_install, to_keep, and to_remove from the given base directory
+    to_install = load_tasks_file(os.path.join(base_dir, 'to_install'), logger)
+    # we do not want to put into rpm transaction what is already installed (it will go to "to_upgrade" bucket)
+    to_install_filtered = [pkg for pkg in to_install if pkg not in api.consume(InstalledRedHatSignedRPM)]
```
bocekm: It would be worth printing a debug msg when not adding a pkg from the to_install file to the dnf plugin to_install list. Because in the end a pkg mentioned in the file may get even removed (e.g. `finger` that is no longer available on RHEL8) and it would be hard to find why the package that was to be installed got removed.

bocekm: Hm, I installed finger, ran the upgrade and got `Package finger-0.17-52.el7.x86_64 is already installed.` in the transaction, even though it should have gotten removed. I'll test it more tomorrow.

Rezney:
@bocekm debug message added...

repos/system_upgrade/el7toel8/actors/rpmtransactionconfigtaskscollector/libraries/scanner.py
Rezney: I iterated through model objects instead of names. Fixed.

repos/system_upgrade/el7toel8/actors/rpmtransactionconfigtaskscollector/libraries/scanner.py
```diff
@@ -19,7 +20,20 @@ def load_tasks_file(path, logger):
...
+    # we do not want to put into rpm transaction what is already installed (it will go to "to_upgrade" bucket)
+    to_install_filtered = [pkg for pkg in to_install if pkg not in rpm_names]
+
+    filtered = set(to_install) - set(to_install_filtered)
+    if filtered:
+        api.current_logger().debug(
+            'The following packages from "to_install" file  will be ignored as they are already installed:'
+            '\n- ' + '\n- '.join(filtered))
+
+
```
bocekm: It didn't go through pylint. Too many blank lines.

Rezney: ok, let's satisfy the linter

bocekm:
Works as expected.
`finger` is being removed despite having it in the `to_install` file. The debug msg is printed.

PR #398, opened 2019-11-05
Call "nsenter" with dnf transaction install

Rezney:
We need to enter "sysvipc" namespace on host to be able
to communicate with udev.
Resolves:
- https://bugzilla.redhat.com/show_bug.cgi?id=1712456
- https://bugzilla.redhat.com/show_bug.cgi?id=1709890

### DIFF
--- a/repos/system_upgrade/el7toel8/libraries/dnfplugin.py
+++ b/repos/system_upgrade/el7toel8/libraries/dnfplugin.py
@@ -89,7 +89,7 @@ def backup_debug_data(context):
             api.current_logger().warn('Failed to copy debugdata. Message: {}'.format(str(e)), exc_info=True)
 
 
-def _transaction(context, stage, target_repoids, tasks, test=False):
+def _transaction(context, stage, target_repoids, tasks, test=False, cmd_prefix=None):
     """
     Perform the actual DNF rpm download via our DNF plugin
     """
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/el7toel8/libraries/dnfplugin.py
```diff
@@ -89,7 +89,7 @@ def backup_debug_data(context):
             api.current_logger().warn('Failed to copy debugdata. Message: {}'.format(str(e)), exc_info=True)
 
 
-def _transaction(context, stage, target_repoids, tasks, test=False):
+def _transaction(context, stage, target_repoids, tasks, test=False, cmd_prefix=False):
```
jmikovic: just a nitpick, you should use empty list instead of False as a default for cmd_prefix if you are expecting lists there

Rezney: makes sense, fixed...

Rezney: dangerous-default-value :) 

"None" it is then...

repos/system_upgrade/el7toel8/libraries/dnfplugin.py
```diff
@@ -106,6 +106,8 @@ def _transaction(context, stage, target_repoids, tasks, test=False):
         ]
         if config.is_verbose():
             cmd.append('-v')
+        if cmd_prefix:
+            cmd[0:0] = cmd_prefix
```
vinzenz: while this works, I think it's easier digestable as
```suggestion
            cmd = cmd_prefix + cmd
```

Rezney: done

pirat89: anyway, it's nice trick to know :)

PR #400, opened 2019-11-07
CheckKernelDriver: Check kernel drivers instead of modules

drehak:
The previous version of `CheckKernelDrivers` compared the list of drivers removed in RHEL 8 against present kernel modules. The problem is, drivers and modules don't always map 1:1 to each other and in some cases, their names differ.

This PR reworks `CheckKernelDrivers` to work with driver names. Those are now produced in `PCIDeviceScanner`, which is also reworked along with the `PCIDevices` model, to which more fields are added. `WhitelistKernelModules` and its model `WhitelistedKernelModules` are removed, since they are no longer needed.


drehak:
Squashed and rebased. DON'T MERGE yet, CheckKernelDrivers still has old tests.

pirat89:
@drehak How are you sure that whitelist is not needed anymore? I am missing little more context.As well - the PR is missing the description.

drehak:
> @drehak How are you sure that whitelist is not needed anymore? I am missing little more context.

According to the `WhitelistKernelModules.process(self)` docstring, the actor was pretty much a little hack to allow upgrades of vagrant machines. Now the upgrade passes anyway, so it's no longer needed. 
> As well - the PR is missing the description.

Yeah, I'll add it.

drehak:
leapp-ci build

repos/system_upgrade/el7toel8/actors/kernel/checkkerneldrivers/actor.py
```diff
@@ -16,59 +16,51 @@ class CheckKernelDrivers(Actor):
...
     def process(self):
-        with open('files/removed_drivers.txt', 'r') as removed:
-            removed_drivers = []
-            whitelisted_modules = set()
-            collected_drivers = set()
+        removed_drivers = set()
+        present_drivers = set()
+        conflicting_drivers = set()
 
+        with open('files/removed_drivers.txt', 'r') as removed:
```
vinzenz: exception handling required, to harden the code

Plus, since you rewrite this actor I'd recommend to move this to a library so you can better unit test the code.

drehak: Alright, handled exceptions and refactored the code a bit.

repos/system_upgrade/el7toel8/actors/kernel/checkkerneldrivers/libraries/library.py
```diff
@@ -0,0 +1,34 @@
...
+                    continue
+                removed.add(token)
+        return removed
+    except (IOError, OSError) as e:
+        raise StopActorExecutionError('Cannot read {}: {}'.format(os.path.abspath(path), str(e)))
+
+
+def get_present_drivers():
+    present = set()
+    for fact in api.consume(PCIDevices):
```
fernflower: Nit: 
```
devices = [d for f in api.consume(PCIDevices) for d in f.devices if d.driver]
return set(devices)
```

repos/system_upgrade/el7toel8/actors/pcidevicesscanner/libraries/pcidevicesscanner.py
```diff
@@ -61,7 +62,7 @@ def produce_pci_devices(producer, devices):
 def scan_pci_devices(producer):
     ''' Scan system PCI Devices '''
     try:
```
fernflower: Nit: you can use checked=False to suppress exception raise
```
output = run(['lspci', '-vmmk'], checked=False).get('stdout', [])
```

drehak: Why would I want to supress that? As far as I know, `lspci` can't exit with a non-zero code.

repos/system_upgrade/el7toel8/actors/kernel/checkkerneldrivers/libraries/library.py
drehak: Good idea, even though this seems a bit opaque. And I'm collecting drivers, not devices.

drehak: Replaced with a set comprehension.

repos/system_upgrade/el7toel8/actors/pcidevicesscanner/libraries/pcidevicesscanner.py
drehak: (Applied after a brief discussion.)

PR #401, opened 2019-11-07
Simplify checking supported OS

bocekm:
- Use the common library function `is_supported_version()` instead of duplicating its functionality
- Rename LEAPP_SKIP_CHECK_OS_RELEASE to LEAPP\_**DEVEL**\_SKIP_CHECK_OS_RELEASE to make it more clear that using this option is not supported

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkosrelease/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkosrelease/actor.py
@@ -1,6 +1,5 @@
 from leapp.actors import Actor
 from leapp.libraries.actor.library import check_os_version, skip_check
-from leapp.libraries.common.config import version
 from leapp.models import OSReleaseFacts
 from leapp.reporting import Report
 from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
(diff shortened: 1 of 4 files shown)


repos/system_upgrade/el7toel8/actors/checkosrelease/actor.py
```diff
@@ -1,18 +1,15 @@
...
-from leapp.libraries.common.config import version
 from leapp.models import OSReleaseFacts
 from leapp.reporting import Report
 from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
 
 
 class CheckOSRelease(Actor):
     """
-    Check if a supported release version of system's OS is in use. If not, inhibit upgrade process.
+    Check if the RHEL 7 minor version in use is supported. If not, inhibit upgrade process.
```
vinzenz: Since this actor is depending on the configuration (IIRC) we should probably rephrase this even more to something like this:
```suggestion
    Check if the current RHEL minor version in use is supported. If not, inhibit upgrade process.
```

repos/system_upgrade/el7toel8/actors/checkosrelease/libraries/library.py
```diff
@@ -13,10 +11,10 @@
 
 def skip_check():
     """ Check if an environment variable was used to skip this actor """
-    if os.getenv('LEAPP_SKIP_CHECK_OS_RELEASE'):
+    if os.getenv('LEAPP_DEVEL_SKIP_CHECK_OS_RELEASE'):
```
vinzenz: @AloisMahdal heads up for the change of the variable name

vinzenz:
Other than the description nit pick LGTM

repos/system_upgrade/el7toel8/actors/checkosrelease/actor.py
bocekm: Isn't the _"current"_ redundant to the _"in use"_ then?

repos/system_upgrade/el7toel8/actors/checkosrelease/libraries/library.py
pirat89: Just pointing out that we had discussions about the names recently and currently there are thoughts about: `LEAPP_DEVEL_UNSUPPORTED` prefix. Keeping that opened until we finish the discussion. I guess, that should happen this tuesday.

PR #402, opened 2019-11-11
add upgrade instruction link to ha cluster check

idevat:
Add a link to documentation that deals with upgrading HA clusters.


shaded-enmity:
thanks @idevat

PR #404, opened 2019-11-14
Inhibit upgrade when it is not supported

drehak:
Current inhibition cases:
- one or more `LEAPP_DEVEL_*` variables are set
- one or more experimental actors are enabled by `--whitelist-experimental`

Requires PR: https://github.com/oamg/leapp/pull/577

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
@@ -1,11 +1,10 @@
-import os
 import json
 
+from leapp import reporting
 from leapp.actors import Actor
 from leapp.models import Report
 from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
 from leapp.utils.audit import get_connection
-from leapp import reporting
 
 
 class UnsupportedUpgradeCheck(Actor):
@@ -34,12 +33,15 @@ class UnsupportedUpgradeCheck(Actor):
                     'These variables are for development purposes only and can interfere with the upgrade '
                     'process in unexpected ways. As such, a successful and safe upgrade process cannot be '
                     'guaranteed and the upgrade is unsupported.\n'
-                    'Found development variables:\n- ' + '\n- '.join([var.name for var in devel_vars]) + '\n'
+                    'You can bypass this error by setting the LEAPP_UNSUPPORTED variable but by doing so, '
+                    'you continue at your own risk.\n'
+                    'Found development variables:\n- {}\n'.format('\n- '.join([v.name for v in devel_vars]))
                 ),
                 reporting.Severity(reporting.Severity.HIGH),
                 reporting.Flags([reporting.Flags.INHIBITOR]),
                 reporting.Tags([reporting.Tags.UPGRADE_PROCESS, reporting.Tags.SANITY]),
-                reporting.Remediation(hint='Invoke leapp without any LEAPP_DEVEL_* environment variables.')
+                reporting.Remediation(hint=('Invoke leapp without any LEAPP_DEVEL_* environment variables '
+                                            'or set LEAPP_UNSUPPORTED=1.'))
             ])
 
         experimental = []
@@ -57,10 +59,13 @@ class UnsupportedUpgradeCheck(Actor):
                     'These actors are unstable or in development and can interfere with the upgrade '
                     'process in unexpected ways. As such, a successful and safe upgrade process cannot be '
                     'guaranteed and the upgrade is unsupported.\n'
+                    'You can bypass this error by setting the LEAPP_UNSUPPORTED variable but by doing so, '
+                    'you continue at your own risk.\n'
                     'Found enabled experimental actors:\n- ' + '\n- '.join(experimental) + '\n'
                 ),
                 reporting.Severity(reporting.Severity.HIGH),
                 reporting.Flags([reporting.Flags.INHIBITOR]),
                 reporting.Tags([reporting.Tags.UPGRADE_PROCESS, reporting.Tags.SANITY]),
-                reporting.Remediation(hint='Invoke leapp without any --whitelist-experimental options.')
+                reporting.Remediation(hint=('Invoke leapp without any --whitelist-experimental options '
+                                            'or set LEAPP_UNSUPPORTED=1.'))
             ])
--- a/repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
@@ -23,39 +23,52 @@ class UnsupportedUpgradeCheck(Actor):
         experimental = bool([v for v in leapp_vars if v.name == 'LEAPP_EXPERIMENTAL' and v.value == '1'])
         override = bool([v for v in leapp_vars if v.name == 'LEAPP_UNSUPPORTED' and v.value == '1'])
 
-        if devel_vars and not override:
+        if override:
             reporting.create_report([
-                reporting.Title('Upgrade inhibited due to usage of development variables'),
+                reporting.Title('Upgrade is unsupported'),
                 reporting.Summary(
-                    'One or more environment variables in the form of LEAPP_DEVEL_* have been detected.\n'
-                    'These variables are for development purposes only and can interfere with the upgrade '
-                    'process in unexpected ways. As such, a successful and safe upgrade process cannot be '
-                    'guaranteed and the upgrade is unsupported.\n'
-                    'You can bypass this error by setting the LEAPP_UNSUPPORTED variable but by doing so, '
-                    'you continue at your own risk.\n'
-                    'Found development variables:\n- {}\n'.format('\n- '.join([v.name for v in devel_vars]))
+                    'Environment variable LEAPP_UNSUPPORTED has been detected. A successful and safe '
+                    'upgrade process cannot be guaranteed. From now on you are continuing at your own '
+                    'risk.\n'
                 ),
                 reporting.Severity(reporting.Severity.HIGH),
-                reporting.Flags([reporting.Flags.INHIBITOR]),
                 reporting.Tags([reporting.Tags.UPGRADE_PROCESS, reporting.Tags.SANITY]),
-                reporting.Remediation(hint=('Invoke leapp without any LEAPP_DEVEL_* environment variables '
-                                            'or set LEAPP_UNSUPPORTED=1.'))
             ])
 
-        if experimental and not override:
-            reporting.create_report([
-                reporting.Title('Upgrade inhibited due to enabled experimental actors'),
-                reporting.Summary(
-                    'One or more enabled experimental actors have been detected.\n'
-                    'These actors are unstable or in development and can interfere with the upgrade '
-                    'process in unexpected ways. As such, a successful and safe upgrade process cannot be '
-                    'guaranteed and the upgrade is unsupported.\n'
-                    'You can bypass this error by setting the LEAPP_UNSUPPORTED variable but by doing so, '
-                    'you continue at your own risk.\n'
-                ),
-                reporting.Severity(reporting.Severity.HIGH),
-                reporting.Flags([reporting.Flags.INHIBITOR]),
-                reporting.Tags([reporting.Tags.UPGRADE_PROCESS, reporting.Tags.SANITY]),
-                reporting.Remediation(hint=('Invoke leapp without any --whitelist-experimental options '
-                                            'or set LEAPP_UNSUPPORTED=1.'))
-            ])
+        else:
+            if devel_vars:
+                reporting.create_report([
+                    reporting.Title('Upgrade inhibited due to usage of development variables'),
+                    reporting.Summary(
+                        'One or more environment variables in the form of LEAPP_DEVEL_* have been detected.\n'
+                        'These variables are for development purposes only and can interfere with the upgrade '
+                        'process in unexpected ways. As such, a successful and safe upgrade process cannot be '
+                        'guaranteed and the upgrade is unsupported.\n'
+                        'You can bypass this error by setting the LEAPP_UNSUPPORTED variable but by doing so, '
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
```diff
@@ -0,0 +1,63 @@
...
+            reporting.create_report([
+                reporting.Title('Upgrade inhibited due to usage of development variables'),
+                reporting.Summary(
+                    'One or more environment variables in the form of LEAPP_DEVEL_* have been detected.\n'
+                    'These variables are for development purposes only and can interfere with the upgrade '
+                    'process in unexpected ways. As such, a successful and safe upgrade process cannot be '
+                    'guaranteed and the upgrade is unsupported.\n'
+                    'Found development variables:\n- ' + '\n- '.join(debug_vars) + '\n'
+                ),
+                reporting.Severity(reporting.Severity.HIGH),
```
vinzenz: I am not sure where this inhibitor comes from. I am out of the loop on this, but according to the task description this should be a warning only, not an inhibitor.

pirat89: @drehak do not use `os.environ`. See `IPUConfig` - that means `self.configuration.leapp_env_vars`. - That means, you can just check whether variable are there or not. As well, when you require new env var to be there, modify the related actor.

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
```diff
@@ -0,0 +1,66 @@
...
+                reporting.Remediation(hint='Invoke leapp without any LEAPP_DEVEL_* environment variables.')
+            ])
+
+        experimental = []
+        with get_connection(None) as db:
+            conf = db.execute("SELECT configuration FROM execution "
+                              "WHERE kind = 'upgrade' OR kind = 'preupgrade' "
+                              "ORDER BY id DESC LIMIT 1").fetchone()
+            if conf:
+                experimental = json.loads(conf[0])["whitelist_experimental"]
```
pirat89: again, something like input parameters, should be the think you should find inside configuration. WDYT? Or maybe better, leapp should inform about that. As already inform about the debug option is used. @vinzenz which one of those two ideas you prefer?

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
```diff
@@ -0,0 +1,66 @@
...
+                    'One or more enabled experimental actors have been detected.\n'
+                    'These actors are unstable or in development and can interfere with the upgrade '
+                    'process in unexpected ways. As such, a successful and safe upgrade process cannot be '
+                    'guaranteed and the upgrade is unsupported.\n'
+                    'Found enabled experimental actors:\n- ' + '\n- '.join(experimental) + '\n'
+                ),
+                reporting.Severity(reporting.Severity.HIGH),
+                reporting.Flags([reporting.Flags.INHIBITOR]),
+                reporting.Tags([reporting.Tags.UPGRADE_PROCESS, reporting.Tags.SANITY]),
+                reporting.Remediation(hint='Invoke leapp without any --whitelist-experimental options.')
```
pirat89: In both cases, the hint is incomplete. We should put there complete hint - think about developers. These are hints to run leapp in supported way. We should inform that when people want to do unsupported upgrade, e.g. for testing, we should inform them what to do. With clarification what it mean does. That they can report bug when they hit issue e.g. because they used the experimental actor, but they cannot expect any support.

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
drehak: Alright, I'm getting them from there now.

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
drehak: I don't fully get what you mean, could you give me an example of a good hint?

pirat89: I mean inform them that they can use LEAPP_UNSUPPORTED as well. Otherwise people will start to ask why they are blocked...

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
```diff
@@ -0,0 +1,66 @@
+import os
+import json
```
fernflower: Wrong import order

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
```diff
@@ -0,0 +1,66 @@
...
+
+from leapp.actors import Actor
+from leapp.models import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+from leapp.utils.audit import get_connection
+from leapp import reporting
```
fernflower: Same here. Wondering why didn't linter catch that

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
```diff
@@ -0,0 +1,66 @@
...
+                experimental = json.loads(conf[0])["whitelist_experimental"]
+        if experimental and not override:
+            reporting.create_report([
+                reporting.Title('Upgrade inhibited due to enabled experimental actors'),
+                reporting.Summary(
+                    'One or more enabled experimental actors have been detected.\n'
+                    'These actors are unstable or in development and can interfere with the upgrade '
+                    'process in unexpected ways. As such, a successful and safe upgrade process cannot be '
+                    'guaranteed and the upgrade is unsupported.\n'
+                    'Found enabled experimental actors:\n- ' + '\n- '.join(experimental) + '\n'
```
fernflower: As per guidelines we agreed to prepare format over string concatenation

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
fernflower: ```hint='Invoke leapp without any --whitelist-experimental options or set LEAPP_UNSUPPORTED=1'```

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
```diff
@@ -0,0 +1,66 @@
...
+                reporting.Severity(reporting.Severity.HIGH),
+                reporting.Flags([reporting.Flags.INHIBITOR]),
+                reporting.Tags([reporting.Tags.UPGRADE_PROCESS, reporting.Tags.SANITY]),
+                reporting.Remediation(hint='Invoke leapp without any LEAPP_DEVEL_* environment variables.')
+            ])
+
+        experimental = []
+        with get_connection(None) as db:
+            conf = db.execute("SELECT configuration FROM execution "
+                              "WHERE kind = 'upgrade' OR kind = 'preupgrade' "
```
fernflower: Why are we selecting kind=upgrade? Do we have post-upgrade reporting as well already?

fernflower: And another stupid question - shouldn't we be querying for some specific execution id instead of select * limit 1?

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
drehak: Alright, there's format now.

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
drehak: os isn't even used, we should check the linter if it works like it should

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
drehak: Reordered.

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
pirat89: @fernflower I would rather ask, why there is communication with db itself from the actor side. That's playground of framework and actors should receive the information in different way.

drehak: @pirat89 @fernflower alright. How can I get this information without querying the database? As far as I can see, information about experimental actors is not in the configuration.

drehak: @fernflower 
>shouldn't we be querying for some specific execution id instead of select * limit 1?

Yeah, that would be better.

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
```diff
@@ -0,0 +1,62 @@
...
+    This can be overridden by setting the variable LEAPP_UNSUPPORTED (at user's own risk).
+    """
+
+    name = 'unsupported_upgrade_check'
+    consumes = ()
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        conf = self.configuration
```
fernflower: Nit: conf is not used anywhere else, I'd abandon it and have
`leapp_vars = self.configuration.leapp_env_vars`

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
```diff
@@ -0,0 +1,62 @@
...
+        leapp_vars = conf.leapp_env_vars
+        devel_vars = [v for v in leapp_vars if v.name.startswith('LEAPP_DEVEL_')]
+        experimental = bool([v for v in leapp_vars if v.name == 'LEAPP_EXPERIMENTAL' and v.value == '1'])
```
bocekm: LEAPP_EXPERIMENTAL env var is not being set anywhere. So if you use e.g. `--whitelist-experimental removeleftoverpackages` the upgrade is not inhibited.

bocekm: Ah, I overlooked the https://github.com/oamg/leapp/pull/577. Nevermind then.

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
drehak: Hint completed.

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
drehak: It's in the configuration now.

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
drehak: No queries now!

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
drehak: Alright. :)

drehak:
@fernflower Please check the newest report. Is it alright to have no hint there? I really don't know...

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
```diff
@@ -0,0 +1,74 @@
...
+        leapp_vars = self.configuration.leapp_env_vars
+        devel_vars = [v for v in leapp_vars if v.name.startswith('LEAPP_DEVEL_')]
+        experimental = bool([v for v in leapp_vars if v.name == 'LEAPP_EXPERIMENTAL' and v.value == '1'])
+        override = bool([v for v in leapp_vars if v.name == 'LEAPP_UNSUPPORTED' and v.value == '1'])
+
+        if override:
+            reporting.create_report([
+                reporting.Title('Upgrade is unsupported'),
+                reporting.Summary(
+                    'Environment variable LEAPP_UNSUPPORTED has been detected. A successful and safe '
```
fernflower: Love the change, thanks! Booting a vm for a fullstack upgrade run, should get the results tomorrow the latest.

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
```diff
@@ -23,39 +23,52 @@ def process(self):
...
+                        'guaranteed and the upgrade is unsupported.\n'
+                        'You can bypass this error by setting the LEAPP_UNSUPPORTED variable but by doing so, '
+                        'you continue at your own risk.\n'
+                    ),
+                    reporting.Severity(reporting.Severity.HIGH),
+                    reporting.Flags([reporting.Flags.INHIBITOR]),
+                    reporting.Tags([reporting.Tags.UPGRADE_PROCESS, reporting.Tags.SANITY]),
+                    reporting.Remediation(hint=('Invoke leapp without any --whitelist-experimental options '
+                                                'or set LEAPP_UNSUPPORTED=1.'))
+                ])
```
pirat89: nitpick: these two reports are pretty same and they could be merged into the one report with generic title and generated/variable summary and hints. I will not block it is really just for people using devel/testing stuff anyway.

repos/system_upgrade/el7toel8/actors/unsupportedupgradecheck/actor.py
fernflower: Sorry for taking too long to confirm, works as expected

PR #405, opened 2019-11-19
WIP: Print warning instead of raising exception

Rezney:
In some cases, the package can be installed only as part of a module
and we cannot put it directly into transaction (e.g. python3-nss)
Until this case is sorted, we will print a warning instead of failing
whole transaction.

### DIFF
--- a/repos/system_upgrade/el7toel8/files/rhel_upgrade.py
+++ b/repos/system_upgrade/el7toel8/files/rhel_upgrade.py
@@ -44,9 +44,9 @@ class RhelUpgradeCommand(dnf.cli.Command):
             except dnf.exceptions.MarkingError:
                 pkgs_notfound.append(pkg_spec)
         if pkgs_notfound:
-            err_str = ('Packages marked by Leapp for installation/removal/upgrade not found '
-                       'in repository metadata: ') + ' '.join(pkgs_notfound)
-            raise dnf.exceptions.MarkingError(err_str)
+            err_str = ('Packages marked by Leapp for {} not found '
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/el7toel8/files/rhel_upgrade.py
```diff
@@ -44,9 +44,9 @@ def _process_packages(self, pkg_set, op):
             except dnf.exceptions.MarkingError:
                 pkgs_notfound.append(pkg_spec)
         if pkgs_notfound:
-            err_str = ('Packages marked by Leapp for installation/removal/upgrade not found '
-                       'in repository metadata: ') + ' '.join(pkgs_notfound)
-            raise dnf.exceptions.MarkingError(err_str)
+            err_str = ('Packages marked by Leapp for {} not found '
+                       'in repositories metadata: '.format(op.__name__) + ' '.join(pkgs_notfound))
+            print('Warning: ' + err_str, file=sys.stderr)
```
Rezney: @j-mracek Any idea how to achieve printing the warnings either at the beginning or at the end? Currently it is put randomly in the middle of the progress. Thx...

pirat89:
This is curretly blocking people from doing an upgrade but obviously the problem with missing rpms is here since the start. Let's merge it now and improve/resolve it in future.

Reported issue #406

repos/system_upgrade/el7toel8/files/rhel_upgrade.py
j-mracek: What about to use stdout.flush()?

PR #409, opened 2019-11-22
Dump grub2-editenv list

Rezney:
before and after setting default kernel to see which one is the
real default


Rezney:
@pirat89 please take a look...

Rezney:
@pirat89 commit message modified...

bocekm:
It seems like the `grub2-editenv` command fails on s390x, see OAMG-2972.

PR #410, opened 2019-11-25
Add "requests" package also to requirements

Rezney:


pirat89:
@jmikovic  This is wrong fix. The requirement should not be inside the leapp-repository as nothing from leapp-repository requires it. The issue is that requirements of leapp have not been installed. This should be fixed.

PR #411, opened 2019-11-25
Add temporary experimental debug grubby actor

Rezney:
This is needed in order to troubleshoot our issue with booting into RHEL7 kernel.


Rezney:
@pirat89 please take a look...

Issue #383, opened 2019-10-24
s390x: network is not running after upgrade on some machines


pirat89:
There is another known issue with networking on some machines (usually those with older HW, where RHEL-ALT cannot be installed. It's related somehow to this issue:
- https://bugzilla.redhat.com/show_bug.cgi?id=1087502

The problem is somewhere in the initrd. We are not able to create initrd with proper content You can find the cww_init script fails in all cases.

Manual workrounds: Network can be provisioned manually on the machine, e.g. like this:
```bash
echo 1 > /sys/bus/ccwgroup/devices/0.0.8000/layer2
echo 1 > /sys/bus/ccwgroup/devices/0.0.8000/online
```
(that's for *enccw0.0.8000* NIC)

When network is running, rebuild the initrd (adding **qeth_l2** driver and everything will be ok after next reboot. Another possibility is the remove of rd.znet from the kernel cmdline, but in such case, network will be broken if VLAN is used.

Those workrounds doesn't seem safe enought to implement, so we are investigating how to build correct initrd.

pirat89:
I put it into the current milestone, but maybe we will not be able to fix it in time, so in such case, I will move it ot the next one.

PR #412, opened 2019-11-27
initrd: have upgrade.target depend on dracut-cmdline.service

lkundrak:
On s390, there's a cmdline hook that sets the ccw.conf contents to
what's been specified by rd.znet parameter. Without it, the ccw_init
udev helper doesn't bind the qeth devices.

Without the qeth devices being available in the leapp session, the
target dracut run doesn't detect that it's necessary to install
the qeth_l2/qeth_l3 drivers into the new hostonly initrd and thus the
upgraded systemd ends up without the network devices.

That us just sad. Let's bring up the dracut-cmdline.service. It fixes
the issue in s390 and should do no harm on other platforms.

Fixes: https://github.com/oamg/leapp-repository/issues/383
Signed-off-by: Lubomir Rintel <lkundrak@v3.sk>


lkundrak:
Note I've now fully tested this on s390, but I didn't test on x86_64 or any other architecture, because I have no idea how to do so. Help appreciated.

PR #413, opened 2019-11-27
Check minimum memory requirements to upgrade to RHEL8

fellipeh:
These actor will check the minimum memory requirements to upgrade to RHEL8, based on these document: https://access.redhat.com/articles/rhel-limits

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkmemory/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkmemory/actor.py
@@ -1,0 +1,20 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import library
+from leapp.models import MemoryInfo, Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckMemory(Actor):
+    """
+    The actor check the size of RAM against RHEL8 minimal hardware requirements
+
+    Using the following resource: https://access.redhat.com/articles/rhel-limits
+    """
+
+    name = 'checkmemory'
+    consumes = (MemoryInfo,)
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        library.process()
--- a/repos/system_upgrade/el7toel8/actors/checkmemory/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/checkmemory/libraries/library.py
@@ -1,0 +1,44 @@
+from leapp import reporting
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.common.config import architecture
+from leapp.libraries.stdlib import api
+from leapp.models import MemoryInfo
+
+min_req_memory = {
+    architecture.ARCH_X86_64: 1536,  # 1.5G
+    architecture.ARCH_ARM64: 2048,  # 2Gb
+    architecture.ARCH_PPC64LE: 2048,  # 2Gb
+    architecture.ARCH_S390X: 1024  # 1Gb
+}
+
+
+def _check_memory(mem_info):
+    msg = {}
+
+    for arch, min_req in iter(min_req_memory.items()):
+        if architecture.matches_architecture(arch):
+            is_ok = mem_info.mem_total >= min_req
+            msg = {} if is_ok else {'detected': mem_info.mem_total,
+                                    'minimal_req': min_req}
+
+    return msg
+
+
+def process():
+    memoryinfo = next(api.consume(MemoryInfo), None)
+    if memoryinfo is None:
+        raise StopActorExecutionError(message=("Missing information about Memory."))
+
+    minimum_req_error = _check_memory(memoryinfo)
+
+    if minimum_req_error:
+        title = 'Minimum memory requirements for RHEL 8 are not met'
+        summary = 'Memory detected: {} KiB, required: {} KiB'.format(minimum_req_error['detected'],
+                                                                     minimum_req_error['minimal_req'])
+        reporting.create_report([
+                          reporting.Title(title),
+                          reporting.Summary(summary),
+                          reporting.Severity(reporting.Severity.HIGH),
+                          reporting.Tags([reporting.Tags.SANITY]),
+                          reporting.Flags([reporting.Flags.INHIBITOR]),
+                      ])
--- a/repos/system_upgrade/el7toel8/actors/checkmemory/tests/test_checkmemory.py
+++ b/repos/system_upgrade/el7toel8/actors/checkmemory/tests/test_checkmemory.py
@@ -1,0 +1,44 @@
+from collections import namedtuple
+
+import pytest
+
+from leapp import reporting
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.actor import library
+from leapp.libraries.common.testutils import create_report_mocked
+from leapp.libraries.common.config import architecture
+from leapp.libraries.stdlib import api
+from leapp.models import MemoryInfo
+
+
+class CurrentActorMocked(object):
+    def __init__(self, arch):
+        self.configuration = namedtuple('configuration', ['architecture'])(arch)
+
+    def __call__(self):
+        return self
+
+
+def test_check_memory_low(monkeypatch):
+    minimum_req_error = []
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(architecture.ARCH_X86_64))
+    minimum_req_error = library._check_memory(MemoryInfo(mem_total=1024))
+    assert minimum_req_error
+
+
+def test_check_memory_high(monkeypatch):
+    minimum_req_error = []
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(architecture.ARCH_X86_64))
+    minimum_req_error = library._check_memory(MemoryInfo(mem_total=16273492))
+    assert not minimum_req_error
+
+
+def test_report(monkeypatch):
+    title_msg = 'Minimum memory requirements for RHEL 8 are not met'
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(architecture.ARCH_X86_64))
+    monkeypatch.setattr(api, 'consume', lambda x: iter([MemoryInfo(mem_total=129)]))
+    monkeypatch.setattr(reporting, "create_report", create_report_mocked())
+    library.process()
+    assert reporting.create_report.called
+    assert title_msg == reporting.create_report.report_fields['title']
+    assert reporting.Flags.INHIBITOR in reporting.create_report.report_fields['flags']
--- a/repos/system_upgrade/el7toel8/actors/scanmemory/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/scanmemory/libraries/library.py
@@ -1,0 +1,24 @@
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.stdlib import api
+from leapp.models import MemoryInfo
+
+
+def _get_memoryinfo(filename='/proc/meminfo'):
+    """ Returns dict of all memory information from /proc/meminfo file """
+
+    try:
+        with open(filename) as fp:
+            # format of the lines: Key:  value unit
+            # e.g.:                MemTotal:   1024 kB
+            return dict((i.split()[0].rstrip(':'), int(i.split()[1])) for i in fp.readlines())
+    except IOError:
+        raise StopActorExecutionError(
+            'Could not read memory values',
+            details={'details': 'Error opening file {}'.format(filename)},
+        )
+
+
+def process():
+    mem_info = _get_memoryinfo()
+    if mem_info:
+        api.produce(MemoryInfo(mem_total=mem_info['MemTotal']))
--- a/repos/system_upgrade/el7toel8/actors/scanmemory/tests/test_scanmemory.py
+++ b/repos/system_upgrade/el7toel8/actors/scanmemory/tests/test_scanmemory.py
@@ -1,0 +1,24 @@
+import mock
+
+from leapp.libraries.actor import library
+from leapp.libraries.common import testutils
+from leapp.libraries.stdlib import api
+from leapp.models import MemoryInfo
+
+
+def test_with_low_memory(monkeypatch):
+    with mock.patch("__builtin__.open", mock.mock_open(read_data="MemTotal: 42 kB")) as mock_proc_meminfo:
+        monkeypatch.setattr(api, 'produce', testutils.produce_mocked())
+        library.process()
+        mock_proc_meminfo.assert_called_once_with('/proc/meminfo')
+        assert api.produce.called == 1
+        assert MemoryInfo(mem_total=42) == api.produce.model_instances[0]
+
+
+def test_with_high_memory(monkeypatch):
+    with mock.patch("__builtin__.open", mock.mock_open(read_data="MemTotal: 424242424242 kB")) as mock_proc_meminfo:
+        monkeypatch.setattr(api, 'produce', testutils.produce_mocked())
+        library.process()
+        mock_proc_meminfo.assert_called_once_with('/proc/meminfo')
+        assert api.produce.called == 1
+        assert MemoryInfo(mem_total=424242424242) == api.produce.model_instances[0]
--- a/repos/system_upgrade/el7toel8/models/memoryinfo.py
+++ b/repos/system_upgrade/el7toel8/models/memoryinfo.py
@@ -1,0 +1,15 @@
+from leapp.models import Model, fields
+from leapp.topics import SystemFactsTopic
+
+
+class MemoryInfo(Model):
+    """
+    Represents information about available memory (KiB)
+    """
+
+    topic = SystemFactsTopic
+
+    mem_total = fields.Nullable(fields.Integer())
+    """
+    Total memory in KiB
+    """
--- a/repos/system_upgrade/el7toel8/actors/scanmemory/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/scanmemory/actor.py
@@ -1,0 +1,16 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import library
+from leapp.models import MemoryInfo
+from leapp.tags import IPUWorkflowTag, FactsPhaseTag
+
+
+class ScanMemory(Actor):
+    """Scan Memory of the machine."""
+
+    name = 'scanmemory'
+    consumes = ()
+    produces = (MemoryInfo,)
+    tags = (IPUWorkflowTag, FactsPhaseTag)
+
+    def process(self):
+        library.process()


repos/system_upgrade/el7toel8/actors/checkminimumrequirements/libraries/library.py
```diff
@@ -0,0 +1,89 @@
...
+        remediation = ('Please check your hardware and provide upgrade, see: ' + URL)
+
+        create_report([
+                          reporting.Title(title),
+                          reporting.Summary(summary),
+                          reporting.Severity(reporting.Severity.HIGH),
+                          reporting.Tags([reporting.Tags.OS_FACTS]),
+                          reporting.Flags([reporting.Flags.INHIBITOR]),
+                          reporting.Remediation(hint=remediation)
+                      ] + [reporting.RelatedResource('minimum_hardware_requirements', i) for i in minimum_req_error])
```
pirat89: There is no resource of this type no but regarding discussion with pavel, you should add resources like `disk`, `memory`, ... based on what requirements are not met.

repos/system_upgrade/el7toel8/actors/checkminimumrequirements/libraries/library.py
```diff
@@ -0,0 +1,89 @@
...
+        URL = ('https://access.redhat.com/articles/rhel-limits')
+        summary = ('The following hardware requirements is not detected:'
+                   '\n{}.'.format('\n     - '.join(minimum_req_error)))
+        remediation = ('Please check your hardware and provide upgrade, see: ' + URL)
+
+        create_report([
+                          reporting.Title(title),
+                          reporting.Summary(summary),
+                          reporting.Severity(reporting.Severity.HIGH),
+                          reporting.Tags([reporting.Tags.OS_FACTS]),
```
pirat89: Actually, that's not `OS_FACTS` but `SANITY`. See other actors that are checking stuff related to HW. E.g. `checkcpu`. If you use something different here, you should consider to change it in other actors as well to keep information provided across the repository consistent.

repos/system_upgrade/el7toel8/actors/checkminimumrequirements/libraries/library.py
```diff
@@ -0,0 +1,89 @@
...
+    _get_kernel_version(hardware_info)
+    _get_memory_info(hardware_info)
+
+    api.produce(hardware_info)
+
+    _check_kernel_version(minimum_req_error, hardware_info)
+    _check_memory(minimum_req_error, hardware_info)
+
+    if minimum_req_error:
+        title = ('Minimum hardware requirements for RHEL 8 not detected. Upgrade cannot proceed.')
```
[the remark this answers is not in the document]
pirat89: Keep in mind we should not use "." (period) in the title.

repos/system_upgrade/el7toel8/models/hardwareinfo.py
```diff
@@ -0,0 +1,23 @@
...
+    The model currently doesn't represent all information about hardware could
+    provide on the machine. Just part of them, in case any other attributes
+    will be needed, the model can be extended.
+    """
+
+    topic = SystemFactsTopic
+
+    mem_total = fields.Nullable(fields.Integer())
+    """
+    Total memory in kib
```
pirat89: Do you mean `KiB`?

repos/system_upgrade/el7toel8/actors/checkminimumrequirements/libraries/library.py
```diff
@@ -0,0 +1,89 @@
...
+            'Total Memory required: {}    found: {}'.format(min_req_data['memory'][architecture.ARCH_ARM64], mem_info))
+    if (architecture.matches_architecture(architecture.ARCH_PPC64LE) and
+            mem_info.mem_total < min_req_data['memory'][architecture.ARCH_PPC64LE]):
+        minimum_req_error.append(
+            'Total Memory required: {}    found: {}'.format(min_req_data['memory'][architecture.ARCH_PPC64LE],
+                                                            mem_info))
+    if (architecture.matches_architecture(architecture.ARCH_S390X) and
+            mem_info.mem_total < min_req_data['memory'][architecture.ARCH_S390X]):
+        minimum_req_error.append(
+            'Total Memory required: {}    found: {}'.format(min_req_data['memory'][architecture.ARCH_S390X], mem_info))
```
pirat89: It doesn't make sense to produce the message here. Why? Only what I want is information what is required, what is discovered (tuple with two values). The information can be represented somewhere else. Just on the one line of the code. Now I see several duplicit lines for which I do not see any reason.

I think much better solution seems to return the tuple if requirements have not been met, or None if everything is ok - I am not telling that even this is ideal, but I would rather now what's wrong and work with it somehow (see the related sources comment) than just have a msg.

repos/system_upgrade/el7toel8/models/hardwareinfo.py
```diff
@@ -0,0 +1,23 @@
...
+    """
+    kernel_version = fields.Nullable(fields.String())
+    """
+    Kernel version
+    """
```
pirat89: How the version of the kernel is supposed to represent HW information? As well - the oldest kernel provided for RHEL 7 has version 3.10. Why to test this in that case? Additionaly it's duplicit information. As well - description doesnt fit. Information about CPU is information about hardware and it is not part of this model, but different one.

repos/system_upgrade/el7toel8/actors/checkminimumrequirements/libraries/library.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def check_minimum_hardware():
+    hardware_info = HardwareInfo()
+    minimum_req_error = []
+
+    _get_kernel_version(hardware_info)
```
pirat89: see my comment about kernel version.

repos/system_upgrade/el7toel8/actors/checkminimumrequirements/actor.py
```diff
@@ -0,0 +1,22 @@
...
+    These actor will check the system against RHEL8 minimal hardware requirements,
+    using the following resource: https://access.redhat.com/articles/rhel-limits
+
+    Note: Disk space will not be checked with these actor, rely on dnf actor to
+    check it.
+    """
+
+    name = 'checkminimumrequirements'
+    consumes = (HardwareInfo, )
+    produces = (Report, HardwareInfo)
```
pirat89: How you can consume and provide msg in one actor? That's wrong. Where is actor scanning the machine?

repos/system_upgrade/el7toel8/actors/checkminimumrequirements/libraries/library.py
```diff
@@ -0,0 +1,89 @@
...
+        minimum_req_error.append('Minimu kernel version: {}    found: {}'.format(
+            min_req_data['kernel_version'], hardware_info.kernel_version))
+
+
+def _get_memory_info(hardware_info):
+    mem_info = dict((i.split()[0].rstrip(':'), int(i.split()[1])) for i in open('/proc/meminfo').readlines())
+    total_mem_kib = mem_info['MemTotal']
+
+    hardware_info.mem_total = total_mem_kib
+
```
pirat89: obiously these functions cannot be part of this actor. This is `check` actor. That is checking something. It is not scanning the system.  Check actors are supposed to decide / check something based on consumed data.

repos/system_upgrade/el7toel8/actors/checkminimumrequirements/libraries/library.py
pirat89: btw, lists of possible / existing resources, flags, tags, ... are here: https://leapp.readthedocs.io/en/latest/actor-rhel7-to-rhel8.html#reporting-tips . about adding new stuff, sync with @shaded-enmity

repos/system_upgrade/el7toel8/actors/checkminimumrequirements/libraries/library.py
fernflower: Besides what Petr said, _check_memory could use some refactor (deduplication and you can have just 1 arch, right, why check for all 4?)
```
def _check_memory(mem_info):
"""
Returns (True, '') if memory check passed or (False, error_msg) if it didn't 
"""
    for arch, min_req in min_req_data['memory'].iteritems():
        if architecture.matches_architecture(arch):
            is_ok = mem_info.mem_total >= min_req:
            msg = '' if is_ok else 'Total memory required: {} found: {}'.format(min_req, mem_info.mem_total)
            return (is_ok, msg)
```

repos/system_upgrade/el7toel8/actors/checkminimumrequirements/libraries/library.py
```diff
@@ -0,0 +1,89 @@
...
+}
+
+
+def _get_kernel_version(hardware_info):
+    hardware_info.kernel_version = re.findall(r'\d+\.\d+\.\d+', os.uname()[2])[0]
+
+
+def _check_kernel_version(minimum_req_error, hardware_info):
+    kernel_ver_info = tuple(map(int, hardware_info.kernel_version.split('.')))
+    if kernel_ver_info < (3, 10, 0):
```
fernflower: Please don't use magic numbers inside functions, minimal kernel version should be taken from some visible constant (min_req_data['kernel_version']?)

repos/system_upgrade/el7toel8/actors/checkminimumrequirements/libraries/library.py
fernflower: Or just create a PR for adding a necessary resource type :)

repos/system_upgrade/el7toel8/actors/checkminimumrequirements/libraries/library.py
```diff
@@ -0,0 +1,89 @@
...
+
+    _check_kernel_version(minimum_req_error, hardware_info)
+    _check_memory(minimum_req_error, hardware_info)
+
+    if minimum_req_error:
+        title = ('Minimum hardware requirements for RHEL 8 not detected. Upgrade cannot proceed.')
+        URL = ('https://access.redhat.com/articles/rhel-limits')
+        summary = ('The following hardware requirements is not detected:'
+                   '\n{}.'.format('\n     - '.join(minimum_req_error)))
+        remediation = ('Please check your hardware and provide upgrade, see: ' + URL)
```
fernflower: Please use string formatting over concatenation

repos/system_upgrade/el7toel8/actors/checkminimumrequirements/libraries/library.py
```diff
@@ -0,0 +1,89 @@
...
+                          reporting.Summary(summary),
+                          reporting.Severity(reporting.Severity.HIGH),
+                          reporting.Tags([reporting.Tags.OS_FACTS]),
+                          reporting.Flags([reporting.Flags.INHIBITOR]),
+                          reporting.Remediation(hint=remediation)
+                      ] + [reporting.RelatedResource('minimum_hardware_requirements', i) for i in minimum_req_error])
+
+
+def process():
+    return check_minimum_hardware()
```
fernflower: iiuc process function should not return anything

repos/system_upgrade/el7toel8/actors/checkmemory/tests/test_checkmemory.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def mocked_get_memoryinfo(filename):
+    """
+    Return lines of the self.filename test file located in the files directory.
+
+    Those files contain /proc/meminfo content from machine.
+    """
+    memory_info = MemoryInfo()
+    mem_info = dict((i.split()[0].rstrip(':'), int(i.split()[1])) for i in open(os.path.join('tests/files', filename), 'rb').readlines())  # noqa: E501; pylint: disable=line-too-long
```
fernflower: You should have a very good reason to skip linter checks. Why disable=line-too-long should be chosen over code refactor?

fernflower: Please use open with context manager.

repos/system_upgrade/el7toel8/actors/scanmemory/libraries/library.py
```diff
@@ -0,0 +1,22 @@
+from leapp.libraries.stdlib import api
+from leapp.models import MemoryInfo
+
+
+def _get_memoryinfo():
+    """Return lines from /proc/meminfo."""
+    # Expecting the file exists on earch system, skipping any check
+    try:
+        return dict((i.split()[0].rstrip(':'), int(i.split()[1])) for i in open('/proc/meminfo').readlines())
```
fernflower: Please use context manager with open. Also exception handling should be in place - what if data isn't parsed the way you expected? Traceback in the framework is not sexy

repos/system_upgrade/el7toel8/actors/checkmemory/libraries/library.py
```diff
@@ -0,0 +1,44 @@
+from leapp.models import MemoryInfo
+from leapp.libraries.common.config import architecture
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.stdlib import api
+from leapp.reporting import create_report
+from leapp import reporting
```
pirat89: This was the issue. You are importing the reporting library and the create_report function as well, but you are mocking just the function inside the reporting library. Not the one imported in the tested library.

repos/system_upgrade/el7toel8/actors/scanmemory/libraries/library.py
```diff
@@ -0,0 +1,22 @@
+from leapp.libraries.stdlib import api
+from leapp.models import MemoryInfo
+
+
+def _get_memoryinfo():
+    """Return lines from /proc/meminfo."""
```
fernflower: Please specify return data type in the docstring. Now you have either a dict or None, shouldn't you unify that?

repos/system_upgrade/el7toel8/actors/checkmemory/libraries/library.py
pirat89: @fellipeh that's the part of the code I have been talking about.

repos/system_upgrade/el7toel8/actors/scanmemory/tests/test_scanmemory.py
```diff
@@ -0,0 +1,38 @@
...
+    def __init__(self, filename):
+        self.filename = filename
+
+    def __call__(self):
+        """
+        Return lines of the self.filename test file located in the files directory.
+
+        Those files contain /proc/meminfo content from machine.
+        """
+        return dict((i.split()[0].rstrip(':'), int(i.split()[1])) for i in open(os.path.join('tests/files', self.filename), 'rb').readlines())  # noqa: E501; pylint: disable=line-too-long
```
fernflower: Please see comment above. 

Another suggestion will be to move code to a library and reuse in both tests.

repos/system_upgrade/el7toel8/models/memoryinfo.py
```diff
@@ -0,0 +1,19 @@
+from leapp.models import Model, fields
+from leapp.topics import SystemFactsTopic
+
+
+class MemoryInfo(Model):
+    """
+    The model represents information about Memory
+
+    The model currently doesn't represent all information about memory could
```
fernflower: I'd keep it simple and leave just "Represents information about available memory (KiB)"

All information about "it's not there yet but can be" is better suited for a comment, not a docstring

repos/system_upgrade/el7toel8/actors/checkmemory/libraries/library.py
```diff
@@ -0,0 +1,44 @@
...
+def process():
+    memoryinfo = next(api.consume(MemoryInfo), None)
+    if memoryinfo is None:
+        raise StopActorExecutionError(message=("Missing information about Memory."))
+
+    minimum_req_error = _check_memory(memoryinfo)
+
+    if minimum_req_error:
+        title = 'Minimum memory requirements for RHEL 8 are not met'
+        summary = 'Memory detected: {} KiB, required: {} KiB'.format(minimum_req_error['memory'][0],
```
fernflower: Maybe change the format of _check_memory output then? A dict {'minimal_req': X, 'detected': Y} is more obvious to use with format then {'memory': [X, Y]}

repos/system_upgrade/el7toel8/actors/checkmemory/tests/test_checkmemory.py
pirat89: fixed.

repos/system_upgrade/el7toel8/actors/scanmemory/libraries/library.py
pirat89: fixed - btw, I removed the try as regarding the comment it didn't make sense - unless the comment is wrong.

repos/system_upgrade/el7toel8/actors/scanmemory/tests/test_scanmemory.py
pirat89: fixed.

repos/system_upgrade/el7toel8/actors/checkmemory/tests/test_checkmemory.py
pirat89: fixed.

repos/system_upgrade/el7toel8/actors/checkmemory/tests/test_checkmemory.py
```diff
@@ -0,0 +1,59 @@
...
+    if mem_info:
+        memory_info.mem_total = mem_info['MemTotal']
+
+    return memory_info
```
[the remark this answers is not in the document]
fernflower: +1, I was probably too frustrated by linter check skipping and missed that

repos/system_upgrade/el7toel8/actors/checkmemory/libraries/library.py
fellipeh: Ok @pirat89 ! Now I understand! :)   Thanks!

repos/system_upgrade/el7toel8/actors/checkmemory/actor.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+class CheckMemory(Actor):
+    """
+    These actor will check the system against RHEL8 minimal hardware requirements,
+    using the following resource: https://access.redhat.com/articles/rhel-limits
+
+    Note: Disk space will not be checked with these actor, rely on dnf actor to
+    check it.
+    """
```
pirat89: As well here. Keep format:
```
Oneline title

Description on multiple lines
```

repos/system_upgrade/el7toel8/actors/checkmemory/actor.py
```diff
@@ -0,0 +1,22 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import library
+from leapp.models import MemoryInfo, Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckMemory(Actor):
+    """
+    These actor will check the system against RHEL8 minimal hardware requirements,
```
pirat89:
```suggestion
    The actor checks the size of RAM against RHEL8 minimal hardware requirements
```

repos/system_upgrade/el7toel8/actors/scanmemory/tests/test_scanmemory.py
```diff
@@ -0,0 +1,39 @@
...
+
+from leapp.libraries.actor import library
+from leapp.libraries.common import testutils
+from leapp.libraries.stdlib import api
+from leapp.models import MemoryInfo
+
+
+class mocked_get_memoryinfo(object):
+    def __init__(self, filename):
+        self.filename = os.path.join('tests', 'files', filename)
```
fernflower: As @pirat89 already mentioned you don't need to create a test file and then parse it to test proper info is produced.

If you consider going with current approach (testing based on files) you should not mock library._get_memoryinfo but rather mock the open function to feed it test file as /proc/meminfo

repos/system_upgrade/el7toel8/actors/scanmemory/tests/test_scanmemory.py
```diff
@@ -0,0 +1,39 @@
...
+        """
+        Return lines of the self.filename test file located in the files directory.
+
+        Those files contain /proc/meminfo content from various machines.
+        """
+        return self._orig_get_memoryinfo(self.filename)
+
+
+def test_low_memory(monkeypatch):
+    mocked_memoryinfo = mocked_get_memoryinfo('meminfo_low')
```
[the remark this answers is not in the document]
pirat89: Actually, that is on many places in other actors as well, so I am ok with that. But yes, it could be eg. parametrized using decorator, which would do the whole thing without need of two almost same functions.

repos/system_upgrade/el7toel8/models/memoryinfo.py
```diff
@@ -0,0 +1,15 @@
+from leapp.models import Model, fields
+from leapp.topics import SystemFactsTopic
+
+
+class MemoryInfo(Model):
+    """
+        Represents information about available memory (KiB)
```
pirat89: Again, indentation.
```suggestion
    Represents information about available memory (KiB)
```

repos/system_upgrade/el7toel8/actors/scanmemory/tests/test_scanmemory.py
pirat89: @fernflower It's small misunderstanding here. My note was related to tests of the `checkmemory` actor, which is not working with files at all and it is just consuming information provided by the scanmemory actor. In this case, I am ok with test files.

fernflower: @pirat89 Still if test files are to be kept in place the mocking of the whole _get_memory_info function is wrong

repos/system_upgrade/el7toel8/actors/scanmemory/libraries/library.py
```diff
@@ -0,0 +1,21 @@
...
+    """ Returns dict of all memory information from /proc/meminfo file """
+
+    # Expecting the file exists on earch system
+    try:
+        with open(filename, 'r') as fp:
+            # format of the lines: Key:  value unit
+            # e.g.:                MemTotal:   1024 kB
+            return dict((i.split()[0].rstrip(':'), int(i.split()[1])) for i in fp.readlines())
+    except IOError:
+        return None
```
pirat89: This is still not resolving the original concerns we reported during previous reviews. Now in case it will fail, we will not see anything and will have to spend extra time with investigation what's the problem. If you have any questions, ping me or @fernflower to resolve them.

repos/system_upgrade/el7toel8/actors/checkmemory/libraries/library.py
```diff
@@ -0,0 +1,44 @@
...
+            msg = {} if is_ok else {'detected': mem_info.mem_total,
+                                    'minimal_req': min_req}
+
+    return msg
+
+
+def process():
+    memoryinfo = next(api.consume(MemoryInfo), None)
+    if memoryinfo is None:
+        raise StopActorExecutionError(message=("Missing information about Memory."))
```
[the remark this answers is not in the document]
pirat89: from my POV, it's ok as it is here. BUT it is just info that is missing the expected message. The problem we reported earlier was in scanner actor. So the error should be raised in scanner actor. So the check actor will not be even processed and the error here is kind of redundant in that case (but can be kept here. just with the point that `Memory`/`memory size`). So still waiting for fix of the part reported earlier.

fellipeh:
Thanks @fernflower ! Now I understand your point, changing the code here with your mock code (it's more, more clear them mine!)

Thanks again!

repos/system_upgrade/el7toel8/actors/scanmemory/libraries/library.py
```diff
@@ -0,0 +1,24 @@
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.stdlib import api
+from leapp.models import MemoryInfo
+
+
+def _get_memoryinfo(filename='/proc/meminfo'):
+    """ Returns dict of all memory information from /proc/meminfo file """
+
+    try:
+        with open(filename, 'r') as fp:
```
fernflower: 'r' mode is redundant

repos/system_upgrade/el7toel8/actors/scanmemory/tests/files/meminfo_high
```diff
@@ -0,0 +1,51 @@
+MemTotal:       16273492 kB
```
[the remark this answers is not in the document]
fernflower: No it is not resolved, the test files are still there

repos/system_upgrade/el7toel8/actors/scanmemory/tests/files/meminfo_low
```diff
@@ -0,0 +1,51 @@
+MemTotal:           1024 kB
```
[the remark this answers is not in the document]
fernflower: Not solved! Please do not resolve conversation until you have fixed everything, it is really hard to track progress

repos/system_upgrade/el7toel8/actors/scanmemory/tests/test_scanmemory.py
```diff
@@ -0,0 +1,24 @@
+import mock
+
+from leapp.libraries.actor import library
+from leapp.libraries.common import testutils
+from leapp.libraries.stdlib import api
+from leapp.models import MemoryInfo
+
+
+def test_with_low_memory(monkeypatch):
+    with mock.patch("__builtin__.open", mock.mock_open(read_data="MemTotal: 42 kB")) as mock_proc_meminfo:
```
fernflower: Open question: why are you checking for MemTotal?

fellipeh: if I understood correctly your question, the goal for these actor is to check the total memory of the machine.. that's the reason I'm checking that.

fernflower: Why are you relying on MemTotal, not MemFree or MemAvailable?

pirat89: @fernflower Because it is HW requirement check. So you are not interested about info how much free or available memory you have.

fernflower: @pirat89 ah, ok, that makes sense then. Thanks!

repos/system_upgrade/el7toel8/actors/scanmemory/libraries/library.py
```diff
@@ -0,0 +1,24 @@
...
+
+    try:
+        with open(filename) as fp:
+            # format of the lines: Key:  value unit
+            # e.g.:                MemTotal:   1024 kB
+            return dict((i.split()[0].rstrip(':'), int(i.split()[1])) for i in fp.readlines())
+    except IOError:
+        raise StopActorExecutionError(
+            'Could not read memory values',
+            details={'details': 'Error open file {}'.format(filename)},
```
fernflower: Nit: "Error opening file"

repos/system_upgrade/el7toel8/actors/scanmemory/tests/test_scanmemory.py
fellipeh: Reading the document (https://access.redhat.com/articles/rhel-limits), I understood minimum required memory as total memory on the machine to rhel8 run, not Available/Free, because it's depends what's is running at the time. 

But is possible I miss understood that...

fernflower: @fellipeh no that's fine, that was only a question. Figured out if I have it others may wonder as well - so won't hurt to clear that once and for all.

fernflower:
leapp-ci build

PR #415, opened 2019-11-28
Grubby debug was failing on s390x

Rezney:
For some reason the sed command was not working on s390x.
Switching to python.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/grubbydebug/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/grubbydebug/actor.py
@@ -1,7 +1,13 @@
+import os
+
 from leapp.actors import Actor
 from leapp.models import TransactionCompleted
 from leapp.tags import RPMUpgradePhaseTag, IPUWorkflowTag, ExperimentalTag
-from leapp.libraries.stdlib import api, run, CalledProcessError
+from leapp.libraries.stdlib import api
+
+
+GRUBBY_FILES = ['/usr/libexec/grubby/grubby-bls', '/usr/libexec/grubby/grubby']
+DEBUG_FLAG = 'set -x\n'
 
 
 class GrubbyDebug(Actor):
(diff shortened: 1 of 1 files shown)


Rezney:
Have to test this first...

AloisMahdal:
e2e tests

AloisMahdal:
e2e tests

Rezney:
Seems to be working fine as @pirat89 witnessed :) @AloisMahdal can we get a clean run here?

AloisMahdal:
> For some reason..

FYI, the reason was that `/usr/bin/grubby` did not exist.

Rezney:
e2e tests

Rezney:
@AloisMahdal traceback or some snippet might be useful...

Rezney:
@AloisMahdal test "murphy-ci/upgrade Pending — lpr=R415" has been in pending state for more than 5 hours...

AloisMahdal:
The reason it was 5h running was because it was in emergency mode due to a known issue: current leapp-repository `U`pstream master is not already compatible with leapp's latest release (`B`rew build, downstream), which is what Murphy CI picks up by default.  (We're going to discuss this choice later.)

Nevertheless,I've re-ran the whole thing manually with leapp `U`pstream master build but I got strange traceback from RHSM; reporting downstream.

As an aside, because the earlier result took over 8h (we're generous with timeout) it overrode the later one which ran only about 1-2h.  Since both results are invalid this almost does not matter now; it's just why the status description is different (`lpr=R415` vs `lpr=R415,lp=U`).

For now I have re-ran it again, hoping that the RHSM issue was an intermittent one.

AloisMahdal:
Now upgrade batches passed but integration+static failed due to Fedora COPR outage.

(I don't know how this works.  It's magic. Apparently calling our CI Murphy was way more fitting than we wanted; this PR is the first one where we're reporting back to GH and it's all on fire. (@vinzenz this was inspired by your remark so explain yourself you sorcerer! :-D))

Note that at this point we saw pass in every batch, with same build set, but I am going to re-run it again now so that if we're lucky in the morning Rezney could see a clean run.

PR #417, opened 2019-12-03
Yet another grub debugging

Rezney:
Unfortunately, so far we were not able to find out what the issue
with booting into old el7 kernel is.

Adding new *temporary* debug options

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/grubbydebug/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/grubbydebug/actor.py
@@ -3,7 +3,7 @@ import os
 from leapp.actors import Actor
 from leapp.models import TransactionCompleted
 from leapp.tags import RPMUpgradePhaseTag, IPUWorkflowTag, ExperimentalTag
-from leapp.libraries.stdlib import api
+from leapp.libraries.stdlib import api, run, CalledProcessError
 
 
 GRUBBY_FILES = ['/usr/libexec/grubby/grubby-bls', '/usr/libexec/grubby/grubby']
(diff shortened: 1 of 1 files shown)


Rezney:
e2e tests

Rezney:
e2e tests

repos/system_upgrade/el7toel8/actors/grubbydebug/actor.py
```diff
@@ -45,3 +45,19 @@ def enable_debug(filename):
...
                     api.current_logger().warning(
                         'Failed to activate debug mode in {}'.format(file_), exc_info=True)
+
+        try:
+            run(['ls', '-la', '/boot'])
+        except CalledProcessError:
+            api.current_logger().warning(
+                'Could not list /boot dir', exc_info=True)
+        try:
+            run(['cat', '/boot/loader/entries/*.conf'])
```
AloisMahdal: The asterisk won't expand (...aand we got `Traceback` in tests.).

```suggestion
            run(['sh', '-c', 'cat /boot/loader/entries/*.conf'])
```

Rezney: Yep, forgot about this :) 

Long live the PR e2e tests \o/

Rezney: Anyway, it is nice that the word "Traceback" is nicely formatted with back-quotes but I would more appreciate seeing the traceback itself :) So I can cover it properly on top of the fixing the command itself...

Rezney:
e2e tests

Rezney:
e2e tests

Rezney:
e2e tests

Rezney:
e2e tests

Rezney:
e2e tests

Rezney:
e2e tests

AloisMahdal:
Quick status update:  with this actor, our test systems seem to have troubles rebooting, get stuck in initramfs and eventually time out.  Through trial and error we've figured out that the `grub2-editenv` call is the culprit: without it the reboot works fine.  We're still investigating it..

Rezney:
e2e tests

Rezney:
@pirat89 this is the last debugging attempt. Let's try if we get something out of it and then delete the experimental actor.

PR #419, opened 2019-12-11
Let's print also default kernel title and index

Rezney:


repos/system_upgrade/el7toel8/actors/forcedefaultboottotargetkernelversion/libraries/forcedefaultboot.py
```diff
@@ -76,6 +76,13 @@ def process():
         api.current_logger().warning('Failed to query grubby for default kernel', exc_info=True)
         return
 
+    for type_ in ('index', 'title'):
+        try:
+            stdlib.run(['grubby', '--default-{}'.format(type_)])['stdout'].strip()
```
pirat89:
```suggestion
            stdlib.run(['grubby', '--default-{}'.format(type_)])
```

PR #423, opened 2020-01-02
Introduce verifydialogs actor

fernflower:
Add an actor that collects information about dialogs
to answer during leapp answer stage.
Also this patch will contain minimal changes to existing
actors with dialogs (ntpd->chrony mostly).

Requires: https://github.com/oamg/leapp/pull/589

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/removeoldpammodulescheck/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/removeoldpammodulescheck/actor.py
@@ -28,7 +28,6 @@ class RemoveOldPAMModulesCheck(Actor):
                     key='confirm',
                     label='Disable pam_pkcs11 module in PAM configuration? '
                           'If no, the upgrade process will be interrupted.',
-                    default=False,
                     description='PAM module pam_pkcs11 is no longer available '
                                 'in RHEL-8 since it was replaced by SSSD.',
                     reason='Leaving this module in PAM configuration may '
@@ -44,7 +43,6 @@ class RemoveOldPAMModulesCheck(Actor):
                     key='confirm',
                     label='Disable pam_krb5 module in PAM configuration? '
                           'If no, the upgrade process will be interrupted.',
-                    default=False,
                     description='PAM module pam_krb5 is no longer available '
                                 'in RHEL-8 since it was replaced by SSSD.',
                     reason='Leaving this module in PAM configuration may '
@@ -63,7 +61,8 @@ class RemoveOldPAMModulesCheck(Actor):
             result = self.confirm(module)
             if result:
                 self.produce_report(module)
-            else:
+            elif result is False:
+                # user specifically chose to disagree with auto disablement
                 self.produce_inhibitor(module)
 
     def confirm(self, module):
@@ -72,7 +71,7 @@ class RemoveOldPAMModulesCheck(Actor):
             'pam_krb5': self.dialogs[1]
         }
 
-        return self.request_answers(questions[module]).get('confirm', False)
+        return self.get_answers(questions[module]).get('confirm')
 
     def produce_report(self, module):
         create_report([
--- a/repos/system_upgrade/el7toel8/actors/verifydialogs/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/verifydialogs/actor.py
@@ -1,0 +1,21 @@
+from leapp.actors import Actor
+from leapp.models import DialogModel
+from leapp.reporting import Report
+from leapp.libraries.actor.library import check_dialogs
+from leapp.tags import ReportPhaseTag, IPUWorkflowTag
+
+
+class VerifyDialogs(Actor):
+    """
+    Check all dialogs and notify that user needs to make some choices.
+
+    Report messages containing all dialogs with questions that need to be answered will be produced.
+    """
+
+    name = 'verify_check_results'
+    consumes = (DialogModel,)
+    produces = (Report,)
+    tags = (ReportPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        check_dialogs(inhibit_if_no_userchoice=self.skip_dialogs)
--- a/repos/system_upgrade/el7toel8/actors/verifydialogs/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/verifydialogs/libraries/library.py
@@ -1,0 +1,20 @@
+from leapp.libraries.stdlib import api
+from leapp.models import DialogModel
+from leapp import reporting
+
+
+def check_dialogs(inhibit_if_no_userchoice=True):
+    results = list(api.consume(DialogModel))
+    for dialog in results:
+        sections = dialog.answerfile_sections.split(',')
+        dialog_resources = [reporting.RelatedResource('dialog', s) for s in sections]
+        dialogs_remediation = ('Please register user choices with leapp answer cli command or by manually editing '
+                               'the answerfile.')
+        report_data = [reporting.Title('Dialog choice required'),
+                       reporting.Severity(reporting.Severity.HIGH),
+                       reporting.Summary(
+                           'One or more sections in answerfile are missing recorded user choices: {}'.format(
+                               '\n'.join(sections))),
+                       reporting.Flags([reporting.Flags.INHIBITOR] if inhibit_if_no_userchoice else []),
+                       reporting.Remediation(hint=dialogs_remediation)]
+        reporting.create_report(report_data + dialog_resources)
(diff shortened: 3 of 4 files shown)


repos/system_upgrade/el7toel8/actors/verifydialogs/libraries/library.py
```diff
@@ -0,0 +1,17 @@
...
+
+def check_dialogs(inhibit_if_no_userchoice=True):
+    results = list(api.consume(DialogModel))
+    for dialog in results:
+        sections = dialog.answerfile_sections.split(',')
+        dialog_resources = [reporting.RelatedResource('dialog', s) for s in sections]
+        report_data = [reporting.Title('Dialog choice required'),
+                       reporting.Severity(reporting.Severity.HIGH),
+                       reporting.Summary(
+                           'Several sections in answerfile are missing recorded user choices: {}'.format(
```
Rezney: I think the wording should be different here as we can have just one section there:

`Summary: Several sections in answerfile are missing recorded user choices: remove_pam_pkcs11_module_check.confirm`

What about: There is/are section(s) missing recorded user choices:

repos/system_upgrade/el7toel8/actors/verifydialogs/actor.py
```diff
@@ -0,0 +1,21 @@
...
+    Report messages containing all dialogs with questions that need to be answered will be produced.
+    """
+
+    name = 'verify_check_results'
+    consumes = (DialogModel,)
+    produces = (Report,)
+    tags = (ReportPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        check_dialogs(inhibit_if_no_userchoice=not self.skip_dialogs)
```
Rezney: Currently, when trying this on the test VM and running `leapp preupgrade` it finishes successfully with 0 exit code. Later, when running `leapp upgrade`, it fails informing users they did not provide any answer.

This probably contrast with what we wanted `preupgrade` to be, to communicate if the command passes, upgrade itself should be ok. What do you think @pirat89 ?

repos/system_upgrade/el7toel8/actors/removeoldpammodulescheck/actor.py
```diff
@@ -44,7 +43,6 @@ class RemoveOldPAMModulesCheck(Actor):
                     key='confirm',
                     label='Disable pam_krb5 module in PAM configuration? '
                           'If no, the upgrade process will be interrupted.',
-                    default=False,
```
Rezney: Could you please elaborate on removing these "defaults"?

fernflower: There should be no defaults in the dialogs that require serious user decision. If we could have a default here - why have a dialog asking the user to take responsibility?

repos/system_upgrade/el7toel8/actors/verifydialogs/actor.py
fernflower: We can't have interactivity during preupgrade stage. I can easily add an inhibitor and this will result in "smth went wrong" every time a preupgrade is run with the dialogs (literally every run on openstack rhel 7.6-updated image with authselect actors enabled). 
I decided not to stress out the user and go without an inhibitor for now. But it can be changed ofc.

repos/system_upgrade/el7toel8/actors/removeoldpammodulescheck/actor.py
Rezney: Ok, fair enough. In the past there was a discussion that we may have something like "--default" to continue with upgrade with suggested choices but the idea was most probably completely abandoned...

repos/system_upgrade/el7toel8/actors/verifydialogs/actor.py
Rezney: This is not much about interactivity I think. We were experiencing the same issue with one of our SSH actors where preupgrade always failed without explicit "PermitRootLogin yes".

I will have to take a look at the dialog again but it should not be difficult to add some workaround in our setup scripts (e.g vagrant, OpenStack), right?

pirat89: @fernflower As you can see, you already have there the ssh probably on major number of systems. What @Rezney is asking, it is not interactivity. The inhibitor should be raised definitely as in case that preupgrade ends with "green", you expect you can do an upgrade. but based on missing answers, it's not true. As well, when you answer everything, it can happen that the process will be inhibited based on anwer. We can prepare for people "elementary" answerfile (iternally only) and they could use it to skip this. Or we could put there default answers if this is really big pain. But the `preupgrade` must tell you already this information. There is not green with toxic waste hidden behind fence.

repos/system_upgrade/el7toel8/actors/removeoldpammodulescheck/actor.py
pirat89: I thought that default will not be used by default,but only when user applies specific option like `--use-default`. Anyway, still it could leat into the situation that people will not read it at all, so removal of default is probably really better from that point.

repos/system_upgrade/el7toel8/actors/verifydialogs/actor.py
pirat89: @fernflower  But looking into the code, it looks that you wrote it well. It's exactly what I would expect, so I am now confused about the discussion at all (O_o). Did I miss something?

fernflower: I get your point, guys, and having slept on this I agree that you are right about the preupgrade idea of showing green report only if you are good to go with the upgrade.
I'll fix that in next commit.

Rezney: @fernflower thanks
@pirat89 Good idea with pre-crafted internal answerfile

repos/system_upgrade/el7toel8/actors/verifydialogs/libraries/library.py
fernflower: Should be resolved in new patch

repos/system_upgrade/el7toel8/actors/verifydialogs/actor.py
fernflower: Done

repos/system_upgrade/el7toel8/actors/removeoldpammodulescheck/actor.py
fernflower: Resolving for now as per discussion with original actor writer having a default=True is a no go here and having default=False won't let upgrade to pass.

repos/system_upgrade/el7toel8/actors/verifydialogs/libraries/library.py
```diff
@@ -0,0 +1,20 @@
...
+from leapp import reporting
+
+
+def check_dialogs(inhibit_if_no_userchoice=True):
+    results = list(api.consume(DialogModel))
+    for dialog in results:
+        sections = dialog.answerfile_sections.split(',')
+        dialog_resources = [reporting.RelatedResource('dialog', s) for s in sections]
+        dialogs_remediation = ('Please register user choices with leapp answer cli command or by manually editing '
+                               'the answerfile.')
```
Rezney: I know the "answerfile" location is quite visible at the end of the log but I would probably mention it in the report remediation hint too. 

_Please register user choices with leapp answer cli command or by manually editing the answerfile in ..._

fernflower: Double checked and for now there is no nice way to retrieve answerfile location from the actor. I don't want to hardcode a magic constant (which is likely to change in the future as per discussion with @pirat89 I agree that /var/log/leapp/answerfile is not a proper way to have an answerfile anyway) and as long as answerfile location is fairly visible I'd leave that change for the next patch.

repos/system_upgrade/el7toel8/actors/verifydialogs/actor.py
pirat89: Review in progess.

Rezney:
@fernflower I would suggest to improve instructions in the answerfile. Type: bool may not be very descriptive for less experienced users.

I uncommented "confirm = yes" and the workflow changed it to "confirm = False" and told me that I do not want to proceed :)

"confirm = True" did the job...

fernflower:
@Rezney this will be improved as a follow-up patch in leapp.

Rezney:
LGTM
I think we should be good to go with this one and framwork PR589. What do you think @pirat89 ?

Rezney:
leapp-ci build

PR #424, opened 2020-01-03
Add tests for OpenSshUsePrivilegeSeparationCheck

drehak:
Processing code is split off to a library so that the exception raise could be tested.
Also documented the related `OpenSshConfig` model.


repos/system_upgrade/el7toel8/models/opensshconfig.py
```diff
@@ -20,12 +22,17 @@ class OpenSshConfig(Model):
     """
     topic = SystemInfoTopic
 
+    # All PermitRootLogin directives.
```
pirat89: please. document it using docstring, - below attributes.

pirat89: see e.g. cpuinfo.py

PR #425, opened 2020-01-06
Makefile: specify chroot by COPR_CHROOT, optional

pirat89:
In some cases we need to build rpms just for specific chroot. E.g.
to make building faster as our rpms are noarch (so build just for
one arch...). Or in case of public COPR, build leapp-repository just
for epel, as there is no repository for Fedora yet.

Possible use:
```
  COPR_CHROOT=epel-7-x86_64 make copr_build
```

Note: this is required to enable building of rpms in public COPR.


pirat89:
The first public build of leapp-repository in the expected public repository is here (nice sentence, :-) ):
 https://copr.fedorainfracloud.org/coprs/g/oamg/leapp/build/1139443/

PR #430, opened 2020-01-09
[POC] RpmScanner: Retry yum call with a fallback locale if it fails

drehak:
Needs a refactor before merging, I'm just trying to wrap the `yum_base.doPackageLists()` call at the moment.


drehak:
leapp-ci build

repos/system_upgrade/el7toel8/actors/rpmscanner/libraries/library.py
```diff
@@ -6,7 +8,18 @@ def get_package_repository_data():
...
+            pkg_repos[pkg.name] = pkg.ui_from_repo.lstrip('@')
+    except ValueError as e:
+        if 'locale' not in str(e):  # reraise if error is not related to locales
+            raise e
+        raise StopActorExecutionError(
+            message='Failed to get installed RPM packages because of an invalid locale',
+            details={
+                'hint': 'Please run leapp with a valid locale. ' +
+                        'You can get a list of installed locales by running `locale -a`.'
+            })
```
pirat89: @fernflower Just curious, how we show these errors to people? I mean, in cockpit. We show it, right?

As well, .. is there anything like `hint`?.. I guess this is just for remediation instruction, right?

fernflower: @shaded-enmity Can you give more details on cockpit reporting?

fernflower: @pirat89 if I got your question right, it is "How do we report workflow errors in the leapp report now?". If that's the case then sadly this won't be reflected in the report at all. I had a framework pr for adding errors to the report https://github.com/oamg/leapp/pull/576

drehak: (There *is* a hint. Discussed in person.)

pirat89: @fernflower thanks for info. We should probably drive that one PR.

PR #432, opened 2020-01-14
pesevenstanner: sort output lists to be consistent

pirat89:
From time to time, unit-tests could randomly fail because the output
list of packages (when log or report msg with list of rpms) was
unsorted. Sort all lists, so output will be more consistent and
readable.

Note: this resolves just one particular actor. Others can be fixed
in another PR(s).


shaded-enmity:
Note that in Python3 `sorted` returns a generator, I'd go with `def sortedl(i): return list(sorted(i))` to prevent future incompatibilities

pirat89:
> Note that in Python3 `sorted` returns a generator, I'd go with `def sortedl(i): return list(sorted(i))` to prevent future incompatibilities

Thanks for info. I missed that when I tried that with Python3 in interactive mode. Yes, we could put function like that in our utils library, as I believe it can be useful on many places.

PR #433, opened 2020-01-15
Update GRUB core on legacy (BIOS) systems.

Rezney:
On legacy (BIOS) systems, GRUB core (located in the gap between the MBR and the
first partition), does not get automatically updated when GRUB is upgraded.

This actor also helps to mitigate an issue with randomly booting into
old RHEL7 kernel.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkgrubcore/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkgrubcore/actor.py
@@ -1,0 +1,53 @@
+from leapp.actors import Actor
+from leapp.libraries.common.config import architecture
+from leapp.models import FirmwareFacts, GrubDevice, UpdateGrub
+from leapp.reporting import Report, create_report
+from leapp import reporting
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+GRUB_SUMMARY = ('On legacy (BIOS) systems, GRUB core (located in the gap between the MBR and the '
+                'first partition) does not get automatically updated when GRUB is upgraded.')
+
+
+class CheckGrubCore(Actor):
+    """
+    Check whether we are on legacy (BIOS) system and instruct Leapp to upgrade GRUB core
+    """
+
+    name = 'check_grub_core'
+    consumes = (FirmwareFacts, GrubDevice)
+    produces = (Report, UpdateGrub)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        if architecture.matches_architecture(architecture.ARCH_S390X):
+            # s390x archs use ZIPL instead of GRUB
+            return
+
+        ff = next(self.consume(FirmwareFacts), None)
+        if ff and ff.firmware == 'bios':
+            dev = next(self.consume(GrubDevice), None)
+            if dev:
+                self.produce(UpdateGrub(grub_device=dev.grub_device))
+                create_report([
+                    reporting.Title(
+                        'GRUB core on {} will be updated during upgrade'.format(dev.grub_device)
+                    ),
+                    reporting.Summary(GRUB_SUMMARY),
+                    reporting.Severity(reporting.Severity.HIGH),
+                    reporting.Tags([reporting.Tags.BOOT]),
+                ])
+            else:
+                create_report([
+                    reporting.Title('Leapp could not identify where GRUB core is located'),
+                    reporting.Summary(
+                        'We assume GRUB core is located on the same device as /boot. Leapp needs to '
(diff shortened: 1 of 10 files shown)


repos/system_upgrade/el7toel8/actors/updategrubcore/libraries/library.py
```diff
@@ -0,0 +1,83 @@
...
+from leapp.exceptions import StopActorExecutionError
+from leapp import reporting
+
+
+def has_grub(blk_dev):
+    """
+    Check whether GRUB is present on block device
+    """
+    try:
+        run(['dd', 'if={}'.format(blk_dev), 'of=/tmp/mbr', 'bs=1', 'count=512'])
```
vinzenz: nitpick: you would want bs=512 & count=1 because you do 512 reads otherwise. Where as bs=512 gives a block size of 512

repos/system_upgrade/el7toel8/actors/updategrubcore/actor.py
```diff
@@ -0,0 +1,24 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.library import update_grub_core
+from leapp.libraries import stdlib
```
vinzenz:
```suggestion
```

repos/system_upgrade/el7toel8/actors/updategrubcore/libraries/library.py
Rezney: aah, what have I done? Of course I wanted `bs=512 & count=1`. Not sure how my mind came up with this :)

Rezney:
Will have to refactor the tests...

Rezney:
@vinzenz I would still like to test it more but overall it should be ready for review...

repos/system_upgrade/el7toel8/actors/checkgrubcore/actor.py
```diff
@@ -0,0 +1,47 @@
...
+                    create_report([
+                        reporting.Title(
+                            'GRUB core on {} will be updated during upgrade'.format(dev.grub_device)
+                        ),
+                        reporting.Summary(GRUB_SUMMARY),
+                        reporting.Severity(reporting.Severity.INFO),
+                        reporting.Tags([reporting.Tags.BOOT]),
+                    ])
+                    self.produce(UpdateGrub(grub_device=dev.grub_device))
+                    return
```
fernflower: Looking at that return - so we are processing the first GrubDevice and skipping the rest?

repos/system_upgrade/el7toel8/actors/updategrubcore/actor.py
```diff
@@ -0,0 +1,22 @@
...
+
+    name = 'update_grub_core'
+    consumes = (TransactionCompleted, UpdateGrub)
+    produces = (Report,)
+    tags = (RPMUpgradePhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        for dev in self.consume(UpdateGrub):
+            update_grub_core(dev.grub_device)
+            return
```
fernflower: Same question about the return as above - if the idea is to process the first dev, why can't we have it like 
```
dev = next(self.consume(UpdateGrub), None)
 if dev:
    update_grub_core(dev.grub_device)```

File: repos/system_upgrade/el7toel8/actors/checkgrubcore/actor.py (hunk 3 above)
Comment (Rezney) [reply repos/system_upgrade/el7toel8/actors/checkgrubcore/actor.py]: we already discussed this offline, this will be improved...

**Comment by Rezney** [comment]
@vinzenz @fernflower all remarks should be hopefully addressed...

**Comment by Rezney** [comment]
e2e tests

**Comment by Rezney** [comment]
e2e tests

**Comment by Rezney** [comment]
@pirat89 @vinzenz @fernflower is this PR good to go?

**Comment by lsthompson** [comment]
This appears as a high risk warning when ELevating from EL7 to 8 and could potentially benefit from a wording tweak, as the summary currently implies that it is a blocker, rather than a cleared blocker (as the actor means it's advice-only).

`Risk Factor: high`
`Title: GRUB core will be updated during upgrade`
`Summary: On legacy (BIOS) systems, GRUB core (located in the gap between the MBR and the first partition) does not get automatically updated when GRUB is upgraded.`
`Key: baa75fad370c42fd037481909201cde9495dacf4`

Potential re-wording could be:

> On legacy (BIOS) systems, GRUB core (located in the gap between the MBR and the first partition) _would otherwise not be_ automatically updated when GRUB is upgraded.

If the title is considered ample explanation, not a problem. Thanks either way!

**Comment by pirat89** [comment]
I guess it's even outdated:
```
 TODO: remove this actor completely after the deprecation period expires
```

PR #434, opened 2020-01-17
Enable IPU with beta|htb content for testing

pirat89:
By default only 'ga' repositories should be used. But for devel purposes it should be possible to use beta and htb repositories as well. To change the product type, set envars:

- LEAPP_DEVEL_SOURCE_PRODUCT_TYPE
- LEAPP_DEVEL_TARGET_PRODUCT_TYPE

Expected values are **beta** and **htb** (and **ga**, but that's default) - it's case-insensitive. *Invalid values* raises `ValueError`. Only repositories of expected product type are used. Others are skipped. The PR requires new repomap file, which should contains additional two fields (without that, e2e tests will fail).

The original model is unchanged. The new fields are used just for filtering and there is no need to use them in any next actor.


pirat89:
P.S. I haven't test it yet, so I expect there are hidden errors.

pirat89:
@bocekm @fernflower : Just letting you know that I did several changes in the `repositoriesmapping`. There is even one FIXME, which I wanted to keep the fix for later, but if you want, I can do it now as well. I am going to check the report now, to see whether it would be better to change the message again.

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -54,44 +54,61 @@ def _prep_repository_access(context, target_userspace):
...
+def _get_expected_product_type(envar_name):
+    """
+    Get expected product type (ga/beta/htb)
+    """
+    # TODO: this function is currently used in another actor as well
+    # - maybe put it into the shared lib? in case we find another use for it..
+    # NOTE: currently we are not printing any debug msg anywhere as LEAPP_DEVEL
+    # envars are already checked
+    val = config.get_env(envar_name, '').lower()
+    return val if val in ('beta', 'htb') else 'ga'
```
bocekm: Stop the execution if there's anything else as a value than these three options.

repos/system_upgrade/el7toel8/actors/repositoriesblacklist/libraries/library.py
```diff
@@ -13,6 +13,11 @@ def _get_list_of_optional_repos():
     repo_map = next(api.consume(RepositoriesMap), None)
     if repo_map:
         for repo in repo_map.repositories:
+            # FIXME: that's not enough when src system is beta/htb as in such
```
bocekm: Add conditions based on the SRC/DST prod type env vars to alter the search string.

repos/system_upgrade/el7toel8/actors/repositoriesmapping/libraries/library.py
```diff
@@ -1,69 +1,90 @@
...
 from leapp.models.fields import ModelViolationError
 
+REPOMAP_FILE = '/etc/leapp/files/repomap.csv'
 
-def inhibit_upgrade(title):
-    summary = 'Read documentation at: https://access.redhat.com/articles/3664871 for more information ' \
-        'about how to retrieve the files'
+
+def _inhibit_upgrade(msg, repo_file):
+    # TODO: change it to the link..
```
bocekm: Remove this TODO

pirat89:
I just realized that original idea with inhibitor as it is used is invalid, as the message is not showed anyway because crashes of other actors. So I will raise error as well.

pirat89:
I will finish it in the morning.

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
pirat89: Done. I will add tests for the function yet.

repos/system_upgrade/el7toel8/actors/repositoriesmapping/libraries/library.py
```diff
@@ -1,69 +1,74 @@
...
 
-from leapp.exceptions import StopActorExecution
-from leapp import reporting
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.common import config
 from leapp.libraries.stdlib import api
 from leapp.models import RepositoriesMap, RepositoryMap
 from leapp.models.fields import ModelViolationError
 
+REPOMAP_FILE = '/etc/leapp/files/repomap.csv'
```
bocekm: You haven't used this constant anywhere. You can use it, instead of passing it from the actor.py.

repos/system_upgrade/el7toel8/actors/repositoriesmapping/libraries/library.py
```diff
@@ -1,69 +1,74 @@
...
+        details={'hint': ('Read documentation at the following link for more'
+                          ' information about how to retrieve the valid files:'
+                          ' https://access.redhat.com/articles/3664871')})
 
-def scan_repositories(path):
+
+def _read_repofile(path):
     if not os.path.isfile(path):
-        inhibit_upgrade('Repositories map file not found ({})'.format(path))
+        _inhibit_upgrade('Repositories map file not found ({}).'.format(path))
```
bocekm:
```suggestion
        _inhibit_upgrade('Repository mapping file not found ({}).'.format(path))
```

repos/system_upgrade/el7toel8/actors/repositoriesmapping/libraries/library.py
```diff
@@ -1,69 +1,74 @@
...
 
     if os.path.getsize(path) == 0:
-        inhibit_upgrade('Repositories map file is invalid ({})'.format(path))
+        _inhibit_upgrade('Repositories map file is invalid ({}).'.format(path))
```
bocekm:
```suggestion
        _inhibit_upgrade('Repository mapping file is invalid ({}).'.format(path))
```

repos/system_upgrade/el7toel8/actors/repositoriesmapping/libraries/library.py
```diff
@@ -1,69 +1,74 @@
...
-        reporting.Tags([reporting.Tags.UPGRADE_PROCESS]),
-        reporting.Flags([reporting.Flags.INHIBITOR])
-    ])
-    raise StopActorExecution()
 
+def _inhibit_upgrade(msg):
+    raise StopActorExecutionError(
+        msg,
+        details={'hint': ('Read documentation at the following link for more'
+                          ' information about how to retrieve the valid files:'
```
bocekm:
```suggestion
                          ' information about how to retrieve the valid file:'
```

repos/system_upgrade/el7toel8/actors/repositoriesmapping/libraries/library.py
```diff
@@ -1,69 +1,74 @@
...
-                    data.line_num, err))
+            )
+        except (ModelViolationError, ValueError) as err:
+            _inhibit_upgrade('Repositories map file is invalid, offending line number: {} ({}).'
+                             ' It is possible the file is out of date.'
+                             .format(line_num, err))
 
     if not repositories:
-        inhibit_upgrade('Repositories map file is invalid ({})'.format(path))
+        _inhibit_upgrade('Repository map file is invalid. Could not find any repository mapping record.')
```
bocekm:
```suggestion
        _inhibit_upgrade('Repository mapping file is invalid. Could not find any repository mapping record.')
```

repos/system_upgrade/el7toel8/actors/repositoriesmapping/libraries/library.py
pirat89: Aah. Yeah. That's artefact from early changes. I will rewrite it

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -58,40 +59,41 @@ def _get_product_certificate_path():
...
-    variant = api.current_actor().configuration.os_release.variant_id
     architecture = api.current_actor().configuration.architecture
     target_version = api.current_actor().configuration.version.target
-    # TODO: so far only base channel is available in rhel8
-    target_channel = 'base'
-    certs_folder = api.get_common_folder_path(PROD_CERTS_FOLDER)
+    target_product_type = get_product_type('target')
+    certs_dir = api.get_common_dir_path(PROD_CERTS_FOLDER)
 
+    # TODO: do we need EUS/... here or is it ga one enough to get eus repos?
```
bocekm: The EUS prod certs have different numbers: https://docs.google.com/document/d/1UQ3a0GG7kN3_-isb6Bd612UX4BgtLRS0sWPkeMMQ6ds/edit#heading=h.z4wi73tiq5hw
I think you can remove the TODO.

pirat89: I know, but the TODO is about: are we able to get eus content using a ga certificate or do we have to use eus certificates?.. as well, missing the detection whether we are on eus or not know. so keeping the todo until we know answers.

pirat89:
added ga certificates for 8.2

bocekm:
It's not possible to upgrade to 8.2 beta now, because beta products don't have any `release` set on the CDN, meaning that when leapp executes `subscription-manager release --set 8.2` (or `8.1`, it's the same), such command fails.
But that's a separate issue to this PR. We'll tackle that later.

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -58,40 +59,41 @@ def _get_product_certificate_path():
...
         }
     }
 
     try:
-        cert = prod_certs[variant][architecture][target_channel]
+        cert = prod_certs[architecture][target_product_type]
     except KeyError as e:
         raise StopActorExecutionError(message=('Failed to determine what certificate to use for {}.'.format(e)))
 
-    return os.path.join(certs_folder, target_version, cert)
```
bocekm: Add check that the cert exists to avoid tracebacks with an error like this:
`IOError: [Errno 2] No such file or directory: u'/etc/leapp/repos.d/system_upgrade/el7toel8/files/prod-certs/8.1/486.pem'`

pirat89:
leapp test

pirat89:
leapp build

pirat89:
leapp-ci build

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
pirat89: Fixed

PR #436, opened 2020-01-17
Add tests for OpenSshProtocolCheck

drehak:
- split functionality into the library
- add tests

Heavily similar to #424, since they consume the same model.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/opensshprotocolcheck/tests/unit_test.py
+++ b/repos/system_upgrade/el7toel8/actors/opensshprotocolcheck/tests/unit_test.py
@@ -1,0 +1,27 @@
+import pytest
+
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.actor import library
+from leapp.models import OpenSshConfig, OpenSshPermitRootLogin, Report
+from leapp.snactor.fixture import current_actor_context
+
+
+def test_no_config(current_actor_context):
+    with pytest.raises(StopActorExecutionError):
+        library.process(iter([]))
+
+
+osprl = OpenSshPermitRootLogin(value='no')
+
+
+@pytest.mark.parametrize('protocol', [None, '1', '2', '1,2', '2,1', '7'])
+def test_protocol(current_actor_context, protocol):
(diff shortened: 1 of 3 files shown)


drehak:
@Jakuje

repos/system_upgrade/el7toel8/actors/opensshprotocolcheck/tests/unit_test.py
```diff
@@ -0,0 +1,32 @@
...
+
+def test_no_config(current_actor_context):
+    with pytest.raises(StopActorExecutionError):
+        library.process(iter([]))
+
+
+osprl = OpenSshPermitRootLogin(value='no')
+
+
+@pytest.mark.parametrize('protocol', ['', '7'])
```
Jakuje: Just a nit that the real values that could be in the config are `1`, `2` or `1,2` so testing these would probably be cleaner, but this does its job fine.

drehak: I see... Updated the tests, thanks.

repos/system_upgrade/el7toel8/actors/opensshprotocolcheck/tests/unit_test.py
```diff
@@ -0,0 +1,32 @@
...
+    with pytest.raises(StopActorExecutionError):
+        library.process(iter([]))
+
+
+osprl = OpenSshPermitRootLogin(value='no')
+
+
+@pytest.mark.parametrize('protocol', ['', '1', '2', '1,2', '2,1', '7'])
+def test_protocol(current_actor_context, protocol):
+    if protocol:
```
fernflower: Nit: I suspect L19-27 can be squashed into `current_actor_context.feed(OpenSshConfig(permit_root_login=[osprl], protocol=protocol))` if no_protocol test case is passed as None.

drehak: Yeah, thanks.

drehak:
Applied feedback and rebased.

PR #441, opened 2020-01-30
verifydialogs: make report message more explicit

fernflower:
Previous "Dialog choice required" turned out to be too cryptic.


drehak:
A required answer with `True` or `False` as available choices passes the upgrade when something else is provided:
```
[remove_pam_pkcs11_module_check]
option = Lol
```
I don't know if it's up to this actor to check that though.

PR #446, opened 2020-01-31
DO NOT MERGE: plumbing for FIPS mode preservation

t184256:
This is a work in progress intended to plumb FIPS-mode awareness through LEAPP.

This is in no way ready for merging or even reviewing, I'm just pushing here for your build facilities to aid my workflow. Comments are welcome though.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
@@ -10,6 +10,14 @@ from leapp.models import BootContent
 def add_boot_entry():
     debug = 'debug' if os.getenv('LEAPP_DEBUG', '0') == '1' else ''
 
+    args = debug + ' enforcing=0 rd.plymouth=0 plymouth.enable=0'
+
+    with open('/proc/cmdline') as f:
+        fips = 'fips=1' in f.read()
+
+    if fips:
+        args += ' fips=1 FIPSTRACE=2'
+
     kernel_dst_path, initram_dst_path = get_boot_file_paths()
     try:
         _remove_old_upgrade_boot_entry(kernel_dst_path)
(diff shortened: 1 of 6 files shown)


repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
```diff
@@ -10,6 +10,14 @@
 def add_boot_entry():
     debug = 'debug' if os.getenv('LEAPP_DEBUG', '0') == '1' else ''
 
+    args = debug + ' enforcing=0 rd.plymouth=0 plymouth.enable=0'
+
+    with open('/proc/cmdline') as f:
+        fips = 'fips=1' in f.read()
+
+    if fips:
```
```suggestion
    fips = next(self.consume(FIPSFacts))
    if fips.enabled:
```
Something like this could be enough and you could remove the _propagatefips_ actor.

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/libraries/modscan.py
```diff
@@ -47,6 +47,13 @@ def _create_dracut_modules():
 
 def _create_initram_packages():
     required_pkgs = _REQUIRED_PACKAGES
+
+    with open('/proc/cmdline') as f:
+        fips = 'fips=1' in f.read()
+
+    if fips:
+        required_pkgs.append('dracut-fips')
```
bocekm: This won't work because the upgrade initramfs we create is based on RHEL 8, and there's no dracut-fips on RHEL 8 anymore: https://github.com/ComplianceAsCode/content/pull/3623
The functionality of the `dracut-fips` package is provided by the `crypto-policies` package and the `fips-mode-setup` tool on RHEL 8.

repos/system_upgrade/el7toel8/actors/addupgradebootentry/libraries/library.py
pirat89: Actually the actor will be probably needed. This is not something what should be in the addupgradebootentry actor. The actor should be consuming special msg, based on which will do the thing. As it is highly possible we will need more stuff in future like that for cmdline - as well, users would needed to be able to extend the IPU solution. But ok to keept it there for now, as solution is not created yet.

bocekm:
@t184256, do you think we can close this one since there's no clear time line for finishing it?

t184256:
@bocekm, use your best judgement. I'm under the impression that 8.4 is the new target.

bocekm:
Yeah, that's many many months ahead. Please reopen it later or create a new one then.

PR #439, opened 2020-01-27
Inhibit upgrade if some links inside / are absolute

drehak:
If `/bin` or `/lib` point to a relative path, there's a fair chance of _dracut_ failing during _InitRamStartPhase_, if not earlier.
We're not sure if this also applies to `/lib64` and `/sbin` but after discussing this in person, we've decided to insist on **all** symbolic links inside `/` being relative.

PR #447, opened 2020-02-03
Use real path for systemd-nspawn in initramfs

drehak:
An alternative to #439, this is an attempt to actually solve the bug rather than inhibiting.
The core issue is that if `/bin` is a link to an absolute path, the resolution of `/bin/systemd-nspawn` to `/usr/bin/systemd-nspawn` fails. So why not just directly use the latter?

Relates to: RHELLEAPP-1481


drehak:
Tested this by upgrading two vagrant VMs, one default and one with `/bin` and `/lib` linked to `/usr/bin` and `/usr/lib`. Both upgraded just fine.
I guess this will need a great deal of testing and carefulness, since dracut/initramfs related code is a sensitive territory and can break stuff in many ways.

pirat89:
I like this idea more. I am going to test it.

drehak:
e2e tests

drehak:
e2e tests

pirat89:
leapp-ci build

bocekm:
@drehak, please restart the travis ci to see if it was just some infra issue.

drehak:
e2e tests

AloisMahdal:
e2e tests

PR #448, opened 2020-02-03
Remove leapp sos plugin

bocekm:
- the plugin has been added to sos in RHEL:
  https://bugzilla.redhat.com/show_bug.cgi?id=1741321
- keeping it also in leapp causes a file conflict


vinzenz:
While I agree on removing all the parts from the spec files I am not sure if we want to remove it completely from the repository just yet.
Shouldn't we keep the sos report python file so far?

bocekm:
Keeping the python file would be useless/duplicate when it's already included in the _sos_ package IMHO.

I've added the DO_NOT_MERGE as I'm still waiting on @shaded-enmity's reply on whether we should allow upgrades of 7.8 only or also older systems (RHELLEAPP-2056). If the latter, then it's gonna be more difficult - I don't know how would we do that, but we would need to ensure that there's no file conflict during the dnf transaction check&test, because the plugin has been added to RHEL 8.2 sos pkg.

AloisMahdal:
e2e tests

AloisMahdal:
Note that the upgrade part of the E2E test is expected to fail because (@Madeyro told me) we have some test code that relies on sos being present, so that part needs to be updated.

(I'll be able to fix it easily once I'll see where exactly the fail is.)

Actually focus of PR should be testable by the `static` batch, ie. RPM's get installed and are sane.

bocekm:
@shaded-enmity, there're I guess two options. We should cover the specific use case of gathering the leapp debug data on RHEL-ALT 7.6 in the documentation. We can tell customers on RHEL-ALT 7.6 either to copy the plugin from somewhere to the sos plugin folder, or to gather these and these files manually.

If the former, then we could have the sos plugin as part of the _leapp-repository_ package (not  _leapp-repository-sos-plugin_), install it to some dir (say _/etc/leapp/files/_) and point customer to copy the plugin from there. But IMHO it would be kinda dirty to have the file laying around somewhere like this.

bocekm:
e2e tests

AloisMahdal:
Sorry, the latest e2e tests were invalid. turns out that for reasons(tm) the change has broken our validator which in turn "thought" that lpr is not installed

New version of the framework is on its way; I'll re-run murphy as well.

AloisMahdal:
e2e tests

AloisMahdal:
(fingers crossed)

AloisMahdal:
Tried manually on openstack; RHEL-7.6:

    root@ci-vm-10-0-136-254 ~]# cat /etc/system-release-cpe
    cpe:/o:redhat:enterprise_linux:7.6:ga:server
    [root@ci-vm-10-0-136-254 ~]# rpm -qa leapp\*
    leapp-repository-deps-0.9.0-0.202002131607Z.34e45e2.remove_sos_plugin.PR448.el7_6.noarch
    leapp-repository-0.9.0-0.202002131607Z.34e45e2.remove_sos_plugin.PR448.el7_6.noarch
    leapp-0.9.0-100.202002130810Z.bc8a1ec.master.el7_6.noarch
    leapp-deps-0.9.0-100.202002130810Z.bc8a1ec.master.el7_6.noarch
    [root@ci-vm-10-0-136-254 ~]# 

and RHEL-7.8:

    [root@ci-vm-10-0-136-96 ~]# cat /etc/system-release-cpe
    cpe:/o:redhat:enterprise_linux:7.8:beta:server
    [root@ci-vm-10-0-136-96 ~]# rpm -qa leapp\*
    leapp-repository-0.9.0-0.202002131607Z.34e45e2.remove_sos_plugin.PR448.el7_6.noarch
    leapp-repository-deps-0.9.0-0.202002131607Z.34e45e2.remove_sos_plugin.PR448.el7_6.noarch
    leapp-0.9.0-100.202002130810Z.bc8a1ec.master.el7_6.noarch
    leapp-deps-0.9.0-100.202002130810Z.bc8a1ec.master.el7_6.noarch
    [root@ci-vm-10-0-136-96 ~]#

pirat89:
Discussed with Lojza. Merging so they can test it.

shaded-enmity:
@bocekm Please outline the process of collecting the data in https://docs.google.com/document/d/1SbK_vVmLLtZJa7WGqtCySgx7WIEqyJWoUpOgcdxcANs/edit#heading=h.1obwmciyjwv7 and we can work from there

PR #450, opened 2020-02-10
Chain events of types Removal and Presence + actor refactor

drehak:
There's a peculiar pair of events in PES:

2020-02-10-191418_824x97_scrot

In RHEL 8.0, `sblim-cmpi-base` gets removed only to get reintroduced in 8.1. A pair of events typed as Removal and Presence is allegedly the best way to represent such changes, but the way events are processed right now, the Presence event doesn't stop the package from being removed. This PR keeps such packages on the target system and upgrades them.

Apart from that, included are some other smaller changes and refactors of the library code, mainly:
* indexing of tasks by enums instead of strings
* recognition of new minor releases
* filtering of events by release before processing
* several one-liners and comprehensions to shorten the code

Also unit tests.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
@@ -113,7 +113,7 @@ def get_transaction_configuration():
 
 def filter_releases_by_target(releases, target):
     return [r for r in releases if version.matches_version(
-        ['<= {}.{}'.format(target[0], target[1])], '{}.{}'.format(r[0], r[1]))]
+        ['<= {}.{}'.format(*target)], '{}.{}'.format(*r))]
 
 
 def get_events(pes_events_filepath):
--- a/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
@@ -37,11 +37,14 @@ def pes_events_scanner(pes_json_filepath):
     """Entrypoint to the library"""
     installed_pkgs = get_installed_pkgs()
     transaction_configuration = get_transaction_configuration()
-    events = get_events(pes_json_filepath)
     arch = api.current_actor().configuration.architecture
-    arch_events = filter_events_by_architecture(events, arch)
-    add_output_pkgs_to_transaction_conf(transaction_configuration, arch_events)
     target = version._version_to_tuple(api.current_actor().configuration.version.target)
+    filtered_releases = filter_releases_by_target(RELEASES, target)
+
+    events = get_events(pes_json_filepath)
+    filtered_events = filter_events_by_releases(events, filtered_releases)
+    arch_events = filter_events_by_architecture(filtered_events, arch)
+    add_output_pkgs_to_transaction_conf(transaction_configuration, arch_events)
     tasks = process_events(target, arch_events, installed_pkgs)
     filter_out_transaction_conf_pkgs(tasks, transaction_configuration)
     produce_messages(tasks)
@@ -108,6 +111,11 @@ def get_transaction_configuration():
     return transaction_configuration
 
 
+def filter_releases_by_target(releases, target):
+    return [r for r in releases if version.matches_version(
+        ['<= {}.{}'.format(target[0], target[1])], '{}.{}'.format(r[0], r[1]))]
+
+
 def get_events(pes_events_filepath):
     """
     Get all the events from the source JSON file exported from PES.
@@ -137,6 +145,10 @@ def filter_events_by_architecture(events, arch):
         if not event.architectures or arch in event.architectures:
             filtered_events.append(event)
     return filtered_events
+
+
+def filter_events_by_releases(events, releases):
+    return [e for e in events if e.to_release in releases]
 
 
 def parse_pes_events_file(path):
--- a/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
@@ -5,7 +5,7 @@ from enum import Enum
 
 from leapp.exceptions import StopActorExecution, StopActorExecutionError
 from leapp import reporting
-from leapp.libraries.common.config import architecture
+from leapp.libraries.common.config import architecture, version
 from leapp.libraries.stdlib import api
 from leapp.libraries.stdlib.config import is_verbose
 from leapp.models import (InstalledRedHatSignedRPM, PESRpmTransactionTasks, RepositoriesMap,
@@ -41,7 +41,8 @@ def pes_events_scanner(pes_json_filepath):
     arch = api.current_actor().configuration.architecture
     arch_events = filter_events_by_architecture(events, arch)
     add_output_pkgs_to_transaction_conf(transaction_configuration, arch_events)
-    tasks = process_events(arch_events, installed_pkgs)
+    target = version._version_to_tuple(api.current_actor().configuration.version.target)
+    tasks = process_events(target, arch_events, installed_pkgs)
     filter_out_transaction_conf_pkgs(tasks, transaction_configuration)
     produce_messages(tasks)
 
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -23,6 +24,15 @@
 RELEASES = ((7, 5), (7, 6), (7, 7), (7, 8), (8, 0), (8, 1))  # TODO: bad, bad hardcode
```
bocekm: When you're at it, add more releases here.
```suggestion
RELEASES = ((7, 5), (7, 6), (7, 7), (7, 8), (7, 9), (8, 0), (8, 1), (8, 2), (8, 3))  # TODO: bad, bad hardcode
```

bocekm:
Oh, ok, it has the WIP label. I was looking at the PR title for "WIP" instead. It seems like we should unify the approach.

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
drehak: Thanks, done.

drehak:
e2e tests

AloisMahdal:
e2e tests

drehak:
e2e tests

drehak:
e2e tests

drehak:
e2e tests

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -20,18 +21,31 @@
                              ])
 
 EVENT_TYPES = ('Present', 'Removed', 'Deprecated', 'Replaced', 'Split', 'Merged', 'Moved', 'Renamed')
-RELEASES = ((7, 5), (7, 6), (7, 7), (7, 8), (8, 0), (8, 1))  # TODO: bad, bad hardcode
+RELEASES = [(7, 5), (7, 6), (7, 7), (7, 8), (8, 0), (8, 1), (8, 2)]  # TODO: bad, bad hardcode
```
vinzenz: can't you build that dynamically based on the information from the file?

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -235,113 +254,107 @@ def parse_architectures(architectures):
...
     return True
 
 
-def add_packages_to_tasks(tasks, packages, key):
-    verb = key[3:].upper()  # 'to_remove' -> 'REMOVE' and so on
+def add_packages_to_tasks(tasks, packages, task_type):
     if packages:
-        api.current_logger().debug('{v} {p}'.format(v=verb, p=', '.join(packages)))
-        tasks[key].update(packages)
+        api.current_logger().debug('{v:7} {p}'.format(v=task_type.name.upper(), p=', '.join(packages)))
```
vinzenz: What's the :7 supposed to be doing? By checking the documentation i couldn't figure it out and my test didn't yield anything either, so what's that?

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -97,6 +111,11 @@ def get_transaction_configuration():
     return transaction_configuration
 
 
+def filter_releases_by_target(releases, target):
+    return [r for r in releases if version.matches_version(
+        ['<= {}.{}'.format(target[0], target[1])], '{}.{}'.format(r[0], r[1]))]
```
vinzenz:
```suggestion
        ['<= {}.{}'.format(*target)], '{}.{}'.format(*r))]
```

Just FYI - Not necessary to use, but it's shorter

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -20,18 +21,31 @@
...
-    events = get_events(pes_json_filepath)
     arch = api.current_actor().configuration.architecture
-    arch_events = filter_events_by_architecture(events, arch)
+    target = version._version_to_tuple(api.current_actor().configuration.version.target)
+    filtered_releases = filter_releases_by_target(RELEASES, target)
+
+    events = get_events(pes_json_filepath)
+    filtered_events = filter_events_by_releases(events, filtered_releases)
+    arch_events = filter_events_by_architecture(filtered_events, arch)
     add_output_pkgs_to_transaction_conf(transaction_configuration, arch_events)
```
vinzenz:
```suggestion

    events = get_events(pes_json_filepath)
    filtered_events = filter_events_by_releases(events, filtered_releases)
    releases = get_releases(events) # This where you would get the releases from the data, doesn't exist ofc
    
    target = version._version_to_tuple(api.current_actor().configuration.version.target)
    filtered_releases = filter_releases_by_target(releases, target)
    arch_events = filter_events_by_architecture(filtered_events, arch)
    add_output_pkgs_to_transaction_conf(transaction_configuration, arch_events)
```

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
drehak: Oh, that's just right padding with spaces so that the packages are justified in the log.

vinzenz: Oh right, thanks

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
drehak: Hm, I could. It's one more list traversal but we won't have to update this later.

drehak: Alright, it's a function now.

drehak:
e2e tests

drehak:
leapp-ci build

drehak:
e2e tests

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -97,6 +112,11 @@ def get_transaction_configuration():
     return transaction_configuration
 
 
+def filter_releases_by_target(releases, target):
+    return [r for r in releases if version.matches_version(
+        ['<= {}.{}'.format(*target)], '{}.{}'.format(*r))]
```
vinzenz: nit pick for future reference, this is through out all iterations the same so it could be made a variable before the loop. But this is not going to be too huge, therefore not needed to change that now. Just wanted to point that out.

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
```diff
@@ -235,113 +264,107 @@ def parse_architectures(architectures):
...
                 api.current_logger().warning(
                     '{p} :: {r} to be removed / currently installed - keeping package'.format(
-                        p=package, r=current['to_install'][package]))
-                current['to_keep'][package] = current['to_install'][package]
-                del tasks['to_remove'][package]
-                del current['to_install'][package]
+                        p=package, r=current[Task.install][package]))
+                current[Task.keep][package] = current[Task.install][package]
+                del tasks[Task.remove][package]
+                del current[Task.install][package]
```
vinzenz: You're here modifying the container you're iterating over this is a bad thing and can lead to problems.
If you would copy the indices (keys I assume) I think this could work.

Line 341:
```python
for package in list(current[Task.install].keys()):
```

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
drehak: Good point, changed it now. ;)

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/library.py
drehak: Alright, buffered the deleted entries into a set as in the block above.

drehak:
e2e tests

drehak:
leapp-ci build

AloisMahdal:
e2e tests

pirat89:
@drehak man, when you want your commits merged, squash it next time ;-) we do not want mess in the master branch

PR #453, opened 2020-02-14
Remove sblim-wbemcli from blacklist

drehak:
This reverts commit 430ccc2f3ca1728b0f61bd498523af1ca57abc21, merged in #364.

Since the conflict between `python3-pywbem` and `sblim-wbemcli` has been fixed and verified, we can now remove `sblim-wbemcli` from `to_remove`.

### DIFF
--- a/etc/leapp/transaction/to_remove
+++ b/etc/leapp/transaction/to_remove
@@ -1,5 +1,1 @@
 ### List of packages (each on new line) to be removed from the upgrade transaction
-
-#OAMG-590
-#https://bugzilla.redhat.com/show_bug.cgi?id=1757855
-sblim-wbemcli


drehak:
e2e tests

AloisMahdal:
e2e tests

drehak:
e2e tests

pirat89:
@AloisMahdal  the failing tests seems completely irrelevant to leapp-repository at all. It is testing some snactor functionality and doesn't provide any results related to this pr. we should maybe think about split of tests? i dunno what could be best solution.

pirat89:
e2e tests

AloisMahdal:
@pirat89 You're right.  The way tests are scoped now is pretty inclusive, but it's because of the way dependencies work in leapp:  change in lp can break lpr --- that's obvious, but what's not obvious is that change in lpr actually can break lp.  (eg. `leapp upgrade` command which afaik is still developed under `leapp`.)

The latter is less likely but still possible, so to err on the safe side we run more tests rather than less.  If you want to change it please don't hesitate to bring it up and we can discuss it further.

AloisMahdal:
Anyway, the reason why test failed was indeed because of long-standing infra issues, which should be resolved now (fingers crossed!) and the tests should even be much faster now.  So it's time to retry.

AloisMahdal:
e2e tests

AloisMahdal:
I've just found out that we actually have reproducer for this in morf (`@oamg590`).  Ie. I will run it as part of `ver` dispatcher along e2e tests.

AloisMahdal:
Also old e2e seems to have been somehow messed up, and instead of running new build i got some strange mixup of results.  I have no idea what's happening but I'm going to try and re-run it once more.

AloisMahdal:
e2e tests

AloisMahdal:
OK, I got traceback that looks like #461.  @drehak  Shouldn't this be rebased?

drehak:
e2e tests

pirat89:
e2e tests

pirat89:
previous tests failed because of internal infrastructure. re-running tests again.

AloisMahdal:
Added also verification with the pre-installed packages; the `ver` batch.

AloisMahdal:
For some reason Petr's trigger above didn't start the regression test.  I've triggered it manually.

PR #456, opened 2020-02-19
Do not use RHSM in the nspawn container

bocekm:
It is not safe to use subscription-manager in a container: https://bugzilla.redhat.com/show_bug.cgi?id=1702691

From the RHSM team:

> There are many problems with trying to share a consumer between two systems, which we avoid generally by putting subscription-manager into "container mode".

> Only the population of redhat.repo works in this state.

From @vojtechsokol:

> The container mode (i.e. /etc/rhsm-host/ exists) means that rhsm is effectively disabled (see [this](https://github.com/candlepin/subscription-manager/blob/f9381573978f5e62ef3c6c6d1f5594e6380efe60/src/rhsm/config.py#L107) and [this](https://github.com/candlepin/subscription-manager/blob/master/src/subscription_manager/managercli.py#L397), especially the second is interresting because of the TODO line), therefore any call to rhsm command will fail.

From @pirat89:

> The refreshed redhat.repo file is only source we can use in this case to get info about repos provided by RH.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
+++ b/repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
@@ -1,20 +1,22 @@
 import itertools
 import os
 
-from leapp.exceptions import StopActorExecutionError
+from leapp.exceptions import StopActorExecution, StopActorExecutionError
 from leapp.libraries.actor import constants
 from leapp.libraries.common import dnfplugin, mounting, overlaygen, rhsm, utils
 from leapp.libraries.common.config import get_product_type
-from leapp.libraries.stdlib import CalledProcessError, api, run, config
-from leapp.models import (RequiredTargetUserspacePackages, SourceRHSMInfo, StorageInfo, TargetRepositories,
-                          TargetUserSpaceInfo, UsedTargetRepositories, UsedTargetRepository, XFSPresence)
+from leapp.libraries.stdlib import CalledProcessError, api, config, run
+from leapp.models import (RequiredTargetUserspacePackages, RHSMInfo,
+                          StorageInfo, TargetRepositories, TargetUserSpaceInfo,
+                          UsedTargetRepositories, UsedTargetRepository,
+                          XFSPresence)
 
 PROD_CERTS_FOLDER = 'prod-certs'
 
 
 def prepare_target_userspace(context, userspace_dir, enabled_repos, packages):
     """
-    Implements the creation of the target userspace.
+    Implement the creation of the target userspace.
     """
     run(['rm', '-rf', userspace_dir])
     _create_target_userspace_directories(userspace_dir)
@@ -57,7 +59,7 @@ def _prep_repository_access(context, target_userspace):
 
 def _get_product_certificate_path():
     """
-    Retrieves the required / used product certificate for RHSM.
+    Retrieve the required / used product certificate for RHSM.
     """
     architecture = api.current_actor().configuration.architecture
     target_version = api.current_actor().configuration.version.target
@@ -130,31 +132,43 @@ def _create_target_userspace_directories(target_userspace):
         )
 
 
-def gather_target_repositories(target_rhsm_info):
+def gather_target_repositories(context):
     """
-    Performs basic checks on requirements for RHSM repositories and returns the list of target repository ids to use
+    Perform basic checks on requirements for RHSM repositories and return the list of target repository ids to use
     during the upgrade.
+
+    :param context: An instance of a mounting.IsolatedActions class
+    :type context: mounting.IsolatedActions class
+    :return: List of target system repoids
+    :rtype: List(string)
     """
+    # Get the RHSM repos available in the RHEL 8 container
+    available_repos = rhsm.get_available_repo_ids(context, releasever=api.current_actor().configuration.version.target)
+
     # FIXME: check that required repo IDs (baseos, appstream)
     # + or check that all required RHEL repo IDs are available.
     if not rhsm.skip_rhsm():
-        if not target_rhsm_info.available_repos or len(target_rhsm_info.available_repos) < 2:
+        if not available_repos or len(available_repos) < 2:
             raise StopActorExecutionError(
-                message='Cannot find required basic RHEL repositories.',
+                message='Cannot find required basic RHEL 8 repositories.',
                 details={
-                    'hint': ('It is required to have RHEL repository on the system'
+                    'hint': ('It is required to have RHEL repositories on the system'
                              ' provided by the subscription-manager. Possibly you'
                              ' are missing a valid SKU for the target system or network'
                              ' connection failed. Check whether your system is attached'
-                             ' to the valid SKU providing target repositories.')
+                             ' to a valid SKU providing RHEL 8 repositories.')
                 }
             )
 
     target_repoids = []
     for target_repo in api.consume(TargetRepositories):
         for rhel_repo in target_repo.rhel_repos:
-            if rhel_repo.repoid in target_rhsm_info.available_repos:
+            if rhel_repo.repoid in available_repos:
                 target_repoids.append(rhel_repo.repoid)
+            else:
+                # TODO: We shall report that the RHEL repos that we deem necessary for the upgrade are not available.
+                # The StopActorExecutionError called above might be moved here.
+                pass
         for custom_repo in target_repo.custom_repos:
             # TODO: complete processing of custom repositories
             # HINT: now it will work only for custom repos that exist
@@ -163,24 +177,63 @@ def gather_target_repositories(target_rhsm_info):
             # + outside of rhsm..
             # #if custom_repo.repoid in available_target_repoids:
             target_repoids.append(custom_repo.repoid)
+    api.current_logger().debug("Gathered target repositories: {}".format(', '.join(target_repoids)))
+    if not target_repoids:
+        raise StopActorExecutionError(
+            message='There are no enabled target repositories for the upgrade process to proceed.',
+            details={'hint': (
+                'Ensure your system is correctly registered with the subscription manager and that'
+                ' your current subscription is entitled to install the requested target version {version}'
+                ).format(version=api.current_actor().configuration.version.target)
+            }
+        )
     return target_repoids
 
 
-def perform():
+def _consume_data():
+    """Wrapper function to consume all input data."""
     packages = {'dnf'}
     for message in api.consume(RequiredTargetUserspacePackages):
         packages.update(message.packages)
 
-    rhsm_info = next(api.consume(SourceRHSMInfo), None)
+    # Get the RHSM information (available repos, attached SKUs, etc.) of the source (RHEL 7) system
+    rhsm_info = next(api.consume(RHSMInfo), None)
     if not rhsm_info and not rhsm.skip_rhsm():
         api.current_logger().warn('Could not receive RHSM information - Is this system registered?')
-        return
+        raise StopActorExecution()
 
     xfs_info = next(api.consume(XFSPresence), XFSPresence())
     storage_info = next(api.consume(StorageInfo), None)
     if not storage_info:
-        api.current_logger.error('No storage info available cannot proceed.')
+        raise StopActorExecutionError('No storage info available cannot proceed.')
+    return packages, rhsm_info, xfs_info, storage_info
 
+
+def _gather_target_repositories(context, rhsm_info, prod_cert_path):
+    """
+    This is wrapper function to gather the target repoids.
+
+    Probably the function could be partially merged into gather_target_repositories
+    and this could be really just wrapper with the switch of certificates.
+    I am keeping that for now as it is as interim step.
+    """
+    rhsm.set_container_mode(context)
+    rhsm.switch_certificate(context, rhsm_info, prod_cert_path)
+    return gather_target_repositories(context)
+
+
+def _create_target_userspace(context, packages, target_repoids):
+    """Create the target userspace."""
+    prepare_target_userspace(context, constants.TARGET_USERSPACE, target_repoids, list(packages))
+    _prep_repository_access(context, constants.TARGET_USERSPACE)
+    dnfplugin.install(constants.TARGET_USERSPACE)
+    # and do not forget to set the rhsm into the container mode again
+    with mounting.NspawnActions(constants.TARGET_USERSPACE) as target_context:
+        rhsm.set_container_mode(target_context)
+
+
+def perform():
+    packages, rhsm_info, xfs_info, storage_info = _consume_data()
     prod_cert_path = _get_product_certificate_path()
     with overlaygen.create_source_overlay(
             mounts_dir=constants.MOUNTS_DIR,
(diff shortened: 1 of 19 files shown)


repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -130,15 +133,22 @@ def _create_target_userspace_directories(target_userspace):
...
 
 
-def gather_target_repositories(target_rhsm_info):
+def gather_target_repositories():
     """
     Performs basic checks on requirements for RHSM repositories and returns the list of target repository ids to use
     during the upgrade.
     """
+    target_rhsm_repoids = []
+    base = yum.YumBase()
```
bocekm: @vinzenz, @pirat89, can I execute this in the nspawn container? Do I need to use the context somehow?

vinzenz: Not like this, if you want to execute it in the nspawn container you need to make it an external script and use the context.run function to run it.
Basically you'd need to move the yum part in a python script and print the output so it can be parsed by the actor, put the script in the tools folder of the actor and run the script with context.run and parse the stdout

bocekm: Thanks, @vinzenz. I feared it could be this kind of cumbersome. I wanted to do it more elegantly using yum python module, but in this case I think I'll just call `yum repoinfo` which also includes the information I need.

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -1,20 +1,23 @@
 import itertools
 import os
 
+import yum
```
pirat89: currently unused import

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -1,20 +1,23 @@
...
 from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.actor import constants
 from leapp.libraries.common import dnfplugin, mounting, overlaygen, rhsm, utils
 from leapp.libraries.common.config import get_product_type
-from leapp.libraries.stdlib import CalledProcessError, api, run, config
-from leapp.models import (RequiredTargetUserspacePackages, SourceRHSMInfo, StorageInfo, TargetRepositories,
-                          TargetUserSpaceInfo, UsedTargetRepositories, UsedTargetRepository, XFSPresence)
+from leapp.libraries.stdlib import CalledProcessError, api, config, run
+from leapp.models import (RequiredTargetUserspacePackages, SourceRHSMInfo,
+                          StorageInfo, TargetRepositories, TargetRHSMInfo,
```
pirat89: Obviously you plan to remove TargetRHSMInfo as there is currently not any consumer. We have not yet the deprecation process, so feel free to remove it if we do not have use for that.

repos/system_upgrade/el7toel8/libraries/rhsm.py
```diff
@@ -108,26 +113,75 @@ def get_attached_skus(context, rhsm_info):
...
+
+
+def get_repos(repos_raw):
+    """
+    Generator providing all the repos available through yum/dnf.
+
+    :rtype: Iterator[:py:class:`leapp.libraries.common.rhsm.Repo`]
+    """
+    # TODO: Log a warning on the following msg that `yum repoinfo` may produce:
+    #  "Repository rhel-7-server-eus-rpms is listed more than once in the configuration"
```
pirat89: @bocekm shouldn't we report that as well? I mean in the report. As it is something what I believe it should not happen on correctly configured system and it could be nice if user resolve that on their own. As it looks like they produce the repo file with the id, that should not appear on the rhel-8 system. as well, if there could be set `enabled=1` - as it's kinda custom repository once it is provided out of the repo file managed by rhsm - so it could happen, they will be able to "install" some rpms that are not supposed to be delivered on rhel-8.

pirat89: of course, that could be part of different PR in future. It doesn't have to be resolved in this one.

repos/system_upgrade/el7toel8/libraries/rhsm.py
```diff
@@ -108,26 +113,75 @@ def get_attached_skus(context, rhsm_info):
...
+    # Split all the available repos per one repo
+    for repo_raw_params in re.findall(
+            r"Repo-id.*?Repo-filename.*?\n",
+            repos_raw,
+            re.DOTALL | re.MULTILINE):
+        yield parse_repo_params(repo_raw_params)
+
+
+def parse_repo_params(repo_raw_params):
+    """Parse multiline string holding repo parameters to distill the important ones."""
```
pirat89: `distill` :) interesting wording

repos/system_upgrade/el7toel8/libraries/rhsm.py
```diff
@@ -108,26 +113,75 @@ def get_attached_skus(context, rhsm_info):
...
+    try:
+        repo.id = get_repo_param(r"^Repo-id\s+:\s+(.*?)(/.*)?$", repo_raw_params, "Repo-id")
+        repo.file = get_repo_param(r"^Repo-filename:\s+(.*?)$", repo_raw_params, "Repo-filename")
+    except ValueError, err:
+        api.current_logger().warn("Failed to parse the '%s' repo parameter of the `yum repoinfo` output", err.args[0])
+
+    return repo
+
+
+def get_repo_param(pattern, repo_raw_params, param):
```
pirat89: reading smoe functions, they looks they have `internal` purposes. Could you distinguish these using the `_` prefix? So it will be more clear for people, what they should/shouldn't use as it's shared library. as well, from the deprecation point of view, we will not have to care about renaming/changing such functions in future if we want.

repos/system_upgrade/el7toel8/libraries/rhsm.py
```diff
@@ -257,12 +296,12 @@ def get_existing_product_certificates(context, rhsm_info):
 
 
 @contextlib.contextmanager
-def switched_certificate(context, rhsm_info, cert_path, version):
+def switch_certificate(context, rhsm_info, cert_path):
```
pirat89: the rename is something what we should consider for deprecation process as well. But as told above, as we do not have set deprecation process yet, I am ok in this pr to go with *hard* remove. If you want, you can keep older name as well otherwise.

repos/system_upgrade/el7toel8/libraries/rhsm.py
```diff
@@ -108,26 +113,75 @@ def get_attached_skus(context, rhsm_info):
...
+def parse_repo_params(repo_raw_params):
+    """Parse multiline string holding repo parameters to distill the important ones."""
+    repo = Repo()
+    try:
+        repo.id = get_repo_param(r"^Repo-id\s+:\s+(.*?)(/.*)?$", repo_raw_params, "Repo-id")
+        repo.file = get_repo_param(r"^Repo-filename:\s+(.*?)$", repo_raw_params, "Repo-filename")
+    except ValueError, err:
+        api.current_logger().warn("Failed to parse the '%s' repo parameter of the `yum repoinfo` output", err.args[0])
+
+    return repo
```
pirat89:
```
from collections import namedtuple
```
```suggestion
    try:
        id = get_repo_param(r"^Repo-id\s+:\s+(.*?)(/.*)?$", repo_raw_params, "Repo-id")
        file = get_repo_param(r"^Repo-filename:\s+(.*?)$", repo_raw_params, "Repo-filename")
        return namedtuple('Repository', ['repoid', 'file'])(id, file)
    except ValueError, err:
        api.current_logger().warn("Failed to parse the '%s' repo parameter of the `yum repoinfo` output", err.args[0])
```

repos/system_upgrade/el7toel8/libraries/rhsm.py
bocekm: I'm detecting it and inhibit the upgrade now. I just hope that I can report an inhibitor from a common library.

repos/system_upgrade/el7toel8/libraries/rhsm.py
bocekm: Yeah, hard remove, similar to others in this PR.

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
bocekm: Yeah, I've removed it.

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -5,16 +5,18 @@
 from leapp.libraries.actor import constants
 from leapp.libraries.common import dnfplugin, mounting, overlaygen, rhsm, utils
 from leapp.libraries.common.config import get_product_type
-from leapp.libraries.stdlib import CalledProcessError, api, run, config
-from leapp.models import (RequiredTargetUserspacePackages, SourceRHSMInfo, StorageInfo, TargetRepositories,
-                          TargetUserSpaceInfo, UsedTargetRepositories, UsedTargetRepository, XFSPresence)
+from leapp.libraries.stdlib import CalledProcessError, api, config, run
+from leapp.models import (RequiredTargetUserspacePackages, RHSMInfo,
+                          StorageInfo, TargetRepositories, TargetRHSMInfo,
```
pirat89: regarding your latest changes, there is not any targetrhsminfo anymore.

repos/system_upgrade/el7toel8/libraries/rhsm.py
```diff
@@ -93,57 +95,121 @@ def _no_op(*args, **kwargs):
...
-        with _handle_rhsm_exceptions():
-            result = context.call(['subscription-manager', 'repos', '--list-enabled'], split=False)
-            rhsm_info.enabled_repos = _RE_REPO_UID.findall(result['stdout'])
+    with _handle_rhsm_exceptions():
+        result = context.call(['subscription-manager', 'repos', '--list-enabled'], split=False)
+        return _RE_REPO_UID.findall(result['stdout'])
 
 
 @with_rhsm
 def _get_repositories_to_use(context, rhsm_info, target_repositories):
```
bocekm: @vinzenz, @pirat89, this function is not used anywhere. And it seems it has not been ever. Do you know something about it? I would just remove it.

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
bocekm: Yeah, that was a leftover. Thanks for pointing it out. It's removed now.

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -130,30 +132,38 @@ def _create_target_userspace_directories(target_userspace):
...
-                message='Cannot find required basic RHEL repositories.',
+                message='Cannot find required basic RHEL 8 repositories.',
                 details={
-                    'hint': ('It is required to have RHEL repository on the system'
+                    'hint': ('It is required to have RHEL repositories on the system'
                              ' provided by the subscription-manager. Possibly you'
                              ' are missing a valid SKU for the target system or network'
                              ' connection failed. Check whether your system is attached'
-                             ' to the valid SKU providing target repositories.')
+                             ' to a valid SKU providing RHEL 8 repositories.')
```
pirat89: just nitpick: replacing `target` wording by `RHEl 8` is little contra-productive as we will have to rewrite it in future again. Many actors could be used in future as well for another system and it is better to learn people what is "target OS" in our case than write the thing two times. But it's minor thing and definitely not blocking the merge.

repos/system_upgrade/el7toel8/libraries/rhsm.py
```diff
@@ -93,81 +95,122 @@ def _no_op(*args, **kwargs):
...
+    duplicates = []
+    for duplicate in re.findall(
+            r'Repository ([^\s]+) is listed more than once', repos_raw_stderr, re.DOTALL | re.MULTILINE):
+        duplicates.append(duplicate)
+
+    if not duplicates:
+        return
+
+    api.current_logger().warn('The following repoids are defined multiple times:\n{0}'
+                              .format('\n'.join(duplicates)))
```
pirat89: I prefer to use the separator we are using on several other places to be consistent and improve readability, e.g. here: https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/actors/vsftpdconfigcheck/libraries/library.py

repos/system_upgrade/el7toel8/libraries/rhsm.py
```diff
@@ -93,81 +95,122 @@ def _no_op(*args, **kwargs):
...
+
+    reporting.create_report([
+        reporting.Title('A YUM/DNF repository defined multiple times'),
+        reporting.Summary(
+            'The `yum repoinfo` command reports that the following repositories are defined multiple times:\n{0}'
+            .format('\n'.join(duplicates))
```
pirat89: same as above.

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
bocekm: IMHO, using `target` in the message just makes a room for doubt or confusion. `RHEL 8` is specific and non-ambiguous. If we ever need to reuse this piece of code, it's trivial to update it.

repos/system_upgrade/el7toel8/libraries/rhsm.py
pirat89: As discussed, I suggest to remove it in separate commit (in this PR). It seems that it's really death-code.

bocekm: Oh, I hope it does not bring death to anyone. Let's get rid of it before it does: https://github.com/oamg/leapp-repository/pull/468

repos/system_upgrade/el7toel8/libraries/rhsm.py
```diff
@@ -291,37 +346,25 @@ def switched_certificate(context, rhsm_info, cert_path, version):
...
-        # Restore backup of product certificates
-        context.call(['rm', '-rf', pki_path], checked=False)
-        context.call(['cp', '-a', pki_backup_path, pki_path], checked=False)
-        unset_release(context)
-        # Restore release - only ga has releases in rhsm
-        if get_product_type('source') == 'ga':
-            restore_release(context, rhsm_info)
+    # Restore product certificates from the backup
+    context.call(['rm', '-rf', pki_path], checked=False)
+    context.call(['cp', '-a', pki_backup_path, pki_path], checked=False)
```
pirat89: with this change, it's not context manager anymore. You will switch the ertificates, restore the original and then continue. contextmanager functions do not make sense without yield

pirat89:
lgtm. waiting for tests

pirat89:
rhsm is not set to container mode

pirat89:
It's broken now in the new way. It's not sure now whether it's because of internal infrastructure or because of rhsm. Investigation in progress.

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -163,55 +177,72 @@ def gather_target_repositories(target_rhsm_info):
...
+    rhsm.switch_certificate(context, rhsm_info, prod_cert_path)
+    return gather_target_repositories(context)
+
+
+def _create_target_userspace(context, packages, target_repoids):
+    """Create the target userspace."""
+    prepare_target_userspace(context, constants.TARGET_USERSPACE, target_repoids, list(packages))
+    _prep_repository_access(context, constants.TARGET_USERSPACE)
+    dnfplugin.install(constants.TARGET_USERSPACE)
+
```
pirat89:
```suggestion
    # and do not forget to set the rhsm into the container mode again
    with mounting.NspawnActions(constants.TARGET_USERSPACE) as target_context:
        rhsm.set_container_mode(target_context)

```

PR #463, opened 2020-03-03
Improve logging for actor ziplconverttoblscfg

smitterl:
On failure, this actor didn't give much info from the failing command.
Handle both exception types separately and log info.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/ziplconverttoblscfg/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/ziplconverttoblscfg/actor.py
@@ -53,12 +53,20 @@ class ZiplConvertToBLSCFG(Actor):
                             # When the files are same, zipl failed - see the switch script
                             raise OSError('Failed to convert the ZIPL configuration to BLS.')
                         context.copy_from('/etc/zipl.conf', '/etc/zipl.conf')
-                    except (OSError, CalledProcessError):
-                        self.log.error(
-                            'Failed to execute zipl-switch-to-blscfg to convert the zipl configuration to BLS',
-                            exc_info=True
+                    except OSError as e:
+                        self.log.error('Could not call zipl-switch-to-blscfg command.',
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/el7toel8/actors/ziplconverttoblscfg/actor.py
```diff
@@ -53,12 +53,20 @@ def process(self):
...
+                        raise StopActorExecutionError(
+                            message='Failed to execute zipl-switch-to-blscfg. Reason: {}'.format(str(e))
+                        )
+                    except CalledProcessError as e:
+                        self.log.error('zipl-switch-to-blscfg execution failed,',
+                                       exc_info=True)
+                        raise StopActorExecutionError(
+                            message='zipl-switch-to-blscfg execution failed with non zero exit code.\nSTDOUT:\n'
+                                    '{stdout}\nSTDERR:\n{stderr}'.format(
+                                        stdout=e.stdout, stderr=e.stderr)
```
bocekm: @smitterl, please put the stdout/stderr to details. The message should be ideally one-liner.

See:

1. https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py#L41

1. https://leapp.readthedocs.io/en/latest/pydoc/leapp.html#leapp.exceptions.StopActorExecutionError

smitterl: Thank you @bocekm 
I updated the code accordingly. My understanding is that `details` can be any dictionary and will be dumped as-is, therefore, I could also add `stdout` to the dict:
`messaging/__init__.py`:
```python
details = json.dumps(details)
```
`models/__init__.py`:
```python
details = fields.Nullable(fields.String())
```
Please, let me know if you disagree.

smitterl:
review please

PR #457, opened 2020-02-20
Inhibit upgrade if multiple kernel-devel pkgs are installed

drehak:
When multiple `kernel-devel` packages are present, leapp attempts to upgrade them, but can't resolve the transaction in most cases. So we've decided to inhibit the upgrade and present the user with hints on removing the surplus devel kernels.

There's also a high chance of unexpected issues occuring on older kernels, so we've decided to inhibit if the machine isn't booted into the latest installed one.

PR #464, opened 2020-03-04
New ipu path RHEL-ALT 7.6 -> 8.2, RHEL 7.8 - 8.2

pirat89:
With this PR, we start to new default upgrade paths:
- RHEL 7.8 -> RHEL 8.2
- RHEL-ALT 7.6 -> RHEL 8.2

For the purposes of RHEL-ALT detection, the `IPUConfig` model has been extended by the `kernel` field, which contains `version-release` of the booted kernel on the source system. E.g.: `3.10.0-100.el7.x86_64`. We already have another one usecase, where the `uname -r` should be consumed in the PR #457 (which can be simplified then.

In the related library (`leapp.libraries.common.config.version`) is added new function `is_rhel_alt` and the function `is_supported_version` reflects the new upgrade paths as well.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/libraries/library.py
@@ -1,6 +1,7 @@
 import os
 
-from leapp import reporting
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.stdlib import run, CalledProcessError
 from leapp.models import EnvVar, OSRelease
 
 CURRENT_TARGET_VERSION = '8.1'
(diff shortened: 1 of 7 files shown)


pirat89:
leapp-ci build

pirat89:
e2e tests

repos/system_upgrade/el7toel8/actors/ipuworkflowconfig/libraries/library.py
```diff
@@ -53,5 +54,25 @@ def get_os_release(path):
...
+        # missing data (same for the report above). We are including errors
+        # in the report too if I rember well. But keeping that for later.
+        # - but both errors are very theoretical. As well return None will
+        # cause an error...
+        reporting.create_report([
+            reporting.Summary(str(e)),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Tags([reporting.Tags.OS_FACTS, reporting.Tags.SANITY]),
+            reporting.Flags([reporting.Flags.INHIBITOR]),
+        ])
```
bocekm: Use `StopActorExecutionError` instead. Report wouldn't get generated anyway.

pirat89: done.

PR #465, opened 2020-03-06
Removed unused consumes in CheckKDEGnome actor

vinzenz:
Currently the CheckKDEGnomeActor consumes InstalledRPM which is not
used at all. Therefore this patch removes the dependency.


bocekm:
leapp-ci build

PR #477, opened 2020-03-18
Inhibit upgrade if FIPS is enabled

drehak:
Due to a multitude of issues we've encountered, we can't ensure a stable and FIPS-compliant upgrade if FIPS is enabled. Because of that, if FIPS is detected as enabled in kernel command line (`fips=1`), we inhibit the upgrade.

Included is a model and an actor for scanning the kernel command line, since we didn't have one.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkfips/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkfips/actor.py
@@ -1,0 +1,35 @@
+from leapp.actors import Actor
+from leapp.exceptions import StopActorExecutionError
+from leapp.models import Report, KernelCmdline
+from leapp.tags import IPUWorkflowTag, ChecksPhaseTag
+from leapp import reporting
+
+
+class CheckFips(Actor):
+    """
+    Inhibit upgrade if FIPS is detected as enabled.
+    """
+
+    name = 'check_fips'
+    consumes = (KernelCmdline,)
+    produces = (Report,)
+    tags = (IPUWorkflowTag, ChecksPhaseTag)
+
+    def process(self):
+        cmdline = next(self.consume(KernelCmdline), None)
+        if not cmdline:
+            raise StopActorExecutionError('Cannot check FIPS state due to missing command line parameters',
+                                          details={'Problem': 'Did not receive a message with kernel command '
+                                                              'line parameters (KernelCmdline)'})
+        for parameter in cmdline.parameters:
+            if parameter.key == 'fips' and parameter.value == '1':
+                title = 'Cannot upgrade a system with FIPS mode enabled'
+                summary = 'Leapp has detected that FIPS is enabled on this system. ' \
+                          'In-place upgrade of systems in FIPS mode is currently unsupported.'
+                reporting.create_report([
+                    reporting.Title(title),
+                    reporting.Summary(summary),
+                    reporting.Severity(reporting.Severity.HIGH),
+                    reporting.Tags([reporting.Tags.SECURITY]),
+                    reporting.Flags([reporting.Flags.INHIBITOR])
+                ])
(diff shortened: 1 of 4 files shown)


drehak:
e2e tests

pirat89:
e2e tests

repos/system_upgrade/el7toel8/actors/checkfips/actor.py
```diff
@@ -0,0 +1,35 @@
...
+        cmdline = next(self.consume(KernelCmdline), None)
+        if not cmdline:
+            raise StopActorExecutionError('Cannot check FIPS state due to missing command line parameters',
+                                          details={'Problem': 'Did not receive a message with kernel command '
+                                                              'line parameters (KernelCmdline)'})
+        for parameter in cmdline.parameters:
+            if parameter.key == 'fips' and parameter.value == '1':
+                title = 'Cannot upgrade with enabled FIPS'
+                summary = 'Leapp has detected that FIPS is enabled on this system. ' \
+                          'Currently, we cannot corretly upgrade FIPS-enabled systems.'
```
t184256: I'd prefer a more neutral `in-place upgrade of systems in FIPS mode is currently unsupported`.

repos/system_upgrade/el7toel8/actors/checkfips/actor.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def process(self):
+        cmdline = next(self.consume(KernelCmdline), None)
+        if not cmdline:
+            raise StopActorExecutionError('Cannot check FIPS state due to missing command line parameters',
+                                          details={'Problem': 'Did not receive a message with kernel command '
+                                                              'line parameters (KernelCmdline)'})
+        for parameter in cmdline.parameters:
+            if parameter.key == 'fips' and parameter.value == '1':
+                title = 'Cannot upgrade with enabled FIPS'
```
t184256: Suggestion: `Cannot upgrade a system with FIPS mode enabled`.

drehak:
@pirat89 
> Ah. Ok, I changed my mind. You are delivering redundant content. You do not need additional actor for scanning the kernel cmdline. See `KernelCmdlineArg`

No, `KernelCmdlineArg` is for kernel command line parameters to be **appended in RHEL8**. If you look more closely at actors that produce this model, each of them just spits out an additional parameter for a safe upgrade. The consumer then tries to append them in final phases.
Apart from this PR, there is no place where the kernel cmdline is scanned.

drehak:
Oh, actually it is scanned in `Biosdevname` - but that's a checks phase actor and it only checks one parameter, it doesn't pass the parameters on. I'll fix that one, since there shouldn't be file reads in checks phase.

repos/system_upgrade/el7toel8/actors/checkfips/actor.py
drehak: Replaced the strings, thanks.

drehak:
e2e tests

AloisMahdal:
the fail is (at least partly) because oamg2929, fixed by #478

drehak:
e2e tests

drehak:
e2e tests

pirat89:
@drehak Ah, you are right. We should think more about names of models.

repos/system_upgrade/el7toel8/models/kernelcmdlineargs.py
```diff
@@ -4,9 +4,20 @@
 
 class KernelCmdlineArg(Model):
     """
-    Single kernel command line argument to include to RHEL-8 kernel cmdline
+    Single kernel command line parameter with/without a value
+
+    When produced alone, represents a parameter to include in RHEL-8 kernel cmdline.
     """
```
pirat89: We will have to change that later, as this looks very confusing. I would tell, we should use different model completely to represent cmdline parameters that should be included into the RHEL-8 kernel cmdline. @oamg/developers it's really wrong to have two different usecases for one model. but the rest is ok. merging.

PR #478, opened 2020-03-20
enablerhsmreposonrhel8: do not set rhsm release for beta/htb

pirat89:
The analogy to b86a44c66a4. The release can be set only for GA
system, as releases are not provided for HTB and Beta composes.


pirat89:
e2e tests

Rezney:
Let's wait for the tests but otherwise looks good to me...

AloisMahdal:
e2e tests

PR #480, opened 2020-03-26
Fix/get target repoids

pirat89:
add repository_scanning library and rewrite scan of repos
    
    We realized that we cannot use YUM reliably to get information about
    repositories in all cases (similar to DNF) with connection to RHSM...
    
    Issues with YUM:
      `yum repolist` downloads metadata about repositories (which we do
      not need) and with set --releasever to 8.x, many URLs of repos
      are invalid. E.g.:
          http://example.com/rhel7/$releasever/os
      will be in evaluated as:
          http://example.com/rhel7/8.x/os
      which causes crash.
    
    Issues with DNF:
      By default, the rhsm plugin for dnf is not installed on the system
      and not sure whether coexistence of yum & dnf is safe when used
      plugins for RHSM in the same time as RHSM itself is very tricky
      and we spent already too much time to fix every feature around.
    
    For these reasons we scan repositories directly. The final code is
    based on @shaded-enmity implementation used in the systemfacts actor.
    Unfortunately, we still have to think about RHSM, in this case, we
    need to use YUM to regenerate the redhat.repo file inside container
    to get RHEL 8 content. Currently the `yum clean all` is used, but
    ideally it should be replaced by something that will not do anythin
    but redeploy redhat.repo file using RHSM.

Related ticket: OAMG-2923

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
+++ b/repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py
@@ -1,7 +1,6 @@
 import errno
 import functools
 import grp
-import json
 import logging
 import os
 import pwd
@@ -10,10 +9,10 @@ import re
 import six
 
 from leapp.libraries.stdlib import CalledProcessError, api, run
-from leapp.libraries.common import utils
+from leapp.libraries.common import repofileutils
 from leapp.models import SysctlVariablesFacts, SysctlVariable, ActiveKernelModulesFacts, ActiveKernelModule, \
-    KernelModuleParameter, UsersFacts, User, GroupsFacts, Group, RepositoriesFacts, RepositoryFile, RepositoryData, \
-    SELinuxFacts, fields, FirewallStatus, FirewallsFacts, FirmwareFacts
+    KernelModuleParameter, UsersFacts, User, GroupsFacts, Group, RepositoriesFacts, \
+    SELinuxFacts, FirewallStatus, FirewallsFacts, FirmwareFacts
 
 
 def aslist(f):
@@ -171,40 +170,9 @@ def get_sysctls_status():
     return SysctlVariablesFacts(sysctl_variables=_get_sysctls())
 
 
-@aslist
-def _get_repositories():
-    def asbool(x):
-        return x == '1'
-
-    @aslist
-    def _parse(r):
-        with open(r, mode='r') as fp:
-            cp = utils.parse_config(fp)
-
-            for section in cp.sections():
-                prepared = {'repoid': section, 'additional_fields': {}}
-                data = dict(cp.items(section))
-                for key in data.keys():
-                    if key in RepositoryData.fields:
-                        if isinstance(RepositoryData.fields[key], fields.Boolean):
-                            data[key] = asbool(data[key])
-                        prepared[key] = data[key]
-                    else:
-                        prepared['additional_fields'][key] = data[key]
-                prepared['additional_fields'] = json.dumps(prepared['additional_fields'])
-                yield RepositoryData(**prepared)
-
-    repos = run(
-        ['find', '/etc/yum.repos.d/', '-type', 'f', '-name', '*.repo'],
-        split=True
-    )['stdout']
-    for repo in repos:
-        yield RepositoryFile(file=repo, data=_parse(repo))
-
-
 def get_repositories_status():
     ''' Get a basic information about YUM repositories installed in the system '''
-    return RepositoriesFacts(repositories=_get_repositories())
+    return RepositoriesFacts(repositories=repofileutils.get_parsed_repofiles())
 
 
 def get_selinux_status():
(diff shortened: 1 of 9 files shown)


pirat89:
@bocekm @drehak: in case you have time, testing would help. I will fix thr rhsm unit-tests tomorrow. But if you want, you can work on additional tests as well. I suggest to finish this one asap and additional unit tests could be added later. I will at least fix the current one.

pirat89:
@vinzenz @jmikovic @bocekm : Guys. I have one mystery here. Look at the failing tests and tell me, how the object, that should be only inside `rhsm_repofile` become part of `other_repofiles` as well - as input parameters generated by `pytest.mark.parametrize`. See the object `0x7f3837580350`. As well, the `other_repofiles` should contain just two objects.

```
other_repofiles = [<leapp.models.repositoriesfacts.RepositoryFile object at 0x7f38375385d0>, <leapp.models.repositoriesfacts.RepositoryFile object at 0x7f38375803d0>, <leapp.models.repositoriesfacts.RepositoryFile object at 0x7f3837580350>]
rhsm_repofile = <leapp.models.repositoriesfacts.RepositoryFile object at 0x7f3837580350>
```
The piece of code:
```python
    @pytest.mark.parametrize('other_repofiles', [
        [],
        [_gen_repofile("foo")],
        [_gen_repofile("foo"), _gen_repofile("bar")],
    ])
    @pytest.mark.parametrize('rhsm_repofile', [
        None,
        _gen_repofile(rhsm._DEFAULT_RHSM_REPOFILE, []),
        _gen_repofile(rhsm._DEFAULT_RHSM_REPOFILE, [_gen_repo("rh-0")]),
        _gen_repofile(rhsm._DEFAULT_RHSM_REPOFILE),
    
    ])
```

repos/system_upgrade/el7toel8/libraries/repository_scanning.py
```diff
@@ -0,0 +1,88 @@
...
+    Return dict of duplicate repositories {repoid: [repofile_path]}
+
+    Repository is defined multiple times if exists in multiple repofiles.
+    Redefinition inside one repository file is ignored (same in DNF).
+
+    :param repofiles:
+    :type repofiles: List(RepositoryFile)
+    :rtype: dict
+    """
+    rf_repos = {repofile: [repo.repoid for repo in repofile.data] for repofile in repofiles}
```
drehak: You can't use model instances as dict keys - they're mutable and therefore not hashable.
It would be better to skip creating an intermediate dict and inverting it, that way you only get models as values - which is fine.

repos/system_upgrade/el7toel8/libraries/repository_scanning.py
```diff
@@ -0,0 +1,88 @@
+import json
```
bocekm: I'd name the file `repofileutils.py` (you can see various Xutils.py libraries there already) or maybe even better just `repofile.py`, similarly to `rhsm.py`.

repos/system_upgrade/el7toel8/libraries/repository_scanning.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+    data = []
+    with open(repofile, mode='r') as fp:
+        cp = utils.parse_config(fp)
+        for repoid in cp.sections():
+            data.append(_parse_repository(repoid, dict(cp.items(repoid))))
+    return RepositoryFile(file=repofile, data=data)
+
+
+def scan_repositories(context=mounting.NotIsolatedActions(base_dir='/')):
```
bocekm:
```suggestion
def get_repofiles(context=mounting.NotIsolatedActions(base_dir='/')):
```
Or `get_parsed_repofiles` if you want emphasize the fact that what you return are objects of the parsed repofiles.

repos/system_upgrade/el7toel8/libraries/repository_scanning.py
```diff
@@ -0,0 +1,88 @@
...
+            else:
+                inv_dict[value].append(key)
+    return inv_dict
+
+
+def get_duplicate_repositories(repofiles):
+    """
+    Return dict of duplicate repositories {repoid: [repofile_path]}
+
+    Repository is defined multiple times if exists in multiple repofiles.
```
bocekm:
```suggestion
    A repository is defined multiple times if it exists in multiple repofiles.
```

repos/system_upgrade/el7toel8/libraries/repository_scanning.py
pirat89: fixed

pirat89:
e2e tests

pirat89:
review please

pirat89:
e2e tests

AloisMahdal:
I have no idea why the upgrade test is pending; on murphy CI it passed normally and the status update function didn't return ..

pirat89:
@AloisMahdal reload the page several times, cleaning cache. Sometime it shows me the task is still running, sometimes it shows me that everything passed. Not sure what is the new github feature :)

AloisMahdal:
nice feature... designed to keep us busy and anxious

pirat89:
*required...

PR #485, opened 2020-04-06
Deploy discover data

vinzenz:
This PR adds automatic deployment of leapp discover --json output to the https://github.com/oamg/githubio-data repository after successful test runs and only when merged to master

https://travis-ci.org/github/vinzenz/leapp-repository/builds/675270923 is a working execution using the current code

**Please note on pull requests the deployment section won't actually run as it has a condition for push to master only.**

Additionally this PR fixed some config warnings from travis


pirat89:
@vinzenz can you be more descriptive? :) (PR descr + commit msg). It's not clear what about it is unless you read the code.

PR #488, opened 2020-04-09
typo fix in the method name of api

ZhukovGreen:
Small typo fix in /repos/system_upgrade/el7toel8/actors/checkcpu/libraries/cpu.py.
This is a bit weird it existed, which means it was never run:)


ZhukovGreen:
@pirat89 merging this and com up with the test in a new MR. Thx

pirat89:
@ZhukovGreen btw, in this case I will not take care so much, by the commit messages should be more descriptive in future. you should be able to see from the commit msg, what exactly happened inside. e.g. the actor name :) but this is really small one, so mergin.

PR #489, opened 2020-04-15
Implement generic CurrentActorMocked and logger_mocked

drehak:
Despite being mocked fairly often in our tests, `current_actor` and `current_logger` didn't have a united mock implementation so far.
This commit implements a generic, fully configurable mocked actor and logger for use in all repository unit tests. All old implementations are removed.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checktargetrepos/tests/test_checktargetrepos.py
+++ b/repos/system_upgrade/el7toel8/actors/checktargetrepos/tests/test_checktargetrepos.py
@@ -8,7 +8,7 @@ from leapp.libraries.actor import library
 from leapp import reporting
 from leapp.libraries.stdlib import api
 from leapp.libraries.common import rhsm
-from leapp.libraries.common.testutils import create_report_mocked
+from leapp.libraries.common.testutils import create_report_mocked, CurrentActorMocked
 
 
 class MockedConsume(object):
@@ -24,18 +24,6 @@ class MockedConsume(object):
 
     def __call__(self, model):
         return iter([msg for msg in self._msgs if isinstance(msg, model)])
-
-
-class CurrentActorMocked(object):
-    def __init__(self, envars=None):
-        if envars:
-            envarsList = [EnvVar(name=key, value=value) for key, value in envars.items()]
-        else:
-            envarsList = []
-        self.configuration = namedtuple('configuration', ['leapp_env_vars'])(envarsList)
-
-    def __call__(self):
-        return self
 
 
 _RHEL_REPOS = [
@@ -72,8 +60,8 @@ def test_checktargetrepos_no_rhsm(monkeypatch, enable_repos, custom_target_repos
     mocked_consume = MockedConsume(_TARGET_REPOS_CUSTOM if custom_target_repos else _TARGET_REPOS_NO_CUSTOM)
     if custom_target_repofile:
         mocked_consume._msgs.append(_CUSTOM_TARGET_REPOFILE)
-    envvars = {'LEAPP_ENABLE_REPOS': 'hill,spencer'} if enable_repos else {}
-    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(envvars))
+    envars = {'LEAPP_ENABLE_REPOS': 'hill,spencer'} if enable_repos else {}
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(envars=envars))
 
     monkeypatch.setattr(reporting, 'create_report', create_report_mocked())
     monkeypatch.setattr(rhsm, 'skip_rhsm', lambda: True)
--- a/repos/system_upgrade/el7toel8/actors/targetuserspacecreator/tests/unit_test.py
+++ b/repos/system_upgrade/el7toel8/actors/targetuserspacecreator/tests/unit_test.py
@@ -6,7 +6,7 @@ import pytest
 from leapp.exceptions import StopActorExecutionError, StopActorExecution
 from leapp.libraries.actor import userspacegen
 from leapp.libraries.common import overlaygen, rhsm
-from leapp.libraries.common.testutils import CurrentActorMocked, produce_mocked
+from leapp.libraries.common.testutils import CurrentActorMocked, produce_mocked, logger_mocked
 from leapp.libraries.common.config import architecture
 from leapp.libraries.stdlib import api
 from leapp import models
@@ -52,25 +52,6 @@ def test_get_product_certificate_path(monkeypatch, result, dst_ver, arch, prod_t
     curr_actor_mocked = CurrentActorMocked(dst_ver=dst_ver, arch=arch, envars=envars)
     monkeypatch.setattr(api, 'current_actor', curr_actor_mocked)
     assert userspacegen._get_product_certificate_path() == result
-
-
-class mocked_logger(object):
-    def __init__(self):
-        self.errmsg = []
-        self.warnmsg = []
-        self.debugmsg = []
-
-    def error(self, msg):
-        self.errmsg.append(msg)
-
-    def warn(self, msg):
-        self.warnmsg.append(msg)
-
-    def debug(self, msg):
-        self.debugmsg.append(msg)
-
-    def __call__(self):
-        return self
 
 
 _PACKAGES_MSGS = [
@@ -173,7 +154,7 @@ def test_consume_data(monkeypatch, raised, no_rhsm, testdata):
                                    testdata.storage,
                                    custom_repofiles)
     monkeypatch.setattr(api, 'consume', mocked_consume)
-    monkeypatch.setattr(api, 'current_logger', mocked_logger())
+    monkeypatch.setattr(api, 'current_logger', logger_mocked())
     monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(envars={'LEAPP_NO_RHSM': no_rhsm}))
     if not xfs:
         xfs = models.XFSPresence()
--- a/repos/system_upgrade/el7toel8/actors/targetuserspacecreator/tests/unit_test.py
+++ b/repos/system_upgrade/el7toel8/actors/targetuserspacecreator/tests/unit_test.py
@@ -6,7 +6,7 @@ import pytest
 from leapp.exceptions import StopActorExecutionError, StopActorExecution
 from leapp.libraries.actor import userspacegen
 from leapp.libraries.common import overlaygen, rhsm
-from leapp.libraries.common import testutils
+from leapp.libraries.common.testutils import CurrentActorMocked, produce_mocked
 from leapp.libraries.common.config import architecture
 from leapp.libraries.stdlib import api
 from leapp import models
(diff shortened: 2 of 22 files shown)


repos/system_upgrade/el7toel8/libraries/testutils.py
```diff
@@ -30,7 +30,31 @@ def __call__(self, report_fields):
...
+        self.dbgmsg.extend(args)
+
+    def info(self, *args):
+        self.infomsg.extend(args)
+
+    def warn(self, *args):
+        self.warnmsg.extend(args)
+
+    def warning(self, *args):
+        self.warnmsg.extend(args)
```
pirat89, line 50: what about just call *warn* instead? :-) but it is nitpick; nothing important.

repos/system_upgrade/el7toel8/libraries/config/tests/test_getenvvars.py
```diff
@@ -32,9 +18,9 @@ def test_get_all_vars(monkeypatch):
     envars = {'LEAPP_DEVEL_SKIP_WIP': '0',
               'LEAPP_DEVEL_SKIP_DIP': '1',
               'LEAPP_DEVEL_SKIP_RIP': '2'}
-    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(envars))
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(envars=envars))
     assert api.current_actor().configuration.leapp_env_vars == get_all_envs()
-    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked())
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked)
```
pirat89, line 23: beaware that this is not equivalent. in you case, when the tested function affects anything inside, you will not recognize it, as you are setting "class" instead of its instance of it. so in case. keep the brackets there rather in all such cases. imagine, that logger should be part of the class as well. in suchcase, you will not be able to access the results after a "test".

repos/system_upgrade/el7toel8/libraries/testutils.py
```diff
@@ -24,6 +28,53 @@ def __call__(self, report_fields):
...
+            'configuration', ['architecture', 'kernel', 'leapp_env_vars', 'os_release', 'version']
+        )(arch, kernel, envarsList, release, version)
+
+    def __call__(self):
+        return self
+
+    def get_common_folder_path(self, folder):
+        return os.path.join(self._common_folder, folder)
+
+
```
pirat89, line 77: as we discussed, other functions are still missing. as well, logger could be part of the generated object as well. so no need to mock it in future separately if people don't want. for methods you do not implement, raise NotImplementedError

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/tests/unit_test.py
```diff
@@ -224,7 +181,7 @@ def test_consume_data(monkeypatch, raised, no_rhsm, testdata):
 
 @pytest.mark.skip(reason="Currently not implemented in the actor. It's TODO.")
 def test_gather_target_repositories(monkeypatch):
-    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked())
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked)
```
ZhukovGreen, line 184: Looks like a bug. This is a factory class, which produces desired callable obj

repos/system_upgrade/el7toel8/libraries/config/tests/test_getenvvars.py
```diff
@@ -45,19 +31,19 @@ def test_get_product_type_valid(monkeypatch):
                   'LEAPP_DEVEL_TARGET_PRODUCT_TYPE': dst}
         exp_src = 'ga' if not src else src.lower()
         exp_dst = 'ga' if not dst else dst.lower()
-        monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(envars))
+        monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(envars=envars))
```
ZhukovGreen, line 34: the same like in https://github.com/oamg/leapp-repository/pull/489/files#r414407185

repos/system_upgrade/el7toel8/libraries/config/tests/test_getenvvars.py
```diff
@@ -45,19 +31,19 @@ def test_get_product_type_valid(monkeypatch):
...
         assert exp_src == get_product_type('source')
         assert exp_dst == get_product_type('target')
     # return 'ga' if envars are not specified
-    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked())
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked)
```
ZhukovGreen, line 38: the same like https://github.com/oamg/leapp-repository/pull/489/files#r414407185

repos/system_upgrade/el7toel8/libraries/config/tests/test_version.py
```diff
@@ -74,7 +60,7 @@ def test_matches_version_pass():
     (False, ['7.5']),
 ])
 def test_matches_source_version(monkeypatch, result, version_list):
-    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked())
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(src_ver='7.6'))
```
ZhukovGreen, line 63: It looks like an antipattern (https://en.wikipedia.org/wiki/Magic_number_%28programming%29#Unnamed_numerical_constants). However not sure if it has the sense to solve since it is everywhere now.

repos/system_upgrade/el7toel8/libraries/testutils.py
```diff
@@ -24,6 +28,53 @@ def __call__(self, report_fields):
             self.report_fields.update(report.to_dict())
 
 
+# class run_mocked(object):
+
+class logger_mocked(object):
```
ZhukovGreen, line 33: I have a big doubt about mocking the logger. Unless we're not testing some magic behind leapp implementation of the logger, we could use, already provided, https://docs.pytest.org/en/latest/logging.html#caplog-fixture. 
/cc @pirat89 @vinzenz

repos/system_upgrade/el7toel8/libraries/testutils.py
ZhukovGreen, line 77: For the PR #494 and #493 I need a mocked consume method and log property implemented. Something like this:
```python
class CurrentActorMocked(object):
    def __init__(
        self,
        arch=architecture.ARCH_X86_64,
        envars=None,
        kernel='3.10.0-957.43.1.el7.x86_64',
        release_id='rhel',
        src_ver='7.8',
        dst_ver='8.1',
        msgs=None,
    ):
        envarsList = (
            [EnvVar(name=k, value=v) for k, v in envars.items()]
            if envars
            else []
        )
        version = namedtuple('Version', ['source', 'target'])(src_ver, dst_ver)
        release = namedtuple('OS_release', ['release_id', 'version_id'])(
            release_id, src_ver
        )

        self._common_folder = '../../files'
        self.configuration = namedtuple(
            'configuration',
            [
                'architecture',
                'kernel',
                'leapp_env_vars',
                'os_release',
                'version',
            ],
        )(arch, kernel, envarsList, release, version)
        self._msgs = [] if not msgs else msgs

    def __call__(self):
        return self

    def get_common_folder_path(self, folder):
        return os.path.join(self._common_folder, folder)

    def consume(self, model):
        return filter(  # pylint:disable=W0110,W1639
            lambda msg: isinstance(msg, model), self._msgs
        )

    @property
    def log(self):
        return logging.getLogger(LEAPP_LOGGER_NAME)
```

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/tests/unit_test.py
```diff
@@ -229,7 +205,8 @@ def test_perform_ok(monkeypatch):
     monkeypatch.setattr(overlaygen, 'create_source_overlay', MockedMountingBase)
     monkeypatch.setattr(userspacegen, '_gather_target_repositories', lambda *x: repoids)
     monkeypatch.setattr(userspacegen, '_create_target_userspace', lambda *x: None)
-    monkeypatch.setattr(api, 'produce', testutils.produce_mocked())
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(envars={'LEAPP_NO_RHSM': '0'}))
```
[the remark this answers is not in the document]
drehak: It seems so. I don't know, it was in the original test code. For some reason it's green on GitHub though.

repos/system_upgrade/el7toel8/actors/checktargetrepos/tests/test_checktargetrepos.py
```diff
@@ -72,8 +60,8 @@ def test_checktargetrepos_no_rhsm(monkeypatch, enable_repos, custom_target_repos
     mocked_consume = MockedConsume(_TARGET_REPOS_CUSTOM if custom_target_repos else _TARGET_REPOS_NO_CUSTOM)
     if custom_target_repofile:
         mocked_consume._msgs.append(_CUSTOM_TARGET_REPOFILE)
-    envvars = {'LEAPP_ENABLE_REPOS': 'hill,spencer'} if enable_repos else {}
-    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(envvars))
+    envars = {'LEAPP_ENABLE_REPOS': 'hill,spencer'} if enable_repos else {}
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(envars=envars))
```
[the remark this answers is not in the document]
drehak: I can't spot the difference.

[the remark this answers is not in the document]
ZhukovGreen: envars don't exists anymore, but directly binded to envars kwarg inside CurrentActorMocked

repos/system_upgrade/el7toel8/libraries/config/tests/test_getenvvars.py
drehak: What's the problem here? There's an argument, so we are creating a new instance here, not just using the default one.

drehak: What's the problem here? There is an argument, so a new instance gets created instead of using the default one.

ZhukovGreen: Not sure already, since my link doesn't work:) Probably it is linked with the redudant envars var, which is used just to pass to CurrentActorMocked. Anyway this is really not important it leaving up to you to decide. From my pov foo(key=key) looks ugly

repos/system_upgrade/el7toel8/actors/checktargetrepos/tests/test_checktargetrepos.py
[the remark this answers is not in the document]
ZhukovGreen: Closing it and leaving up to you, whether to change or not

drehak:
@pirat89 @ZhukovGreen Okay, I've thrown the logger out, added API methods and addressed feedback.
One test doesn't pass due to missing `log()`. Right now I don't know how to go around that.

repos/system_upgrade/el7toel8/libraries/testutils.py
ZhukovGreen: @drehak what about this implementation of the log property?

repos/system_upgrade/el7toel8/libraries/testutils.py
```diff
@@ -24,6 +28,111 @@ def __call__(self, report_fields):
...
+    def common_tools_paths(self):
+        raise NotImplementedError
+
+    @property
+    def tools_paths(self):
+        raise NotImplementedError
+
+    @staticmethod
+    def _get_folder_path(directories, name):
+        raise NotImplementedError
```
pirat89, line 86: In case of private methods, I am not sure whether we need to mock them. WDYT?

repos/system_upgrade/el7toel8/libraries/config/tests/test_version.py
pirat89: I believe that this is ok.

repos/system_upgrade/el7toel8/actors/checktargetrepos/tests/test_checktargetrepos.py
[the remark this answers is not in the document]
pirat89: to be honest, previous 2line code was more readable for me. but that's just matter of taste. both are ok.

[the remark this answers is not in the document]
ZhukovGreen: Sorry, about my black formatter:) The point here was in removing envars variable and directly pass its value to the CurrentactorMocked envpars kwarg. Multiline was done just because 79 line length limitation.

ZhukovGreen:
@pirat89 @drehak to follow up our discussion with @vinzenz I suggest now to add just:
```python
    @property
    def log(self):
        return logging.getLogger(__name__)
```
to the `CurrentActorMocked`. + leave mocked_actor with TODO to unify the test of log messages.
So we can merge quickly.

drehak:
@pirat89 @ZhukovGreen nice, thanks for the tips!

PR #491, opened 2020-04-20
grubdevname: Read block device directly instead of through dd

drehak:
The previously used dd command prints out binary data. If debug is enabled, the binary data gets displayed, which looks bad and potentially can break terminal output.
This more direct method has less overhead and no output at all.


Rezney:
Looks good to me. I will not approve yet and wait for @pirat89 to see if it is ok to merge.

PR #493, opened 2020-04-23
Chore/tests/sctpconfigread

ZhukovGreen:
Closes OAMG-1285

Introduces unit tests for `sctpconfigread` actor.
There are two specific points to discuss here:
1. ~~Possible conflict with PR #489~~ /cc @drehak
2. A new way of mocking builtin open is introduced, which looks safer implementation.
3. I don't understand the need to use the mocked logger from here https://github.com/oamg/leapp-repository/pull/489/files#diff-18f04a4c5ff25e505ff29b8c482d2c5fR33 . There is a pytest fixture `caplog`(https://docs.pytest.org/en/latest/logging.html#caplog-fixture) which looks like could cover all our needs. /cc @drehak @pirat89 

Coverage report:

```bash
----------- coverage: platform linux, python 3.8.2-final-0 -----------
Name                                        Stmts   Miss  Cover
---------------------------------------------------------------
tests/test_unit_sctpconfigread_sctpdlm.py      22      0   100%
tests/test_unit_sctpconfigread_sctplib.py      56      0   100%
---------------------------------------------------------------
TOTAL                                          78      0   100%
```

/cc @bocekm 

Update:
- agreed with @pirat89 to Merge it after PR #489 /cc @drehak

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/sctpconfigread/libraries/sctpdlm.py
+++ b/repos/system_upgrade/el7toel8/actors/sctpconfigread/libraries/sctpdlm.py
@@ -7,12 +7,15 @@ from leapp.libraries.stdlib import api
 from leapp.libraries.common import utils
 
 
-def check_dlm_cfgfile():
-    """Parse DLM config file"""
-    fname = "/etc/dlm/dlm.conf"
+def check_dlm_cfgfile(_open=open):
+    """Parse DLM config file.
+    :param _open: object behind opening a file. Might be replaced
+        by mocked one for the purpose of testing
+    """
+    fname = '/etc/dlm/dlm.conf'
 
     try:
-        with open(fname, 'r') as fp:
+        with _open(fname, 'r') as fp:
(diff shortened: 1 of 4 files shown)


ZhukovGreen:
Hurray! tests passed finally

ZhukovGreen:
@pirat89 
- fixed quotes to use single qoutes
- simplified commit history

bocekm:
@ZhukovGreen, out of curiosity I've looked at the use of Chore in git commits and I found this: https://karma-runner.github.io/0.10/dev/git-commit-msg.html. If it's what you follow, should this PR be of type Test instead of Chore? :)

ZhukovGreen:
@bocekm you're right. I was not exactly followed those framework, but rather the one I get used to. But generally, this is very useful convention. First if you see chore, you  already know that the introduced change shouldn't affect the application (that's why I had chore/tests or chore/docs).
As a result you could see branches strctured in more systematic way, i.e. (what we have now)
image

PR #496, opened 2020-04-29
Add msgs arg to CurrentActorMocked

ZhukovGreen:
A fix to the MR #489 which is already merged. The reason is that the new mock Actor object is not tested ref #497 
This allows defining messages which will be filtered against model inside consume method


pirat89:
for me it's mystery how did part dissapeared in the orig PR, as I saw it there several times. thank for catch.

PR #501, opened 2020-05-13
Chore/improve/allow running tests on fedora

ZhukovGreen:
After this MR, it is possible to run tests directly on Fedora OS. 

- [x] Squash commits after review 
- [x] Decide should we add this functionality test to travis?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,6 +83,8 @@ help:
 	@echo "                         - can be changed by the COPR_CONFIG env"
 	@echo "  install-deps           create python virtualenv and install there"
 	@echo "                         leapp-repository with dependencies"
+	@echo "  install-deps-fedora    create python virtualenv and install there"
+	@echo "                         leapp-repository with dependencies for Fedora OS"
 	@echo "  lint                   lint source code"
 	@echo "  test                   lint source code and run tests"
 	@echo "  test_no_lint           run tests without linting the source code"
@@ -199,6 +201,21 @@ install-deps:
 	pip install --upgrade -r requirements.txt
 	python utils/install_actor_deps.py --actor=$(ACTOR)
 
+install-deps-fedora:
+	@# Check the necessary rpms are installed for py3 (and py2 below)
+	if ! rpm -q git findutils python3-virtualenv; then \
+		if ! dnf install -y git findutils python3-virtualenv; then \
+			echo 'Please install the following rpms via the command: ' \
+				'sudo dnf install -y git findutils python3-virtualenv'; \
+			exit 1; \
+		fi; \
+	fi
+	@# Prepare the virtual environment
+	virtualenv --system-site-packages --python /usr/bin/python tut
+	. tut/bin/activate ; \
+	pip install --upgrade setuptools; \
+	pip install --upgrade -r requirements.txt; \
+
 lint:
 	. tut/bin/activate; \
 	echo "--- Linting ... ---" && \
@@ -229,4 +246,4 @@ dashboard_data:
 	python ../../../utils/dashboard-json-dump.py > ../../../discover.json; \
 	popd
 
-.PHONY: help build clean prepare source srpm copr_build print_release register install-deps lint test_no_lint test dashboard_data
+.PHONY: help build clean prepare source srpm copr_build print_release register install-deps install-deps-fedora lint test_no_lint test dashboard_data
--- a/repos/system_upgrade/el7toel8/actors/rpmscanner/libraries/library.py
+++ b/repos/system_upgrade/el7toel8/actors/rpmscanner/libraries/library.py
@@ -1,11 +1,25 @@
+import warnings
+
 from leapp.exceptions import StopActorExecutionError
+
+no_yum = False
+no_yum_warning_msg = "package `yum` is unavailable"
+try:
+    import yum
+except ImportError:
+    no_yum = True
+    warnings.warn(no_yum_warning_msg, ImportWarning)
 
 
 def get_package_repository_data():
-    """ Return dictionary mapping package name with repository from which it was installed """
-    # import has to be inside the function to avoid troubles with non-existing
-    # module in Python3 (where we do not need this function anymore)
-    import yum  # pylint: disable=import-outside-toplevel
+    """ Return dictionary mapping package name with repository from which it was installed.
+    Note:
+        There's no yum module for py3. The dnf module could have been used
+        instead but there's a bug in dnf preventing us to do so:
+        https://bugzilla.redhat.com/show_bug.cgi?id=1789840
+    """
+    if no_yum:
+        raise StopActorExecutionError(message=no_yum_warning_msg)
     yum_base = yum.YumBase()
     pkg_repos = {}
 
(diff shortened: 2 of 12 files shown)


repos/system_upgrade/el7toel8/actors/rpmscanner/libraries/library.py
```diff
@@ -1,11 +1,25 @@
...
 
 def get_package_repository_data():
-    """ Return dictionary mapping package name with repository from which it was installed """
-    # import has to be inside the function to avoid troubles with non-existing
-    # module in Python3 (where we do not need this function anymore)
-    import yum  # pylint: disable=import-outside-toplevel
+    """ Return dictionary mapping package name with repository from which it was installed.
+    Note:
+        There's no yum module for py3. The dnf module could have been used
+        instead but there's abug in dnf preventing us to do so:
```
vojtechsokol, line 18:
```suggestion
        instead but there's a bug in dnf preventing us to do so:
```

repos/system_upgrade/el7toel8/actors/rpmscanner/tests/test_rpmscanner.py
```diff
@@ -5,10 +5,22 @@
...
 
-@pytest.mark.skipif(sys.version_info.major >= 3,
-                    reason="There's no yum module for py3. The dnf module could have been used instead but there's a"
-                           " bug in dnf preventing us to do so: https://bugzilla.redhat.com/show_bug.cgi?id=1789840")
+
+@pytest.mark.skipif(
+    sys.version_info.major >= 3,
+    reason=(
+        "There's no yum module for py3. The dnf module could have been used "
+        "instead but there's abug in dnf preventing us to do so: "
```
vojtechsokol, line 19:
```suggestion
        "instead but there's a bug in dnf preventing us to do so: "
```

ZhukovGreen:
@vojtechsokol thank you for the review

repos/system_upgrade/el7toel8/actors/rpmscanner/libraries/library.py
```diff
@@ -1,11 +1,25 @@
...
+        https://bugzilla.redhat.com/show_bug.cgi?id=1789840
+    """
+    if no_yum:
+        raise UsageError(message=no_yum_warning_msg)
```
pirat89, line 22: I guess this is not exception you want to use here. I think it's purely for the Leapp itself regarding problem with 'how leapp is executed from the commandline'. So I am not sure we should use it in this way. Maybe good time to document Leapp exceptions finally. In this case, use `StopActorExecutionError`.

setup.cfg
```diff
@@ -10,4 +10,4 @@ application-import-names = leapp
 # check only for correct group type and abc sorting without group
 import-order-style = pep8
 # default values + ignore strict alphabetical sorting within one module import
-ignore = E121,E123,E126,E133,E241,E242,E226,W503,W504,W505,I101
+ignore = E121,E123,E126,E133,E241,E242,E226,W503,W504,W505,I101,E741
```
pirat89, line 13: It depends. But in majority number of cases I will disagree. Using instead `line` instead of `l` is not big deal. As well, the code itself is really bad when you call `l` in block of code. The `line` seems much more clear so I prefer to drop this one change.

pirat89: Discussed in person. This error is raised on fedora but it is not raised in our CI (older version of linter in centos container, maybe?).. From that point, just create an issue, so we will fix it in future and put it to the ignore list for now.

ZhukovGreen: Done https://github.com/oamg/leapp-repository/issues/502 @pirat89

pirat89:
@ZhukovGreen not working for me. First, I am not able to run tests, no linters because of missing deps, which I am missing - and I am not able to install on fedora, as required rpms are not supported anymore nor delivered anymore (e.g. the one for pyudev). To be able to install all deps, the pip installation of additional deps is needed - which should be covered by this PR in such case regarding description. In case that any other steps are needed, added please into the description.

pirat89:
Thinking about that, what about add something like `print-pip-deps` so we would be able to get all deps from various actors for pip. Benefits:
-  no need to require root on your system (fedora) when everything is installed through pip in the expected pyenv.
- can be handled from the main makefile easily as well

ZhukovGreen:
Hey @pirat89, now I think it could be really merged.
Tested on

```Dockerfile
FROM fedora:31 as base

FROM base as install-base-deps
RUN dnf update -y
RUN dnf install -y make

FROM install-base-deps as install-repo-deps
WORKDIR /leapp-repository
COPY ./requirements.txt .
COPY ./Makefile .
COPY ./utils/install_actor_deps.py ./utils/install_actor_deps.py
COPY ./repos ./repos
RUN PYTHON_VENV=python3 make install-deps-fedora
RUN dnf clean all

FROM install-repo-deps as copy-src
COPY . .
RUN source tut/bin/activate && snactor repo find --path .

ENTRYPOINT ["bash"]
```

Then just `make test`

Makefile
```diff
@@ -199,6 +199,13 @@ install-deps:
 	pip install --upgrade -r requirements.txt
 	python utils/install_actor_deps.py --actor=$(ACTOR)
 
+install-deps-fedora:
+	dnf install -y git findutils python3-virtualenv
```
pirat89, line 203: Expecting this command should never be executed under root on real system, it will be better to do it like that (so in case of CI, ... - it will be done under the root, and in other case, people will now what they should do):
```suggestion
    if ! rpm -q git findutils python3-virtualenv; then \
        if ! dnf install -y git findutils python3-virtualenv; then \
            echo 'Please install the following rpms via the command: ' \
                'sudo dnf install -y git findutils python3-virtualenv'; \
            exit 1; \
        fi; \
    fi
```

Makefile
```diff
@@ -199,6 +199,13 @@ install-deps:
...
+	virtualenv --system-site-packages --python /usr/bin/$(_PYTHON_VENV) tut
```
pirat89, line 204: Regarding the installation of rpms above, in case the `_PYTHON_VENV` is not set, the default system python should be used (Python 3.7?). But not hard requirement from me.

pirat89: But to be honest, I do not know know how to do it nicely without impact on other targets. So feel free to keep it like that, just document that properly. WDYT @bocekm ?

ZhukovGreen: This concern resolved here https://github.com/oamg/leapp-repository/pull/501/commits/6f447e4484942cd0b0dcee90a11e858fc063c00b

If no PYHTON_VENV or explictly poinitng to python2, then it installs python2

Makefile
```diff
@@ -199,6 +201,23 @@ install-deps:
 	pip install --upgrade -r requirements.txt
 	python utils/install_actor_deps.py --actor=$(ACTOR)
 
+install-deps-fedora:
+	@dnf install -y git findutils python3-virtualenv || ( \
+	echo 'Please install the following rpms via the command:' \
```
bocekm, line 206:
```
$ make install-deps-fedora 
Error: This command has to be run with superuser privileges (under the root user on most systems).
Please install the following rpms via the command: sudo dnf install -y git findutils python3-virtualenv
make: *** [Makefile:205: install-deps-fedora] Error 1
```
Do not require root.

pirat89: It should be ok now.

ZhukovGreen: Should work now

pirat89:
:champagne:

PR #508, opened 2020-06-08
[DO NOT MERGE] - Upgrades on AWS POC

Rezney:
Let's discuss this POC approach. This is an MVP, we could probably use repomapping too as the repoids should not change and are different from the "base" ones.

Leapp RHUI package will be most probably maintained by RHUI team.

I am going to explain this further on tomorrow's meeting.

### DIFF
--- a/repos/system_upgrade/el7toel8/libraries/dnfplugin.py
+++ b/repos/system_upgrade/el7toel8/libraries/dnfplugin.py
@@ -7,9 +7,12 @@ import shutil
 from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.common import guards, mounting, overlaygen, rhsm, utils
 from leapp.libraries.stdlib import CalledProcessError, api, config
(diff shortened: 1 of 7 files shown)


repos/system_upgrade/el7toel8/libraries/dnfplugin.py
```diff
@@ -109,6 +112,8 @@ def _transaction(context, stage, target_repoids, tasks, test=False, cmd_prefix=N
             cmd.append('-v')
         if rhsm.skip_rhsm():
             cmd += ['--disableplugin', 'subscription-manager']
+            if stage == 'upgrade':
+                cmd += ['--disableplugin', 'amazon-id']
```
Rezney, line 116: of course we want to disable the Amazon plugin only if we are in cloud

Rezney:
Closing in favor of https://github.com/oamg/leapp-repository/pull/539

PR #513, opened 2020-06-17
disable grubdevname for s390x

ZhukovGreen:
ref JIRA OAMG-3131

This MR introduces two things:
1. Decorator skip_actor_execution_if(*conditions) for any Actors classes, which allows to skip the actor execution
2. Skip execution of grubdevname actor for s390x cpu architecture with the help of introduced decorator

During the work there were one enhancement ticket created:
https://github.com/oamg/leapp-repository/issues/514

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/grubdevname/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/grubdevname/actor.py
@@ -1,9 +1,14 @@
+import platform
+
 from leapp.actors import Actor
 from leapp.libraries.actor.grubdevname import get_grub_device
+from leapp.libraries.common.config import architecture
+from leapp.libraries.common.utils import skip_actor_execution_if
 from leapp.models import GrubDevice
 from leapp.tags import FactsPhaseTag, IPUWorkflowTag
 
 
+@skip_actor_execution_if(platform.machine() == architecture.ARCH_S390X)
 class Grubdevname(Actor):
     """
     Get name of block device where GRUB is located
--- a/repos/system_upgrade/el7toel8/libraries/utils.py
+++ b/repos/system_upgrade/el7toel8/libraries/utils.py
@@ -3,6 +3,7 @@ import sys
 
 import six
 
+from leapp.actors import Actor
 from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.common import mounting
 from leapp.libraries.stdlib import STDOUT, CalledProcessError, api, config, run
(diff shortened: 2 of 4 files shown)


ZhukovGreen:
review please

ZhukovGreen:
e2e tests

ZhukovGreen:
e2e tests

repos/system_upgrade/el7toel8/libraries/utils.py
```diff
@@ -168,3 +169,47 @@ def read_file(path):
     """
     with open(path, 'r') as f:
         return f.read()
+
+
+def _blank_fn(self, *args, **kwargs):
+    return None
+
+
+def skip_actor_execution_if(*conditions):
```
pirat89, line 178: I think that function (or decorator) like that should be something what is implemented inside the framework. Better to say, that we discussed existence of a method for Actor class like `is_executable` (I do not remember the name, @vinzenz ?). Instead of this, I prefer just to add the condition into the process function of the actor now as it is used on several places already to not complicate it, like:
```
def process():
        return
```

Another thing is, that if we decide that the functionality will be part of leapp framework, we will require to use it instead of this decorator and in such a case, people will have additional work to do (as we deprecate that). WDYT @vinzenz?

repos/system_upgrade/el7toel8/actors/grubdevname/actor.py
```diff
@@ -1,9 +1,14 @@
...
+
 from leapp.actors import Actor
 from leapp.libraries.actor.grubdevname import get_grub_device
+from leapp.libraries.common.config import architecture
+from leapp.libraries.common.utils import skip_actor_execution_if
 from leapp.models import GrubDevice
 from leapp.tags import FactsPhaseTag, IPUWorkflowTag
 
 
+@skip_actor_execution_if(platform.machine() == architecture.ARCH_S390X)
```
pirat89, line 11: Read my commet about the decorator. Anyway, you should not use `platform.machine` anywhere. Use the configuration function: `architecture.matches_architecture(architecture.ARCH_S390X)`

repos/system_upgrade/el7toel8/libraries/utils.py
ZhukovGreen: I was thinking about the place to introduce this decorator. I think it shouldn’t be in the Leapp framework. It looks like serves specific needs of IPU only

vinzenz: > I was thinking about the place to introduce this decorator. I think it shouldn’t be in the Leapp framework. It looks like serves specific needs of IPU only

I tend to disagree with you here. There is nothing IPU specific about being able to skip actors.

ZhukovGreen: OK, then I think the easiest way to solve as proposed by @pirat89 .

ZhukovGreen: I will close the MR and do this in a new one

ZhukovGreen:
closed in favor of https://github.com/oamg/leapp-repository/pull/515

PR #518, opened 2020-06-26
Add tests for PersistentNetNames and PersistentNetNamesInitramfs

drehak:
Both actors are almost identical except for the name and phase.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/persistentnetnames/tests/test_persistentnetnames.py
+++ b/repos/system_upgrade/el7toel8/actors/persistentnetnames/tests/test_persistentnetnames.py
@@ -1,0 +1,37 @@
+import pytest
+
+from leapp.libraries.common import persistentnetnames
+from leapp.models import Interface, PCIAddress, PersistentNetNamesFacts
+from leapp.snactor.fixture import current_actor_context
+
+
+def interface_mocked(i=0):
+    return Interface(
+        name='n{}'.format(i),
+        devpath='dp{}'.format(i),
+        driver='d{}'.format(i),
+        vendor='v{}'.format(i),
+        pci_info=PCIAddress(
+            domain='pd{}'.format(i),
+            bus='pb{}'.format(i),
+            function='pf{}'.format(i),
+            device='pd{}'.format(i)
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/el7toel8/actors/persistentnetnames/tests/test_persistentnetnames.py
```diff
@@ -0,0 +1,37 @@
+import pytest
+
+from leapp.libraries.common import persistentnetnames
+from leapp.models import Interface, PCIAddress, PersistentNetNamesFacts
+from leapp.snactor.fixture import current_actor_context
+
+
+def interface_mocked(i=0):
+    return Interface(
+        name='n'+str(i),
```
ZhukovGreen, line 10: Just out of curiosity compared with the way I would do that and this is the result:
```python

In [2]: %timeit "n{}".format(1)                                                                                         
156 ns ± 0.644 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [3]: %timeit "n" + str(1)                                                                                            
168 ns ± 0.253 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```
** this is not th requirement to change **

drehak: No, this is a good point. Thanks!

drehak:
Am I doing this right?
```
---------- coverage: platform linux2, python 2.7.18-final-0 ----------
Name                                                                           Stmts   Miss  Cover
--------------------------------------------------------------------------------------------------
actors/persistentnetnames/actor.py                                                11      0   100%
actors/persistentnetnames/tests/test_persistentnetnames.py                        16      0   100%
actors/persistentnetnamesinitramfs/actor.py                                       11      0   100%
actors/persistentnetnamesinitramfs/tests/test_persistentnetnamesinitramfs.py      16      0   100%
--------------------------------------------------------------------------------------------------
TOTAL                                                                             54      0   100%
```

drehak:
@ZhukovGreen Thanks! :)

PR #519, opened 2020-06-30
Improve messages about transaction changes

ZhukovGreen:
Related: OAMG-888
Rephrase the reporting content about skipped packages when they use unsupported repositories/packages. Also, apply renaming to prohibited stuff 

# TODO
- [x] rename the word backlisted into something more relevant
- [x] remove TODO note after confirmation from Michal
- [x] Squash some commits
- [x] Add functionality do not exclude repositories when enabled via `--enablerepo`
- [x]  Apply deprecation proccess for old models 

# Example report with --enablerepo
```
Risk Factor: medium 
Title: Added not supported by Red Hat repository 
Summary: The following repository added to the upgrade, but it is not supported by the Red Hat.
- codeready-builder-for-rhel-8-x86_64-rpms
...
```
# Example report without --enablerepo
```
Risk Factor: info 
Title: Excluded RHEL 8 repositories 
Summary: The following repositories are not supported by Red Hat and are excluded from the list of repositories used during the upgrade. 
- codeready-builder-for-rhel-8-x86_64-rpms 
For more information, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/package_manifest/codereadylinuxbuilder-repository. 
Remediation: [hint] At your own risk! Use leapp upgrade --enablerepo codeready-builder-for-rhel-8-x86_64-rpms 
Risk Factor: high 
Title: Packages will not be installed 
Summary: 4 packages will not be installed or upgraded due to repositories unknown to leapp: 
- libselinux-python3 
- postfix-cdb - postfix-pcre - postfix-sqlite Remediation: [hint] Please file a bug in http://bugzilla.redhat.com/ for leapp-repository component ofthe Red Hat Enterprise Linux 7 product. 
Risk Factor: high 
Title: Packages will not be installed 
Summary: 2 packages will be skipped because they are available only in RHEL 8 repositories that are intentionally excluded from the list of repositories used during the upgrade. See the report message titled "Excluded RHEL 8 repositories" for details. 
 The list of these packages (repositories: codeready-builder-for-rhel-8-x86_64-rpms): 
- python3-pyxattr 
- rpcgen 
Remediation: [hint] At your own risk!  Use leapp upgrade --enablerepo codeready-builder-for-rhel-8-x86_64-rpms 
```


pirat89:
@ZhukovGreen friendly reminder that you should write here all related information for the PR. e.g. what was the issue of the previous solution and why we are changing it. point to the JIRA task is good for our team, but not enough for others.

ZhukovGreen:
Sure! @pirat89 as far as I will get the whole scope enough to be able to formalize that, I will add the description:)))

ZhukovGreen:
@bocekm Michal, the MR is still in draft, but, in terms of rephrasing of the message, could you please glance on it. THe CI passes now, but due to some magic not appeared MR checks (https://travis-ci.org/github/oamg/leapp-repository/builds/705819865)

ZhukovGreen:
@bocekm marking it as ready for a review. Please take a look

ZhukovGreen:
@bocekm Hi Michal, pls review. The also added a snippet from the leapp-report

repos/system_upgrade/el7toel8/actors/repositoriesexcluded/actor.py
```diff
@@ -0,0 +1,23 @@
...
+from leapp.reporting import Report
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+
+
+class RepositoriesExcludedList(Actor):
+    """
+    Generate list of repository IDs that should be ignored by Leapp during upgrade process
+    """
+
+    name = 'repositories_excluded_list'
```
bocekm, line 17:
```suggestion
    name = 'exclude_repositories'
```
The original name (repositories_blacklist) was not good at all. Because most of the actors are named using a verb (e.g. check) or noun explaining some kind of action (e.g. scanner), I think we can name it even better now. I'd go with the verb here.

repos/system_upgrade/el7toel8/actors/repositoriesexcluded/libraries/repositoriesexcluded.py
```diff
@@ -0,0 +1,82 @@
...
+            reposid_excluded_list,
+        )
+        api.produce(RepositoriesExcluded(repoids=reposid_excluded_list))
+
+        report = [
+            reporting.Title("Excluded RHEL 8 repositories"),
+            reporting.Summary(
+                "The following repositories\n- {}\n are not supported by the "
+                "Red Hat and should be excluded from the list of repositories "
+                "used during the upgrade.".format(
```
bocekm, line 64:
```suggestion
               "The following repositories are not supported by "
                "Red Hat and are excluded from the list of repositories "
                "used during the upgrade:\n- {}.".format(
```

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/peseventsscanner.py
```diff
@@ -407,27 +407,43 @@ def filter_out_out_pkgs(event_in_pkgs, event_out_pkgs):
...
-    if blacklisted_pkgs:
-        report_skipped_packages('packages will not be installed due to blacklisted repositories:',
-                                blacklisted_pkgs)
+    if excluded_pkgs:
+        report_skipped_packages(
+            message=SKIPPED_PKGS_MSG.format(', '.join(excluded_repos)),
+            packages=excluded_pkgs,
+            remediation=(
+                'AT YOUR OWN RISK!!! Use leapp upgrade {}'.format(' '.join(
+                    ['--enablerepo ' + repo for repo in excluded_repos])
```
bocekm, line 443: Have you verified that this really works as expected? I.e. that the list of excluded repositories is not used in any actor?

ZhukovGreen: @bocekm thanks for pointing out. I took it from somewhere (already not remember) as a fact. After testing this option, it reveals as not working at all. So the fix implemented now

repos/system_upgrade/el7toel8/actors/peseventsscanner/actor.py
```diff
@@ -21,7 +21,7 @@ class PesEventsScanner(Actor):
     """
 
     name = 'pes_events_scanner'
-    consumes = (InstalledRedHatSignedRPM, RepositoriesBlacklisted, RepositoriesMap, RpmTransactionTasks)
+    consumes = (InstalledRedHatSignedRPM, RepositoriesExcluded, RepositoriesMap, RpmTransactionTasks)
```
pirat89, line 24: @ZhukovGreen this is against our policy. Follow the deprecation process. You can add the RepositoriesExcluded but you have to keep support of RepositoriesBlacklisted and set it as deprecated.

ZhukovGreen: Thanks for pointing out. I will adopt the deprecation process. @pirat89

ZhukovGreen:
@bocekm @pirat89 guys, I deprecated the old model, please review

ZhukovGreen:
e2e tests

ZhukovGreen:
@AloisMahdal pls help to troubleshoot murphy-ci . I lost in logs:)

repos/system_upgrade/el7toel8/actors/repositoriesexcluded/libraries/repositoriesexcluded.py
```diff
@@ -0,0 +1,110 @@
...
+    suffix = "optional-rpms"
+    if sys_type != "ga":
+        suffix = "optional-{}-rpms".format(sys_type)
+    for enabled_repo in custom_target_repos:
+        if enabled_repo.repoid == repo.to_repoid:
+            if repo.to_repoid not in already_reported_skips:
+                already_reported_skips.append(repo.to_repoid)
+                api.current_logger().info(
+                    "Skipping %s repository from exclusion due to directly "
+                    "specified in --enablerepo flag"
```
bocekm, line 26:
```suggestion
                    "Skipping %s repository from exclusion as it is directly "
                    "specified in the --enablerepo option"
```

repos/system_upgrade/el7toel8/actors/repositoriesexcluded/libraries/repositoriesexcluded.py
```diff
@@ -0,0 +1,110 @@
...
+                )
+                report = [
+                    reporting.Title(
+                        "Added not supported by Red Hat repository"
+                    ),
+                    reporting.Summary(
+                        "The following repository added to the upgrade, "
+                        "but it is not supported by the Red Hat.:\n- {}".format(
```
bocekm, line 34:
```suggestion
                        "The following repository has been used for the upgrade, "
                        "but it is not supported by Red Hat.:\n- {}\nFor more information, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/package_manifest/codereadylinuxbuilder-repository.".format(
```

repos/system_upgrade/el7toel8/actors/repositoriesexcluded/libraries/repositoriesexcluded.py
```diff
@@ -0,0 +1,110 @@
...
+                            enabled_repo.repoid
+                        )
+                    ),
+                    reporting.Severity(reporting.Severity.MEDIUM),
+                    reporting.Tags([reporting.Tags.REPOSITORY]),
+                ]
+                reporting.create_report(report)
```
bocekm, line 41: Frankly, I don't like adding this whole reporting feature to the `_is_optional_repo()` function because then the name of the function does not fit what it is doing.

Please create a new function like:
```
def _report_using_unsupported_repo(opt_repo_mapping):
    custom_target_repos = tuple(api.consume(CustomTargetRepository))
    for repo_to_enable in custom_target_repos:
        if repo.to_repoid in opt_repo_mapping.values():
            <create_report>
```
and call it from the `process()`, like:
```
opt_repo_mapping = _get_optional_repo_mapping() # See my other comment below on this function
_report_using_unsupported_repo(opt_repo_mapping)
```

repos/system_upgrade/el7toel8/actors/repositoriesexcluded/libraries/repositoriesexcluded.py
```diff
@@ -0,0 +1,110 @@
...
+    if repo_map:
+        for repo in repo_map.repositories:
+            if _is_optional_repo(repo, custom_target_repos):
+                opt_repo[repo.from_repoid] = repo.to_repoid
+    return opt_repo
+
+
+def _get_disabled_optional_repo():
+    """
+    Return a list of disabled optional repositories available on the system.
```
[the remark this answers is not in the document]
ZhukovGreen:
```
    """Get a list of repoids do not use during the upgrade.

    These are such RHEL 8 repositories that mapped to disabled RHEL7 Optional
    repositories.
    """
```

repos/system_upgrade/el7toel8/actors/repositoriesexcluded/actor.py
```diff
@@ -0,0 +1,24 @@
+from leapp.actors import Actor
```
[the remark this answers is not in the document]
ZhukovGreen: Thanks, this makes sense

repos/system_upgrade/el7toel8/actors/repositoriesexcluded/libraries/repositoriesexcluded.py
ZhukovGreen: Agree, sounds better

repos/system_upgrade/el7toel8/actors/repositoriesexcluded/libraries/repositoriesexcluded.py
ZhukovGreen: After refactoring this message no longer needed

repos/system_upgrade/el7toel8/actors/repositoriesexcluded/libraries/repositoriesexcluded.py
```diff
@@ -0,0 +1,110 @@
...
+                "Red Hat and are excluded from the list of repositories "
+                "used during the upgrade.\n- {}".format(
+                    "\n- ".join(reposid_excluded_list)
+                )
+            ),
+            reporting.Severity(reporting.Severity.INFO),
+            reporting.Tags([reporting.Tags.REPOSITORY]),
+            reporting.Flags([reporting.Flags.FAILURE]),
+            reporting.Remediation(
+                hint="AT YOUR OWN RISK!!! Use leapp upgrade {}".format(
```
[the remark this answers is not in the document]
ZhukovGreen: I refactored it to the proposed edition.

ZhukovGreen:
e2e tests

ZhukovGreen:
e2e tests

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/peseventsscanner.py
```diff
@@ -475,44 +478,55 @@ def map_repositories(packages):
...
+    Filter out those PES events that conflict with the higher priority transaction configuration files.
 
     Output packages from an event are added to packages for removal only if all input packages are already there.
 
     :param transaction_configuration: RpmTransactionTasks model instance with pkgs to install, keep and remove based
                                       on the user configuration files
     :param events: List of Event tuples, where each event contains event type and input/output pkgs
     """
-    message = 'Marking packages for removal:\n'
+    message = 'The following RHEL 8 packages will not be installed:\n'
```
AloisMahdal, line 528: Mentioning the target distro by name means the text need to become dynamic once we spit off to el8toel9.

pirat89: Not so big problem. It's highly possible the actor will have to be rewritten because of modularity anyway in the future.

requirements.txt
```diff
@@ -5,9 +5,7 @@ flake8-import-order
 funcsigs==1.0.2
 mock==2.0.0
 pylint
-# pytest is fixed at 3.6.4 because never version (3.7.*) requires scandir, which depends on gcc
-pytest==3.6.4
-pytest-ordering==0.5
+pytest==4.6.11
```
pirat89, line 8: What is the reasoning for this change? The change is not documented anywhere and regarding the previous comment, I expect the 3.6.4 version should be kept. If there is not real reasoning for this, the old pytest should be kept.

ZhukovGreen: There is a reason. But it was hard to find:) In pytest 3.6 the caplog implementation doesn't have attribute messages (see https://travis-ci.org/github/oamg/leapp-repository/jobs/705807738#L4052). So I bumped the pytest to a higher version. Moreover, we are no longer restricted by the gcc, since we added it to the dependencies

repos/system_upgrade/el7toel8/actors/repositoriesexcluded/libraries/repositoriesexcluded.py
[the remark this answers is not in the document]
lspackova: @bocekm sorry for the delayed response (I'm on PTO). I second your concern. I suggest we use admonitions (WARNING, IMPORTANT, NOTE - ordered by severity) - for which we can use upper case, and then follow with the text. In this case, I suggest something like:
WARNING: Use at your own risk.

requirements.txt
pirat89: OK. So please document it. Nor the commit, nor the file have a comment about the change to make it clear for the others.

repos/system_upgrade/el7toel8/actors/peseventsscanner/actor.py
pirat89: @ZhukovGreen I am still missting the `RepositoriesBlacklisted` in the consumed set. Deprecation policy tells, you have to keep the deprecated functionality until removed.

pirat89:
@ZhukovGreen https://github.com/pirat89/leapp/tree/improve_deprecation_doc - I have started to improve the doc for deprecation process. If you need more guidance, it's good to look at the last commit

repos/system_upgrade/el7toel8/actors/peseventsscanner/actor.py
bocekm: I don't think *our* actor needs to consume it. That's not how I understand deprecation. If we deprecate a model, it's definition still needs to exist (of course), we still need to produce the messages of this model. Nothing more.

pirat89: @bocekm they have to, otherwise it is not deprecated model and we could remove it.

bocekm: You're right. If any third party actor produces the message through which they want to blacklist a repository and if we stop consuming/processing the message, we would stop excluding such their repository.

repos/system_upgrade/el7toel8/actors/excluderepositories/libraries/excluderepositories.py
```diff
@@ -0,0 +1,137 @@
...
+
+        report = [
+            reporting.Title("Excluded RHEL 8 repositories"),
+            reporting.Summary(
+                "The following repositories are not supported by "
+                "Red Hat and are excluded from the list of repositories "
+                "used during the upgrade.\n- {}\nFor more information, see "
+                "https://access.redhat.com/documentation/en-us/"
+                "red_hat_enterprise_linux/8/html/package_manifest/"
+                "codereadylinuxbuilder-repository.".format(
```
pirat89, line 122: Use `reporting.ExternalLink` instead

repos/system_upgrade/el7toel8/actors/excluderepositories/libraries/excluderepositories.py
```diff
@@ -0,0 +1,137 @@
...
+
+def _report_using_unsupported_repos(repo_ids):
+    if manually_enabled_repos_in_use:
+        report = [
+            reporting.Title("Using repository not supported by Red Hat"),
+            reporting.Summary(
+                "The following repository has been used for the "
+                "upgrade, but it is not supported by the Red Hat.:\n"
+                "- {}\nFor more information, see "
+                "https://access.redhat.com/documentation/en-us/"
```
pirat89, line 22: And here the external link should be used as well.

ZhukovGreen:
closing in favour #543 #545 and #546

PR #520, opened 2020-07-03
Make sure "default.target.wants" dir exists

Rezney:
In case nothing is enabled in the default target, the directory
does not exist yet.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/createresumeservice/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/createresumeservice/actor.py
@@ -1,6 +1,7 @@
 import shutil
 import os
 
+from leapp.exceptions import StopActorExecutionError
 from leapp.actors import Actor
 from leapp.tags import FinalizationPhaseTag, IPUWorkflowTag
 from leapp.reporting import Report, create_report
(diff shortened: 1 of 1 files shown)


Rezney:
@pirat89 rebased

bocekm:
@pirat89, @Rezney, how critical is this. Do you want it in the next release? Is there a related Jira task?

pirat89:
@bocekm  without this fix, machines with the missing directory will not finish the whole upgrade correctly, as the last phase will not be executed because of the crash. Despite that fact, it's not critical for us now, but who knows

Rezney:
@bocekm @pirat89 ticket #3537 (there is also subtask for QE). Well, actually I was kind of surprised that we did not hit it yet (or got reported). But otherwise, IIRC we were not able to reach the machine when we hit this (either due to the selinux context not relabelled or crash into emergency console) so severity is critical but likelihood is probably much lower.

pirat89 (dismissed):
LGTM. It's only about to get waive from @oamg/qe.

pirat89:
e2e tests

Rezney:
@AloisMahdal not sure how far you are with the QE task but this PR is blocking our public clouds work. Could this be merged as it is not going to the current release anyway?

pirat89 (approved):

drehak:
Manually reproduced and verified.

PR #521, opened 2020-07-03
WIP: bind configuration check

pemensik:
Bind configuration check, currently only checking whether problem is found.
It reports only errors in configuration in facts now.


repos/system_upgrade/el7toel8/actors/bindupdate/actor.py
```diff
@@ -0,0 +1,34 @@
...
+        return False
+
+    def process(self):
+        if not self.has_package(InstalledRedHatSignedRPM):
+            self.log.debug('bind is not installed')
+            return
+
+        for bindfacts in self.consume(BindFacts):
+            modified = updates.update_facts(bindfacts)
+            self.log.info('{0} BIND configuration files modified.'.format(modified))
```
ZhukovGreen, line 34: Please use logger specific string formatting (https://docs.python.org/3/howto/logging-cookbook.html#use-of-alternative-formatting-styles)
```
>>> logger.error('This is an%s %s %s', 'other,', 'ERROR,', 'message')
2010-10-28 15:19:29,833 foo.bar ERROR This is another, ERROR, message
>>>
```

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
```diff
@@ -0,0 +1,86 @@
...
+from leapp.libraries.stdlib import run
+
+# Callback for walk function
+callbacks = {
+    'dnssec-lookaside': isccfg.ModifyState.callback_comment_out,
+}
+
+
+def paths_from_issues(issues):
+    """ Extract paths from list of BindConfigIssuesModel """
```
ZhukovGreen, line 11: docstrings should follow PEP-257

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
```diff
@@ -0,0 +1,86 @@
...
+
+def parser_file(parser, path):
+    for cfg in parser.FILES_TO_CHECK:
+        if cfg.path == path:
+            return cfg
+    return None
+
+
+def debug_log(log, text):
+    if log is not None:
```
ZhukovGreen, line 26: just ```if log:```?

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
```diff
@@ -0,0 +1,86 @@
...
+        log.debug(text)
+
+
+def make_backup(path, backup_suffix='.leapp'):
+    """ Make backup of a file before modification """
```
ZhukovGreen, line 31: docstrings formatting

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
```diff
@@ -0,0 +1,86 @@
...
+    backup_path = path + backup_suffix
+    run(['cp', '--preserve=all', path, backup_path])
+
+
+def update_section(parser, section):
+    """ Modify one section
+    :ptype section: ConfigSection
```
ZhukovGreen, line 38: `ptype` - I've never seen this before. Also googling didn't help. Maye you meant just `:type section: ...`

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
```diff
@@ -0,0 +1,86 @@
...
+
+def update_config(parser, cfg):
+    """ Modify contents of file accoriding to rules
+    :ptype cfg: ConfigFile
+    :returns str: Modified config contents
+    """
+    return update_section(parser, cfg.root_section())
+
+
+def update_file(parser, path, log=None, write=True):
```
ZhukovGreen, line 54: what is the purpose of having a log as an argument?

ZhukovGreen: typical way of logging within this repo is to use api.current_logger() logger obj. It does some logger naming magic for you.

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
```diff
@@ -0,0 +1,86 @@
...
+    """ Prepare modified content for the file, make backup and rewrite it
+    :param parser: IscConfigParser
+    :param path: String with path to a file
+    :param log: Log instance with debug(str) method or None
+    :param write: True to allow file modification, false to only return modification status
+    """
+    cfg = parser_file(parser, path)
+    modified = update_config(parser, cfg)
+    if modified != cfg.buffer:
+        debug_log(log, '{0} needs modification'.format(path))
```
ZhukovGreen, line 64: use logger specific str formatting

ZhukovGreen: btw the way the most useful article for me explaining why it should be passed to the logger instead of str is here https://reinout.vanrees.org/weblog/2015/06/05/logging-formatting.html

repos/system_upgrade/el7toel8/actors/checkbind/actor.py
```diff
@@ -0,0 +1,49 @@
...
+        if not self.has_package(InstalledRedHatSignedRPM):
+            self.log.debug('bind is not installed')
+            return
+
+        facts = model.get_facts('/etc/named.conf')
+        issues = model.get_messages(facts)
+
+        if issues is not None:
+            api.produce(facts)
+            issues.extend([
```
ZhukovGreen, line 39: issues, I suppose, is not that good naming as you're extending it with the reporting attributes.

repos/system_upgrade/el7toel8/actors/checkbind/actor.py
```diff
@@ -0,0 +1,49 @@
...
+                reporting.Severity(reporting.Severity.HIGH),
+                reporting.Tags([reporting.Tags.SERVICES, reporting.Tags.NETWORK]),
+                reporting.Flags([reporting.Flags.INHIBITOR]),
+            ])
+            reporting.create_report(issues)
+            self.log.info('BIND configuration issues were found.')
+        else:
+            self.log.info('BIND configuration seems compatible.')
+
+        pass
```
ZhukovGreen, line 49: old artefact?

repos/system_upgrade/el7toel8/actors/checkbind/libraries/model.py
```diff
@@ -0,0 +1,92 @@
...
+
+        arg = statement.var(1)
+        if arg.type() == arg.TYPE_BARE and arg.value() in ['auto', 'yes']:
+            # auto or yes statement
+            add_statement(statement, state)
+        # dnssec-lookaside "." trust-anchor "dlv.isc.org";
+        elif (arg.type() == arg.TYPE_QSTRING and arg.value() == '"."'
+              and statement.var(2).value() == 'trust-anchor'
+              and statement.var(3).invalue() == 'dlv.isc.org'):
+            add_statement(statement, state)
```
ZhukovGreen, line 29: seems too big clause under the try. Consider use try-except-finally structure

repos/system_upgrade/el7toel8/actors/checkbind/libraries/model.py
```diff
@@ -0,0 +1,92 @@
...
+    except IndexError:
+        pass
```
ZhukovGreen, line 31: maybe some log message?

repos/system_upgrade/el7toel8/actors/checkbind/libraries/model.py
```diff
@@ -0,0 +1,92 @@
...
+    return facts
+
+
+def get_messages(facts):
+    if facts.dnssec_lookaside:
+        return [
+            reporting.Title('BIND configuration issues found'),
+            reporting.Summary('BIND configuration contains no longer accepted statements: dnssec-lookaside')
+                ]
+    return None
```
ZhukovGreen, line 92: No need in this line
```
In [1]: def foo():
   ...:     """"""
   ...: 

In [2]: foo()

In [3]: foo() is None
Out[3]: True
```

repos/system_upgrade/el7toel8/libraries/bindparser.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+
+        views = {}
+
+        for cfg in self.FILES_TO_CHECK:
+            v = self.find_views_file(cfg)
+            views.update(v)
+        return views
+
+    pass
```
ZhukovGreen, line 63: drop?

repos/system_upgrade/el7toel8/libraries/isccfg.py
```diff
@@ -0,0 +1,959 @@
...
+
+    def __str__(self):
+        return self.buffer
+
+    def __repr__(self):
+        return 'ConfigFile {0} ({1})'.format(
+                self.path, self.buffer)
+
+    def load(self, path):
+        with open(path, 'r') as f:
```
ZhukovGreen, line 38: 'r' is default and may be ommited

repos/system_upgrade/el7toel8/libraries/isccfg.py
```diff
@@ -0,0 +1,959 @@
...
+            self.buffer = self.original = f.read()
+            f.close()
```
ZhukovGreen, line 40: with open context closes the file for you, no need to do it manually

repos/system_upgrade/el7toel8/libraries/isccfg.py
```diff
@@ -0,0 +1,959 @@
...
+        :type config: ConfigFile
+        :param kind: type of this section
+        """
+        self.config = config
+        self.name = name
+        self.start = start
+        self.end = end
+        self.ctext = self.original_value()   # a copy for modification
+        self.parser = parser
+        if kind is None:
```
ZhukovGreen, line 86: maybe `if not kind`?

repos/system_upgrade/el7toel8/libraries/isccfg.py
```diff
@@ -0,0 +1,959 @@
...
+
+    def children(self, comments=False):
+        """
+        Return list of items inside this section
+        """
+        start = self.start
+        if self.type() == self.TYPE_BLOCK:
+            start += 1
+        return list(IscIterator(self.parser, self, comments, start))
+    pass
```
ZhukovGreen, line 137: drop?

repos/system_upgrade/el7toel8/libraries/isccfg.py
```diff
@@ -0,0 +1,959 @@
...
+            statement = next(self.iter)
+            while statement:
+                vl.append(statement)
+                if self.parser.is_terminal(statement):
+                    return ConfigVariableSection(vl, None, parent=self.section)
+                statement = next(self.iter)
+        except StopIteration:
+            if vl:
+                return ConfigVariableSection(vl, None, parent=self.section)
+        raise StopIteration
```
ZhukovGreen, line 233:
```suggestion
                return ConfigVariableSection(vl, None, parent=self.section)
            raise
```

repos/system_upgrade/el7toel8/libraries/tests/test_isccfg.py
```diff
@@ -0,0 +1,317 @@
+#!/usr/bin/env python
+#
+# Tests for bind configuration parsing
+
+try:
+    from leapp.libraries.common import isccfg
+except ImportError:
+    # For easier testing, allow having it out-of-directory
+    import isccfg
```
ZhukovGreen, line 9: I don't understand it. Do you modify a system path?

repos/system_upgrade/el7toel8/actors/checkbind/libraries/model.py
```diff
@@ -1,19 +1,19 @@
...
     name = statement.var(0).value()
     if name in state:
         state[name].append((stmt_text, statement.config.path))
     else:
         state[name] = [(stmt_text, statement.config.path)]
 
+
 def find_dnssec_lookaside(statement, state):
     try:
         assert statement.var(0).value() == 'dnssec-lookaside'
```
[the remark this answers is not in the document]
ZhukovGreen: just adding reasoning:
- it could introducing security risks and bugs into your applications
- syntax quirk that makes it easy to write useless assertions (assertions could be globally disabled by the interpreter)
https://dbader.org/blog/python-assert-tutorial

repos/system_upgrade/el7toel8/libraries/isccfg.py
pemensik: would this change raise always StopIteration exception? I want to ignore StopIteration "errors", but not any other. It seems okay to me

repos/system_upgrade/el7toel8/libraries/tests/test_isccfg.py
pemensik: No. This allowed me testing the same file out of  repos structure by *pytest-3 test_isccfg* on Fedora. Because I had a lot of code to cleanup, debugging via Makefile was quite painful to me. I needed debugger ocassionally. This was easy workaround.

I don't modify path. But any python file in leapp repos needs running under make for some stupid reason. This allowed me to use pytest directly in different directory. I would like better fix, but this just worked enough.

repos/system_upgrade/el7toel8/actors/checkbind/libraries/model.py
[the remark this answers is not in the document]
pemensik: okay, a bit dumb. It is good practice to leave assertions in place in C. Just disable them on compile time. I admit it would not work in python.

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
pemensik: Okay, I thought I read that somewhere. But type is probably correct version.

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
pemensik: okay, its purpose was to have some way to log. logger member is not available in implementation detail. api.current_logger() should work instead.

repos/system_upgrade/el7toel8/actors/checkbind/libraries/model.py
pemensik: I don't get what I am supposed to do. For cases when keyword would have unexpected parameters, that means for example *dnssec-lookaside ".";* It would be invalid configuration for named, but I think it should not be fatal error. It might be permitted in future versions.

This code will add statement to state:
dnssec-lookaside yes;
dnssec-lookaside auto;
dnssec-lookaside "." trust-anchor "dlv.isc.org";

But it would ignore:
dnssec-lookaside no;

Another option might be removing try and let it crash hard, when unexpected configuration is found. But I don't know how to use finally in this case.

repos/system_upgrade/el7toel8/libraries/isccfg.py
pemensik: ok

repos/system_upgrade/el7toel8/libraries/isccfg.py
pemensik: Since kind may contain numeric constant, I would keep None as special value to notify. Value 0 should not pass this test, even it is not used.

repos/system_upgrade/el7toel8/libraries/isccfg.py
pemensik: ok

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
pemensik: Removed that altogether and used global logger.

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
pemensik: I would like instead link to logger methods supported in LEAPP. I had trouble finding it. Could you point me to logging classes docs?

repos/system_upgrade/el7toel8/actors/checkbind/libraries/model.py
pemensik: Ok, resolved by just checking for dnssec-lookaside no; which is the only supported configuration forwad.

repos/system_upgrade/el7toel8/actors/bindupdate/actor.py
pirat89: @ZhukovGreen This should be ok. In our guidelines, the format string is suggested. We have plenty of examples in our code where this is used even for logging.

repos/system_upgrade/el7toel8/libraries/tests/test_isccfg.py
pirat89: @pemensik This is already improved by @ZhukovGreen . You can run pytest nowadays without the makefile: https://leapp.readthedocs.io/en/latest/unit-testing.html#actor-s-tests

repos/system_upgrade/el7toel8/actors/bindupdate/actor.py
ZhukovGreen: fine to me

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
ZhukovGreen: from leapp.libraries.stdlib import api
api.current_logger

ZhukovGreen: No need to use logger specific str formatting, because in our codebase we have a lot of similar examples. However, I would use a proper way:)

ZhukovGreen: Feel free to close

repos/system_upgrade/el7toel8/libraries/isccfg.py
ZhukovGreen: you're right

repos/system_upgrade/el7toel8/actors/checkbind/libraries/model.py
pemensik: Well, that is not true. It complains when some return path returns something and another does not. It has to return something. Changed it to [], it still evaluates to false.

repos/system_upgrade/el7toel8/libraries/isccfg.py
pemensik: does it hurt to provide default mode? It seems a bit more clear to me, because I did not know it is the default.

repos/system_upgrade/el7toel8/actors/checkbind/libraries/model.py
ZhukovGreen: ok

repos/system_upgrade/el7toel8/libraries/isccfg.py
ZhukovGreen: no hurt:)

vojtechsokol:
leapp-ci build

Rezney:
@pemensik Hello, could you please rebase the PR? Thanks...

Rezney:
@pemensik Thank you very much for the rebase, could you please also fix the import warnings in the unit tests?

repos/system_upgrade/el7toel8/actors/bindupdate/actor.py
```diff
@@ -0,0 +1,34 @@
...
+
+    name = 'bind_update'
+    consumes = (InstalledRedHatSignedRPM, BindFacts)
+    produces = ()
+    tags = (PreparationPhaseTag, IPUWorkflowTag)
+
+    pkg_names = {'bind', 'bind-sdb', 'bind-pkcs11'}
+
+    def has_package(self, t_rpms):
+        """Replacement for broken leapp.libraries.common.rpms.has_package."""
```
fernflower, line 20: Hmm it should not be broken, and if it is - we (the leapp team) should fix it as it's a library function. Is there an open issue?

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
```diff
@@ -0,0 +1,85 @@
...
+    for cfg in parser.FILES_TO_CHECK:
+        if cfg.path == path:
+            return cfg
+    return None
+
+
+def make_backup(path, backup_suffix='.leapp'):
+    """Make backup of a file before modification."""
+    backup_path = path + backup_suffix
+    run(['cp', '--preserve=all', path, backup_path])
```
fernflower, line 28: This can potentially fail and raise CommandError. Please wrap this in try/except

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
```diff
@@ -0,0 +1,85 @@
...
+
+# Callback for walk function
+callbacks = {
+    'dnssec-lookaside': isccfg.ModifyState.callback_comment_out,
+}
+
+
+def paths_from_issues(issues):
+    """Extract paths from list of BindConfigIssuesModel."""
+    paths = []
```
fernflower, line 12: Nit: `return [issue.path for issue in issues]`

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
```diff
@@ -0,0 +1,85 @@
...
+    for issue in issues:
+        paths.append(issue.path)
+    return paths
+
+
+def parser_file(parser, path):
+    for cfg in parser.FILES_TO_CHECK:
+        if cfg.path == path:
+            return cfg
+    return None
```
fernflower, line 22: Nit: return next((cfg for cfg in parser.FILES_TO_CHECK if cfg.path == path), None)

repos/system_upgrade/el7toel8/actors/checkbind/actor.py
```diff
@@ -0,0 +1,42 @@
...
+        report = model.make_report(facts)
+
+        if report:
+            api.produce(facts)
+            self.log.info('BIND configuration issues were found.')
+            reporting.create_report(report)
+        else:
+            self.log.debug('BIND configuration seems compatible.')
+
+        pass
```
fernflower, line 42: Why do you need `pass` here?

repos/system_upgrade/el7toel8/actors/checkbind/libraries/model.py
```diff
@@ -0,0 +1,108 @@
...
+
+        api.current_logger().debug('Found state: "%s", files: "%s"',
+                                   repr(state), files)
+
+    facts = convert_found_state(state, list(files))
+    return facts
+
+
+def make_report(facts):
+    summary_messages = []
```
fernflower, line 91: Could you please change this a bit, so that returned value from all branches would have consistent typing? Something like

```
report = []
... 
if summary_messages:
    report.extend([reporting.Title(), reporting.Summary()])
return report
```

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
```diff
@@ -0,0 +1,85 @@
...
+    :param path: String to main configuration file
+    :returns: number of modified files
+    """
+    parser = isccfg.IscConfigParser(path)
+    modified_files = 0
+    if facts.dnssec_lookaside:
+        for model in facts.dnssec_lookaside:
+            if update_file(parser, model.path):
+                modified_files += 1
+    return modified_files
```
fernflower, line 85: Just asking: would it make sense to keep track of which exact files were modified? And store this information in the report?

repos/system_upgrade/el7toel8/libraries/tests/test_isccfg.py
fernflower: Yeah, this should be fixed. Sorry you had the user-unfriendly experiences with our tests, it should be better now.

fernflower (changes requested):
Loved the decent test coverage, thanks! :) Left a few comments.

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
pirat89: to be honest, I like the original way more than the proposed oneliner :) but it's matter of taste. both are good

repos/system_upgrade/el7toel8/actors/bindupdate/actor.py
pemensik: No issue known, I think this function were copied originally from checkpostfix actor. Not sure anymore, I will test using proper library.

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
pemensik: If the backup fails, some error has to be raised. If even making backup copy fails, I do not think it should continue updating anything. How should be failed actor marked? Is there correct error to be raised? I expect actor-level try/except should be used, is it?

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
pemensik: Unless it is functionally different from my version, I would like keep it as it is. Seems more readable to me.

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
pemensik: Good idea, I just wanted to log something. But I guess list of modified files should be kept in a single place.

pirat89:
e2e tests

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
pirat89: @pemensik  actor should never crash in the hard way. in the worst valid case it should raise `StopActorExecutionError` exception with proper error msg and possible details explanations (that exception has additional special handler on the framework lvl)

Rezney:
@pemensik We are good to merge this. I would like to just ask you for one more rebase and squash the 34 commits. Thanks.

repos/system_upgrade/el7toel8/actors/checkbind/libraries/model.py
```diff
@@ -0,0 +1,111 @@
...
+    report = []
+    if facts.dnssec_lookaside:
+        summary_messages.append('BIND configuration contains no longer accepted statements: dnssec-lookaside.')
+    if facts.listen_on_v6_missing:
+        summary_messages.append('Default value of listen-on-v6 have changed, but it is not present in configuration.'
+                                ' named service will now listen on INET6 sockets also.')
+
+    if summary_messages:
+        summary = ' '.join(summary_messages)
+        report.extend(
```
Rezney, line 103: Did not pass final testing:

```
Process Process-371:
    self.run()
  File "/usr/lib/python2.7/site-packages/leapp/actors/__init__.py", line 290, in run
    self.process(*args)
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/checkbind/actor.py", line 33, in process
    report = model.make_report(facts)
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/checkbind/libraries/model.py", line 108, in make_report
    reporting.Flags([reporting.Flags.INHIBITOR]),
TypeError: extend() takes exactly one argument (5 given)
```

Please provide a fix + add unit test and do not hesitate to try a simple integration test. Thank you.

Rezney (changes requested):

Rezney:
@pemensik Please fix the ImportError, hopefully it is the last issue and we could go ahead and merge it. Thanks.

Rezney (dismissed):

repos/system_upgrade/el7toel8/actors/bindupdate/libraries/updates.py
fernflower: Sorry for late response, missed the ping (twice!). How about the usual approach we take, by catching CallesProcessError and raising StopActorExecutionError (if you believe that there is no sense in proceeding)?

```
from leapp.libraries.stdlib import api, CalledProcessError, run

try:
    run(['cp', '--preserve=all', path, backup_path])
except CalledProcessError as exc:
    raise StopActorExecutionError('Could not create a backup copy', 'details': 'An exception during backup raised {}'.format(str(exc)))
```

fernflower (commented):
@pemensik So sorry for late response.
The PR is good to go, please fix the only one remark we have about exception handling.

Rezney (approved):

fernflower (commented):
Fix commit message please, s/expection/exception'

I have concerns about morf-leapp-bind test failure, the 7to8 upgrade is currently inhibited because of "BIND configuration contains no longer accepted statements: dnssec-lookaside." inhibitor.
If we merge this patch this test will start failing all the time, do we have a plan here?
@mreznik @mkluson @mmacura311

fernflower:
Okay, we came to an agreement to merge this test now and come up with the fix to the playbooks later. So merging, thanks for working on it everyone!

PR #522, opened 2020-07-03
Allow run tests on not RPM based systems

ZhukovGreen:
Very small MR introducing the possibility to run tests on non-RPM based linux distros

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkleftoverpackages/tests/test_checkleftover_checkleftoverpackages.py
+++ b/repos/system_upgrade/el7toel8/actors/checkleftoverpackages/tests/test_checkleftover_checkleftoverpackages.py
@@ -1,7 +1,0 @@
-from leapp.models import LeftoverPackages
-from leapp.snactor.fixture import current_actor_context
-
-
-def test_actor_execution(current_actor_context):
-    current_actor_context.run()
-    assert current_actor_context.consume(LeftoverPackages)
--- a/repos/system_upgrade/el7toel8/actors/removeresumeservice/tests/test_removeresumeservice.py
+++ b/repos/system_upgrade/el7toel8/actors/removeresumeservice/tests/test_removeresumeservice.py
@@ -1,16 +1,18 @@
-import os
 import errno
+import os
 
-import distro
 import pytest
-
-from leapp.snactor.fixture import current_actor_context
 
 
 @pytest.mark.skipif(
-    distro.linux_distribution()[0] == 'Fedora',
-    reason='default.target.wants does not exists on Fedora distro',
-    )
+    not os.path.isdir('/etc/systemd/system/default.target.wants')
+    or not os.getuid() == 0,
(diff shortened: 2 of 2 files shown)


ZhukovGreen:
please review

ZhukovGreen:
review please

pirat89:
@examon Seems the @leapp-bot is buggy as the review is not requested in the PR (just the comment is created...). I am going to set it, but pls check the bot.

repos/system_upgrade/el7toel8/actors/checkleftoverpackages/tests/test_checkleftover_checkleftoverpackages.py
```diff
@@ -1,7 +1,14 @@
+import distro
+import pytest
+
 from leapp.models import LeftoverPackages
-from leapp.snactor.fixture import current_actor_context
 
 
+@pytest.mark.skipif(
+    distro.linux_distribution()[0]
+    not in ('Fedora', 'CentOS Linux', 'Red Hat Enterprise Linux Server'),
```
pirat89, line 9: the PR is not doing what is telling.  there are more rpm based systems than in the list (e.g. [here](https://en.wikipedia.org/wiki/Category:RPM-based_Linux_distributions)). instead of this, rather check existence of  the `/usr/bin/rpm` file, if you want to be more precise. Anyway, this is still wrong as the test is just skipped instead of to be allowed to run. Basically this is incorrect test at all, as it is affected by environment and it not testing anything except the point it does not crash during the execution.

repos/system_upgrade/el7toel8/actors/removeresumeservice/tests/test_removeresumeservice.py
```diff
@@ -8,8 +8,8 @@
 
 
 @pytest.mark.skipif(
-    distro.linux_distribution()[0] == 'Fedora',
-    reason='default.target.wants does not exists on Fedora distro',
+    distro.linux_distribution()[0] not in ('CentOS Linux', 'Red Hat Enterprise Linux Server'),
+    reason='default.target.wants does not exists.',
```
pirat89, line 12: this is not connected to rpm distributions.

repos/system_upgrade/el7toel8/actors/checkleftoverpackages/tests/test_checkleftover_checkleftoverpackages.py
ZhukovGreen: then I suggest removing it @pirat89 . Agree with your statements

repos/system_upgrade/el7toel8/actors/removeresumeservice/tests/test_removeresumeservice.py
ZhukovGreen: Solving it in a different manner, where we are testing whether the default.targets.wants really exists, the test running under the root and at the end, it is marked as a destructive test

PR #527, opened 2020-07-13
WIP: POC - Inband debug data in initramfs

vinzenz:
TBD

Signed-off-by: Vinzenz Feenstra <vfeenstr@redhat.com>

PR #529, opened 2020-07-15
Add support for in-band console diagnostics

AloisMahdal:
Based on @vinzenz's earlier pull request #527.

Specification
=============

When upgrade fails in initramfs, log files and other assets will be tarballed and, using a special encoding method based on Base64, dumped to the console.

The encoding is designed to be resillient towards various forms of corruption such as other things writing to the console (mid-line) or line breaking (often seen on s390x machines in our Beaker pool).

By default, the dump functionality is disabled, but can be enabled by adding kernel option `rd.upgrade.inband`:

 *  If kernel option value is `onerror`, dump is done only if upgrade (systemd-nspawn) finished with non-zero exit status.

 *  If kernel option value is `always`, dump is done regardless of the exit status.

 *  If kernel option value is `never`, dump is not done, regardless of the exit status.

 *  If kernel option is not present, set to empty string or an unknown value, the dump is not done.  (Same as `never`.)

In order to retrieve the dumped tarball, one has to collect console log (UTF-8 is assumed) and run decoder (Python 3) included in the utils/ subdirectory:

    utils/ibdmp-decode console.log ibdmp.tar.xz

This will result in ibdmp.tar.xz, containing the set of files encoded by the upgrade dracut module.

Implementation notes
====================

In the dracut module, the `collect_and_dump_debug_data()` function is where the tarball is created.  This function passes the tarball to `ibdmp()`, which takes care of the encoding and dumping.

To make sure the data is recoverable even if the console log is affected by irrelevant writes or later corruption, the `ibdmp()` function employs lot of redundancy:  after splitting the base64-encoded tarball to small chunks and pre-pending it with some meta-data, it will perform 3 dumps, each separated with progressively increasing delay.

That is, each dump consists of one "dump header" and a large number of small (40 characters by default) "chunks".

The decoder will first find all meta-data headers (learning how many chunks to look for), then collect all chunks from all dumps and merge them according to prevalency, ie. if same chunk appears twice with value "foo" but once with value "bar", the second value is disregarded.

Finally all the chunks are joinded and decoded, resulting in original tarball file, which is finally checked against MD5 sum included in the dump header.

The number of dumps and chunk size can be controlled by integers `IBDMP_ITER` (number of dumps, 3 by default) and `IBDMB_CHUNKSIZE` length of each chunk in characters, 40 by default).  These parameters are purely internal constants; there's no intent to let user control them.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
+++ b/repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
@@ -17,6 +17,154 @@ NSPAWN_OPTS="--capability=all --bind=/sys --bind=/dev --bind=/dev/pts --bind=/ru
 [ -d /dev/mapper ] && NSPAWN_OPTS="$NSPAWN_OPTS --bind=/dev/mapper"
 export NSPAWN_OPTS="$NSPAWN_OPTS --bind=/run/udev --keep-unit --register=no --timezone=off --resolv-conf=off"
 
+#
+# Temp for collecting and preparing tarbal
+#
+LEAPP_DEBUG_TMP="/tmp/leapp-debug-root"
+
+#
+# Number of times to emit all chunks
+#
+# To avoid spammy parts of console log, second and later emissions
+# take longer delay in-between.  For example, with N being 3,
+# first emission is done immediately, second after 10s, and the
+# third one after 20s.
+#
+IBDMP_ITER=3
+
+#
+# Size of one payload chunk
+#
+# IOW, amount of characters in a single chunk of the base64-encoded
+# payload.   (By base64 standard, these characters are inherently ASCII,
+# so ie. they correspond to bytes.)
+#
+IBDMP_CHUNKSIZE=40
+
+collect_and_dump_debug_data() {
+    #
+    # Collect various debug files and dump tarball using ibdmp
+    #
+    local tmp=$LEAPP_DEBUG_TMP
+    local data=$tmp/data
+    mkdir -p "$data" || { echo >&2 "fatal: cannot create leapp dump data dir: $data"; exit 4; }
+    journalctl -amo verbose >"$data/journalctl.log"
+    mkdir -p "$data/var/lib/leapp"
+    mkdir -p "$data/var/log"
+    cp -vr "$NEWROOT/var/lib/leapp/leapp.db" \
+          "$data/var/lib/leapp"
+    cp -vr "$NEWROOT/var/log/leapp" \
+          "$data/var/log"
+    tar -cJf "$tmp/data.tar.xz" "$data"
+    ibdmp "$tmp/data.tar.xz"
+    rm -r "$tmp"
+}
+
+want_inband_dump() {
+    #
+    # True if dump collection is needed given leapp exit status $1 and kernel option
+    #
+    local leapp_es=$1
+    local mode
+    local kopt
+    kopt=$(getarg 'rd.upgrade.inband')
+    case $kopt in
+        always|never|onerror)   mode="$kopt" ;;
+        "")                     mode="never" ;;
+        *)  warn "ignoring unknown value of rd.upgrade.inband (dump will be disabled): '$kopt'"
+            return 2 ;;
+    esac
+    case $mode:$leapp_es in
+        always:*)   return 0 ;;
+        never:*)    return 1 ;;
+        onerror:0)  return 1 ;;
+        onerror:*)  return 0 ;;
+    esac
+}
+
+ibdmp() {
+    #
+    # Dump tarball $1 in base64 to stdout
+    #
+    # Tarball is encoded in a way that:
+    #
+    #   * final data can be printed to plain text terminal,
+    #   * tarball can be restored by scanning the saved
+    #     terminal output,
+    #   * corruptions caused by extra terminal noise
+    #     (extra lines, extra characters within lines,
+    #     line splits..) can be corrected.
+    #
+    # That is,
+    #
+    #   1. encode tarball using base64
+    #
+    #   2. pre-pend line `chunks=CHUNKS,md5=MD5` where
+    #      MD5 is the MD5 digest of original tarball and
+    #      CHUNKS is number of upcoming Base64 chunks
+    #
+    #   3. decorate each chunk with prefix `N:` where
+    #      N is number of given chunk.
+    #
+    #   4. Finally print all lines (pre-pended "header"
+    #      line and all chunks) several times, where
+    #      every iteration should be prefixed by
+    #      `_ibdmp:I/TTL|` and suffixed by `|`.
+    #      where `I` is iteration number and `TTL` is
+    #      total iteration numbers.
+    #
+    # Decoder should look for strings like this:
+    #
+    #     _ibdmp:I/J|CN:PAYLOAD|
+    #
+    # where I, J and CN are integers and PAYLOAD is a slice of a
+    # base64 string.
+    #
+    # Here, I represents number of iteration, J total of iterations
+    # ($IBDMP_ITER), and CN is number of given chunk within this
+    # iteration.  CN goes from 1 up to number of chunks (CHUNKS)
+    # predicted by header.
+    #
+    # Each set corresponds to one dump of the tarball and error
+    # correction is achieved by merging sets using these rules:
+    #
+    #    1. each set has to contain header (`chunks=CHUNKS,
+    #       md5=MD5`) prevalent header wins.
+    #
+    #    2. each set has to contain number of chunks
+    #       as per header
+    #
+    #    3. chunks are numbered so they can be compared across
+    #       sets; prevalent chunk wins.
+    #
+    # Finally the merged set of chunks is decoded as base64.
+    # Resulting data has to match md5 sum or we're hosed.
+    #
+    local tarball=$1
+    local tmp=$LEAPP_DEBUG_TMP/ibdmp
+    local md5
+    local i
+    mkdir -p "$tmp"
+    base64 -w "$IBDMP_CHUNKSIZE" "$tarball" > "$tmp/b64"
+    md5=$(md5sum "$tarball" | sed 's/ .*//')
+    chunks=$(wc -l <"$tmp/b64")
+    (
+        set +x
+        echo "chunks=$chunks,md5=$md5"
+        cnum=1
+        while read -r chunk; do
+            echo "$cnum:$chunk"
+            ((cnum++))
+        done <"$tmp/b64"
+    ) >"$tmp/report"
+    i=0
+    while test "$i" -lt "$IBDMP_ITER"; do
+        sleep "$((i * 10))"
+        ((i++))
+        sed "s%^%_ibdmp:$i/$IBDMP_ITER|%; s%$%|%; " <"$tmp/report"
+    done
+}
+
 
 do_upgrade() {
     local args="" rv=0
@@ -61,6 +209,11 @@ do_upgrade() {
         # need to handle more stuff around storage at all.
         /usr/bin/systemd-nspawn $NSPAWN_OPTS -D $NEWROOT /usr/bin/bash -c "mount -a; /usr/bin/python3 $LEAPP3_BIN upgrade --resume $args"
         rv=$?
+    fi
+
+    # Dump debug data in case something went wrong
+    if want_inband_dump "$rv"; then
+        collect_and_dump_debug_data
     fi
 
     # NOTE: THIS SHOULD BE AGAIN PART OF LEAPP IDEALLY
--- a/repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/module-setup.sh
+++ b/repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/module-setup.sh
@@ -4,10 +4,12 @@ LEAPPBIN=/usr/bin/leapp
 LEAPP_VERSION=0.8.10
 
 check() {
-    # [ -x $LEAPPBIN ] || return 1
-    # return 255
-    #[ -x "$LEAPPBIN" ]
-    # NOTE: we don't need $LEAPPBIN FOR NOW as part of of initrd
+    require_binaries base64 || return 1
+    require_binaries tar || return 1
+    require_binaries xz || return 1
+    require_binaries md5sum || return 1
+    require_binaries wc || return 1
+    # 0 enables by default, 255 only on request
     return 0
 }
 
@@ -73,6 +75,12 @@ install() {
 
     # install this one to ensure we are able to sync write
     inst_binary sync
+    # install in-band debugging utilities
+    inst_binary base64
+    inst_binary tar
+    inst_binary xz
+    inst_binary md5sum
+    inst_binary wc
 
     # script to actually run the upgrader binary
     inst_hook upgrade 49 "$moddir/mount_usr.sh"
--- a/utils/ibdmp-decode
+++ b/utils/ibdmp-decode
@@ -1,0 +1,265 @@
+#!/usr/bin/python3
+
+import base64
+import collections
+import hashlib
+import re
+import sys
+
+DEBUG = False
+RE_HEADER = r'_ibdmp:[0-9]+/[0-9]+[|]chunks=[0-9]+,md5=[0-9a-f]+[|]'
+RE_CHUNK = r'_ibdmp:[0-9]+/[0-9]+[|][0-9]+:[A-Za-z0-9+/=]+[|]'
+
+
+def USAGE():
+    self = sys.argv[0].split('/')[-1]
+    lines = [
+        "usage: %s path/to/console.log path/to/target.tar.xz" % self,
+        "",
+        "Decode debug tarball emited by leapp's initramfs in-band",
+        "console debugger, ibdmp().",
+    ]
+    sys.stderr.writelines('%s\n' % l for l in lines)
+    sys.exit(2)
+
+
+def LOG_DEBUG(msg):
+    if DEBUG:
+        sys.stderr.write('DEBUG: %s\n' % msg)
+
+
+def LOG_WARN(msg):
+    sys.stderr.write('%s\n' % msg)
+
+
+class IbdmpDecodeError(ValueError):
+    pass
+
+
+class UsageError(ValueError):
+    pass
+
+
+class _Chunk:
+
+    @classmethod
+    def from_raw1(cls, raw_chunk):
+        if not raw_chunk.startswith('_ibdmp:'):
+            LOG_WARN("invalid chunk payload (no '_ibdmp:'?): %s"
+                     % raw_chunk)
+            raise IbdmpDecodeError(raw_chunk)
+        areas = raw_chunk.split('|')
+        parts = areas[1].split(':')
+        return cls(
+            ordinal=int(parts[0]),
+            payload=str(parts[1]),
+        )
+
+    def __init__(self, ordinal, payload):
+        self.ordinal = ordinal
+        self.payload = payload
+
+
+class Header:
+    """
+    Chunk set header
+    """
+
+    @classmethod
+    def from_rawN(cls, raw_headers):
+        """
+        Initialize chunk header from header chunk candidates
+
+        raw_headers is a list of strings that contain encoded chunk
+        parameters for the whole chunk set, ie. number of chunks, number
+        of iterations, and MD5 hash of the content encoded in the chunk set.
+
+        Raw header chunks can be corrupted so this factory will choose
+        winner based on prevalence.
+
+        For chunk set example in ChunkCounter.__init__ corresponding
+        raw headers could look similar to this:
+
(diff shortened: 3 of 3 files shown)


pirat89:
be aware of trailing spaces :)

examon:
leapp-ci build

AloisMahdal:
s390x, dumping 3 times, wrapping at 40 chars:

fulll console.log: 3.0M
 * size without in-band-debug: 962K
 * in-band-debug: 865K
 * spam caused by in-band debug function running in `set -x` mode: 1.3M

@vinzenz I think it's bearable, and if we neutralize the `set -x` spam; we would "just double" the console.log.

AloisMahdal:
The good news is that here's the decoder (*warning:* outputs binary! call with redirect, eg. `./inband_dec.py console.log > console.tar.xz`):

    #!/usr/bin/python3

    import base64
    import collections
    import hashlib
    import re
    import sys

    RE_HEADER = r'_leapp_debug:[0-9]+/[0-9]+[|]chunks=[0-9]+,md5=[0-9a-f]+[|]'
    RE_CHUNK = r'_leapp_debug:[0-9]+/[0-9]+[|][0-9]+:[A-Za-z0-9+/=]+[|]'


    class Chunk:

        @classmethod
        def from_raw(cls, raw):
            assert raw.startswith('_leapp_debug:')
            areas = raw.split('|')
            parts = areas[1].split(':')
            return cls(
                cord=int(parts[0]),
                cdata=str(parts[1]),
            )

        def __init__(self, cord, cdata):
            self.cord = cord
            self.cdata = cdata

        def __hash__(self):
            return hash((self.cord, self.cdata))


    class Header:

        @classmethod
        def from_raw(cls, raw):
            parts = raw.split('|')
            _, stats = parts[0].split(':')
            pairs = parts[1].split(',')
            assert pairs[0].startswith('chunks=')
            assert pairs[1].startswith('md5=')
            return cls(
                chunks=int(pairs[0].split('=')[1]),
                md5=str(pairs[1].split('=')[1]),
                csets=int(stats.split('/')[1]),
            )

        def __init__(self, chunks, md5, csets):
            self.chunks = chunks
            self.md5 = md5
            self.csets = csets

        def __eq__(self, othr):
            return (self.chunks, self.md5) == (othr.chunks, othr.md5)

        def __hash__(self):
            return hash((self.chunks, self.md5))


    class ChunkCounter:

        def __init__(self, header, chunks_raw):
            self.header = header
            self._chunks = [None] * header.chunks
    #       sys.stderr.write('header.chunks=%r\n' % header.chunks)
            for cr in chunks_raw:
                c = Chunk.from_raw(cr)
    #           sys.stderr.write('c.cord=%r\n' % c.cord)
                if self._chunks[c.cord-1] is None:
                    self._chunks[c.cord-1] = collections.Counter([c])
                else:
                    self._chunks[c.cord-1].update([c])

        @property
        def chunks(self):
            out = []
            i = 0
            for c in self._chunks:
                i += 1
                if c is None:
                    sys.stderr.write('Missing chunk id: %d/%d\n'
                                     % (i, self.header.chunk))
                out.append(c.most_common()[0][0].cdata)
            return out

        def dump64(self):
            for c in self.chunks:
                print(c)

        def decode(self):
            tarball = base64.b64decode(''.join(self.chunks))
            tar_md5 = hashlib.md5(tarball).hexdigest()
            assert tar_md5 == self.header.md5
            sys.stdout.buffer.write(tarball)


    if __name__ == '__main__':
        headers_raw = []
        chunks_raw = []
        with open(sys.argv[1]) as f:
            for line in f.readlines():
                for m in re.findall(RE_HEADER, line):
                    headers_raw.append(m)
                for m in re.findall(RE_CHUNK, line):
                    chunks_raw.append(m)

        headers_cnt = collections.Counter([
            Header.from_raw(h) for h in headers_raw
        ])
        header = headers_cnt.most_common()[0][0]
        ccounter = ChunkCounter(header, chunks_raw)
    #   ccounter.dump64()
        ccounter.decode()
    #   print(chunks_raw)

the bad news is that I've decoded my data and turns out that it's only journalctl.log.

Ie. the the above computation **only counts journtalctl.log**.  for some reason my collector does not pick up any code.

vinzenz:
Tested on my machine works perfectly fine. Follow up needed to make this optional however.

AloisMahdal:
Where should I place the decoder?  cc @Rezney

pirat89:
@AloisMahdal e.g. into the utils or into the snippets repository seems ok to me.

AloisMahdal:
Improvements since the last version:

 *  When searching for chunks it works with window of 2 lines so line breaks can be fixed.
 *  Added some meaningful debugging and error messaging and exit status policy.
 *  Couple of fixes.
 *  Added Python docstrings & Bash docstrings.

What still needs to be done:

 *  RTD docs (part of the original spec in our Jira); the docstrings are probably not suitable.
 *  Make it optional as @vinzenz suggested above.

AloisMahdal:
@vinzenz I'd like to make the dumper optional.  I suggest 3 modes affecting whether or not `collect_and_dump_debug_data()` is called:
 * off - never,
 * on error - ie. on leapp error,
 * always - always (OAMG-QE would use this for a time to help debug long-standing issues such as rhel7 kernel boot).

I'm not sure what's the proper way of controlling options in the initramfs (kernel options? via leapp?) and how to access them in the dracut code.

WDYT?  Cc @shaded-enmity, @Rezney, @pirat89, anyone else?

Rezney:
@AloisMahdal thanks for working on this. I like the 3 mode idea. For controlling, a kernel cmdline option should be ok. I will leave the proper access in dracut question for @vinzenz

AloisMahdal:
I've added the kernel option logic (separate commit for now, i will squash it prior to review):

 *  if `rd.upgrade.inband=never` - dump is never produced, regardless of leapp exit status.
 *  if `rd.upgrade.inband=always` - dump is always produced, regardless of leapp exit status.
 *  if `rd.upgrade.inband=onerror` - dump is produced, if exit status was other than zero.
 *  without the option at all, we act as if the value was `never`,
 *  if the option was provided but with unknown value, warning is printed and we act as if the value was `never`.

@vinzenz @Rezney, others, WDYT?

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -17,6 +17,149 @@ NSPAWN_OPTS="--capability=all --bind=/sys --bind=/dev --bind=/dev/pts --bind=/ru
...
+# take longer delay in-between.  For example, with N being 3,
+# first emission is done immediately, second after 10s, and the
+# third one after 20s.
+#
+IBDMP_ITER=3
+
+#
+# Size of one chunk payload
+#
+IBDMB_CHUNKSIZE=40
```
shaded-enmity, line 38: The whole archive will be dumped to the console `IBDMP_ITER` times? 

As for `IBDMB_CHUNKSIZE`:
1. Shouldn't it be `IBDMP_CHUNKSIZE`?
2. It's size in bytes / characters?

vinzenz: > The whole archive will be dumped to the console `IBDMP_ITER` times?
> 
> As for `IBDMB_CHUNKSIZE`:
> 
> 1. Shouldn't it be `IBDMP_CHUNKSIZE`?
Good catch
> 2. It's size in bytes / characters?
Since the dump is supposed to be BASE64 and the data is supposed to be UTF-8, the set of characters would cause size in bytes == number of characters

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -17,6 +17,149 @@ NSPAWN_OPTS="--capability=all --bind=/sys --bind=/dev --bind=/dev/pts --bind=/ru
...
+    mkdir -p "$data" || { echo >&2 "fatal: cannot create leapp dump data dir: $data"; exit 4; }
+    journalctl -amo verbose >"$data/journalctl.log"
+    mkdir -p "$data/var/lib/leapp"
+    mkdir -p "$data/var/log"
+    cp -vr "$NEWROOT/var/lib/leapp/leapp.db" \
+          "$data/var/lib/leapp"
+    cp -vr "$NEWROOT/var/log/leapp" \
+          "$data/var/log"
+    tar -cJf "$tmp/data.tar.xz" "$data"
+    ibdmp "$tmp/data.tar.xz"
```
shaded-enmity, line 55: Hmm, I'm thinking of two possible modes here:

1. We keep the data file in well known location (`rd.inband=keep`?)
2. We remove the data after they have been dumped (think systems where `/tmp` is not `tmpfs` and we're potentially leaking data to `/tmp`)

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -17,6 +17,149 @@ NSPAWN_OPTS="--capability=all --bind=/sys --bind=/dev --bind=/dev/pts --bind=/ru
...
+    local tarball=$1
+    local tmp=$LEAPP_DEBUG_TMP/ibdmp
+    local md5
+    local i
+    mkdir -p "$tmp"
+    base64 -w "$IBDMB_CHUNKSIZE" "$tarball" > "$tmp/b64"
+    md5=$(md5sum "$tarball" | sed 's/ .*//')
+    chunks=$(wc -l <"$tmp/b64")
+    (
+        set +x
```
shaded-enmity, line 147: This doesn't need to be here, right?

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/module-setup.sh
```diff
@@ -73,6 +76,13 @@ install() {
 
     # install this one to ensure we are able to sync write
     inst_binary sync
+    # install in-band debugging utilities
+    inst_binary base64
+    inst_binary tar
+    inst_binary xz
+    inst_binary sha256sum
```
shaded-enmity, line 83: I don't see `sha256sum` being used anywhere in the code, is it a dependency of a dependency?

utils/ibdmp-decode
```diff
@@ -0,0 +1,261 @@
...
+            ordinal=int(parts[0]),
+            payload=str(parts[1]),
+        )
+
+    def __init__(self, ordinal, payload):
+        self.ordinal = ordinal
+        self.payload = payload
+
+    def __hash__(self):
+        return hash((self.ordinal, self.payload))
```
shaded-enmity, line 60: Why is the `ordinal` part of the hash? Isn't the `payload` the identity element (content-addressable)?

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
pirat89: @AloisMahdal you should probably check the script by shellcheck

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
AloisMahdal: I think it does.  --  Note that `+x` *disables* the shell debug mode.

So if I drop it, the shell will re-print every upcoming arithmetic and `echo`, making the log **massive**.

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
AloisMahdal: > Hmm, I'm thinking of two possible modes here:
> 
> 1. We keep the data file in well known location (`rd.inband=keep`?)

I'd keep it as RFE for next PR.

> 2. We remove the data after they have been dumped (think systems where `/tmp` is not `tmpfs` and we're potentially leaking data to `/tmp`)

Good point, seems relevant to our overall function of the initramfs phase.

Ie. if leaking data matters, I imagine we should have general policy that applies not just to `collect_and_dump...*` but to all of our dracut code.  (The `keep` thing should be treated in the same context.)

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/module-setup.sh
AloisMahdal: Good point, removing.

utils/ibdmp-decode
AloisMahdal: Not if it's corrupted.

(That is, if i remember my thinking -- from 5 months ago -- correctly.)

AloisMahdal: Umm, you are probably right.

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
AloisMahdal: yeah, it's number of b64-enoced chars per chunk, which by definition will be ascii, ie. bytes.

I will clarify it in the docstring.

FWIW, this is not directly about the console size footprint (although it does affect how much meta-data there will be) but more about "how big a string can i safely print without something messing with it".  (Ie. the first value is roughly half of the "standard" 80-wide terminal.)

AloisMahdal: > @AloisMahdal you should probably check the script by shellcheck

I have (and the code *I*'m adding is shellcheck-ok), but the typo would not be caught by ShellCheck (it was everywhere).

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
shaded-enmity: oh right, the perks of shell using `+` to disable things and `-` to disable things :D

utils/ibdmp-decode
shaded-enmity: Hmm, is this part even needed? I don't see `_Chunk` being hashed anywhere.

utils/ibdmp-decode
```diff
@@ -0,0 +1,261 @@
...
+            md5=str(pairs[1].split('=')[1]),
+            csets=int(stats.split('/')[1]),
+        )
+
+    def __init__(self, chunks, md5, csets):
+        self.chunks = chunks
+        self.md5 = md5
+        self.csets = csets
+
+    def __eq__(self, othr):
```
shaded-enmity, line 122: Make sure to also provide `__neq__` to save some headaches later on (right, @pirat89 ? :))

utils/ibdmp-decode
```diff
@@ -0,0 +1,261 @@
...
+        LOG_DEBUG("header winner: %s" % winner)
+        return winner
+
+    @classmethod
+    def _from_raw1(cls, raw_header):
+        parts = raw_header.split('|')
+        _, stats = parts[0].split(':')
+        pairs = parts[1].split(',')
+        try:
+            assert pairs[0].startswith('chunks=')
```
shaded-enmity, line 105: Don't `assert` → `except AssertionError` in non-test code. Assertions are skipped when Python is running with `-O` or `-OO` so they shouldn't be relied on for control flow (pretty much as all exception handling).

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
AloisMahdal: Yeah, no blame though, POSIX might be older than modern use of `+` and `-` in mathematics :D

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
AloisMahdal: Anyway, I've fixed the typo 827e66e1 and clarified the docstring 5b4c85c

utils/ibdmp-decode
AloisMahdal: I was under the impression that one does not need to do that in Python 3 anymore... but I might be wrong.  I've added `__neq__` in abf4c26

utils/ibdmp-decode
AloisMahdal: OK, 18c9380 replaces asserts (here but also in `_Chunk.from_raw1` around L50) with `if` and uses "classic" `LOG_WARN` + `IbdmbDecodeError` combo.

And yes, it's `IbdmbDecodeError`, and yes, it's `b`, and yes, it's the same f-ing typo again -- see, @pirat89 ? -- even pylint does not help :D  fixed in 430badd

utils/ibdmp-decode
AloisMahdal: I think the hashability is needed because the use of `collections.Counter` relies on it.

Ie. check out `ChunkCounter.__init__` -- I'm loading the chunks into a huge dictionary (`ChunkCounter._bagset`) where keys are chunk ordinals, and each value is `collections.Counter` -- that way I can load chunks in any order and `collections.Counter` will do the maths.

You're probably right that the `ordinal` does not need to be part of the hash just for this to work.  OTOH, I would not say the payload is content-addressable: one could have "same" chunk at different positions and semantically it would not really be "the same chunk" -- so even if it's not functionally necessary (`_Chunk` is internal, after all) it seems more correct to include it.

WDYY?

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
AloisMahdal: After some pondering, I'm simply adding

    rm -r "$tmp"

after the dump. 9543e99

That should solve the potential privacy risk of leaving the dump around.

AloisMahdal:
I found a minor bug in the decoder chunk loading code: in case N chunks are completely lost (in all dumps), when loading chunks, the decoder would act as if there's N less overall chunks to be loaded (it would look for indices 1 .. total-N, instead of 1..total).  Ie. the number of expected chunks would be lowered by the fact that when raw-loading, some were never seen.

This would affect only cases where the data is badly corrupted anyway (in which case the MD5 check would fail), but it would make debug log misleading, so I went ahead and fixed it in 531f3c2;  now the loader will always use the number from the header.

AloisMahdal:
I found a minor bug in the decoder chunk loading code: in case N chunks are completely lost (in all dumps), when loading chunks, the decoder would act as if there's N less overall chunks to be loaded (it would look for indices 1 .. total-N, instead of 1..total).  Ie. the number of expected chunks would be lowered by the fact that when raw-loading, some were never seen.

This would affect only cases where the data is badly corrupted anyway (in which case the MD5 check would fail), but it would make debug log misleading, so I went ahead and fixed it in 531f3c2;  now the loader will always use the number from the header.

utils/ibdmp-decode
shaded-enmity: So on line 173 I see:
```
self._bagset[c.ordinal].update([c.payload])
```

So it seems that `c` doesn't need to be hashable because it's members are put into the structure and not `c` itself?

shaded-enmity: nvm it probably gets hashed at line 130

AloisMahdal: L130 is hashing `Header.chunks` but that's just int

AloisMahdal: IOW, you're right, neither `Header` nor `_Chunk` actually need `__hash__` -- it's likely just remainder from some earlier attempt ising `set` or something.

AloisMahdal:
...aand squashed commits from review back to one.

AloisMahdal:
....aand rebased on `master`.

shaded-enmity (commented):
Code-wise LGTM, still need to try it out. :+1:

pirat89:
leapp-ci build

pirat89:
leapp-ci build

shaded-enmity (approved):
LGTM. One thing that I just realized is documentation, but let's not block on it - you did a good job describing the interface in the commit message and the docs live in different repository anyway. Can you create a task to update https://github.com/oamg/leapp/tree/master/docs/source ?

PR #531, opened 2020-07-16
Add actor localreposinhibit

ZhukovGreen:
ref OAMG-3610

We are unable currently to handle local repos, specified
by baseurl=file:///.+

This actor inhibits the upgrade if such repos exists

# TODO

- [ ] squash commits before merge

# report inside the shell

```
                     UPGRADE INHIBITED                      

    1. Inhibitor: Local repository identified.
```

# content  snippet from leapp-report.txt
```
Risk Factor: high (inhibitor)
Title: Local repository identified.
Summary: Local repository &s with baseurl %s were found. Currently leapp is not supporting this option.
Remediation: [hint] By using Apache HTTP Server you can expose your local repo via http. For details see external link.
```

# test coverage
```

----------- coverage: platform linux, python 3.8.3-final-0 -----------
Name    Stmts   Miss  Cover   Missing
-------------------------------------
-------------------------------------
TOTAL      28      0   100%

2 files skipped due to complete coverage.

Required test coverage of 95.0% reached. Total coverage: 100.00%
```

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
@@ -54,8 +54,8 @@ class LocalReposInhibit(Actor):
                     reporting.Remediation(
                         hint=(
                             "By using Apache HTTP Server you can expose "
-                            "your local repository via http. For details "
-                            "see external link."
+                            "your local repository via http. See the linked "
+                            "article for details. "
                         )
                     ),
                     reporting.ExternalLink(
--- a/repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
@@ -41,12 +41,12 @@ class LocalReposInhibit(Actor):
         if self.file_baseurl_in_use():
             warn_msg = (
                 "Local repository found (baseurl starts with file:///). "
-                "Currently leapp is not supporting this option."
+                "Currently leapp does not support this option."
             )
             self.log.warning(warn_msg)
             reporting.create_report(
                 [
-                    reporting.Title("Local repository identified"),
+                    reporting.Title("Local repository detected"),
                     reporting.Summary(warn_msg),
                     reporting.Severity(reporting.Severity.HIGH),
                     reporting.Tags([reporting.Tags.REPOSITORY]),
--- a/repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
@@ -19,10 +19,8 @@ class LocalReposInhibit(Actor):
     tags = (IPUWorkflowTag, TargetTransactionChecksPhaseTag)
 
     def process(self):
-        # fmt: off
         used_target_repos = next(self.consume(UsedTargetRepositories)).repos
         target_repos = next(self.consume(TMPTargetRepositoriesFacts)).repositories
-        # fmt: on
         target_repo_id_to_url_map = {
             repo.repoid: repo.baseurl
             for repofile in target_repos
--- a/repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
@@ -3,8 +3,10 @@ from leapp.actors import Actor
 from leapp.models import TMPTargetRepositoriesFacts, UsedTargetRepositories
 from leapp.reporting import Report
 from leapp.tags import IPUWorkflowTag, TargetTransactionChecksPhaseTag
+from leapp.utils.deprecation import suppress_deprecation
 
 
+@suppress_deprecation(TMPTargetRepositoriesFacts)
 class LocalReposInhibit(Actor):
     """Inhibits the upgrade if local repositories were found."""
 
--- a/repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
@@ -1,0 +1,64 @@
+from leapp import reporting
+from leapp.actors import Actor
+from leapp.models import TMPTargetRepositoriesFacts, UsedTargetRepositories
+from leapp.reporting import Report
+from leapp.tags import IPUWorkflowTag, TargetTransactionChecksPhaseTag
+
+
+class LocalReposInhibit(Actor):
+    """Inhibits the upgrade if local repositories were found."""
+
+    name = "local_repos_inhibit"
+    consumes = (
+        UsedTargetRepositories,
+        TMPTargetRepositoriesFacts,
+    )
+    produces = (Report,)
+    tags = (IPUWorkflowTag, TargetTransactionChecksPhaseTag)
+
+    def process(self):
+        # fmt: off
+        used_target_repos = next(self.consume(UsedTargetRepositories)).repos
+        target_repos = next(self.consume(TMPTargetRepositoriesFacts)).repositories
+        # fmt: on
+        target_repo_id_to_url_map = {
+            repo.repoid: repo.baseurl
+            for repofile in target_repos
+            for repo in repofile.data
+            if repo.baseurl
+        }
+        if any(
+            target_repo_id_to_url_map[repo.repoid].startswith("file:")
+            for repo in used_target_repos
+        ):
+            warn_msg = (
+                "Local repository found (baseurl starts with file:///). "
+                "Currently leapp is not supporting this option."
+            )
+            self.log.warning(warn_msg)
+            reporting.create_report(
+                [
+                    reporting.Title("Local repository identified"),
+                    reporting.Summary(warn_msg),
+                    reporting.Severity(reporting.Severity.HIGH),
+                    reporting.Tags([reporting.Tags.REPOSITORY]),
+                    reporting.Flags([reporting.Flags.INHIBITOR]),
+                    reporting.Remediation(
+                        hint=(
+                            "By using Apache HTTP Server you can expose "
+                            "your local repository via http. For details "
+                            "see external link."
+                        )
+                    ),
+                    reporting.ExternalLink(
+                        title=(
+                            "Customizing your Red Hat Enterprise Linux "
+                            "in-place upgrade"
+                        ),
+                        url=(
+                            "https://access.redhat.com/articles/4977891/"
+                            "#repos-known-issues"
+                        ),
+                    ),
+                ]
+            )
--- a/repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
@@ -18,7 +18,12 @@ class LocalReposInhibit(Actor):
     produces = (Report,)
     tags = (IPUWorkflowTag, TargetTransactionChecksPhaseTag)
 
-    def process(self):
+    def file_baseurl_in_use(self):
+        """Check if any of target repos is local.
+
+        UsedTargetRepositories doesn't contain baseurl attribute. So gathering
+        them from model TMPTargetRepositoriesFacts.
+        """
         used_target_repos = next(self.consume(UsedTargetRepositories)).repos
         target_repos = next(self.consume(TMPTargetRepositoriesFacts)).repositories
         target_repo_id_to_url_map = {
@@ -27,10 +32,13 @@ class LocalReposInhibit(Actor):
             for repo in repofile.data
             if repo.baseurl
         }
-        if any(
+        return any(
             target_repo_id_to_url_map[repo.repoid].startswith("file:")
             for repo in used_target_repos
-        ):
+        )
+
+    def process(self):
+        if self.file_baseurl_in_use():
             warn_msg = (
                 "Local repository found (baseurl starts with file:///). "
                 "Currently leapp is not supporting this option."
--- a/repos/system_upgrade/el7toel8/actors/localreposinhibit/tests/test_unit_localreposinhibit.py
+++ b/repos/system_upgrade/el7toel8/actors/localreposinhibit/tests/test_unit_localreposinhibit.py
@@ -19,34 +19,35 @@ def test_unit_localreposinhibit(current_actor_context, baseurl, exp_msgs_len):
 
     :type current_actor_context: ActorContext
     """
-    current_actor_context.feed(
-        TMPTargetRepositoriesFacts(
-            repositories=[
-                RepositoryFile(
-                    file="the/path/to/some/file",
-                    data=[
-                        RepositoryData(
-                            name="BASEOS",
-                            baseurl=(
-                                "http://example.com/path/to/repo/BaseOS/x86_64/os/"
+    with pytest.deprecated_call():
+        current_actor_context.feed(
+            TMPTargetRepositoriesFacts(
+                repositories=[
+                    RepositoryFile(
+                        file="the/path/to/some/file",
+                        data=[
+                            RepositoryData(
+                                name="BASEOS",
+                                baseurl=(
+                                    "http://example.com/path/to/repo/BaseOS/x86_64/os/"
+                                ),
+                                repoid="BASEOS",
                             ),
-                            repoid="BASEOS",
-                        ),
-                        RepositoryData(
-                            name="APPSTREAM",
-                            baseurl=(
-                                "http://example.com/path/to/repo/AppStream/x86_64/os/"
+                            RepositoryData(
+                                name="APPSTREAM",
+                                baseurl=(
+                                    "http://example.com/path/to/repo/AppStream/x86_64/os/"
+                                ),
+                                repoid="APPSTREAM",
                             ),
-                            repoid="APPSTREAM",
-                        ),
-                        RepositoryData(
-                            name="CRB", repoid="CRB", baseurl=baseurl
-                        ),
-                    ],
-                )
-            ]
+                            RepositoryData(
+                                name="CRB", repoid="CRB", baseurl=baseurl
+                            ),
+                        ],
+                    )
+                ]
+            )
         )
-    )
     current_actor_context.feed(
         UsedTargetRepositories(
             repos=[
--- a/repos/system_upgrade/el7toel8/actors/localreposinhibit/tests/test_unit_localreposinhibit.py
+++ b/repos/system_upgrade/el7toel8/actors/localreposinhibit/tests/test_unit_localreposinhibit.py
@@ -1,0 +1,59 @@
+import pytest
+
+from leapp.models import (
+    RepositoryData,
+    RepositoryFile,
+    TMPTargetRepositoriesFacts,
+    UsedTargetRepositories,
+    UsedTargetRepository,
+)
+from leapp.snactor.fixture import ActorContext
+
+
+@pytest.mark.parametrize(
+    ("baseurl", "exp_msgs_len"),
+    [("file:///root/crb", 1), ("http://localhost/crb", 0)],
+)
+def test_unit_localreposinhibit(current_actor_context, baseurl, exp_msgs_len):
+    """Ensure the Report is generated when local path is used as a baseurl.
+
+    :type current_actor_context: ActorContext
+    """
+    current_actor_context.feed(
+        TMPTargetRepositoriesFacts(
+            repositories=[
+                RepositoryFile(
+                    file="the/path/to/some/file",
+                    data=[
+                        RepositoryData(
+                            name="BASEOS",
+                            baseurl=(
+                                "http://example.com/path/to/repo/BaseOS/x86_64/os/"
+                            ),
+                            repoid="BASEOS",
+                        ),
+                        RepositoryData(
+                            name="APPSTREAM",
+                            baseurl=(
+                                "http://example.com/path/to/repo/AppStream/x86_64/os/"
+                            ),
+                            repoid="APPSTREAM",
+                        ),
+                        RepositoryData(
+                            name="CRB", repoid="CRB", baseurl=baseurl
+                        ),
+                    ],
+                )
+            ]
+        )
+    )
+    current_actor_context.feed(
+        UsedTargetRepositories(
+            repos=[
+                UsedTargetRepository(repoid="BASEOS"),
+                UsedTargetRepository(repoid="CRB"),
+            ]
+        )
+    )
+    current_actor_context.run()
+    assert len(current_actor_context.messages()) == exp_msgs_len
--- a/repos/system_upgrade/el7toel8/actors/targetuserspacecreator/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/targetuserspacecreator/actor.py
@@ -4,7 +4,7 @@ from leapp.libraries.common.config import get_env, version
 from leapp.models import (CustomTargetRepositoryFile, RepositoriesMap, RequiredTargetUserspacePackages,
                           RHSMInfo, StorageInfo, TargetRepositories,
                           TargetUserSpaceInfo, UsedTargetRepositories,
-                          XFSPresence, Report)
+                          XFSPresence, Report, TMPTargetRepositoriesFacts)
 from leapp.tags import IPUWorkflowTag, TargetTransactionFactsPhaseTag
 
 
@@ -22,7 +22,7 @@ class TargetUserspaceCreator(Actor):
     name = 'target_userspace_creator'
     consumes = (CustomTargetRepositoryFile, RepositoriesMap, RequiredTargetUserspacePackages,
                 StorageInfo, RHSMInfo, TargetRepositories, XFSPresence)
-    produces = (TargetUserSpaceInfo, UsedTargetRepositories, Report)
+    produces = (TargetUserSpaceInfo, UsedTargetRepositories, Report, TMPTargetRepositoriesFacts, )
     tags = (IPUWorkflowTag, TargetTransactionFactsPhaseTag)
 
     def process(self):
--- a/repos/system_upgrade/el7toel8/actors/targetuserspacecreator/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/targetuserspacecreator/actor.py
@@ -6,8 +6,10 @@ from leapp.models import (CustomTargetRepositoryFile, RepositoriesMap, RequiredT
                           TargetUserSpaceInfo, UsedTargetRepositories,
                           XFSPresence, Report, TMPTargetRepositoriesFacts)
 from leapp.tags import IPUWorkflowTag, TargetTransactionFactsPhaseTag
+from leapp.utils.deprecation import suppress_deprecation
 
 
+@suppress_deprecation(TMPTargetRepositoriesFacts)
 class TargetUserspaceCreator(Actor):
     """
     Initializes a directory to be populated as a minimal environment to run binaries from the target system.
--- a/repos/system_upgrade/el7toel8/models/repositoriesfacts.py
+++ b/repos/system_upgrade/el7toel8/models/repositoriesfacts.py
@@ -29,7 +29,7 @@ class RepositoriesFacts(Model):
 
 
 @deprecated(
-    since="2020-02-01",
+    since="2020-09-01",
     message=(
         "The model is temporary and not assumed to be used in any "
         "other actors."
--- a/repos/system_upgrade/el7toel8/models/repositoriesfacts.py
+++ b/repos/system_upgrade/el7toel8/models/repositoriesfacts.py
@@ -1,5 +1,6 @@
 from leapp.models import Model, fields
 from leapp.topics import SystemFactsTopic
+from leapp.utils.deprecation import deprecated
 
 
 class RepositoryData(Model):
@@ -26,10 +27,18 @@ class RepositoriesFacts(Model):
 
     repositories = fields.List(fields.Model(RepositoryFile))
 
+
+@deprecated(
+    since="2020-02-01",
+    message=(
+        "The model is temporary and not assumed to be used in any "
+        "other actors."
+    ),
+)
 class TMPTargetRepositoriesFacts(RepositoriesFacts):
-    """
-    Do not consume this model anywhere outside of localreposinhibit
+    """Do not consume this model anywhere outside of localreposinhibit.
 
     The model is temporary and will be removed in close future
     """
+
     pass
--- a/repos/system_upgrade/el7toel8/models/repositoriesfacts.py
+++ b/repos/system_upgrade/el7toel8/models/repositoriesfacts.py
@@ -25,3 +25,11 @@ class RepositoriesFacts(Model):
     topic = SystemFactsTopic
 
     repositories = fields.List(fields.Model(RepositoryFile))
+
+class TMPTargetRepositoriesFacts(RepositoriesFacts):
+    """
+    Do not consume this model anywhere outside of localreposinhibit
+
+    The model is temporary and will be removed in close future
+    """
+    pass
--- a/repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
+++ b/repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
@@ -10,7 +10,7 @@ from leapp.libraries.stdlib import CalledProcessError, api, config, run
 from leapp.models import (CustomTargetRepositoryFile, RequiredTargetUserspacePackages, RHSMInfo,
                           StorageInfo, TargetRepositories, TargetUserSpaceInfo,
                           UsedTargetRepositories, UsedTargetRepository,
-                          XFSPresence)
+                          XFSPresence, TMPTargetRepositoriesFacts)
 
 # TODO: "refactor" (modify) the library significantly
 # The current shape is really bad and ineffective (duplicit parsing
@@ -421,6 +421,11 @@ def perform():
         with overlay.nspawn() as context:
             target_repoids = _gather_target_repositories(context, indata, prod_cert_path)
             _create_target_userspace(context, indata.packages, target_repoids)
+            # TODO: this is tmp solution as proper one needs significant refactoring
+            # refactorings..
+            target_repo_facts = repofileutils.get_parsed_repofiles(context)
+            api.produce(TMPTargetRepositoriesFacts(repositories=target_repo_facts))
+            # ## fixme ends here
             api.produce(UsedTargetRepositories(
                 repos=[UsedTargetRepository(repoid=repo) for repo in target_repoids]))
             api.produce(TargetUserSpaceInfo(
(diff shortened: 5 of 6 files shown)


ZhukovGreen:
e2e tests

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,80 @@
...
+                    "Local repository %r found. Referring to baseurl: %s. "
+                    "Currently leapp not supporting this option",
+                    custom_repo.repoid,
+                    custom_repo.baseurl,
+                )
+                reporting.create_report(
+                    [
+                        reporting.Title("Local repository identified."),
+                        reporting.Summary(
+                            "Local repository &s with baseurl %s were found. "
```
shaded-enmity, line 58: Missing string interpolation?

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,80 @@
...
+            if custom_repo.baseurl.startswith("file:"):
+                self.log.warning(
+                    "Local repository %r found. Referring to baseurl: %s. "
+                    "Currently leapp not supporting this option",
+                    custom_repo.repoid,
+                    custom_repo.baseurl,
+                )
+                reporting.create_report(
+                    [
+                        reporting.Title("Local repository identified."),
```
shaded-enmity, line 56:
```suggestion
                        reporting.Title("Local repository identified"),
```

ZhukovGreen:
@AloisMahdal pls help me investigate ht eproblem behin murphy ci

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,83 @@
...
+                                "your local repo via http. For details see "
+                                "external link."
+                            )
+                        ),
+                        reporting.ExternalLink(
+                            title=(
+                                "Customizing your Red Hat Enterprise Linux "
+                                "in-place upgrade"
+                            ),
+                            url="https://access.redhat.com/articles/4977891",
```
bocekm, line 79:
```suggestion
                            url="https://access.redhat.com/articles/4977891#repos-known-issues",
```

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,83 @@
...
+                    custom_repo.repoid,
+                    custom_repo.baseurl,
+                )
+                reporting.create_report(
+                    [
+                        reporting.Title("Local repository identified"),
+                        reporting.Summary(
+                            (
+                                "Local repository {} with baseurl {} were "
+                                "found. Currently leapp is not supporting "
```
bocekm, line 60:
```suggestion
                                "found. Currently leapp does not support "
```

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,83 @@
...
+                    "Currently leapp not supporting this option",
+                    custom_repo.repoid,
+                    custom_repo.baseurl,
+                )
+                reporting.create_report(
+                    [
+                        reporting.Title("Local repository identified"),
+                        reporting.Summary(
+                            (
+                                "Local repository {} with baseurl {} were "
```
bocekm, line 59:
```suggestion
                                "Local repository {} with baseurl {} "
```
It's singular. But the `was` can be omitted.

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,83 @@
...
+                                "this option."
+                            ).format(custom_repo.repoid, custom_repo.baseurl)
+                        ),
+                        reporting.Severity(reporting.Severity.HIGH),
+                        reporting.Tags([reporting.Tags.REPOSITORY]),
+                        reporting.Flags([reporting.Flags.INHIBITOR]),
+                        reporting.Remediation(
+                            hint=(
+                                "By using Apache HTTP Server you can expose "
+                                "your local repo via http. For details see "
```
bocekm, line 70:
```suggestion
                                "your local repository via http. For details see "
```

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,83 @@
...
+
+    name = "local_repos_inhibit"
+    consumes = (UsedTargetRepositories, TargetRepositories)
+    produces = (Report,)
+    tags = (IPUWorkflowTag, TargetTransactionChecksPhaseTag)
+
+    def process(self):
+        used_target_repos = next(
+            self.consume(UsedTargetRepositories)
+        ).repos  # type: List[UsedTargetRepository]
```
bocekm, line 37: Is there a reason to not have it on one line? Like:
```suggestion
        used_target_repos = next(self.consume(UsedTargetRepositories)).repos  # type: List[UsedTargetRepository]
```
The line below is the same.

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,83 @@
...
+        custom_target_repos = next(
+            self.consume(TargetRepositories)
+        ).custom_repos  # type: List[CustomTargetRepository]
+        for custom_repo in filter(  # pylint: disable-msg=deprecated-lambda
+            lambda custom_target_repo: custom_target_repo.repoid
+            in (used_tg_repo.repoid for used_tg_repo in used_target_repos)
+            and custom_target_repo.baseurl,
+            custom_target_repos,
+        ):
```
bocekm, line 46: This is super compact, but super unreadable. I suggest using a small and simple function instead:
```suggestion
        for custom_repo in _get_custom_repos():
```
Remember, clean code has functions with a few lines. And each function has a name that describes what it does. Then, you can read the function calls as easily as a story in a book.
We can debate what number of lines is optimal/maximum for a function, whether 10, 20 or 30. But this one has 50 lines.

AloisMahdal:
By *leapp-repository-0.10.0-0.202007171213Z.f4189aa.feature_inhibit_upgrade_if_local_repo_present.PR531.el7_6*:

    [2020-07-17T11:17:16-0400] phase.START check stray output # test
    [2020-07-17T11:17:16-0400]   assert.FAIL: there is no stray error output
    [2020-07-17T11:17:16-0400]   test.INFO: =====BEGIN PIPE STRAY=====
    [2020-07-17T11:17:16-0400]   test.INFO: Process Process-65:
    [2020-07-17T11:17:16-0400]   test.INFO: Traceback (most recent call last):
    [2020-07-17T11:17:16-0400]   test.INFO:   File "/usr/lib64/python2.7/multiprocessing/process.py", line 258, in _bootstrap
    [2020-07-17T11:17:16-0400]   test.INFO:     self.run()
    [2020-07-17T11:17:16-0400]   test.INFO:   File "/usr/lib64/python2.7/multiprocessing/process.py", line 114, in run
    [2020-07-17T11:17:16-0400]   test.INFO:     self._target(*self._args, **self._kwargs)
    [2020-07-17T11:17:17-0400]   test.INFO:   File "/usr/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 30, in inspect_actor
    [2020-07-17T11:17:17-0400]   test.INFO:     definition.load()
    [2020-07-17T11:17:17-0400]   test.INFO:   File "/usr/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 179, in load
    [2020-07-17T11:17:17-0400]   test.INFO:     self._module = importer.find_module(name).load_module(name)
    [2020-07-17T11:17:17-0400]   test.INFO:   File "/usr/lib64/python2.7/pkgutil.py", line 246, in load_module
    [2020-07-17T11:17:17-0400]   test.INFO:     mod = imp.load_module(fullname, self.file, self.filename, self.etc)
    [2020-07-17T11:17:17-0400]   test.INFO:   File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/localreposinhibit/actor.py", line 1, in <module>
    [2020-07-17T11:17:17-0400]   test.INFO:     from typing import List
    [2020-07-17T11:17:17-0400]   test.INFO: ImportError: No module named typing
    [2020-07-17T11:17:17-0400]   test.INFO: Inspection of actor in actors/localreposinhibit failed=====NO_NEWLINE_AT_EOF=====
    [2020-07-17T11:17:17-0400]   test.INFO: =====END PIPE STRAY=====
    [2020-07-17T11:17:17-0400] phase.END.FAIL (d=1) check stray output # test

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
ZhukovGreen: Considered, thx

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
ZhukovGreen: agree, thx

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
ZhukovGreen: I try to stick to 79, since, at first it is not violating 119 limits, second, it is much friendly for doing code reviews on mobile phones. Being a parent of two kids, this is a great advantage to me:)
Howe ver I know, sometimes the formatting becomes pretty ugly due to this constraint. This is an example:)

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
ZhukovGreen: Thanks for advising on this

ZhukovGreen:
e2e tests

pirat89:
@oamg/developers  @oamg/qe do not forget the PR should not be merged until the QE work is done. See jira tickets for more info.

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,86 @@
+from leapp import reporting
+from leapp.actors import Actor
+from leapp.models import TargetRepositories, UsedTargetRepositories
+from leapp.reporting import Report
+from leapp.tags import IPUWorkflowTag, TargetTransactionChecksPhaseTag
+
+
+class LocalReposInhibit(Actor):
+    """Inhibits the upgrade if local repositories were found.
```
pirat89, line 9:
```suggestion
    """
    Inhibits the upgrade if local repositories were found.
```

it's style used in majority code in leapp project it looks really much better. but not blocker from me. if you change it, change it everywhere in this PR, it's on more places.

pirat89:
@ZhukovGreen  I think that the solution does not cover all cases. I will check it properly next morning.

ZhukovGreen:
@pirat89 Sounds interesting

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,86 @@
...
+
+    Currently we're not supporting local repositories during the upgrade.
```
ZhukovGreen, line 11: @pirat89 I would like to make sure you checked this docstrings, because it is describes why only custom repos to be checked

pirat89: Yes, this is nothing against I wrote above. I am focused around custom repositories only as well.

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,86 @@
...
+        return filter(  # pylint: disable-msg=deprecated-lambda, filter-builtin-not-iterating
+            lambda custom_target_repo: custom_target_repo.repoid
+            in used_target_repos_ids
+            and custom_target_repo.baseurl,
+            custom_target_repos,
+        )
+
+    def process(self):
+        for custom_repo in self.get_used_custom_repos():
+            if custom_repo.baseurl.startswith("file:"):
```
pirat89, line 48: it should be checked against the real info on the system. repositories enabled using `--enablerepo` do not have set the baseurl in custom repos (for various valid reasons; e.g. the repository is not defined in the repodir before run of the leapp - e.g. when the repofile is generated/copied/... into the /etc/yum.repos.d dir dynamically during the run of the leapp; e.g if you want to enable specific RH repository that is not mapped by our repomap file, ...).

repos/system_upgrade/el7toel8/actors/localreposinhibit/tests/test_unit_localreposinhibit.py
```diff
@@ -0,0 +1,50 @@
...
+                        "http://example.com/path/to/repo/BaseOS/x86_64/os/"
+                    ),
+                    repoid="BASEOS",
+                ),
+                CustomTargetRepository(
+                    baseurl=(
+                        "http://example.com/path/to/repo/AppStream/x86_64/os/"
+                    ),
+                    repoid="APPSTREAM",
+                ),
```
pirat89, line 35: missing `CustomTargetRepository` without defined baseurl (only repoid is mandatory in case of this object)

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
ZhukovGreen: @pirat89  This style is accepted in https://www.python.org/dev/peps/pep-0257/#id17 . I will fix it in all other places . Thx

ZhukovGreen:
Marking it as a draft one again, since after the @pirat89 review it requires gradual changes

ZhukovGreen:
@pirat89 the proposed solution not covering the cases when the local repo was added to the leapp repofile. 
```python
ipdb> used_custom_repos
[]
ipdb> used_custom_repo_ids
set([u'APPSTREAM', u'CRB', u'BASEOS'])
```

Now thinking how to capture both options.

pirat89:
> @pirat89 the proposed solution not covering the cases when the local repo was added to the leapp repofile.
> 
> ```python
> ipdb> used_custom_repos
> []
> ipdb> used_custom_repo_ids
> set([u'APPSTREAM', u'CRB', u'BASEOS'])
> ```
> 
> Now thinking how to capture both options.

That's weird, it should cover it as the scan happens (or should happen) after the context is prepared (iow, it contains the custom file as well). But there is always way place for bugs. I haven't tested that yet.

ZhukovGreen:
@pirat89 I found how to resolve it. Now implementing it

ZhukovGreen:
@pirat89 Petr, please do a quick draft review of the concept. I tested all variants and it works for me. Tomorrow I will fix the tests and if you're ok, we can merge it

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,64 @@
...
+        target_repos = next(self.consume(TMPTargetRepositoriesFacts)).repositories
+        # fmt: on
+        target_repo_id_to_url_map = {
+            repo.repoid: repo.baseurl
+            for repofile in target_repos
+            for repo in repofile.data
+            if repo.baseurl
+        }
+        if any(
+            target_repo_id_to_url_map[repo.repoid].startswith("file:")
```
ZhukovGreen, line 31: I assuming that it is always the repo.repoid exists inside target_repo_id_to_url_map. If there could cases when this is not true, I have to redesign a code a bit @pirat89

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -1,86 +1,64 @@
...
+            for repo in used_target_repos
+        ):
+            warn_msg = (
+                "Local repository found (baseurl starts with file:///). "
+                "Currently leapp not supporting this option."
```
pirat89, line 36: typo
```suggestion
                "Currently leapp is not supporting this option."
```

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
pirat89: @ZhukovGreen no. In your case you could hit error because not every repository have baseurl.

repos/system_upgrade/el7toel8/models/repositoriesfacts.py
```diff
@@ -25,3 +25,11 @@ class RepositoriesFacts(Model):
     topic = SystemFactsTopic
 
     repositories = fields.List(fields.Model(RepositoryFile))
+
```
pirat89, line 28: add the deprecation and suprress deprecations decorators to make it more clear that this is deprecated to be used by anyone else.

pirat89:
> @pirat89 Petr, please do a quick draft review of the concept. I tested all variants and it works for me. Tomorrow I will fix the tests and if you're ok, we can merge it

Just one issue described in the comment above and keep in mind to set the deprecation for the temporary model, but the approach seems good to me otherwise.

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,64 @@
...
+    )
+    produces = (Report,)
+    tags = (IPUWorkflowTag, TargetTransactionChecksPhaseTag)
+
+    def process(self):
+        # fmt: off
+        used_target_repos = next(self.consume(UsedTargetRepositories)).repos
+        target_repos = next(self.consume(TMPTargetRepositoriesFacts)).repositories
+        # fmt: on
+        target_repo_id_to_url_map = {
```
ZhukovGreen, line 24: @bocekm Michal, what do you think to leave this logic here, without moving to a separate function? It is simplified and, from my POV, looks readable like it is.

repos/system_upgrade/el7toel8/models/repositoriesfacts.py
ZhukovGreen: github don't like forced pushes similar to me. Moving your comment to a right place

repos/system_upgrade/el7toel8/models/repositoriesfacts.py
```diff
@@ -25,3 +26,19 @@ class RepositoriesFacts(Model):
...
+
+@deprecated(
+    since="2020-02-01",
+    message=(
+        "The model is temporary and not assumed to be used in any "
+        "other actors."
+    ),
+)
+class TMPTargetRepositoriesFacts(RepositoriesFacts):
```
ZhukovGreen, line 38: @pirat89 
> add the deprecation and suprress deprecations decorators to make it more clear that this is deprecated to be used by anyone else. The second one should be used properly, so there will not be generated any message in the report about used deprecated model.

ZhukovGreen: @pirat89 I don't think the idea with suppress decorator is good. I would like to be notified if our actors use something that is going to be deprecated. Why suppress this?

bocekm: Because we deprecated something intentionally and per the deprecation process we still need to use it. I think it's correct if the warning about using something deprecated is shown only to someone outside of the Leapp team (external contributors, customers). If we don't suppress it, the warning would be shown to **everybody** who uses leapp for the next 6 months. 

The question is how does the **leapp** team get notified that the 6 months period has elapsed and we are good to remove it. This is something we need to figure out, but it's just an internal topic/problem IMHO.

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,64 @@
...
+    name = "local_repos_inhibit"
+    consumes = (
+        UsedTargetRepositories,
+        TMPTargetRepositoriesFacts,
+    )
+    produces = (Report,)
+    tags = (IPUWorkflowTag, TargetTransactionChecksPhaseTag)
+
+    def process(self):
+        # fmt: off
```
bocekm, line 20: Is this for the black plugin?

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
bocekm: Frankly, it's still less readable than I'd like to see.
Ideally, the condition could look like:
`if _file_baseurl_in_use()`
and put the lines 20-33 in that function.

repos/system_upgrade/el7toel8/models/repositoriesfacts.py
ZhukovGreen: this is exactly the reason, why I was thinking that there shouldn't be any difference between the leapp team and everybody. I saw many examples when something deprecated internally and still being raised as a deprecated warning to everyone (pytest, aiohttp of the top of my head...). Anyway, for me was important to bring this thought to the team.

ZhukovGreen: Added the suppress decorator in order do not slowdown the process of the review
https://github.com/oamg/leapp-repository/pull/531/commits/5af0c474211dcc05248b507da98e500906eefaca

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
ZhukovGreen: yes, I removed it now

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,72 @@
...
+    name = "local_repos_inhibit"
+    consumes = (
+        UsedTargetRepositories,
+        TMPTargetRepositoriesFacts,
+    )
+    produces = (Report,)
+    tags = (IPUWorkflowTag, TargetTransactionChecksPhaseTag)
+
+    def file_baseurl_in_use(self):
+        """Check if any of target repos is local.
```
pirat89, line 22: Just reminder, that the style used in the majority of code in leapp repositories is:
```
"""
Check if any....
"""

File: repos/system_upgrade/el7toel8/actors/targetuserspacecreator/actor.py [hunk 20]
```diff
@@ -22,7 +22,7 @@ class TargetUserspaceCreator(Actor):
     name = 'target_userspace_creator'
     consumes = (CustomTargetRepositoryFile, RepositoriesMap, RequiredTargetUserspacePackages,
                 StorageInfo, RHSMInfo, TargetRepositories, XFSPresence)
-    produces = (TargetUserSpaceInfo, UsedTargetRepositories, Report)
+    produces = (TargetUserSpaceInfo, UsedTargetRepositories, Report, TMPTargetRepositoriesFacts, )
```
pirat89, line 25: @ZhukovGreen Missing `@suppress_deprecation(TMPTargetRepositoriesFacts)` in this actor.

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
bocekm: It's not such a big deal, but you're right. It's covered in our python coding guidelines: https://leapp.readthedocs.io/en/latest/python-coding-guidelines.html#docstrings.

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,72 @@
...
+                    reporting.Title("Local repository identified"),
+                    reporting.Summary(warn_msg),
+                    reporting.Severity(reporting.Severity.HIGH),
+                    reporting.Tags([reporting.Tags.REPOSITORY]),
+                    reporting.Flags([reporting.Flags.INHIBITOR]),
+                    reporting.Remediation(
+                        hint=(
+                            "By using Apache HTTP Server you can expose "
+                            "your local repository via http. For details "
+                            "see external link."
```
bocekm, line 58:
```suggestion
                            "your local repository via http. See the linked article for details."
```

repos/system_upgrade/el7toel8/models/repositoriesfacts.py
```diff
@@ -25,3 +26,19 @@ class RepositoriesFacts(Model):
     topic = SystemFactsTopic
 
     repositories = fields.List(fields.Model(RepositoryFile))
+
+
+@deprecated(
+    since="2020-02-01",
```
bocekm, line 32: Do we really want to cheat the deprecation process this way? I'd just use the current date.

ZhukovGreen: @bocekm I was trying to follow the draft version of the docs related to the deprecation process (https://github.com/pirat89/leapp/commit/92690c36c2b61b7ce09eb268e839b504d8d1806d#diff-42e97d6a6a1f32cee83f15fb67893b98R17-R18) .  Thus I added this minimum 6 month period. However, there is no difference, wether it is today or today + 180 days. both opttions will result in the report with the severity HIGH (https://github.com/oamg/leapp/commit/b8f11beb060b589c88ac6e93479e357b0607889a#diff-46ef8e2c58eae2fcd84596e29d65b614R302-R303)

pirat89: @ZhukovGreen From my POV, I do not see reason to set any lower risk during the 6m period - as after 6m period the functionality will be whenever dropped. I will just update the documentation probably. But it can be discussed e.g. on cabal mtg. @oamg/developers

pirat89: ha. I have forgotten to put the topic on our cabal mtg. my bad :( I will do it now

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
AloisMahdal: It's actually inconsistent in this one string

    """Check if any
    
    Lorem ipsum blah blah..
    """

ZhukovGreen: This is the most basic python convention for multiline docstrings. Why do you need to irritate people with this exception? A lot of IDEs utilities rely on this convention. As well as the sphinx autodoc. I acknowledge that this is now in our guidelines, so my current message is a kind of product of emotions.

ZhukovGreen:
e2e tests

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
bocekm: If I recall correctly the reasoning back then was related to the fact that the first line would have no whitespaces before the sentence but all the other lines of the multiline string would have X number of whitespaces. We were planning to parse it or already do parse it (maybe for the [dashboard](https://oamg.github.io/leapp-dashboard/), @vinzenz?) and this whitespace inconsistency could cause troubles.

repos/system_upgrade/el7toel8/models/repositoriesfacts.py
bocekm: I'm ok with leaving the functionality in place for 6 months, so I'd use the "current" date, e.g. 2020-09-01. Let's deal with it now to not block the PR. Waiting for cabal would be just wasting last days before the final release.

ZhukovGreen:
e2e tests

AloisMahdal:
Looks almost ready; test is also ready; I ran verification test with via **murphy-ci/ver** batch.

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,72 @@
...
+    def process(self):
+        if self.file_baseurl_in_use():
+            warn_msg = (
+                "Local repository found (baseurl starts with file:///). "
+                "Currently leapp is not supporting this option."
+            )
+            self.log.warning(warn_msg)
+            reporting.create_report(
+                [
+                    reporting.Title("Local repository identified"),
```
bocekm, line 49:
```suggestion
                    reporting.Title("Local repository detected"),
```

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,72 @@
...
+        return any(
+            target_repo_id_to_url_map[repo.repoid].startswith("file:")
+            for repo in used_target_repos
+        )
+
+    def process(self):
+        if self.file_baseurl_in_use():
+            warn_msg = (
+                "Local repository found (baseurl starts with file:///). "
+                "Currently leapp is not supporting this option."
```
bocekm, line 44:
```suggestion
                "Currently leapp does not support this option."
```

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
```diff
@@ -0,0 +1,72 @@
...
+                    reporting.Tags([reporting.Tags.REPOSITORY]),
+                    reporting.Flags([reporting.Flags.INHIBITOR]),
+                    reporting.Remediation(
+                        hint=(
+                            "By using Apache HTTP Server you can expose "
+                            "your local repository via http. See the linked "
+                            "article for details. "
+                        )
+                    ),
+                    reporting.ExternalLink(
```
bocekm, line 61: The external link does not show up in the txt report. @shaded-enmity, @pirat89, is that expected?
```
Risk Factor: high (inhibitor)
Title: Local repository identified
Summary: Local repository found (baseurl starts with file:///). Currently leapp is not supporting this option.
Remediation: [hint] By using Apache HTTP Server you can expose your local repository via http. See the linked article for details. 
```

ZhukovGreen: I see at https://github.com/oamg/leapp/blob/af8730d5076659bcdc4ab0c2b09b6b408e95dcd3/leapp/utils/report.py#L91 that this is looks expected. JSON report should be complete, but txt one omitting some fields

ZhukovGreen:
@AloisMahdal merging?

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
pirat89: I guess that the `strip` of the string could resolve such problems if we have any, but not sure as I am not familiar with python parser. My point is mainly about the `aesthetics`. And the majority of our code uses the new line in docstring, so to be consistent. I do not see so big benefit to change our coding guidelines to be honest. One more enter per docstring is not something what seems too much annoying in editor from my POV. And even when I am fun of the newline start in docstring (it really looks much better..) I am not telling that this is blocker. Just it looks really weird from my POV.

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
pirat89: @bocekm It's expected now. I already raised that topic in the past, to print the external link in .txt report as well, but not sure why it is not there till this day. I guess that it's just because we missed that. But not sure really.

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
pirat89: Btw, one of topics about that I remeber was that many people want to put empty space before and after the docstring
```
""" Docstring. """
```
Using the new line always automatically resolves this problem and you have not to write people they have to remove the whitespaces around the docstring.

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
ZhukovGreen: ref https://github.com/oamg/leapp-repository/issues/559

ZhukovGreen: Any further discussion, let's keep in #559, because it is out of the scope of this MR

repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
ZhukovGreen: If it is not breaking something in dashboard (@vinzenz ), then I suggest closing this thread in a separate ticket, because if I am testing now standard flake8 pre-commit hooks, with the docstrings in the following form
```
"""
Some sdsdsd
sdsdsds
""""
```

I see the error:
```
D205 1 blank line required between summary line and description
D400 First line should end with a period
```
So the user won't be able to commit the wrong form of the docstrings

pirat89: Yes, but it's because it should be correctly like this:
```
Some sdsdsd.

sdsdsds
```
right?

ZhukovGreen: @pirat89 now I see, you're right. So the docstrings linter accepts your variant and I'm really surprised about it. So both versions work ok:
```
"""Summary.

details
"""
```
and
```
"""
Summary.

details
"""
```

AloisMahdal:
I'm investigating the latest fail.

AloisMahdal:
For some reason in the second case leapp won't inhibit on a `file://` repo like this:

    [myBASEOS]
    name=myBASEOS
    baseurl=file:////tmp/xcase.runpath.2yAisXWw/xcase-relics/repo_enabled_from_leapp_cmdline/repo
    enabled=0
    gpgcheck=0

the extra slashes are weird but i assume they are valid (because `//tmp` is a valid path).  Not 100% sure if it's because of this.

pirat89:
@AloisMahdal Extras slashes should not be problem unless DNF has troubles with them.

AloisMahdal:
Turns out that for the second test case the repo must be valid *and* point to a path on a non-virtual fs.  In my case the repo was in /tmp which caused dnf/leapp to blow up earlier.  Thanks @pirat89 for help with investigation.

I've moved the repo to rootfs and re-ran the **murphy-ci/ver**.

AloisMahdal:
Finally, passed with *leapp-repository-0.10.0-0.202008071648Z.aadb76c.feature_inhibit_upgrade_if_local_repo_present.PR531.el7_6*

Rezney:
Please bear in mind that there can be also a mirrorlist instead of baseurl. Not sure if the original issue applies also to mirrorlists but it breaks the upgrade when they are used (e.g. in AWS).

```
        target_repo_id_to_url_map = {
            repo.repoid: repo.baseurl
            for repofile in target_repos
            for repo in repofile.data
            if repo.baseurl
        }
```

Above we gather only baseurl repos in the map but then we try to access it by repoid from target repos (with mirrorlist) which fails:

```
  File "/usr/lib/python2.7/site-packages/leapp/utils/deprecation.py", line 32, in process_wrapper
    return target_item(*args, **kwargs)
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/localreposinhibit/actor.py", line 44, in process
    if self.file_baseurl_in_use():
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/localreposinhibit/actor.py", line 40, in file_baseurl_in_use
    for repo in used_target_repos
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/localreposinhibit/actor.py", line 40, in <genexpr>
    for repo in used_target_repos
KeyError: u'rhui-client-config-server-8'
```

@AloisMahdal let's try to add also mirrorlists to the tests.

bocekm:
Thanks for the report, Michal. The repo mirrorlists haven't been considered in this solution. I've reported the issue internally as OAMG-3795.

Rezney:
@bocekm thanks for the info. I was just about to start creating ticket :)

pirat89:
@bocekm @Rezney The mirrorlist and metalink have not been considered to handle at all, as use of these does not make any sense for local repos.

PR #533, opened 2020-07-17
Make authconfig->authselect actors non-experimental

bocekm:
- Re-reverting the commit 51a93e7 that was reverted by e30fdc0
- The dialogs feature of the leapp framework is working like a charm now

The original message from Ina:
- With introduction of new dialogs processing mechanism these actors
can finally be included in the workflow.
May cause the need to update upgrade testing automation as in several
cases (openstack rhel 7.6 image) these actors are hit 100% of the time
so adding a call to `leapp answer` will be inevitable.

Co-authored-by: Inessa Vasilevskaya <ivasilev@redhat.com>


examon:
leapp-ci build

examon:
leapp-ci build

PR #534, opened 2020-07-22
fix: Make PreparePythonWorkround more robust

vinzenz:
Previously multiple executions of leapp after a failing in the initram
phase caused this actor to fail. This patch fixes up the issue and
additionally adds some unit test.

BZ-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1858479
OAMG-Issue: OAMG-3640


vinzenz:
e2e tests

pirat89:
Ah, I think I see the problem. It's probably connected to issue in tests related to removed/moved /var/log/leapp directory. It should be already fixed. I will re-run the tests

pirat89:
e2e tests

PR #537, opened 2020-07-24
add IBM z15 systems as supported by RHEL 8

sharkcz:


pirat89:
Green linghts. Merging

PR #540, opened 2020-07-30
Multi-command remediations are not supported

shaded-enmity:
Before we have a mechanism to provide users with either/or choice
for remediations we need to make sure that there is always one and only
one remediation command

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/kernel/checkinstalleddevelkernels/libraries/checkinstalleddevelkernels.py
+++ b/repos/system_upgrade/el7toel8/actors/kernel/checkinstalleddevelkernels/libraries/checkinstalleddevelkernels.py
@@ -27,8 +27,10 @@ def process():
         summary = ('DNF cannot produce a valid upgrade transaction when'
                    ' multiple kernel-devel packages are installed.')
         hint = ('Remove all but one kernel-devel packages before running Leapp again.')
-        commands = ([['yum', '-y', 'remove', '{n}-{v}.{r}'.format(
-            n=pkg.name, v=pkg.version, r=pkg.release)] for pkg in pkgs[:-1]])
+        all_but_latest_kernel_devel = pkgs[:-1]
+        packages = ['{n}-{v}-{r}'.format(n=pkg.name, v=pkg.version, r=pkg.release)
+                    for pkg in all_but_latest_kernel_devel]
+        commands = [['yum', '-y', 'remove'] + packages]
         reporting.create_report([
             reporting.Title(title),
             reporting.Summary(summary),
--- a/repos/system_upgrade/el7toel8/actors/reportsettargetrelease/libraries/reportsettargetrelease.py
+++ b/repos/system_upgrade/el7toel8/actors/reportsettargetrelease/libraries/reportsettargetrelease.py
@@ -25,11 +25,10 @@ def _report_set_release():
 def _report_unhandled_release():
     # TODO: set the POST group after it's created.
     target_version = api.current_actor().configuration.version.target
-    commands = [
-        ['subscription-manager', 'release', '--unset'],
-        ['subscription-manager', 'release', '--set', target_version],
-    ]
-    hint = 'Set the new release (or unset it) after the upgrade using subscription-manager.'
+    hint_command = 'subscription-manager release --set {}'.format(target_version)
+    # FIXME: This should use Dialogs and Answers to offer post-upgrade remediation
+    # so that users can choose whether to --set or --unset the release number
+    hint = 'Set the new release (or unset it) after the upgrade using subscription-manager: ' + hint_command
     reporting.create_report([
         reporting.Title(
             'The subscription-manager release is going to be kept as it is during the upgrade'),
@@ -41,7 +40,7 @@ def _report_unhandled_release():
             ' after the upgrade.'
         ),
         reporting.Severity(reporting.Severity.LOW),
-        reporting.Remediation(commands=commands, hint=hint),
+        reporting.Remediation(hint=hint),
         reporting.Tags([reporting.Tags.UPGRADE_PROCESS]),
         reporting.RelatedResource('package', 'subscription-manager')
     ])
--- a/repos/system_upgrade/el7toel8/actors/checkrootsymlinks/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkrootsymlinks/actor.py
@@ -28,6 +28,10 @@ class CheckRootSymlinks(Actor):
         absolute_links = [item for item in rootdir.items if item.target and os.path.isabs(item.target)]
 
         if absolute_links:
+            commands = [' '.join(['ln', '-snf',
+                                  os.path.relpath(item.target, '/'),
+                                  os.path.join('/', item.name)]) for item in absolute_links]
+            remediation = [['sh', '-c', ' && '.join(commands)]]
             reporting.create_report([
                 reporting.Title('Upgrade requires links in root directory to be relative'),
                 reporting.Summary(
@@ -37,7 +41,5 @@ class CheckRootSymlinks(Actor):
                 ),
                 reporting.Severity(reporting.Severity.HIGH),
                 reporting.Flags([reporting.Flags.INHIBITOR]),
-                reporting.Remediation(commands=[['ln', '-snf',
-                                                 os.path.relpath(item.target, '/'),
-                                                 os.path.join('/', item.name)] for item in absolute_links])
+                reporting.Remediation(commands=remediation)
             ])
--- a/repos/system_upgrade/el7toel8/actors/verifydialogs/libraries/verifydialogs.py
+++ b/repos/system_upgrade/el7toel8/actors/verifydialogs/libraries/verifydialogs.py
@@ -12,8 +12,9 @@ def check_dialogs(inhibit_if_no_userchoice=True):
         dialog_resources = [reporting.RelatedResource('dialog', s) for s in sections]
         dialogs_remediation = ('Please register user choices with leapp answer cli command or by manually editing '
                                'the answerfile.')
+        # FIXME: Enable more choices once we can do multi-command remediations
         cmd_remediation = [['leapp', 'answer', '--section', "{}={}".format(s, choice)]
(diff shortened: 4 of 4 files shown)


repos/system_upgrade/el7toel8/actors/reportsettargetrelease/libraries/reportsettargetrelease.py
```diff
@@ -26,8 +26,9 @@ def _report_unhandled_release():
     # TODO: set the POST group after it's created.
     target_version = api.current_actor().configuration.version.target
     commands = [
-        ['subscription-manager', 'release', '--unset'],
-        ['subscription-manager', 'release', '--set', target_version],
+        ['sh', '-c',
+         ('subscription-manager release --unset &&'
+          'subscription-manager release --set {}'.format(target_version))]
```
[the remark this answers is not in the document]
shaded-enmity: I thought we needed to execute both, so I used `&&` to string the actions together. However, `&&` is a shell primitive, hence the `sh -c 'command1 && command2'`. If we need to execute just one `sh -c` becomes redundant. 

The question now is, do we want to suggest unset or set? :)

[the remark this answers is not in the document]
bocekm: I expect (maybe blindly) that the default/typical scenario is to not have the release locked. So I vote for the unset.

> I thought we needed to execute both, so I used && to string the actions together. However, && is a shell primitive, hence the sh -c 'command1 && command2'. If we need to execute just one sh -c becomes redundant.

I see. Yeah, I can't come up with a better solution. If @pirat89 can't as well, then I think it's ok :)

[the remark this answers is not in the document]
pirat89: yep. Just one of them should be executed. the execution of both of them is nonsense. This is one of cases which we discussed we will create multiple remediation instructions per report, to be able to tell that just one of cmds should be executed.  In case of systems waiting for certifications (or in case of EUS) it is expected to have locked release to a specific release. Installing the newest one, menas usually that such users will miss certificatipon they need.

repos/system_upgrade/el7toel8/actors/kernel/checkinstalleddevelkernels/libraries/checkinstalleddevelkernels.py
```diff
@@ -27,8 +27,8 @@ def process():
         summary = ('DNF cannot produce a valid upgrade transaction when'
                    ' multiple kernel-devel packages are installed.')
         hint = ('Remove all but one kernel-devel packages before running Leapp again.')
-        commands = ([['yum', '-y', 'remove', '{n}-{v}.{r}'.format(
-            n=pkg.name, v=pkg.version, r=pkg.release)] for pkg in pkgs[:-1]])
+        packages = ['{n}-{v}.{r}'.format(n=pkg.name, v=pkg.version, r=pkg.release) for pkg in pkgs[:-1]]
+        commands = [['yum', '-y', 'remove'] + packages]
```
pirat89, line 31: @shaded-enmity This is wrong too. This will lead in sutation you remove all devel-kernels. But you are supposed to pick one that should be kept and remove the others. This command does a different think.

pirat89: last change made it clear completely.

repos/system_upgrade/el7toel8/actors/reportsettargetrelease/libraries/reportsettargetrelease.py
```diff
@@ -25,11 +25,10 @@ def _report_set_release():
...
     target_version = api.current_actor().configuration.version.target
-    commands = [
-        ['subscription-manager', 'release', '--unset'],
-        ['subscription-manager', 'release', '--set', target_version],
-    ]
-    hint = 'Set the new release (or unset it) after the upgrade using subscription-manager.'
+    hint_command = 'subscription-manager release --set {}'.format(target_version)
+    # FIXME: This should use Dialogs and Answers to offer post-upgrade remediation
+    # so that users can choose whether to --set or --unset the release number
+    hint = 'Set the new release (or unset it) after the upgrade using subscription-manager: ' + hint_command
```
pirat89, line 31: the used command refers just to the set of the release. What about something like:
```suggestion
    hint_cmd_set = 'subscription-manager release --set {}'.format(target_version)
    hint_cmd_unset = 'subscription-manager release --unset'
    # FIXME: This should use Dialogs and Answers to offer post-upgrade remediation
    # so that users can choose whether to --set or --unset the release number
    hint = 'Set the new release (cmd [1]) or unset it (cmd [2]) after the upgrade using subscription-manager: \n[1] {set_cmd}\n[2] {unset_cmd}'.format(set_cmd=hint_cmd_set, unset_cmd=hint_cmd_unset)
```
Formatting needs a love little bit to meet <120 chars per line.

pirat89: To make it clear, the current messaging is not blocker of this PR. Just pointing this out in case you like that more.

bocekm:
e2e tests

repos/system_upgrade/el7toel8/actors/kernel/checkinstalleddevelkernels/libraries/checkinstalleddevelkernels.py
```diff
@@ -27,8 +27,10 @@ def process():
         summary = ('DNF cannot produce a valid upgrade transaction when'
                    ' multiple kernel-devel packages are installed.')
         hint = ('Remove all but one kernel-devel packages before running Leapp again.')
-        commands = ([['yum', '-y', 'remove', '{n}-{v}.{r}'.format(
-            n=pkg.name, v=pkg.version, r=pkg.release)] for pkg in pkgs[:-1]])
+        all_but_latest_kernel_devel = pkgs[:-1]
+        packages = ['{n}-{v}.{r}'.format(n=pkg.name, v=pkg.version, r=pkg.release)
```
bocekm, line 31:
```suggestion
        packages = ['{n}-{v}-{r}'.format(n=pkg.name, v=pkg.version, r=pkg.release)
```
There's a bug:
```
Remediation: [hint] Remove all but one kernel-devel packages before running Leapp again.
[command] yum -y remove kernel-devel-3.10.0.1062.4.3.el7
```
There should be a dash instead of period. This is the correct pkg name: `kernel-devel-3.10.0-1062.4.3.el7`.

bocekm: @AloisMahdal, I think it would be a great addition to the tests. If there's a remediation command, test that when applied, it fixes the reported issue.

pirat89:
@shaded-enmity  btw, dialogs are not covered and creates report with two commands as well.
```
   {   
      "severity": "high", 
      "title": "Missing required answers in the answer file", 
      "timeStamp": "2020-08-07T14:40:06.137283Z", 
      "hostname": "localhost.localdomain", 
      "detail": {
        "related_resources": [
          {
            "scheme": "dialog", 
            "title": "remove_pam_pkcs11_module_check.confirm"
          }
        ],
        "remediations": [
          {
            "type": "hint", 
            "context": "Please register user choices with leapp answer cli command or by manually editing the answerfile."
          },
          {
            "type": "command", 
            "context": [
              "leapp", 
              "answer", 
              "--section", 
              "remove_pam_pkcs11_module_check.confirm=True"
            ]
          },
          {
            "type": "command", 
            "context": [
              "leapp", 
              "answer", 
              "--section", 
              "remove_pam_pkcs11_module_check.confirm=False"
            ]
          }
        ]
      },  
      "actor": "verify_check_results", 
      "summary": "One or more sections in answerfile are missing user choices: remove_pam_pkcs11_module_check.confirm\nFor more information consult https://leapp.readthedocs.io/en/latest/dialogs.html", 
      "audience": "sysadmin", 
      "flags": [
        "inhibitor"
      ],  
      "id": "656ac5dc426d068395f4a41dccb54bfb5fe4ca632441507d26a9df3d26e43592"
    },

```

AloisMahdal:
This branch is made from outdated master (that's why integration fails.)

AloisMahdal:
The integration fail is irrelevant (caused by chronyd AVC which we only see once in a blue moon).  Rest of the integration is OK.

AloisMahdal:
> The integration fail is irrelevant (caused by chronyd AVC which we only see once in a blue moon). Rest of the integration is OK.

to clarify, none of the above constitutes verification.  verification is separate; still working on it..

AloisMahdal:
#531 got merged in the meantime so this branch is behind.  Because its test already got promoted to murphy-ci/integration, now the batch is failing.  Could you please rebase again?

repos/system_upgrade/el7toel8/actors/kernel/checkinstalleddevelkernels/libraries/checkinstalleddevelkernels.py
bocekm: OAMG-3761 created based on my comment above.

bocekm:
Rebased.

AloisMahdal:
e2e tests

AloisMahdal:
leapp-ci build

AloisMahdal:
1. extended tests for touched 3 actors to produce multiple commands
 2. added common check to assert that there are no multiple command "objects" in JSON
 3. re-ran all integration tests and upgrade tests with *leapp-repository-0.10.0-0.202008131746Z.f52d0ce.remedies.PR540.el7_6* on rhel-7.9 x86_64 and rhel-alt-7.6 aarch64

All *passed*.  Ie. **ack** for merging f52d0ce.

AloisMahdal:
Also note that **continuous-integration/tft-ci/e2e** fails because of test issue and can be waived.  (AFAIK **murphy-ci/upgrade** is super-set of the tft/e2e test so virtually no coverage is lost.)

PR #525, opened 2020-07-08
Get repofile directories from dnf config, do not search subdirectories

drehak:
yum/dnf does not look into subdirectories, so leapp should not do that either to avoid reporting false poitives.

PR #541, opened 2020-07-30
Skip test which rely on dnf package if dnf is unavailable

ZhukovGreen:
This is a small MR linked with recent #525 (already merged). After merging #525, the tests are failing on systems which don't have dnf package.

This MR resolves this problem


drehak:
Is it okay to skip the test? IMHO a better solution would be just to mock `get_repodirs` as a function or a lambda returning `['/etc/yum.repos.d']`. That's the only part that requires importing `dnf`, so the rest of the should run just like before.

drehak:
Oh, I see it's merged now. Nevermind then.

bocekm:
I also think this could be done better. Setting a constant in the code which is used **only** in a test seems unclean to me.

ZhukovGreen:
@bocekm @drehak agree with you guys, I did it too fast. Fixing it right now

ZhukovGreen:
@fellipeh sorry Fellipe, for urging you

Issue #544, opened 2020-07-30
Improve clarity and provide useful remediations when user uses optional rhel7 repos


ZhukovGreen:
The current situation is that our users are confused by the messages related to filtering out unsupported repositories.

# Acceptance criteria
- [ ] Make clear messages why do we removing the optional repos
- [ ] Provide report messages
- [ ] Provide useful remediation

ZhukovGreen:
Initial god MR was here https://github.com/oamg/leapp-repository/pull/519

PR #543, opened 2020-07-30
Chore/improve msgs about transaction changes

ZhukovGreen:
ref #544 

First MR which improves the clarity of the messages regarding unsupported repos

P.S. Bumping the requirements is required due to outdated caplog fixture + after merged #524 we are no longer restricted to pytest 3.6.4

PR #545, opened 2020-07-30
Chore/rename blacklist

ZhukovGreen:
Ref #544 

This MR is next in raw after #543 and is optional, but we decided in team to make it happen.

# TODO
- [x] merge #543 and rebase on top
- [ ] ensure deprecation policy is applied
- [ ] make sure all the references of the blacklisted renamed after the rebasing

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/peseventsscanner.py
+++ b/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/peseventsscanner.py
@@ -9,7 +9,7 @@ from leapp.libraries.common.config import architecture, version
 from leapp.libraries.stdlib import api
 from leapp.libraries.stdlib.config import is_verbose
 from leapp.models import (InstalledRedHatSignedRPM, PESRpmTransactionTasks, RepositoriesMap,
-                          RepositoriesSetupTasks, RpmTransactionTasks, RepositoriesBlacklisted)
+                          RepositoriesSetupTasks, RpmTransactionTasks, RepositoriesExcluded)
 
 
 Event = namedtuple('Event', ['id',            # int
@@ -407,7 +407,7 @@ def process_events(releases, events, installed_pkgs):
 
     map_repositories(tasks[Task.install])
     map_repositories(tasks[Task.keep])
-    filter_out_pkgs_in_blacklisted_repos(tasks[Task.install])
+    filter_out_pkgs_in_excluded_repos(tasks[Task.install])
     resolve_conflicting_requests(tasks)
 
     return tasks
@@ -448,29 +448,34 @@ SKIPPED_PKGS_MSG = (
 )
 
 
-def filter_out_pkgs_in_blacklisted_repos(to_install):
+def filter_out_pkgs_in_excluded_repos(to_install):
     """
-    Do not install packages that are available in blacklisted repositories
+    Do not install packages that are available in excluded repositories
 
-    No need to filter out the to_keep packages as the blacklisted repos won't get enabled - that is ensured in the
+    No need to filter out the to_keep packages as the excluded repos won't get enabled - that is ensured in the
     setuptargetrepos actor. So even if they fall into the 'yum upgrade' bucket, they won't be available thus upgraded.
     """
     # FIXME The to_install contains just a limited subset of packages - those that are *not* currently installed and
     #   are to be installed. But we should also warn about the packages that *are* installed.
-    blacklisted_pkg_repo_pairs = set()
-    blacklisted_repos = get_repositories_blacklisted()
+    excluded_pkg_repo_pairs = set()
+    excluded_repos = get_excluded_repositories()
     for pkg, repo in to_install.items():
-        if repo in blacklisted_repos:
-            blacklisted_pkg_repo_pairs.add((pkg, repo))
+        if repo in excluded_repos:
+            excluded_pkg_repo_pairs.add((pkg, repo))
 
-    for pkg, _ in blacklisted_pkg_repo_pairs:
+    for pkg, _ in excluded_pkg_repo_pairs:
         del to_install[pkg]
 
-    if blacklisted_pkg_repo_pairs:
+    if excluded_pkg_repo_pairs:
         report_skipped_packages(
             title='Packages available in excluded repositories will not be installed',
             message=SKIPPED_PKGS_MSG,
-            package_repo_pairs=blacklisted_pkg_repo_pairs,
+            package_repo_pairs=excluded_pkg_repo_pairs,
+            remediation=(
+                'At your own risk! Use leapp upgrade {}'.format(' '.join(
+                    ['--enablerepo ' + repo for repo in excluded_repos])
+                )
+            )
         )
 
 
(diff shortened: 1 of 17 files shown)


AloisMahdal:
Are we really renaming just "blacklist"?  shouldn't we rename the whole "blacklist/whitelist" pairs?  cc @bocekm.

If this is a global action, please also make sure the replacements are clear with regards to other side of the concept and other terms already used.  I mean, eg. in some places "exclude" may conflict with some component-related term, or similar case may arise for "include".

(Maybe you already did all of that, in that case ignore my comment.)

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/peseventsscanner.py
```diff
@@ -475,44 +486,55 @@ def map_repositories(packages):
 
     if repo_without_mapping:
         report_skipped_packages('packages will not be installed or upgraded due to repositories unknown to leapp:',
-                                repo_without_mapping)
+                                repo_without_mapping,
+                                "Please file a bug in http://bugzilla.redhat.com/ for leapp-repository component of"
+                                "the Red Hat Enterprise Linux 7 product.")
```
bocekm, line 491: This and many other changes to this file do not pertain to the PR topic - renaming blacklist.

repos/system_upgrade/el7toel8/actors/redhatsignedrpmcheck/libraries/redhatsignedrpmcheck.py
```diff
@@ -8,22 +8,21 @@
...
-    unsigned_packages = ' '.join(packages)
-    remediation = ['yum', '-y', 'remove', '{}'.format(unsigned_packages)]
-    title = 'Packages not signed by Red Hat found in the system'
-    summary = 'The following packages have not been signed by Red Hat ' \
-              'and may be removed in the upgrade process:\n{}'.format(unsigned_packages_new_line)
+    title = 'Packages not signed by Red Hat found on the system'
+    summary = ('The following packages have not been signed by Red Hat'
+               ' and may be removed during the upgrade process in case Red Hat-signed'
+               ' packages to be removed during the upgrade depend on them:\n{}'
+               .format(unsigned_packages_new_line))
```
bocekm, line 19: Also, this file changes are not related to blacklist renaming.

ZhukovGreen: @bocekm It is too early to review it. Since it also contains commits from #543

ZhukovGreen:
e2e tests

ZhukovGreen:
@pirat89 could you be more specific here, please. Actor RepositoriesExcluded already uses the new model, so no actions here. Then, we're renaming RepositoriesBlacklisted model to RepositoriesExcluded. Our `external API` assumes the model itself, so I don't remove RepositoriesBlacklisted, but, marking it as deprecated. What else should remain?

pirat89:
@ZhukovGreen the same thing I am repeating all the time: keep its functionality ;-) When I use RepositoriesBlacklisted model, will be the functionality the same as before this PR? If not, the model is not deprecated correctly.

bocekm:
Just to make it clear so that you're not under stress. This does not need to go to the next release. Ccing @shaded-enmity in case he thinks otherwise.

ZhukovGreen:
Sorry, @pirat89 I disagree to follow something that is not published yet. I want to wait when we'll actually describe our deprecation policy. Because I have another opinion on this topic. What you're suggesting seems too complex. I think it is enough to make an old model just as pointer to the new one, with adding a deprecated decorator. Then refactor all the actors which are using the old model, to make them use the new one. If someone decided to consume the old model, then he will be notified that it is deprecated and use the new one. In this case, there is no reason for him to keep using the deprecated and he will switch to the new one. And old actors already refactored, so no need to produce the old model at all

pirat89:
@ZhukovGreen  but the deprecation process is already published. Just it was not described so explicitly as it was expected originally that it is clear. changes I am doing, are about to clarify it better. In this PR, the deprecation is not implemented as the functionality is not kept. When the functionality is not kept, you do not need to refer to anything, because it's just incompatible change. It's even more safe to drop the functionality completely and keep application crashing immadiately. Imagine that user is using msg A to do something totally crucial to be able to boot on the upgraded system. Imagine the model A is kept, functionality is dropped or replaced by model B. in such case, user will do the upgrade and....ends with unbootable system. this is unacceptable. as well, it is not acceptable to change something from day to day for user. That's why the deprecation process exists. Only difference in our project in comparison to other projects is, that in oether projects the functionality is dropped based on a release version (e.g. in case of semantic versioning, functionality is usually kept for the whole major version, and dropped with bump of the major version). In our project is dropped in version, that is created after the specific amount of type. That's basically only difference.

bocekm:
I kind of like the Artem's approach from https://github.com/oamg/leapp-repository/pull/545#issuecomment-668547457. It would keep the functionality.
With every release, we ourselves inevitable change behaviour of our actors which also consume certain messages.
Petr, it seems to me you are more strict about keeping all the functionality that follows consumption of a deprecated model than about keeping functionality that does not consume anything deprecated. But maybe I don't get clearly all the aspects you're trying to explain.
I think this is a good topic to discuss with the whole team on Cabal.

pirat89:
@bocekm I do not understand what yare you trying to tell. Maybe we could sync over video call, to sync better. Anyway, the cabal, ML or different mtg for a discussion is good idea. From my pov it's more political than technical discussion, as the technical POV is clear about what the deprecation means really. I have more feeling we are discussing just political stuff here to do something differently than it's typical meaning of deprecation. So ML or separate mtg seems better for this. But unless we have anything else for cabal, why not use that time?

For the purpose of the discussion, I will try to give you this one example of deprecation - think please about this case before the discussion: Imagine deprecation of RPM inside a specific RHEL X major version. E.g. ifconfig. That one is deprecated, but people could use the `ip` utility from the `iproute` rpm instead. Do you think the replacement of `ifconfig` binary file by bash print `Use ip from the iproute` can be called deprecated functionality? Imagine this change and imagine tousands other scripts and apps that could fail because of this, including point that you are not able to change code in 3rd party apps. Does it means you deprecated the functionality, or you removed the functionality? Does it mean that you cannot use your system for ~months or you cannot update your system until all other apps/scripts are fixed (including waiting for vendors)? There is no difference in our project. Think please guys about this example. Currently, as I understand what you are trying to tell me, you think that this is correct deprecation and we could do it like that. I am telling, that if you want this, do not use the `deprecate` wording anywhere and call it *additional notification about removed functionality*. @bocekm , @ZhukovGreen

shaded-enmity:
Models are quintessential part of the Leapp API - as a customer who needs to customize the upgrade workflow to suit the needs of my organization I'm going to inadvertently end up using models provided by `leapp-repository`. Depending on the size of my organization, I may be done writing my actors now, but those actors may still be used by different part of my organization a year later. This is quite common pattern we see - develop customizations in a test environment and start upgrading in particular maintenance windows.

API's are contracts between the ones who create them and the ones who use them. If you give me a volatile API that changes on a whim, more than likely I'll start looking for a different, more stable API. And here comes the core idea of the deprecation process - we guarantee stability for at least 6 month period since first *notice*. This is important because it fundamentally changes any planning and projections I'm able to make.

Let's compare:
**Deprecation with eager removal**
1. I see a message that something that I've been using is deprecated
2. My code no longer works and my machines may start blowing up
3. The person who originally wrote these actors may have very well left the company
4. I need to learn Leapp & investigate the issue **now**, while the clock on the wall is ticking

**Deprecation with late removal**
1. I see a message that something that I've been using is deprecated
2. My code still works and I'm just being annoyed with deprecation message(s)
3. I contact the person who wrote those actors to "rebase" them on top of the new API
4. I can continue with upgrades as normal

Sometimes we may need to remove something without marking it as deprecated first, but there has to be a very serious technical reason to do that.

ZhukovGreen:
e2e tests

bocekm:
@shaded-enmity, thanks for the write up. We actually met on a mtg with Vinny, Petr and Artem earlier today to clarify what's needed. In the end, all of us agreed with what you've described as **Deprecation with late removal**.

Rezney:
@ZhukovGreen could you please rebase the PR? @pirat89 do I understand it correctly that you did reach an agreement here?

pirat89 (changes requested):
@Rezney I guess so. However, it looks that still some actors do not handle the original functionality (e.g. setuptargetrepos). So still waiting for the fix.

pirat89:
@ZhukovGreen I even see that at least some tests are changed to test the new functionality only. Cover by unit-tests the old and the new functionality as well.

ZhukovGreen:
Closed (consulted with @Rezney )

PR #546, opened 2020-07-30
Feature/keep manually enabled repos

ZhukovGreen:
ref #544 

This MR provides the user with the possibility to manually enable the rhel8 repository by specifying it next to --enablerepo option. By doing this, the repo will be shielded from exclusion (even if it is optional on rhel 7).

# TODO
- [ ] Merge #545 and rebase on top
- [ ] Clarify what is manually enabled repo


repos/system_upgrade/el7toel8/actors/excluderepositories/libraries/excluderepositories.py
```diff
@@ -0,0 +1,145 @@
...
+            if repo.repoid in opt_repo_mapping and not repo.enabled:
+                repos_to_exclude.append(opt_repo_mapping[repo.repoid])
+    return repos_to_exclude
+
+
+def process():
+    """Exclude the RHEL8 CRB repo if the RHEL7 optional repo is not enabled.
+
+    If manually enabled RHEL8 repo (by --enablerepo option) the repo do not
+    excluded.
```
bocekm, line 104: Please reword this sentence to be more understandable.

repos/system_upgrade/el7toel8/actors/excluderepositories/libraries/excluderepositories.py
```diff
@@ -0,0 +1,145 @@
...
+from leapp.libraries.common.config import get_product_type
+from leapp.libraries.stdlib import api
+from leapp.models import (
+    CustomTargetRepository,
+    RepositoriesExcluded,
+    RepositoriesFacts,
+    RepositoriesMap,
+)
+
+manually_enabled_repos_in_use = []
```
bocekm, line 11: This global var is quite confusing. What is it supposed to hold? All the repoids passed through the `--enablerepo` option?
It is hard for me to untangle it's purpose even if I find the place where this var is being filled. From what I understand, it should be either empty or holds **just** the CRB repo. But it's hard for me to read the code to verify that.
Instead of a global var, I'd rather see a function with a proper name that returns this list.

For example, it could be like this in `process()`:
```python
repos_to_exclude = _get_repos_to_exclude()
manually_enabled_repos = _get_manually_enabled_repos()
# The manually enabled repos override the default exclusion
repos_to_exclude = list(set(repos_to_exclude) - set(manually_enabled_repos))
overriden_repos_to_exclude = list(set(repos_to_exclude) & set(manually_enabled_repos)) 
_report_using_unsupported_repos(overriden_repos_to_exclude)
```

repos/system_upgrade/el7toel8/actors/excluderepositories/libraries/excluderepositories.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def _report_using_unsupported_repos(repo_ids):
+    if manually_enabled_repos_in_use:
```
bocekm, line 15:
```suggestion
    if repo_ids:
```
Since you call the function this way `_report_using_unsupported_repos(manually_enabled_repos_in_use)`, the `repo_ids` == `manually_enabled_repos_in_use`.

repos/system_upgrade/el7toel8/actors/excluderepositories/libraries/excluderepositories.py
```diff
@@ -0,0 +1,145 @@
...
+            ),
+            reporting.Remediation(
+                hint="At your own risk! Use leapp upgrade {}".format(
+                    " ".join(
+                        ["--enablerepo " + repo for repo in repos_to_exclude]
+                    )
+                )
+            ),
+        ]
+        reporting.create_report(report)
```
bocekm, line 145: The `process()` function is the highes level function/entry point of the library. Thus, from the clean code perspective, it should not be doing the low level stuff. Add this reporting functionality into a separate function like `report_excluded_repos()` and just call it from `process()`.

repos/system_upgrade/el7toel8/actors/excluderepositories/libraries/excluderepositories.py
```diff
@@ -0,0 +1,145 @@
...
+                repo, custom_target_repos
+            ):
+                opt_repo_mapping[repo.from_repoid] = repo.to_repoid
+    return opt_repo_mapping
+
+
+def _get_repos_to_exclude():
+    """Get a list of repoids to not use during the upgrade.
+
+    These are such RHEL 8 repositories that mapped to disabled RHEL7 Optional
```
bocekm, line 87:
```suggestion
    These are such RHEL 8 repositories that are mapped to disabled RHEL 7 Optional
```

repos/system_upgrade/el7toel8/actors/excluderepositories/libraries/excluderepositories.py
```diff
@@ -0,0 +1,145 @@
...
+                manually_enabled_repos_in_use.append(repo.to_repoid)
+                api.current_logger().info(
+                    "Manually enabled repository %s found. ", repo.to_repoid
+                )
+            return True
+    return False
+
+
+def _get_optional_repo_mapping():
+    """Get a dict of optional repositories based on RepositoriesMap: { 'from_repoid' : 'to_repoid'}.
```
bocekm, line 64:
```suggestion
    """Get a mapping of RHEL 7 Optional repos to corresponding RHEL 8 repos.
```

repos/system_upgrade/el7toel8/actors/excluderepositories/libraries/excluderepositories.py
```diff
@@ -0,0 +1,145 @@
...
+
+    It consumes RepositoriesMap messages and create map (dict) of optional repositories
+    on RHEL 7 system to CRB repositories on RHEL 8. See the RepositoriesMap model..
```
bocekm, line 67:
```suggestion
    It consumes RepositoriesMap messages and creates a mapping of 'Optional' repositories
    on RHEL 7 to CRB repositories on RHEL 8. It returns the mapping as a dict {'from_repoid' : 'to_repoid'}.
```

pirat89:
@ZhukovGreen @bocekm  The description seems confusing to me. The manually enabled repos are all custom repos. Including those from the special custom repo file. Thinking about that, I am not sure whether we should do difference between them, as from the upgrade POV, they are practically same. Both represented just by one msg type. Thinking about that, maybe we should postpone this PR after the release. WDYT?

ZhukovGreen:
@pirat89 /cc @bocekm I think manually enabled repo is very natural definition for repos which you're adding via both options:
1. Manually add them to the leapp repofile
2. Manually add them via --enablerepo flag
Why not? What is TODO here is to ensure I have this described and maybe updated some relevant docs. The MR is still under the development, so let's write all the things to resolve

ZhukovGreen:
closed in favor of #623

Issue #556, opened 2020-08-04
Make pytest.mark.parametrize more readable in this particular case


ZhukovGreen:
Current situation:

```python
   # valid cases without RHSM (== skip_rhsm)                                   
    (None, '1', testInData(_PACKAGES_MSGS, None, _XFS_MSG, _STORAGEINFO_MSG, None)),
    (None, '1', testInData(_PACKAGES_MSGS, None, None, _STORAGEINFO_MSG, None)),
    (None, '1', testInData([], None, _XFS_MSG, _STORAGEINFO_MSG, None)),        
    (None, '1', testInData([], None, None, _STORAGEINFO_MSG, None)),            
    (None, '1', testInData(_PACKAGES_MSGS, None, _XFS_MSG, _STORAGEINFO_MSG, _CTRF_MSGS)),
    (None, '1', testInData(_PACKAGES_MSGS, None, None, _STORAGEINFO_MSG, _CTRF_MSGS)),
    (None, '1', testInData([], None, _XFS_MSG, _STORAGEINFO_MSG, _CTRF_MSGS)),  
    (None, '1', testInData([], None, None, _STORAGEINFO_MSG, _CTRF_MSGS)),  
```

Example of more readable version:
```python
@pytest.mark.parametrize('raised,no_rhsm,testdata', [                           
    # valid cases with RHSM                                                     
    (None, '0', testInData(_PACKAGES_MSGS   , _RHSMINFO_MSG, _XFS_MSG,  _STORAGEINFO_MSG, None)),
    (None, '0', testInData(_PACKAGES_MSGS[0], _RHSMINFO_MSG, _XFS_MSG,  _STORAGEINFO_MSG, None)),
    (None, '0', testInData([],                _RHSMINFO_MSG, _XFS_MSG,  _STORAGEINFO_MSG, None)),
    (None, '0', testInData(_PACKAGES_MSGS,    _RHSMINFO_MSG, None,      _STORAGEINFO_MSG, None)),
    (None, '0', testInData(_PACKAGES_MSGS,    _RHSMINFO_MSG, _XFS_MSG,  _STORAGEINFO_MSG, _CTRF_MSGS)),
    (None, '0', testInData(_PACKAGES_MSGS[0], _RHSMINFO_MSG, _XFS_MSG,  _STORAGEINFO_MSG, _CTRF_MSGS)),
[...]

PR #557, opened 2020-08-05
targetuserspacecreator: improve readability of test cases

pirat89:
Test cases for the test_consume_data test are hard to read and it is
easy to do a mistake when updating these test cases. To improve
the readability of the code and make it more friendly for updates
in future, use the table-like formatting, ignoring linter errors.

Closes: #556


pirat89:
I cannot help myself, but really. I do not care it's not "pythonic" code style. It's much more readable than "pythonic way". I head to touch this part 3times in last days and this would be really much better to update than anything else.

PR #576, opened 2020-08-13
Minor fixes automatically fixed by pre-commit linter

ZhukovGreen:
Ref #575 
New pre-commit base linter automatic fixes


repos/common/.leapp/info
```diff
@@ -1 +1 @@
-{"messages": {}, "name": "common", "id": "efcf9016-f2d1-4609-9329-a298e6587b3c"}
\ No newline at end of file
+{"messages": {}, "name": "common", "id": "efcf9016-f2d1-4609-9329-a298e6587b3c"}
```
pirat89, line 1: @ZhukovGreen This should be fixed in leap first as this is the real output from leapp and it should not be different in our repositorz.

repos/system_upgrade/el7toel8/actors/detectgrubconfigerror/tests/files/grub.correct_trailing_space
```diff
@@ -1,3 +1,3 @@
 GRUB_TERMINAL_OUTPUT="console"
-GRUB_CMDLINE_LINUX="console=tty0 crashkernel=auto console=ttyS0,115200n8 no_timer_check net.ifnames=0" 
-GRUB_DISABLE_RECOVERY="true"
\ No newline at end of file
+GRUB_CMDLINE_LINUX="console=tty0 crashkernel=auto console=ttyS0,115200n8 no_timer_check net.ifnames=0"
```
pirat89, line 2: regarding the name of the file and the fact this is possibly a real example, it should not be modified by this PR. discuss it with @jmikovic and @Rezney  - If the file reflect the real word, it really should not be touched in any way.

repos/system_upgrade/el7toel8/actors/multipathconfread/tests/files/already_updated.conf
```diff
@@ -105,7 +105,7 @@ blacklist_exceptions {
 		product "S/390x"
 	}
 }
-		
+
```
pirat89, line 108: @ZhukovGreen thinking about that, I guess that the PR should affect only code. it should skip all data files completely. really we should not care about any whitespaces here etc. it's not code, it's just data.

utils/install_actor_deps.py
```diff
@@ -30,7 +30,7 @@ def install(path):
 
 
 def install_actor_deps(actor, directory):
-    for root, dirs, files in os.walk(directory):
+    for root, dirs, _ in os.walk(directory):
```
pirat89, line 33: the use of `_` in our project is kind of prohibited because of conflict with the `_` function. use the dummy_files instead for example.

ZhukovGreen: Could you please elaborate on this. I see the `_` variable inside /home/azhukov/Dropbox/code/leapp/leapp/utils/i18n.py, which is used only once in leapp/snactor/__init__.py
```
from leapp.utils.i18n import _  # noqa; pylint: disable=redefined-builtin
```
Also `_` is already used in 7 places inside this repo and it is a typical python convention to skip the variable when unpacking

repos/common/.leapp/info
ZhukovGreen: Added this to ignore

repos/system_upgrade/el7toel8/actors/detectgrubconfigerror/tests/files/grub.correct_trailing_space
ZhukovGreen: Added this to ignore, the file suffix speaks on its own. Sorry for missing this

repos/system_upgrade/el7toel8/actors/multipathconfread/tests/files/already_updated.conf
ZhukovGreen: @pirat89 it is good to have pre-commit working on all files. There are a lot of useful fixes that might be done for example in yml files or shell scripts. Let's just exclude from linting those files that are outputs of other programs.

ZhukovGreen (commented):
Adding my answers to the code review from Petr

ZhukovGreen:
@pirat89 I am not yet sure if I properly setup excluded filter, please review it. I think some of the datafiles are still ok to modify, but let me know if I am wrong.

utils/install_actor_deps.py
pirat89: use of dummy prefix/suffix is kind of typical as well, right because of this. It is possible in future that we will need to deal with it, it's better to not use it. Originally it was mentioned in our guidelines as well. Not sure whether it is there as well, anyway, do not use it. As well, it could happen in future that someone will need to start to use such dummy variable. So that's good reason why to put there something more clear about what's inside.

Existence of `_` in our project doesn't mean we want to start to use it everywhere. Maybe it's just a mistake that reviewers did and passed these variables, but they shouldn't.

repos/system_upgrade/el7toel8/actors/multipathconfread/tests/files/already_updated.conf
pirat89: conf file is not code. from that point, I do not agree we should use linting here. yml files - ok, makes sense, as many people do mistake in yml files. shell scripts are code again, so this doesn't have to be ignored, but again, we should check the filename has expected suffix or shebang. Checking all files, will be from my point of view rather annoying than helpful. can be discussed in team (@oamg/developers)

utils/install_actor_deps.py
pirat89: actually, this is one of things, that pre-commit hooks could detect ;-) but again, people should be guided how to ignore it for bundled libraries.

repos/system_upgrade/el7toel8/actors/multipathconfread/tests/files/already_updated.conf
pirat89: thinking about that, I would skip probably everything inside the "files" directories. these files should not be restricted by us at all.

repos/system_upgrade/el7toel8/actors/detectgrubconfigerror/tests/files/grub.correct_trailing_space
vinzenz: This is based on a real world problem, I remember that this has been an issue in the past which had to be solved.

ZhukovGreen:
@pirat89 current set of excluded files are here:
https://github.com/oamg/leapp-repository/pull/575/files#diff-82c3b36709926de3a1b27bea0d8258ccR2

vojtechsokol:
leapp-ci build

PR #579, opened 2020-08-18
Require new leapp framework capability 1.3

bocekm:
Based on the bump of the framework capability in https://github.com/oamg/leapp/pull/640


pirat89:
Yep, that should be part of #531 . We missed that during the review. Merging.

pirat89 (approved):

PR #581, opened 2020-08-24
LocalReposInhibit: Check mirrorlist and metalink too

### LINKED ISSUE #580: Repo mirrorlist breaks the upgrade
Please bear in mind that there can be also a mirrorlist instead of baseurl. Not sure if the original issue applies also to mirrorlists but it breaks the upgrade when they are used (e.g. in AWS).

```
        target_repo_id_to_url_map = {
            repo.repoid: repo.baseurl
            for repofile in target_repos
            for repo in repofile.data
            if repo.baseurl
        }
```

Above we gather only baseurl repos in the map but then we try to access it by repoid from target repos (with mirrorlist) which fails:

```
  File "/usr/lib/python2.7/site-packages/leapp/utils/deprecation.py", line 32, in process_wrapper
    return target_item(*args, **kwargs)
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/localreposinhibit/actor.py", line 44, in process
    if self.file_baseurl_in_use():
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/localreposinhibit/actor.py", line 40, in file_baseurl_in_use
    for repo in used_target_repos
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/localreposinhibit/actor.py", line 40, in <genexpr>
    for repo in used_target_repos
KeyError: u'rhui-client-config-server-8'
```

@AloisMahdal let's try to add also mirrorlists to the tests.

_Originally posted by @Rezney in https://github.com/oamg/leapp-repository/pull/531#issuecomment-677880614_

drehak:
A repository can have an URL defined in the field `mirrorlist` or `metalink` instead of `baseurl`. Mirrorlist and metalink are mutually exclusive, though baseurl can still be present as a fallback.

Important notes:
- I'm not sure if the bug with addresses starting with `file:///` occurs when it's a metalink/mirrorlist address. Should we regard those as local repos too?
- `baseurl`, `mirrorlist` and `metalink` are all nullable fields but I could find no checks if there's not a forbidden combination of them, e.g. mirrorlist with metalink or none of these fields at all. I know such repos would be invalid, but there are many places in our codebase where they would traceback. Maybe we should check that somewhere.

Closes #580

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/localreposinhibit/actor.py
@@ -27,10 +27,9 @@ class LocalReposInhibit(Actor):
         used_target_repos = next(self.consume(UsedTargetRepositories)).repos
         target_repos = next(self.consume(TMPTargetRepositoriesFacts)).repositories
         target_repo_id_to_url_map = {
-            repo.repoid: repo.baseurl
+            repo.repoid: repo.mirrorlist or repo.metalink or repo.baseurl or ""
             for repofile in target_repos
             for repo in repofile.data
-            if repo.baseurl
         }
         return any(
             target_repo_id_to_url_map[repo.repoid].startswith("file:")
--- a/repos/system_upgrade/el7toel8/actors/localreposinhibit/tests/test_unit_localreposinhibit.py
+++ b/repos/system_upgrade/el7toel8/actors/localreposinhibit/tests/test_unit_localreposinhibit.py
@@ -11,10 +11,21 @@ from leapp.snactor.fixture import ActorContext
 
 
 @pytest.mark.parametrize(
-    ("baseurl", "exp_msgs_len"),
-    [("file:///root/crb", 1), ("http://localhost/crb", 0)],
+    ("baseurl", "mirrorlist", "metalink", "exp_msgs_len"),
+    [
+        ("file:///root/crb", None, None, 1),
+        ("http://localhost/crb", None, None, 0),
+        (None, "file:///root/crb", None, 1),
+        (None, "http://localhost/crb", None, 0),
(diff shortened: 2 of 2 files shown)


ZhukovGreen (changes requested):
Generally looks good to me. But please fix CI first

drehak:
@ZhukovGreen The CI problems that popped up are existing code not passing trough the new checks in pylint 2.6, released 5 days ago. The linked PR addresses this.

examon:
leapp-ci build

ZhukovGreen:
@pirat89 petr, could you pls glance on it. I have some concerns regarding the description in this MR, maybe I miss something. Otherwise looks good to me

pirat89:
> - I'm not sure if the bug with addresses starting with file:/// occurs when it's a metalink/mirrorlist address. Should we regard those as local repos too?

Yes, it will occur for the others as well.

> - baseurl, mirrorlist and metalink are all nullable fields but I could find no checks if there's not a forbidden combination of them, e.g. mirrorlist with metalink or none of these fields at all. I know such repos would be invalid, but there are many places in our codebase where they would traceback. Maybe we should check that somewhere.

No, we are not checking it. We are expecting that repositories are wel defined regarding DNF/YUM. We cannot check every piece of system whether is already broken by user or not. As this is one of very basic things (to have properly set repofiles), we do not check that. Difference is in question of duplicated repoids - which we are checking - but it's different story.

ZhukovGreen (approved):
After @pirat89 answer, I feel safe to approve it. But I believe it is required additional verification task @mmacura311 , right?

drehak:
@ZhukovGreen @mmacura311 Thanks a lot, guys!

PR #582, opened 2020-08-25
pylint: Disable checks introduced in pylint 2.6

drehak:
pylint 2.6 was released on 2020-08-20 and introduced two new checks which we cannot use due to Python 2:
  - `super-with-arguments` (R1725)
  - `raise-missing-from` (W0707)

Release notes: http://pylint.pycqa.org/en/latest/whatsnew/2.6.html

Blocks #581.


ZhukovGreen (approved):
Good

@Rezney FYI

PR #584, opened 2020-09-02
Move GRUB device detection into library

Rezney:
We have to start detecting GRUB device at the same stage
as we update GRUB core in order to not hit issue with device
name swap.

Related BZ:

https://bugzilla.redhat.com/show_bug.cgi?id=1871162

Other info:

https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkremovedenvvars/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkremovedenvvars/actor.py
@@ -1,0 +1,19 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import checkremovedenvvars
+from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckRemovedEnvVars(Actor):
+    """
+    Check for usage of removed environment variables and inhibit the upgrade
+    if they are used.
+    """
+
+    name = 'check_removed_envvars'
+    consumes = ()
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        checkremovedenvvars.process()
--- a/repos/system_upgrade/el7toel8/libraries/tests/test_grub.py
+++ b/repos/system_upgrade/el7toel8/libraries/tests/test_grub.py
@@ -1,0 +1,104 @@
+import os
+
+import pytest
+
+from leapp.exceptions import StopActorExecution
+from leapp.libraries.common import grub
+from leapp.libraries.stdlib import CalledProcessError, api
+from leapp.libraries.common.testutils import logger_mocked
+
+BOOT_PARTITION = '/dev/vda1'
+BOOT_DEVICE = '/dev/vda'
+
+VALID_DD = b'GRUB GeomHard DiskRead Error'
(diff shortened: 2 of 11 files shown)


repos/system_upgrade/el7toel8/models/grubdevice.py
```diff
@@ -1,13 +1,22 @@
 from leapp.models import Model, fields
 from leapp.topics import SystemFactsTopic
+from leapp.utils.deprecation import deprecated
 
 
+@deprecated(
+    since="2020-09-01",
+    message=("The model is deprecated as we moved the GRUB device detection into library")
```
ZhukovGreen, line 8: Maybe extend the message with the advice what the developer should do

ZhukovGreen (commented):
Just a little comment

repos/system_upgrade/el7toel8/models/grubdevice.py
Rezney: to use the library :) But I will expand the message for the completeness sake...

repos/system_upgrade/el7toel8/actors/checkremovedenvvars/actor.py
```diff
@@ -0,0 +1,19 @@
...
+class CheckRemovedEnvVars(Actor):
+    """
+    Check for usage of removed environment variables and inhibit the upgrade
+    if they are used.
+    """
+
+    name = 'check_removed_envvars'
+    consumes = ()
+    produces = (Report,)
+    tags = (RPMUpgradePhaseTag, IPUWorkflowTag)
```
pirat89, line 16:
```suggestion
    tags = (ChecksPhaseTag, IPUWorkflowTag)
```

repos/system_upgrade/el7toel8/actors/checkremovedenvvars/actor.py
```diff
@@ -0,0 +1,19 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import checkremovedenvvars
+from leapp.reporting import Report
+from leapp.tags import RPMUpgradePhaseTag, IPUWorkflowTag
```
pirat89, line 4:
```suggestion
from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
```

pirat89 (changes requested):

repos/system_upgrade/el7toel8/libraries/tests/test_grub.py
```diff
@@ -26,6 +23,21 @@ def raise_call_error(args=None):
...
+        self.warnmsg = None
+
+    def info(self, msg):
+        self.infomsg = msg
+
+    def warn(self, msg):
+        self.warnmsg = msg
+
+    def __call__(self):
+        return self
```
pirat89, line 38: uset testutils library

pirat89 (approved):
lgtm. just code review in this case - without own testing.

Rezney:
e2e tests

Rezney:
e2e tests

Rezney:
Verified as part of OAMG-3543. Gen1 instances contain two storage devices which often swaps its name upon reboot. We are now able to detect the correct GRUB device with moving its detection right before the actual GRUB core update.

PR #588, opened 2020-09-15
Remove invalid XFS mount options from target system fstab

drehak:
Several XFS mount options have been removed from the kernel between RHEL 7 and 8. On RHEL 7, the options are deprecated and have no effect. However, on RHEL 8, their occurence will lead to an unbootable machine.
If this actor detects any of the options in `/etc/fstab`, it removes them during the upgrade. The user is notified about this in the preupgrade report and a "read-only" copy of the new `/etc/fstab` is provided for inspection.

The list of removed options is taken from `xfs(5)` on RHEL 8.

Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1870813


repos/system_upgrade/el7toel8/actors/checkfstaboptions/actor.py
```diff
@@ -0,0 +1,74 @@
...
+        found = defaultdict(list)
+        for entry in storage_info.fstab:
+            if entry.fs_vfstype == 'xfs':
+                for option in entry.fs_mntops.split(','):
+                    option_name = option.split('=')[0]
+                    if option_name in self.REMOVED_XFS_OPTIONS:
+                        found[entry.fs_spec].append(option_name)
+
+        if found:
+            reporting.create_report([
```
Rezney, line 55: I am actually missing the inhibitor flag :)

repos/system_upgrade/el7toel8/actors/checkfstaboptions/tests/test_checkfstaboptions.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def test_removed_multiple(current_actor_context):
+    fstab = MOUNTS_CORRECT_1 + MOUNTS_INVALID + MOUNTS_CORRECT_2
+    current_actor_context.feed(StorageInfo(fstab=fstab))
+    current_actor_context.run()
+    assert current_actor_context.consume(Report)
+    for string in ['/dev/mapper/rhel_ibm--p8--kvm--03--guest--03-root', 'nobarrier',
+                   'UUID=c3ae2d1a-8088-45c2-9482-4aacb2c075cf', 'barrier', 'delaylog', 'ihashsize']:
+        assert string in current_actor_context.consume(Report)[0].report['summary']
```
Rezney, line 87: Same for the tests. We may want to check if the inhibitor flag was raised.

Rezney (changes requested):

repos/system_upgrade/el7toel8/actors/checkfstaboptions/actor.py
drehak: Ah, my bad.

repos/system_upgrade/el7toel8/actors/checkfstaboptions/tests/test_checkfstaboptions.py
drehak: Alright.
Added severity and remediation too, somehow I missed those.

repos/system_upgrade/el7toel8/actors/checkfstaboptions/actor.py
```diff
@@ -0,0 +1,76 @@
...
+                reporting.Title(
+                    'Removed mount options found in /etc/fstab'
```
pirat89, line 57: @drehak the title is kind of encrypted. You are referring just to XFS, but title says it's generic.

pirat89: I am curious whether additional FSs have some changes as well.

drehak: Yeah, I deliberately decided to keep the actor more generic because other FS types might have removed mount options too - now or later. At this moment it only covers XFS.

Rezney:
e2e tests

examon:
e2e tests

examon:
leapp-ci build

AloisMahdal:
/me is checking out the murphy-ci/integration fail..

AloisMahdal:
...aaand it's a test error (introduced by me yesterday -- sorry) and you can safely ignore it.

drehak:
@sandeen Hi Eric,

as you are the maintainer of `xfsprogs`, your review could be of help here. According to `xfs(5)`, several options have been removed from the kernel between RHEL 7 and RHEL 8. This actor inhibits the upgrade from RHEL 7 to 8 if one of the removed options (for example **nobarrier**) occurs in `/etc/fstab` . I have a few minor questions:

* Is the list of removed options taken from the manual page correct? Are the kernel versions correct? 
* Do the options apply to XFS mounts only? If no, have they been removed for other types of mounts as well?

sandeen:
> @sandeen Hi Eric,

>     * Is the list of removed options taken from the manual page correct? Are the kernel versions correct?

Yes, though the versions refer to upstream versions, not RHEL.  barrier/nobarrier was removed prior to RHEL8 GA.
The others were removed in v4.0 so they are also missing from RHEL8.

>     * Do the options apply to XFS mounts only? If no, have they been removed for other types of mounts as well?

the xfs(5) man page only refers to xfs options.  Other filesystems may have similarly named mount options, but they would not be affected by the xfs deprecation schedule.

> This actor inhibits the upgrade from RHEL 7 to 8 if one of the removed options (for example nobarrier) occurs in /etc/fstab .

It would be unfortunate to prohibit an upgrade due to this, because for i.e. barriers, the handling is simply automatic now, and the mount options have been ignored since 2016.  Are these sorts of things documented, possibly with workarounds?  i.e. it would be 100% safe to simply remove any barrier/nobarrier options from xfs fstab entries, and proceed with the upgrade.

drehak:
@sandeen 
> It would be unfortunate to prohibit an upgrade due to this, because for i.e. barriers, the handling is simply automatic now, and the mount options have been ignored since 2016.

We still need to take care of this one way or another since an upgraded machine with these options can't boot to RHEL 8:
https://bugzilla.redhat.com/show_bug.cgi?id=1870813
Maybe we could handle the removal of options automatically, although that's riskier.

> Are these sorts of things documented, possibly with workarounds?

Yes. When the (pre)upgrade is stopped, a report is generated which contains an entry about the problem and info about how to fix the problem. Relevant code in this PR:
https://github.com/oamg/leapp-repository/pull/588/files#diff-71536c617bd27e6e0f0289ad25950efdR74
We don't have a central registry of these potential problems yet, although one is planned before next release.

sandeen:
It really is OK to simply remove nobarrier/barrier and nodelaylog/delaylog - the former is automatically set properly now, and the latter was really only for debug/testing, there is no "nodelaylog" anymore, it was only used for testing.

If it were me, I'd remove the others as well and perhaps just note it - if I can help with documentation on any of this just let me know.

repos/system_upgrade/el7toel8/actors/checkfstaboptions/actor.py
fernflower: The report title is too cryptic for a bystander, I'd rename to "Unsupported mount options in /etc/fstab detected"

repos/system_upgrade/el7toel8/actors/checkfstaboptions/actor.py
```diff
@@ -0,0 +1,76 @@
...
+
+    The list of removed options is taken from the xfs(5) manual page on RHEL 8.
+    """
+
+    name = 'check_fstab_options'
+    consumes = (StorageInfo,)
+    produces = (Report,)
+    tags = (IPUWorkflowTag, ChecksPhaseTag)
+
+    REMOVED_XFS_OPTIONS = [
```
ZhukovGreen, line 26: why do you need this constant inside the class variable?

ZhukovGreen: Also, if you would decide to move it in the module scope check this https://stackoverflow.com/questions/33700927/tuples-vs-lists-for-module-level-constants-in-python disucssion. It is nice to understand when you might need tuple ot list as a constant

repos/system_upgrade/el7toel8/actors/checkfstaboptions/actor.py
```diff
@@ -0,0 +1,76 @@
...
+                'Could not check fstab entries', details={'details': 'No StorageInfo facts found.'}
+            )
+
+        found = defaultdict(list)
+        for entry in storage_info.fstab:
+            if entry.fs_vfstype == 'xfs':
+                for option in entry.fs_mntops.split(','):
+                    option_name = option.split('=')[0]
+                    if option_name in self.REMOVED_XFS_OPTIONS:
+                        found[entry.fs_spec].append(option_name)
```
ZhukovGreen, line 52: nothing bad here, just FYI you could simplify the construction from:

for i in iterable:
if condition in terable:
do something

into

for i in filter(or comprhension)(condition, iterable)

in this case you will have a smaller number of lines and more efficient code (looping through generator)

repos/system_upgrade/el7toel8/actors/checkfstaboptions/tests/test_checkfstaboptions.py
```diff
@@ -0,0 +1,89 @@
+from leapp.models import FstabEntry, StorageInfo
+from leapp.snactor.fixture import current_actor_context
+from leapp.reporting import Report
+
+
+MOUNTS_CORRECT_1 = [
```
ZhukovGreen, line 6: watch the previous comment about constants

ZhukovGreen: maybe to add that tuples are usually more memory efficient and faster (https://stackoverflow.com/questions/1708510/list-vs-tuple-when-to-use-each)

repos/system_upgrade/el7toel8/actors/checkfstaboptions/tests/test_checkfstaboptions.py
```diff
@@ -0,0 +1,89 @@
...
+        fs_file='/boot',
+        fs_vfstype='ext4',
+        fs_mntops='defaults',
+        fs_freq='1',
+        fs_passno='2'
+    )
+]
+
+
+def test_ok(current_actor_context):
```
ZhukovGreen, line 64: it is might be a good idea to add docstrings (or comments) to your tests in order to explain what do you actually testing

ZhukovGreen (commented):
This was a surface review, so leaving just comment here. The code seems readable enough and might be merged. Some minor things which I commented, might be fixed or not. It is up to you @drehak .

repos/system_upgrade/el7toel8/actors/checkfstaboptions/actor.py
fernflower: Using filter is against our guidelines unless it can't be avoided.

Readable code is preferred over preliminary optimizations

ZhukovGreen: @fernflower I am sorry Inna, but I can't see this in our guidelines

fernflower: @ZhukovGreen https://github.com/oamg/leapp-guidelines/blob/master/python-coding-guidelines.md#7-avoid-map-reduce-and-too-complex-comprehensions

ZhukovGreen: @fernflower I don't see in my comment any mention of filter or too complex comprehension. Anyway, this is not a strict demand as well as any other from my review

repos/system_upgrade/el7toel8/actors/checkfstaboptions/tests/test_checkfstaboptions.py
drehak: Thanks, it's a tuple now.

drehak:
The code has been rewritten to remove the options instead of just reporting about them. At the moment, installing the new `/etc/fstab` to our RHEL 8 initramdisk doesn't work yet.

drehak:
The modified `/etc/fstab` now gets into the initramdisk. The only thing left to implement is to get it on the target system too.

(During my upgrade testing with `nobarrier` set on the root mount, things went smoothly right until the second reboot, where it got stuck in a boot loop. Booting into the old el7 kernel and back somehow let the boot process continue, although root got mounted as read only. Indeed, the old `/etc/fstab` was still there. This is unfeasible, of course.)

drehak:
Finally done! (Sorry for the delays, it was hard to get the details right.)
The focus of this PR has shifted a lot so now the forbidden mount options are actually removed from /etc/fstab during the upgrade. To be specific, if any of the options are detected:
- during the report phase, a report entry is created. The new `/etc/fstab` is also written to `/var/log/leapp/fstab.new` with a brief comment. Any changes to this file are not reflected during the upgrade process. The file is then cleaned up in later phases.
- the new `/etc/fstab` is added into the upgrade initramdisk, so that no (re)mounts will fail.
- from inside the initramdisk, `/etc/fstab` is overwritten on the system, ensuring a smooth RHEL 8 boot.

I've already done several successful upgrades with this PR (with/without the options) so I'm certain it works. :)

@Rezney @pirat89 @ZhukovGreen @fernflower a review would be appreciated!

cc @sandeen

repos/system_upgrade/el7toel8/actors/checkfstaboptions/tests/test_checkfstaboptions.py
drehak: Basically fixed this one, in the current tests it's more obvious.

repos/system_upgrade/el7toel8/actors/checkfstaboptions/actor.py
drehak: Resolving (the code is no longer there).

repos/system_upgrade/el7toel8/actors/checkfstab/actor.py
```diff
@@ -0,0 +1,81 @@
...
+from leapp.libraries.stdlib import api
+from leapp.models import Report, FstabSignal
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+from leapp import reporting
+
+
+PREFACE = """# Below are the modified contents of /etc/fstab that have been generated for
+# the upgraded system. Changes to this file will not be reflected. If you
+# believe the file has been modified incorrectly, please report a bug and
+# attach this file and the current /etc/ftsab.
```
pirat89, line 16:
```suggestion
# attach this file and the current /etc/fstab.
```

pirat89, line 16: The information that user should check the file and in case of invalid content the bu should be reported is only here. This should be already part of the report as well - how a user realize they are supposed to check the file?

As well, in case there is a bug in the generated / modified fstab, we should encourage users to update the fstab so the removed options are not present and run leapp again - so we will not do the modifications.

repos/system_upgrade/el7toel8/actors/checkfstab/actor.py
```diff
@@ -0,0 +1,81 @@
...
+            reporting.Title(
+                '/etc/fstab will be modified during upgrade'
+            ),
+            reporting.Summary(
+                'In RHEL 8, several mount options for XFS are removed and their usage in /etc/fstab would '
+                'result in an unbootable system. Since the options are already deprecated and do not have '
+                'any effect on the mounts, leapp will remove them.\n'
+                'A copy of the new /etc/fstab has been written to {}. '.format(FSTAB_LOGFILE) +
+                'Changes to this file will not be reflected in the upgrade process.'
+            ),
```
pirat89, line 70: I am missing the instruction (maybe remediation hint?) people should take a look into the file  to check it's correct, or they can update the file manually to drop the deprecated/removed options already and re-run leapp.

repos/system_upgrade/el7toel8/models/fstabsignal.py
```diff
@@ -0,0 +1,6 @@
+from leapp.models import Model
+from leapp.topics import BootPrepTopic
+
+
+class FstabSignal(Model):
```
pirat89, line 5: this is nothing telling model. whoever will see it, it will miss the purpose of the model. Missing docstring, model name is not intuitive.

repos/system_upgrade/el7toel8/actors/checkfstab/actor.py
```diff
@@ -0,0 +1,81 @@
...
+
+    name = 'check_fstab'
+    consumes = ()
+    produces = (Report, FstabSignal)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        fstab_old = []
+        try:
+            with open('/etc/fstab') as f:
```
pirat89, line 40: you cannot work with files in the check phase. expecte logic is - get all required data about the fstab during the scanner phase. work with the data during the check phase.

repos/system_upgrade/el7toel8/actors/removeinvalidfstaboptions/actor.py
```diff
@@ -0,0 +1,29 @@
...
+        if not next(self.consume(FstabSignal), None):
+            return
+
+        with open('/etc/fstab', 'r') as fstab:
+            lines = fstab.readlines()
+        with open('/etc/fstab', 'w') as fstab:
+            fstab.writelines(drop_xfs_options(lines))
+        # clean up
+        if os.path.exists('/var/log/etc/fstab.new'):
+            os.remove('/var/log/etc/fstab.new')
```
pirat89, line 29:
```suggestion
```

The cleanup doesn't make sense. Why to remove log file?... keep it as it is

repos/system_upgrade/el7toel8/actors/initramdiskgenerator/libraries/initramgen.py
```diff
@@ -65,16 +66,25 @@ def _install_files(context):
...
         context.copy_to('/etc/dasd.conf', '/etc/dasd.conf')
-        return ['/etc/dasd.conf']
-    return []
+        files.append('/etc/dasd.conf')
+
+    if next(api.consume(FstabSignal), None):
+        with open('/etc/fstab') as fstab7:
+            with context.open('/etc/fstab', 'w') as fstab8:
+                fstab8.writelines(drop_xfs_options(fstab7.readlines()))
+        files.append('/etc/fstab')
```
pirat89, line 85: what if user / vendor .. will need to change the fstab file as well? thinking abou that, this is not a good solution from my pov. and what is the content of the /etc/fstab inside the container by default? is it same as the original one already? if bydefault it is, the whole operation should happen just above this one file. as well, I guess separate function could be better - this file is always present inside the initrd and only thing here is that we want the upgradet versions inside the container. any else ideas welcomed @oamg/developers

pirat89 (changes requested):

repos/system_upgrade/el7toel8/actors/removeinvalidfstaboptions/actor.py
Rezney: + "leapp" is missing in the path...

pirat89: @Rezney good catch!

pirat89: in the same time...it was supposed to be `/var/log/leapp/fstab.new` if I remember the code in the PR well..

Rezney (commented):
Looks good to me. Just some minor suggestions / question. Feel free to ignore the "content/contents" comments, it is most probably just matter of personal taste and would require too many changes.

Tested and it seems to be working fine.

@pirat89 were all your concerns addressed already?

drehak:
@Rezney Is this still wanted?

pirat89:
@drehak that's a question. I mean, we have another task, where we would possibly need to update the fstab, so if that will be happening, I guess we will revive this as well. But I expec the resulting solution will work with the fstab itself, instead of creating the `FstabContent` in the end. Let's keep it for a while yet until we begin work on additional 2 issues connected with fstab currently.

fernflower:
There has been no activity for quite some time now, and no response to questions. In the attempt to keep an up-to-date PRs overview I am closing this PR, please reopen if this change is still needed.

PR #589, opened 2020-09-17
Fix run of unit tests in case on system with installed pytest

pirat89:
In case the host system has installed e.g. leapp or pytest already,
the host system libraries are used. In our case, different pytest,
different leapp... than we expact to use for running of unit tests.

To resolve this, do not call pytest dierectly, but load the pytest
library through python.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -240,7 +240,7 @@ test_no_lint:
 	cd repos/system_upgrade/el7toel8/; \
 	snactor workflow sanity-check ipu && \
 	cd - && \
-	pytest $(REPORT_ARG) $(ACTOR_PATH) $(LIBRARY_PATH)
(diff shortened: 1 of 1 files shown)


ZhukovGreen (approved):
Good catch, thx for this. However such issues could occur again for any globally installed package, which interfere the requirements txt. We'll fix this completely only when switching to poetry

pirat89:
@ZhukovGreen in such case, poetry is becoming more attractive to me :)

pirat89:
@ZhukovGreen Thanks for this fix to you man :)

ZhukovGreen:
@pirat89 Remember the automatic generation of py3 and py2 environments as well:)))

PR #590, opened 2020-09-23
Do not use the deprecated logger.warn method

pirat89:
The logger.warn method has been deprecated since Python 3.3.
Developers are supposed to use the logger.warning method instead.
See:
    https://github.com/python/cpython/blob/3.3/Lib/logging/__init__.py#L1252-L1253
    https://bugs.python.org/issue13235

Deprecates:
    produce_mocked.warn from leapp.libraries.common.testutils


pirat89:
Let's call it cute:
```
AttributeError: module 'warnings' has no attribute 'warning'
```

For `logging` use `warning`, for `warnings` use `warn`. That's design.

pirat89:
@Rezney  ok, it seems it could work :)

pirat89:
e2e tests

examon:
leapp-ci build

Rezney (approved):
Looks good to me. Let's wait for the last part of the pipeline to pass.

pirat89:
leapp-ci build

vojtechsokol:
e2e tests

PR #591, opened 2020-09-24
Azure hybrid image handling

Rezney:
Detect Azure hybrid (BIOS/EFI) image and convert
/boot/grub2/grubenv symlink pointing to
/boot/efi/EFI/redhat/grubenv to a normal file
as GRUB cannot read the symlink if it is pointing
to different partition (in case of BIOS).

Consequently, we do not need the "KernelCmdlineArg"
with "root_delay" anymore.


pirat89:
leapp-ci build

repos/system_upgrade/el7toel8/actors/cloud/grubenvtofile/libraries/grubenvtofile.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def grubenv_to_file():
+    try:
+        run(['unlink', BIOS_PATH])
+    except CalledProcessError as err:
+        api.current_logger().warning('Could not unlink {}: {}'.format(BIOS_PATH, str(err)))
+        return
+    try:
+        run(['cp', EFI_PATH, BIOS_PATH])
```
pirat89, line 15: Maybe to be sure, this could be better:
```suggestion
        run(['cp', '-a', EFI_PATH, BIOS_PATH])
```

repos/system_upgrade/el7toel8/actors/cloud/grubenvtofile/actor.py
```diff
@@ -0,0 +1,26 @@
...
+
+    name = 'grubenvtofile'
+    consumes = (HybridImageDetected,)
+    produces = ()
+    tags = (FinalizationPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        grubenv_msg = next(self.consume(HybridImageDetected), None)
+
+        if grubenv_msg:
```
pirat89, line 25: I guess this is not correct. If you create (produce) the default `HybridImageDetected`, this will be always false. Extend please the model (see suggestion around the mode).

repos/system_upgrade/el7toel8/models/grubenv.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class HybridImageDetected(Model):
+    """
+    Model used for instructing Leapp to convert "grubenv" symlink
+    into a regular file in case of hybrid (BIOS/EFI) images using BIOS
+    on Azure.
+
+    This model will likely be removed in the future.
+    """
```
pirat89, line 12: Would't make sense to mark the mode to be deprecated in case it's expected it will be removed in future (regarding the docstring).

repos/system_upgrade/el7toel8/models/grubenv.py
```diff
@@ -0,0 +1,13 @@
...
+    topic = SystemFactsTopic
```
pirat89, line 13: Really do not produce *empty* msg. Instead:
```suggestion
    topic = SystemFactsTopic
    
    detected = fields.Boolean(default=False)
```

In such case, maybe even the name of the class could be changed, like `HybridImageDetected`. The point is, that when you detect it, you should create non-default msg. Maybe it's *inherited* from the `TransactionCompleted` model, but that one has different purpose than this new one.

pirat89 (changes requested):
Mostly just suggestions and tips. But I guess the stuff around the `HybridImageDetected` model should really be changed.

repos/system_upgrade/el7toel8/actors/cloud/grubenvtofile/actor.py
Rezney: @pirat89 could you please explain why it would be false? It is obviously working for me :) Why to complicate it with some field? We either send it or not. The message name is self-explanatory.

repos/system_upgrade/el7toel8/actors/cloud/grubenvtofile/libraries/grubenvtofile.py
Rezney: I agree. Originally wanted to do but as the metadata were copied ok...

repos/system_upgrade/el7toel8/actors/cloud/checkrhui/actor.py
```diff
@@ -66,10 +66,6 @@ def process(self):
                     # We have to disable Amazon-id plugin in the initramdisk phase as the network
                     # is down at the time
                     self.produce(DNFPluginTask(name='amazon-id', disable_in=['upgrade']))
-                if provider == 'azure':
-                    # This option is important on Azure. Due to the special Azure hybrid image
-                    # GRUB cannot find grubenv and is getting just default options.
-                    self.produce(KernelCmdlineArg(**{'key': 'rootdelay', 'value': '300'}))
```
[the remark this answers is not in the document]
Rezney: It was actually never about sending `root_delay` in `KernelCmdlineArg` but about sending *any* such message. Whenever we run `grubby --add-args` $kernelopts in GRUB config file gets expanded and we did not need `grubenv` (with `root_delay` as it was always there) that much in this case. So when this was merged we were under impression we missed just the important `root_delay` argument and GRUB used just some default to boot it but it was actually the dark magic behind the curtains which made it work.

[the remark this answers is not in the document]
Rezney: But we can still sync in person...

repos/system_upgrade/el7toel8/actors/cloud/checkhybridimage/actor.py
```diff
@@ -0,0 +1,19 @@
+from leapp.actors import Actor
+from leapp.models import InstalledRPM, HybridImageDetected, FirmwareFacts
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+from leapp.libraries.actor.checkhybridimage import check_hybrid_image
+
+
+class CheckHybridImage(Actor):
+    """
+    Check if the system is running in Azure cloud.
+    This check is implemented because
```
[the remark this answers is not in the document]
Rezney: Yep,I did :)

repos/system_upgrade/el7toel8/actors/cloud/checkhybridimage/libraries/checkhybridimage.py
```diff
@@ -0,0 +1,58 @@
...
+BIOS_PATH = '/boot/grub2/grubenv'
+EFI_REL_PATH = '../efi/EFI/redhat/grubenv'
+
+
+def is_grubenv_symlink_to_efi():
+    """
+    Check whether '/boot/grub2/grubenv' is a relative symlink to
+    '/boot/efi/EFI/redhat/grubenv'.
+    """
+    return os.path.islink(BIOS_PATH) and os.readlink(BIOS_PATH) == EFI_REL_PATH
```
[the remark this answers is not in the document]
Rezney: Yes, in case of the official Azure image it will always be like that. There are kickstarts for this. IMHO we can't predict what users will do. I think they can do literally everything :)

repos/system_upgrade/el7toel8/actors/cloud/checkrhui/actor.py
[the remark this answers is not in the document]
pirat89: Explained. Thanks :)

repos/system_upgrade/el7toel8/models/grubenv.py
Rezney: Realized we will have to keep it anyway...

repos/system_upgrade/el7toel8/actors/cloud/checkhybridimage/libraries/checkhybridimage.py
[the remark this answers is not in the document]
Rezney: Implemented as discussed offline.

repos/system_upgrade/el7toel8/actors/cloud/grubenvtofile/actor.py
```diff
@@ -0,0 +1,26 @@
...
+
+    name = 'grubenvtofile'
+    consumes = (HybridImage,)
+    produces = ()
+    tags = (FinalizationPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        grubenv_msg = next(self.consume(HybridImage), None)
+
+        if grubenv_msg and grubenv_msg.detected:
```
pirat89, line 25: What about this (↓)? Anyway, not blocker. I am ok with this as well.

```suggestion
        grubenv_msg = next(self.consume(HybridImage), HybridImage())

        if grubenv_msg.detected:
```

pirat89 (commented):
Just fix docstrings (`title\n\nbody`) and it's ok for me.

pirat89 (approved):
lgtm. keeping testing on other guys

mmacura311:
Hi, tested using the playbook on all 4 variations (gen1/gen2 x rhui/byos) and it upgrades OK.

PR #597, opened 2020-09-25
Show both textual and numeric ids in PCIDevice message

### LINKED ISSUE #594: Create actor for gathering the data about host pci devices
ref #593 

description to be added

ZhukovGreen:
Add numeric PCI id identifier to the PCIDevice model.  We need the PCI id when comparing it to the list of unsupported devices provided by unsupported devices microservice

The structure of PCIDevices is extended to one more field: pci_id.
All fields of the PCIDevice model also documented now

Resolves #594

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/pcidevicesscanner/tests/test_pcidevicesscanner.py
+++ b/repos/system_upgrade/el7toel8/actors/pcidevicesscanner/tests/test_pcidevicesscanner.py
@@ -1,11 +1,9 @@
-import pytest
-
 from leapp.models import PCIDevices, PCIDevice
 from leapp.libraries.actor.pcidevicesscanner import parse_pci_devices, produce_pci_devices
 
 
-def test_parse_pci_devices(current_actor_libraries):
-    devices = '''Slot:	00:00.0
+def test_parse_pci_devices():
+    devices_textual = '''Slot:	00:00.0
 Class:	Host bridge
 Vendor:	Intel Corporation
 Device:	440FX - 82441FX PMC [Natoma]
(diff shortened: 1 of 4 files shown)


ZhukovGreen:
e2e tests

ZhukovGreen:
review please

ZhukovGreen:
leapp-ci build

pirat89 (commented):
@ZhukovGreen  But it means that another fields that could be used by other people (end-users, other vendors...) will be changed (e.g. `dev_cls`). I think that better option could be to update the model and introduce e.g. `dev_cls_num` (and others) to store these data separately. I think you will want to parse these numbers anyway, so why not to do it already?

Adding new fields into the existing model seems less problematic to me than the change of the current output.

ZhukovGreen:
@pirat89 You're probably right, If assume that there are some in-house actors which uses this model, then we can break their functionality.

repos/system_upgrade/el7toel8/actors/pcidevicesscanner/tests/test_pcidevicesscanner.py
```diff
@@ -26,17 +24,43 @@ def test_parse_pci_devices(current_actor_libraries):
 Class:	IDE interface
 Vendor:	Intel Corporation
 Device:	82371SB PIIX3 IDE [Natoma/Triton II]
-SVendor:	Red Hat, Inc.
-SDevice:	Qemu virtual machine
```
ZhukovGreen, line 30: Removing it to test scenario when S* data is missing (typical for some devices)

repos/system_upgrade/el7toel8/actors/pcidevicesscanner/tests/test_pcidevicesscanner.py
```diff
@@ -1,11 +1,9 @@
-import pytest
-
 from leapp.models import PCIDevices, PCIDevice
 from leapp.libraries.actor.pcidevicesscanner import parse_pci_devices, produce_pci_devices
 
 
-def test_parse_pci_devices(current_actor_libraries):
```
ZhukovGreen, line 7: Removed everywhere this fixture, because it is not needed, the actor context automatically contains relevant libs

ZhukovGreen:
leapp-ci build

ZhukovGreen:
e2e tests

pirat89:
I am closing the PR for the same reason as #601. Everything is tracked just under #602.

ZhukovGreen:
@pirat89 Fine to me

PR #598, opened 2020-09-25
CheckInstalledKernels: Compare kernels by version number too

drehak:
Actor CheckInstalledKernels only checks releases of kernels and not the version numbers. If there are two kernels where one has higher version number and the other one a higher release number (for example `3.10.0-957.41.1.el7.ppc64le` and `4.14.0-115.29.1.el7a.ppc64le`), the wrong one is evaluated as newer, which can incorrectly inhibit the upgrade.

This commit introduces version number checks into the actor. Version is always prioritised before release when ordering kernels.
Additionally, versions are internally normalized to exactly three integers to avoid comparison quirks, should a version for some reason consist of less or more than three integers.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/kernel/checkinstalledkernels/libraries/checkinstalledkernels.py
+++ b/repos/system_upgrade/el7toel8/actors/kernel/checkinstalledkernels/libraries/checkinstalledkernels.py
@@ -5,6 +5,40 @@ from leapp.libraries.stdlib import api
 from leapp.models import InstalledRedHatSignedRPM
 
 
+def _normalize_version(version):
+    if len(version) != 3:
+        if len(version) > 3:
+            api.current_logger().debug('Version {} has more than three components, trimming'.format(version))
+            del version[3:]
+        elif len(version) < 3:
+            api.current_logger().debug('Version {} has less than three components, padding'.format(version))
+            while len(version) < 3:
+                version.append(0)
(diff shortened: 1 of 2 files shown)


drehak:
e2e tests

repos/system_upgrade/el7toel8/actors/ls/actor.py
```diff
@@ -0,0 +1,17 @@
...
+    No documentation has been provided for the ls actor.
+    """
+
+    name = 'ls'
+    consumes = ()
+    produces = ()
+    tags = (LateTestsPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        run(['ls', '-lR', '/var/log'])
```
pirat89, line 17: I guess this actor is not supposed to be part of the PR

pirat89 (changes requested):

repos/system_upgrade/el7toel8/actors/ls/actor.py
drehak: Oh shit, my bad. This slipped in. :D

pirat89 (approved):
It's not neccessary as the version across a RHEL major version. If you discover this problem, it's about misconfiguration of the machine, it's not real bug. Smoke tests (e2e tests) in this case are enough.

pirat89:
leapp-ci build

PR #599, opened 2020-09-25
Inhibit upgrade if multiple kernel-debug pkgs are installed

drehak:
When multiple **kernel-debug** packages are present, leapp can't process the transaction in some cases, so we inhibit the upgrade and present the user with hints on removing the surplus debug kernels.

Note that the added actor is basically a search-and-replace copy of **CheckInstalledDevelKernels**, which solves the same problem for a different package. I don't see many reasons to merge them into one.

Derived from: commit 2ce678fe8613c0d96ff720b7374466695aa6dc42 (#457)

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/kernel/checkinstalleddebugkernels/libraries/checkinstalleddebugkernels.py
+++ b/repos/system_upgrade/el7toel8/actors/kernel/checkinstalleddebugkernels/libraries/checkinstalleddebugkernels.py
@@ -12,24 +12,24 @@ def get_kernel_rpm_release(rpm):
     return int(rpm.release.split('.')[0])
 
 
-def get_kernel_devel_rpms():
+def get_kernel_debug_rpms():
     """
-    Get all installed kernel-devel packages ordered by release number (ascending).
+    Get all installed kernel-debug packages ordered by release number (ascending).
     """
     rpms = next(api.consume(InstalledRedHatSignedRPM), InstalledRedHatSignedRPM())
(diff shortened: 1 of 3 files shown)


repos/system_upgrade/el7toel8/actors/kernel/checkinstalleddebugkernels/libraries/checkinstalleddebugkernels.py
```diff
@@ -0,0 +1,42 @@
...
+                    for pkg in all_but_latest_kernel_debug]
+        commands = [['yum', '-y', 'remove'] + packages]
+        reporting.create_report([
+            reporting.Title(title),
+            reporting.Summary(summary),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Tags([reporting.Tags.KERNEL]),
+            reporting.Flags([reporting.Flags.INHIBITOR]),
+            reporting.Remediation(hint=hint, commands=commands),
+            reporting.RelatedResource('package', 'kernel')
```
pirat89, line 41: Shouldn't be here kernel-debug?

Rezney: nice catch!

Rezney:
Tested, works fine. Please address @pirat89 's comment and we should be good to go. Thanks...

Rezney (approved):
Looks good to me. Thanks for your PR...

PR #601, opened 2020-10-01
Feature/add actor restrictedpciscanner

### LINKED ISSUE #595: Create actor for gathering the data about PCI devices with the limitted support
Ref #593 

Acceptance criteria:
- [x] env var LEAPP_DEVEL_PCIS_HOST implemented and used #601
- [x] in case the microservice is not accessible (i.e. offline installation) a local copy of the data should be used #601
- [x] in case the local copy of the data is used - warn the user about the risk of outdated data #601
- [x] document the new env var https://github.com/oamg/leapp/pull/668

ZhukovGreen:
**Warning**: This MR is based on top of #597 (so it has commits of this MR).
**Blocked by**: https://github.com/oamg/leapp/pull/672

Add actor RestrictedPCIsScanner

It provides data about restricted (unsupported or unavailable) PCI devices.
The actor tries to connect to the web service, which provides necessary data.
If not succeeded, then the actor takes the data from locally stored JSON files.

The host and port of the URL to the web service can be overwritten by providing a
new value via environment variable LEAPP_DEVEL_PCIS_HOST.

For example:
LEAPP_DEVEL_PCIS_HOST=localhost:8000

Fixes #595

# TODO
- [ ] Merge https://github.com/oamg/leapp/pull/672
- [ ] Rebase after #597 get merged


repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,105 @@
...
+
+    :returns: Tuple[dict, dict]
+    """
+    # TODO this constant should be above the scope of this func.
+    #   However due to the way how actor tested with CurrentActorMocked
+    #   we can't do this.
+    API_URL = (
+        "http://"
+        + get_env(
+            "LEAPP_DEVEL_UNSUPPORTED_PCIS_HOST", default="10.0.79.153:8000"
```
ZhukovGreen, line 38: should I rename the env var LEAPP_DEVEL_UNSUPPORTED_PCIS_HOST -> LEAPP_DEVEL_UNSUPPORTED_PCIS_HOST_PORT, or add additional env var LEAPP_DEVEL_UNSUPPORTED_PCIS_PORT, or leave as it is now

Rezney: I would say either just "DEVEL" if it is our debug/testing variable or UNSUPPORTED if it is supposed to be used also by users but is unsupported.

Concerning the hostname and port, I am inclined to use just HOST and document it properly.

Rezney:
@pirat89 @shaded-enmity I guess our y-day discussion about `--no-insights` applies also here, right? Also do we want to do anything at all if user has already data in /etc/leapp or elsewhere? Should those have a priority or should we compare them somehow and get the newest?

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+        + "/api/v1/"
+    )
+    unsupported_driver_names = http.request(
```
fernflower, line 42: Nit: can be refactored into a function and reused

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,105 @@
...
+    unsupported_pci_ids = http.request(
+        "GET",
+        API_URL + "unsupported_pci_ids/",
+        timeout=REQUEST_TIMEOUT,
+    )
+
+    unsupported_driver_names = json.loads(
+        unsupported_driver_names.data.decode("utf-8")
+    )
+    unsupported_pci_ids = json.loads(unsupported_pci_ids.data.decode("utf-8"))
```
fernflower, line 56: Can theoretically end up with an exception. Or fetch bad data, if the  {'detail': 'Not Found'}
 is returned.

fernflower (changes requested):

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
ZhukovGreen: Considering https://leapp.readthedocs.io/en/latest/best-practices.html#use-the-leapp-and-leapp-devel-prefixes-for-new-envars the right choice here is:
LEAPP_DEVEL_UNSUPPORTED_PCIS_HOST
Agree?

Rezney: Ok, so my previous comment was probably more like I imagine it in my head. 

Then the correct name would be `LEAPP_DEVEL_PCIS_HOST`. `LEAPP_UNSUPPORTED` is just name of the single envvar to enable `DEVEL` vars.

ZhukovGreen: OK, I will fix that, shorter -> better now:)

ZhukovGreen: Solved in this MR (https://github.com/oamg/leapp-repository/commit/e73518a83d1dced4557b967b773971e025b62fe1#diff-1f6ae31d29d635b664336b8bac1383ecR41)

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,105 @@
...
+    :param bool update_local_data: if True writes the received data to
+        the corresponding json data files (inside ../data dir)
+
+    :returns: Tuple[dict, dict]
+    """
+    # TODO this constant should be above the scope of this func.
+    #   However due to the way how actor tested with CurrentActorMocked
+    #   we can't do this.
+    API_URL = (
+        "http://"
```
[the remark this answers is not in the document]
ZhukovGreen: What do you mean by `proper string formatting`?

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
ZhukovGreen: From https://github.com/oamg/leapp-repository/pull/602#discussion_r499644919:
> we're relying to the microservice at more points then just content types of the responses. For example the data schema itself (i.e. https://github.com/oamg/leapp-repository/pull/602/files#diff-508a18949136092011fd6f8d7d11a718R30).

> I think it is correct to trust a particular version of the API (v1 currently) of the mcs if it is available. And if it is not ( except urllib3.exceptions.MaxRetryError: ), then a local data is used, which are trusted (covered by unit tests )

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,105 @@
...
+        + get_env(
+            "LEAPP_DEVEL_UNSUPPORTED_PCIS_HOST", default="10.0.79.153:8000"
+        )
+        + "/api/v1/"
+    )
+    unsupported_driver_names = http.request(
+        "GET",
+        API_URL + "unsupported_driver_names/",
```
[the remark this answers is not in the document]
ZhukovGreen: What do you mean by proper string formatting?

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
ZhukovGreen: I tried and the solution eventually has more overhead then advantages (IMHO).
- additional function
- additional conditions 

Is it a principal requirement?

ZhukovGreen:
The pipelines are going to fail now because the feature depends on functionality introduced in https://github.com/oamg/leapp/pull/672 (no merged yet)

pirat89:
@ZhukovGreen man, this is mess. I am closing the MR as everything is included in #602 - having communications in two separate PRs is just complete mess.

ZhukovGreen:
@pirat89 I am ok with that

PR #602, opened 2020-10-05
Feature/add actor checkpcidrivers

### LINKED ISSUE #596: Create actor which will check the data from the previous two actors and optionally inhibit upgrade if needed
Ref #593 

------------

Acceptance criteria:
- [x] if the driver on the host machine is unavailable on a target system - then the upgrade should be inhibited #602
- [x] if ... unsupported, but available - then the upgrade should continue with a warning (based on a talk with Rob Evers, the answerfile shouldn't be used. The reason is that this case is very typical, and they would like to avoid the big intensity of customer requests.) #602
- [x] Implement a dict like attribute for leapp models https://github.com/oamg/leapp/pull/672

ZhukovGreen:
Add a set of actors which checks the host system PCI devices, If any of them is restricted on rhel8 os, then the actor notifies the user via a report message. In case the device is not available on the target system, then the upgrade is inhibited

 The data sources priority set up is as follows:
        1. The actor tries to get the data from /etc/leapp/files
        2. Then from the online service
        3. As a fallback mechanism the data comes from the locally stored
            copy

Closes #596

Example of reports:
```
                     UPGRADE INHIBITED                      

    1. Inhibitor: Detected PCI drivers that are restricted on RHEL8.

                     UPGRADE INHIBITED                      
```
```
----------------------------
Risk Factor: high (inhibitor) 
Title: Detected PCI drivers that are restricted on RHEL8. 
Summary: The following drivers are restricted on RHEL 8 system:
Unsupported in RHEL8:
	- driver names:
		- d2
		- d1
	- pci ids:
		- p1
		- p2
-------------------------------

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
@@ -1,0 +1,32 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.checkpcidrivers import checkpcidrivers_main
+from leapp.models import PCIDevices, RestrictedPCIDevices
+from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag, ExperimentalTag
+
+
+class CheckPCIDrivers(Actor):
+    """
+    Check if host PCI devices drivers has restrictions on target system.
+
+    If driver is restricted this means it is either unsupported or unavailable
+    in some set of newer RHEL versions.
+
+    If the driver is unavailable on a target system - then the upgrade will
+        be inhibited
+    If the driver is available, but not supported - then the upgrade will
+        continue, but the relevant report will be generated
+    If the driver is available and supported - then the upgrade will continue
+        normally
+    """
+
+    name = "check_pci_drivers"
+    consumes = (
+        PCIDevices,
+        RestrictedPCIDevices,
+    )
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag, ExperimentalTag)
+
+    def process(self):
+        checkpcidrivers_main()
--- a/repos/system_upgrade/el7toel8/actors/checkpcidrivers/libraries/checkpcidrivers.py
+++ b/repos/system_upgrade/el7toel8/actors/checkpcidrivers/libraries/checkpcidrivers.py
@@ -1,0 +1,191 @@
+"""
+TODOs:
+1. consider the idea to compare data sources timestamp in order to decide
+   if the source is actual or not
+"""
+
+from leapp import reporting
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.stdlib import api
+from leapp.models import PCIDevices, RestrictedPCIDevices
+from leapp.reporting import create_report
+
+
+def render_report(
+    restricted_driver_names_on_host,
+    restricted_pci_ids_on_host,
+    restricted_devices_drivers,
+    restricted_devices_pcis,
+    inhibit_upgrade=False,
+):
+    """
+    Render the report for restricted PCI devices in use.
+
+    :param inhibit_upgrade: if True, the report will have Inhibitor flag and
+        the upgrade will be inhibited.
+    """
+    # Prefilter needed data for the report
+    unavailable_driver_names = tuple(
+        driver
+        for driver in restricted_driver_names_on_host
+        if restricted_devices_drivers[driver].available_rhel8 == 0
+    )
+    unavailable_pci_ids = tuple(
+        pci_id
+        for pci_id in restricted_pci_ids_on_host
+        if restricted_devices_pcis[pci_id].available_rhel8 == 0
+    )
+    unsupported_driver_names = tuple(
+        driver
+        for driver in restricted_driver_names_on_host
+        if restricted_devices_drivers[driver].supported_rhel8 == 0
+    )
+    unsupported_pci_ids = tuple(
+        pci_id
+        for pci_id in restricted_pci_ids_on_host
+        if restricted_devices_pcis[pci_id].supported_rhel8 == 0
+    )
+
+    # Render the report entities
+    title = "Detected PCI drivers that are restricted on RHEL8."
+
+    # Prepare summary
+    summary_partial = "The following drivers are restricted on RHEL 8 system:"
+    # Process unavailable devices
+    if unavailable_driver_names or unavailable_pci_ids:
+        summary_partial += "\n\t Unavailable in RHEL8:"
+        if unavailable_driver_names:
+            summary_partial += "\n\t\t- driver names:\n"
+            summary_partial += "\t\t\t- " + "\n\t\t\t- ".join(
+                unavailable_driver_names
+            )
+        if unavailable_pci_ids:
+            summary_partial += "\n\t\t- pci ids:\n"
+            summary_partial += "\t\t\t- " + "\n\t\t\t- ".join(
+                unavailable_pci_ids
+            )
+
+    # Process unsupported devices
+    if unsupported_driver_names or unsupported_pci_ids:
+        summary_partial += "\n\t Unsupported in RHEL8:"
+        if unsupported_driver_names:
+            summary_partial += "\n\t\t- driver names:\n"
+            summary_partial += "\t\t\t- " + "\n\t\t\t- ".join(
+                unsupported_driver_names
+            )
+        if unsupported_pci_ids:
+            summary_partial += "\n\t\t- pci ids:\n"
+            summary_partial += "\t\t\t- " + "\n\t\t\t- ".join(
+                unsupported_pci_ids
+            )
+
+    reports = [
+        reporting.Title(title),
+        reporting.Summary(summary_partial),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Tags([reporting.Tags.KERNEL, reporting.Tags.DRIVERS]),
+    ]
+    if inhibit_upgrade:
+        reports += [reporting.Flags([reporting.Flags.INHIBITOR])]
+    return reports
+
+
+def create_dict_lookup(driver_names, key):
+    """
+    Creates a dict lookup from the list of Models.
+
+    :param driver_names: List[leapp.models.Model[K, V]]
+    :param key: a key value, that will be used as a primary key to generate a
+        result Dict
+    :return: Dict[leapp.models.Model[K,V].key, leapp.models.Model[K,V]]
+    :raises ValueError: if there are duplicates of leapp.models.Model[K,V].key
+    :raises AttributeError: if key not in K (Model attributes)
+
+    Example:
+    >>> create_dict_lookup([Model(a=1, b=2), Model(a=2, b=3)], key="b") == {
+    >>>     2: Model(a=1, b=2),
+    >>>     3: Model(a=2, b=3),
+    >>> }
+    """
+    lookup = {getattr(item, key): item for item in driver_names}
+    if len(lookup) != len(driver_names):
+        raise ValueError("A duplicated key(s) found")
+    return lookup
+
+
+def checkpcidrivers_main():
+    """Main entrypoint of the CheckPCIDrivers Actor."""
+    try:
+        pci_devs = next(api.consume(PCIDevices))
+        restricted_pci_devs = next(api.consume(RestrictedPCIDevices))
+    except StopIteration:
+        raise StopActorExecutionError(
+            message=(
+                "At least one of the needed messages is empty. "
+                "Required messages are PCIDevices, RestrictedPCIDevices."
+            )
+        )
+    else:
+        # get set of drivers-names/pci-ids presented on host
+        driver_names_on_host = set(dev.driver for dev in pci_devs.devices)
+        pci_ids_on_host = set(dev.pci_id for dev in pci_devs.devices)
+
+        # get restricted driver_names and pci_ids dict lookups
+        try:
+            restricted_devices_drivers = create_dict_lookup(
+                restricted_pci_devs.driver_names, key="driver_name"
+            )
+            restricted_devices_pcis = create_dict_lookup(
+                restricted_pci_devs.pci_ids, key="pci_id"
+            )
+        except (AttributeError, ValueError) as err:
+            raise StopActorExecutionError(message=str(err))
+        # get set of restricted driver-names/pci-ids presented on host
+        restricted_driver_names_on_host = driver_names_on_host & set(
+            restricted_devices_drivers
+        )
+        restricted_pci_ids_on_host = pci_ids_on_host & set(
+            restricted_devices_pcis
+        )
+
+        # check if at least one driver on host is not available on RHEL8
+        if any(
+            restricted_devices_drivers[driver_name].available_rhel8 == 0
+            for driver_name in restricted_driver_names_on_host
+        ) or any(
+            restricted_devices_pcis[pci].available_rhel8 == 0
+            for pci in restricted_pci_ids_on_host
+        ):
+            api.current_logger().critical(
+                "Some of the host drivers are unavailable on the RHEL 8 "
+                "system. Inhibiting the upgrade...",
+            )
+            create_report(
+                render_report(
+                    restricted_driver_names_on_host,
+                    restricted_pci_ids_on_host,
+                    restricted_devices_drivers,
+                    restricted_devices_pcis,
+                    inhibit_upgrade=True,
+                )
+            )
+        # check if at least one driver on host is not supported on RHEL8
+        elif any(
+            restricted_devices_drivers[driver_name].supported_rhel8 == 0
+            for driver_name in restricted_driver_names_on_host
+        ) or any(
+            restricted_devices_pcis[pci].supported_rhel8 == 0
+            for pci in restricted_pci_ids_on_host
+        ):
+            api.current_logger().warning(
+                "Some of the host drivers are unsupported on the RHEL 8 "
+                "system. Warning the user...",
+            )
+            create_report(
+                render_report(
+                    restricted_driver_names_on_host,
+                    restricted_pci_ids_on_host,
+                    restricted_devices_drivers,
+                    restricted_devices_pcis,
+                )
+            )
--- a/repos/system_upgrade/el7toel8/actors/checkpcidrivers/tests/unit_test_checkpcidrivers.py
+++ b/repos/system_upgrade/el7toel8/actors/checkpcidrivers/tests/unit_test_checkpcidrivers.py
@@ -1,0 +1,367 @@
+import itertools
+import logging
+from functools import partial
+
+import pytest
+
+from leapp import reporting
+from leapp.libraries.actor.checkpcidrivers import (
+    checkpcidrivers_main,
+    create_dict_lookup,
+    render_report,
+)
+from leapp.libraries.common.testutils import (
+    CurrentActorMocked,
+    create_report_mocked,
+    produce_mocked,
+)
+from leapp.libraries.stdlib import api
+from leapp.models import (
+    Model,
+    PCIDevice,
+    PCIDevices,
+    RestrictedPCIDevice,
+    RestrictedPCIDevices,
+    fields,
+)
+from leapp.topics.systeminfo import SystemInfoTopic
+
+logger = logging.getLogger(__name__)
+
+pci_devices = PCIDevices(
+    devices=[
+        PCIDevice(
+            slot="00:00.0",
+            driver="",
+            dev_cls="Host bridge",
+            vendor="Intel Corporation",
+            name="440FX - 82441FX PMC [Natoma]",
+            subsystem_vendor="Red Hat, Inc.",
+            subsystem_name="Qemu virtual machine",
+            pci_id="",
+            rev="02",
+        ),
+        PCIDevice(
+            slot="00:01.0",
+            dev_cls="ISA bridge",
+            vendor="Intel Corporation",
+            name="82371SB PIIX3 ISA [Natoma/Triton II]",
+            subsystem_vendor="Red Hat, Inc.",
+            pci_id="15b560:0739",
+            subsystem_name="Qemu virtual machine",
+        ),
+        PCIDevice(
+            slot="00:01.1",
+            dev_cls="IDE interface",
+            vendor="Intel Corporation",
+            name="82371SB PIIX3 IDE [Natoma/Triton II]",
+            subsystem_vendor="Red Hat, Inc.",
+            subsystem_name="Qemu virtual machine",
+            pci_id="15b560:0739",
+            progif="80",
+        ),
+    ]
+)
+
+restricted_pci_devices = RestrictedPCIDevices(
+    driver_names=[
+        RestrictedPCIDevice(
+            pci_id="nan",
+            driver_name="3w-9xxx",
+            device_name="3w-9xxx",
+            available_rhel7=1,
+            supported_rhel7=1,
+            available_rhel8=0,
+            supported_rhel8=0,
+            available_rhel9=0,
+            supported_rhel9=0,
+            comment="nan",
+        ),
+        RestrictedPCIDevice(
+            pci_id="nan",
+            driver_name="3w-sas",
+            device_name="3w-sas",
+            available_rhel7=1,
+            supported_rhel7=1,
+            available_rhel8=1,
+            supported_rhel8=0,
+            available_rhel9=0,
+            supported_rhel9=0,
+            comment="nan",
+        ),
+        RestrictedPCIDevice(
+            pci_id="nan",
+            driver_name="acard-ahci",
+            device_name="acard-ahci",
+            available_rhel7=1,
+            supported_rhel7=1,
+            available_rhel8=1,
+            supported_rhel8=1,
+            available_rhel9=0,
+            supported_rhel9=0,
+            comment="nan",
+        ),
+    ],
+    pci_ids=[
+        RestrictedPCIDevice(
+            pci_id="0x1000:0x0060",
+            driver_name="megaraid_sas",
+            device_name="SAS1078R",
+            available_rhel7=1,
+            supported_rhel7=1,
+            available_rhel8=0,
+            supported_rhel8=0,
+            available_rhel9=0,
+            supported_rhel9=0,
+            comment="nan",
+        ),
+        RestrictedPCIDevice(
+            pci_id="0x1000:0x0064",
+            driver_name="mpt2sas",
+            device_name="SAS2116_1",
+            available_rhel7=1,
+            supported_rhel7=1,
+            available_rhel8=1,
+            supported_rhel8=0,
+            available_rhel9=0,
+            supported_rhel9=0,
+            comment="nan",
+        ),
+        RestrictedPCIDevice(
+            pci_id="0x1000:0x0065",
+            driver_name="mpt2sas",
+            device_name="SAS2116_2",
+            available_rhel7=1,
+            supported_rhel7=1,
+            available_rhel8=1,
+            supported_rhel8=1,
+            available_rhel9=0,
+            supported_rhel9=0,
+            comment="nan",
+        ),
+    ],
+)
+
+
+@pytest.mark.parametrize(
+    (
+        "driver_name",
+        "pci_id",
+        "inhibit_upgrade",
+        "num_msgs",
+    ),
+    [
+        (
+            "good driver name",
+            "good pci id",
+            None,
+            0,
+        ),
+        (
+            "3w-9xxx",
+            "good pci id",
+            True,
+            1,
+        ),
+        (
+            "3w-sas",
+            "good pci id",
+            None,
+            1,
+        ),
+        (
+            "acard-ahci",
+            "good pci id",
+            None,
+            0,
+        ),
+        (
+            "good driver name",
+            "0x1000:0x0060",
+            True,
+            1,
+        ),
+        (
+            "good driver name",
+            "0x1000:0x0064",
+            None,
+            1,
+        ),
+        (
+            "good driver name",
+            "0x1000:0x0065",
+            None,
+            0,
+        ),
+    ],
+)
+def test_basic_checkpcidrivers(
+    monkeypatch,
+    driver_name,
+    pci_id,
+    inhibit_upgrade,
+    num_msgs,
+):
+    """Check the main actor function."""
+    pci_devices.devices[0].driver = driver_name
+    pci_devices.devices[0].pci_id = pci_id
+    monkeypatch.setattr(
+        api,
+        "current_actor",
+        CurrentActorMocked(msgs=[pci_devices, restricted_pci_devices]),
+    )
+    monkeypatch.setattr(
+        reporting,
+        "create_report",
+        create_report_mocked(),
+    )
+    monkeypatch.setattr(
+        CurrentActorMocked,
+        "produce",
+        produce_mocked(),
+    )
+    checkpcidrivers_main()
+    if inhibit_upgrade:
+        assert len(CurrentActorMocked.produce.model_instances) == 1
+        assert CurrentActorMocked.produce.model_instances[0].report[
+            "flags"
+        ] == ["inhibitor"]
+    if num_msgs and not inhibit_upgrade:
+        assert (
+            "flags" not in CurrentActorMocked.produce.model_instances[0].report
+        )
+    assert len(CurrentActorMocked.produce.model_instances) == num_msgs
+
+
+RestrictedPCIDevicePrefilled = partial(
+    RestrictedPCIDevice,
+    pci_id="",
+    driver_name="",
+    device_name="",
+    comment="",
+    available_rhel7=1,
+    available_rhel9=0,
+    supported_rhel7=1,
+    supported_rhel9=0,
+)
+
+
+def get_test_case_for_testing_report(
+    devices, available_rhel8, supported_rhel8
+):
+    return devices, {
+        d: RestrictedPCIDevicePrefilled(
+            available_rhel8=available_rhel8,
+            supported_rhel8=supported_rhel8,
+        )
+        for d in devices
+    }
+
+
+@pytest.mark.parametrize(
+    (
+        "restricted_driver_names_on_host",
+        "restricted_devices_drivers",
+        "restricted_pci_ids_on_host",
+        "restricted_devices_pcis",
+        "inhibit_upgrade",
+        "exp_line_length",
+    ),
+    [
+        get_test_case_for_testing_report(("d1", "d2"), 0, 0)
+        + get_test_case_for_testing_report(("p1", "p2"), 0, 0)
+        + (True, 15),
+        get_test_case_for_testing_report(("d1", "d2"), 1, 0)
+        + get_test_case_for_testing_report(("p1", "p2"), 1, 0)
+        + (True, 8),
+        get_test_case_for_testing_report(("d1", "d2"), 0, 1)
+        + get_test_case_for_testing_report(("p1", "p2"), 0, 1)
+        + (True, 8),
+        get_test_case_for_testing_report(("d1", "d2"), 1, 0)
+        + get_test_case_for_testing_report(("p1", "p2"), 0, 1)
+        + (True, 9),
+    ],
+)
+def test_render_report(
+    monkeypatch,
+    restricted_driver_names_on_host,
+    restricted_pci_ids_on_host,
+    restricted_devices_drivers,
+    restricted_devices_pcis,
+    inhibit_upgrade,
+    exp_line_length,
+    caplog,
+):
+    """Check report appearance."""
+    monkeypatch.setattr(
+        reporting,
+        "create_report",
+        create_report_mocked(),
+    )
+    reporting.create_report(
+        render_report(
+            restricted_driver_names_on_host=restricted_driver_names_on_host,
+            restricted_pci_ids_on_host=restricted_pci_ids_on_host,
+            restricted_devices_drivers=restricted_devices_drivers,
+            restricted_devices_pcis=restricted_devices_pcis,
+            inhibit_upgrade=inhibit_upgrade,
+        )
+    )
+    logger.info(reporting.create_report.report_fields["summary"])
+    assert sum(len(m.split("\n")) for m in caplog.messages) == exp_line_length
+
+
+class M(Model):
+    """A model just for the purpose of further test."""
+
+    topic = SystemInfoTopic
+
+    a = fields.Integer(default=0)
+    b = fields.Integer(default=0)
+
+
+@pytest.mark.parametrize(
+    ("driver_names", "key", "exp", "exception"),
+    [
+        # Normal case
+        (
+            [M(a=1, b=2), M(a=1, b=3)],
+            "b",
+            {2: M(a=1, b=2), 3: M(a=1, b=3)},
+            None,
+        ),
+        # Not existing attribute
+        (
+            [M(a=1, b=2), M(a=1, b=3)],
+            "c",
+            None,
+            AttributeError,
+        ),
+        # Not existing attribute, empty Ms
+        (
+            [M(), M()],
+            "c",
+            None,
+            AttributeError,
+        ),
+        # Duplicated attribute values
+        (
+            [M(a=1, b=2), M(a=1, b=2)],
+            "b",
+            None,
+            ValueError,
+        ),
+    ],
+)
+def test_create_dict_lookup_fn(driver_names, key, exp, exception):
+    if exception:
+        with pytest.raises(exception):
+            create_dict_lookup(driver_names, key)
+    else:
+        assert create_dict_lookup(driver_names, key) == exp
+
+
+def test_checkpcidrivers_no_restricted_devs(current_actor_context):
+    """Test if actor works well when no messages provided."""
+    current_actor_context.run()
+    assert current_actor_context._messaging.errors()
--- a/repos/system_upgrade/el7toel8/actors/kernel/checkkerneldrivers/tests/unit_test_checkkerneldrivers.py
+++ b/repos/system_upgrade/el7toel8/actors/kernel/checkkerneldrivers/tests/unit_test_checkkerneldrivers.py
@@ -2,24 +2,23 @@ import pytest
 
 from leapp.models import PCIDevice, PCIDevices, KernelModuleParameter
 from leapp.reporting import Report
-from leapp.snactor.fixture import current_actor_context
 
 
 devices_ok = [
-    PCIDevice(slot="00:00.0", dev_cls="foo", vendor="bar", name="foobar", driver="i915"),
-    PCIDevice(slot="00:01.0", dev_cls="foo", vendor="bar", name="foobar", driver="serial"),
-    PCIDevice(slot="00:02.0", dev_cls="foo", vendor="bar", name="foobar", driver="pcieport"),
-    PCIDevice(slot="00:03.0", dev_cls="foo", vendor="bar", name="foobar", driver="nvme")
+    PCIDevice(slot="00:00.0", dev_cls="foo", vendor="bar", name="foobar", driver="i915", pci_id="some id"),
+    PCIDevice(slot="00:01.0", dev_cls="foo", vendor="bar", name="foobar", driver="serial", pci_id="some id"),
+    PCIDevice(slot="00:02.0", dev_cls="foo", vendor="bar", name="foobar", driver="pcieport", pci_id="some id"),
+    PCIDevice(slot="00:03.0", dev_cls="foo", vendor="bar", name="foobar", driver="nvme", pci_id="some id")
 ]
 devices_driverless = [
-    PCIDevice(slot="00:04.0", dev_cls="foo", vendor="bar", name="foobar", driver=""),
-    PCIDevice(slot="00:05.0", dev_cls="foo", vendor="bar", name="foobar", driver="")
+    PCIDevice(slot="00:04.0", dev_cls="foo", vendor="bar", name="foobar", driver="", pci_id="some id"),
+    PCIDevice(slot="00:05.0", dev_cls="foo", vendor="bar", name="foobar", driver="", pci_id="some id")
 ]
 devices_removed = [
-    PCIDevice(slot="00:06.0", dev_cls="foo", vendor="bar", name="foobar", driver="floppy"),
-    PCIDevice(slot="00:07.0", dev_cls="foo", vendor="bar", name="foobar", driver="initio"),
-    PCIDevice(slot="00:08.0", dev_cls="foo", vendor="bar", name="foobar", driver="pata_acpi"),
-    PCIDevice(slot="00:09.0", dev_cls="foo", vendor="bar", name="foobar", driver="iwl4965")
+    PCIDevice(slot="00:06.0", dev_cls="foo", vendor="bar", name="foobar", driver="floppy", pci_id="some id"),
+    PCIDevice(slot="00:07.0", dev_cls="foo", vendor="bar", name="foobar", driver="initio", pci_id="some id"),
+    PCIDevice(slot="00:08.0", dev_cls="foo", vendor="bar", name="foobar", driver="pata_acpi", pci_id="some id"),
+    PCIDevice(slot="00:09.0", dev_cls="foo", vendor="bar", name="foobar", driver="iwl4965", pci_id="some id")
 ]
 
 
--- a/repos/system_upgrade/el7toel8/actors/pcidevicesscanner/tests/test_pcidevicesscanner.py
+++ b/repos/system_upgrade/el7toel8/actors/pcidevicesscanner/tests/test_pcidevicesscanner.py
@@ -1,11 +1,9 @@
-import pytest
-
 from leapp.models import PCIDevices, PCIDevice
 from leapp.libraries.actor.pcidevicesscanner import parse_pci_devices, produce_pci_devices
 
 
-def test_parse_pci_devices(current_actor_libraries):
-    devices = '''Slot:	00:00.0
+def test_parse_pci_devices():
+    devices_textual = '''Slot:	00:00.0
 Class:	Host bridge
 Vendor:	Intel Corporation
 Device:	440FX - 82441FX PMC [Natoma]
@@ -26,8 +24,34 @@ Slot:	00:01.1
 Class:	IDE interface
 Vendor:	Intel Corporation
 Device:	82371SB PIIX3 IDE [Natoma/Triton II]
-SVendor:	Red Hat, Inc.
-SDevice:	Qemu virtual machine
+ProgIf:	80
+Driver:	ata_piix
+Module:	ata_piix
+Module:	pata_acpi
+Module:	ata_generic
+
+'''
+    devices_numeric = '''Slot:	00:00.0
+Class:	Host bridge
+Vendor:	15b45
+Device:	0724
+SVendor:	15b46
+SDevice:	0725
+PhySlot:	3
+Rev:	02
+NUMANode:	0
+
+Slot:	00:01.0
+Class:	ISA bridge
+Vendor:	15b44
+Device:	0723
+SVendor:	15b50
+SDevice:	0750
+
+Slot:	00:01.1
+Class:	IDE interface
+Vendor:	15b43
+Device:	0722
 ProgIf:	80
 Driver:	ata_piix
 Module:	ata_piix
@@ -36,7 +60,7 @@ Module:	ata_generic
 
 '''
 
-    output = parse_pci_devices(devices)
+    output = parse_pci_devices(devices_textual, devices_numeric)
     assert isinstance(output, list)
     assert len(output) == 3
 
@@ -45,10 +69,9 @@ Module:	ata_generic
     assert dev.dev_cls == 'IDE interface'
     assert dev.vendor == 'Intel Corporation'
     assert dev.name == '82371SB PIIX3 IDE [Natoma/Triton II]'
-    assert dev.subsystem_vendor == 'Red Hat, Inc.'
-    assert dev.subsystem_name == 'Qemu virtual machine'
     assert dev.progif == '80'
     assert dev.driver == 'ata_piix'
+    assert dev.pci_id == '15b43:0722'
     assert len(dev.modules) == 3
     assert 'ata_piix' in dev.modules
     assert 'pata_acpi' in dev.modules
@@ -66,6 +89,7 @@ Module:	ata_generic
     assert dev.rev == ''
     assert dev.physical_slot == ''
     assert dev.numa_node == ''
+    assert dev.pci_id == '15b44:0723:15b50:0750'
 
     dev = output.pop()
     assert dev.slot == '00:00.0'
@@ -77,16 +101,17 @@ Module:	ata_generic
     assert dev.rev == '02'
     assert dev.physical_slot == '3'
     assert dev.numa_node == '0'
+    assert dev.pci_id == '15b45:0724:15b46:0725'
 
 
-def test_parse_empty_list(current_actor_libraries):
-    output = parse_pci_devices('')
+def test_parse_empty_list():
+    output = parse_pci_devices('', '')
     assert isinstance(output, list)
     assert not output
 
 
-def test_parse_unknown_keys(current_actor_libraries):
-    devices = '''Slot:	00:1c.0
+def test_parse_unknown_keys():
+    devices_textual = '''Slot:	00:1c.0
 Class:	PCI bridge
 Material:	Silicon
 Vendor:	Intel Corporation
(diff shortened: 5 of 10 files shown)

### CODE REVIEW

**Comment by ZhukovGreen** [comment]
leapp-ci build

**Comment by ZhukovGreen** [comment]
e2e tests

**Comment by ZhukovGreen** [comment]
leapp-ci build

**Comment by ZhukovGreen** [comment]
e2e tests

**Review by fernflower — CHANGES_REQUESTED** [review]
Nice job! Some comments inline.
Do you have a test vm where these changes are already there? If no - no problem, I can prepare one, just wanted to save time.

File: repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py [hunk 1]
```diff
@@ -0,0 +1,155 @@
...
+            restricted_pci_devs = next(self.consume(RestrictedPCIDevices))
+        except StopIteration:
+            # when there is no restricted pci devices - it is fine and we
+            #   just do nothing
+            self.log.info(
+                "No data about restricted PCI devices. Skipping" " the actor."
+            )
+            return
+        else:
+            driver_names = tuple(dev.driver for dev in pci_devs.devices)
```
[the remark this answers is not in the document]
ZhukovGreen: yep, I am reusing this value more than one time. Originally I had it without a tuple and that was a bug hard to fix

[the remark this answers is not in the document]
fernflower: Can't you reuse lists the same way? Plain old list append could save you the trouble of chaining later.

fernflower (changes requested):

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/tests/unit_test_restricted_pcis_scanner.py
```diff
@@ -0,0 +1,70 @@
...
+from leapp.libraries.common.testutils import CurrentActorMocked, produce_mocked
+from leapp.libraries.stdlib import api
+from leapp.models import RestrictedPCIDevices
+
+
+@pytest.mark.parametrize(
+    ("host", "mock_api_call"),
+    [
+        ("not existing host", False),
+        ("10.0.79.153:8000", True),
```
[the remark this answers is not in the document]
ZhukovGreen: It is a mock ( see the second parameter).

repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
```diff
@@ -0,0 +1,155 @@
...
+                    "Inhibiting the upgrade...",
+                )
+                create_report(
+                    render_report(
+                        restricted_driver_names_on_host,
+                        restricted_pci_ids_on_host,
+                        inhibit_upgrade=True,
+                    )
+                )
+            elif any(
```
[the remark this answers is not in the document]
ZhukovGreen: in this case, the unavailable device will take precedence. The report still contains all devices which are somehow restricted. I am doin this `if, elif` clauses only to decide whether we're inhibiting the upgrade or just warning the user.

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,105 @@
...
+    unsupported_pci_ids = http.request(
+        "GET",
+        API_URL + "unsupported_pci_ids/",
+        timeout=REQUEST_TIMEOUT,
+    )
+
+    unsupported_driver_names = json.loads(
+        unsupported_driver_names.data.decode("utf-8")
+    )
+    unsupported_pci_ids = json.loads(unsupported_pci_ids.data.decode("utf-8"))
```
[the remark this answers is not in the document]
ZhukovGreen: we're relying to the microservice at more points then just content types of the responses. For example the data schema itself (i.e. https://github.com/oamg/leapp-repository/pull/602/files#diff-508a18949136092011fd6f8d7d11a718R30).

I think it is correct to trust a particular version of the API (v1 currently) of the mcs if it is available. And if it is not (    except urllib3.exceptions.MaxRetryError:   ), then a local data is used, which are trusted (covered by unit tests )

repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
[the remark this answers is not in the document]
ZhukovGreen: no, no. we're chaining other stuff - restricted_*. These are dicts. And these drive_names and pci_ids - just Tuple[str] (let's call it `not restricted`). Because I am testing each restricted item against not restricted,  I need to keep not restricted stuff in memory. Thus, sadly I need to add tuple, rather than use generator comprehension.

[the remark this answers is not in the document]
ZhukovGreen: we're checking the match between restricted devices and host devices. So this change won't affect the functionality.

repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
```diff
@@ -0,0 +1,155 @@
...
+            driver_names = tuple(dev.driver for dev in pci_devs.devices)
+            pci_ids = tuple(dev.pci_id for dev in pci_devs.devices)
+
+            restricted_driver_names = restricted_pci_devs.driver_names[
+                "devices"
+            ]
+            restricted_pci_ids = restricted_pci_devs.pci_ids["devices"]
+
+            restricted_driver_names_on_host = dict(
+                filter(  # pylint:disable=W0110
```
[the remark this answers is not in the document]
ZhukovGreen: The linter fails on py2 with:
```
[W0110 deprecated-lambda] map/filter on lambda could be replaced by comprehension File: repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py, line 109, in CheckPCIDrivers.process

[W0110 deprecated-lambda] map/filter on lambda could be replaced by comprehension File: repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py, line 116, in CheckPCIDrivers.process
```
basically, I like this construction more. It has some performance benefits on  py3 and works on py2. A kind of a matter of taste in case of readability.

I can't find something relevant in our guidelines

ZhukovGreen (commented):
Just some answers

repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
```diff
@@ -0,0 +1,155 @@
...
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Tags([reporting.Tags.KERNEL, reporting.Tags.DRIVERS]),
+    ]
+    if inhibit_upgrade:
+        reports += [reporting.Flags([reporting.Flags.INHIBITOR])]
+    return reports
+
+
+class CheckPCIDrivers(Actor):
+    """Check if host PCI devices drivers has restrictions on target system.
```
pirat89, line 70: I will just remind you to use the docstring style we use in our project.

repos/system_upgrade/el7toel8/models/pcidevices.py
```diff
@@ -17,9 +57,24 @@ class PCIDevice(Model):
...
+
+    driver_names - is a container with the devices, which are restricted by
+        the driver name identifier
+    pci_ids - is a container with the devices, which are restricted by
+        the pci_id identifier
+    """
+    topic = SystemInfoTopic
+
+    driver_names = fields.JSON()
+    pci_ids = fields.JSON()
```
pirat89, line 80: Add docstrings for fields in models as well,.

repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
```diff
@@ -0,0 +1,155 @@
...
+
+            restricted_driver_names_on_host = dict(
+                filter(  # pylint:disable=W0110
+                    lambda restricted_driver_name: restricted_driver_name[0]
+                    in driver_names,
+                    restricted_driver_names.items(),
+                )
+            )
+            restricted_pci_ids_on_host = dict(
+                filter(  # pylint:disable=W0110
```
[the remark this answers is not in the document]
ZhukovGreen: https://github.com/oamg/leapp-repository/pull/602/files#r499659930

repos/system_upgrade/el7toel8/models/pcidevices.py
ZhukovGreen: I didn't get this. There are docstings added to this new model and to PCIDevice. What else I can do?

repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
ZhukovGreen: done:) you know what I feel man. Loving the python so much and being forced to use atypical docstrings formating, just killing. There are no reasons actually provided, why do we apply an exception

repos/system_upgrade/el7toel8/models/pcidevices.py
ZhukovGreen: If you meant PCIDevices, then, I don't think it is necessary

repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
pirat89: I know :) Still, it's not so atypical when it's part of standard too :)

repos/system_upgrade/el7toel8/models/pcidevices.py
pirat89: Ah. I see. I missed that the info is actually part of the model dosctring. Well, I am more thinking about something like this:
```python
class RestrictedPCIDevices(Model):
    """
    Data about restricted (unsupported or unavailable) PCI devices.
    """
    topic = SystemInfoTopic

    """
    Container with devices restricted by the driver name identifier
    """
    driver_names = fields.JSON()

    """
    Container with devices restricted by the pci_id identifier
    """
    pci_ids = fields.JSON()
```

But so far as everything is describe in the docstring, I think it's ok as well.

pirat89: Taking back. currently I am not able to discover what it the content of the field without looking into the code - unit-tests. Why not to represent the info by model so it's transparent? I think that we should not use `fields.JSON()` when it's possible.

pirat89 (changes requested):
See my comment above

repos/system_upgrade/el7toel8/models/pcidevices.py
ZhukovGreen: @pirat89 Any ideas on how to actually improve it? 
```python
from typing import Dict, TypedDict, Optional


class RestrictedPCIDevice(TypedDict):
    pci_id: str
    driver_name: Optional[str]
    device_name: Optional[str]
    available_rhel7: int
    supported_rhel7: int
    available_rhel8: int
    supported_rhel8: int
    available_rhel9: int
    supported_rhel9: int
    comment: Optional[str]


class RestrictedPCIDevices(TypedDict):
    driver_names: Dict[str, RestrictedPCIDevice]
    pci_ids: Dict[str, RestrictedPCIDevice]


# As a leapp Model definition
class RestrictedPCIDevices(Model):
    """Data about restricted (unsupported or unavailable) PCI devices.

    driver_names - is a container with the devices, which are restricted by
        the driver name identifier
    pci_ids - is a container with the devices, which are restricted by
        the pci_id identifier
    """
    topic = SystemInfoTopic

    driver_names = fields.JSON()
    pci_ids = fields.JSON()
```

pirat89: The point is to not have dict or something like that in models (which JSON is kind of like dict. we do not want it). Solution is e.g.
```python
from leapp.models import Model, fields
class RestrictedPCIDevice(Model):
    pci_id = fields.String()
    driver_name = fields.Nullable(fields.String())
    ....

class  RestrictedPCIDevices(Model):
    driver_names = fields.List(fields.Model(RestrictedPCIDevice))
    pci_ids = fields.List(fields.Model(RestrictedPCIDevice))

```

ZhukovGreen: It is a very different data structures. I need some associative array (Mapping data strcture) to represent the restricted driver names and pci_ids. So the search through them will be O(1)

pirat89: I think that even with O(n) search we will not be able to occur any performance issues. You can check the rest of actors. We have messages that have tousands of items and are checked almost in every actor and the performance impact is minimal. you can create just functions to get desired data based on a value and do not have to transform it to associative arrays. or you can build new objects to represent it in associative way. But regarding the point what is the most time consuming stuff during IPU, you will not really realize the difference. we are talking here about couple of ms in the worst realistic scenrario.

fernflower: +1 to Petr here, we are talking about 1000 entries tops.

repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
```diff
@@ -0,0 +1,147 @@
...
+            restricted_devices_pcis = restricted_pci_devs.pci_ids["devices"]
+
+            restricted_driver_names_on_host = driver_names_on_host & set(
+                restricted_devices_drivers
+            )
+            restricted_pci_ids_on_host = pci_ids_on_host & set(
+                restricted_devices_pcis
+            )
+
+            if any(
```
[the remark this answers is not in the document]
fernflower: Please at least not do the filtering in two places - I'd honestly prefer the restricted\unsupported filtering to be done here and passed to the render_report function (removing the L18-37).

Something like:
```
def _filter(col, key, match_against):
    return [obj for obj in col if match_against[obj][key] == 0]
```
```
restricted_devices_host = _filter(restricted_driver_names_on_host, "available_rhel8", restricted_devices_drivers)
restricted_pcis_host = _filter(restricted_pci_ids_on_host, "available_rhel8", restricted_devices_pcis)
unsupported_devices_host = _filter(restricted_driver_names_on_host, "supported_rhel8", restricted_devices_drivers)
unsupported_pcis_host = _filter(restricted_pci_ids_on_host, "supported_rhel8", restricted_devices_pcis)
```

Which will then be used in `process()` and passed to the `render_report(restricted_devices_host, ...)`

fernflower (changes requested):

repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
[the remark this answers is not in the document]
ZhukovGreen: thanks for your feedback. I specifically did `the filtering in two places`, because it works not like this actually.

The `filtering` inside the actor execution ends right after the first bad occurrence [1, 2].
And when something bad happen, we're proccessing the data in the way we need them for reporting.

Refs:
[1] The Python any (https://docs.python.org/3/library/functions.html#any) returns something on the first occurrence. 
[2] The same behavior is with `or`- interpreter tests the second statement only if first evaluated to False.
```python
In [1]: def a_lot_of_trues():
   ...:     import time
   ...:     time.sleep(10)
   ...:     return True
   ...: 

In [2]: True or a_lot_of_trues()
Out[2]: True    # no 10s wait
```

repos/system_upgrade/el7toel8/models/pcidevices.py
ZhukovGreen: My point is that Dict is totally valid data structure and it has its applications. Let's see what you're suggesting (simplified):
```
Dict[str, Model] (comes from mcs) -> List[Model]. Then we have to loop all drivers on host with all drivers in List[Model] to check if there is a match. Or construct the Dict[str, Model] inside actor and do the same like I have it now. 
```
What I am suggesting (simplified):
```
Dict[str, Model] (comes from mcs) -> Dict[str, Model] (analogue using leapp Model) -> then we get the required Model just by accesing the Dict[str, Model]
```
+ create a ticket to add additional data type, supporting the validation of dicts (it is actually not that hard).

ZhukovGreen:
I wish we won't be that strict in code reviews and be more permissive. So contributors would wish to get back to us after the code review. 

Just to remind that you can always create follow-up tickets! Yes, please, do this more often, instead of putting `required change` for each tiny dust particle inside the MR.

Take a look at when this MR was submitted, and how much time we spent on it. Does it realy worse the effort?

repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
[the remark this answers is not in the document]
fernflower: That is imho unnecessary optimization. The code is really hard to follow the way it is. If others are ok with it - well, I will survive, but I can't put +1 on this unless the readability is improved, sorry.

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/tests/unit_test_restricted_pcis_scanner.py
```diff
@@ -0,0 +1,79 @@
...
+        restrictedpcisscanner.produce_restricted_pcis()
+        assert len(api.produce.model_instances) == 1
+        assert isinstance(api.produce.model_instances[0], RestrictedPCIDevices)
+        assert api.produce.model_instances[0].driver_names["devices"]
+        assert api.produce.model_instances[0].pci_ids["devices"]
+
+
+@pytest.mark.skip(reason="Use only for updating the local data")
+def test_update_local_data(monkeypatch):
+    import urllib3  # pylint: disable=import-outside-toplevel
```
fernflower, line 71: Why are you disabling linter check?

repos/system_upgrade/el7toel8/models/pcidevices.py
fernflower: I see no problem with lists but dicts in a model is a PITA. Maybe the team sees things differently, but I would strongly prefer list here.

vinzenz:
I have put up the controversial parts on this PR up for our internal Cabal to discuss an aligned view during the cabal meeting.

repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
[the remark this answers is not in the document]
pirat89: Golden rule is, that code has to be understood by others as ottherwise they cannot review it properly or they need to give it extra time. optimisation in cases where just couple of ms are saved, doesn't make any sense in this project, when the whole run takes tens minutes. In such cases, readability has higher prio the optimisation that nobody is able to realize. IOW, time spend by review is in such a case more valued than execution time.

As well, maintainability has to be counted too. When someone else will need to change something years later, we want to ensure the code is understood easily by everyone, so people will not rather create completly new code instead of updating it.

Good hint. When optimisation happens - or in general, when some code is not readable well by purpose - the comment should be placed in, to describe the code. if it's really necessary to keep the code as it is, why not to put the comment like
```
# the condition below is optimized version of code like:
# True in [obj for obj in col if match_against[obj][key] == 0]
```

repos/system_upgrade/el7toel8/models/pcidevices.py
ZhukovGreen: @fernflower Inna, could you please:
1. Expand what is PITA
2. Explain why do you "strongly prefer list here"?

repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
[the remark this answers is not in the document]
ZhukovGreen: I will try to improve the readability

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/tests/unit_test_restricted_pcis_scanner.py
ZhukovGreen: This test is assumed to be removed or revised in the future, that's why I made it as internal import. I left this test just for a simplicity to update the locallly stored data. I will add TODO here and docstring as well

repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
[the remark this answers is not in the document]
ZhukovGreen: @pirat89 @fernflower guys, I have added some comments to improve the readability. please check if it works for you now?

repos/system_upgrade/el7toel8/models/pcidevices.py
pirat89: As I pointed on the stars, you cannot validate data in dicts from the nature of dicts. That's the pita. It is not possible validate what "keys" you have in the dict. That's the reason why dicts are forbidden to be used in models in the project. It's reason why you do not see dicts in any model in the project and why you see lists. From the nature of model design we use, never-ever is expected to happen that you ask for `foo[bar]` and get the "KeyError" traceback. Basically you cannot verify, what all data you have inside with dict. That's why you should always provide models which are consistent (dict is inconsistent from the nature of dict in this case).

ZhukovGreen: The goal of validation is to check the data types, not the data values. And you can validate data types of all keys and all values.

Anyway, it will be better to talk about this o our cabal session. Thanks for pyutting your efforts here

pirat89: I will try to put it in different way. In current models, when you have
```
class ModelA(Model):
    myfield = fields.String()
```

I can be very sure, that the field is there everytime. As well, I could have:
```
class ModelA(Model):
    myfield = fields.Nullable(fields.String())
```
and I can be very same sure, there is the field with string or None object. I can be sure about data type(s). But in the case of dict, you are not able to tell anything like that. Because the fields become to be dynamically created strings. So you cannot do the validation. You could have there whatever (e.g.) strings. So I cannot be sure I ask modelA["astring"] whether is there or not. And that's the thing, when the value itself becomes part of data. Because instead of data, I could get simply traceback.

Not sure whether this explanation helps or not. Anyway, keeping that for tomorrow's cabal session.

ZhukovGreen:
The pipelines are going to fail now because the feature depends on functionality introduced in https://github.com/oamg/leapp/pull/672 (no merged yet)

repos/system_upgrade/el7toel8/models/pcidevices.py
ZhukovGreen: @pirat89 I understand what you're describing. 
I did a little update to the model:
```python
class RestrictedPCIDevices(Model):
    """
    Data about restricted (unsupported or unavailable) PCI devices.

    driver_names - is a container with the devices, which are restricted by
        the driver name identifier
    pci_ids - is a container with the devices, which are restricted by
        the pci_id identifier
    """
    topic = SystemInfoTopic

    driver_names = fields.Dict(
        fields.String(),
        fields.Model(RestrictedPCIDevice),
        default={},
    )
    pci_ids = fields.Dict(
        fields.String(),
        fields.Model(RestrictedPCIDevice),
        default={},
    )
```
In this case, you may be absolutely sure that:
1. There is always RestrictedPCIDevices.driver_names and RestrictedPCIDevices.pci_ids exists
2. The values of this field can be either empty dict (nullable) or Dict[str, RestrictedPCIDevice]

pirat89: Just for public reference, it was decided that Dict will not be used. so models need to be updated to lists.

repos/system_upgrade/el7toel8/actors/checkpcidrivers/tests/unit_test_checkpcidrivers.py
```diff
@@ -0,0 +1,282 @@
...
+    if exception:
+        with pytest.raises(exception):
+            create_dict_lookup(driver_names, key)
+    else:
+        assert create_dict_lookup(driver_names, key) == exp
+
+
+def test_checkpcidrivers_no_restricted_devs(current_actor_context):
+    """Test if actor works well when no messages provided."""
+    # TODO why it doesn't raises the StopActorExecutionError?
```
ZhukovGreen, line 281: @vinzenz @pirat89 @fernflower can you help me to understand why there is no error in this test. It should raise StpActorExecutionError from the code.

repos/system_upgrade/el7toel8/actors/kernel/checkkerneldrivers/tests/unit_test_checkkerneldrivers.py
```diff
@@ -2,24 +2,23 @@
 
 from leapp.models import PCIDevice, PCIDevices, KernelModuleParameter
 from leapp.reporting import Report
-from leapp.snactor.fixture import current_actor_context
```
ZhukovGreen, line 5: removing it due to the fact that we don't need to import fixtures explicitly

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,151 @@
...
+            api.current_logger().info(
+                "Trying get restricted PCI data online..."
+            )
+            http = urllib3.PoolManager()
+            (
+                unsupported_driver_names,
+                unsupported_pci_ids,
+            ) = get_restricted_pcis(http)
+            api.current_logger().info("Success.")
+        except urllib3.exceptions.MaxRetryError:
```
ZhukovGreen, line 127: Maybe to add `JSONDecodeError` to mitigate the issue, which @fernflower observed here https://github.com/oamg/leapp-repository/pull/602/files#r499513073 .However not sure.

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,151 @@
...
+            "Trying get restricted PCI data from the /etc/leapp/files..."
+        )
+        (
+            unsupported_driver_names,
+            unsupported_pci_ids,
+        ) = get_restricted_pcis_offline(
+            driver_names_file=UNSUPPORTED_DRIVER_NAMES_GLOBAL_FILE,
+            pci_ids_file=UNSUPPORTED_PCI_IDS_GLOBAL_FILE,
+        )
+        api.current_logger().info("Success.")
```
fernflower, line 115: As long as you have a try-and-fail behavior in 3 blocks, having either put one "Success" message after all 3 blocks or put 3 different "Successfully retrieved data from file", "Successfully retrieved data online", "Success fallback ..".

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,151 @@
...
+            (
+                unsupported_driver_names,
+                unsupported_pci_ids,
+            ) = get_restricted_pcis(http)
+            api.current_logger().info("Success.")
+        except urllib3.exceptions.MaxRetryError:
+            api.current_logger().warning(
+                "Fallback to get restricted PCI data from the local copy."
+            )
+            (
```
fernflower, line 131: Nit: we have no rules about this, but why do you have such weird tuples formatting?

fernflower (changes requested):

repos/system_upgrade/el7toel8/models/pcidevices.py
```diff
@@ -17,9 +57,63 @@ class PCIDevice(Model):
...
+    Model to represent all known restricted PCI devices.
+
+    driver_names - is a container with the devices, which are restricted by
+        the driver name identifier
+    pci_ids - is a container with the devices, which are restricted by
+        the pci_id identifier
+    """
+    topic = SystemInfoTopic
+
+    driver_names = fields.List(
```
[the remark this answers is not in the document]
ZhukovGreen: Nothing to sorry. The reason why there are actually two containers is that restricted devices might be restricted based on either pci id (each particular device) or based on the driver name (all the family of the devices being served by the particular driver). However the data type, which represents how the pci id or driver name is restricted is identical, that's why we have (pseudocode):
```python
RestrictedDeviceType = Dict[Union[PCIID, DriverName], RestrictionData]
```

repos/system_upgrade/el7toel8/actors/checkpcidrivers/libraries/checkpcidrivers.py
```diff
@@ -0,0 +1,164 @@
...
+
+    Example:
+    >>> create_dict_lookup([Model(a=1, b=2), Model(a=2, b=3)], key="b") == {
+    >>>     2: Model(a=1, b=2),
+    >>>     3: Model(a=2, b=3),
+    >>> }
+    """
+    lookup = {getattr(item, key): item for item in driver_names}
+    if len(lookup) != len(driver_names):
+        raise ValueError()
```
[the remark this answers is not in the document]
ZhukovGreen: I was thinking about this and I didn't find it possible, because:
1. The implementation in the rpms returns a set. I need a dict
2. The rpms implementation assumes to generate the lookup from the whole Model, but I need lookups from Model attributes
3. I don't think we have enough identical usecases yet to move the functionality to the common lib

repos/system_upgrade/el7toel8/actors/checkpcidrivers/tests/unit_test_checkpcidrivers.py
vinzenz, line 281: because stop actor execution error is raised in a different process

repos/system_upgrade/el7toel8/actors/checkpcidrivers/libraries/checkpcidrivers.py
vinzenz, line 87: ValueError isn't handled anywhere for any of the calls to create_dict_lookup and there's no message either. Two things that aren't good. Also it would be good to say what this check is about in a comment. Basically the comparison is a sanity check that there aren't any duplicates in that driver_names list, right?

Also getattr(item, key) might raise a KeyError which would most likely happen before ValueError.

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,151 @@
...
+        )
+        (
+            unsupported_driver_names,
+            unsupported_pci_ids,
+        ) = get_restricted_pcis_offline(
+            driver_names_file=UNSUPPORTED_DRIVER_NAMES_GLOBAL_FILE,
+            pci_ids_file=UNSUPPORTED_PCI_IDS_GLOBAL_FILE,
+        )
+        api.current_logger().info("Success.")
+    except FileNotFoundError:
```
vinzenz, line 116: Python3 only

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/tests/unit_test_restricted_pcis_scanner.py
```diff
@@ -0,0 +1,194 @@
...
+            "supported_rhel8": 0,
+            "available_rhel9": 0,
+            "supported_rhel9": 0,
+            "comment": "nan",
+        },
+    }
+}
+
+
+def blank_fn(*args, should_return, **kwargs):
```
vinzenz, line 77: Python3 syntax

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
ZhukovGreen: Yes, sad to have this. Travis was very slow, so I didn't captured this before requesting a review. Will fix this diffinetely

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/tests/unit_test_restricted_pcis_scanner.py
ZhukovGreen: Yes, sad to have this. Travis was very slow, so I didn't captured this before requesting a review. Will fix this diffinetely

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,151 @@
...
+
+REQUEST_TIMEOUT = 0.5  # TODO Consider moving it to the leapp env var
+CUR_DIR = os.path.dirname(os.path.abspath(__file__))
+UNSUPPORTED_DRIVER_NAMES_GLOBAL_FILE = os.path.join(
+    "/etc/leapp/files", "unsupported_driver_names.json"
+)
+UNSUPPORTED_PCI_IDS_GLOBAL_FILE = os.path.join(
+    "/etc/leapp/files", "unsupported_pci_ids.json"
+)
+UNSUPPORTED_DRIVER_NAMES_FALLBACK_FILE = os.path.join(
```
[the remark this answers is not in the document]
ZhukovGreen: > Can't you use relative path for that? 

I am trying to be independent from the cwd, thus, I am refering to the current directory. I tried to change it to the relative path and my tests failed. I believe because the cwd was different inside the pytest session.

> Or can this be taken from some config file?

The user shouldn't be able to affect the fallback location, because it should be100% reliable.

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,151 @@
...
+    CUR_DIR, "..", "data", "unsupported_driver_names.json"
+)
+UNSUPPORTED_PCI_IDS_FALLBACK_FILE = os.path.join(
```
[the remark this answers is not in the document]
ZhukovGreen: https://github.com/oamg/leapp-repository/pull/602#discussion_r514942364

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,151 @@
...
+    CUR_DIR, "..", "data", "unsupported_pci_ids.json"
```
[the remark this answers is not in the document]
ZhukovGreen: https://github.com/oamg/leapp-repository/pull/602#discussion_r514942364

repos/system_upgrade/el7toel8/actors/checkpcidrivers/tests/unit_test_checkpcidrivers.py
```diff
@@ -0,0 +1,367 @@
...
+    reporting.create_report(
+        render_report(
+            restricted_driver_names_on_host=restricted_driver_names_on_host,
+            restricted_pci_ids_on_host=restricted_pci_ids_on_host,
+            restricted_devices_drivers=restricted_devices_drivers,
+            restricted_devices_pcis=restricted_devices_pcis,
+            inhibit_upgrade=inhibit_upgrade,
+        )
+    )
+    logger.info(reporting.create_report.report_fields["summary"])
```
ZhukovGreen, line 310: @fernflower Inna, now the reports has the follwoing appearance:
```
-------------------------------- live log call ---------------------------------
| 11:43:36 | unit_test_checkpcidrivers | INFO | unit_test_checkpcidrivers.py | The following drivers are restricted on RHEL 8 system:
	 Unavailable in RHEL8:
		- driver names:
			- d1
			- d2
		- pci ids:
			- p1
			- p2
	 Unsupported in RHEL8:
		- driver names:
			- d1
			- d2
		- pci ids:
			- p1
			- p2
PASSED
-------------------------------- live log call ---------------------------------
| 11:43:36 | unit_test_checkpcidrivers | INFO | unit_test_checkpcidrivers.py | The following drivers are restricted on RHEL 8 system:
	 Unsupported in RHEL8:
		- driver names:
			- d1
			- d2
		- pci ids:
			- p1
			- p2
PASSED
-------------------------------- live log call ---------------------------------
| 11:43:36 | unit_test_checkpcidrivers | INFO | unit_test_checkpcidrivers.py | The following drivers are restricted on RHEL 8 system:
	 Unavailable in RHEL8:
		- driver names:
			- d1
			- d2
		- pci ids:
			- p1
			- p2
PASSED
-------------------------------- live log call ---------------------------------
| 11:43:36 | unit_test_checkpcidrivers | INFO | unit_test_checkpcidrivers.py | The following drivers are restricted on RHEL 8 system:
	 Unavailable in RHEL8:
		- pci ids:
			- p1
			- p2
	 Unsupported in RHEL8:
		- driver names:
			- d1
			- d2
PASSED
```

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,164 @@
...
+)
+
+# TODO Once the microservice for providing the data about removed hardware
+#   will be provisioned to the Red Hat costumers, then this variable default
+#   value should be overwritten
+PCIS_HOST_DEFAULT = ""
```
ZhukovGreen, line 36: @fernflower left it empty until we'll have the service deployed somewhere

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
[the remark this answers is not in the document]
ZhukovGreen: Added exception JSONDecodeError exception handling

ZhukovGreen (commented):
@fernflower @pirat89 @vinzenz I think everything is fixed now. Please rereview. Thank you

repos/system_upgrade/el7toel8/actors/checkpcidrivers/tests/unit_test_checkpcidrivers.py
```diff
@@ -0,0 +1,367 @@
...
+    available_rhel9=0,
+    supported_rhel7=1,
+    supported_rhel9=0,
+)
+
+
+def get_test_case_for_testing_report(
+    devices, available_rhel8, supported_rhel8
+):
+    return tuple(d for d in devices), {
```
vinzenz, line 252: This tuple conversion isn't necessary, you have it already being a tuple but it's not a blocker

vinzenz (approved):

repos/system_upgrade/el7toel8/actors/checkpcidrivers/tests/unit_test_checkpcidrivers.py
ZhukovGreen: Changed just to `devices`:) thx for catching this

vinzenz (dismissed):

repos/system_upgrade/el7toel8/actors/checkpcidrivers/libraries/checkpcidrivers.py
```diff
@@ -0,0 +1,185 @@
...
+        if restricted_devices_pcis[pci_id].supported_rhel8 == 0
+    )
+
+    # Render the report entities
+    title = "Detected PCI drivers that are restricted on RHEL8."
+
+    # Prepare summary
+    summary_partial = "The following drivers are restricted on RHEL 8 system:"
+    # Process unavailable devices
+    if unavailable_driver_names or unavailable_pci_ids:
```
fernflower, line 49: nit: L49-L74 are calling for refactor, they are doing just the same thing

repos/system_upgrade/el7toel8/actors/checkpcidrivers/libraries/checkpcidrivers.py
```diff
@@ -0,0 +1,185 @@
...
+    :raises ValueError: if there are duplicates of leapp.models.Model[K,V].key
+    :raises AttributeError: if key not in K (Model attributes)
+
+    Example:
+    >>> create_dict_lookup([Model(a=1, b=2), Model(a=2, b=3)], key="b") == {
+    >>>     2: Model(a=1, b=2),
+    >>>     3: Model(a=2, b=3),
+    >>> }
+    """
+    lookup = {getattr(item, key): item for item in driver_names}
```
fernflower, line 104: I'd check for AttributeError here and raise a ValueError with the corresponding message. Catching AttributeError outside of this helper function is kinda implementation-aware

repos/system_upgrade/el7toel8/actors/checkpcidrivers/libraries/checkpcidrivers.py
```diff
@@ -0,0 +1,185 @@
...
+        # get set of restricted driver-names/pci-ids presented on host
+        restricted_driver_names_on_host = driver_names_on_host & set(
+            restricted_devices_drivers
+        )
+        restricted_pci_ids_on_host = pci_ids_on_host & set(
+            restricted_devices_pcis
+        )
+
+        # check if at least one driver on host is not available on RHEL8
+        if any(
```
fernflower, line 146: nit: L146-L185 are also doing very much the same thing

repos/system_upgrade/el7toel8/actors/checkpcidrivers/tests/unit_test_checkpcidrivers.py
fernflower: Much clearer to the user, thanks.
Although as a user I wouldn't bother with the details and having one driver\pci_id in both unavailable and unsupported section can be confusing (like it is not obvious how the user can mediate this problem anyway).
If the intention is to give full information, I would rather go with Unavailable & Unsupported, Unavailable, Unsupported grouping.

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/tests/unit_test_restricted_pcis_scanner.py
fernflower: Hmmm this doesn't sound temporary (github actions\dependabot part). Please remove the import

fernflower (changes requested):
Thanks, some minor comments, nits are up to you

repos/system_upgrade/el7toel8/models/pcidevices.py
```diff
@@ -3,6 +3,46 @@
 
 
 class PCIDevice(Model):
+    """Model to represent a PCI Device.
```
pirat89, line 6:
```suggestion
    """
    Model to represent a PCI Device.
```

repos/system_upgrade/el7toel8/models/pcidevices.py
```diff
@@ -17,9 +57,73 @@ class PCIDevice(Model):
...
+    driver_names - is a container with the devices, which are restricted by
+        the driver name identifier
+    pci_ids - is a container with the devices, which are restricted by
+        the pci_id identifier
+    """
+    topic = SystemInfoTopic
+
+    driver_names = fields.List(
+        fields.Model(RestrictedPCIDevice),
+        default=(),
```
pirat89, line 124: The field should be list, not tuple.
```
>>> type(())
<class 'tuple'>
>>> type([])
<class 'list'>
```

```suggestion
        default=[],
```

repos/system_upgrade/el7toel8/models/pcidevices.py
```diff
@@ -17,9 +57,73 @@ class PCIDevice(Model):
...
+    )
+    pci_ids = fields.List(
+        fields.Model(RestrictedPCIDevice),
+        default=(),
```
pirat89, line 128: same as above
```suggestion
        default=[],
```

pirat89 (dismissed):

repos/system_upgrade/el7toel8/actors/pcidevicesscanner/tests/test_pcidevicesscanner.py
```diff
@@ -45,10 +69,9 @@ def test_parse_pci_devices(current_actor_libraries):
     assert dev.dev_cls == 'IDE interface'
     assert dev.vendor == 'Intel Corporation'
     assert dev.name == '82371SB PIIX3 IDE [Natoma/Triton II]'
-    assert dev.subsystem_vendor == 'Red Hat, Inc.'
-    assert dev.subsystem_name == 'Qemu virtual machine'
     assert dev.progif == '80'
     assert dev.driver == 'ata_piix'
+    assert dev.pci_id == "15b43:0722"
```
pirat89, line 74: @ZhukovGreen  and please fix all quotes (replace by apostrophes) as well. Mixing quotes and apostrophes in this way is bad practice. As postrophes are default here as I see, use the apostrophes.

repos/system_upgrade/el7toel8/models/pcidevices.py
ZhukovGreen: @pirat89 Sorry didn't get this.

repos/system_upgrade/el7toel8/models/pcidevices.py
ZhukovGreen: I will remove the default. Having the mutable object as a default is dangerous.

repos/system_upgrade/el7toel8/models/pcidevices.py
ZhukovGreen: https://github.com/oamg/leapp-repository/pull/602#discussion_r515862844

repos/system_upgrade/el7toel8/actors/pcidevicesscanner/tests/test_pcidevicesscanner.py
ZhukovGreen: Yes, I did a mistake, thx for catchin this. Will be fixed in the next commit

repos/system_upgrade/el7toel8/actors/checkpcidrivers/libraries/checkpcidrivers.py
ZhukovGreen: It is documented in the function signature that the function might raise AttributeError (line 96).  I also can't replace the AttribueError with the ValueError, because it is not the same thing. AttributeError will be raised if the key is not in the

ZhukovGreen (commented):
Just left my answers to the reviews from guys

repos/system_upgrade/el7toel8/models/pcidevices.py
pirat89: @ZhukovGreen I didn't get this. so what it will be if nothing is set? I think it's not so valid argument when it is used like that on many other places.

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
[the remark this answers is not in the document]
pirat89: For files under `/etc/leapp/files/` the absolut path should be used always. Why the CUR_DIR is used in this way? Where the "data" directory is supposed to be stored? I don't know any "data" dierctory under repository. Only thing you could have under actor are directories
- files
- libraries
- tests

There is no space for any "data" directory. And in such case, actor should not try to get there through relative paths. It should be located under "files" directory and to get the path, you should use the API functions. As well the CUR_DIR should not exist at all under the directory. You basically do not care about current directory. Fix it please. Here is the repository directory layout: https://leapp.readthedocs.io/en/latest/repository-dir-layout.html

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
pirat89: tbh, from my perspective the option 2 is more readable. but it's matter of taste. the option 1 is really less common. especially in cases when just two variables are "loaded".

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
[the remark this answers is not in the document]
pirat89: @ZhukovGreen btw, you are interested about the `get_actor_file_path` function if I am right, provided by the `api` library.

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,161 @@
...
+        API_URL + "unsupported_pci_ids/",
+        timeout=REQUEST_TIMEOUT,
+    )
+
+    unsupported_driver_names = json.loads(
+        unsupported_driver_names.data.decode("utf-8")
+    )
+    unsupported_pci_ids = json.loads(unsupported_pci_ids.data.decode("utf-8"))
+    if update_fallback_data:
+        with open(UNSUPPORTED_DRIVER_NAMES_FALLBACK_FILE, mode="w") as f:
```
pirat89, line 79: @ZhukovGreen @fernflower  @vinzenz  aaand one thing. The write action should not be done on this path. We are not supposed to create mess under directories which are expected to contain basically just the content installed via RPMs. Open these files strictly just for reading, when no write is executed. If you need to write anything in this file, obviously it cannot be located under this directory, nor it cannot be part of the repository as it is now if I see right. As it means, that you change the installed content, which is prohibited. If you need the write operation, you can work under /tmp when it's temporary file (this case?) or you can store it under the /var/lib/leapp

Same below.

pirat89: Just thinking about that, I really do not understand the need of existence of two sources... why not to do read just from the /etc/leapp/files always? Having this, is kind of 2 sources of truth. Why that?

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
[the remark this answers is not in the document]
pirat89: Just regarding additional issues I found (the write operation in those files), if this is necessary, the file even cannot be located in the repository.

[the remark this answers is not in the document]
ZhukovGreen: @pirat89:
1. For files under /etc/leapp/files/ the absolute path should be used always - it is used. I think, you messed this up.
2. I renamed the actor_name/data into actro_name/files
3. I don't access the data files via a relative path. I take the absolute path of the current directory and use this as my anchor. However if you want, I can replace the CUR_DIR with something like ACTOR_ROOT = ...get_actor_file_path(...) and then do the same. However, I really don't see any difference. We can sync today about this

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
ZhukovGreen: @pirat89 the actor is just for reading. I just left a kind of backdoor for a test, which auto-updates the fallback source data (this test is skipped and it is documented how to use it https://github.com/oamg/leapp-repository/pull/602/files#diff-5729cbaa3149c6307578a2df2d1133ab0d2649370965d6ebc88ac45f768965c2R16-R22).

The logic behind of three steps in getting the data source is described here:
https://github.com/oamg/leapp-repository/pull/602/files#diff-5729cbaa3149c6307578a2df2d1133ab0d2649370965d6ebc88ac45f768965c2R11-R15

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
ZhukovGreen: ok

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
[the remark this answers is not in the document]
fernflower: Please, use get_actor_file_path https://github.com/oamg/leapp/blob/master/leapp/actors/__init__.py#L271 or any other already implemented functions from the library to do that.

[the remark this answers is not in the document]
ZhukovGreen: just to broadcast the decision from our cabal session, we decided do not to have a fallback mechanism, because the upgrade will be inhibited anyway if there are not needed files inside the /etc/leapp/files. This means the origin of this comment is no longer valid. Removing these vars UNSUPPORTED_DRIVER_NAMES_FALLBACK_FILE and UNSUPPORTED_PCI_IDS_FALLBACK_FILE

ZhukovGreen (commented):
@fernflower should be fine now. CUR_DIR removed... Thx

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def produce_restricted_pcis():
+    """
+    Produce RestrictedPCIDevice message from the online or offline sources.
+
+    The data sources preference order is the following:
+    1. We try to get the data from the /etc/leapp/files
+    2. We try to get the data from the only web service
+    3. Fallback mechanism, which provides the data from the data, stored inside
```
fernflower, line 98: If there is no fallback anymore - the comment doesn't make sense

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/actor.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+class RestrictedPCIsScanner(Actor):
+    """
+    Provides data about restricted (unsupported or/and unavailable) devices.
+
+    The data sources priority set up is as following:
+    1. The actor tries to get the data from /etc/leapp/files
+    2. Then from the online service
+    3. As a fallback mechanism the data comes from the locally stored
```
fernflower, line 14: Same here - no fallback anymore, please update comment

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -0,0 +1,137 @@
...
+)
+
+# TODO Once the microservice for providing the data about removed hardware
+#   will be provisioned to the Red Hat costumers, then this variable default
+#   value should be overwritten
+PCIS_HOST_DEFAULT = ""
+API_VERSION = "v1"
+
+
+def get_restricted_pcis(http, update_fallback_data=False):
```
fernflower, line 34: no more fallback, the flag doesn't make sense anymore

fernflower (changes requested):
LGTM once you fix the remaining fallback logic in comments and tests

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
ZhukovGreen: resolved in https://github.com/oamg/leapp-repository/pull/602/commits/06e630c346b953bbe1dc9d26628be33ae194abfa

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/actor.py
ZhukovGreen: resolved in https://github.com/oamg/leapp-repository/pull/602/commits/06e630c346b953bbe1dc9d26628be33ae194abfa

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
ZhukovGreen: resolved in https://github.com/oamg/leapp-repository/pull/602/commits/06e630c346b953bbe1dc9d26628be33ae194abfa

fernflower (approved):
lgtm, let's wait for @pirat89 or @vinzenz

ZhukovGreen:
Hurray!!!! please let me squash fixups:)

ZhukovGreen:
So it is ready for @oamg/qe !

pirat89:
I do not see presence of any problems I reported. So keeping that on @fernflower and @vinzenz . No blocker from me.

vinzenz (approved):

ZhukovGreen:
FYI commits from 
ZhukovGreen added 4 commits on Oct 27 
is just rebasing to the recent master

ZhukovGreen:
@oamg/developers can't figure out what's wrong with the continuous-integration/oamg-ci/e2e tests

ZhukovGreen:
@pirat89 @Rezney @vinzenz @fernflower  I believe we can merge it. We managed all verification tests to PASS (thx to @mmacura311 and @shaded-enmity ).

pirat89:
@ZhukovGreen  we discussed that yesterday. We will merge it after the new release is created.

ZhukovGreen:
@pirat89 Thx Petr

fernflower:
@ZhukovGreen Can you please squash commits that make sense to be squashed? Like those that don't have description\bug fix ones probably?

ZhukovGreen:
rebased, squashed, thx @fernflower

pirat89:
@ZhukovGreen new rebase is needed.

pirat89:
leapp-ci build

pirat89:
This has been merged as part of #654 . closing

PR #604, opened 2020-10-09
Drop events with identical initial releases and input packages

drehak:
Some package transitions have to be represented via events that have the same initial release and the same set of input packages. In such a conflict, keep the event with the highest target release or ID and drop the rest.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/peseventsscanner.py
+++ b/repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/peseventsscanner.py
@@ -1,6 +1,6 @@
 import json
 import os
-from collections import namedtuple
+from collections import namedtuple, defaultdict
 from enum import IntEnum
 
 from leapp.exceptions import StopActorExecution, StopActorExecutionError
@@ -12,7 +12,8 @@ from leapp.models import (InstalledRedHatSignedRPM, PESRpmTransactionTasks, Repo
                           RepositoriesSetupTasks, RpmTransactionTasks, RepositoriesBlacklisted)
 
 
-Event = namedtuple('Event', ['action',        # An instance of Action
+Event = namedtuple('Event', ['id',            # int
+                             'action',        # An instance of Action
                              'in_pkgs',       # A dictionary with packages in format {<pkg_name>: <repository>}
                              'out_pkgs',      # A dictionary with packages in format {<pkg_name>: <repository>}
                              'from_release',  # A tuple representing a release in format (major, minor)
@@ -55,6 +56,7 @@ def pes_events_scanner(pes_json_filepath):
     arch_events = filter_events_by_architecture(filtered_events, arch)
 
     add_output_pkgs_to_transaction_conf(transaction_configuration, arch_events)
+    drop_conflicting_release_events(arch_events)
     tasks = process_events(filtered_releases, arch_events, installed_pkgs)
     filter_out_transaction_conf_pkgs(tasks, transaction_configuration)
     produce_messages(tasks)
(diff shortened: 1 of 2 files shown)


Rezney:
```python
Process Process-236:
    self.run()
  File "/usr/lib/python2.7/site-packages/leapp/actors/__init__.py", line 335, in run
    self.process(*args)
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/peseventsscanner/actor.py", line 29, in process
    pes_events_scanner('/etc/leapp/files/pes-events.json')
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/peseventsscanner/libraries/peseventsscanner.py", line 59, in pes_events_scanner
    tasks = process_events(filtered_releases, arch_events, installed_pkgs)
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/peseventsscanner/libraries/peseventsscanner.py", line 338, in process_events
    drop_conflicting_release_events(release_events)
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/peseventsscanner/libraries/peseventsscanner.py", line 313, in drop_conflicting_release_events
    input_events.sort(key=lambda e: (e.to_release, e.event_id))
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/peseventsscanner/libraries/peseventsscanner.py", line 313, in <lambda>
    input_events.sort(key=lambda e: (e.to_release, e.event_id))
AttributeError: 'Event' object has no attribute 'event_id'

```

drehak:
Added unit tests, flake8 + pytest pass. Now only to test it on a real system.

repos/system_upgrade/el7toel8/actors/peseventsscanner/libraries/peseventsscanner.py
```diff
@@ -285,6 +288,36 @@ def add_packages_to_tasks(tasks, packages, task_type):
...
+    return ','.join('{}:{}'.format(p[0], p[1]) for p in sorted(packages.items()))
+
+
+def drop_conflicting_release_events(events):
+    """
+    In case of events with identical initial release and input packages, drop those with older target releases.
+
+    In case of identical target releases too, drop events with lower IDs.
+    """
+    events_by_input = defaultdict(list)  # {input packages: [event]}
```
pirat89, line 307: @drehak wow! That's cool :)

pirat89 (commented):
@drehak No time to go through the whole actor, so I ask. What in case of upgrade from lower-version release? Right now, in case of RHEL-ALT, I want to do upgrade from RHEL 7.6.  what happens when I have installed package A, which contains duplicate entries for RHEL 7.7, 7.8, ... in this case, obviously I want to use the RHEL 7.7 one, not the newest one. What's the behaviour in the scope of the whole actor?

drehak:
> @drehak No time to go through the whole actor, so I ask. What in a case of an upgrade from lower-version release? Right now, in case of RHEL-ALT, I want to do upgrade from RHEL 7.6. what happens when I have installed package A, which contains duplicate entries for RHEL 7.7, 7.8, ... in this case, obviously I want to use the RHEL 7.7 one, not the newest one. What's the behaviour in the scope of the whole actor?

@pirat89 The behaviour is that all events with release higher than target get dropped before processing, see lines 54 and 55.

drehak:
It works!
Reproducer: If you install **quagga** and run preupgrade, the Removal event for 7.7 - >8.0 gets dropped, the Replacement event for 7.7 -> 8.1 is applied and **frr** gets correctly installed.

pirat89:
@drehak discussed in person. It seems the fix is correct, so far as the input data for `drop_conflicting_release_events` are pre-filtered in the `pes_events_scanner`. The possible issue I tried to pictured above is probably a bug, but not related to this PR. Just code review. I haven't tested the PR personally, yet.

drehak:
Yeah, it's in a great need of refactoring.

AloisMahdal:
I've run extended regression downstream (including  maximum installation) and all passed.

Ie. from QE POV this could get into the next release.  cc @Rezney

Rezney (approved):

PR #605, opened 2020-10-09
Fixing log storing after upgrade in initramdisk

vinzenz:
Previously, when running leapp upgrade during the initram disk, log
files weren't saved if /var/log was residing on a different partition.
This PR fixes this issue by running the saving process of the log file
in a container and mounting all partition before hand. The same approach
is used for running leapp, options to systemd-nspawn are shared through
a global environment variable.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
+++ b/repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
@@ -13,6 +13,11 @@ export LEAPP3_BIN=$LEAPPHOME/leapp3
 
 export NEWROOT=${NEWROOT:-"/sysroot"}
 
+NSPAWN_OPTS="--capability=all --bind=/sys --bind=/dev --bind=/dev/pts --bind=/run/systemd --bind=/proc"
+[ -d /dev/mapper ] NSPAWN_OPTS="$NSPAWN_OPTS --bind=/dev/mapper"
+export NSPAWN_OPTS="$NSPAWN_OPTS --bind=/run/udev --keep-unit --register=no --timezone=off --resolv-conf=off"
+
+
 do_upgrade() {
     local args="" rv=0
     #FIXME: set here params we would like to possible use...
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -78,13 +81,27 @@ save_journal() {
...
-    local logfile="$NEWROOT/var/log/leapp/leapp-upgrade.log"
+    local logfile="/sysroot/tmp-leapp-upgrade.log"
+
+    # Create logfile if it doesn't exist
+    [ -n $logfile ] && > $logfile
+
+    # If file exists save the journal
+    if [ -e $logfile ]; then
+        # Add a separator
+        echo "### LEAPP reboot ###" > $logfile
```
Rezney, line 92:
```suggestion
        echo "### LEAPP reboot ###" >> $logfile
```
Not sure if there is any chance for anything to be there yet...

vinzenz: not in this one, and I want it to be empty.

pirat89:
@vinzenz Could you remove the trailing whitespaces?

pirat89:
I realised I can push too, so it can be squashed on merge.

pirat89:
leapp-ci build

Rezney:
leapp-ci build

Rezney:
The "e2e" test fail was caused by a time out connecting back to the machine. Restarting.

Rezney:
Tested on Azure where the issue was originally reported and I can confirm with this patch we get the logs from initramfs phase and "leapp-upgrade.log: No such file or directory" error is gone. Thanks for the fix!

repos/system_upgrade/el7toel8/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -121,6 +139,9 @@ result=$?
 ##### safe the data and remount $NEWROOT as it was previously mounted #####
 save_journal
 
+#FIXME: for debugging purposes; this will be removed or redefined in future
```
Rezney, line 142: @vinzenz do you want to remove it in this PR or later?

vinzenz: It's the same comment on the other places, I just duplicated it as well - I want to leave it there.

Rezney (approved):
Verified...

PR #606, opened 2020-10-12
Provide also a hint to check if repomap file is up-to-date

Rezney:
In order to enable upgrades on RHUI we need an updated repomap file. It can happen users have an old one from previous upgrades and thus missing the needed mapping.


pirat89 (approved):

repos/system_upgrade/el7toel8/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -387,7 +387,7 @@ def gather_target_repositories(context, indata):
                 ' ensure the custom repository file is provided regarding the documentation with'
                 ' properly defined repositories or in case repositories are already defined'
                 ' in any repofiles under /etc/yum.repos.d/ directory, use the --enablerepo option'
-                ' for leapp'
+                ' for leapp. Also make sure "/etc/leapp/files/repomap.csv" file is up-to-date.'
```
pirat89, line 390:
```suggestion
                ' for leapp. Also make sure the "/etc/leapp/files/repomap.csv" file is up-to-date.'
```

PR #607, opened 2020-10-12
Enable "oddjobd" service

Rezney:
Enable "oddjobd" service after authselect convertion, otherwise mkhomedir does not work properly.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/authselectapply/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/authselectapply/actor.py
@@ -53,6 +53,11 @@ class AuthselectApply(Actor):
             ] + resources)  # pylint: disable-msg=too-many-arguments
             return
 
+        try:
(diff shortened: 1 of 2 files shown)


AloisMahdal:
Note that the verification job runs a test that is not 100% done yet and will fail because of a known test issue.  (We're checking for faillock tally migration but turns out that that can't work.)   I'll be keeping an eye on the job and report the correct result here.

AloisMahdal:
I've reviewed the **murphy-ci/ver** result and I can confirm that the oddjobd.service is running and mkhomedir works for all users.

(The failure is as expected due to test limitation which will be resolved before promoting the test.)

AloisMahdal:
So @Rezney  has confirmed the above force-push was just user-facing message text change so the verification applies.

Looks ready to merge..?

pirat89 (approved):
lgtm.

PR #609, opened 2020-10-14
packaging: Use replace __python macro by __python2

pirat89:
On F33+ the __python macro has become an error:
  https://fedoraproject.org/wiki/Changes/PythonMacroError
Use strictly just __python2 macro (for leapp-repository, we cannot
use __python3 unfortunately).


Rezney (approved):
Works for me...

PR #610, opened 2020-10-15
Drop nssglibccheck actor (re-enable winbind)

AloisMahdal:
This actor inhibits upgrade if winbind/wins is used in nsswitch.conf.

The actor was added because of issues in samba and glibc which -- under certain circumstances -- could cause upgrade to crash.

 *  Bug 1754575 - samba: Remove NSS wins and winbind dependency on librt/libpthread as workaround for glibc defect. (edit)
    https://bugzilla.redhat.com/show_bug.cgi?id=1754575

 *  Bug 1748197 - glibc: Reduce IFUNC resolver usage in libpthread and librt (edit)
    https://bugzilla.redhat.com/show_bug.cgi?id=1748197

In RHEL-8.2, samba packages have been updated to work around this issue so currently supported upgrade paths won't be affected, and therefore this actor is no longer needed.


drehak:
Sounds reasonable.

drehak (approved):

pirat89 (approved):

PR #614, opened 2020-11-02
Use more accurate tense in the SELinux messaging

AloisMahdal:
After `leapp preupgrade`, reviewing results will show this report:

    SElinux relabeling has been scheduled
    SElinux relabeling has been scheduled as the status was permissive/enforcing.

This looks like leapp has already scheduled the relabeling (ie. `touch /.autorelabel`) which is not accurate and would be illegal.  The proposal is to change the tense:

    SElinux relabeling will be scheduled
    SElinux relabeling will be scheduled as the status is permissive/enforcing.

(This is consistent with another messaging from the same actor.)


pirat89 (dismissed):

fernflower:
Could you please rebase against recent master? I'd like to have this merged.

fernflower (changes requested):
Please rebase.

fernflower (dismissed):

fernflower (approved):

PR #615, opened 2020-11-09
Do not inhibit upgrade with SCA enabled manifest

Rezney:
https://bugzilla.redhat.com/show_bug.cgi?id=1873312

### DIFF
--- a/repos/system_upgrade/el7toel8/models/rhsminfo.py
+++ b/repos/system_upgrade/el7toel8/models/rhsminfo.py
@@ -18,3 +18,5 @@ class RHSMInfo(Model):
     """ Repositories that are enabled on the current system through the subscription-manager. """
     existing_product_certificates = fields.List(fields.String(), default=[])
     """ Product certificates that are currently installed on the system. """
+    sca_detected = fields.Boolean(default=False)
+    """ Info about whether SCA manifest was used or not. """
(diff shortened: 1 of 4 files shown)


repos/system_upgrade/el7toel8/models/rhsminfo.py
```diff
@@ -18,3 +18,5 @@ class RHSMInfo(Model):
     """ Repositories that are enabled on the current system through the subscription-manager. """
     existing_product_certificates = fields.List(fields.String(), default=[])
     """ Product certificates that are currently installed on the system. """
+    status = fields.String(default='')
+    """ RHSM status output. """
```
pirat89, line 22: @mreznik wouldn't be better in this case just boolean field like `is_sca`? or it seems it could be even generic "mode" and haviing choices `CAM` and `SCA`. But mdoe could be confusing a little bit for someone, so `is_sca` / `sca_detected` / ... seems to be a better solution. As it is clearly providing info usable by other actors, without need of additional output_parsing in other actors in future (if any actor need it in future).

Rezney: @pirat89 changed...

Rezney:
Re-tested and seems to be working fine.

Rezney:
e2e tests

pirat89 (approved):
LGTM and works as expected. Waiting for verification by qe

pirat89:
Verified by RHSM QE.

PR #616, opened 2020-11-10
Docstring refactoring: use quotes instead of apostrophe

pirat89:
In case of docstring, quotes should be used always, officialy.
Replace apostrophes by quotes in case of docstrings.

Ref: https://www.python.org/dev/peps/pep-0257


drehak:
Sounds reasonable.
(See: https://www.python.org/dev/peps/pep-0257)

drehak (approved):

PR #617, opened 2020-11-11
Added "-y" option to yum remove ipa-server in actor

calvingsmith:
Automated remediation fails as yum remove is expecting input. Added the "-y" option to fix


vinzenz (approved):

vinzenz:
@calvingsmith Thanks a lot for your contribution

PR #620, opened 2020-11-12
Use Groups instead of report Flags and report Tags

fernflower:


Makefile
```diff
@@ -53,6 +53,11 @@ TIMESTAMP:=$${__TIMESTAMP:-$(shell /bin/date -u "+%Y%m%d%H%MZ")}
 SHORT_SHA=`git rev-parse --short HEAD`
 BRANCH=`git rev-parse --abbrev-ref HEAD | tr -- '-/' '_'`
 
+# In case the top commit Depends-On some framework patch that is not merged yet - override framework master with the
+# proper version
+REQ_LEAPP_PR=$(shell git show HEAD | grep Depends-On | grep -Eo '[[:digit:]]*')
```
vinzenz, line 58:
```suggestion
REQ_LEAPP_PR=`git show HEAD | grep Depends-On | grep -Eo '[[:digit:]]*'`
```
Try this

fernflower: The backticks fail to properly resolve in the if condition. I tried it in various flavors, even adding direct shell invocation with `$(shell )` and could not fix that. 
```
if [[ ! -z "`git show HEAD | grep Depends-On | grep -Eo '[[:digit:]]*'`" ]] && [ `curl "https://api.github.com/repos/oamg/leapp/pulls/`git show HEAD | grep Depends-On | grep -Eo '[[:digit:]]*'`" | python -c "import json,sys; print (json.load(sys.stdin)['closed_at'] is None)"` = 'True' ]; then \
	echo "Leapp-repository depends on the yet unmerged pr of the framework #`git show HEAD | grep Depends-On | grep -Eo '[[:digit:]]*'`, installing it.." && \
	pip install "git+https://github.com/oamg/leapp.git@refs/pull/`git show HEAD | grep Depends-On | grep -Eo '[[:digit:]]*'`/head"; \
fi
/bin/sh: command substitution: line 0: unexpected EOF while looking for matching `"'
/bin/sh: command substitution: line 1: syntax error: unexpected end of file
/bin/sh: command substitution: line 0: syntax error near unexpected token `json.load'
/bin/sh: command substitution: line 0: `" | python -c "import json,sys; print (json.load(sys.stdin)['closed_at'] is None)"'
```

Makefile
```diff
@@ -53,6 +53,11 @@ TIMESTAMP:=$${__TIMESTAMP:-$(shell /bin/date -u "+%Y%m%d%H%MZ")}
 SHORT_SHA=`git rev-parse --short HEAD`
 BRANCH=`git rev-parse --abbrev-ref HEAD | tr -- '-/' '_'`
 
+# In case the top commit Depends-On some framework patch that is not merged yet - override framework master with the
+# proper version
+REQ_LEAPP_PR:=$(shell git show HEAD | grep Depends-On | grep -Eo '[[:digit:]]*')
+REQ_LEAPP_PR_IS_OPEN:=$(shell curl "https://api.github.com/repos/oamg/leapp/pulls/${REQ_LEAPP_PR}" | python -c "import json,sys; print (json.load(sys.stdin)['closed_at'] is None)")
```
vinzenz, line 59:
```suggestion
REQ_LEAPP_PR=$(shell git show HEAD | grep Depends-On | grep -Eo '[[:digit:]]*')
REQ_LEAPP_PR_IS_OPEN=$(shell curl "https://api.github.com/repos/oamg/leapp/pulls/${REQ_LEAPP_PR}" | python -c "import json,sys; print (json.load(sys.stdin)['closed_at'] is None)")
```

fernflower: As I said, I have tried with both Makefile var flavors - I am getting the failure in travis (apparently REQ_LEAPP_PR is empty and REQ_LEAPP_PR_IS_OPEN is badly processed because of that) which I can't reproduce locally if I run the build an run commands directly. 
Still trying though

fernflower:
recheck

PR #621, opened 2020-11-14
A humble attempt to respect framework dependencies.

fernflower:
If the top commit in a branch has a Depends-On comment,
then the corresponding PR of the framework is checked whether
or not it has been already merged, and in case it's still open
the proper version of leapp will be installed into the venv
as part of the install-deps-* target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,6 +53,13 @@ TIMESTAMP:=$${__TIMESTAMP:-$(shell /bin/date -u "+%Y%m%d%H%MZ")}
 SHORT_SHA=`git rev-parse --short HEAD`
 BRANCH=`git rev-parse --abbrev-ref HEAD | tr -- '-/' '_'`
 
+# The dependent framework PR connection will be taken from the top commit's depends-on message.
+REQ_LEAPP_PR=$(shell git log master..HEAD | grep -m1 -iE '^[[:space:]]*Depends-On:[[:space:]]*.*[[:digit:]]+[[:space:]]*$$' | grep -Eo '*[[:digit:]]*')
+# NOTE(ivasilev) In case of travis relying on top commit is a no go as a top commit will be a merge commit.
+ifdef CI
+	REQ_LEAPP_PR=$(shell git log master..HEAD | grep -m1 -iE '^[[:space:]]*Depends-On:[[:space:]]*.*[[:digit:]]+[[:space:]]*$$' | grep -Eo '[[:digit:]]*')
+endif
+
 # In case anyone would like to add any other suffix, just make it possible
 _SUFFIX=`if test -n "$$SUFFIX"; then echo ".$${SUFFIX}"; fi; `
 
@@ -203,8 +210,14 @@ install-deps:
 	case $(_PYTHON_VENV) in python3*) yum install gcc -y; esac
 	virtualenv --system-site-packages -p /usr/bin/$(_PYTHON_VENV) $(VENVNAME); \
 	. $(VENVNAME)/bin/activate; \
+	pip install -U pip; \
 	pip install --upgrade setuptools; \
-	pip install --upgrade -r requirements.txt
+	pip install --upgrade -r requirements.txt \
+	# In case the top commit Depends-On some yet unmerged framework patch - override master leapp with the proper version
+	if [[ ! -z "$(REQ_LEAPP_PR)" ]] ; then \
+		echo "Leapp-repository depends on the yet unmerged pr of the framework #$(REQ_LEAPP_PR), installing it.." && \
+		$(VENVNAME)/bin/pip install -U "git+https://github.com/oamg/leapp.git@refs/pull/$(REQ_LEAPP_PR)/head"; \
+	fi
 	python utils/install_actor_deps.py --actor=$(ACTOR)
 
 install-deps-fedora:
@@ -219,8 +232,14 @@ install-deps-fedora:
 	@# Prepare the virtual environment
 	virtualenv --system-site-packages --python /usr/bin/python $(VENVNAME)
 	. $(VENVNAME)/bin/activate ; \
+	pip install -U pip; \
 	pip install --upgrade setuptools; \
 	pip install --upgrade -r requirements.txt; \
+	# In case the top commit Depends-On some yet unmerged framework patch - override master leapp with the proper version
+	if [[ ! -z "$(REQ_LEAPP_PR)" ]] ; then \
+		echo "Leapp-repository depends on the yet unmerged pr of the framework #$(REQ_LEAPP_PR), installing it.." && \
+		$(VENVNAME)/bin/pip install -U "git+https://github.com/oamg/leapp.git@refs/pull/$(REQ_LEAPP_PR)/head"; \
+	fi
 
 lint:
 	. $(VENVNAME)/bin/activate; \
--- a/utils/docker-tests/Dockerfile
+++ b/utils/docker-tests/Dockerfile
@@ -5,5 +5,11 @@ VOLUME /payload
 RUN yum update -y && \
     yum install python-virtualenv make git -y
 
+# NOTE(ivasilev) Unless we upgrade pip an obsolete v 9 will be on the system
+# and we need at least 10.0.1 to install a not-yet-merged framework pr in case
+# it's necessary
+RUN curl "https://bootstrap.pypa.io/get-pip.py" -o "get-pip.py" && \
+    python get-pip.py
+
 WORKDIR /payload
 ENTRYPOINT make install-deps && make test
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,13 +16,13 @@ install:
 jobs:
   include:
     - script:
-        - docker run --rm -ti -v ${PWD}:/payload --env=PYTHON_VENV leapp-tests
+        - docker run --env CI=$CI --rm -ti -v ${PWD}:/payload --env=PYTHON_VENV leapp-tests
       env: PYTHON_VENV=python2.7
     - script:
-        - docker run --rm -ti -v ${PWD}:/payload --env=PYTHON_VENV leapp-tests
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -53,6 +53,11 @@ TIMESTAMP:=$${__TIMESTAMP:-$(shell /bin/date -u "+%Y%m%d%H%MZ")}
 SHORT_SHA=`git rev-parse --short HEAD`
 BRANCH=`git rev-parse --abbrev-ref HEAD | tr -- '-/' '_'`
 
+# In case the top commit Depends-On some framework patch that is not merged yet - override framework master with the
+# proper version
+DEPENDS_ON_PR=$(shell git show HEAD -q | grep Depends-On | grep -Eo '[[:digit:]]*')
+DEPENDS_ON_PR_IS_OPEN=$(shell curl "https://api.github.com/repos/oamg/leapp/pulls/${DEPENDS_ON_PR}" | python -c "import json,sys; print (json.load(sys.stdin)['closed_at'] is None)")
```
pirat89, line 59: @fernflower just quick view. I think these vars should have more specific name. like: REQ_LEAPP_PR - depends on PR is not intuitive. As well, do not use please `$(...)` syntax in Makefile for subshell. It could be missleading for people that are not familiar with Makefile. Use backticks instead.

fernflower: Hmm @pirat89 but backticks don't work for my specific use case - what I need is not an as-is command substitution behavior the backticks give me, but the result of the shell command, and afaik this can only be achieved by a subshell command execution the way I have it now. 
Besides we already have shell command invocation in Makefile, so I would not call it me bending the rules https://github.com/oamg/leapp-repository/blob/master/Makefile#L52

Could you give me a working example with backticks for my specific case?

vojtechsokol: I'm OK with using `$(...)` in the Makefile:
1. Argument that people unfamiliar with Makefile syntax is void - people unfamiliar with `${SUBSTITUTE_ANYTHING}` syntax should learn it first and then they can work with `${SUBSTITUTE_ANYTHING}`
2. The syntax is aleady used in multiple places in this Makefile, the same holds for `leapp`'s Makefile
3. I'm hugely biased against backticks, they should have died really long ago, but that's no argument...

Makefile
```diff
@@ -53,6 +53,11 @@ TIMESTAMP:=$${__TIMESTAMP:-$(shell /bin/date -u "+%Y%m%d%H%MZ")}
 SHORT_SHA=`git rev-parse --short HEAD`
 BRANCH=`git rev-parse --abbrev-ref HEAD | tr -- '-/' '_'`
 
+# In case the top commit Depends-On some framework patch that is not merged yet - override framework master with the
+# proper version
+REQ_LEAPP_PR=$(shell git show HEAD -q | grep Depends-On | grep -Eo '[[:digit:]]*')
```
vojtechsokol, line 58: The check for `Depends-On` keyword only in top commit of the PR is somewhat cumbersome, what if the PR author:
1. needs to add another commit? He has to add this keyword to every commit message.
2. forgets to add this keyword to the last comment? He has to add another commit or rebase with modified commit message.
3. wnats to change (or even remove) the PR dependency? Same as above

It seems to me that better approach would be querying API and checking PR description for this keyword. This way the PR author (or even repo owner) can easily add/remove the dependecy and also the dependency has better visibility. If implemented this way I would also prefer to have this dependency on its own line. e.g. checking something like `^Depends-On: [[:digit:]]*$`

utils/docker-tests/Dockerfile
```diff
@@ -5,5 +5,9 @@ VOLUME /payload
 RUN yum update -y && \
     yum install python-virtualenv make git -y
 
+# XXX FIXME couldn't figure out a proper way to get pip v10+ on the system
```
vojtechsokol, line 8: Is pip v10+ really needed? If yes, I wouldn't mark it as a `FIXME`, if installation by a package manager is not a viable option, installation using `get-pip.py` script is pretty standard.

vinzenz: I am just wondering why easy_install can't get the job done

fernflower: @vojtechsokol Yeah, I need at least 10.0.1 to get the job done (specifically I am interested in https://github.com/pypa/pip/commit/737dfbfc71490b41ab92c6d49b0c303c0e197f46)
@vinzenz you mean I should install pip with easy_install?

Makefile
fernflower: 1. That's true, but tbh I see no real problem with this - if 2 or more commits are part of some big change saying clearly that they actually are related to it is common practice. 
2. As we work on feature branches any issue like this is resolved with `git commit -- amend`\`git rebase -i HEAD~2` and a force push. 
3. Same as 2 - just fix the commit message.

But I really liked your idea about github api and even tried to implement it. Unfortunately I had to revert to the old "check commit message for depends-on" behavior because of the following concerns:
- checking gihub api requires the code to be already pushed to the pr branch. Which makes it impossible to run local checks on committed-but-not-yet-pushed code. Running on not-committed code will succeed though.
- you still need to get the dependency information from somewhere. The easiest way I could think of was by adding a label like "requires-XXX", where XXX is number of repository PR. Though it has nicer visibility and any team member can do it, it requires a manual action which can be easily overlooked.
- github api has a certain limit for unauthorized requests, something like 60 per hour. This might not be an issue for travis (for authorized it is 5000 per hour), but if local runs will heavily use api there is a possibility this unauthorized requests limit can be exhausted. Well, I did hit it yesterday night after some really intensive debugging and testing.

I'd say having a combination of 2 (check api for travis + depend on commit message for local run) is a nice follow up, but I'd like to have the initial version simple and working.

fernflower: Thanks for the regex suggestion, fixed https://github.com/oamg/leapp-repository/pull/621/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R57

Makefile
```diff
@@ -203,8 +211,14 @@ install-deps:
 	case $(_PYTHON_VENV) in python3*) yum install gcc -y; esac
 	virtualenv --system-site-packages -p /usr/bin/$(_PYTHON_VENV) $(VENVNAME); \
 	. $(VENVNAME)/bin/activate; \
+	pip install -U pip; \
 	pip install --upgrade setuptools; \
-	pip install --upgrade -r requirements.txt
+	pip install --upgrade -r requirements.txt \
+	# In case the top commit Depends-On some yet unmerged framework patch - override master leapp with the proper version
+	if [[ ! -z "$(REQ_LEAPP_PR)" ]] && [ $(REQ_LEAPP_PR_IS_OPEN) = 'True' ]; then \
```
pirat89, line 218: I am thinking now whether the second part of the condition is correct. I haven't tested it but thinking about:

```suggestion
	if [[ ! -z "$(REQ_LEAPP_PR)" ]] && [[ "$(REQ_LEAPP_PR_IS_OPEN)" == "True" ]]; then \
```

Makefile
```diff
@@ -53,6 +53,14 @@ TIMESTAMP:=$${__TIMESTAMP:-$(shell /bin/date -u "+%Y%m%d%H%MZ")}
 SHORT_SHA=`git rev-parse --short HEAD`
 BRANCH=`git rev-parse --abbrev-ref HEAD | tr -- '-/' '_'`
 
+# The dependent framework PR connection will be taken from the top commit's depends-on message.
+REQ_LEAPP_PR=$(shell git log -n 1 HEAD | grep '^.*Depends-On:.*[[:digit:]]*$$' | grep -Eo '*[[:digit:]]*')
```
pirat89, line 57: That could hit even string like " 232  Depends-on: 333 PR. but takes 555... 555" - where the result will be multiple numbers:
```bash
echo  " 232  Depends-On: 333 PR. but takes 555... 555"  | grep '^.*Depends-On:.*[[:digit:]]*$' | grep -Eo '*[[:digit:]]*'
232
333
555
555
```

```suggestion
REQ_LEAPP_PR=$(shell git log -n 1 HEAD | grep -E '^[[:space:]]*Depends-On:[[:space:]]*[[:digit:]]+[[:space:]]*$$' | grep -Eo '*[[:digit:]]*')
```

Makefile
pirat89: @fernflower  I agree with @vojtechsokol. The need of the "Depends-On" in every commit keep it in the last one seems not good. Why are you checking just the last commit? Why not to check all commits different from master?

fernflower: Because checking last one is way easier than figuring out what commits are in this pr and implementing that separately for travis and local only run.
Again - that's a nice follow up but we have a blocker for proper unittest run right now.

Makefile
fernflower: @pirat89 this won't match the `Depends-On: https://direct/link/to/pr/XXX`.
Although I like the idea of making regex stricter, thanks, I guess `grep -m1 -iE '^[[:space:]]*Depends-On:[[:space:]]*.*[[:digit:]]+[[:space:]]*$$'` should be ok for both cases (url and just the number)

Makefile
fernflower: Yeah, you are right, will fix in next commit. Thanks a lot!

Makefile
fernflower: Should be fixed now

Makefile
fernflower: Should be fixed now.

Makefile
fernflower: @pirat89 implemented as you suggested (checking all commits between HEAD..master for Depends-On) and looks like it works https://travis-ci.org/github/oamg/leapp-repository/jobs/747420365 :)

fernflower:
e2e tests

fernflower:
e2e tests

vinzenz:
leapp-ci build

vinzenz (approved):

PR #622, opened 2020-11-23
If msgs is None, make sure mocked consume works

ZhukovGreen:
The real api.consume returns empty iterator when
no messages were fed. However in the mocked api.consume
version, this leads to an error.

This is fixed now

### DIFF
--- a/repos/system_upgrade/el7toel8/libraries/testutils.py
+++ b/repos/system_upgrade/el7toel8/libraries/testutils.py
@@ -71,7 +71,7 @@ class CurrentActorMocked(object):  # pylint:disable=R0904
         self.configuration = namedtuple(
             'configuration', ['architecture', 'kernel', 'leapp_env_vars', 'os_release', 'version']
         )(arch, kernel, envarsList, release, version)
-        self._msgs = msgs
+        self._msgs = msgs or []
 
     def __call__(self):
         return self


ZhukovGreen:
@oamg/developers pls review

repos/system_upgrade/el7toel8/libraries/testutils.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, arch=architecture.ARCH_X86_64, envars=None, kernel='3.10.0-95
         self.configuration = namedtuple(
             'configuration', ['architecture', 'kernel', 'leapp_env_vars', 'os_release', 'version']
         )(arch, kernel, envarsList, release, version)
-        self._msgs = msgs
+        self._msgs = msgs if msgs else []
```
vinzenz, line 74:
```suggestion
        self._msgs = msgs or []
```

vinzenz: alternatively you could probably set msgs=() in the constructor

ZhukovGreen: Thx!

ZhukovGreen:
e2e tests

ZhukovGreen:
leapp-ci build

pirat89 (approved):

PR #625, opened 2020-12-01
persistentnetneamesconfig: crash on nonexistent iface

pirat89:
We have to deal with situation when:
      - we detect a network device on src system
      - we cannot detect the network device inside the upgrade env
    
    This happens very easily, when the nerwork card need specific driver
    which is by default not included in the initramfs. As we are not
    able to deal with any possible driver (especially from the other
    vendors), let's rather ensure the actor doesn't crash in such a case
    and rather prints a warning message(s).


pirat89:
The current issue is hit by these two devices:
```
            {                                                                   
                "devpath": "/devices/pci0000:80/0000:80:01.0/0000:81:00.0/net/ib0",
                "driver": "mlx5_core[ib_ipoib]",                                   
                "mac": "00:00:10:86:fe:80:00:00:00:00:00:00:98:03:9b:03:00:9d:6f:08",
                "name": "ib0",                                                     
                "pci_info": {                                                   
                    "bus": "81",                                                
                    "device": "00",                                             
                    "domain": "0000",                                           
                    "function": "0"                                             
                },                                                              
                "vendor": "0x15b3"                                              
            },                                                                  
               
            {                                                                   
                "devpath": "/devices/pci0000:80/0000:80:01.0/0000:81:00.1/net/ib1",
                "driver": "mlx5_core[ib_ipoib]",                                
                "mac": "00:00:18:86:fe:80:00:00:00:00:00:00:98:03:9b:03:00:9d:6f:09",
                "name": "ib1",                                                  
                "pci_info": {                                                      
                    "bus": "81",                                                
                    "device": "00",                                                
                    "domain": "0000",                                              
                    "function": "1"                                                
                },                                                                 
                "vendor": "0x15b3"                                              
            }
```

which are present on RHEL 7 but cannot be detected inside the upgrade initramfs. I guess it's generic problem as possibly this could be hit by any network device that requires specific drivers.

pirat89:
Currently I have not been able to discover `mlx5_ib` module inside the upgrade initramfs. In case of target RHEL 8 initramfs, I can see much more new modules / files around *infiniband* inside the intiramfs in comparison to RHEL 7 initramfs. Maybe it's blind alley but I am curious why these files are not inside the upgrade initramfs. Possibly we are missing some config files when creating the initramfs?

pirat89:
@oamg/developers  this PR is just for the generic part of the problem - ensure the actor handling the net interfaces names is not crashing when a device cannot be detected inside the upgrad initramfs. Handling of infiniband card will be solved in separate PR. but currently this should be already enough for now.

repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/tests/test_persistentnetnamesconfig.py
```diff
@@ -114,3 +137,32 @@ def test_renamed_swap_eth(monkeypatch, current_actor_context):
...
+    ]
+    monkeypatch.setattr(persistentnetnamesconfig, 'generate_link_file', generate_link_file_mocked)
+    monkeypatch.setattr(persistentnetnamesconfig.api, 'current_actor', CurrentActorMocked(msgs=msgs))
+    monkeypatch.setattr(persistentnetnamesconfig.api, 'current_logger', logger_mocked())
+    monkeypatch.setattr(persistentnetnamesconfig.api, 'produce', produce_mocked())
+    persistentnetnamesconfig.process()
+
+    for prod_models in [RenamedInterfaces, InitrdIncludes]:
+        any(isinstance(i, prod_models) for i in persistentnetnamesconfig.api.produce.model_instances)
+    assert any(['Some network devices' in x for x in persistentnetnamesconfig.api.current_logger.warnmsg])
```
fernflower, line 168: Just wondering if logging a warning is enough in that case. Maybe we should report it as well?

vinzenz (dismissed):
LGTM

repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/libraries/persistentnetnamesconfig.py
```diff
@@ -23,3 +33,57 @@ def generate_link_file(interface):
...
+
+                initrd_files.append(generate_link_file(iface))
+
+    if missing_ifaces:
+        msg = (
+            'Some network devices have not been detected inside the'
+            ' upgrade environment and so related network interfaces'
+            ' could be renamed on the upgraded system.'
+        )
+        # Note:
```
fernflower, line 81: nit: I'd honestly prefere NOTE(pstodulk) as this saves a git blame call

repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/tests/test_persistentnetnamesconfig.py
vinzenz: I was thinking the same thing, but we should fix it first. We can make a follow up on that. We need to decide on what to tell the user what they should do in that case.

pirat89: I guess it's enough now regarding the rest of actor (where the warning is used only as well in cases we cannot handle). But in future I think we should generate post-upgrade reports as well.

repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/libraries/persistentnetnamesconfig.py
vinzenz: To me it doesn't really matter, who made the note. It helps to understand the code.

repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/tests/test_persistentnetnamesconfig.py
```diff
@@ -1,7 +1,30 @@
...
+    PersistentNetNamesFactsInitramfs,
+    RenamedInterface,
+    RenamedInterfaces,
+)
+
+TEST_DIR = os.path.dirname(os.path.abspath(__file__))
+CUR_DIR = ""
+
+
+@pytest.fixture
```
fernflower, line 22: Hmm sorry if I got it wrong, are you using adjust_cwd mainly for figuring out full path to proper crashed_interfaces.json?
In that case you could theoretically use PYTEST_CURRENT_TEST https://stackoverflow.com/questions/55913423/pytest-get-the-file-path-of-current-test-file

repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/libraries/persistentnetnamesconfig.py
pirat89: fixed :)

repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/tests/test_persistentnetnamesconfig.py
fernflower: Sure, I have no problem with that. Just please leave smth like `TODO add proper reporting later`

repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/tests/test_persistentnetnamesconfig.py
pirat89: I am using that as it was told earlier we should use it. Maybe I missed something? (@ZhukovGreen)
Anyway, as far as it works now, let's keep it.

repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/tests/test_persistentnetnamesconfig.py
pirat89: the TODO is added.

fernflower (approved):
code lgtm. 
Didn't have chance to properly test that in action though, so not merging.

repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/libraries/persistentnetnamesconfig.py
```diff
@@ -23,3 +33,60 @@ def generate_link_file(interface):
...
+
+                initrd_files.append(generate_link_file(iface))
+
+    if missing_ifaces:
+        msg = (
+            'Some network devices have not been detected inside the'
+            ' upgrade environment and so related network interfaces'
+            ' could be renamed on the upgraded system.'
+        )
+        # Note(pstodulk):
```
msekletar, line 81: I think this could be fixed. Basically we just need to have an actor that checks for presence of kernel drivers in upgrade initramfs after the image is generated. We either have the driver in the image, or RHEL kernel config (for next major version) says it is built-in or it is not present for some reason and in such case we can warn the user and abort the upgrade.

msekletar (approved):
LGTM (modulo minor suggestions, nothing that should block the merge though)

@pirat89 thank you for working on this!

pirat89:
leapp-ci build

repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/libraries/persistentnetnamesconfig.py
pirat89: Just for the record: discussed in person. basically I agree, but we are going to keep it as it is because of finished testing. We will improve it later once we update the actor to create the report.

examon:
leapp-ci build

pirat89:
leapp-ci build

PR #626, opened 2020-12-03
to_remove: Add initial-setup

vinzenz:
Previously if the initial-setup package was installed on a system, the
upgrade caused that the user was asked to accept the EULA and other
questions, during the first boot into the upgraded system. This causes
the system to stop the startup and it gets stuck on this point unless
someone interactively answers these questions.

This patch adds the initial-setup package to the list of rpms to be
removed during the upgrade transaction to ensure that this is no longer
happening.

Issue: OAMG-1531


examon:
e2e tests

vojtechsokol:
e2e tests

Rezney (approved):

examon:
leapp-ci build

pirat89:
leapp-ci build

PR #627, opened 2020-12-07
Check if grub.cfg is in /boot/efi

holser:
There are cases when /boot/efi is mounted but grub.cfg is not there as
EFI is not used. In that case leapp will fail with the following error

Details: Command ['/usr/sbin/grubby', '--remove-kernel',
'/boot/vmlinuz-upgrade.x86_64', '-c', '/boot/efi/EFI/redhat/grub.cfg']
failed with exit code 1.: error opening /boot/efi/EFI/redhat/grub.cfg
for read: No such file or directory

This patch adds file check to condition to allow leapp to skip such
cases

BZ-Url: https://bugzilla.redhat.com/show_bug.cgi?id=1905247

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/addupgradebootentry/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/addupgradebootentry/actor.py
@@ -34,6 +34,6 @@ class AddUpgradeBootEntry(Actor):
 
         # related to issue with hybrid BIOS and UEFI images
         # https://bugzilla.redhat.com/show_bug.cgi?id=1667028
-        if ff.firmware == 'bios' and os.path.ismount('/boot/efi'):
+        if ff.firmware == 'bios' and os.path.ismount('/boot/efi') and os.path.isfile('/boot/efi/EFI/redhat/grub.cfg'):
             configs = ['/boot/grub2/grub.cfg', '/boot/efi/EFI/redhat/grub.cfg']
         add_boot_entry(configs)


pirat89:
@holser Waiting for the bootloader team, as this doesn't seem as a valid configuration. The proposed patch could create additional bugs right now.

jngrosso:
Can some explain why this does not seem like a Valid Config. I am looking for scenarios to test...We have customers on OSP 10 and 13 with older hardware and they may prefer the legacy BIOS choice over EFI . I was also looking for a list of supported boot configurations. My tests was RHEL 7.9 to RHEL 8,2.  Is there a way to escalate this we  have a maintenance release coming and we would like to put this leapp fix in if we can have the bootloader team approve this PR or let us know why it should not be added.

examon:
leapp-ci build

pirat89:
@jngrosso because it's like 5th hackish solution how to detect BIOS/EFI and work with bootloader. Maybe the solution works for OSP, but it's not told that with that we do not break something else (e.g. hybrid BIOS/EFI bootloding...) and we do not want to add something blindly just now - we are close to release too and already have troubles to deliver what we want regarding missing people - especially QE. You can see how many PRs are waiting for testing... So we want to be sure that bootloader team is ok with that first.

pirat89 (approved):
So discussed with bootloader team (@jhlavac) it should be ok. It's suggested to do additional changes (and we discovered possible issue with `removeupgradebootentry`) but let's keep another changes for later. This change is unrelated to that.

Moving to testing.

mkluson:
e2e tests

examon:
leapp-ci build

pirat89:
leapp-ci build

pirat89:
Verified by @jngrosso as working

AloisMahdal:
I've checked the *murphy-ci/integration* and the only failures are because of 2 broken machines -- ie not even installed rhel-7.9.  That said, 26/28 tests did run and pass.

PR #629, opened 2021-01-09
Makefile: Fix leapp with proper version install in case of py3

fernflower:
pip3 --upgrade behaviour was defferent from pip2 --upgrade which
caused the leapp version specified by REQ_LEAPP_PR not to be prefered
over leapp master.
This patch changes --upgrade option to --ignore-installed which forcefully
updates leapp to REQ_LEAPP_PR.
Also install-deps-fedora was taught to respect PYTHON_VENV option. The
unability to catch pip3 weird behavior was partly because no true python3
env was created with PYTHON_VENV=python3.6 make install-fedora-deps.


vinzenz (approved):

pirat89 (approved):
lgtm and do not see any issue.

Issue #630, opened 2021-01-11
SR-IOV Virtual Function (VF) devices should not be considered as persistent device


krsacme:
**Actual behavior**
Baremetal Node hangs during upgrade

**To Reproduce**
Steps to reproduce the behavior

1. install RHEL 7.8
2. setup SR-IOV VF devices
3. run leapp upgrade

**Expected behavior**
Upgrade should be successful to RHEL 8.2

- OS and version: 
RHEL 7.8

Attach (or provide link to) log files if applicable (optional - may contain confidential information):
****
![Screenshot from 2020-08-06 20-27-30](https://user-images.githubusercontent.com/6107458/104150809-222ce480-5401-11eb-92b7-0c9ceb03b460.png)

[leapp.db.tar.gz](https://github.com/oamg/leapp-repository/files/5794248/leapp.db.tar.gz)

**Additional context**

krsacme:
I am NOT able to reproduce the issue with the recent version of leapp
​packages, even without this PR changes.

While trying to reproduce the issue in standalone setup, with RHEL7.9,
​the issue of hang does not occur on upgrade. It was reproducible in
​RHEL7.8, refer to the analysis in the BZ [1]. The leapp package used in
​the reproducible version was v0.10.0. 

I see there are lot of changes in the file [2] in which the Exception
​was observed during reboot. I am not able to trace which change has
​fixed this issue but, the issue is not reproducible anymore with latest
​repo. This means that we can close this PR and remove the workaround in
​the TripleO to reset the VFs.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1866372#c5
[2] https://github.com/oamg/leapp-repository/blob/v0.10.0/repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/actor.py#L50

PR #631, opened 2021-01-11
Exclude SR-IOV VF devices from physical devices

krsacme:
Fixes: https://github.com/oamg/leapp-repository/issues/630

Exclude SR-IOV VF devices from the persistent list as it has to be created on every reboot.

Relate BZ - [#1866372](https://bugzilla.redhat.com/show_bug.cgi?id=1866372#c5)

### DIFF
--- a/repos/system_upgrade/el7toel8/libraries/persistentnetnames.py
+++ b/repos/system_upgrade/el7toel8/libraries/persistentnetnames.py
@@ -1,3 +1,4 @@
+import os
 import pyudev
 
 from leapp.models import PCIAddress, Interface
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/el7toel8/libraries/persistentnetnames.py
```diff
@@ -1,3 +1,4 @@
+import os
 import pyudev
```
vinzenz, line 2:
```suggestion
import os

import pyudev
```

krsacme:
@vinzenz I could not make sense out of the ci error. Can you point out the actual errors on CI?

pirat89:
@krsacme I will check. just in the meanwhile, I will re-run the tests as our CI is broken oftenly..

pirat89:
leapp-ci build

pirat89:
leapp-ci build

pirat89:
leapp-ci build

krsacme:
I am NOT able to reproduce the issue with the recent version of leapp
​packages, even without this PR changes.

While trying to reproduce the issue in standalone setup, with RHEL7.9,
​the issue of hang does not occur on upgrade. It was reproducible in
​RHEL7.8, refer to the analysis in the BZ [1]. The leapp package used in
​the reproducible version was v0.10.0. 

I see there are lot of changes in the file [2] in which the Exception
​was observed during reboot. I am not able to trace which change has
​fixed this issue but, the issue is not reproducible anymore with latest
​repo. This means that we can close this PR and remove the workaround in
​the TripleO to reset the VFs.


[1] https://bugzilla.redhat.com/show_bug.cgi?id=1866372#c5
[2] https://github.com/oamg/leapp-repository/blob/v0.10.0/repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/actor.py#L50

PR #632, opened 2021-01-11
storagescanner: Inhibit if any CIFS entries found in /etc/fstab

fernflower:
Due to networking not available in initramfs phase CIFS upgrade attempt
with CIFS entries in the /etc/fstab has to be inhibited for now.
Also changed a behavior of error reporting in _get_fstab_info
function, instead of find-first-error-and-report-it now all errors
of the same type will be reported under a single report entry.

OAMG-4107

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkcifs/tests/test_checkcifs.py
+++ b/repos/system_upgrade/el7toel8/actors/checkcifs/tests/test_checkcifs.py
@@ -1,0 +1,34 @@
+from leapp.snactor.fixture import current_actor_context
+from leapp.models import StorageInfo, FstabEntry
+from leapp.reporting import Report
+
+
+def test_actor_with_fstab_entry(current_actor_context):
+    with_fstab_entry = [FstabEntry(fs_spec="//10.20.30.42/share1", fs_file="/mnt/win_share1",
+                                   fs_vfstype="cifs",
+                                   fs_mntops="credentials=/etc/win-credentials,file_mode=0755,dir_mode=0755",
+                                   fs_freq="0", fs_passno="0"),
+                        FstabEntry(fs_spec="//10.20.30.42/share2", fs_file="/mnt/win_share2",
+                                   fs_vfstype="cifs",
+                                   fs_mntops="credentials=/etc/win-credentials,file_mode=0755,dir_mode=0755",
+                                   fs_freq="0", fs_passno="0"),
+                        FstabEntry(fs_spec="/dev/mapper/fedora-home", fs_file="/home",
+                                   fs_vfstype="ext4",
+                                   fs_mntops="defaults,x-systemd.device-timeout=0",
+                                   fs_freq="1", fs_passno="2")]
+    current_actor_context.feed(StorageInfo(fstab=with_fstab_entry))
+    current_actor_context.run()
+    report_fields = current_actor_context.consume(Report)[0].report
+    assert 'inhibitor' in report_fields['flags']
+    assert report_fields['severity'] == 'high'
+    assert report_fields['title'] == "Use of CIFS detected. Upgrade can't proceed"
+
+
+def test_actor_no_cifs(current_actor_context):
+    with_fstab_entry = [FstabEntry(fs_spec="/dev/mapper/fedora-home", fs_file="/home",
+                                   fs_vfstype="ext4",
+                                   fs_mntops="defaults,x-systemd.device-timeout=0",
+                                   fs_freq="1", fs_passno="2")]
+    current_actor_context.feed(StorageInfo(fstab=with_fstab_entry))
+    current_actor_context.run()
+    assert not current_actor_context.consume(Report)
--- a/repos/system_upgrade/el7toel8/actors/checkcifs/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkcifs/actor.py
@@ -1,0 +1,21 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.checkcifs import checkcifs
+from leapp.models import StorageInfo
+from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckCIFS(Actor):
+    """
+    Check if CIFS filesystem is in use. If yes, inhibit the upgrade process.
+
+    Actor looks for CIFS in /ets/fstab.
+    If there is a CIFS entry, the upgrade is inhibited.
+    """
+    name = "check_cifs"
+    consumes = (StorageInfo,)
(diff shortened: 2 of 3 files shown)


repos/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py
```diff
@@ -75,65 +75,75 @@ def _get_partitions_info(partitions_path):
...
-                            'The /etc/fstab file must have at least 4 values and at most 6 per line, your file on the'
-                            ' line: {} have {} values, please edit and fix this, for further information read'
-                            ' fstab man page (man 5 fstab). '.format(line, len(entries))
-                        )
-                    summary = (
-                        'The fstab configuration file seems to be invalid. You need to fix it to be able to proceed'
-                        ' with the upgrade process.'
+    """ Collect storage info from /etc/fstab file
+        Returns a list of successfully processed fstab entries, errors_dict
+    """
```
pirat89, line 81:
```suggestion
    """
    Collect storage info from /etc/fstab file
    
    Return a list of successfully processed fstab entries and errors_dict
    """
```
Actually it returns tuple of two lists.

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py
```diff
@@ -75,65 +75,75 @@ def _get_partitions_info(partitions_path):
...
+    if not _is_file_readable(fstab_path):
+        return [], {}
```
pirat89, line 83: I am thinking about the change of the logic here. If the file is not readable, we should probably report error / inhibit upgrade as we are not able to check the very basic stuff otherwise. but that could be following PR.

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py
```diff
@@ -236,9 +246,23 @@ def _get_systemd_mount_info():
...
+        remediation = '\n'.join([e['remediation'] for e in errs])
+        reporting.create_report([
+            reporting.Title(title),
+            reporting.Summary(errs[0]['summary']),
+            reporting.Severity(errs[0]['severity']),
+            reporting.Tags([reporting.Tags.FILESYSTEM]),
+            reporting.Flags(errs[0]['flags']),
+            reporting.Remediation(hint=remediation),
+            reporting.RelatedResource('file', '/etc/fstab')
+        ])
```
pirat89, line 262: This logic is wrong from my POV. If I understand it well, it's exactly opposite of what we want. There is just one problem: invalid fstab. The number of wrong lines is just a parameter of the issue. And only thing we are interested about in this one particular actor is valid/invalid/readable fstab file.

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py
```diff
@@ -75,65 +75,75 @@ def _get_partitions_info(partitions_path):
...
+
+            fs_spec, fs_file, fs_vfstype, fs_mntops, fs_freq, fs_passno = entries
+            # NOTE(ivasilev) As specified in OAMG-4107, before networking is working in the initramfs stage
+            # the upgrade should be inhibited if any CIFS entries are found.
+            if fs_vfstype == 'cifs':
+                summary = 'CIFS entries are not supported at the moment.'
+                _add_error('CIFS entry encountered in /etc/fstab', summary,
+                           'Please comment out the CIFS entry at line {}.'.format(line))
+                api.current_logger().error(summary)
+                continue
```
pirat89, line 142: The logic about evalutaion of data should not be here. This actor should provide only what's inside the fstab file. Next to CIFS, we need to inhibit even NFS, etc. The actual check for the CIFS paritions should be implemented in the separate actor. It can be analogy to `checknfs` or even maybe better, the `checknfs` actor should be extended.

pirat89 (changes requested):
This implementation is not good. From several POVs, but mainly from the point of design. The actor should just parse the fstab file and deliver all data via the message. That's all what should be in this actor from my pov.

repos/system_upgrade/el7toel8/actors/storagescanner/libraries/storagescanner.py
pirat89: Of course, in case of extension of the NFS inhibitor, it would make sense to rename the actor.

fernflower: I'd rather have a different actor if that's ok.

fernflower:
@pirat89 changed the approach, covered with a simple test. Thank you!

repos/system_upgrade/el7toel8/actors/checkcifs/tests/test_checkcifs.py
```diff
@@ -0,0 +1,24 @@
...
+                        FstabEntry(fs_spec="/dev/mapper/fedora-home", fs_file="/home",
+                                   fs_vfstype="ext4",
+                                   fs_mntops="defaults,x-systemd.device-timeout=0",
+                                   fs_freq="1", fs_passno="2")]
+    current_actor_context.feed(StorageInfo(fstab=with_fstab_entry))
+    current_actor_context.run()
+    report_fields = current_actor_context.consume(Report)[0].report
+    assert 'inhibitor' in report_fields['flags']
+    assert report_fields['severity'] == 'high'
+    assert report_fields['title'] == "Use of CIFS detected. Upgrade can't proceed"
```
pirat89, line 24: lgtm now. just waiting for another test case as discussed for no-cifs detection. thanks for the quick change!

fernflower: Done. Hope I understood you correctly and you asked for 'it's ok' case test

pirat89: @fernflower  yes :) thanks

pirat89 (approved):
Lgtm and regarding unit-tests I believe the actor is doing what it should do.

pirat89:
leapp-ci build

examon:
leapp-ci build

pirat89 (approved):

pirat89:
@oamg/qe  to trigger re-build, just write: "leapp-ci build" (without quotes). anyway, this has to be rebased first to make installation possible nowadays

pirat89 (dismissed):

pirat89 (approved):
Rebased against the master. Verification has been done (OAMG-4107). Waiting for the finished tests and will merge it.

PR #633, opened 2021-01-14
Fix XFS ftype=0 workaround for "non-XFS" rootfs

pirat89:
We have realized that in case the /var/lib/leapp is on partition
with XFS ftype=0, we need to treat rootfs partition as the XFS ftype=0 one,
doesn't matter what actual FS is used for rootfs.

This implementation treat rootfs like that always if any XFS ftype=0
has been detected (just to save time now). It's even possible, the
root cause had a bigger scope and we will need to treat any FS with
our workaround in future. But this will definitely fix the bug for
now.

rhbz: #1911802

### DIFF
--- a/repos/system_upgrade/el7toel8/libraries/overlaygen.py
+++ b/repos/system_upgrade/el7toel8/libraries/overlaygen.py
@@ -63,7 +63,7 @@ def _prepare_required_mounts(scratch_dir, mounts_dir, mount_points, xfs_info):
 
     mount_names = [mount_point.fs_file for mount_point in mount_points]
 
-    for mountpoint in xfs_info.mountpoints_without_ftype:
+    for mountpoint in set(xfs_info.mountpoints_without_ftype + ['/']):
         if mountpoint in mount_names:
             image = _create_mount_disk_image(disk_images_directory, mountpoint)
             result[mountpoint] = mounting.LoopMount(source=image, target=_mount_dir(mounts_dir, mountpoint))


pirat89:
In case you want to test the devel build for this PR, do not forget here is the public COPR repository where our builds are created, so installing `leapp-repository*.PR633.*` provides the up-to-date build for this PR.

- https://copr.fedorainfracloud.org/coprs/g/oamg/leapp/repo/epel-7/group_oamg-leapp-epel-7.repo
- https://copr.fedorainfracloud.org/coprs/g/oamg/leapp/

repos/system_upgrade/el7toel8/libraries/overlaygen.py
```diff
@@ -63,7 +63,7 @@ def _prepare_required_mounts(scratch_dir, mounts_dir, mount_points, xfs_info):
 
     mount_names = [mount_point.fs_file for mount_point in mount_points]
 
-    for mountpoint in xfs_info.mountpoints_without_ftype:
+    for mountpoint in set(xfs_info.mountpoints_without_ftype + ['/']):
```
pirat89, line 66: I should add a comment here why the '/' is here..

pirat89:
leapp-ci build

pirat89:
leapp-ci build

examon:
leapp-ci build

repos/system_upgrade/el7toel8/libraries/overlaygen.py
pirat89: huh... I closed that by mistake *facepalm*

pirat89:
The issue is going to be resolved here https://github.com/oamg/leapp-repository/pull/633 as I've removed the original branch by mistake and cannot reopen it anymore.

PR #634, opened 2021-01-15
ziplconvertoblcfg: Ensure we won't delete /boot

vinzenz:
During the conversiion of zipl bootmanager configuration to the BLS
format, we bind mount /boot into the target userspace. Previously, the
bind mount cleanup removed everything from the mounted /boot directory.
This patch fixes the problem by making it avoid running cleanup actors
before and after mount operations.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/ziplconverttoblscfg/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/ziplconverttoblscfg/actor.py
@@ -38,7 +38,9 @@ class ZiplConvertToBLSCFG(Actor):
         # replace the original boot directory inside the container by the host one
         # - as we cannot use zipl* pointing anywhere else than default directory
         # - no, --bls-directory is not solution
-        with mounting.BindMount(source='/boot', target=os.path.join(userspace.path, 'boot')):
+        with mounting.BindMount(source='/boot',
+                                target=os.path.join(userspace.path, 'boot'),
+                                config=mounting.MountConfig.AttachOnly):
             userspace_zipl_conf = os.path.join(userspace.path, 'etc', 'zipl.conf')
             if os.path.exists(userspace_zipl_conf):
                 os.remove(userspace_zipl_conf)


vinzenz:
The culprint here is this:

```python
with mounting.BindMount(source='/boot', target=....)
```

Using this will fix it:
```python
with mounting.BindMont(source='/boot', target=..., config=mounting.MountConfig.AttachOnly):
```

pirat89 (approved):
The fix seems clear. Actually, the current solution fixes just this one issue already reported, but it doesn't deal with the generic problem that the used bindmount by default removes the content of the directory. Regarding our timeframe, this is good fix right now. But in future we should make it more generic to ensure this will not happen by accident when a custom actor is applied by someoen who is not aware of this behaviour. Even our current implementation can lead into the same issue before the reboot itself. The another more possible issue vector:
  1. leapp crashes when /boot is bindmounted inside targetuserspace
  2. another run of leapp results in a try to remove targetuserspace where the /boot is bind mounted
  3. the /boot is removed again

The fix could be to check whether any else mountpoint is inside the dierctory which are are going to remove to prevent this problem. iow, if we want to remove /a and /a/b is mounted inside, /a cannot be removed (it fails anyway in the end, after the files inside are removed...).

@vinzenz  Giving an approval, but expecting creation of the followup ticket to make work with the mounting library more safe. I expect we are not going to do the work right now.

Tested on system where the issue has been reproduced.

examon:
leapp-ci build

AloisMahdal:
FTR; the fix currently breaks all s390x, even with "normal" partition layouts, going to emergency mode.

We're currently investigating it with @vinzenz and @Rezney

pirat89:
The bug has been confirmed to be fixed by https://github.com/oamg/leapp-repository/commit/a6445b391a01bf17d3ad8229ca1185b10479f467 . The inhibitor will be dropped by PR https://github.com/oamg/leapp-repository/pull/991 . Closing this one

PR #637, opened 2021-01-25
docker-tests: pin get-pip.py version in Dockerfile

fernflower:
This should fix recent failures with the latest get-pip.py not
being py2-friendly.


examon:
leapp-ci build

pirat89 (approved):

PR #638, opened 2021-01-26
persistentnetnamesdisable: Fix the issue when 1 eth nic in multiple n…

winndows:
…ics env

We should only inhibit the upgrade when multiple eth nics exist.
Neverthless, for 1 eth nic in multiple nics env, we should not
fail the upgrade.

Signed-off-by: Liao Pingfang <liao.pingfang@zte.com.cn>

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/persistentnetnamesdisable/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/persistentnetnamesdisable/actor.py
@@ -38,7 +38,7 @@ class PersistentNetNamesDisable(Actor):
 
         if self.single_eth0(interfaces):
             self.disable_persistent_naming()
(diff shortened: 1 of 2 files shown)


examon:
leapp-ci build

examon:
leapp-ci build

examon:
leapp-ci build

pirat89:
@winndows Hi, thank you for the fix. It seems you are right. However, regarding we have a deadline soon and our CI is broken for two weeks, we will get to it after the release. Thank you for the patience.

pirat89:
@winndows Also please rebase the branch regarding the current master to fix the problem with `get-pip.py` https://github.com/oamg/leapp-repository/pull/637

PR #641, opened 2021-02-01
s390 inhibit if /boot is not on a separate partition

vinzenz:
Previously, when the upgrade is performed on a s390x machine with zipl
and /boot is not on a separate partition, the content of /boot got
deleted during the upgrade process. Since the potential solution for
this problem turned out to fail the upgrade for all s390 upgrades, it
has been decided to create an actor that inhibits the upgrade for the
detected scenario that caused the deletion of the /boot content.

This patch introduces a new check actor that inhibits the upgrade when:
- The architecture is s390X
- /boot is not a mounted partition

In case all of the above is true, the upgrade is inhibited with a
report.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checknonmountboots390/libraries/checknonmountboots390.py
+++ b/repos/system_upgrade/el7toel8/actors/checknonmountboots390/libraries/checknonmountboots390.py
@@ -1,0 +1,27 @@
+import os
+
+from leapp import reporting
+from leapp.libraries.common.config import architecture
+
+
+def perform_check():
+    if not architecture.matches_architecture(architecture.ARCH_S390X):
+        return
+
+    if os.path.ismount('/boot'):
+        return
+
+    data = [
+        reporting.Title('Leapp detected known issue related to /boot on s390x architecture'),
+        reporting.Summary((
+            'Due to a bug in the Leapp code, there is a situation when the upgrade process'
+            ' removes content of /boot when the directory is not on a separate partition and'
+            ' the system is running on S390x architecture. To avoid this from happening, we'
+            ' are inhibiting the upgrade process in this release until the issue has been fixed.'
+        )),
+        reporting.Flags([reporting.Flags.INHIBITOR]),
+        reporting.Tags([reporting.Tags.FILESYSTEM, reporting.Tags.UPGRADE_PROCESS, reporting.Tags.BOOT]),
+        reporting.Severity(reporting.Severity.HIGH),
+    ]
+
+    reporting.create_report(data)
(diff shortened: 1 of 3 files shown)


Rezney:
@AloisMahdal could you please verify this on some s390x?

repos/system_upgrade/el7toel8/actors/checknonmountboots390/libraries/checknonmountboots390.py
```diff
@@ -0,0 +1,26 @@
...
+
+    if os.path.ismount('/boot'):
+        return
+
+    data = [
+        reporting.Title('Found /boot is not on a separate partition'),
+        reporting.Summary((
+            'Due to some bug in the leapp code if /boot is not on a separate partition, the upgrade'
+            ' process erases the content of /boot. The problem could not be fixed in time and the upgrade'
+            ' has been inhibited to prevent data loss.'
```
Rezney, line 19: What about: "Due to a bug in the Leapp code, there is a situation when the upgrade process removes content of /boot when the directory is not on a separate partition and the system is running on S390x architecture. To avoid this from happening, we are inhibiting the upgrade process in this release until the issue has been fixed.

repos/system_upgrade/el7toel8/actors/checknonmountboots390/libraries/checknonmountboots390.py
```diff
@@ -0,0 +1,26 @@
...
+
+def perform_check():
+    if not architecture.matches_architecture(architecture.ARCH_S390X):
+        return
+
+    if os.path.ismount('/boot'):
+        return
+
+    data = [
+        reporting.Title('Found /boot is not on a separate partition'),
```
Rezney, line 15:
```suggestion
        Leapp detected known issue related to /boot on s390x architecture 
```

Rezney (commented):
Left couple of suggestions but looks good to me as a temporary measure.

Rezney (approved):

AloisMahdal:
With

 * *leapp-0.11.1-100.202102021032Z.d9ded67.master.el7_8*
 * *leapp-repository-0.12.0-0.202102020940Z.aaa020a.inhibit_s390_non_mount_boot.PR641.el7_8*

...upgrade is inhibited when, and only when the system is s390x *and* /boot is merged:

    Risk Factor: high (inhibitor)
    Title: Leapp detected known issue related to /boot on s390x architecture
    Summary: Due to a bug in the Leapp code, there is a situation when the upgrade process removes content of /boot when the directory is not on a separate partition and the system is running on S390x architecture. To avoid this from happening, we are inhibiting the upgrade process in this release until the issue has been fixed.
    Key: b2a53a503ba293a1d8cb87008113a0eb54fcab33

IOW, from QE POV the PR is good to merge.

PR #644, opened 2021-02-08
Use Groups instead of report Flags and report Tags

fernflower:


fernflower:
Accidentally closed https://github.com/oamg/leapp-repository/pull/620 and deleted the branch by adding extra whitespace around ':' in `git push ivasilev reportgroupspostrelease :reportgroups -f` 
So reopening same pr after rebase here.

fernflower:
Decided to take a different approach that won;t change the interface for actor writers. So closing.

PR #647, opened 2021-02-27
Make pylint 2.7.0 and 2.7.1 happy again

pirat89:
List of changes

- disable `use-a-generator` - cannot be fixed because the wanted change is incompatible with Python2
- update `_run_cmd` in selinux actors to set the return statement expicitly (the mock function looks weird, but let's ignore it)
- disable `R1726` & `R1727` in affected actors; the pylint's report is correct, however I guess that in this case the weird asserts have documentation purposes (@vinzenz  ?)

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -41,6 +41,7 @@ disable=
  bad-option-value,  # python 2 doesn't have import-outside-toplevel, but in some case we need to import outside toplevel
  super-with-arguments,  # required in python 2
  raise-missing-from,  # no 'raise from' in python 2
+ use-a-generator, # cannot be modified because of Python2 support
 
 [FORMAT]
 # Maximum number of characters on a single line.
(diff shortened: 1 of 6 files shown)


pirat89:
leapp-ci build

.pylintrc
```diff
@@ -41,6 +41,7 @@ disable=
  bad-option-value,  # python 2 doesn't have import-outside-toplevel, but in some case we need to import outside toplevel
  super-with-arguments,  # required in python 2
  raise-missing-from,  # no 'raise from' in python 2
+ use-a-generator, # cannot by modified because of Python2 support
```
drehak, line 44:
```suggestion
 use-a-generator, # cannot be modified because of Python2 support
```

drehak (approved):
A tiny typo, otherwise LGTM.

.pylintrc
pirat89: @drehak  thanks. fixed. re-approve pls

drehak (approved):
Yup.

pirat89:
Merging as the internal CI is broken and this change touches just unit-tests.

PR #648, opened 2021-03-02
WIP: Feature/add actor checkpcidrivers (pes-cloud edition)

drehak:
Same as #602, except `RestrictedPCIsScanner` is modified to grab data from pes-cloud.

The URL isn't live yet, so until then, you have to override it using the `LEAPP_DEVEL_PCIS_HOST` variable and point it to the host or IP of your existing instance. (You'll also have to change `server_name` in the service's nginx config.)

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
+++ b/repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
@@ -48,18 +48,19 @@ def get_restricted_pcis(http):
             default=PCIS_HOST_DEFAULT,
         ),
     )
+    headers = urllib3.make_headers(basic_auth="drehak@redhat.com:redhat")
 
     unsupported_driver_names = http.request(
         "GET",
         API_URL + "unsupported_driver_names.json",
         timeout=REQUEST_TIMEOUT,
-        headers={},
+        headers=headers,
     )
     unsupported_pci_ids = http.request(
         "GET",
         API_URL + "unsupported_pci_ids.json",
         timeout=REQUEST_TIMEOUT,
-        headers={},
+        headers=headers,
     )
 
     unsupported_driver_names = json.loads(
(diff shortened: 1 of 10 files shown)


drehak:
I've written a short guide on how to test this PR with **pes-cloud**, even when it's running locally.
(accessible only from RH network)

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
```diff
@@ -46,22 +42,22 @@ def get_restricted_pcis(http):
     # TODO this constant should be above the scope of this func.
     #   However due to the way how actor tested with CurrentActorMocked
     #   we can't do this.
-    API_URL = "http://{host_port}/api/{api_version}/".format(
+    API_URL = "http://{host_port}/api/pes/".format(
```
fernflower, line 45: `host_port` variable naming is puzzling to be honest, it took some time to go from "why is it referencing the port only" to "oh, it means hostname_with_port".
Can we have good old `host` or `server_url` here?

fernflower (commented):
One minor nit, otherwise lgtm.
Still needs to be tested with cloud pes though.

repos/system_upgrade/el7toel8/actors/restrictedpcisscanner/libraries/restrictedpcisscanner.py
drehak: Sure, I'll change it.

repos/system_upgrade/el7toel8/models/pcidevices.py
```diff
@@ -17,9 +58,71 @@ class PCIDevice(Model):
...
+
+    pci_id = fields.Nullable(fields.String())
+    driver_name = fields.Nullable(fields.String())
+    device_name = fields.Nullable(fields.String())
+    available_rhel7 = fields.Integer()
+    supported_rhel7 = fields.Integer()
+    available_rhel8 = fields.Integer()
+    supported_rhel8 = fields.Integer()
+    available_rhel9 = fields.Integer()
+    supported_rhel9 = fields.Integer()
```
pirat89, line 98: just nitpick. why not use bool instead? The current logic is binary and using integer just to tell whether it True or False seems useless. It's a nitpick anyway, I really do not require to transform it to bool, but it seems to me as a bad practice to use integer like that.

drehak:
Closing in lieu of #654.

PR #649, opened 2021-03-03
tmt first blood

pirat89:
just some playing with tmt...


thrix:
if I run
```make test```
in the work directory
```/var/tmp/tmt/run-081/unittests/discover/default/tests/tests/smoketests```
it returns 0
so ./test.sh executes in the test folder
```
[mvadkert@freedom /var/tmp/tmt/run-081/unittests/discover/default/tests/tests/smoketests] (tmt-poc)$ make test
make: 'test' is up to date.
```

pirat89:
@thrix it's really like that. it seems that "make" has an error, as `make test` without the Makefile should file instead of doing whatever. I am trying to find any note about that before I report BZ

thrix:
> @thrix it's really like that. ~it seems that "make" has an error, as `make test` without the Makefile should file instead of doing whatever. I am trying to find any note about that before I report BZ~ after additional research, it seems it is expected behaviour. I do not understand why, but...

yeah, suprised me also ...

pirat89:
Closing. This was kinda experimental branch and current work is far beyond this one.

PR #656, opened 2021-04-14
CI: travis: fix docker pull limit (centos registry)

pirat89:
Regarding the pull rate limit from the docker.io hub (100 pulls / 6h),
let's look at alternatives. There are two ways (see two commits) how
to achieve this.

1) Use the docker but pull from centos registry instead of from docker.io:
-  registry.centos.org/centos:7

This can be a little bit annoying when you want to remove an image and you
cannot write just "centos:7" but has to write all other stuff. (passed)

2) start to use podman, as docker is over. From my point that's better solution as docker is kind of history now. (failed, see comments)

Note: Tested in my VM. Now waiting for results on travis. Once the tests passes for the first commit, I will push the second one.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ os: linux
 dist: xenial
 env:
   global:
-    - CONTAINER=centos:7
+    - CONTAINER=registry.centos.org/centos:7
 services:
   - docker
 
--- a/utils/docker-tests/Dockerfile
+++ b/utils/docker-tests/Dockerfile
@@ -1,4 +1,4 @@
-FROM centos:7
+FROM registry.centos.org/centos:7
 
 VOLUME /payload
 


pirat89:
The solution 1) seems working: https://travis-ci.org/github/oamg/leapp-repository/builds/767013201

Going on solution 2)

pirat89:
Ok. There is not good way still to use podman on travis. We could do something like [here](https://github.com/junaruga/podman-experiment/blob/master/.travis.yml) but currently it seems it will be better postpone it until podman is supported natively on travis. So skipping the 2) option.

vinzenz (approved):

pirat89:
leapp-ci build

pirat89:
Note: The CI server is currently jammed by a huge queue of pending tasks regarding some recent outages. It seems it will work ok in 4+ hours. So let's merge this tomorrow..

pirat89:
leapp-ci build

.travis.yml
```diff
@@ -5,7 +5,7 @@ os: linux
 dist: xenial
 env:
   global:
-    - CONTAINER=centos:7
+    - CONTAINER=registry.centos.org/centos:7
```
vinzenz, line 8: this is not needed

pirat89: I do not understand. In the travis.yml file, you can see the docker pull command. In case I do not change this variable, the docker pull command fails on the limit pull limit rate.

pirat89: @vinzenz Or do you mean to drop the docker pull command as well? As I understand, the explicit pull is not necessary and then we could drop the CONTAINER envar completely.

vinzenz: Nevermind, I forgot that it is a bit different in leapp-repository

PR #658, opened 2021-04-21
WIP (experimental) inject failure

AloisMahdal:
This is **never to be merged** -- the purepose of this PR is to force
upgrade to fail in order to help with testing of #529.

Once #529 is verified or a better method is found, please drop this PR.


pirat89:
switching to draft to make clear this is nothing people should spend time on it.

pirat89:
btw, when you need a build, you do not need to create a PR just to have it. `make copr_build` from a side branch will do the work.

AloisMahdal:
> btw, when you need a build, you do not need to create a PR just to have it. `make copr_build` from a side branch will do the work.

ahh, thanks.  i can avoid spamming GitHub next time :)

AloisMahdal:
...wow, works like a charm -- and I can even run MORF test against this (`lpr=Finbandfail`)  thanks for the tip!

PR #660, opened 2021-04-26
amultiple attempts to fetch data from server and increased timeout

pirat89:
The original timout for the fetching of data was too low (just 5s).
Keep 5s for connection to the server and at least 30 seconds to read
the data.

As well, try to fetch data multiple times to reduce problems with
random network glitches. Currently 3 attempts per fetch of a data
file.

In case the reading of data is too slow (no connection timeout)
increase the timeout by 10s per attempt.

### DIFF
--- a/repos/system_upgrade/el7toel8/libraries/fetch.py
+++ b/repos/system_upgrade/el7toel8/libraries/fetch.py
@@ -7,7 +7,8 @@ from leapp.libraries.common.config import get_env
 from leapp.libraries.stdlib import api
 
 SERVICE_HOST_DEFAULT = "https://cert.cloud.redhat.com"
-REQUEST_TIMEOUT = 5
+REQUEST_TIMEOUT = (5, 30)
+MAX_ATTEMPTS = 3
 
 
 def _raise_error(local_path, details):
@@ -19,6 +20,32 @@ def _raise_error(local_path, details):
             " for more information about how to retrieve the file.")
 
     raise StopActorExecutionError(summary, details={'details': details, 'hint': hint})
+
+
+def _request_data(service_path, cert, proxies, timeout=REQUEST_TIMEOUT):
+    logger = api.current_logger()
+    attempt = 0
+    while True:
+        attempt += 1
+        try:
+            return requests.get(service_path, cert=cert, proxies=proxies, timeout=REQUEST_TIMEOUT)
+        except requests.Timeout as e:
+            etype_msg = 'Connection timeout'
+            if isinstance(e, requests.ReadTimeout):
+                etype_msg = 'Read timeout'
+                # reading is slow, increase the time limit for the reading
+                timeout[1] += 10
+            if attempt > MAX_ATTEMPTS:
+                logger.warning(
+                    'Attempt {} of {} to get {} failed: {}.'
+                    .format(MAX_ATTEMPTS, MAX_ATTEMPTS, service_path, etype_msg)
+                )
+                raise
+
+            logger.info(
+                'Attempt {} of {} to get {} failed: {}. Retrying...'
+                .format(attempt, MAX_ATTEMPTS, service_path, etype_msg)
+            )
 
 
 def read_or_fetch(filename, directory="/etc/leapp/files", service=None, allow_empty=False):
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/el7toel8/libraries/fetch.py
```diff
@@ -21,6 +22,34 @@ def _raise_error(local_path, details):
     raise StopActorExecutionError(summary, details={'details': details, 'hint': hint})
 
 
+def _request_data(service_path, cert, proxies, timeout=REQUEST_TIMEOUT):
+    logger = api.current_logger()
+    attempt = 0
+    while True:
+        attempt += 1
```
vinzenz, line 29:
```suggestion
    for attempt range(1, MAX_ATTEMPTS + 1):
```

repos/system_upgrade/el7toel8/libraries/fetch.py
```diff
@@ -21,6 +22,34 @@ def _raise_error(local_path, details):
...
+                    'Attempt {} of {} to get {} failed: {}.'
+                    .format(MAX_ATTEMPTS, MAX_ATTEMPTS, service_path, etype_msg)
+                )
+                raise
+
+            logger.info(
+                'Attempt {} of {} to get {} failed: {}. Retrying...'
+                .format(attempt, MAX_ATTEMPTS, service_path, etype_msg)
+            )
+    return response
```
vinzenz, line 50:
```suggestion
```

repos/system_upgrade/el7toel8/libraries/fetch.py
```diff
@@ -21,6 +22,34 @@ def _raise_error(local_path, details):
...
 
 
+def _request_data(service_path, cert, proxies, timeout=REQUEST_TIMEOUT):
+    logger = api.current_logger()
+    attempt = 0
+    while True:
+        attempt += 1
+        try:
+            response = requests.get(service_path, cert=cert, proxies=proxies, timeout=REQUEST_TIMEOUT)
+            break
```
vinzenz, line 32:
```suggestion
            return requests.get(service_path, cert=cert, proxies=proxies, timeout=REQUEST_TIMEOUT)
```

repos/system_upgrade/el7toel8/libraries/fetch.py
```diff
@@ -52,7 +81,7 @@ def read_or_fetch(filename, directory="/etc/leapp/files", service=None, allow_em
     cert = ("/etc/pki/consumer/cert.pem", "/etc/pki/consumer/key.pem")
     response = None
     try:
-        response = requests.get(service_path, cert=cert, proxies=proxies, timeout=REQUEST_TIMEOUT)
+        response = _request_data(service_path, cert=cert, proxies=proxies)
     except requests.RequestException as e:
         logger.error(e)
         _raise_error(local_path, "Could not fetch {f} from {sp} (unreachable address).".format(
```
vinzenz, line 87:
```
        _raise_error(local_path, "Could not fetch {f} from {sp} (unreachable address).".format(
```
'(unreachable address)' is probably misleading in case of ReadTimeout, no? I know it's unrelated, just happen to see this

pirat89: It is. I know about that. But anyway, it's already planned to do bunch of changes (refactoring, ....) around, so I vote to keep the message for now and fix it in the follow up (not sure whether the change of the message will not break the tests again..)

repos/system_upgrade/el7toel8/libraries/fetch.py
pirat89: I was thinking about that, but I was afraid that linter will be blaming missing return in the end of the function. I will try it

vinzenz: Well it would cause to be return None automatically (that's how it works in python)
However before that, you would end up raising an exception.

vinzenz: There are only two exit points in this exception, that is the return and the re-raising of the exception

pirat89: I know, but linters have (had?) problems with it because of FP. Anyway, it seems it's passing like that.

repos/system_upgrade/el7toel8/libraries/fetch.py
pirat89: done

repos/system_upgrade/el7toel8/libraries/fetch.py
pirat89: I prefer to keep it stupid, They way it's written it's clear there are just two possibilities how the function ends, without need to think about that.

Rezney (approved):
Tested. Seems to be working fine.

PR #664, opened 2021-05-11
[WIP] [completely broken now] Refactoring & fixing the fetching of data and PCI stuff

pirat89:
## Summary

The current implementation around fetching of data and especially around PCI stuff is not fulfilling our needs as:
* it can be used on RHEL 7 only now
* it's not generic - it cannot be used in the multiple repositories in the same time; the code duplication and modification is required now
* the proposed models around restricted PCIs, reflecting the real data structure, is not usable and suitable to use for a generic processing
* the code is not python2 & python3 compatible
* ...etc

The goal of this PR is to create **clean code** that will be generic for the use in upgrades.

## TODOs etc.
- [ ] propose new model design for the restricted PCI devices with which we could work in reasonabe way
- [ ] Update the fetch function to return expected output on python2 & python3
  - now it returns different stuff on py2 & py3 which breaks actors... 
- [ ] Update the restrictedPCIscanner actor properly to address the changes in models
- [ ] ...

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkpcidrivers/actor.py
@@ -7,17 +7,14 @@ from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
 
 class CheckPCIDrivers(Actor):
     """
-    Check if host PCI devices drivers has restrictions on target system.
+    Check if detected PCI devices are supported on a target system.
 
-    If driver is restricted this means it is either unsupported or unavailable
-    in some set of newer RHEL versions.
+    Inhibit the ugprade if any detected drivers are unavailable on the target
+    system.
 
-    If the driver is unavailable on a target system - then the upgrade will
-        be inhibited
-    If the driver is available, but not supported - then the upgrade will
-        continue, but the relevant report will be generated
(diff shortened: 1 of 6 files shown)


pirat89:
@vinzenz that's the current draft

MichalHe:
The problems that are mentioned in the PR description about the fetch function (`read_or_fetch`), are addressed in the #717. Tested this on RHEL8 (python3.6) machine - truncated output of `leapp upgrade` logs:
```
WARNING leapp.workflow.FactsCollection.restricted_pcis_scanner: File /etc/leapp/files/unsupported_driver_names.json does not exist, falling back to online service
WARNING leapp.workflow.FactsCollection.restricted_pcis_scanner: File <URL>/unsupported_driver_names.json successfully retrieved and read (X bytes)
WARNING leapp.workflow.FactsCollection.restricted_pcis_scanner: File /etc/leapp/files/unsupported_pci_ids.json does not exist, falling back to online service
WARNING leapp.workflow.FactsCollection.restricted_pcis_scanner: File <URL>/unsupported_pci_ids.json successfully retrieved and read (X bytes)
INFO    leapp.workflow.FactsCollection: Executing actor biosdevname 
``` 
The upgrade process did not crash on pci related data fetching and reading.

PR #665, opened 2021-05-12
Set the default shell used in Makefile to bash

vojtechsokol:
The default shell in Makefile is `sh` and since bashisms are used in the
Makefile, some of the targets fail, depending on setting of the system
(not an issue on vanilla Fedora where `/bin/sh` is a symlink to `bash`).


pirat89 (approved):
pls next time create the branch in your fork.

PR #666, opened 2021-05-13
Removed drivers actor needs to check all devices

shaded-enmity:
not only PCI devices

Ping @mmacura311 @pirat89 @drehak


pirat89 (approved):
discussed in person on the mtg. Let's consider it finally ok :D

pirat89 (approved):
rebased and updated component tests.

PR #636, opened 2021-01-20
[WIP] Handle modules in PES event scanner

drehak:
Now that modular packages are represented in PES data, we need to integrate that data into our PES event scanner actor.
Larger refactors are necessary because a dict in the format `{'name': 'repository'}` is not enough to handle packagesets with modular packages.

Work in progress, please do not review.

PR #667, opened 2021-05-17
Modularity pt. 1: Include module data in InstalledRPM model

drehak:
To calculate RPM transactions involving modular packages, we first need data about existing modular packages. Here we get the data through the `dnf` Python module and add information about module streams to the `InstalledRPM` model, which we extend with two new nullable fields.
Another implementation of mapping of repositories to packages using the `dnf` module has also been added. This is necessary for future upgrades from RHEL 8 to 9. (The code in question has been buggy on RHEL 7 but on RHEL 8, it works correctly.) The existing implementation using the `yum` module will still be used on RHEL 7.

Code taken from #636 (that one is pretty messy).

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1789840


pirat89:
review in progress

pirat89 (dismissed):

pirat89 (approved):

PR #670, opened 2021-05-24
Fix the pip installation in Dockerfile for unit-tests

pirat89:
The easy_install script is not usable anymore since the html page
changed the format and the script is not able to discover required
python packages anymore. Instead of that, install pip from centos
(includes installation of the EPEL repository), update it to version
we require and continue..

As well, to omit possible additional problems in future with pip for
Python3, let's install in the container python3-pip and update it
to the very same version (v20.3.4) too.


Rezney (approved):

Rezney:
The change is related just to the Travis CI which passed.

PR #672, opened 2021-06-07
The Modularity integration

vinzenz:
This patch introduces support for modularity.

    PES events will drive what module streams are going to be enabled and
    reset. Additionally users can now request module streams to be enabled
    or reset via the `RPMTransactionTasks` message by filling the
    `modules_to_enable` or `modules_to_reset` respectively.

NOTE(pirat89): the reset of module-streams is not working yet and has been commented out in this pr from our upgrade dnf plugin. The fix will be delivered in a following PR

### DIFF
--- a/repos/system_upgrade/common/actors/filterrpmtransactionevents/actor.py
+++ b/repos/system_upgrade/common/actors/filterrpmtransactionevents/actor.py
@@ -1,7 +1,10 @@
 from leapp.actors import Actor
-from leapp.models import (FilteredRpmTransactionTasks,
-                          InstalledRedHatSignedRPM, PESRpmTransactionTasks,
-                          RpmTransactionTasks)
+from leapp.models import (
+    FilteredRpmTransactionTasks,
+    InstalledRedHatSignedRPM,
+    PESRpmTransactionTasks,
+    RpmTransactionTasks
+)
 from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
 
 
@@ -29,11 +32,15 @@ class FilterRpmTransactionTasks(Actor):
         to_remove = set()
         to_keep = set()
         to_upgrade = set()
+        modules_to_enable = {}
+        modules_to_reset = {}
         for event in self.consume(RpmTransactionTasks, PESRpmTransactionTasks):
             local_rpms.update(event.local_rpms)
             to_install.update(event.to_install)
             to_remove.update(installed_pkgs.intersection(event.to_remove))
             to_keep.update(installed_pkgs.intersection(event.to_keep))
+            modules_to_enable.update({'{}:{}'.format(m.name, m.stream): m for m in event.modules_to_enable})
+            modules_to_reset.update({'{}:{}'.format(m.name, m.stream): m for m in event.modules_to_reset})
 
         to_remove.difference_update(to_keep)
 
@@ -45,4 +52,6 @@ class FilterRpmTransactionTasks(Actor):
             to_install=list(to_install),
             to_remove=list(to_remove),
             to_keep=list(to_keep),
-            to_upgrade=list(to_upgrade)))
+            to_upgrade=list(to_upgrade),
+            modules_to_reset=list(modules_to_reset.values()),
+            modules_to_enable=list(modules_to_enable.values())))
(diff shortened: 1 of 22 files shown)


repos/system_upgrade/common/actors/filterrpmtransactionevents/actor.py
```diff
@@ -29,11 +29,15 @@ def process(self):
...
         to_upgrade = set()
+        modules_to_enable = set()
+        modules_to_disable = set()
         for event in self.consume(RpmTransactionTasks, PESRpmTransactionTasks):
             local_rpms.update(event.local_rpms)
             to_install.update(event.to_install)
             to_remove.update(installed_pkgs.intersection(event.to_remove))
             to_keep.update(installed_pkgs.intersection(event.to_keep))
+            modules_to_enable.update(event.modules_to_enable)
+            modules_to_disable.update(event.modules_to_enable)
```
MichalHe, line 40: This might be a typo:
```suggestion
            modules_to_disable.update(event.modules_to_disable)
```

evgeni:
Thanks @vinzenz! This is really the missing piece to make #733 nice and structured.

Couple of questions:
* Why did you use the `modules_to_enable=[Module(name='postgresql', stream='12')]` syntax, over a more simple `modules_to_enable=['postgresql:12']`? If I read `libraries/dnfplugin.py` correctly, you end up flattening these objects to strings anyways? As a CLI user, the `module:stream` syntax is already known to me and feels more natural?
* Is there a way to add integration tests that test the following scenario: 
  * Start with EL7 that has PostgreSQL (or Ruby) installed from the Base OS
  * Upgrade to EL8 and use a non-default module stream (like `postgresql:12` or `ruby:2.7`)
  * Verify that the correct module stream is enabled
  * This is the scenario where I need `dnf module reset` in #733 as we upgrade from EL7 with default Ruby to EL8 with Ruby 2.7 (instead of 2.5) -- I believe this should *just* work as LEAPP instructs DNF to do everything in one transaction, but having a test is always good ;-)

repos/system_upgrade/common/actors/peseventsscanner/libraries/peseventsscanner.py
```diff
@@ -339,13 +385,34 @@ def parse_action(action_id):
...
+
 def parse_packageset(packageset):
     """
     Get "input" or "output" packages and their repositories from each PES event.
 
-    :return: A dictionary with packages in format {<pkg_name>: <repository>}
+    :return: set of Package tuples
     """
-    return {p['name']: p['repository'] for p in packageset.get('package', [])}
+    return {parse_package(p) for p in packageset.get('package', packageset.get('packages', []))}
```
evgeni, line 406: This raises a `TypeError: unhashable type: 'list'`

This works:
```suggestion
    return set(parse_package(p) for p in packageset.get('package', packageset.get('packages', [])))
```

evgeni, line 406: This raises `TypeError: unhashable type: 'list'`

vinzenz: Yeah I know, working on it

repos/system_upgrade/common/models/rpmtransactiontasks.py
```diff
@@ -10,6 +10,8 @@ class RpmTransactionTasks(Model):
     to_keep = fields.List(fields.String(), default=[])
     to_remove = fields.List(fields.String(), default=[])
     to_upgrade = fields.List(fields.String(), default=[])
+    modules_to_enable = fields.List(fields.Model(Module), default=[])
+    modules_to_reset = fields.List(fields.Model(Module), default=[])
```
pirat89, line 14: we should finally cover model's fields by docstrings. but not blocker for this pr.

pirat89 (commented):
Review still in progress

repos/system_upgrade/common/libraries/module.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    if not dnf:
+        return []
+
+    base = _create_or_get_dnf_base()
+    modules = get_modules(base)
+
+    # if modules are not supported (RHEL 7), base.sack._moduleContainer won't exist
+    # luckily in that case modules are empty and the element won't even be accessed
+    return [m for m in modules if base.sack._moduleContainer.isEnabled(m)]
```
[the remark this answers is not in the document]
vinzenz: because it's not about consistency but how it is used. get_enabled_modules() just produces it, get_modules is used elsewhere too where the base was already created. get_enabled_modules isn't.

PR #674, opened 2021-06-07
Fix XFS ftype=0 workaround for "non-XFS" rootfs

pirat89:
We have realized that in case the /var/lib/leapp is on partition
with XFS ftype=0, we need to treat rootfs partition as the XFS ftype=0 one,
doesn't matter what actual FS is used for rootfs.

This implementation treat rootfs like that always if any XFS ftype=0
has been detected (just to save time now). It's even possible, the
root cause had a bigger scope and we will need to treat any FS with
our workaround in future. But this will definitely fix the bug for
now.

rhbz: #1911802

### DIFF
--- a/repos/system_upgrade/common/libraries/overlaygen.py
+++ b/repos/system_upgrade/common/libraries/overlaygen.py
@@ -63,7 +63,16 @@ def _prepare_required_mounts(scratch_dir, mounts_dir, mount_points, xfs_info):
 
     mount_names = [mount_point.fs_file for mount_point in mount_points]
 
-    for mountpoint in xfs_info.mountpoints_without_ftype:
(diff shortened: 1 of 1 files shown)


vinzenz (dismissed):

pirat89:
leapp-ci build

vinzenz (approved):

pirat89:
Build has been created manually.

mkluson:
leapp-ci build

pirat89:
@vinzenz  @oamg/developers  re-approve pls. Rebase against the master branch.

vinzenz (approved):

pirat89:
Verified by @williamfsanford . Merging

PR #675, opened 2021-06-10
Github action for copr build & tests launch

fernflower:
Shamelessly copied from leapp to fix minor things later.

### DIFF
--- a/.github/workflows/tmt-tests.yml
+++ b/.github/workflows/tmt-tests.yml
@@ -1,0 +1,142 @@
+name: tmt@TF
+
+on:
+  issue_comment:
+    types:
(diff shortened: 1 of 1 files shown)


vojtechsokol:
The copr build fails because the Makefile uses `copr` command instead of `copr-cli`. On systems with copr installed from rpm it's not a problem (because copr is a symlink to copr-cli) but when installed from pip only `copr-cli` is available on the system.

jmikovic:
I opened [PR](https://github.com/oamg/leapp-repository/pull/676) to fix the Makefile

jmikovic (approved):

PR #678, opened 2021-06-22
xfsinfoscanner: Do not use systemd-mount data

vinzenz:
The data from systemd-mount is not relevant for the xfsinfoscanner. All
the parititions mounted are reported through /proc/mounts and the output
as is from systemd-mount --list is not giving us any additionally useful
output. It rather caused problems by having PCI addresses as mount
paths, which caused failures when we tried to get the information if the
xfs partition has ftype=0.

### DIFF
--- a/repos/system_upgrade/common/actors/xfsinfoscanner/libraries/xfsinfoscanner.py
+++ b/repos/system_upgrade/common/actors/xfsinfoscanner/libraries/xfsinfoscanner.py
@@ -20,15 +20,6 @@ def scan_xfs_mount(data):
     return mountpoints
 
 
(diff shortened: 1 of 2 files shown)


pirat89 (approved):
make sense and change is clear. Is there a related BZ? If so, it has to be verified by QE, so not merging it yet.

pirat89:
builds created automatically.

pirat89:
all for now finished integration tests passed. the last one is about `leapp-data` only (fetching data, ...) 83% done without any issue. however this one will probably run another couple of hours and do not believe in any valuable data from it regarding the change. so feel free to take it as passed.

PR #681, opened 2021-06-22
grubdevname: Python3 compatibility fix

vinzenz:


pirat89 (commented):
we should cover this change by unit-test

vinzenz:
it is, already covered that's how I figured out that i broke python2 with
the initial implementation

pirat89:
aaah, great!

Rezney (approved):
The tests and the upgrade itself passed locally.

pirat89:
ahh..how we just simply overlooked the failed tests? @vinzenz  @Rezney 

btw, I am almost sure I've seen passed tests before. so I am really confused I see the fail right now. Am I the only one who remember the very same thing or you realized the green test results too? O_o

PR #684, opened 2021-06-24
fix the bug that the persistentnetnamesdisable actor cannot obtain the virtio NIC

linux-gcc:
Because the ID_PATH(virtio-pci-xxx:xx:x) of the vritio network card obtained by systemd-219-42 is at the beginning,
the persistentnetnamesdisable actor fails to obtain the information of this type of network card.


vinzenz (changes requested):
LGTM however we should extend tests

linux-gcc:
oamg-ci/e2e
Travis CI - Pull Request

linux-gcc:
> LGTM however we should extend tests
Do you mean to pass the oamg-ci/e2e test?
@vinzenz

pirat89:
@linux-gcc  I guess that @vinzenz  is talking about unit-tests.

linux-gcc:
I have done a test based on sytemd-219-42 .
what else do I need to do now ?

@pirat89 @vinzenz

pirat89:
@linux-gcc  rebase please. we did significant changes - the file you updated has been moved into the common leapp repo (instead of el7toel8)

linux-gcc:
@vinzenz Should I rebase based on which branch?

pirat89:
@linux-gcc  always against the master. you need to fetch the up-to-date state of the master branch to see the changes. currently the library is in the `repos/system_upgrade/common/libraries/` directory.

pirat89:
@linux-gcc just be aware that by mistake we broke two actors yesterday, so you will need to rebase soon again (when we fix them) otherwise tests will always fail.

linux-gcc:
@pirat89 I have rebase, please review the code. 4e82fae
thanks!

linux-gcc:
@pirat89 
I have rebase, why the test still reports an error? May I ask why this is caused?

pirat89:
@linux-gcc  Interesting. The problem is related to oamg/leapp/pull/718 (which should be fixed already by that PR). Currently travis became dead for this project (since this morning?), so @vinzenz is working on the fix of the CI.

In case you are running tests locally, you need to remove the `tut` directory and install all dependencies again to fix the problem (on fedora: `make install-deps-fedora`).

linux-gcc:
@pirat89  ok, thanks!

linux-gcc:
@pirat89  ok, thanks!

pirat89:
@linux-gcc  Hi, the CI should be working again - just rebase against the current master.

linux-gcc:
@pirat89 I have rebase. I have rebase. Please help review the code. thanks!

pirat89 (approved):
The change seems good to me.

pirat89:
@linux-gcc 
> I have done a test based on sytemd-219-42 .
> what else do I need to do now ?

Just to be sure, you are talking about an automatic test? Ping me please over #leapp irc channel so we can finally resolve this.

linux-gcc:
> Just to be sure, you are talking about an automatic test? Ping me please over #leapp irc channel so we can finally resolve this.
I just did manual testing based on systemd-219-42 before

pirat89:
@linux-gcc ahh... now I guess I understand it really properly. I thought we are talking about behaviour of systemd present since 219-42 till the last systemd on rhel 7.9 (`systemd-219-78`). Just friendly reminder that you have to have at least `rhel-7.6` to be able to run `leapp` - which means at least `systemd-219-62`. The mentioned systemd release (42) is invalid for leapp, as we are strict the leapp can be run just on RHEL 7.6+ (which mean fully upgraded 7.6, not only several rpms) - which means that `leapp` can never be run with this version of systemd so users cannot even see this issue.

@linux-gcc What is the benefit of this PR in this case? I want to be sure this helps to resolve something and it's not just a death code with no value for the project.

## My testing with the current upstream build

My system:
```
[root@localhost ~]# udevadm info /sys/class/block/vda
P: /devices/pci0000:00/0000:00:05.0/virtio2/block/vda
N: vda
S: disk/by-path/pci-0000:00:05.0
S: disk/by-path/virtio-pci-0000:00:05.0
E: DEVLINKS=/dev/disk/by-path/pci-0000:00:05.0 /dev/disk/by-path/virtio-pci-0000:00:05.0
E: DEVNAME=/dev/vda
E: DEVPATH=/devices/pci0000:00/0000:00:05.0/virtio2/block/vda
E: DEVTYPE=disk
E: ID_PART_TABLE_TYPE=dos
E: ID_PATH=pci-0000:00:05.0
E: ID_PATH_TAG=pci-0000_00_05_0
E: MAJOR=253
E: MINOR=0
E: SUBSYSTEM=block
E: TAGS=:systemd:
E: USEC_INITIALIZED=5830
```

Running the same code on RHEL 7.4 with the unsupported version of systemd, I've been successful to reproduce the problem (see the empty `pci_info`):
```
[root@localhost ~]# python -i psysnet.py 
>>> from pprint import pprint
>>> pprint([i for i in interfaces()])
[{'devpath': u'/devices/pci0000:00/0000:00:02.0/virtio0/net/eth0',
  'driver': u'virtio_net',
  'mac': '52:54:00:8b:35:4f',
  'name': u'eth0',
  'pci_info': {},
  'vendor': u'0x1af4'}]
>>> 
```

But I am not able to reproduce the problem with `systemd-219-62` and newer:
```bash

[root@localhost ~]# python -i psysnet.py 
>>> from pprint import pprint
>>> pprint([i for i in interfaces()])
[{'devpath': u'/devices/pci0000:00/0000:00:02.0/virtio0/net/eth0',
  'driver': u'virtio_net',
  'mac': '52:54:00:8b:35:4f',
  'name': u'eth0',
  'pci_info': {'bus': u'00',
               'device': u'02',
               'domain': u'0000',
               'function': u'0'},
  'vendor': u'0x1af4'}]
>>> 
[root@localhost ~]# rpm -q systemd
systemd-219-78.el7_9.3.x86_64
```

@oamg/developers The change doesn't look harmful but I do not see currently any benefit of this PR, wdyt?

linux-gcc:
Why can't leapp be used in rhel 7.2? Can it be used on rhel 7.2 after solving this problem? 
I think that if this problem is solved, leapp can be used in versions lower than rhel 7.6. 
@pirat89

pirat89:
> Why can't leapp be used in rhel 7.2? Can it be used on rhel 7.2 after solving this problem?
> I think that if this problem is solved, leapp can be used in versions lower than rhel 7.6.
> @pirat89

You are missing bunch of other changes on the system which could lead into various problems. 7.2? There is not even provided `overlay2` - which is presented since rhel 7.4. Nowadays depending on your system setup, you could hit various issues breaking the upgrade - or inhibiting the upgrade because of errors - which are fixed since RHEL 7.8.

linux-gcc:
> > Why can't leapp be used in rhel 7.2? Can it be used on rhel 7.2 after solving this problem?
> > I think that if this problem is solved, leapp can be used in versions lower than rhel 7.6.
> > @pirat89
> 
> @linux-gcc You are missing bunch of other changes on the system which could lead into various problems. e.g. on RHEL 7.2 there is not even provided `overlay2` - which is presented since rhel 7.4. Nowadays depending on your system setup, you could hit various issues breaking the upgrade - or inhibiting the upgrade because of errors - which are fixed since RHEL 7.8.

At present, I only encounter this problem when using leapp on 7.4, so I think that after solving this problem, it should be able to be used directly on 7.4?

vinzenz:
Hi @linux-gcc 
We have been internally discussing this change request of yours and we came to the conclusion that it has unfortunately no benefit to our project. We did check with the systemd team to verify that the code changes aren't required in RHEL 7.6 and higher. 
We know that you tried to bring support for a lower version of RHEL, however there are other problems involved, including problems with the overlayfs feature which is one of our biggest problems that needs to be supported in order for our system to work. 
Therefore we have unfortunately to close this pull request.

PR #686, opened 2021-06-28
Remove duplicated definition of models

pirat89:
Several models are defined multipletimes. Remove the files that are
supposed to not exist already. The duplicated models:
* RequiredTargetUserspacePackages
* RequiredUpgradeInitramPackages
* TargetUserSpaceInfo
* UpgradeDracutModule


pirat89:
@vinzenz just for the curiosity do we have any check for redefinitions in models? It's not urgent now, but probably good feature candidate in future

pirat89:
I've realized all these are duplicated:
```
class RequiredTargetUserspacePackages(Model):
class RequiredUpgradeInitramPackages(Model):
class TargetUserSpaceInfo(Model):
class UpgradeDracutModule(Model):
```

I am going to update the PR now.

Rezney (approved):

PR #687, opened 2021-06-29
Makefile: update building of packages

pirat89:
* Added the DIST_VERSION option to be able to specify for what dist
  we should create the packages. This is preparation for building
  of packages in future. E.g. we will need to build deps packages
  for el8, but also for el9. The default value is: 7.

* Build the deps rpms locally instead of via the COPR. The deps
  packages do not have any buildrequires nor content - they are
  just metapackages to set expected provides and requirements.
  As well, the building process is now faster because of that
  by 50%.

* New temporary default for the COPR_CHROOT is "rhel-7-x86_64".
  We will remove it in future, but currently we plan to enable
  additional chroots in one of our COPRs and want to prevent
  people from building problems.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ SHELL=/bin/bash
 
 __PKGNAME=$${_PKGNAME:-leapp-repository}
 VENVNAME ?= tut
+DIST_VERSION ?= 7
 PKGNAME=leapp-repository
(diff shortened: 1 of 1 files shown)


Rezney (approved):

pirat89:
ping @oamg/developers  just to be sure you are aware of the change. rebase your branches please

PR #688, opened 2021-06-30
grubdevname: Fix for py3 compatibility change

vinzenz:
In a previous change a test breaking change was accidentally merged.
This patch fixes the problem.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/grubdevname/libraries/grubdevname.py
+++ b/repos/system_upgrade/el7toel8/actors/grubdevname/libraries/grubdevname.py
@@ -20,8 +20,9 @@ def has_grub(blk_dev):
         raise StopActorExecution()
     os.close(blk)
     test = 'GRUB'
-    if bytes is not str:
+    if not isinstance(mbr, str):
         test = test.encode('utf-8')
+
     return test in mbr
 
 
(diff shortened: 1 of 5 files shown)


repos/system_upgrade/el7toel8/actors/grubdevname/libraries/grubdevname.py
```diff
@@ -20,8 +20,9 @@ def has_grub(blk_dev):
         raise StopActorExecution()
     os.close(blk)
     test = 'GRUB'
-    if bytes is not str:
+    if not isinstance(mbr, str):
```
pirat89, line 23: This condition is on two places, I suggest to update the other one too.

vinzenz:
that is a different problem. the condition works but the unit test on py3
returns a string even though bytes should be returned.

so this is trying to fix the symptoms
i couldn't figure out why this happens though

repos/system_upgrade/el7toel8/actors/grubdevname/libraries/grubdevname.py
pirat89: mentioned below. it's ok

pirat89 (approved):

PR #692, opened 2021-07-14
Add tests for the scansubscriptionmanagerinfo actor

MichalHe:
Added tests for the `scansubscriptionmanagerinfo` verifying its functionality. There are overall 9 new tests that together assert all parts of the model produced by the actor.

### DIFF
--- a/repos/system_upgrade/common/actors/scansubscriptionmanagerinfo/tests/test_scansubscriptionmanagementinfo.py
+++ b/repos/system_upgrade/common/actors/scansubscriptionmanagerinfo/tests/test_scansubscriptionmanagementinfo.py
@@ -1,0 +1,43 @@
+import pytest
+
+from leapp.libraries.actor import scanrhsm
+from leapp.libraries.common import rhsm
+from leapp.libraries.common.testutils import CurrentActorMocked, produce_mocked
+from leapp.libraries.stdlib import api
+from leapp.models import RepositoryData, RepositoryFile, RHSMInfo
+
+
+def mocked_get_rhsm_info(context):
+    assert context, 'The actor did not provide library with valid context.'
+    info = RHSMInfo()
+    info.attached_skus = ['SKU1', 'SKU2']
+    info.available_repos = ['Repo1', 'Repo2']
+    info.enabled_repos = ['Repo2']
+    info.release = '7.9'
+    info.existing_product_certificates = ['Cert1', 'Cert2', 'Cert3']
+    info.sca_detected = True
+    return info
+
+
+def test_scansubscriptionmanagementinfo(monkeypatch):
+    actor_producs = produce_mocked()
+
+    monkeypatch.setattr(rhsm, 'scan_rhsm_info', mocked_get_rhsm_info)
+    monkeypatch.setattr(api, 'produce', actor_producs)
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked())
+
+    scanrhsm.scan()
+
+    assert actor_producs.model_instances, 'The actor did not produce any message.'
+    assert len(actor_producs.model_instances) == 1, 'The actor produced more messages than expected.'
+
+    message = actor_producs.model_instances[0]
+
+    # The actor does not do much more than calling the `rhsm` library (which has its own tests),
+    # just check that the message has not changed
+    assert message.attached_skus == ['SKU1', 'SKU2']
+    assert message.available_repos == ['Repo1', 'Repo2']
+    assert message.enabled_repos == ['Repo2']
+    assert message.release == '7.9'
+    assert message.existing_product_certificates == ['Cert1', 'Cert2', 'Cert3']
+    assert message.sca_detected
--- a/repos/system_upgrade/common/libraries/tests/test_rhsm.py
+++ b/repos/system_upgrade/common/libraries/tests/test_rhsm.py
@@ -1,16 +1,67 @@
 from collections import namedtuple
+import os
 
 import pytest
 
 from leapp import reporting
 from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.common import repofileutils, rhsm
-from leapp.libraries.common.testutils import create_report_mocked, logger_mocked
+from leapp.libraries.common.testutils import (
+    create_report_mocked,
+    CurrentActorMocked,
+    logger_mocked
+)
 from leapp.libraries.stdlib import CalledProcessError, api
 from leapp.models import RepositoryFile, RepositoryData
 
 Repository = namedtuple('Repository', ['repoid', 'file'])
 LIST_SEPARATOR = '\n    - '
+
+# External commands called by the RHSM library
+CMD_RHSM_LIST_CONSUMED = ('subscription-manager', 'list', '--consumed')
+CMD_RHSM_STATUS = ('subscription-manager', 'status')
+CMD_RHSM_RELEASE = ('subscription-manager', 'release')
+CMD_RHSM_LIST_ENABLED_REPOS = ('subscription-manager', 'repos', '--list-enabled')
+
+RHSM_STATUS_OUTPUT_NOSCA = '''
++-------------------------------------------+
+   System Status Details
++-------------------------------------------+
+Overall Status: Current
+
+System Purpose Status: Not Specified
+'''
+
+RHSM_STATUS_OUTPUT_SCA = '''
++-------------------------------------------+
+   System Status Details
++-------------------------------------------+
+Overall Status: Current
+
+System Purpose Status: Matched
+
+Content Access Mode is set to Simple Content Access
+'''
+
+# Used to simulate realistic output of RHSM, therefore carries more information than `Repository` namedtuple
+RHSMRepositoryEntry = namedtuple('RHSMRepositoryEntry', ('id', 'name', 'url', 'enabled'))  # For clarity purposes
+RHSM_ENABLED_REPOS = [
+    RHSMRepositoryEntry(
+        id='rhel-8-for-x86_64-appstream-rpms',
+        name='Appstream',
+        url='some_url',
+        enabled='1'),
+    RHSMRepositoryEntry(
+        id='satellite-tools-6.6-for-rhel-8-x86_64-rpms',
+        name='Satellite',
+        url='some_url',
+        enabled='1'),
+    RHSMRepositoryEntry(
+        id='rhel-8-for-x86_64-baseos-rpms',
+        name='Base',
+        url='some_url',
+        enabled='1')
+]
 
 
 class IsolatedActionsMocked(object):
@@ -19,11 +70,43 @@ class IsolatedActionsMocked(object):
         self.call_return = {'stdout': call_stdout, 'stderr': None}
         self.raise_err = raise_err
 
-    def call(self, cmd, *args):
+        # A map from called commands to their mocked output
+        self.mocked_command_call_outputs = dict()
+
+    def call(self, cmd, *args, **dummy_kwargs):
         self.commands_called.append(cmd)
         if self.raise_err:
             raise_call_error(cmd)
-        return self.call_return
+
+        return self.mocked_command_call_outputs.get(
+            tuple(cmd),  # Cast to tuple, as list is not hashable
+            self.call_return)
+
+    def add_mocked_command_call_with_stdout(self, cmd, stdout):
+        # We cast `cmd` from list to tuple, as a list cannot be hashed
+        self.mocked_command_call_outputs[tuple(cmd)] = {
+            'stdout': stdout,
+            'stderr': None}
+
+    def full_path(self, path):
+        return path
+
+
+@pytest.fixture
+def actor_mocked(monkeypatch):
+    """
+    Fixture providing a mocked actor that was already used to monkeypatch api.current_actor.
+
+    Introduced to reduce repetition inside tests.
+    """
+    actor = CurrentActorMocked()
+    monkeypatch.setattr(api, 'current_actor', actor)
+    return actor
+
+
+@pytest.fixture
+def context_mocked():
+    return IsolatedActionsMocked()
 
 
 def raise_call_error(args=None, exit_code=1):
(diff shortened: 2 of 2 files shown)


repos/system_upgrade/common/actors/scansubscriptionmanagerinfo/tests/test_scansubscriptionmanagementinfo.py
```diff
@@ -0,0 +1,349 @@
...
+from collections import namedtuple
+
+import pytest
+
+from leapp.libraries import stdlib
+from leapp.libraries.actor import scanrhsm
+from leapp.libraries.stdlib import api
+from leapp.libraries.common import mounting
+from leapp.libraries.common import repofileutils
+from leapp.libraries.common import rhsm
```
pirat89, line 12:
```suggestion
from leapp.libraries.common import (
    mounting,
    repofileutils,
    rhsm,
)
```

repos/system_upgrade/common/actors/scansubscriptionmanagerinfo/tests/test_scansubscriptionmanagementinfo.py
```diff
@@ -0,0 +1,349 @@
...
++-------------------------------------------+
+   System Status Details
++-------------------------------------------+
+Overall Status: Current
+
+System Purpose Status: Matched
+'''
+
+# Not the actual satelitte output (but it should suffice)
+RHSM_STATUS_OUTPUT_SATELLITE = '''
```
pirat89, line 32: This is about the SCA independent on satellite.
```suggestion
RHSM_STATUS_OUTPUT_SCA = '''
```

repos/system_upgrade/common/actors/scansubscriptionmanagerinfo/tests/test_scansubscriptionmanagementinfo.py
```diff
@@ -0,0 +1,349 @@
...
+    _, actor_products, external_command_runner = common_monkeypatched_handles
+
+    _original_get_env = rhsm.get_env
+
+    def mocked_get_env(env_variable, default):
+        if env_variable == 'LEAPP_NO_RHSM':
+            return '1'
+        return _original_get_env(env_variable, default)
+
+    monkeypatch.setattr(rhsm, 'get_env', mocked_get_env)
```
pirat89, line 177: replace by  `MockedCurrentActor` (see the `envars` parameter)

repos/system_upgrade/common/actors/scansubscriptionmanagerinfo/tests/test_scansubscriptionmanagementinfo.py
```diff
@@ -0,0 +1,349 @@
...
+        self.calls_performed_counter = 0
+
+    def add_mocked_command_with_stdout(self, command, stdout):
+        self.mocked_command_calls[command] = {'stdout': stdout}
+
+    def __call__(self, command, **kwargs):
+        # Ignore kwargs
+        default_result = {'stdout': ''}
+        self.calls_performed_counter += 1
+        return self.mocked_command_calls.get(tuple(command), default_result)
```
pirat89, line 76: Take baseically the current file and extend the `common/libraries/tests/test_rhsm.py` (see the `IsolatedActionsMocked`).

repos/system_upgrade/common/actors/scansubscriptionmanagerinfo/tests/test_scansubscriptionmanagementinfo.py
```diff
@@ -0,0 +1,349 @@
...
+
+    command_runner.add_mocked_command_with_stdout(RHSM_LIST_CONSUMED, 'SKU: 598339696910')
+
+    scanrhsm.scan()
+    assert command_runner.calls_performed_counter > 0, 'Some calls to subscription-manager were expected.'
+    assert actor_products.model_instances, 'The actor did not produce a message.'
+
+    produced_message = actor_products.model_instances[0]
+    assert produced_message.attached_skus, 'Actor failed to parse SKU.'
+    assert len(produced_message.attached_skus) == 1, \
```
pirat89, line 117: Do not use `\` to split/wrap....(you know)... lines.

repos/system_upgrade/common/actors/scansubscriptionmanagerinfo/tests/test_scansubscriptionmanagementinfo.py
```diff
@@ -0,0 +1,349 @@
...
+    """Verifies that the actor is able to correctly retrieve existing product certificates."""
+    _original_isdir = os.path.isdir
+
+    def mocked_isdir(path):
+        # Only return True for these two specific paths specific to this test
+        if path == '/etc/pki/product-default':
+            return True
+        if path == '/etc/pki/product':
+            return False
+        return _original_isdir(path)
```
pirat89, line 293: for others `_original_XX` - unit-tests should not work with the real system. The environment should be fully mocked... so unit-tests do not touch anything real in the environment...

pirat89 (changes requested):

pirat89:
This is very good set of tests to extend tests for the `rhsm.py` library. We discussed move & update of the tests to put them into `common/libraries/tests/test_rhsm.py` and test in the actor can be very minimalistic after that.

repos/system_upgrade/common/actors/scansubscriptionmanagerinfo/tests/test_scansubscriptionmanagementinfo.py
MichalHe: Suggested changes were rendered obsolete after moving the introduced tests into the library tests. However, the general idea of not repeating import paths was taken into account and applied when moving the tests.

repos/system_upgrade/common/actors/scansubscriptionmanagerinfo/tests/test_scansubscriptionmanagementinfo.py
MichalHe: Thanks for the tip :+1:  (new changes/commits already doesn't mock `get_env`).

repos/system_upgrade/common/actors/scansubscriptionmanagerinfo/tests/test_scansubscriptionmanagementinfo.py
MichalHe: Suggested changes were implemented in the most recent commits.

repos/system_upgrade/common/actors/scansubscriptionmanagerinfo/tests/test_scansubscriptionmanagementinfo.py
MichalHe: New version does not contain any usage of `\` to achieve multi-line asserts.

repos/system_upgrade/common/actors/scansubscriptionmanagerinfo/tests/test_scansubscriptionmanagementinfo.py
MichalHe: Valuable lesson learned :+1:. The newest version throws an exception when the module attempts to communicate with (file)system in a way that it is not supposed/predicted to.

repos/system_upgrade/common/libraries/tests/test_rhsm.py
```diff
@@ -1,29 +1,119 @@
...
+'''
+
+RHSM_STATUS_OUTPUT_SATELLITE = '''
++-------------------------------------------+
+   System Status Details
++-------------------------------------------+
+Overall Status: Current
+
+System Purpose Status: Not Specified
+'''
```
pirat89, line 42: I can have the vary same output on non-satellite subscriptions:
```
[root@localhost ~]# rhsm status
+-------------------------------------------+
   System Status Details
+-------------------------------------------+
Overall Status: Current

System Purpose Status: Not Specified
```

Difference between `RHSM_STATUS_OUTPUT_(NON)SATELLITE` outputs is just in the purpose of the system, which is set via the `syspurpose` command (see the [configuring system purpose](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/performing_an_advanced_rhel_installation/configuring-system-purpose-advanced_installing-rhel-as-an-experienced-user) article). What about this?
 - RHSM_STATUS_OUTPUT_NOSCA_NOPURPOSE
 - RHSM_STATUS_OUTPUT_NOSCA_PURPOSE

Or something shorter. I would just remove the satellite/non-satellite at all as I am able to get both outputs with & without satellite.

repos/system_upgrade/common/libraries/tests/test_rhsm.py
```diff
@@ -1,29 +1,119 @@
...
 from leapp.models import RepositoryFile, RepositoryData
 
 Repository = namedtuple('Repository', ['repoid', 'file'])
 LIST_SEPARATOR = '\n    - '
 
+# External commands called by the RHSM library
+RHSM_LIST_CONSUMED = ('subscription-manager', 'list', '--consumed')
+RHSM_STATUS = ('subscription-manager', 'status')
+RHSM_RELEASE = ('subscription-manager', 'release')
+RHSM_LIST_ENABLED_REPOS = ('subscription-manager', 'repos', '--list-enabled')
```
pirat89, line 24: just nitpick, what about  to refer to the command in the name of these constants, co it's clear from the name these are commands? e.g.

```suggestion
CMD_RHSM_LIST_CONSUMED = ('subscription-manager', 'list', '--consumed')
CMD_RHSM_STATUS = ('subscription-manager', 'status')
CMD_RHSM_RELEASE = ('subscription-manager', 'release')
CMD_RHSM_LIST_ENABLED_REPOS = ('subscription-manager', 'repos', '--list-enabled')
```

but I am ok with original names as well.

repos/system_upgrade/common/libraries/tests/test_rhsm.py
```diff
@@ -1,29 +1,119 @@
...
+
+    def full_path(self, path):
+        return path
+
+
+@pytest.fixture
+def actor_mocked(monkeypatch):
+    """Fixture providing a mocked actor that was already used to monkeypatch api.current_actor.
+    Introduced to reduce repetition inside tests.
+    """
```
pirat89, line 108:
```suggestion
    """
    Fixture providing a mocked actor that was already used to monkeypatch api.current_actor.
    
    Introduced to reduce repetition inside tests.
    """
```

We expect always one of these formats of docstrings
```
"""
title.

body
"""
```
or (without body)
```
"""
title.
"""
```
or
```
"""Title."""
```

repos/system_upgrade/common/libraries/tests/test_rhsm.py
```diff
@@ -119,3 +209,169 @@ def test_inhibit_on_duplicate_repos_no_dups(monkeypatch):
...
+
+def test_get_existing_product_certificates(monkeypatch, actor_mocked, context_mocked):
+    """Verifies that the library is able to correctly retrieve existing product certificates."""
+
+    def mocked_isdir(path):
+        # Only return True for these two specific paths specific to this test
+        if path == '/etc/pki/product-default':
+            return True
+        if path == '/etc/pki/product':
+            return False
```
pirat89, line 325: Comment says it should return `Tru`, but it returns `False` in the second case. Shouldn't it be rather (?):
```suggestion
        # Only return True for these two specific paths specific to this test
        if path in ['/etc/pki/product-default', '/etc/pki/product']:
            return True
```

Or fix the comment. Regarding the stuff in the bottom, I guess you wanted a different comment. However the setup like /etc/pki/product is file seems not reflecting the real world. But I am ok with the test too. However in such a case, I would expect even the case when both are dirs as in the real world, as this is the most common use case. Real scenarious are more valuable for us to test than very unexpected scenarious. Regarding amount of cases covered in this test function, a second one could be much more simple (e.g. expected two certs are discovered, one in product, one in product-default, no other problems).

repos/system_upgrade/common/libraries/tests/test_rhsm.py
```diff
@@ -119,3 +209,169 @@ def test_inhibit_on_duplicate_repos_no_dups(monkeypatch):
...
+        if path == '/etc/pki/product-default':
+            return False
+
+        # Handle imaginary files inside the '/etc/pki/product-default'
+        dirname = os.path.dirname(path)
+        if dirname == '/etc/pki/product-default':
+            filename = os.path.basename(path)
+            if filename in product_default_directory_contents['files']:
+                return True
+            return False
```
pirat89, line 357: Regarding the amount of the code, wouldn't be a better to have just a list of paths which returns true? E.g.
```
ISFILE_PATHS = [
    '/etc/pki/product',
    '/etc/pki/product-default/cert2',
    '/etc/pki/product-default/cert3',
]
if path in ISFILE_PATHS:
    return True
if path == '/etc/pki/product-default'
   return False
```
I guess we will not extend `product_default_directory_contents` significantly in future.

pirat89 (changes requested):
Some small changes yet, but it looks good otherwise.

repos/system_upgrade/common/libraries/tests/test_rhsm.py
MichalHe: Looking at it now, I absolutely agree that it is absolutely unnecessary to have satellite/non-satellite outputs. Removed them altogether and kept only `RHSM_STATUS_OUTPUT_SCA` and `RHSM_STATUS_OUTPUT_NOSCA`.

repos/system_upgrade/common/libraries/tests/test_rhsm.py
MichalHe: I feel that this would improve the readability of the code, therefore, I incorporated the suggested changes. Thanks!

repos/system_upgrade/common/libraries/tests/test_rhsm.py
MichalHe: Understood, fixed and absorbed, so it should not be a problem in the future.

repos/system_upgrade/common/libraries/tests/test_rhsm.py
MichalHe: Hmm... the entire `test_get_existing_product_certificates` is riddled with unnecessary logic. I guess I tried to be smart and test whether the library can cope with some edge case scenarios. Going to rewrite the test to contain a real world scenario that will be tested thoroughly.

MichalHe: Tests for the `get_existing_product_certificates` now consist of two tests - one thoroughly testing common scenario, one testing edge case when one of the cert directories is missing.

repos/system_upgrade/common/actors/scansubscriptionmanagerinfo/tests/test_scansubscriptionmanagementinfo.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+def mocked_get_rhsm_info(context):
+    assert context, 'The actor did not provide library with valid context.'
+    info = RHSMInfo()
+    info.attached_skus = ['SKU1', 'SKU2']
+    info.available_repos = ['Repo1', 'Repo2']
+    info.enabled_repos = ['Repo2']
+    info.release = '7.9'
+    info.existing_product_certificates.extend(['Cert1', 'Cert2', 'Cert3'])
```
pirat89, line 17: why extend when you set what should be inside? The assert below expects excatly the same list. The extend doesn't not make sense in such a case. If there is specific reason for the extend, just let me know.

```suggestion
    info.existing_product_certificates = ['Cert1', 'Cert2', 'Cert3']
```

pirat89 (changes requested):
seems good. one last change and please sqash all commits into one with a proper commit msg.

repos/system_upgrade/common/actors/scansubscriptionmanagerinfo/tests/test_scansubscriptionmanagementinfo.py
MichalHe: Sorry :confused:, I don't get it why did I write it this way and there is no reason why do not use direct assignment.

pirat89 (approved):
Prepared for the merge! Thanks for your work - let's be your first PR merged! :1st_place_medal:

PR #697, opened 2021-07-27
Fix the python workaround for IPU 8 -> 9

pirat89:
Some packages already use the `--keep-foreign` option for alternatives
in their scriptlets, some not. For these that handles alretnatives
correctly, we need to drop such a files from the alternatives command
executed in the actor.

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/pythonthreetmpworkaround/actor.py
+++ b/repos/system_upgrade/el8toel9/actors/pythonthreetmpworkaround/actor.py
@@ -30,14 +30,24 @@ class PythonThreeTmpWorkaround(Actor):
 
         cmd = [
             'alternatives', '--install', '/usr/bin/python3', 'python3', '/usr/bin/python3.9', '1000000',
-            '--slave', '/usr/share/man/man1/python3.1.gz', 'python3-man', '/usr/share/man/man1/python3.9.1.gz',
-            '--slave', '/usr/bin/pip3', 'pip3', '/usr/bin/pip3.9',
-            '--slave', '/usr/bin/pip-3', 'pip-3', '/usr/bin/pip-3.9',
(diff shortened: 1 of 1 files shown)


pirat89:
The crash without the fix looks like this:
```
[  966.399244] upgrade[40533]: 2021-07-27 09:08:09.77  DEBUG    PID: 39245 leapp.workflow.RPMUpgrade.pythonthreetmpworkaround: External command has started: ['alternatives', '--install', '/usr/bin/python3', 'python3', '/usr/bin/python3.9', '1000000', '--slave', '/usr/share/man/man1/python3.1.gz', 'python3-man', '/usr/share/man/man1/python3.9.1.gz', '--slave', '/usr/bin/pip3', 'pip3', '/usr/bin/pip3.9', '--slave', '/usr/bin/pip-3', 'pip-3', '/usr/bin/pip-3.9', '--slave', '/usr/bin/easy_install-3', 'easy_install-3', '/usr/bin/easy_install-3.9', '--slave', '/usr/bin/pydoc3', 'pydoc3', '/usr/bin/pydoc3.9', '--slave', '/usr/bin/pydoc-3', 'pydoc-3', '/usr/bin/pydoc3.9', '--slave', '/usr/bin/pyvenv-3', 'pyvenv-3', '/usr/bin/pyvenv-3.9']
[  966.652295] systemd-journald[225]: Compressed data object 716 -> 300 using ZSTD
[  966.469736] upgrade[40533]: failed to link /usr/bin/pip3 -> /etc/alternatives/pip3: /usr/bin/pip3 exists and it is either not a symlink or --keep-foreign was set and link points outside /etc/alternatives
```

See especially the last line. This fix drops all slaves that exist.

vinzenz (approved):

pirat89:
Manually tested it works again.

PR #698, opened 2021-07-30
Fix copying of files into the target userspace (preupgrade task)

pirat89:
The original implementation copied the files into the wrong
container (using the context of the first container that is
used to created the target userspace environement).

Setup the right context for the _copy_files() function so the
files are copyied into the target userspace container
as expected.


Rezney (approved):
Looks good to me. Tested internally.

PR #699, opened 2021-07-30
checkrhui: Produce TargetUserSpacePreupgradeTasks + refactoring

pirat89:
The RequiredTargetUserspacePackages model has been deprecated
in favor of the TargetUserSpacePreupgradeTasks model. Keeping
the original functionality + produce the new msg as well.

Additionally the actor is a little bit refactored.

### DIFF
--- a/repos/system_upgrade/common/actors/cloud/checkrhui/actor.py
+++ b/repos/system_upgrade/common/actors/cloud/checkrhui/actor.py
@@ -1,19 +1,21 @@
+from leapp import reporting
 from leapp.actors import Actor
-from leapp.libraries.common.rpms import has_package
+from leapp.libraries.common import rhsm, rhui, rpms
 from leapp.models import (
     DNFPluginTask,
     InstalledRPM,
     KernelCmdlineArg,
     RHUIInfo,
-    RequiredTargetUserspacePackages,
+    RequiredTargetUserspacePackages,  # deprecated
     RpmTransactionTasks,
+    TargetUserSpacePreupgradeTasks,
 )
 from leapp.reporting import Report, create_report
-from leapp import reporting
 from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
-from leapp.libraries.common import rhsm, rhui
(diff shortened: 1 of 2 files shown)


MichalHe (approved):
Looks OK to me.

repos/system_upgrade/common/actors/cloud/checkrhui/actor.py
```diff
@@ -68,8 +72,13 @@ def process(self):
                     self.produce(DNFPluginTask(name='amazon-id', disable_in=['upgrade']))
                 # if RHEL7 and RHEL8 packages differ, we cannot rely on simply updating them
                 if info['el7_pkg'] != info['el8_pkg']:
-                    self.produce(RpmTransactionTasks(to_install=[info['el8_pkg']]))
-                    self.produce(RpmTransactionTasks(to_remove=[info['el7_pkg']]))
-                self.produce(RHUIInfo(provider=provider))
-                self.produce(RequiredTargetUserspacePackages(packages=[info['el8_pkg']]))
+                    self.produce(RpmTransactionTasks(
+                        to_install=[info['el8_pkg']],
```
shaded-enmity, line 76: Since this is in `common`, do we want to keep `el8_pkg` and `el7_pkg`? Renaming those to `old_pkg` and `new_pkg` would make much more sense in `common` context.

pirat89: @shaded-enmity Good point. We could define the new/old variables in the begining of the begining and than just use these variables everywhere to work with the map in the share library.

pirat89:
@MichalHe you can ignore this one for now. we know it's need to be rebased but  currently this is postponed for the month after the release.

pirat89:
@fernflower not sure whether the RHUI actors have been already refactored. this should really happen still if they are not using new up-to-date standards instead of deprecated functionality. let's keep it for the next release, where I would like to propose to drop some deprecated functionalities to clean the code.

pirat89:
Closing in favor of PR #1057

PR #700, opened 2021-08-02
Inhibit the upgrade on multiple Zipl rescue boot entries.

MichalHe:
When multiple rescue boot entries are present in the Zipl configuration the system fails boot when rebooting after `leapp upgrade`. This PR adds the necessary infrastructure that deals with the boot entries and inhibits if multiple zipl rescue boot entries are found.

[RHBZ #1783542](https://bugzilla.redhat.com/show_bug.cgi?id=1783542)

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/ziplcheckmultiplerescueentries/libraries/ziplcheckmultiplerescueentries.py
+++ b/repos/system_upgrade/el7toel8/actors/ziplcheckmultiplerescueentries/libraries/ziplcheckmultiplerescueentries.py
@@ -1,0 +1,45 @@
+from leapp import reporting
+from leapp.libraries.common.config import architecture
+
+FMT_LIST_SEPARATOR = '\n    - '
+
+
+def inhibit_if_multiple_zipl_rescue_entries_present(bootloader_config):
+    """
+    Inhibits the upgrade if we are running on s390x and the bootloader configuration
+    contains multiple rescue boot entries.
+
+    A boot entry is recognized as a rescue entry when its title contains the `rescue` substring.
+
+    :param SourceBootloaderConfiguration bootloader_config: The configuration of the source boot loader.
+    """
+
+    if not architecture.matches_architecture(architecture.ARCH_S390X):
+        # Zipl is used only on s390x
+        return
+
+    # Keep the whole information about boot entries not just their count as
+    # we want to provide user with the details
+    rescue_entries = []
+    for boot_entry in bootloader_config.entries:
+        if 'rescue' in boot_entry.title.lower():
+            rescue_entries.append(boot_entry)
+
+    if len(rescue_entries) > 1:
+        # Prepare the list of available rescue entries for user
+        rescue_entries_text = ''
+        for rescue_entry in rescue_entries:
+            rescue_entries_text += '{0}{1}'.format(FMT_LIST_SEPARATOR, rescue_entry.title)
+        zipl_config_path = '/etc/zipl.conf'
+
+        reporting.create_report([
+            reporting.Title('Multiple rescue boot entries present in the bootloader configuration.'),
+            reporting.Summary(
+                'The zipl configuration file {0} contains multiple rescue boot entries:{1}'
+                .format(zipl_config_path, rescue_entries_text)
+            ),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Tags([reporting.Tags.BOOT]),
+            reporting.Remediation(hint='Remove rescue boot entries from the configuration and leave just one.'),
+            reporting.Flags([reporting.Flags.INHIBITOR])
+        ])
--- a/repos/system_upgrade/el7toel8/actors/ziplcheckmultiplerescueentries/tests/test_ziplcheckmultiplerescueentries.py
+++ b/repos/system_upgrade/el7toel8/actors/ziplcheckmultiplerescueentries/tests/test_ziplcheckmultiplerescueentries.py
@@ -1,0 +1,75 @@
+import pytest
+
+from leapp import reporting
+from leapp.libraries.actor import ziplcheckmultiplerescueentries
+from leapp.libraries.actor.ziplcheckmultiplerescueentries import inhibit_if_multiple_zipl_rescue_entries_present
+from leapp.libraries.common.testutils import create_report_mocked, CurrentActorMocked
+from leapp.libraries.common.config import architecture
+from leapp.libraries.stdlib import api
+from leapp.models import BootEntry, SourceBootLoaderConfiguration
+from leapp.snactor.fixture import current_actor_context
+
+
+def test_inhibition_multiple_rescue_entries_present(monkeypatch):
+    """Tests whether the upgrade process is inhibited when multiple rescue boot entries are present."""
+    mocked_report = create_report_mocked()
+    monkeypatch.setattr(architecture, 'matches_architecture', lambda dummy: True)
+    monkeypatch.setattr(reporting, 'create_report', mocked_report)
+
+    boot_entries = [
+        BootEntry(title='entry_1'),
+        BootEntry(title='entry_1_Rescue'),
+        BootEntry(title='entry_2'),
+        BootEntry(title='entry_2_rescue-ver2.3'),  # Typically is the `rescue` substring surrounded
+    ]
+
+    inhibit_if_multiple_zipl_rescue_entries_present(SourceBootLoaderConfiguration(entries=boot_entries))
+
+    assert mocked_report.called, 'Report should be created when multiple rescue entries are present.'
+
+    fail_description = 'The correct rescue entries are not present in the report summary.'
+    report_summary = mocked_report.report_fields['summary']
+    for expected_rescue_entry in ['entry_1_Rescue', 'entry_2_rescue-ver2.3']:
+        assert expected_rescue_entry in report_summary, fail_description
+
+    fail_description = 'Upgrade should be inhibited on multiple rescue entries.'
+    assert 'inhibitor' in mocked_report.report_fields['flags'], fail_description
+
+
+def test_inhibition_multiple_rescue_entries_not_present(monkeypatch):
+    """Tests whether the upgrade process is not inhibited when multiple rescue boot entries are not present."""
+    monkeypatch.setattr(architecture, 'matches_architecture', lambda dummy: True)
+    monkeypatch.setattr(reporting, 'create_report', create_report_mocked())
+
+    boot_entries = [
+        BootEntry(title='entry_1'),
+        BootEntry(title='entry_2'),
+        BootEntry(title='entry_2_rescue-ver2.3'),
+    ]
+
+    inhibit_if_multiple_zipl_rescue_entries_present(SourceBootLoaderConfiguration(entries=boot_entries))
+
+    assert not reporting.create_report.called, 'Report was created, even if multiple rescue entries were not present.'
+
+
+@pytest.mark.parametrize(
+    ('arch',),
+    [(arch,) for arch in architecture.ARCH_SUPPORTED]
+)
+def test_checks_performed_only_on_s390x_arch(arch, monkeypatch):
+    """Tests whether the actor doesn't perform different architectures than s390x."""
+
+    should_perform = False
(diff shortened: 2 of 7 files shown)


repos/system_upgrade/el7toel8/actors/bootentriesscanner/actor.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class Bootentriesscanner(Actor):
+    """
+    Scans the bootloader configuration for available boot entries.
+    """
+
+    name = 'boot_entries_scanner'
+    consumes = ()
+    produces = (BootEntry, )
```
vinzenz, line 14:
```suggestion
    produces = (BootEntry,)
```

repos/system_upgrade/el7toel8/actors/bootentriesscanner/tests/test_bootentryscanner.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+
+from leapp.libraries import stdlib
+from leapp.libraries.actor import bootentriesscanner
+from leapp.libraries.common.testutils import produce_mocked
+
+
+GRUBBY_INFO_ALL_STDOUT = '''
```
vinzenz, line 8: Grubby output really starts with an empty line?

repos/system_upgrade/el7toel8/actors/ziplcheckmultiplerescueentries/libraries/ziplcheckmultiplerescueentries.py
```diff
@@ -0,0 +1,36 @@
...
+
+        reporting.create_report([
+            reporting.Title('Multiple rescue boot entries present in the bootloader configuration.'),
+            reporting.Summary(
+                'The zipl configuration file {0} contains multiple rescue boot entries:{1}'
+                .format(zipl_config_path, rescue_entries_text)
+            ),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Tags([reporting.Tags.BOOT]),
+            reporting.Remediation(hint='Remove rescue boot entries from the configuration and leave just one.'),
```
vinzenz, line 34: It would be good, if we could link to a KB (knowledge base) article where it is described how to remove an entry.

repos/system_upgrade/el7toel8/actors/bootentriesscanner/actor.py
MichalHe: This was implemented in the new version. Thanks!

repos/system_upgrade/el7toel8/actors/bootentriesscanner/tests/test_bootentryscanner.py
MichalHe: Nice catch (although I should have noticed it myself - lesson learned), thank you.

MichalHe:
Tested with 4e18364 and the VM provided by bug the reporter:
Machine RHEL ALT 7.6 on s390x:
- [x] `leapp upgrade` fails and boots into dracut console when testing with current master 
- [x] boot entry detection works (tested with `snactor  run bootentriesscanner --print-output`) 
- [x] upgrade process gets inhibited should there be more than one rescue entries
- [x] after removal of the extra rescue entries the `leapp upgrade` does not fail and boots normally to the system

repos/system_upgrade/el7toel8/actors/ziplcheckmultiplerescueentries/tests/test_ziplcheckmultiplerescueentries.py
```diff
@@ -0,0 +1,69 @@
...
+def test_inhibition_multiple_rescue_entries_present(monkeypatch):
+    """Tests whether the upgrade process is inhibited when multiple rescue boot entries are present."""
+    mocked_report = create_report_mocked()
+    monkeypatch.setattr(reporting, 'create_report', mocked_report)
+
+    boot_entries = [
+        BootEntry(title='entry_1'),
+        BootEntry(title='entry_1_Rescue'),
+        BootEntry(title='entry_2'),
+        BootEntry(title='entry_2_rescue-ver2.3'),  # Typically the `rescue` substring surrounded
```
vinzenz, line 22:
```suggestion
        BootEntry(title='entry_2_rescue-ver2.3'),  # Typically is the `rescue` substring surrounded
```

MichalHe:
Performed final testing before submitting for a code review:
Using the VM linked in the issue ticket (RHEL ALT 7.6), the problematic system state was easily reproduced by repeating the steps described in the ticket/bugzilla. Leapp (2271b75) recognized multiple rescue entries as an inhibitor. After removing the extra rescue entries the upgrade process finished successfully, without any errors.

repos/system_upgrade/el7toel8/actors/bootentriesscanner/libraries/bootentriesscanner.py
```diff
@@ -0,0 +1,47 @@
...
+        return text[1:-1]
+    return text
+
+
+def scan_boot_entries():
+    """
+    Scans the available boot entries are informs other actors about the findings
+    via a :class:`BootEntry` message.
+    """
+    grubby_output = run(CMD_GRUBBY_INFO_ALL, split=True)
```
vinzenz, line 30: I think you want to check that this is not failing by catching the `leapp.libraries.stdlib.CalledProcessError` exception.
Since you didn't specify otherwise, `run`  assumed checked=True which is fine. However you need to handle the exception or, if grubby for whatever reason fails, you will be causing an error in the upgrade and it will exit after finishing the current phase.

repos/system_upgrade/el7toel8/actors/bootentriesscanner/libraries/bootentriesscanner.py
```diff
@@ -0,0 +1,47 @@
...
+    for output_line in grubby_output['stdout']:
+        # For now it is sufficient to look only for the titles as that is the only
+        # information we use. If need be, we would have to parse the structure
+        # of the grubby output into sections according to the `index` lines
+        if output_line.startswith('title='):
+            boot_entry = output_line[6:]  # Remove the `title=` prefix
+
+            # On s390x grubby produces quotes only when needed (whitespace in
+            # the configuration values), on x86 the values are quoted either way
+            boot_entry = remove_quotes_if_used(boot_entry)
```
vinzenz, line 43:
```suggestion
            boot_entry = boot_entry.strip('\'"')
```

vinzenz: I don't think that you want any quotes around it

MichalHe: TIL: U can specify characters to be stripped via an optional argument. 
Awesome! Thank you.

repos/system_upgrade/el7toel8/actors/bootentriesscanner/libraries/bootentriesscanner.py
MichalHe: Just a quick question (so I know how to approach similar matters in the future):
- on `x86` and others - failing to call `grubby`, and therefore, failing to retrieve a list of boot entries is not that big of a problem, as there are no problems known with rescue entries etc. My proposed solution would be to log an error (and probably inform the user via report too) and move on.
- on `s390x` however, we know that without knowing what boot entries there are, the system may end up unable to boot. I suppose, that we definitely want to inform the user about this via a report entry. Just to be sure - should we also preemptively inhibit the upgrade? (I would not say so.)

The lesson I would like to take from this as a novice: In a scenario like this (when some error means that we cannot guarantee a safe upgrade process, but also strictly doesn't mean trouble with the IPU) what stand should I take?

MichalHe: The error is handled in the latest commit. Thank you.

repos/system_upgrade/el7toel8/actors/bootentriesscanner/libraries/bootentriesscanner.py
MichalHe: Fixed in the latest commit.

repos/system_upgrade/el7toel8/actors/ziplcheckmultiplerescueentries/tests/test_ziplcheckmultiplerescueentries.py
```diff
@@ -0,0 +1,69 @@
...
+
+    assert not reporting.create_report.called, 'Report was created, even if multiple rescue entries were not present.'
+
+
+def test_actor_does_nothing_on_non_s390x_architecture(monkeypatch, current_actor_context):
+    """Tests whether the actor doesn't perform different architectures than s390x."""
+    def matches_architecture_mocked(arch):
+        if arch == architecture.ARCH_X86_64:
+            return True
+        return False
```
pirat89, line 58: The test will pass only on intel arch. In case someone execute it on a non-intel arch, test is broken. In this case, the correct way of testing would be telling what system the actor is expected to be executed during the test. E.g. tell the actor the system is s390x and see the result. Of course for such a test you cannot use the component test (`current_actor_context`), but you have to write the unit test instead.

pirat89 (changes requested):

repos/system_upgrade/el7toel8/models/bootentry.py
```diff
@@ -0,0 +1,8 @@
+from leapp.models import Model, fields
+from leapp.topics import SystemInfoTopic
+
+
+class BootEntry(Model):
+    topic = SystemInfoTopic
+
+    title = fields.String()
```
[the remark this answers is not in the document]
pirat89: (and of course, docstrings! :))

repos/system_upgrade/el7toel8/actors/ziplcheckmultiplerescueentries/libraries/ziplcheckmultiplerescueentries.py
MichalHe: We've discussed this internally - a new ticket documenting this issue has been created.

repos/system_upgrade/el7toel8/actors/ziplcheckmultiplerescueentries/tests/test_ziplcheckmultiplerescueentries.py
MichalHe: This is fixed in the latest version of this PR.

repos/system_upgrade/el7toel8/models/bootentry.py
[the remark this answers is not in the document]
MichalHe: The requested changes are implemented in the latest version of the PR.

repos/system_upgrade/el7toel8/actors/ziplcheckmultiplerescueentries/tests/test_ziplcheckmultiplerescueentries.py
```diff
@@ -0,0 +1,75 @@
...
+    [(arch,) for arch in architecture.ARCH_SUPPORTED]
+)
+def test_checks_performed_only_on_s390x_arch(arch, monkeypatch):
+    """Tests whether the actor doesn't perform different architectures than s390x."""
+
+    should_perform = False
+    if arch == architecture.ARCH_S390X:  # Rescue entries should be checked only on s390x.
+        should_perform = True
+
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(arch))
```
pirat89, line 66:
```suggestion
    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(arch=arch))
```

Use please keywords for arguments of CurrentActor mocked - it's more safe against possible changes in the class.

MichalHe: Understood, fixed in the latest version of the PR.

pirat89 (approved):
Nice! Let's fun begin :) Pushed for testing.

pirat89:
Verified by @smitterl . Merging

PR #701, opened 2021-08-02
Major packaging changes: leapp-repository renaming + leapp version locking

pirat89:
The leapp-repository rpm is going to be renamed based on the system
it is installed on:
  - RHEL 7  -> leapp-upgrade-el7toel8
  - RHEL 8  -> leapp-upgrade-el8toel9

The original naming has not been clear to people and we wanted
provide clearly what is the purpose of the RPM. The new naming
schema is:
   - **leapp-{purpose}-{main_reponame}**

In this case, purpose of our repositories is now just IPU between
major versions of RHEL (so the leapp-upgrade part). For IPU 7 > 8 the
main leapp repository is el7toel8, ...

((As well this naming syncs our project with [convert2rhel](https://github.com/oamg/convert2rhel/) (hi guys!)
where the `leapp-convert` name will be used in future when they
switch to leapp.))

Additionally set conflicts between:
  - leapp-upgrade-el7toel8
  - leapp-upgrade-el8toel9

packages to ensure they are not replaced during the upgrade as we
are super interested about keeping the installed leapp rpms untouched
during the upgrade process (excluding deps metapackages) to be sure
we finish the upgrade with the same RPMs as we started with.

And use the `setup` macro instead of `autosetup`. We do not expect
to apply any paches in the upstream and automatic application of patches
has been fragile in downstream when we played around.

IMPORTANT NOTES FOR TESTING
Regarding the set Obsoletes and Provides, you cannot downgrade nor upgrade to packages for this PR from the master builds. That's because of the the set to limited version-release. After the merge, this should work normally as expected in future. This is to prevent automatic replacement of upstream packages by these new from this PR before the merged. To be able to test the behaviour, install packages from older PRs (or any PR with builds older then for this one) and then you can try the experience to update. But the downgrade is basically not possible now. Or you can remove leapp packages and then install new one using dnf install leapp-upgrade.


pirat89:
Prepared for the review. I am keeping it in the draft state to ensure we do not merge it before the testing tools and CI is adapted to the changes.

pirat89:
@oamg/developers We should probably create an actor that informs users about expected manual action after the upgrade & that we changed their configuration of the packager to exclude leapp rpms from any transaction.

pirat89:
I have realized one error. I have forgotten to set the changes on the host so packages could be replaced e.g. during the first boot phase running the `dnf update` cmd. Output from the upgraded system:

```
[root@localhost ~]# uname -a
Linux localhost.localdomain 4.18.0-305.el8.x86_64 #1 SMP Thu Apr 29 08:54:30 EDT 2021 x86_64 x86_64 x86_64 GNU/Linux
[root@localhost ~]# rpm -qa | grep leapp
python2-leapp-0.12.1-0.202108021531Z.b0e30cb.packaging_pt2.PR727.el7.noarch
leapp-0.12.1-0.202108021531Z.b0e30cb.packaging_pt2.PR727.el7.noarch
leapp-upgrade-el7toel8-0.14.0-0.202108031919Z.1f6897b0.new_packaging.PR701.el7.noarch
leapp-deps-el8-5.0.8-0.202108031919Z.1f6897b0.new_packaging.PR701.el8.noarch
leapp-repository-deps-el8-5.0.8-0.202108031919Z.1f6897b0.new_packaging.PR701.el8.noarch
[root@localhost ~]# dnf config-manager --dump | grep exclude
disable_excludes = 
exclude = 
excludepkgs = 
```

The actor performing the upgrade should call the functioned from the dfconfig library. that should resolve the problem.

pirat89:
The issues is fixed. The change is now persistent after the upgrade as well:
```
# uname -a
Linux localhost.localdomain 4.18.0-305.12.1.el8_4.x86_64 #1 SMP Mon Jul 26 08:06:24 EDT 2021 x86_64 x86_64 x86_64 GNU/Linux
# grep exclude /etc/dnf/dnf.conf 
exclude=python2-leapp,snactor,leapp-upgrade-el7toel8,leapp
```

pirat89:
@oamg/developers  prepared for the review.

packaging/leapp-el7toel8-deps.spec
```diff
@@ -2,7 +2,14 @@
 %global  lrdname  leapp-repository-deps-el%{rhel}
 %global  ldname   leapp-deps-el%{rhel}
 
-%define leapp_repo_deps  5
+%if 0%{?rhel} == 8
+    %define lpr_name_src leapp-upgrade-el7toel8-deps
```
MichalHe, line 6: If I interpret the intention correctly, shouldn't there be (if we are running on RHEL7, we want the 7->8 path and its dependencies):
```suggestion
%if 0%{?rhel} == 7
    %define lpr_name_src leapp-upgrade-el7toel8-deps
```
Similar check as suggested is performed in `leapp-repository.spec`.

pirat89: @MichalHe no, these packages are expected to be created for the **target** system. The deps packages for the system where the leapp / leapp-repository packages are part of the *main* spec files. It's better described in the doc here: https://leapp.readthedocs.io/en/latest/dependencies-leapp-repository.html
(the doc is expected to be updated consequently)

repos/system_upgrade/common/libraries/dnfconfig.py
```diff
@@ -0,0 +1,109 @@
...
+        raise StopActorExecutionError(
+            message='Invalid DNF configuration data (missing [main])',
+            details=data,
+        )
+
+    output_data = {}
+    for line in data[main_start:]:
+        if '=' not in line:
+            # TODO: could we skip this check completely?
+            continue
```
MichalHe, line 62: I did a little research:
1. The `config-manager` plugin dumps only the contents of `[main]` section when no repos supplied on command line. The code path that handles `--dump` is here: [DNF plugin code](https://github.com/rpm-software-management/dnf-plugins-core/blob/master/plugins/config_manager.py#L147).
2. The plugins uses `self.base.conf.dump()` to display the contents of the section ([link to implementation](https://github.com/rpm-software-management/dnf/blob/master/dnf/conf/config.py#L161)). The implementation outputs lines with `Key=Value` format (the `=` is always present), therefore, we should not need to check it.

pirat89: I've changed the implementation a little bit. Despite the current DNF implementation, I am worried the print could be in future changed and we could miss that. So instead of if-continue, added try/except block with a warning and comment.

PR #703, opened 2021-08-06
Add detection of enabled yum plugins.

MichalHe:
The IPU requires `subscription-manager` and `product-id` yum plugins to be enabled in order to proceed without errors about no enabled target repositories. This PR adds the necessary infrastructure for the detection and reporting when the required plugins are not being loaded by yum.

The introduced changes consist of:
- add a new model `YumConfig` that carries the information about detected `enabled_plugins`,
- add a new actor `YumConfigScanner` that takes care of extracting the current yum configuration from the system,
- add a new actor `CheckYumPluginsEnabled` that checks that the plugins are loaded (available), and if not, informs the user about this fact, along with information where to enable them.

Relates: https://bugzilla.redhat.com/show_bug.cgi?id=1913829

### DIFF
--- a/repos/system_upgrade/common/actors/checkyumpluginsenabled/libraries/checkyumpluginsenabled.py
+++ b/repos/system_upgrade/common/actors/checkyumpluginsenabled/libraries/checkyumpluginsenabled.py
@@ -1,0 +1,72 @@
+import os
+
+from leapp import reporting
+from leapp.libraries.common.config.version import get_source_major_version
+from leapp.libraries.common.rhsm import skip_rhsm
+
+# If LEAPP_NO_RHSM is set, subscription-manager and product-id will not be
+# considered as required when checking whether the required plugins are enabled.
+REQUIRED_YUM_PLUGINS = {'subscription-manager', 'product-id'}
+FMT_LIST_SEPARATOR = '\n    - '
+
+
+def check_required_yum_plugins_enabled(yum_config):
+    """
+    Checks whether the yum plugins required by the IPU are enabled.
+
+    If they are not enabled, a report is produced informing the user about it.
+
+    :param yum_config: YumConfig
+    """
+
+    missing_required_plugins = REQUIRED_YUM_PLUGINS - set(yum_config.enabled_plugins)
+
+    if skip_rhsm():
+        missing_required_plugins -= {'subscription-manager', 'product-id'}
+
+    if missing_required_plugins:
+        missing_required_plugins_text = ''
+        for missing_plugin in missing_required_plugins:
+            missing_required_plugins_text += '{0}{1}'.format(FMT_LIST_SEPARATOR, missing_plugin)
+
+        if get_source_major_version() == '7':
+            pkg_manager = 'YUM'
+            pkg_manager_config_path = '/etc/yum.conf'
+            plugin_configs_dir = '/etc/yum/pluginconf.d'
+        else:
+            # On RHEL8+ the yum package might not be installed
+            pkg_manager = 'DNF'
+            pkg_manager_config_path = '/etc/dnf/dnf.conf'
+            plugin_configs_dir = '/etc/dnf/plugins'
+
+        # pkg_manager_config_path - enable/disable plugins globally
+        # subscription_manager_plugin_conf, product_id_plugin_conf - plugins can be disabled individually
+        subscription_manager_plugin_conf = os.path.join(plugin_configs_dir, 'subscription-manager.conf')
+        product_id_plugin_conf = os.path.join(plugin_configs_dir, 'product-id.conf')
+
+        remediation_commands = [
+            'sed -i \'s/^plugins=0/plugins=1/\' \'{0}\''.format(pkg_manager_config_path),
+            'sed -i \'s/^enabled=0/enabled=1/\' \'{0}\''.format(subscription_manager_plugin_conf),
+            'sed -i \'s/^enabled=0/enabled=1/\' \'{0}\''.format(product_id_plugin_conf)
+        ]
+
+        reporting.create_report([
+            reporting.Title('Required {0} plugins are not being loaded.'.format(pkg_manager)),
+            reporting.Summary(
+                'The following {0} plugins are not being loaded: {1}'.format(pkg_manager,
+                                                                             missing_required_plugins_text)
+            ),
+            reporting.Remediation(
+                hint='If you have yum plugins globally disabled, please enable them by editing the {0}. '
+                     'Individually, the {1} plugins can be enabled in their corresponding configurations found at: {2}'
+                     .format(pkg_manager_config_path, pkg_manager, plugin_configs_dir),
+                # Provide all commands as one due to problems with satellites
+                commands=[['bash', '-c', '"{0}"'.format('; '.join(remediation_commands))]]
+            ),
+            reporting.RelatedResource('file', pkg_manager_config_path),
+            reporting.RelatedResource('file', subscription_manager_plugin_conf),
+            reporting.RelatedResource('file', product_id_plugin_conf),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Flags([reporting.Flags.INHIBITOR]),
+            reporting.Tags([reporting.Tags.REPOSITORY]),
+        ])
--- a/repos/system_upgrade/common/actors/checkyumpluginsenabled/tests/test_checkyumpluginsenabled.py
+++ b/repos/system_upgrade/common/actors/checkyumpluginsenabled/tests/test_checkyumpluginsenabled.py
@@ -1,0 +1,39 @@
+import pytest
+
+from leapp import reporting
+from leapp.libraries.actor.checkyumpluginsenabled import check_required_yum_plugins_enabled
+from leapp.libraries.common.testutils import create_report_mocked, CurrentActorMocked
+from leapp.libraries.stdlib import api
+from leapp.models import YumConfig
+
+
+def test_report_when_missing_required_plugins(monkeypatch):
+    """Test whether a report entry is created when any of the required YUM plugins are missing."""
+    yum_config = YumConfig(enabled_plugins=['product-id', 'some-user-plugin'])
+
+    actor_reports = create_report_mocked()
+
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked())
+    monkeypatch.setattr(reporting, 'create_report', actor_reports)
+
+    check_required_yum_plugins_enabled(yum_config)
+
+    assert actor_reports.called, 'Report wasn\'t created when required a plugin is missing.'
+
+    fail_description = 'The missing required plugin is not mentioned in the report.'
+    assert 'subscription-manager' in actor_reports.report_fields['summary'], fail_description
+
+    fail_description = 'The upgrade should be inhibited when plugins are not enabled.'
+    assert 'inhibitor' in actor_reports.report_fields['flags'], fail_description
+
+
+def test_nothing_is_reported_when_rhsm_disabled(monkeypatch):
+    actor_mocked = CurrentActorMocked(envars={'LEAPP_NO_RHSM': '1'})
+
+    monkeypatch.setattr(api, 'current_actor', actor_mocked)
+    monkeypatch.setattr(reporting, 'create_report', create_report_mocked())
+
+    yum_config = YumConfig(enabled_plugins=[])
+    check_required_yum_plugins_enabled(yum_config)
+
+    assert not reporting.create_report.called, 'Report was created even if LEAPP_NO_RHSM was set'
--- a/repos/system_upgrade/common/actors/yumconfigscanner/libraries/yumconfigscanner.py
+++ b/repos/system_upgrade/common/actors/yumconfigscanner/libraries/yumconfigscanner.py
@@ -1,0 +1,82 @@
+import re
+
+from leapp.libraries.common.config.version import get_source_major_version
+from leapp.libraries.stdlib import api, run
+from leapp.models import YumConfig
+
+# When the output spans multiple lines, each of the lines after the first one
+# start with a '    <SPACES>    : '
+YUM_LOADED_PLUGINS_NEXT_LINE_START = ' +: '
+
+
+def _parse_loaded_plugins(yum_output):
+    """
+    Retrieves a list of plugins that are being loaded when calling yum.
+
+    :param dict yum_output: The result of running the yum command.
+    :rtype: list
+    :returns: A list of plugins that are being loaded when calling yum.
+    """
+    # YUM might break the information about loaded plugins into multiple lines,
+    # we need to concaternate the list ourselves
+    loaded_plugins_str = ''
+    for line in yum_output['stdout']:
+        if line.startswith('Loaded plugins:'):
+            # We have found the first line that contains the plugins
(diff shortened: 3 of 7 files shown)


MichalHe:
Implemented features tested with 16f40b4, with the following results
System RHEL7.9, `subscription-manager` configured not to auto-enable plugins:
Scenario | Report entry created | Remediation command works
------------ | --------------------------- | ----------------------------------------
All plugins disabled (/etc/yum.conf:plugins=0) | Yes, reported `subscription-manager`, `product-id` not being loaded | yes
Only `subscription-manager` disabled (/etc/yum/pluginconf.d/subscription-manager.conf:enabled=0) | Yes, reported `subscription-manager` not being loaded | yes
None of the required plugins disabled | No | Not applicable

repos/system_upgrade/common/actors/checkyumpluginsenabled/libraries/checkyumpluginsenabled.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    Checks whether the yum plugins required by the IPU are enabled.
+
+    If they are not enabled, a report is produced informing the user about it.
+    Does not inhibit the IPU.
+
+    :param yum_config: YumConfig
+    """
+
+    if skip_rhsm():
```
vinzenz, line 18: Just a comment on this line: While currently true, that if skip_rhsm() is true we don't need to do any checks, we might want to have it remove the plugins from the required list if skip_rhsm is true instead. But right now I think this might be acceptable as is.

repos/system_upgrade/common/actors/checkyumpluginsenabled/libraries/checkyumpluginsenabled.py
```diff
@@ -0,0 +1,50 @@
+from leapp import reporting
+from leapp.libraries.common.rhsm import skip_rhsm
+
+REQUIRED_YUM_PLUGINS = ['subscription-manager', 'product-id']
```
vinzenz, line 4: Make it set right away, there's no need to convert it later 
```suggestion
REQUIRED_YUM_PLUGINS = {'subscription-manager', 'product-id'}
```

repos/system_upgrade/common/actors/checkyumpluginsenabled/libraries/checkyumpluginsenabled.py
```diff
@@ -0,0 +1,50 @@
...
+    If they are not enabled, a report is produced informing the user about it.
+    Does not inhibit the IPU.
+
+    :param yum_config: YumConfig
+    """
+
+    if skip_rhsm():
+        return
+
+    missing_required_plugins = set(REQUIRED_YUM_PLUGINS) - set(yum_config.enabled_plugins)
```
vinzenz, line 21: In case you apply the comment about making `REQUIRED_YUM_PLUGINS` a set already.
```suggestion
    missing_required_plugins = REQUIRED_YUM_PLUGINS - set(yum_config.enabled_plugins)
```

repos/system_upgrade/common/actors/yumconfigscanner/libraries/yumconfigscanner.py
```diff
@@ -0,0 +1,45 @@
...
+
+    # We rely on yum itself to report what plugins are used when it is invoked.
+    # An alternative approach would be to check /usr/lib/yum-plugins/ (install
+    # path for yum plugins) and parse corresponding configurations from
+    # /etc/yum/pluginconf.d/
+
+    if get_source_major_version() == '7':
+        yum_cmd = ['yum']
+    else:
+        yum_cmd = ['yum', '-v']  # On RHEL8 we need to supply an extra switch
```
vinzenz, line 22: Since you know it's RHEL8+
```suggestion
        yum_cmd = ['dnf', '-v']  # On RHEL8 we need to supply an extra switch
```

repos/system_upgrade/common/actors/yumconfigscanner/tests/test_yumconfigscanner.py
```diff
@@ -0,0 +1,42 @@
...
+CMD_YUM_OUTPUT = '''Loaded plugins: langpacks, my plugin, subscription-manager, product-id
+Usage: yum [options] COMMAND
+'''
+
+
+@pytest.mark.parametrize(
+    ('source_major_version', 'yum_command'),
+    [
+        ('7', ['yum']),
+        ('8', ['yum', '-v']),
```
vinzenz, line 14:
```suggestion
        ('8', ['dnf', '-v']),
```

MichalHe: Applied in the latest version. Thanks.

repos/system_upgrade/common/actors/yumconfigscanner/libraries/yumconfigscanner.py
MichalHe: Applied in the latest version. Thanks.

repos/system_upgrade/common/actors/checkyumpluginsenabled/libraries/checkyumpluginsenabled.py
MichalHe: Applied in the latest version. Thank you.

repos/system_upgrade/common/actors/checkyumpluginsenabled/libraries/checkyumpluginsenabled.py
MichalHe: Implemented in the most recent commit.

repos/system_upgrade/common/actors/checkyumpluginsenabled/libraries/checkyumpluginsenabled.py
MichalHe: Thanks for the info. I've changed in in such a way that the currently required plugins just get removed from the set of required plugins if `LEAPP_NO_RHSM` is set. 

The change was motivated by the fact that although the old implementation works, it does not really reflects the intention - when the `LEAPP_NO_RHSM` is set that means we should not use the `subscription-manager`. Since we are not supposed to be using it, we should not require its yum plugins. The old implementation suggested that when when we are not using `subscription-manager`, we do not require _any_ plugins which is wrong from the semantic point of view.

Really appreciate the comment that enabled me to see it in this fashion. Thanks!

repos/system_upgrade/common/actors/checkyumpluginsenabled/tests/test_checkyumpluginsenabled.py
```diff
@@ -0,0 +1,36 @@
...
+
+from leapp import reporting
+from leapp.libraries.actor.checkyumpluginsenabled import check_required_yum_plugins_enabled
+from leapp.libraries.common.testutils import create_report_mocked, CurrentActorMocked
+from leapp.libraries.stdlib import api
+from leapp.models import YumConfig
+
+
+def test_report_when_missing_required_plugins(monkeypatch):
+    """Test whether a report entry is created when some of YUM required plugins are missing."""
```
pirat89, line 11:
```suggestion
    """Test whether a report entry is created when any required YUM plugins are missing."""
```

repos/system_upgrade/common/actors/yumconfigscanner/libraries/yumconfigscanner.py
```diff
@@ -0,0 +1,45 @@
...
+        yum_cmd = ['yum']
+    else:
+        yum_cmd = ['dnf', '-v']  # On RHEL8 we need to supply an extra switch
+
+    yum_output = run(yum_cmd, split=True, checked=False)  # The yum command will certainly fail (does not matter).
+
+    loaded_plugins = []
+    # Find the line that lists all the loaded plugins
+    for line in yum_output['stdout']:
+        if line.startswith('Loaded plugins:'):
```
pirat89, line 29: this could be broken as well. The  list could be split into the several lines so detection will not be possibly correct (see the [rhbz#1964311](https://bugzilla.redhat.com/show_bug.cgi?id=1964311) ticket where people reported this problem). I am not sure whether it happens only in the interactive console or always. But we should definitely find a time to check this behaviour to be sure. If this can happen during the upgrade, we need to find a more robust solution. E.g. check whether additional lines start with "\s+:\s" and in such a case process sequence of all these lines.
E.g.
```
# yum 2>&1|head -2
Loaded plugins: enabled_repos_upload, package_upload, product-id, search-
              : disabled-repos, subscription-manager
```
Pay the special attention to the split of the `search-disabled-repos` plugin.

pirat89 (changes requested):
This is partial review as I have to look at additional stuff now. I am sorry I was not able to finish it completely, but at least some stuff to work on for now.

repos/system_upgrade/common/actors/yumconfigscanner/libraries/yumconfigscanner.py
MichalHe: After a little bit of investigation about at what conditions precisely can the list of plugins be broken into multiple lines:
The `YUM` code responsible for handling this is [here](https://github.com/rpm-software-management/yum/blob/4ed25525ee4781907bd204018c27f44948ed83fe/yum/plugins.py#L222) (same code as present on RHEL7). To perform the formatting of the output they use the `textwrap.fill` function from **stdlib**. They do not modify the default values of the parameters responsible for configuring the text breaking, therefore:
- `break_long_words = True` implies that should some plugin have longer name than `termwidth` - some formatting prefix, the plugin name gets broken. 
- `break_on_hyphens = True` allows plugin names to be broken on hyphens (with hyphen being kept at the end of the line)
Furthermore we know, that the list of plugins is always separated by a `,`
Therefore, the parsing can be done as suggested above (checking for presence `\s+: `) to get the block of yum plugins, recreate the string containing the plugins (remove newlines and formatting prefixes, and concatenating the lines together) and breaking the result on `, `.

MichalHe: This should be addressed in the latest version of this PR.

repos/system_upgrade/common/actors/checkyumpluginsenabled/libraries/checkyumpluginsenabled.py
```diff
@@ -0,0 +1,53 @@
...
+                'The following YUM plugins are not being loaded: {0}'.format(missing_required_plugins_text)
+            ),
+            reporting.Remediation(
+                hint='If you have yum plugins globally disabled, please enable them by editing the {0}. '
+                     'Individually, the YUM plugins can be enabled in their corresponding configurations found at: {1}'
+                     .format(yum_config_path, plugin_configs_dir),
+                commands=[
+                    ['sed', '-i', '\'s/^plugins=0/plugins=1/\'', '/etc/yum.conf'],
+                    ['sed', '-i', '\'s/^enabled=0/enabled=1/\'', '/etc/yum/pluginconf.d/subscription-manager.conf'],
+                    ['sed', '-i', '\'s/^enabled=0/enabled=1/\'', '/etc/yum/pluginconf.d/product-id.conf']]
```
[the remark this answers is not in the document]
MichalHe: The requested changes are incorporated in the latest version of this PR.

pirat89 (approved):
Great work! Lgtm. Moving to verification.

pirat89 (changes requested):
This could be broken on RHEL 8 if `yum` is not installed. Discussed in person, the code is going to be updated.

MichalHe:
I've tested the changes with the following findings.
### RHEL7 (team vagrant), `/etc/yum.conf:plugins=0` (plugins globally disabled)
IPU inhibited with inhibitor:
```
Title: Required YUM plugins are not being loaded.
Summary: The following YUM plugins are not being loaded:
    - product-id
    - subscription-manager
Remediation: [hint] If you have yum plugins globally disabled, please enable them by editing the /etc/yum.conf. Individually, the YUM plugins can be enabled in their corresponding configurations found at: /etc/yum/pluginconf.d                                                       
[command] bash -c sed -i 's/^plugins=0/plugins=1/' '/etc/yum.conf'; sed -i 's/^enabled=0/enabled=1/' '/etc/yum/pluginconf.d/subscription-manager.conf'; sed -i 's/^enabled=0/enabled=1/' '/etc/yum/pluginconf.d/product-id.conf'
```
### RHEL8 `/etc/dnf/dnf.conf:plugins=0` (plugins globally disabled)
```
Title: Required DNF plugins are not being loaded.
Summary: The following DNF plugins are not being loaded: 
    - subscription-manager
    - product-id
Remediation: [hint] If you have yum plugins globally disabled, please enable them by editing the /etc/dnf/dnf.conf. Individually, the DNF plugins can be enabled in their corresponding configurations found at: /etc/dnf/plugins
[command] bash -c sed -i 's/^plugins=0/plugins=1/' '/etc/dnf/dnf.conf'; sed -i 's/^enabled=0/enabled=1/' '/etc/dnf/plugins/subscription-manager.conf'; sed -i 's/^enabled=0/enabled=1/' '/etc/dnf/plugins/product-id.conf"
```
I've verified that the referenced files exists, however the bash command is missing quotes around the sed commands (fixed by the most recent force push).

MichalHe:
Output after the missing quotes around provided script:
```
Risk Factor: medium (inhibitor)
Title: Required DNF plugins are not being loaded.
Summary: The following DNF plugins are not being loaded: 
    - subscription-manager
    - product-id
Remediation: [hint] If you have yum plugins globally disabled, please enable them by editing the /etc/dnf/dnf.conf. Individually, the DNF plugins can be enabled in their corresponding configurations found at: /etc/dnf/plugins
[command] bash -c "sed -i 's/^plugins=0/plugins=1/' '/etc/dnf/dnf.conf'; sed -i 's/^enabled=0/enabled=1/' '/etc/dnf/plugins/subscription-manager.conf'; sed -i 's/^enabled=0/enabled=1/' '/etc/dnf/plugins/product-id.conf'"
```
The given command works (enabled required plugins).

pirat89 (approved):
Good work!!

repos/system_upgrade/common/actors/checkyumpluginsenabled/libraries/checkyumpluginsenabled.py
```diff
@@ -0,0 +1,72 @@
...
+            pkg_manager_config_path = '/etc/yum.conf'
+            plugin_configs_dir = '/etc/yum/pluginconf.d'
+        else:
+            # On RHEL8+ the yum package might not be installed
+            pkg_manager = 'DNF'
+            pkg_manager_config_path = '/etc/dnf/dnf.conf'
+            plugin_configs_dir = '/etc/dnf/plugins'
+
+        # pkg_manager_config_path(plugins=1) - Allows YUM to load plugins
+        # subscription_manager_plugin_conf, product_id_plugin_conf - Plugins can be enabled plugins individually
```
fernflower, line 43: Hmm this comment is a bit cryptic imho, especially the 'Plugins can be enabled plugins'. Could you please rephrase that?

MichalHe: Thanks for catching this, the comment was indeed cryptic. I rephrased this (latest push).

pirat89:
done

PR #668, opened 2021-05-18
Split (move) generic stuff for upgrades to the upgrade/common repo from el7toel8

pirat89:
This is PoC right now. With still a lot of work that needs to be done. See the `splitcomms` file for some notes we (@Rezney ) made. This is even not done yet. Some crucial topics needs to be resolved (e.g. what to do with IPUConfig and various constants in shared libraries).

Currently we want to have inside the common repo just content that is going to be really used in multiple upgrades. Additional content can be moved to the common repo later when needed. As well, we need to sync with several teams to check whether some content, that we keep inside the common repo, should be kept specific just for particular upgrade.

This is preparation step to be able to introduce the `el8toel9` leapp repository.

Sorry for making the msg less informative. I will polish it when switching from the draft to PR.

PR #704, opened 2021-08-10
PR668 Follow up.

MichalHe:
This PR implements noncritical changes suggested in code review to the #668 as wells as some minor refactoring (e.g. docs fixes) found when doing additional pass through the code.

### DIFF
--- a/repos/system_upgrade/common/libraries/config/version.py
+++ b/repos/system_upgrade/common/libraries/config/version.py
@@ -18,12 +18,45 @@ _SUPPORTED_VERSIONS = {
 }
 
 
+def get_major_version(version):
+    """
+    Return the major version from the given version string.
+
+    Versioning schema: MAJOR.MINOR.PATCH
+    It doesn't matter how many dots are present. Everything until the first dot is returned. E.g.:
+       8.1.0 => 8
+       7.9   => 7
+       7     => 7
+
+    :param str version: The version string according to the versioning schema described.
+    :rtype: str
+    :returns: The major version from the given version string.
+    """
+    return version.split('.')[0]
+
+
 def get_source_major_version():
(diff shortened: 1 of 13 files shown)


repos/system_upgrade/common/libraries/config/version.py
```diff
@@ -18,12 +18,16 @@
...
+
+
 def get_source_major_version():
-    return api.current_actor().configuration.version.source.split('.')[0]
+    return get_major_version(api.current_actor().configuration.version.source)
 
 
 def get_target_major_version():
-    return api.current_actor().configuration.version.target.split('.')[0]
+    return get_major_version(api.current_actor().configuration.version.target)
```
pirat89, line 30: Just a nitpick. I've just realized we have forgotten to add the docstrings to the functions. Would you mind to add them? So people will understand for sure what is the major version. e.g.

```python
def get_major_version(version):
    """
    Return the major version from the string given string.

    Versioning schema: MAJOR.MINOR.PATCH
    It doesn't matter how many dots are present. Everything until the first dot is return. E.g.:
       8.1.0 => 8
       7.9    => 7
       7       => 7
    """
    return version.split('.')[0]


def get_source_major_version():
    """
    Return the major version of the source (original) system
    """
    return get_major_version(api.current_actor().configuration.version.source)


def get_target_major_version():
    """
    Return the major version of the target system
    """
    return get_major_version(api.current_actor().configuration.version.target)
```

pirat89: It seems ok otherwise.

pirat89 (approved):
lgtm

repos/system_upgrade/common/libraries/config/version.py
pirat89: Thanks :) Approved.

PR #708, opened 2021-08-20
An attempt to support leapp-repository->leapp dependencies

fernflower:
This should teach github action to trigger relevant leapp copr build
if a specific leapp pr number is specified as part of rerun comment.

For example, '/rerun 42' should trigger testing farm service testing for
the leapp-repository version from the pr together with leapp version from
pr 42.

https://issues.redhat.com/browse/OAMG-4988

### DIFF
--- a/.github/workflows/tmt-tests.yml
+++ b/.github/workflows/tmt-tests.yml
@@ -11,7 +11,7 @@ jobs:
     name: Run tmt tests on Testing Farm service
     if: |
       github.event.issue.pull_request
-      && github.event.comment.body == '/rerun'
+      && startsWith(github.event.comment.body, '/rerun')
       && contains(fromJson('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.comment.author_association)
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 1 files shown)


fernflower:
@vojtechsokol Please review. The logs of test runs for (/rerun 686) and (/rerun) are in JIRA task and my fork.

vojtechsokol (commented):
I think it would be clearer (DRY) to not duplicate `sched_test` and `comment` steps and just modify the already existing steps. E.g. the for the `sched_test` step leave the `request.json` file as it is and add something like
```sh
jq '.environments[0].artifacts += [{"type": "fedora-copr-build", "id": "lepp_copr_id"}]' < request.json
```
and something similar for the `comment` step.

Also `request.json` in `sched_test_leapp` step omits `.environments.variables` list, is this intentional, and if yes, what is the reason behind that?

fernflower:
@vojtechsokol I don't need any environment vars to be passed to the playbook. I wanted to go on and remove environment vars for the request completely, but then it seemed as unrelated change so I went for omitting env vars only in newly written code instead.
Env vars were introduced for the sole purpose of possible embedding the leapp dependencies logic in the playbook, once we have real leapp/leapp-repository deps support we don't need them anymore.

fernflower:
@vojtechsokol Fixed, test run on my fork passed https://github.com/fernflower/leapp-repository/pull/1#issuecomment-916479331

vojtechsokol:
Just a nitpick - the steps that are repeated for leapp and leapp repository (`checkout`\\`checkout_leapp`, `ref_sha`\\`ref_sha_leapp`, `copr_build`\\`copr_build_leapp`, `link_copr`\\`link_copr_leapp`) have the same name, which makes it a bit confusing when looking at the [log](https://github.com/fernflower/leapp-repository/runs/3461823748?check_suite_focus=true). If I wasn't aware of what's going on under the hood, I would be asking why some of the steps are run twice.

Other than that LGTM.

vojtechsokol (approved):

PR #712, opened 2021-08-26
Improve NFS check

Rezney:
We cannot check just for an "nfs" string as there can also be "nfs3" and "nfs4*".


MichalHe (approved):
Looks great to me :+1:.

Rezney:
@mkluson I guess we should be good to go here with just unit tests covering this. WDYT?

mkluson:
Hi @Rezney, we can consider this solely upstream change, so unit tests will be sufficient.

PR #716, opened 2021-09-09
checkcpu: RHEL-9 minimal supported IBM Z system is z14

sharkcz:
Documentation reference will be updated once it will exist.


pirat89 (commented):
@sharkcz add pls the fixme comment so we will improve our chances we do not forget to update the link later :-)

```
FIXME(dhorak): update the URL to the document once it exists
```

pirat89 (approved):

PR #717, opened 2021-09-09
fetch_or_read: Allow to specify what encoding is the file using.

MichalHe:
The requests library relies on Content-Encoding header to determine what
the encoding of the response is. As the response is compressed, the
actual encoding of the uncompressed received data is unknown, so it relies on
heuristics to determine the actual encoding. This patch prevents
the library from guessing and enforces the response content to be
decoded as utf-8 text.

### DIFF
--- a/repos/system_upgrade/common/actors/repositoriesmapping/tests/unit_test_repositoriesmapping.py
+++ b/repos/system_upgrade/common/actors/repositoriesmapping/tests/unit_test_repositoriesmapping.py
@@ -112,25 +112,25 @@ def test_scan_valid_file_with_comments(monkeypatch, arch, src_type, dst_type):
 
 
 class ResponseMocked(object):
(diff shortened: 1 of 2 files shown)


pirat89:
seems good. I will run yet QE tests on this and we can merge that.

pirat89 (approved):

pirat89:
P.S. the original issue nowadays is not reproducible on RHEL 8 python3.6  - regarding the python3 rpm changelog, the problem in `json.loads` has been probably resolved as a side-effect of a different bug. However we still want to merge this PR to be sure we are not longer affected at all.

pirat89:
Seems everything OK. Only failed tests are regarding the test utility which cannot install leapp (maybe because of the changed naming schema? which is expected and we are ok to go with it).

PR #720, opened 2021-09-15
WIP: Set releasever in /etc to the target version

Rezney:
Set target release version in cases where the "/etc/{yum,dnf}/var/releasever" file exists.

If the file exists, whatever is in there, is attached into repository URL. In case the file is
empty, an empty string is attached to the URL causing it to be invalid, also when it is set to
the source version (e.g 7.7), the URL on the target system is again invalid. In order to avoid
such scenario, we popullate the file with the target system version whenever it exists.


Rezney:
@pirat89 For the moment, it is without tests + ignore the unfinished "create_report". I did not re-test after I did some changes but it should be ready.

repos/system_upgrade/common/actors/setetcreleasever/actor.py
```diff
@@ -0,0 +1,29 @@
...
+    empty, an empty string is attached to the URL causing it to be invalid, also when it is set to
+    the source version (e.g 7.7), the URL on the target system is again invalid. In order to avoid
+    such scenario, we popullate the file with the target system version whenever it exists.
+
+    """
+
+    name = 'set_etc_releasever'
+    consumes = (PackageManagerInfo,)
+    produces = ()
+    tags = (IPUWorkflowTag, FirstBootPhaseTag)
```
pirat89, line 21: These actions should be done during the application phase `ApplicationsPhaseTag`. There is no reason why it should happen so late (we do not need an internet connection to do it). The firstbootphase should be used only by actors that needs a booted target OS (or internet connection nowadays).

```suggestion
    tags = (IPUWorkflowTag, ApplicationsPhaseTag)
```

repos/system_upgrade/common/actors/setetcreleasever/libraries/setetcreleasever.py
```diff
@@ -0,0 +1,22 @@
+from leapp.libraries.stdlib import api
+from leapp.libraries.common.config.version import get_source_major_version
+
+
+def _get_default_pkg_manager():
+    return "yum" if get_source_major_version() == '7' else "dnf"
+
+
+def _set_releasever(releasever):
+    releasever_path = '/etc/{}/vars/releasever'.format(_get_default_pkg_manager())
```
pirat89, line 10: no need to guess whether yum or dnf is used. The target OS has always DNF. Btw, it could be defined as constant. As well `yum` paths are symlinked to dnf.

```suggestion
def _set_releasever(releasever):
    releasever_path = '/etc/dnf/vars/releasever'
```

pirat89 (commented):
Some changes should be done. But it looks ok otherwise.

repos/system_upgrade/common/actors/checkpkgmanager/actor.py
```diff
@@ -0,0 +1,21 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import checkpackagemanager
+from leapp.models import Report, PackageManagerInfo
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckPackageManager(Actor):
+    """ Check package manager info in order to report a message or inhibit the upgrade when needed """
```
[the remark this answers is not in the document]
Rezney: I was commenting both, today's and potentially future use cases but I can change it.

repos/system_upgrade/common/actors/setetcreleasever/actor.py
Rezney: Fixed.

repos/system_upgrade/common/actors/setetcreleasever/actor.py
```diff
@@ -0,0 +1,29 @@
...
+
+class SetEtcReleaseVer(Actor):
+    """
+    Set target release version in cases where the "/etc/{yum,dnf}/var/releasever" file exists
+
+    If the file exists, whatever is in there, is attached into repository URL. In case the file is
+    empty, an empty string is attached to the URL causing it to be invalid, also when it is set to
+    the source version (e.g 7.7), the URL on the target system is again invalid. In order to avoid
+    such scenario, we popullate the file with the target system version whenever it exists.
+
```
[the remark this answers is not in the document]
Rezney: Nice improvement. Thanks.

Rezney:
@pirat89 this should be ready.

repos/system_upgrade/common/actors/scanpkgmanager/libraries/scanpkgmanager.py
```diff
@@ -0,0 +1,32 @@
...
+from leapp.libraries.common.config.version import get_source_major_version
+
+
+def _get_releasever_path():
+    default_manager = 'yum' if get_source_major_version() == '7' else 'dnf'
+    return '/etc/{}/vars/releasever'.format(default_manager)
+
+
+def _releasever_exists(releasever_path):
+    return bool(os.path.isfile(releasever_path))
```
pirat89, line 14: The conversion to bool is not necessary as function returns bool.
```suggestion
    return os.path.isfile(releasever_path)
```

repos/system_upgrade/common/actors/setetcreleasever/actor.py
```diff
@@ -0,0 +1,32 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import setetcreleasever
+from leapp.models import PkgManagerInfo
+from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
+
+
+class SetEtcReleaseVer(Actor):
+    """
+    Set target release version in case the "/etc/{yum,dnf}/var/releasever" file exists
```
pirat89, line 9:
```
"""
One line title

body multiline
```

repos/system_upgrade/common/models/packagemanagerinfo.py
```diff
@@ -0,0 +1,14 @@
...
+class PkgManagerInfo(Model):
+    """
+    Package manager (yum/dnf) related info
+
+    We expect to have only one single message of this kind produced
+    """
+    topic = SystemInfoTopic
+    # separate field for "/etc/{yum,dnf}/vars/releasever" file as we also care about its existence,
+    # otherwise the info could have been taken from global configuration.
+    etc_releasever = fields.Nullable(fields.String())
```
pirat89, line 14:
```suggestion
    # separate field for "/etc/{yum,dnf}/vars/releasever" file as we also care about its existence,
    # otherwise the info could have been taken from global configuration.
    etc_releasever = fields.Nullable(fields.String())
    """
    Contain the first line of /etc/{yum,dnf}/vars/releasever file or None if the file does not exist.
    
    In case the value is empty string, it means the file exists but it is empty. In such a case the original configuration
    seems obviously broken.
    """
```

repos/system_upgrade/common/actors/scanpkgmanager/libraries/scanpkgmanager.py
Rezney: Ouch :D This happens when you are just blindly following linter...

repos/system_upgrade/common/actors/setetcreleasever/libraries/setetcreleasever.py
```diff
@@ -0,0 +1,17 @@
...
+    with open(releasever_path, 'w') as fo:
+        fo.write(releasever)
+
+
+def process(facts):
+    target_ver = api.current_actor().configuration.version.target
+
+    # if "/etc/{yum,dnf}/vars/releasever" file exists, let's set it to our
+    # target version.
+    if facts.etc_releasever:
```
pirat89, line 16: The comment is not accurate. To make the comment accurate, the code would need to look like this:
```python
    if facts.etc_releasever is not None:
```

currently it does not do the change even when the file exists but it is an empty (which points to a broken configuration) as in such a case the value contains empty string, so nothing happens. From what I understand from our discussion, I guess in such a case we want to set the value as well.

PR #721, opened 2021-09-16
Respect LEAPP_NO_NETWORK_RENAMING

fernflower:
This patch disable network interface renaming in case an
environment variable was specified.

OAMG-4323


pirat89 (dismissed):
lgtm. I've been thinking about complete skip (putting the check to the beginning of the function) but this is ok as well.

repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/libraries/persistentnetnamesconfig.py
```diff
@@ -64,7 +64,7 @@ def process():
                 )
                 continue
 
-            if rhel7_name != rhel8_name:
+            if rhel7_name != rhel8_name and not os.getenv('LEAPP_NO_NETWORK_RENAMING'):
```
shaded-enmity, line 67: Is this gonna work? This actor executes in the `ApplicationPhase` so unless we're forwarding environment variables to the initrd the original environment settings would have been already lost.

Rezney: @shaded-enmity @fernflower 

We can use this one (it is not only about DEVEL env vars nowadays so we may want to change the docstring):

https://github.com/oamg/leapp-repository/blob/22061422b8fc300b4a4c64a1c82b898cd441e6ca/repos/system_upgrade/common/actors/ipuworkflowconfig/libraries/ipuworkflowconfig.py#L45

Rezney: It is then accessed through the Leapp configuration api rather then consuming a message.

E.g.

`env_vars = self.configuration.leapp_env_vars`

pirat89: Aaah. You are right guys. I completely overlooked this. Thanks for another eyes :)

shaded-enmity: Let's also do some checking on the actual value, let's say the variable needs to be set `LEAPP_NO_NETWORK_RENAMING=1` to make it very clear what is the expected outcome. I can imagine people experimenting with various values `LEAPP_NO_NETWORK_RENAMING` and scratching their heads because _any_ value would trigger that code path.

pirat89: ok, so let's do this:
```
if get_env('LEAPP_NO_NETWORK_RENAMING', '0') == '1':
```

repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/libraries/persistentnetnamesconfig.py
```diff
@@ -2,6 +2,7 @@
 import os
 import re
 
+import leapp.libraries.common.config as leapp_config
```
pirat89, line 5: just one point, if I remember well, we told we do not want people to use "renaming" of libs for imports (unless you need it to resolve duplicate names).

fernflower: Should be fixed

repos/system_upgrade/el7toel8/actors/persistentnetnamesconfig/libraries/persistentnetnamesconfig.py
fernflower: Done

pirat89 (approved):
lgtm. waiting for the verification.

PR #722, opened 2021-09-17
Introduce fast_lint make target.

MichalHe:
Our `Makefile` currently contains the `lint` target that executes the linter on the entire repository which takes a considerable amount of time. In order to improve developer experience via significantly shortening the time required to get linting results on the code they introduced in their PR a new target has been added in this PR to the `Makefile`.

The `fast_lint` target runs the linter only on files that have been subject to changes compared to the master branch.  Moreover the linter is executed with the option enabling parallel execution to further decrease the feedback time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -266,6 +266,23 @@ test_no_lint:
 
 test: lint test_no_lint
 
+fast_lint:
+	@. $(VENVNAME)/bin/activate; \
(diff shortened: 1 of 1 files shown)


shaded-enmity:
`grep '\.py'` to prevent grep using the dot as wildcard for any character

MichalHe:
The issue with dot not being escaped and thus matching any character which was pointed out by @shaded-enmity has been fixed with 1188e9e. Thanks for pointing it out.

pirat89 (approved):

PR #723, opened 2021-09-17
Silence some linter checks

fernflower:
- consider-using-f-string check ignored as is incompatible with py2;
- consider-using-in seems worth keeping, so did minor refactoring to
respect it.


fernflower:
@vojtechsokol Don't quite understand the failure here (tested locally it works). 
Why is there an issue with container run `docker build -t leapp-tests -f utils/docker-tests/Dockerfile utils/docker-tests && PYTHON_VENV=python3.6  docker run --rm -ti -v ${PWD}:/payload --env=PYTHON_VENV leapp-tests`? 
It behaves as if PYTHON_VENV variable wasn't set..

pirat89 (approved):

PR #724, opened 2021-09-21
upgrade: Restore leapp environment variables on --resume

vinzenz:
This patch will restore leapp environment variables from the IPUConfig
message. This will allow to easier and more naturally access environment
from actors during the upgrade even after reboot etc.

### DIFF
--- a/commands/upgrade/util.py
+++ b/commands/upgrade/util.py
@@ -8,9 +8,19 @@ from datetime import datetime
 from leapp.config import get_config
 from leapp.exceptions import CommandError
 from leapp.repository.scan import find_and_scan_repositories
-from leapp.utils.audit import get_connection, get_checkpoints
+from leapp.utils.audit import get_connection, get_checkpoints, get_messages
 from leapp.utils.output import report_unsupported
 from leapp.utils.report import fetch_upgrade_report_messages, generate_report_file
+
+
+def restore_leapp_env_vars(context):
+    """
+    Restores leapp environment variables from the `IPUConfig` message.
+    """
+    messages = get_messages(('IPUConfig',), context)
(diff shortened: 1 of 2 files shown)


commands/upgrade/util.py
```diff
@@ -8,11 +8,21 @@
...
 from leapp.utils.output import report_unsupported
 from leapp.utils.report import fetch_upgrade_report_messages, generate_report_file
 
 
+def restore_leapp_env_vars(context):
+    """
+    Restores leapp environment variables from the `IPUConfig` message.
+    """
+    messages = get_messages(('IPUConfig',), context)
+    leapp_env_vars = json.loads((messages or [{}])[0].get('message', {}).get('data', '{}')).get('leapp_env_vars', {})
```
MichalHe, line 21:
```suggestion
    leapp_env_vars = json.loads((messages or [{}])[0].get('message', {}).get('data', {})).get('leapp_env_vars', {})
```
In case the dictionary has no `data` field  the default value is provided which is a string. String has no `get` method, therefore, it would result in a crash.

vinzenz: @MichalHe  Look again :-)

MichalHe: Oh :eyes: :sweat: sorry, I've did not correctly parse where the parenthesis end.

MichalHe:
Manually tested with the `LEAPP_OVL_SIZE` environment variable used as a test variable. The variable value gets correctly injected back to the environment when running `leapp --resume`.

MichalHe (approved):
lgtm.

pirat89 (approved):
lgtm and otuput from @MichalHe  looks good too

PR #727, opened 2021-09-25
Improve "btrfs" actor's remediation message

Rezney:
Improve "btrfs" actor's remediation message


repos/system_upgrade/el7toel8/actors/checkbtrfs/actor.py
```diff
@@ -19,6 +19,14 @@ class CheckBtrfs(Actor):
     tags = (ChecksPhaseTag, IPUWorkflowTag)
 
     def process(self):
+
+        hint = (
+            'In order to unload the module from the running system, check the accompanied command. '
+            'To prevent a module from being loaded altogether, you may consult the following KCS solution: '
+            'https://access.redhat.com/solutions/41278.'
```
vinzenz, line 26: I don't think you want the link here and have it as 'External Link' Also the period interferes with the link

vinzenz (changes requested):

repos/system_upgrade/el7toel8/actors/checkbtrfs/actor.py
pirat89: however, people reading just txt versions do not see any links - these are visible only in JSON, cockpit, and satellite right now. probably we could finally change that

Rezney: > however, people reading just txt versions do not see any links - these are visible only in JSON, cockpit, and satellite right now. probably we could finally change that

yep, that was my motivation for doing it like that...

vinzenz: The right approach for me would be to generate the text version with URLs then and not duplicate it here

Rezney: We discussed it with @pirat89 offline and we will remove the link part for now and open a ticket for improving the text part.

PR #728, opened 2021-09-27
[WIP] Add the --channel cli option for target channel specification.

MichalHe:
This PR adds the `--channel` option for a user to specify preferred target channel.

Depends on oamg/leapp#686 which enables command option to specify a list of valid choices for its value. 

TODO:
- [ ] rebase against master once repomapping is merged
- [ ] fix the guards checking the _LEAPP\_DEVEL\_TARGET\_PRODUCT\_TYPE_ in the userspacegen library.
- [ ] add docs to tests


repos/system_upgrade/common/libraries/config/__init__.py
```diff
@@ -43,3 +48,44 @@ def get_product_type(sys_type):
...
+
+    :param default: Value to be returned if no target product type has been specified when running leapp.
+    :type default: str
+    :returns: The user-specified target channel or default if no channel was specified.
+    :rtype: str
+    """
+
+    devel_target_product_type = get_env('LEAPP_DEVEL_TARGET_PRODUCT_TYPE')
+    if devel_target_product_type:
+        supported_devel_target_channels = SUPPORTED_TARGET_CHANNELS + {'beta'}
```
Rezney, line 76:
```python
[root@leapp-20211005121234 ~]# python
Python 2.7.5 (default, Mar 20 2020, 17:08:22) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-39)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> SUPPORTED_TARGET_CHANNELS = {'ga', 'tuv', 'e4s', 'eus', 'aus'}
>>> 
>>> SUPPORTED_TARGET_CHANNELS + {'beta'}
TypeError: unsupported operand type(s) for +: 'set' and 'set'
>>> 
>>> 
```

Rezney (changes requested):

repos/system_upgrade/common/libraries/config/__init__.py
vinzenz, line 76:
```suggestion
        supported_devel_target_channels = SUPPORTED_TARGET_CHANNELS | {'beta'}
```

repos/system_upgrade/common/actors/peseventsscanner/libraries/peseventsscanner_repomap.py
```diff
@@ -0,0 +1,300 @@
...
+        self.repositories = repo_map.repositories
+        self.mapping = repo_map.mapping
+        # FIXME(pstodulk): what about default_channel -> fallback_channel
+        # hardcoded always as ga? instead of list of channels..
+        # it'd be possibly confusing naming now...
+        self.default_channels = default_channels or ['ga']
+
+        # Make self.prio_channel None if the user did not specify any target channels, so that self.default_channels
+        # will be used instead
+        self.prio_channel = prio_channel or get_target_product_channel(default=None)
```
vinzenz, line 52: Somehow I don't understand why the `prio_channel` parameter exists, it's not used anywhere and even if there is a reason for it, why it supersedes the thing provided on the command line or environment variable. This seems counter intuitive for me. If we specify the `prio_channel` parameter to be 'ga' anywhere in the code, then a user won't be able to override that anymore. Is that intended?

repos/system_upgrade/common/libraries/config/__init__.py
```diff
@@ -43,3 +48,44 @@ def get_product_type(sys_type):
...
+                                  .format(channel_specified_via, actual_value, ", ".join(supported_values)))
+
+
+def get_target_product_channel(default='ga'):
+    """
+    Get target product channel specified when running the IPU or default if no channel was specified.
+
+    The target channel can be specified via:
+        - Using the environment variable LEAPP_DEVEL_TARGET_PRODUCT_TYPE (devel variable with higher priority than
+        other any way of specifying target channel).
```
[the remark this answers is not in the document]
MichalHe: Many thanks for reviewing, fixed.

repos/system_upgrade/common/libraries/config/__init__.py
MichalHe: Many thanks for reviewing :+1:, fixed.

MichalHe: Many thanks for reviewing +1, fixed.

repos/system_upgrade/common/actors/peseventsscanner/libraries/peseventsscanner_repomap.py
MichalHe: The parameter was present in the original new repomapping implementation, so I let it there even when it was clear that the desired target channel will be taken from environmental vars, thinking it would be cool if the actors that do want to use the library have easy way to override it if the need it.

I really appreciate you raising this concern -  it made me rethink what we are trying achieve in the bigger picture (we want the `src_repoid -> dst_repoid` translation to be consistent across all actors needing it), and this hypothetically useful parameter conflicts it. Therefore, I will remove it. Thanks :+1:.

Rezney:
@MichalHe you are free to rebase against the merged repomapping...

PR #729, opened 2021-10-04
actor: Check the existence of the custom network-scripts

liangwen12year:
RHEL 9 does not support the legacy network-scripts package that was
deprecated in RHEL 8. As a result, if the local scripts of network
scripts are detected on the system, user should move these custom
scripts to the user local directory (e.g. /opt or /usr/local/sbin) in
order to proceed with the upgrade into RHEL 9.

Signed-off-by: Wen Liang <liangwen12year@gmail.com>

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/actor.py
+++ b/repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/actor.py
@@ -1,0 +1,18 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import customnetworkscripts
+from leapp.models import Report
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+
+
+class CheckCustomNetworkScripts(Actor):
+    """
+    Check the existence of custom network-scripts and warn user about possible
+    manual intervention requirements.
+    """
+
+    name = "check_custom_network_scripts"
+    produces = (Report,)
+    tags = (IPUWorkflowTag, FactsPhaseTag)
+
+    def process(self):
+        customnetworkscripts.process()
--- a/repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
+++ b/repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
@@ -1,0 +1,58 @@
+import os
+
+from leapp import reporting
+
+CUSTOM_NETWORK_SCRIPTS = [
+    "/sbin/ifup-local",
+    "/sbin/ifup-pre-local",
+    "/sbin/ifdown-local",
+    "/sbin/ifdown-pre-local",
+]
+DOC_URL = "https://red.ht/upgrading-RHEL-8-to-RHEL-9-network-scripts"
+
+
+def generate_report(existing_custom_network_scripts):
+    """ Generate reports informing user about possible manual intervention required """
+
+    # Show documentation url if custom network-scripts detected
+    title = "custom network-scripts detected"
+    summary = (
+        "RHEL 9 does not support the legacy network-scripts package that was"
+        " deprecated in RHEL 8. Custom network-scripts have been detected."
+    )
+
+    reporting.create_report(
+        [
+            reporting.Title(title),
+            reporting.Summary(summary),
+            reporting.Remediation(
+                hint=(
+                    "Migrate the custom network-scripts to NetworkManager dispatcher"
+                    " scripts manually before the ugprade. Follow instructions in the"
+                    " official documentation."
+                )
+            ),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Flags([reporting.Flags.INHIBITOR]),
+            reporting.Tags([reporting.Tags.NETWORK, reporting.Tags.SERVICES]),
+            reporting.ExternalLink(
+                title=(
+                    "Upgrading from RHEL 8 to 9 - migrating custom network-scripts to"
+                    " NetworkManager dispatcher scripts"
+                ),
+                url=DOC_URL,
+            ),
+        ]
+        + [
+            reporting.RelatedResource("file", fname)
+            for fname in existing_custom_network_scripts
+        ]
+    )
+
+
+def process():
+    existing_custom_network_scripts = [
+        fname for fname in CUSTOM_NETWORK_SCRIPTS if os.path.isfile(fname)
+    ]
+    if existing_custom_network_scripts:
+        generate_report(existing_custom_network_scripts)
--- a/repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/tests/unit_test_customnetworkscripts.py
+++ b/repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/tests/unit_test_customnetworkscripts.py
@@ -1,0 +1,19 @@
+import os
+
+from leapp import reporting
+from leapp.libraries.actor import customnetworkscripts
+from leapp.libraries.common import testutils
+
+
+def test_customnetworkscripts_exists(monkeypatch):
+    monkeypatch.setattr(os.path, "isfile", lambda dummy: True)
+    monkeypatch.setattr(reporting, "create_report", testutils.create_report_mocked())
+    customnetworkscripts.process()
+    assert reporting.create_report.called
+
+
+def test_customnetworkscripts_not_found(monkeypatch):
+    monkeypatch.setattr(os.path, "isfile", lambda dummy: False)
+    monkeypatch.setattr(reporting, 'create_report', testutils.create_report_mocked())
+    customnetworkscripts.process()
+    assert not reporting.create_report.called


repos/system_upgrade/el7toel8/actors/checklocalnetworkscripts/actor.py
```diff
@@ -0,0 +1,28 @@
...
+
+class LocalNetworkScriptsCheck(Actor):
+    """
+    Check for existence of network-scripts local scripts and warn user about possible
+    manual intervention requirements.
+    """
+
+    name = "local_network-scripts_check"
+    produces = (Report,)
+    tags = (IPUWorkflowTag, ChecksPhaseTag)
```
pirat89, line 19: The actor has to be executed in the `FactsPhaseTag` phase instead.
* in case an actor has to work with the system it cannot be part of check-phases. (imagine that actors executed in the checksphase should be possible to execute on a remote machine, so they can work just with information delivered via messages and cannot call any commands on the system, so cannot check whether particular files exists). Ideally it should be 2 actors (one getting the data and producing a msg for the check actor, and check actor that create the inhibitor), but in this case, let's rather keep it just one actor and move it to the factscollection phase.

repos/system_upgrade/el7toel8/actors/checklocalnetworkscripts/libraries/library.py
```diff
@@ -0,0 +1,26 @@
...
+        [
+            reporting.Title(title),
+            reporting.Summary(summary),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Flags([reporting.Flags.INHIBITOR]),
+            reporting.Tags([reporting.Tags.SERVICES]),
+            reporting.ExternalLink(
+                title="Please migrate custom network-scripts to NetworkManager dispatcher scripts.",
+                url=DOC_URL,
+            ),
```
pirat89, line 24: Add please resources fields:
```
reporting.RelatedResource('file', </abs/path/to/file>)
```
You can mention all files you test for the existence (one per the resource always) or just those that really exists on the system - both solutions are ok. but add the field.

As well, I suggest to add `reporting.Remediation` with `hint` pointing people they have to migrate the local  network-scripts to NetworkManager dispatcher scripts following the official documentation.

repos/system_upgrade/el7toel8/actors/checklocalnetworkscripts/actor.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def process(self):
+        if (
+            os.path.isfile("/sbin/ifup-local")
+            or os.path.isfile("/sbin/ifup-pre-local")
+            or os.path.isfile("/sbin/ifdown-local")
+            or os.path.isfile("/sbin/ifdown-pre-local")
+        ):
+            generate_report()
```
pirat89, line 28: for the purpose of unit-tests, it's better to put the whole process functionality in the private library. e.g see the scandasd actor: https://github.com/oamg/leapp-repository/blob/b7ebfca50c3ad895d9893541f912b8197469d408/repos/system_upgrade/el7toel8/actors/scandasd/actor.py#L24

as well, please add some simple unit tests. look at the very same actor for inspiration how to mock `os.path.isfile` and parametrize decorator. Plus this one actor to see how to mock reporting in unit tests: https://github.com/oamg/leapp-repository/blob/b7ebfca50c3ad895d9893541f912b8197469d408/repos/system_upgrade/el7toel8/actors/checkcpu/tests/test_check_cpu_checkcpu.py#L44

repos/system_upgrade/el7toel8/actors/checklocalnetworkscripts/libraries/library.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def generate_report():
+    """ Generate reports informing user about possible manual intervention requirements """
+
+    DOC_URL = "https://red.ht/upgrading-RHEL-8-to-RHEL-9-network-scripts"
+
+    # Show documentation url if local scripts of network-scripts installed
+    title = "local scripts of network-scripts detected"
+    summary = "Upgrade will be stopped. Additional steps might be required.\n\n"
```
pirat89, line 12: The summary should tell users all they need to know. Reading the summary, it's not helpful in any way. Way is the upgrade stopped? What steps are required? The thing that upgrade is stopped should not even appear in the summary, as that should be obvious from the the point the upgrade is inhibited. Instead of that write summary about the problem. e.g. something like: "The local network-scripts has been detected but these are not supported nor functional on the target system anymore." (my example could be wrong regarding this particular problem, but just giving you a point what is kind of expected to read in summary)

repos/system_upgrade/el7toel8/actors/checklocalnetworkscripts/libraries/library.py
```diff
@@ -0,0 +1,26 @@
+import os
+from leapp import reporting
+
+
+def generate_report():
+    """ Generate reports informing user about possible manual intervention requirements """
+
+    DOC_URL = "https://red.ht/upgrading-RHEL-8-to-RHEL-9-network-scripts"
```
pirat89, line 8: The document is invalid now. I guess it will not be available until the RHEL 9 release and this actor is now in the repository for the upgrade RHEL 7 to RHEL 8. What's the expectation? In case it's just for IPU 8 -> 9, it should be in the `el8toel9` repository, not `el7toel8`.

repos/system_upgrade/el7toel8/actors/checklocalnetworkscripts/libraries/library.py
```diff
@@ -0,0 +1,26 @@
...
+
+    reporting.create_report(
+        [
+            reporting.Title(title),
+            reporting.Summary(summary),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Flags([reporting.Flags.INHIBITOR]),
+            reporting.Tags([reporting.Tags.SERVICES]),
+            reporting.ExternalLink(
+                title="Please migrate custom network-scripts to NetworkManager dispatcher scripts.",
```
pirat89, line 22: Ideally title of the article, not whatever text - it should not be a text "do please this" or "do this", but title of the link.

pirat89 (changes requested):

pirat89:
Hi @liangwen12year , please go through my comments and questions.

liangwen12year:
> Hi @liangwen12year , please go through my comments and questions.

Thanks for the comments, I will update properly.

repos/system_upgrade/el7toel8/actors/checklocalnetworkscripts/libraries/library.py
liangwen12year: Thanks, I moved to `el8toel9` dir.

repos/system_upgrade/el7toel8/actors/checklocalnetworkscripts/libraries/library.py
liangwen12year: Updated to use the title of the link.

repos/system_upgrade/el7toel8/actors/checklocalnetworkscripts/libraries/library.py
liangwen12year: Updated. I explained why the upgrade stops and  what steps are required.

repos/system_upgrade/el7toel8/actors/checklocalnetworkscripts/actor.py
liangwen12year: Thanks for the explanation, I used `FactsPhaseTag` phase instead.

repos/system_upgrade/el7toel8/actors/checklocalnetworkscripts/libraries/library.py
liangwen12year: Resource fields and `reporting.Remediation` were added.

repos/system_upgrade/el7toel8/actors/checklocalnetworkscripts/actor.py
liangwen12year: I included the whole process functionality in the private library. And unit tests were also added.

liangwen12year:
@pirat89 , I addressed all your requested changes and comments

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/tests/unit_test_customnetworkscripts.py
```diff
@@ -0,0 +1,27 @@
...
+    monkeypatch.setattr(os.path, "isfile", lambda dummy: True)
+    monkeypatch.setattr(reporting, "create_report", testutils.create_report_mocked())
+    customnetworkscripts.process()
+    assert reporting.create_report.called
+
+
+def test_customnetworkscripts_not_found(monkeypatch):
+    monkeypatch.setattr(os.path, "isfile", lambda dummy: False)
+    customnetworkscripts.process()
+    assert not reporting.create_report.called
```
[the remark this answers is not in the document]
pirat89: @liangwen12year  in this case, the logic is wrong. The `create_report` is by default function, so you cannot check whether it has been called like that. Basically you do this:

```python
def my_function():
      print('I have been called')


assert not my_function.called
```

which will basically result in something like:
```
AttributeError: 'function' object has no attribute 'called'
```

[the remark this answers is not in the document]
pirat89: basically you can see it when you run the tests locally

```
monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f12dd1787d0>

    def test_customnetworkscripts_not_found(monkeypatch):
        monkeypatch.setattr(os.path, "isfile", lambda dummy: False)
        customnetworkscripts.process()
>       assert not reporting.create_report.called
E       AttributeError: 'function' object has no attribute 'called'
```

pirat89:
@liangwen12year  see my comments above. fixse are needed. I approved the run of the tests in the PR so you will be now aware of the problems right after the commit.

pirat89:
@liangwen12year btw, when we can expect the article mentioned in the PR will be accessible / written / published?

liangwen12year:
> @liangwen12year btw, when we can expect the article mentioned in the PR will be accessible / written / published?

Around RHEL-9 GA in the next spring.

pirat89:
@liangwen12year why so late? this could be relevant already for RHEL 9 beta, couldn't?

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/actor.py
```diff
@@ -0,0 +1,21 @@
...
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+
+
+class CheckCustomNetworkScripts(Actor):
+    """
+    Check the existence of custom network-scripts and warn user about possible
+    manual intervention requirements.
+    """
+
+    name = "custom_network_scripts_check"
```
pirat89, line 16: Regarding the issues, I suggest to be consistent in the name of the actor and strings used fot the class name and directory. E.g.

```suggestion
    name = "check_custom_network_scripts"
```

liangwen12year: Thanks for suggestion, I updated.

pirat89 (dismissed):
It lookks good to me now :)

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
```diff
@@ -0,0 +1,52 @@
...
+        ]
+    )
+
+
+def process():
+    if (
+        os.path.isfile("/sbin/ifup-local")
+        or os.path.isfile("/sbin/ifup-pre-local")
+        or os.path.isfile("/sbin/ifdown-local")
+        or os.path.isfile("/sbin/ifdown-pre-local")
```
tyll, line 50: Ideally, the list of filenames should be defined only once, for example:

```python
CUSTOM_NETWORK_SCRIPTS = ["/sbin/ifup-local", ...]
...
for file in CUSTOM_NETWORK_SCRIPTS:
    reporting.RelatedResource("file", file)
...
if any(os.path.isfile(file) for file in CUSTOM_NETWORK_SCRIPTS):
    generate_report()
```

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
```diff
@@ -0,0 +1,51 @@
...
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Flags([reporting.Flags.INHIBITOR]),
+            reporting.Tags([reporting.Tags.SERVICES]),
+            reporting.ExternalLink(
+                title="Upgrading from RHEL 8 to 9 - migrating custom network-scripts \
+to NetworkManager dispatcher scripts",
+                url=DOC_URL,
+            ),
+        ]
+        + [reporting.RelatedResource("file", file) for file in CUSTOM_NETWORK_SCRIPTS]
```
pirat89, line 45: @liangwen12year https://leapp.readthedocs.io/en/latest/python-coding-guidelines.html#avoid-shadowing-python-builtins

Do not use the `file` varname. You can use e.g. `fname`, `fpath`, `filename`, ...

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+
+
+def process():
+    if any(os.path.isfile(file) for file in CUSTOM_NETWORK_SCRIPTS):
```
pirat89, line 50: same here

liangwen12year: Updated.

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
liangwen12year: Thanks for commenting, updated.

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def generate_report():
+    """ Generate reports informing user about possible manual intervention requirements """
+
+    DOC_URL = "https://red.ht/upgrading-RHEL-8-to-RHEL-9-network-scripts"
+
+    # Show documentation url if custom network-scripts detected
+    title = "custom network-scripts detected"
+    summary = "RHEL 9 does not support the legacy network-scripts package that was \
```
fernflower, line 20: Nit: That's super minor, but could you please use string concatenation instead of \?
Like
```
summary = ("RHEL 9 does not support the legacy network-scripts package that was"
           "deprecated in RHEL 8, and the custom network-scripts have been detected. In order to"
           "proceed with the upgrade, you can move the custom network-scripts, for example, to"
           "the `/opt/network-scripts/` directory."
)
```

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
liangwen12year: I updated the patch, thanks.

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
```diff
@@ -0,0 +1,51 @@
...
+    "/sbin/ifup-pre-local",
+    "/sbin/ifdown-local",
+    "/sbin/ifdown-pre-local",
+]
+
+
+def generate_report():
+    """ Generate reports informing user about possible manual intervention requirements """
+
+    DOC_URL = "https://red.ht/upgrading-RHEL-8-to-RHEL-9-network-scripts"
```
fernflower, line 16: Could you move DOC_URL to the same level as CUSTOM_NETWORK_SCRIPTS constant?

fernflower (commented):
lgtm, some nits.
I launched integration tests for your pr as well.

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
pirat89: i missed that. thanks for the socnd eyes @fernflower :)

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
```diff
@@ -0,0 +1,51 @@
...
+                    " scripts manually before the ugprade. Follow instructions in the"
+                    " official documentation."
+                )
+            ),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Flags([reporting.Flags.INHIBITOR]),
+            reporting.Tags([reporting.Tags.SERVICES]),
+            reporting.ExternalLink(
+                title="Upgrading from RHEL 8 to 9 - migrating custom network-scripts \
+to NetworkManager dispatcher scripts",
```
pirat89, line 41: same here

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
liangwen12year: I moved it to the same level as  CUSTOM_NETWORK_SCRIPTS constant.

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
liangwen12year: I changed it into string concatenation.

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
liangwen12year: I removed the `\`.

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
```diff
@@ -0,0 +1,53 @@
...
+                ),
+                url=DOC_URL,
+            ),
+        ]
+        + [reporting.RelatedResource("file", fname) for fname in CUSTOM_NETWORK_SCRIPTS]
+    )
+
+
+def process():
+    if any(os.path.isfile(fname) for fname in CUSTOM_NETWORK_SCRIPTS):
```
fernflower, line 52: Hmm this way you will report all files specified in CUSTOM_NETWORK_SCRIPTS regardless of how many actually exist. Is it intended?
Maybe some additional os.path.isfile check should be added to generate_report, to filter out only actually existing files?

liangwen12year: Ideally, we should generate report as long as the system is detecting any one of the following scripts:
```
    "/sbin/ifup-local"
    "/sbin/ifup-pre-local"
    "/sbin/ifdown-local"
    "/sbin/ifdown-pre-local"
```

> Maybe some additional os.path.isfile check should be added to generate_report, to filter out only actually existing files?
what do you mean ?

pirat89: The generated report should ideally contain information relevant just to the actual system. What @fernflower means is that line 47 should look like this to print only relevant (present) resources:
```
        + [reporting.RelatedResource("file", fname) for fname in CUSTOM_NETWORK_SCRIPTS if os.path.exists(fname)]
```

So scripts that are not used on the system, are not reported.

liangwen12year: Ok, thanks for pointing out, I updated.

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def generate_report():
+    """ Generate reports informing user about possible manual intervention requirements """
+
+    # Show documentation url if custom network-scripts detected
+    title = "custom network-scripts detected"
+    summary = (
+        "RHEL 9 does not support the legacy network-scripts package that was"
+        " deprecated in RHEL 8, and the custom network-scripts have been"
```
tyll, line 21:
```suggestion
        " deprecated in RHEL 8. Custom network-scripts have been"
```

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
```diff
@@ -0,0 +1,59 @@
...
+        " detected. In order to proceed with the upgrade, you can move the"
+        " custom network-scripts, for example, to the `/opt/network-scripts/`"
```
tyll, line 23: The advise to move the scripts to `/opt/network-scripts/` contradicts the advise to move to dispatcher scripts (which should be the only recommended way). The summary does not seem to be the right place to mentioned the remediation step, so I would remove this sentence here.

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
tyll: There is no need to check the files twice if `generate_report()` can be modified to accept an argument (which it probably can):

```
def process():
    custom_network_scripts = [
    "/sbin/ifup-local",
    "/sbin/ifup-pre-local",
    "/sbin/ifdown-local",
    "/sbin/ifdown-pre-local",
]
    related_resources = [reporting.RelatedResource("file", filename) for filename in custom_network_scripts]
    if related_resources:
       generate_report(related_resources)
```

pirat89: @tyll it has to be Python3.6+ compatible as it's the default python for RHEL 8.

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
liangwen12year: I removed the "advice".

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
```diff
@@ -0,0 +1,56 @@
...
+    "/sbin/ifup-local",
+    "/sbin/ifup-pre-local",
+    "/sbin/ifdown-local",
+    "/sbin/ifdown-pre-local",
+]
+DOC_URL = "https://red.ht/upgrading-RHEL-8-to-RHEL-9-network-scripts"
+
+
+def generate_report(existed_custom_network_scripts):
+    """ Generate reports informing user about possible manual intervention requirements """
```
fernflower, line 15: s/requirements/required

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
```diff
@@ -0,0 +1,56 @@
...
+
+from leapp import reporting
+
+CUSTOM_NETWORK_SCRIPTS = [
+    "/sbin/ifup-local",
+    "/sbin/ifup-pre-local",
+    "/sbin/ifdown-local",
+    "/sbin/ifdown-pre-local",
+]
+DOC_URL = "https://red.ht/upgrading-RHEL-8-to-RHEL-9-network-scripts"
```
fernflower, line 11: Sorry for maybe stupid question - this link is broken because there is no such article yet and will be updated once it appears?

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
```diff
@@ -0,0 +1,56 @@
...
+        + [
+            reporting.RelatedResource("file", fname)
+            for fname in existed_custom_network_scripts
+        ]
+    )
+
+
+def process():
+    existed_custom_network_scripts = [fname for fname in CUSTOM_NETWORK_SCRIPTS if os.path.isfile(fname)]
+    if existed_custom_network_scripts:
```
fernflower, line 55: Please rename it to either `existing_custom_network_scripts` or `discovered_custom_network_scripts` or just `custom_network_scripts`.

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
```diff
@@ -0,0 +1,56 @@
...
+            reporting.Remediation(
+                hint=(
+                    "Migrate the custom network-scripts to NetworkManager dispatcher"
+                    " scripts manually before the ugprade. Follow instructions in the"
+                    " official documentation."
+                )
+            ),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Flags([reporting.Flags.INHIBITOR]),
+            reporting.Tags([reporting.Tags.SERVICES]),
```
fernflower, line 37: Would it make sense to add reporting.Tags.NETWORK here as well?

fernflower (commented):
Some suggestions and one open question, otherwise lgtm.

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
liangwen12year: The link is reserved for publishing the article, and the Docs Team decide to publish it around RHEL-9 GA in the next spring.

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
liangwen12year: Yes. I added the tag `reporting.Tags.NETWORK`.

repos/system_upgrade/el8toel9/actors/checkcustomnetworkscripts/libraries/customnetworkscripts.py
pirat89: @fernflower that's expected. the link will be generated just ~days before the release. We can expect this to be common for el8toel9 until RHEL 9 GA.

liangwen12year:
@fernflower , all your suggestions are addressed. Thanks.

fernflower (commented):
lgtm, will approve once the integration tests pass. Thank you!

fernflower (approved):
Tests passed, all comments addressed, approving.

PR #732, opened 2021-10-06
Add sssd el8toel9 actor

pbrezina:
Check SSSD configuration for changes in RHEL9 and report them in model.

Implicit files domain is disabled by default. This may affect local
smartcard authentication if there is not explicit files domain created.

If there is no files domain and smartcard authentication is enabled,
we will notify the administrator.

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/sssdfacts/libraries/sssdfacts8to9.py
+++ b/repos/system_upgrade/el8toel9/actors/sssdfacts/libraries/sssdfacts8to9.py
@@ -1,0 +1,43 @@
+from leapp.models import SSSDConfig8to9
+
+
+class SSSDFactsLibrary(object):
+    """
+    Helper library from SSSDFacts actor to allow unit testing.
+    """
+    def __init__(self, config):
+        self.config = config
+
+    def process(self):
+        """
+        Check SSSD configuration for the following options:
+
+        - sssd/enable_files_domain
+        - pam/pam_cert_auth
+        - explicit files provider domain present
+        """
+        facts = SSSDConfig8to9(
+            enable_files_domain_set=False,
+            explicit_files_domain=False,
+            pam_cert_auth=False,
+        )
+
+        facts.enable_files_domain_set = self.config.has_option('sssd', 'enable_files_domain')
+        facts.pam_cert_auth = self.config.getboolean('pam', 'pam_cert_auth', fallback=False)
+
+        # Lookup domain with id_provider == files
+        for section in self.config.sections():
+            # Not a domain config.
+            if not section.startswith('domain/'):
+                continue
+
+            # Malformed config?
+            if not self.config.has_option(section, 'id_provider'):
+                continue
+
+            val = self.config.get(section, 'id_provider', fallback=None)
+            if val == 'files':
+                facts.explicit_files_domain = True
+                break
+
+        return facts
--- a/repos/system_upgrade/el8toel9/actors/sssdfacts/tests/unit_test_sssdfacts_8to9.py
+++ b/repos/system_upgrade/el8toel9/actors/sssdfacts/tests/unit_test_sssdfacts_8to9.py
@@ -1,0 +1,167 @@
+import textwrap
+
+from six import StringIO
+
+from leapp.libraries.actor.sssdfacts8to9 import SSSDFactsLibrary
+from leapp.libraries.common import utils
+
+
+def get_config(content):
+    return utils.parse_config(StringIO(textwrap.dedent(content)))
+
+
+def test_empty_config():
+    config = utils.parse_config()
+    facts = SSSDFactsLibrary(config).process()
+
+    assert not facts.enable_files_domain_set
+    assert not facts.explicit_files_domain
+    assert not facts.pam_cert_auth
+
+
+def test_enable_files_domain_set__notset():
+    config = get_config("""
+    [sssd]
+    """)
+
+    facts = SSSDFactsLibrary(config).process()
+
+    assert not facts.enable_files_domain_set
+    assert not facts.explicit_files_domain
+    assert not facts.pam_cert_auth
+
+
+def test_enable_files_domain_set__set_true():
+    config = get_config("""
+    [sssd]
+    enable_files_domain = true
+    """)
+
+    facts = SSSDFactsLibrary(config).process()
+
+    assert facts.enable_files_domain_set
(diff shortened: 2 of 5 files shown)


pbrezina:
Apparently the actors names cannot be the same as in el7toel8. Do you have any recommendation?

pbrezina:
@vinzenz I rebased the patches. Has there been any progress on the pytest issue?

vinzenz:
> @vinzenz I rebased the patches. Has there been any progress on the pytest issue?

I need to work on that, sorry for that. I just put it aside before the release or the new leapp version. I will do this ASAP.

vinzenz:
I fixed the problem by renaming your library. The problem cannot be solved in any timely fashion in another way at this point. 

The problem lies in the way how we execute the tests. We do load all repositories and once an actor has been loaded, their libraries are preloaded into the pytest process. This manifests only for unit tests. We do not have this problem for leapp itself. By having a unique name of the library, we can avoid this, but this implies some follow up work for us to detect these things and at least give a hint so we know where the problem is.

Initially I thought that the problem lies deeper, however leapp is doing everything right, just unfortunately we cannot create new processes for tests in every actor at this moment.

vinzenz (approved):

repos/system_upgrade/el8toel9/actors/sssdfacts/libraries/sssdfacts8to9.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class SSSDFactsLibrary(object):
+    """
+    Helper library from SSSDFacts actor to allow unit testing.
+    """
+    def __init__(self, config):
+        self.config = config
+
+    def get_fallback(self, section, key, fallback):
```
fernflower, line 11: Maybe instead of having 2 dedicated functions just use configparser's fallback option inplace?
Like change occurrences  of get_fallback to `self.config.get(section, key, fallback=fallback)` and getboolean_fallback change to `self.config.getboolean(section, key, fallback=fallback)`?

https://docs.python.org/3/library/configparser.html#fallback-values

repos/system_upgrade/el8toel9/actors/sssdfacts/libraries/sssdfacts8to9.py
```diff
@@ -0,0 +1,53 @@
...
+        if not self.config.has_option(section, key):
+            return fallback
+        return self.config.get(section, key)
+
+    def getboolean_fallback(self, section, key, fallback):
```
fernflower, line 16: Please see comment above

repos/system_upgrade/el8toel9/actors/sssdfacts/libraries/sssdfacts8to9.py
```diff
@@ -0,0 +1,53 @@
...
+        - explicit files provider domain present
+        """
+        facts = SSSDConfig8to9(
+            enable_files_domain_set=False,
+            explicit_files_domain=False,
+            pam_cert_auth=False,
+        )
+
+        facts.enable_files_domain_set = self.config.has_option('sssd', 'enable_files_domain')
+        facts.pam_cert_auth = self.getboolean_fallback('pam', 'pam_cert_auth', fallback=False)
```
fernflower, line 36: Can use configparser's native functions here

repos/system_upgrade/el8toel9/actors/sssdfacts/libraries/sssdfacts8to9.py
```diff
@@ -0,0 +1,53 @@
...
+        for section in self.config.sections():
+            # Not a domain config.
+            if not section.startswith('domain/'):
+                continue
+
+            # Malformed config?
+            if not self.config.has_option(section, 'id_provider'):
+                continue
+
+            val = self.get_fallback(section, 'id_provider', fallback=None)
```
fernflower, line 48: Same here

repos/system_upgrade/el8toel9/actors/sssdfacts/tests/unit_test_sssdfacts_8to9.py
```diff
@@ -0,0 +1,168 @@
+import textwrap
+
+from six import StringIO
+
+from leapp.libraries.actor.sssdfacts8to9 import SSSDFactsLibrary
+from leapp.libraries.common import utils
+
+
+def get_config(content):
+    parser = utils.parse_config(StringIO(textwrap.dedent(content)))
```
fernflower, line 10: Nit: can return straightahead, no need for variable assignment

fernflower (commented):
I absolutely love the thorough comments and test coverage, great job!
Some comments inside, otherwise lgtm.

repos/system_upgrade/el8toel9/actors/sssdfacts/libraries/sssdfacts8to9.py
pbrezina: @vinzenz You implemented custom fallback in order to support python2 as well. Is this still needed since the python2 tests are not run anymore?

vinzenz: I did, yes and I actually started writing an answer to this already but forgot to send it how it looks like.

We shouldn't need it anymore, so we could revert those things, if you don't mind back to the python3 version.

pbrezina: Done.

repos/system_upgrade/el8toel9/actors/sssdfacts/tests/unit_test_sssdfacts_8to9.py
pbrezina: Done.

repos/system_upgrade/el8toel9/actors/sssdfacts/libraries/sssdfacts8to9.py
pbrezina: Done.

pbrezina:
Tests pass now. Is there anything needed from my side?

vinzenz:
@pbrezina  Hi, nothing missing from your end, but we'll need your QE to contribute integration tests as discussed on the JIRA ticket of your team. The PR itself is in a very good shape.

Rezney (approved):

PR #733, opened 2021-10-07
add Satellite el7toel8 upgrade actor

evgeni:
this is mostly for me and @vinzenz to have something to talk about :)

### DIFF
--- a/repos/system_upgrade/common/libraries/testutils.py
+++ b/repos/system_upgrade/common/libraries/testutils.py
@@ -1,6 +1,6 @@
-from collections import namedtuple
 import logging
 import os
+from collections import namedtuple
 
 from leapp.libraries.common.config import architecture
 from leapp.models import EnvVar
@@ -20,14 +20,22 @@ class produce_mocked(object):
 class create_report_mocked(object):
     def __init__(self):
         self.called = 0
-        self.report_fields = {}
+        self.reports = []
 
     def __call__(self, report_fields):
         self.called += 1
+        full_report = {}
         # iterate list of report primitives (classes)
         for report in report_fields:
             # last element of path is our field name
-            self.report_fields.update(report.to_dict())
+            full_report.update(report.to_dict())
+        self.reports.append(full_report)
+
+    @property
+    def report_fields(self):
+        if self.reports:
+            return self.reports[-1]
+        return {}
 
 
 class logger_mocked(object):
--- a/repos/system_upgrade/el7toel8/tools/handle-postgresql-legacy-actions
+++ b/repos/system_upgrade/el7toel8/tools/handle-postgresql-legacy-actions
@@ -1,0 +1,6 @@
+#!/usr/bin/bash -e
+
+# lets assume the new PostgreSQL package will just fix this
+if [ -L /usr/libexec/initscripts/legacy-actions/postgresql ]; then
+  rm -f /usr/libexec/initscripts/legacy-actions/postgresql
+fi
--- a/repos/system_upgrade/common/actors/createresumeservice/files/leapp_resume.service
+++ b/repos/system_upgrade/common/actors/createresumeservice/files/leapp_resume.service
@@ -11,3 +11,5 @@ Type=oneshot
 # FIXME: this is temporary workround for Python3
 ExecStart=/root/tmp_leapp_py3/leapp3 upgrade --resume
 StandardOutput=journal+console
+# FIXME: this shouldn't be needed, but Satellite upgrade runs installer, and that's slow
+TimeoutStartSec=infinity
--- a/repos/system_upgrade/el7toel8/actors/satellite_upgrade_check/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/satellite_upgrade_check/actor.py
@@ -1,0 +1,22 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.satellite_upgrade_check import satellite_upgrade_check
+from leapp.models import Report, SatelliteFacts
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class SatelliteUpgradeCheck(Actor):
+    """
+    Check state of Satellite system before upgrade
+    """
+
+    name = 'satellite_upgrade_check'
+    consumes = (SatelliteFacts,)
+    produces = (Report,)
+    tags = (IPUWorkflowTag, ChecksPhaseTag)
+
+    def process(self):
+        facts = next(self.consume(SatelliteFacts), None)
+        if not facts or not facts.has_foreman:
+            return
+
+        satellite_upgrade_check(facts)
--- a/repos/system_upgrade/el7toel8/actors/satellite_upgrade_check/libraries/satellite_upgrade_check.py
+++ b/repos/system_upgrade/el7toel8/actors/satellite_upgrade_check/libraries/satellite_upgrade_check.py
@@ -1,0 +1,53 @@
+import textwrap
+
+from leapp import reporting
+
+
+def satellite_upgrade_check(facts):
+    if facts.postgresql.local_postgresql:
+        if facts.postgresql.old_var_lib_pgsql_data:
+            title = "Old PostgreSQL data found in /var/lib/pgsql/data"
+            summary = """
+            The upgrade wants to move PostgreSQL data to /var/lib/pgsql/data,
+            but this directory already exists on your system.
+            Please make sure /var/lib/pgsql/data doesn't exist prior to the upgrade.
+            """
+            reporting.create_report([
+                reporting.Title(title),
+                reporting.Summary(textwrap.dedent(summary).strip()),
+                reporting.Severity(reporting.Severity.HIGH),
+                reporting.Tags([]),
+                reporting.Flags([reporting.Flags.INHIBITOR])
+            ])
+
+        title = "Satellite PostgreSQL data migration"
+        flags = []
+        severity = reporting.Severity.MEDIUM
+
+        if facts.postgresql.same_partition:
+            summary = "Your PostgreSQL data will be automatically migrated."
+        else:
+            scl_psql_path = '/var/opt/rh/rh-postgresql12/lib/pgsql/data/'
+            if facts.postgresql.space_required > facts.postgresql.space_available:
+                storage_message = """You currently don't have enough free storage to move the data.
+                Automatic moving cannot be performed."""
+                flags = [reporting.Flags.INHIBITOR]
+                severity = reporting.Severity.HIGH
+            else:
+                storage_message = """You currently have enough free storage to move the data.
+                This operation can be performed by the upgrade process."""
+            summary = """
+            Your PostgreSQL data in {} is currently on a dedicated volume.
+            PostgreSQL on RHEL8 expects the data to live in /var/lib/pgsql/data.
+            {}
+            However, instead of moving the data over, you might want to consider manually adapting your mounts,
+            so that the contents of {} are available in /var/lib/pgsql/data.
+            """.format(scl_psql_path, storage_message, scl_psql_path)
+
+        reporting.create_report([
+            reporting.Title(title),
+            reporting.Summary(textwrap.dedent(summary).strip()),
+            reporting.Severity(severity),
+            reporting.Tags([]),
+            reporting.Flags(flags)
+        ])
--- a/repos/system_upgrade/el7toel8/actors/satellite_upgrade_check/tests/unit_test_satellite_upgrade_check.py
+++ b/repos/system_upgrade/el7toel8/actors/satellite_upgrade_check/tests/unit_test_satellite_upgrade_check.py
@@ -1,0 +1,79 @@
+from leapp import reporting
+from leapp.libraries.actor.satellite_upgrade_check import satellite_upgrade_check
+from leapp.libraries.common.testutils import create_report_mocked
+from leapp.models import SatelliteFacts, SatellitePostgresqlFacts
+
+
+def test_old_data(monkeypatch):
+    monkeypatch.setattr(reporting, 'create_report', create_report_mocked())
+
+    satellite_upgrade_check(SatelliteFacts(has_foreman=True,
+                            postgresql=SatellitePostgresqlFacts(local_postgresql=True, old_var_lib_pgsql_data=True)))
+
+    assert reporting.create_report.called == 2
+
+    expected_title = 'Old PostgreSQL data found in /var/lib/pgsql/data'
+    assert next((report for report in reporting.create_report.reports if report.get('title') == expected_title), None)
+
+    expected_title = 'Satellite PostgreSQL data migration'
+    assert next((report for report in reporting.create_report.reports if report.get('title') == expected_title), None)
+
+
+def test_no_old_data(monkeypatch):
+    monkeypatch.setattr(reporting, 'create_report', create_report_mocked())
+
+    satellite_upgrade_check(SatelliteFacts(has_foreman=True,
+                            postgresql=SatellitePostgresqlFacts(local_postgresql=True, old_var_lib_pgsql_data=False)))
+
+    assert reporting.create_report.called == 1
+
+    expected_title = 'Satellite PostgreSQL data migration'
+
+    assert expected_title == reporting.create_report.report_fields['title']
+
+
+def test_same_disk(monkeypatch):
+    monkeypatch.setattr(reporting, 'create_report', create_report_mocked())
+
+    satellite_upgrade_check(SatelliteFacts(has_foreman=True,
+                            postgresql=SatellitePostgresqlFacts(local_postgresql=True, same_partition=True)))
+
+    assert reporting.create_report.called == 1
+
+    expected_title = 'Satellite PostgreSQL data migration'
+    expected_summary = 'Your PostgreSQL data will be automatically migrated.'
+
+    assert expected_title == reporting.create_report.report_fields['title']
+    assert expected_summary == reporting.create_report.report_fields['summary']
+
+
+def test_different_disk_sufficient_storage(monkeypatch):
+    monkeypatch.setattr(reporting, 'create_report', create_report_mocked())
+
+    satellite_upgrade_check(SatelliteFacts(has_foreman=True,
+                            postgresql=SatellitePostgresqlFacts(local_postgresql=True, same_partition=False,
+                                                                space_required=5, space_available=10)))
+
+    assert reporting.create_report.called == 1
+
+    expected_title = 'Satellite PostgreSQL data migration'
+    expected_summary = 'You currently have enough free storage to move the data'
+
+    assert expected_title == reporting.create_report.report_fields['title']
+    assert expected_summary in reporting.create_report.report_fields['summary']
+
+
+def test_different_disk_insufficient_storage(monkeypatch):
+    monkeypatch.setattr(reporting, 'create_report', create_report_mocked())
+
+    satellite_upgrade_check(SatelliteFacts(has_foreman=True,
+                            postgresql=SatellitePostgresqlFacts(local_postgresql=True, same_partition=False,
+                                                                space_required=10, space_available=5)))
+
+    assert reporting.create_report.called == 1
+
+    expected_title = 'Satellite PostgreSQL data migration'
+    expected_summary = "You currently don't have enough free storage to move the data"
+
+    assert expected_title == reporting.create_report.report_fields['title']
+    assert expected_summary in reporting.create_report.report_fields['summary']
--- a/repos/system_upgrade/el7toel8/actors/satellite_upgrade_data_migration/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/satellite_upgrade_data_migration/actor.py
@@ -1,0 +1,43 @@
+import glob
+import os
+import shutil
+
+from leapp.actors import Actor
+from leapp.models import SatelliteFacts
+from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
+
+POSTGRESQL_DATA_PATH = '/var/lib/pgsql/data/'
+POSTGRESQL_SCL_DATA_PATH = '/var/opt/rh/rh-postgresql12/lib/pgsql/data/'
+
+SYSTEMD_WANTS_BASE = '/etc/systemd/system/multi-user.target.wants/'
+SERVICES_TO_DISABLE = ['dynflow-sidekiq@*', 'foreman', 'foreman-proxy',
+                       'httpd', 'postgresql', 'pulpcore-api', 'pulpcore-content',
+                       'pulpcore-worker@*', 'tomcat']
+
+
+class SatelliteUpgradeDataMigration(Actor):
+    """
+    Reconfigure Satellite services and migrate PostgreSQL data
+    """
+
+    name = 'satellite_upgrade_data_migration'
+    consumes = (SatelliteFacts,)
+    produces = ()
+    tags = (IPUWorkflowTag, ApplicationsPhaseTag)
+
+    def process(self):
+        facts = next(self.consume(SatelliteFacts), None)
+        if not facts or not facts.has_foreman:
+            return
+
+        # disable services, will be re-enabled by the installer
+        for service_name in SERVICES_TO_DISABLE:
+            for service in glob.glob(os.path.join(SYSTEMD_WANTS_BASE, '{}.service'.format(service_name))):
+                os.unlink(service)
+
+        if facts.postgresql.local_postgresql and os.path.exists(POSTGRESQL_SCL_DATA_PATH):
+            # remove empty PostgreSQL data from the package
+            if os.path.exists(POSTGRESQL_DATA_PATH):
+                os.rmdir(POSTGRESQL_DATA_PATH)
+            # move PostgreSQL data to the new home
+            shutil.move(POSTGRESQL_SCL_DATA_PATH, POSTGRESQL_DATA_PATH)
--- a/repos/system_upgrade/el7toel8/actors/satellite_upgrade_facts/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/satellite_upgrade_facts/actor.py
@@ -1,0 +1,134 @@
+import os
+
+from leapp.actors import Actor
+from leapp.libraries.common.rpms import has_package
+from leapp.libraries.stdlib import run
+from leapp.models import (
+    DNFWorkaround,
+    InstalledRPM,
+    Module,
+    RepositoriesSetupTasks,
+    RpmTransactionTasks,
+    SatelliteFacts,
+    SatellitePostgresqlFacts
+)
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+
+POSTGRESQL_SCL_DATA_PATH = '/var/opt/rh/rh-postgresql12/lib/pgsql/data/'
+
+
+class SatelliteUpgradeFacts(Actor):
+    """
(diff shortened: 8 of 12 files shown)


repos/system_upgrade/el7toel8/actors/satelliteupgradefacts/actor.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+
+    name = 'satellite_upgrade_facts'
+    consumes = (InstalledRPM, )
+    produces = (RepositoriesSetupTasks, RpmTransactionTasks, SatelliteFacts)
+    tags = (IPUWorkflowTag, FactsPhaseTag)
+
+    def process(self):
+        local_postgresql = has_package(InstalledRPM, 'rh-postgresql12-postgresql-server')
+        self.produce(SatelliteFacts(local_postgresql=local_postgresql))
```
pirat89, line 19: btw, why to produce new satellite facts if the only fact is get from other facts? this can be performed in the other actor as well. What is the added value with the new model in this case? basically the first line provides the same information as the second one.

evgeni: So far "because I wanted to play with facts and a boolean looked simple to act upon", but yeah, sure, you're right, there is right now nothing that consuming the `InstalledRPM` fact directly wouldn't do for us.

repos/system_upgrade/el7toel8/actors/satelliteupgradedatamigration/actor.py
```diff
@@ -0,0 +1,36 @@
...
+        if os.path.exists(POSTGRESQL_DATA_PATH):
+            os.rmdir(POSTGRESQL_DATA_PATH)
+        if not os.path.exists('/var/lib/pgsql'):
+            os.mkdir('/var/lib/pgsql')
+        if os.path.exists(POSTGRESQL_SCL_DATA_PATH):
+            os.rename(POSTGRESQL_SCL_DATA_PATH, POSTGRESQL_DATA_PATH)
+        # fix back scenario files, might not be needed once we don't purge installer
+        for rpmsave_file in glob.glob('/etc/foreman-installer/scenarios.d/*.rpmsave'):
+            os.rename(rpmsave_file, rpmsave_file.replace('.rpmsave', ''))
+        # this hack is not needed after 6.10, as Foreman 3.1 has a fix in the installer for it
```
ekohl, line 34: 6.10 is technically Foreman 2.5. Given 6.10 is already in beta, can we drop the workaround now?

repos/system_upgrade/el7toel8/actors/satelliteupgradefacts/actor.py
```diff
@@ -0,0 +1,41 @@
...
+            # until this is on the CDN
+            to_install=['rubygem-foreman_maintain', 'foreman-installer-katello'],
+            modules_to_enable=[Module(name='ruby', stream='2.7')]
+            )
+        )
+        if local_postgresql:
+            self.produce(RpmTransactionTasks(
+                modules_to_enable=[Module(name='postgresql', stream='12')],
+                to_remove=['rh-postgresql12-runtime'],
+                to_install=['postgresql', 'postgresql-server']
```
ekohl, line 32: In some environments we've also seen `-contrib` installed. Would it be lost after this?

repos/system_upgrade/el7toel8/actors/satelliteupgradedatamigration/actor.py
evgeni: 2.5 < 3.1? ;)
"after 6.10" means "with 7.0"

evgeni:
With #672 merged, I think we can start talking about the actual contents of this PR.

repos/system_upgrade/el7toel8/actors/satelliteupgradefacts/actor.py
```diff
@@ -0,0 +1,133 @@
...
+        self.produce(RpmTransactionTasks(
+            to_remove=['tfm-runtime', 'tfm-pulpcore-runtime', 'rh-redis5-runtime', 'rh-ruby27-runtime', 'rh-python38-runtime'],
+            # you'll need to define the repos in leapp_upgrade_repositories.repo
+            # until this is on the CDN
+            to_install=['rubygem-foreman_maintain', 'foreman-installer-katello'],
+            modules_to_enable=[Module(name='ruby', stream='2.7'), Module(name='python38', stream='3.8')]
+            )
+        )
+
+        if has_package(InstalledRPM, 'katello'):
```
ekohl, line 40: Technically this is only needed for candlepin, not katello itself, right?

repos/system_upgrade/el7toel8/actors/satelliteupgradefacts/actor.py
```diff
@@ -0,0 +1,133 @@
...
+
+        local_postgresql = has_package(InstalledRPM, 'rh-postgresql12-postgresql-server')
+        postgresql_contrib = has_package(InstalledRPM, 'rh-postgresql12-postgresql-contrib')
+        postgresql_evr = has_package(InstalledRPM, 'rh-postgresql12-postgresql-evr')
+        # remove packages we don't need anymore
+        self.produce(RpmTransactionTasks(
+            to_remove=['tfm-runtime', 'tfm-pulpcore-runtime', 'rh-redis5-runtime', 'rh-ruby27-runtime', 'rh-python38-runtime'],
+            # you'll need to define the repos in leapp_upgrade_repositories.repo
+            # until this is on the CDN
+            to_install=['rubygem-foreman_maintain', 'foreman-installer-katello'],
```
ekohl, line 35: Isn't `foreman-installer-katello` here covered by lines 61 & 62 now?

repos/system_upgrade/el7toel8/actors/satelliteupgradefacts/actor.py
```diff
@@ -0,0 +1,133 @@
...
+                if pkg.name.startswith('tfm-rubygem-hammer') or pkg.name.startswith('tfm-rubygem-foreman') or pkg.name.startswith('tfm-rubygem-katello') or pkg.name.startswith('tfm-rubygem-smart_proxy'):
+                    additional_rpms.append(pkg.name.replace('tfm-rubygem-', 'rubygem-'))
+                elif pkg.name.startswith('tfm-pulpcore-python3-pulp'):
+                    additional_rpms.append(pkg.name.replace('tfm-pulpcore-python3-', 'python38-'))
+                elif pkg.name.startswith('foreman-installer') or pkg.name.startswith('satellite-installer'):
+                    additional_rpms.append(pkg.name)
+        self.produce(RpmTransactionTasks(to_install=additional_rpms))
+
+        if local_postgresql:
+            scl_psql_path = '/var/opt/rh/rh-postgresql12/lib/pgsql/data/'
```
ekohl, line 66: There's a `POSTGRESQL_SCL_DATA_PATH` constant in the other actor. Could this be unified?

repos/system_upgrade/el7toel8/actors/satelliteupgradeprepare/actor.py
```diff
@@ -0,0 +1,27 @@
...
+from leapp.libraries.common.rpms import has_package
+from leapp.models import InstalledRPM
+from leapp.tags import IPUWorkflowTag, PreparationPhaseTag
+
+
+class SatellitePrepare(Actor):
+    """
+    Handle migration of the yum configuration files.
+    RPM cannot handle replacement of directories by symlinks by default
+    without the %pretrans scriptlet. As yum package is packaged wrong,
```
ekohl, line 12: Which yum package? Is this for the rh-postgresql12-postgresql-syspaths package?

repos/system_upgrade/el7toel8/tools/handle-postgresql-legacy-actions
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/bash -e
+
+# lets asume the new package will just fix this
```
ekohl, line 3:
```suggestion
# lets assume the new package will just fix this
```

repos/system_upgrade/el7toel8/actors/satelliteupgradefacts/actor.py
vinzenz: I think this is just used as an detection mechanism AFAICS

repos/system_upgrade/el7toel8/actors/satelliteupgradefacts/actor.py
```diff
@@ -0,0 +1,133 @@
...
+            else:
+                summary = "Your PostgreSQL data will be automatically migrated."
+
+            reporting.create_report([
+                reporting.Title(title),
+                reporting.Summary(textwrap.dedent(summary)),
+                reporting.Severity(severity),
+                reporting.Tags([]),
+                reporting.Flags(flags)
+            ])
```
vinzenz, line 106: Reports need to be performed by a separate actor "Check" actor, however the trigger needs to be collected in here and you will have to create a custom model to send as a message. The so called check actor, is responsible for producing the report and may only use the information provided by the message. This is a convention with in leapp.

repos/system_upgrade/el7toel8/actors/satelliteupgradefacts/actor.py
```diff
@@ -0,0 +1,133 @@
...
+                    )
+                )
+
+        # technically, this needs to be ansible-2.9-for-rhel-8-x86_64-rpms, but my test
+        # system doesn't have it right now - FIXME
+        # there is no repo for satellite yet, but that also will need enablement
+        # probably should find out what's enabled today and map that nicely?
+        self.produce(RepositoriesSetupTasks(to_enable=['ansible-2-for-rhel-8-x86_64-rpms']))
+        self.produce(RepositoriesSetupTasks(to_enable=['satellite-7.0-for-rhel-8-x86_64-rpms']))
+        self.produce(RepositoriesSetupTasks(to_enable=['satellite-maintenance-7.0-for-rhel-8-x86_64-rpms']))
```
vinzenz, line 133:
```suggestion
        self.produce(RepositoriesSetupTasks(to_enable=['ansible-2-for-rhel-8-x86_64-rpms',
                                                       'satellite-7.0-for-rhel-8-x86_64-rpms',
                                                       'satellite-maintenance-7.0-for-rhel-8-x86_64-rpms']))
```

repos/system_upgrade/el7toel8/actors/satelliteupgradefacts/actor.py
```diff
@@ -0,0 +1,133 @@
...
+                    to_remove=['rh-postgresql12-postgresql-contrib'],
+                    to_install=['postgresql-contrib']
+                    )
+                )
+            if postgresql_evr:
+                self.produce(RpmTransactionTasks(
+                    to_remove=['rh-postgresql12-postgresql-evr'],
+                    to_install=['postgresql-evr']
+                    )
+                )
```
vinzenz, line 125: All these RpmTransactionTasks can be combined into one big one, you'd just collect all the things to remove and add and make one big one out of it.

repos/system_upgrade/el7toel8/actors/satelliteupgrader/actor.py
```diff
@@ -0,0 +1,27 @@
...
+    No documentation has been provided for the satellite_upgrader actor.
+    """
+
+    name = 'satellite_upgrader'
+    consumes = (InstalledRPM, )
+    produces = ()
+    tags = (IPUWorkflowTag, FirstBootPhaseTag.After)
+
+    def process(self):
+        has_foreman = has_package(InstalledRPM, 'foreman') or has_package(InstalledRPM, 'foreman-proxy')
```
vinzenz, line 22: You should rather use a custom message from your facts collection instead.

repos/system_upgrade/el7toel8/actors/satelliteupgradefacts/actor.py
evgeni: It depends on how you look at this. My intention, with these line was "ensure katello and katello-selinux are installed", which only is needed if "katello" was already installed, but has the dependency (via candlepin) on the PKI modules.

repos/system_upgrade/el7toel8/actors/satelliteupgradefacts/actor.py
evgeni: it is, yeah

repos/system_upgrade/el7toel8/actors/satelliteupgrader/actor.py
evgeni: In https://github.com/oamg/leapp-repository/pull/733/files/6d36dff7bff36d223d1adc1569accbd35d3d377e#diff-f91027c92012eac0394869e41f6b9e7c7dda170dd750466bb081d8f4de87e819 @pirat89 said not to (but back then I wasn't using the same "fact" in multiple places).

I don't mind either way, just tell me what the official convention is.

repos/system_upgrade/el7toel8/actors/satelliteupgradefacts/actor.py
ekohl: I should also clarify that I'd like to get rid of the need to install the `katello` package, but perhaps I'm being premature here. The dependency chain is katello-selinux -> candlepin-selinux -> candlepin -> some package from the pki-core module. Perhaps that should be included in a code comment and that's sufficient?

repos/system_upgrade/el7toel8/actors/satelliteupgradedatamigration/actor.py
```diff
@@ -0,0 +1,36 @@
...
+    tags = (IPUWorkflowTag, ApplicationsPhaseTag)
+
+    def process(self):
+        has_foreman = has_package(InstalledRPM, 'foreman') or has_package(InstalledRPM, 'foreman-proxy')
+        if not has_foreman:
+            return
+
+        # move PostgreSQL data to the new home
+        if os.path.exists(POSTGRESQL_DATA_PATH):
+            os.rmdir(POSTGRESQL_DATA_PATH)
```
vinzenz, line 32: If it is not empty, this will raise an exception

repos/system_upgrade/el7toel8/actors/satelliteupgrader/actor.py
evgeni: (after some IRC convo: has_package is not really performant, and if we can save calling it multiple times with the same input, we win)

repos/system_upgrade/el7toel8/actors/satelliteupgradedatamigration/actor.py
```diff
@@ -0,0 +1,36 @@
...
+        if not os.path.exists('/var/lib/pgsql'):
+            os.mkdir('/var/lib/pgsql')
```
ekohl, line 34: Will the RPM later correct any permissions on this?

repos/system_upgrade/el7toel8/actors/satelliteupgrader/actor.py
fernflower: @evgeni Hmm would it help if we had has_packages? =)

repos/system_upgrade/el7toel8/actors/satelliteupgradedatamigration/actor.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def process(self):
+        has_foreman = has_package(InstalledRPM, 'foreman') or has_package(InstalledRPM, 'foreman-proxy')
+        if not has_foreman:
+            return
+
+        # move PostgreSQL data to the new home
+        if os.path.exists(POSTGRESQL_DATA_PATH):
+            os.rmdir(POSTGRESQL_DATA_PATH)
+        if not os.path.exists('/var/lib/pgsql'):
```
ekohl, line 33: It's a bit odd you have a constant for `/var/lib/pgsql/data/` but here you hardcode the path. Should this get the parent from `POSTGRESQL_DATA_PATH`?

repos/system_upgrade/el7toel8/actors/satelliteupgrader/actor.py
ekohl: My first thought with that: does that implement AND or OR? Would that be an option?

vinzenz: Well you would still use has_package, but within the facts actor, and gather all necessary information there. 
Later actors will just consume already collected information.

e.g. if there's no message at all, then the actors can assume they don't need to do anything and just return, as there's no satellite. 

Something like this:
```python
class SatellitePostgresqlFacts(Model):
     topic = SystemFactsTopic

    same_partition = fields.Boolean(...)
    """ Whether or not target and source postgresql data will stay on the same partition """
    space_required = fields.Nullable(fields.Integer(...))
    """ How many bytes are required on the target partition """
    space_available = fields.Nullable(fields.Integer(...))
    """ How many bytes are available on the target partition """

class SatelliteFacts(Model):
     topic = SystemFactsTopic
     
    has_foreman = fields.Boolean(...)
    postgresql = fields.Model(SatellitePostgresqlFacts, ...)
    """ Foreman related PostgreSQL facts """
```

Then you can do this:
```Python
self.produce(SatelliteFacts(
    has_foreman=(has_package(InstalledRPM, 'foreman') or has_package(InstalledRPM, 'foreman-proxy'),
    postgresql=SatellitePostgresqlFacts(
        same_partition=on_same_partition
        space_required=None if on_same_partition else bytes_required,
        space_available=None if on_same_partition else bytes_available,
   ),
))
```

repos/system_upgrade/el7toel8/actors/satelliteupgradedatamigration/actor.py
evgeni: in reality, this part is probably not needed anymore now that we properly upgrade SCL PostgreSQL to modular PostgreSQL

repos/system_upgrade/el7toel8/actors/satelliteupgradedatamigration/actor.py
```diff
@@ -0,0 +1,35 @@
...
+    tags = (IPUWorkflowTag, ApplicationsPhaseTag)
+
+    def process(self):
+        has_foreman = has_package(InstalledRPM, 'foreman') or has_package(InstalledRPM, 'foreman-proxy')
+        if not has_foreman:
+            return
+
+        # remove empty PostgreSQL data from the package
+        if os.path.exists(POSTGRESQL_DATA_PATH):
+            os.rmdir(POSTGRESQL_DATA_PATH)
```
ekohl, line 32: Should this removal only happen if `POSTGRESQL_SCL_DATA_PATH` exists? If we don't need to move it, there's no reason to `rmdir`.

evgeni: nah, I plan to guard that all by "if satellite_local_postgresql", just didn't yet.

evgeni:
@vinzenz I think I've addressed all of your comments, would you mind having another look please?

repos/system_upgrade/el7toel8/actors/satelliteupgrader/actor.py
vinzenz: To group postgresql related items together. It wasn't meant to be used as a message but as a way of structuring the data. Models aren't all messages. We have been talking about this within the team already, that we want eventually mark messages as such, by having them derive from a different base class. To make clear which Models are intended to be used as messages and which are supposed to be just solely meant to structure the data.

repos/system_upgrade/el7toel8/libraries/satelliteutils.py
```diff
@@ -0,0 +1,23 @@
...
+from leapp.libraries.common import mounting
+from leapp.libraries.stdlib import api, CalledProcessError
+
+
+def apply_postgresql_workaround(context=None):
+    """
+    Applies a workaround on the system to allow the upgrade to succeed for PostgreSQL.
+    """
+    script_path = api.get_tool_path('handle-postgresql-legacy-actions')
+    cmd = ['/bin/bash', '-c', script_path]
```
vinzenz, line 11: One of the things I don't like about this, is that this is now executed always. I am still thinking of how to best mitigate this. It should be only executed if satellite is installed. That's one of the 'issues' with this workaround.
You're now also executing the the removal of that folder if postgresql (kind of any that has a symlink there) is installed even without any satellite around. 
I have to think of something.

vinzenz: Just be sure, there's nothing wrong with your implementation - I am just trying to find a good way to not execute it if there's no satellite. Without having to consume the satellite message everywhere.

repos/system_upgrade/el7toel8/actors/satelliteupgrader/actor.py
evgeni: thanks for the explanation!

evgeni:
```
ImportError: cannot import name 'satelliteutils' from 'leapp.libraries.common' (/payload/tut/lib/python3.9/site-packages/leapp/libraries/common/__init__.py)
```

:confused:

Rezney:
@evgeni Hi, what is status of this PR? Is it still WIP? Thanks...

evgeni:
@Rezney it is not, sorry, forgot to drop the label

repos/system_upgrade/el7toel8/actors/satelliteupgradefacts/tests/unit_test_satelliteupgradefacts.py
```diff
@@ -0,0 +1,89 @@
...
+
+def fake_package(pkg_name):
+    return RPM(name=pkg_name, version='0.1', release='1.sm01', epoch='1', packager=RH_PACKAGER, arch='noarch',
+               pgpsig='RSA/SHA256, Mon 01 Jan 1970 00:00:00 AM -03, Key ID 199e2f91fd431d51')
+
+
+foreman_RPM = fake_package('foreman')
+foremanproxy_RPM = fake_package('foreman-proxy')
+katello_RPM = fake_package('katello')
+postgresql_RPM = fake_package('rh-postgresql12-postgresql-server')
```
vinzenz, line 17: Global 'constants' are supposed to be all upper case.

repos/system_upgrade/el7toel8/actors/satelliteupgrader/actor.py
```diff
@@ -0,0 +1,27 @@
...
+    consumes = (SatelliteFacts, )
+    produces = ()
+    tags = (IPUWorkflowTag, FirstBootPhaseTag)
+
+    def process(self):
+        facts = next(self.consume(SatelliteFacts), None)
+        if not facts or not facts.has_foreman:
+            return
+
+        api.current_logger().info('Running the installer. This can take a while.')
```
vinzenz, line 22:
```suggestion
        api.current_actor().show_message('Running the installer. This can take a while.')
```

repos/system_upgrade/common/libraries/testutils.py
```diff
@@ -21,13 +21,17 @@ class create_report_mocked(object):
     def __init__(self):
         self.called = 0
         self.report_fields = {}
```
evgeni, line 23: technically, `report_fields` can become a `@property` returning `self.reports[-1]`

repos/system_upgrade/el7toel8/actors/satelliteupgrader/actor.py
```diff
@@ -0,0 +1,27 @@
...
+        try:
+            run(['foreman-installer'])
+        except CalledProcessError:
+            err_msg = 'Could not run the installer, please inspect the logs in /var/log/foreman-installer!'
+            api.current_logger().error(err_msg)
```
vinzenz, line 27:
```suggestion
        except OSError as e:
            api.current_logger().error('Failed to run `foreman-installer`: {}'.format(str(e))
        except CalledProcessError:
            api.current_logger().error(
                'Could not run the installer, please inspect the logs in /var/log/foreman-installer!'
            )
```

repos/system_upgrade/el7toel8/actors/satelliteupgradefacts/actor.py
fernflower: Technically you could create a shared library to share code between 2 actors (like `repos/system_upgrade/el7toel8/libraries/satelliteupgradeutils.py` and then import it as `import leapp.libraries.common.satelliteupgradeutils as lib`). But for one single constant that is used for report generation only this may be an overkill, so up to you.

Anyway, could you please make this string a constant 
`POSTGRESQL_SCL_DATA_PATH = '/var/opt/rh/rh-postgresql12/lib/pgsql/data/'` as well?

repos/system_upgrade/el7toel8/actors/satelliteupgrader/actor.py
evgeni: In which case would this raise an `OSError` (just curious)?

repos/system_upgrade/el7toel8/actors/satelliteupgradecheck/actor.py
```diff
@@ -0,0 +1,22 @@
...
+from leapp.models import Report, SatelliteFacts
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class SatelliteUpgradeCheck(Actor):
+    """
+    Check state of Satellite system before upgrade
+    """
+
+    name = 'satelliteupgradecheck'
```
fernflower, line 12: Could you please change it to 'satellite_upgrade_check' for better readability?

repos/system_upgrade/el7toel8/actors/satelliteupgradecheck/actor.py
```diff
@@ -0,0 +1,22 @@
...
+    consumes = (SatelliteFacts,)
+    produces = (Report,)
+    tags = (IPUWorkflowTag, ChecksPhaseTag)
+
+    def process(self):
+        facts = next(self.consume(SatelliteFacts), None)
+        if not facts or not facts.has_foreman:
+            return
+
+        satelliteupgradecheck(facts)
```
fernflower, line 22: Same here, could you please rename the module to have underscores `satellite_upgrade_check`? Sorry I missed that in the first review.

fernflower (commented):
Don't see anything else, just couple of super minor things. lgtm

repos/system_upgrade/el7toel8/actors/satelliteupgradecheck/actor.py
vinzenz: This is autogenerated

repos/system_upgrade/el7toel8/actors/satelliteupgradecheck/actor.py
vinzenz: You didn't we refactored it today, that's why

repos/system_upgrade/el7toel8/actors/satelliteupgrader/actor.py
vinzenz: Problems with access rights (e.g. not +x), missing binary, anything before the process actually can be created

evgeni:
Thanks @fernflower and @vinzenz!
Incorporated all the latest suggestions and the tests still pass :sweat_smile:

vinzenz (approved):

fernflower (dismissed):
lgtm, thanks for your work and patience.

evgeni:
hah, dropped one last FIXME, which I forgot was still in there \o/

fernflower (approved):

fernflower:
No, it's not related. Awesome work, merging

PR #736, opened 2021-10-10
Enable support for SAP upgrades on AWS using RHUI

Rezney:
Adds RHUI specific data needed for RHEL 7 > RHEL 8 upgrade using RHUI. Special "leapp-rhui-aws-sap-e4s" package is needed in the RHEL 7 client repository.


MichalHe:
Hi, yes, that is all that needs to be done in order to set the target channel from the code.

(Just a note) The only thing to be aware of is to set the variable early enough (before `PESEventsScanner` or `SetupTargetRepos` is executed).

PR #739, opened 2021-10-12
Enable support for SAP upgrades on Azure using RHUI

Rezney:
Adds RHUI specific data needed for RHEL 7 > RHEL 8 upgrade using RHUI. Special "leapp-rhui-azure-sap" package is needed in the RHEL 7 client repository.


alexxa:
LGTM. I was able to upgrade an Azure SAP 7.9 instance with SAP installed to 8.2 successfully.
Thanks @Rezney

PR #741, opened 2021-10-12
Disable database sync before reboot by default

vinzenz:
This patch sets the LEAPP_DEVEL_DATABASE_SYNC_OFF by default for
`preupgrade` and for `upgrade` for the pre-reboot part of the upgrade,
unless the environment variable has been set already by the user.

### DIFF
--- a/commands/upgrade/util.py
+++ b/commands/upgrade/util.py
@@ -1,3 +1,4 @@
+import functools
 import itertools
 import json
 import os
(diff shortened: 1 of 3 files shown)


commands/upgrade/util.py
```diff
@@ -8,11 +9,30 @@
...
 from leapp.utils.output import report_unsupported
 from leapp.utils.report import fetch_upgrade_report_messages, generate_report_file
 
 
+def disable_database_sync():
+    def disable_db_sync_decorator(f):
+        @functools.wraps(f)
+        def wrapper(*args, **kwargs):
+            try:
+                saved = os.environ.get('LEAPP_DEVEL_DATABASE_SYNC_OFF', None)
```
fernflower, line 23: Nit: os.environ.get('LEAPP_DEVEL_DATABASE_SYNC_OFF') should return None by default

fernflower (commented):
Could you please rebase that against master?

commands/upgrade/util.py
vinzenz: Yes it does, but it is more obvious to the reader what you expect it to return.

pirat89 (approved):
tested manually. the sync is disabled before the first reboot. after the reboot it's enabled. LEAPP_DEVEL_... envar is not recorded by any actor unless it's set directly by user

PR #744, opened 2021-10-14
Modify the supported upgrade path for RHEL 7.9 systems running SAP HANA

Rezney:
Modify the supported upgrade path for RHEL 7.9 systems running SAP HANA from RHEL 8.4 to RHEL 8.2


pirat89:
Merging after the discussion with @Rezney

PR #745, opened 2021-10-18
Fix: Random crashes during the output decoding of dnf commands

vinzenz:
Previously the output of DNF causes crashes during the upgrade.
This was caused by incomplete UTF-8 byte sequences being decoded on the
fly, for this purpose we're now starting to use an incremental decoder,
that should solve the problem.


Rezney (dismissed):
Looks good to me. Tested and it works fine. @mmacura311 @mkluson can we run this PR also against non-Intel archs? Thanks...

pirat89:
@vinzenz  rebase please

Rezney (approved):

PR #746, opened 2021-10-18
Add CentOS signed packages support

andrewlukoshko:
This PR adds CentOS 7 and 8 keys as valid


evgeni:
I was pondering about LEAPP and CentOS a bit, and after thinking about this issue a bit, came up with the following alternative:
https://github.com/oamg/leapp-repository/pull/876

What do you think?

fernflower:
Do we still need this PR?

pirat89:
I think we could close it as it will be covered by https://github.com/oamg/leapp-repository/pull/876. @andrewlukoshko  wdyt?

andrewlukoshko:
Sure. https://github.com/oamg/leapp-repository/pull/876 is a better way to do this.

fernflower:
Thanks! So closing.

PR #753, opened 2021-11-02
Remove upgrade_paths_map specification from code

fernflower:
This one should be loaded from some config.
For POC will be loading from leapp.conf that should have some
form of defaults, later will think of a better place to store that data.

Depends-On: https://github.com/oamg/leapp/pull/742


fernflower:
Looks like after transition to running test via github action leapp dependency hack is not working properly. Instead of expected
```
if [[ ! -z "740" ]] ; then \
	echo "Leapp-repository depends on the yet unmerged pr of the framework #740, installing it.." && \
	tut/bin/pip install -I "git+https://github.com/oamg/leapp.git@refs/pull/740/head"; \
fi
Leapp-repository depends on the yet unmerged pr of the framework #740, installing it..
```
Getting 
```
 if [[ ! -z "" ]] ; then \
	echo "Leapp-repository depends on the yet unmerged pr of the framework #, installing it.." && \
	tut/bin/pip install -I "git+https://github.com/oamg/leapp.git@refs/pull//head"; \
fi
python utils/install_actor_deps.py --actor=
```
@vojtechsokol Do you have any idea why is that happening?

pirat89 (changes requested):
i am missing the config file (which is something i do not agree basically with, to be honsest, and it needs a discussion first i believe) in the pr. the pr opened for leapp is bad by design as the logic cannot be part of the framework / utility.

packaging/leapp-repository.spec
```diff
@@ -182,6 +182,7 @@ install -m 0755 -d %{buildroot}%{_sysconfdir}/leapp/repos.d/
 install -m 0755 -d %{buildroot}%{_sysconfdir}/leapp/transaction/
 install -m 0755 -d %{buildroot}%{_sysconfdir}/leapp/files/
 install -m 0644 etc/leapp/transaction/* %{buildroot}%{_sysconfdir}/leapp/transaction
+install -m 0644 -p upgrade_paths.json %{buildroot}%{_sysconfdir}/leapp/files/upgrade_paths.json
```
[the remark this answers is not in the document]
pirat89: @fernflower you are missing the the file in the `%files` section. If you check the logs from the build, for sure there is a warning or error log speaking about "installed" files not mentioned in the `%files` section.

repos/system_upgrade/common/actors/ipuworkflowconfig/actor.py
```diff
@@ -18,8 +19,8 @@ class IPUWorkflowConfig(Actor):
     tags = (IPUWorkflowTag,)
 
     def process(self):
-        flavour = ipuworkflowconfig.get_upgrade_flavour()
-        target_version = ipuworkflowconfig.get_target_version(flavour)
+        flavour = upgrade_path.get_upgrade_flavour()
```
fernflower, line 22: @pirat89 Wondering if I should move all code from process into the library and leave here just a single func call. WDYT?

pirat89 (dismissed):
The current proposal basically breaks the design. there is not need for any other actor to access the "upgrade_path" library and worse, it tells actors can access content which is not persistent (depends on an actor calling functions from the library, it's possible to obtain different values during the upgrade process). The only actor that makes sense to call these functions is the IPUWorkflowConfig actor so moving any of these functions from the original private library basically doesn make sense.

fernflower:
Will put there some details to clarify the problem for other reviewers:
Main concern here that my patch introduces is the non-persistent values returned by get_target_release function on source and destination machine. Now no other actor uses the library, but they in theory somebody may unknowingly use this now public library one day for some actor, and if it is running on upgraded system will get non-expected value.

Although there are plenty of ifs in the description of a potential problem, after discussion with @pirat89 I am not sure we should leave that for reviewers only to catch. If public library way is to go some safeguard should be implemented.

conftest.py
```diff
@@ -33,7 +33,7 @@ def _load_and_add_repo(manager, repo_path):
 
 def pytest_collectstart(collector):
     if collector.nodeid:
-        current_repo_basedir = find_repository_basedir(str(collector.fspath))
+        current_repo_basedir = find_repository_basedir(str(collector.fspath)) or '.'
```
vinzenz, line 36:
```suggestion
        current_repo_basedir = find_repository_basedir(str(collector.fspath))
        if not current_repo_basedir:
            return
```

commands/upgrade/upgrade_path.py
```diff
@@ -0,0 +1,124 @@
...
+LEAPP_UPGRADE_FLAVOUR_SAP_HANA = 'saphana'
+# XXX FIXME NOTE(ivasilev) Need to agree on a hardcoded path to store the file (and package it properly)
+# or implement config file for leapp-repository.
+# We can't use api.get_common_file_path functions here as there is no actor context
+# XXX FIXME Hardcoding just for the sake of testing to deal with other issues
+LEAPP_UPGRADE_PATHS = 'upgrade_paths.json'
+
+
+# XXX FIXME NOTE(ivasilev) Shamelessly copied from leapp.libraries.common.config as could not include shared actor
+# library at the command level
```
pirat89, line 21: I think that this functino is so generic the comment is not needed.

fernflower: I removed XXX FIXME but left the note - imho it's good to know that actor library can't be loaded at the command level.

fernflower:
py3.6 Unit tests failures should be fixed by https://github.com/oamg/leapp-repository/pull/766

fernflower:
Argh, e2e tests has broken leapp dependency - it failed because of leapp*PR742* leapp*master* was installed.
@vojtechsokol could you take a look? That's obviously some regression.

fernflower:
@pirat89 Please review

pirat89:
@fernflower  I will get to it tonight/tomorrow

commands/preupgrade/__init__.py
```diff
@@ -23,6 +23,11 @@
              help='Set preferred channel for the IPU target.',
              choices=['ga', 'tuv', 'e4s', 'eus', 'aus'],
              value_type=str.lower)  # This allows the choices to be case insensitive
+@command_opt('flavour', default=upgrade_path.LEAPP_UPGRADE_FLAVOUR_DEFAULT,
+             choices=[upgrade_path.LEAPP_UPGRADE_FLAVOUR_DEFAULT, upgrade_path.LEAPP_UPGRADE_FLAVOUR_SAP_HANA],
+             help='Specify upgrade path flavor')
+# NOTE(ivasilev) choices argument here will depend on flavor, so not specifying defaults
```
pirat89, line 29: @fernflower why the falvour option? I thought this is going to be detected automatically only.

commands/preupgrade/__init__.py
```diff
@@ -23,6 +23,11 @@
...
+@command_opt('to', help='Specify RHEL version to upgrade to')
```
pirat89, line 30: We are using `source`/`original` and `target` keywords everywhere - presentations, docs, reports, .... let's stick to it. `to` is not so much descriptive.

```suggestion
@command_opt('target', help='Specify RHEL version to upgrade to')
```

commands/preupgrade/__init__.py
fernflower: Yeah, that should be detected automatically. Don't see a case (apart some weird testing probably) when we would want to force SAP on a non sap machine. Will get rid of it, thanks.

commands/preupgrade/__init__.py
fernflower: Ok, will change.

fernflower:
leapp ci-build

Makefile
```diff
@@ -14,7 +14,7 @@ LIBRARY_PATH=
 REPORT_ARG=
 REPOSITORIES ?= $(shell ls $(_SYSUPG_REPOS) | xargs echo | tr " " ",")
 SYSUPG_TEST_PATHS=$(shell echo $(REPOSITORIES) | sed -r "s|(,\\|^)| $(_SYSUPG_REPOS)/|g")
-TEST_PATHS:=commands repos/common $(SYSUPG_TEST_PATHS)
+TEST_PATHS:=commands repos/common tests/commands $(SYSUPG_TEST_PATHS)
```
pirat89, line 17: @fernflower  move `tests/commands` under `commands/tests` and drop this change. That's the path where we expect tests for commands right now (the spec reflects that as well).

fernflower: Hmm ok, wasn't aware we had agreed on command tests path.

commands/preupgrade/__init__.py
```diff
@@ -23,6 +23,8 @@
              help='Set preferred channel for the IPU target.',
              choices=['ga', 'tuv', 'e4s', 'eus', 'aus'],
              value_type=str.lower)  # This allows the choices to be case insensitive
+# NOTE(ivasilev) choices argument here will depend on flavor, so not specifying defaults
+@command_opt('target', help='Specify RHEL version to upgrade to')
```
pirat89, line 27: @fernflower  I'd like to add choices here to be able to write what are the allowed parameters

pirat89: @fernflower ah...now I got an idea why we do not want that. e.g. because of openstack, tests, .... because in such a case, they should be able to upgrade even to an unsupported systems... maybe we should do a short brainstorming about that.

fernflower: Actually if you are referring to my comment - my comment is a leftover of --flavor parameter, as argparse doesn't allow choices of argX to be dependent on value of argY. 
It is possible to add choices argument now if we want this.

pirat89: I've been thinking about specifying the list of possible values jsut by a function, but I am afraid that people could be confused that on some systems we could print e.g. `[8.4, 8.6]` and on some systems `[8.2]` because of different detected flavours.

commands/tests/test_upgrade_paths.py
```diff
@@ -0,0 +1,45 @@
...
+def test_vet_upgrade_path(mock_open, monkeypatch):
+    monkeypatch.setattr(upgrade_path, 'get_os_release_version_id', lambda x: '8.6')
+
+    # make sure env var LEAPP_DEVEL_TARGET_RELEASE takes precedence
+    # when env var set to a bad version - abort the upgrade
+    args = mock.Mock(target='9.0')
+    monkeypatch.setenv('LEAPP_DEVEL_TARGET_RELEASE', '1.2.3shouldfail')
+    with pytest.raises(CommandError):
+        upgrade_path.vet_upgrade_path(args)
+    # no env var set, --to is set to proper version
```
pirat89, line 39:
```suggestion
    # no env var set, --target is set to proper version
```

commands/tests/test_upgrade_paths.py
```diff
@@ -0,0 +1,45 @@
...
+    monkeypatch.delenv('LEAPP_DEVEL_TARGET_RELEASE', raising=False)
+    assert upgrade_path.vet_upgrade_path(args) == ('9.0', 'default')
+    # env var is set to proper version, --to is set to a bad one - go on with the upgrade
```
pirat89, line 42:
```suggestion
    # env var is set to proper version, --target is set to a bad one - go on with the upgrade
```

commands/upgrade/upgrade_path.py
```diff
@@ -0,0 +1,124 @@
...
+    if detect_sap_hana():
+        return LEAPP_UPGRADE_FLAVOUR_SAP_HANA
+    return LEAPP_UPGRADE_FLAVOUR_DEFAULT
+
+
+def get_os_release_version_id(filepath):
+    """
+    Retrieve data about System OS release from provided file.
+
+    :return: `str` version_id
```
pirat89, line 58: @fernflower  we should maybe consolidate rules about docstrings :) currently we are using 3 docstring formats around our projects. in this file we are using two. maybe good topic for the next year :D

commands/upgrade/upgrade_path.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    Make sure the user requested upgrade_path is a supported one.
+
+    :raises: `CommandError` if the specified upgrade_path is not supported
+    :return: `tuple` (target_release, flavor)
+    """
+    flavor = get_upgrade_flavour()
+    target_release = os.getenv('LEAPP_DEVEL_TARGET_RELEASE') or args.target or get_target_version(flavor)
+    supported_target_versions = get_supported_target_versions(flavor)
+    if target_release not in supported_target_versions:
```
pirat89, line 118: I haven't checked whole code yet, but when `LEAPP_DEVEL_TARGET_RELEASE` is specified and contains data in wrong format, why should it be tested whether points to supported version? Using `LEAPP_DEVEL_TARGET_RELEASE=9.9` right now should be ok. No need for any additional check. Of course, data like `9.9.4`, `whatever`, .. should be raised as invalid.

pirat89 (changes requested):
Currently it's broken for both commands (preupgrade, upgrade)
```
[root@localhost ~]# leapp preupgrade
  File "/usr/bin/leapp", line 9, in <module>
    load_entry_point('leapp==0.13.0', 'console_scripts', 'leapp')()
  File "/usr/lib/python2.7/site-packages/leapp/cli/__init__.py", line 37, in main
    cli.command.execute('leapp version {}'.format(VERSION))
  File "/usr/lib/python2.7/site-packages/leapp/utils/clicmd.py", line 106, in execute
    args.func(args)
  File "/usr/lib/python2.7/site-packages/leapp/utils/clicmd.py", line 128, in called
    self.target(args)
  File "/usr/lib/python2.7/site-packages/leapp/cli/commands/upgrade/breadcrumbs.py", line 94, in wrapper
    return f(*args, breadcrumbs=breadcrumbs, **kwargs)
  File "/usr/lib/python2.7/site-packages/leapp/cli/commands/preupgrade/__init__.py", line 33, in preupgrade
    configuration = util.prepare_configuration(args)
  File "/usr/lib/python2.7/site-packages/leapp/cli/commands/upgrade/util.py", line 180, in prepare_configuration
    target_version, flavor = upgrade_path.vet_upgrade_path(args)
  File "/usr/lib/python2.7/site-packages/leapp/cli/commands/upgrade/upgrade_path.py", line 116, in vet_upgrade_path
    target_release = os.getenv('LEAPP_DEVEL_TARGET_RELEASE') or args.target or get_target_version(flavor)
  File "/usr/lib/python2.7/site-packages/leapp/cli/commands/upgrade/upgrade_path.py", line 104, in get_target_version
    target_versions = get_supported_target_versions(flavour)
  File "/usr/lib/python2.7/site-packages/leapp/cli/commands/upgrade/upgrade_path.py", line 92, in get_supported_target_versions
    target_versions = get_target_versions_from_config(current_version_id, flavour)
  File "/usr/lib/python2.7/site-packages/leapp/cli/commands/upgrade/upgrade_path.py", line 81, in get_target_versions_from_config
    upgrade_paths_map = get_upgrade_paths_config()
  File "/usr/lib/python2.7/site-packages/leapp/cli/commands/upgrade/upgrade_path.py", line 68, in get_upgrade_paths_config
    from leapp.utils import path  # noqa: C415; pylint: disable=import-outside-toplevel
ImportError: cannot import name path
```

commands/upgrade/upgrade_path.py
pirat89:
```
[root@localhost ~]# LEAPP_DEVEL_TARGET_RELEASE=7.1 leapp preupgrade --debug

Error: Upgrade to 7.1 for default upgrade path is not supported, possible choices are 8.4
```

In this case I would expect the warning msg in best, but upgrade should continue. Of course, it will fail in this case. but e.g. 8.3 or 8.5 should not be any problem

pirat89:
@fernflower I've added a unit test for the issue. Let's sync tomorrow over vcall. We can do a "hacking sesssion" to get to it to the final state.

vojtechsokol:
@fernflower For the failing e2e tests please see a comment I left in the issue you created in the upgrade playbook repo.

commands/tests/test_upgrade_paths.py
```diff
@@ -0,0 +1,54 @@
...
+            return_value={'default': {'7.9': ['8.4'], '8.6': ['9.0'], '7': ['8.4'], '8': ['9.0']}})
+def test_vet_upgrade_path(mock_open, monkeypatch):
+    monkeypatch.setattr(upgrade_path, 'get_os_release_version_id', lambda x: '8.6')
+
+    # make sure env var LEAPP_DEVEL_TARGET_RELEASE takes precedence
+    # when env var set to a bad version - abort the upgrade
+    args = mock.Mock(target='9.0')
+    monkeypatch.setenv('LEAPP_DEVEL_TARGET_RELEASE', '1.2.3shouldfail')
+    with pytest.raises(CommandError):
+        upgrade_path.vet_upgrade_path(args)
```
pirat89, line 38: this part of the test is correct. in case of invalid data format, the error should be still raised. But with explanation that "X.Y" format is expected.

commands/tests/test_upgrade_paths.py
```diff
@@ -0,0 +1,54 @@
...
+
+    monkeypatch.setenv('LEAPP_DEVEL_TARGET_RELEASE', '1.2.3')
+    assert upgrade_path.get_target_version('default') == '9.0'
+
+    monkeypatch.delenv('LEAPP_DEVEL_TARGET_RELEASE', raising=True)
+    # unsupported path
+    monkeypatch.setattr(upgrade_path, 'get_os_release_version_id', lambda x: '8.5')
+    assert upgrade_path.get_target_version('default') == '9.0'
+
+
```
pirat89, line 27: qw ahouls have a test for different flavour as well. at least a simple one, so it's visible that not everything works only for "default".

commands/preupgrade/__init__.py
pirat89: @Rezney @fernflower  @vinzenz @shaded-enmity : btw, still not sure about printing the choices. I mean from the point that based on detected flavours, people will see different choices. or maybe we could have even "dynamic" help. I mean. description detected flavour and than the choices.. something like:
```
Specify the target version of RHEL system. See URL  for the list of supported upgrade paths. For your system, we detected {} "flavour" (different flavours could mean different possible choices where to upgrade).
```
Take that just as an idea - definitely this wording is not good. E.g. how to explain what the flavour is..

repos/system_upgrade/common/actors/ipuworkflowconfig/libraries/ipuworkflowconfig.py
```diff
@@ -89,41 +40,42 @@ def get_booted_kernel():
...
+        leapp_env_vars=get_env_vars(),
+        os_release=os_release,
+        architecture=platform.machine(),
+        version=Version(
+            source=os_release.version_id,
+            target=target_version
+            ),
+        kernel=get_booted_kernel(),
+        flavour=flavour
+        ))
```
pirat89, line 81: Just small reformatting if you don't mind. But it's nitpick. Keeping on you
```suggestion
        version=Version(
            source=os_release.version_id,
            target=target_version
        ),
        kernel=get_booted_kernel(),
        flavour=flavour
    ))
```

commands/preupgrade/__init__.py
Rezney: Despite feeling that the dynamic approach might not be exactly "kosher" in the context of Linux command help, this way (adding dynamic choices), we would be telling users what target system versions they can upgrade to without sending them to read documentation straight away.

What about: "Specify RHEL version to upgrade to. For this system (SAP flavour), Leapp detected the following available paths: ... ..." ?

commands/upgrade/upgrade_path.py
fernflower: I got the idea wrong then, will make LEAPP_DEVEL_TARGET_RELEASE be independent on allowed upgrade paths. Thanks

commands/tests/test_upgrade_paths.py
fernflower: I'd say it's visible for the upgrade_paths_map but ok, I'll add another flavor and a test.
Oh and if it were up to me I'd rename 'flavour' to 'flavor', but there are plenty of occurrences already.

commands/preupgrade/__init__.py
fernflower: I think of adding this to choices and generating a dynamic "Here are available choices for detected {} flavor" message. This way argparse will take care of vetting the arguments and warning the user for us. wdyt?

commands/upgrade/upgrade_path.py
fernflower: And what's wrong with this one, you mean :return: in code?

commands/upgrade/upgrade_path.py
fernflower: Should be fixed now and covered by unit tests

commands/tests/test_upgrade_paths.py
fernflower: Should be fixed

fernflower:
leapp-ci build

fernflower:
leapp-ci build

fernflower:
e2e seems stuck, retriggering

fernflower:
leapp ci-build

fernflower:
leapp ci-build

fernflower:
leapp ci-build

fernflower:
leapp ci-build

fernflower:
I have absolutely no idea why e2e is failing. I can't reproduce the failure locally, tmt tests (e2e and e2e-based morf tests) are also ok http://artifacts.osci.redhat.com/testing-farm/7a8e0994-51a0-4231-8155-bafb99b74cc9/ .
@vinzenz @vojtechsokol Do you have any clues?

vojtechsokol:
The failure was caused by old leapp being used for the upgrade. It should be already fixed and I alrady triggered a rerun, however because of the recent AWS outage it will take some to report results.

fernflower:
Integration tests for the feature - https://gitlab.cee.redhat.com/oamg/tmt-plans/-/merge_requests/18

fernflower:
leapp ci-build

fernflower:
leapp ci-build

fernflower:
@vojtechsokol Can't schedule a rerun of leapp-ci build.
The failure I see from the one you reissued manually is still incomprehensible to me :(

fernflower:
leapp-ci build

commands/command_utils.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+# NOTE(ivasilev) Borrowed from https://semver.org/
+SEMVER_REGEX = re.compile((r"^(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*)"
+                           r"(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)"
+                           r"(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?"
+                           r"(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$"))
+
+
+def check_semver(version):
```
pirat89, line 23: @fernflower my question is, why? As you know, that only X.Y are expected, why to do something like this? there is no possible to find RHEL 8.yada or 8.7.3, as far as I know. So what is the reason behind such canon?

fernflower: Well it was you who wanted version checking, no? =)
Or did you mean "check that X and Y are numbers and be done with it"?

fernflower: Should be fixed now.

Rezney:
@fernflower Great job! Looks already pretty much refined to me. Could you please rebase against the latest master so I can give it some testing (using the recently merged 8.6 certs). Thanks...

fernflower:
@Rezney Rebased, squashed commits, changed commit message a bit.

Rezney:
When I put the following 7 > 8 paths into `upgrade_paths.json`:
```
"7.9": ["8.4", "8.6"]
``` 
and do **not** use the `--target` option, then the system will get by default upgraded to `8.4`. I believe the default should be the latest, so in this case `8.6`. @shaded-enmity WDYT?

fernflower:
@Rezney that's the correct behavior - this patch does not support real N\N-1 upgrades, it is just a necessary refactor to deliver one.

fernflower:
Or actually I can add another commit (should be minor one) to deliver N\N-1. Would you prefer this done in 1 PR? @Rezney

Rezney:
@fernflower Aah, I see. Makes sense. In this case, I will leave the decision to @mkluson.

fernflower:
@Rezney I like 'no decision' choice more - merge this now and then ask @mkluson's opinion on the second part. the real N\N-1 enablement =)

Rezney:
@fernflower Fair enough. Let's wait for the `e2e` test to finish.

Rezney (approved):
Looks good to me...

evgeni:
This change means that the current `leap-repository` master builds can't be used with the released `leapp` 0.13.0 even tho `leap-repository` wants "framework 2.0" and `leapp` 0.13.0 provides it.

fernflower:
@evgeni Yeah, unfortunately that's expected - this patch requires few changes in leapp itself, that are not present in released leapp.
If you want to test bleeding edge leapp-repository you should take recent master, not the latest released.

evgeni:
Not saying this is unexpected, just that the built RPMs should express this ;)

fernflower:
> Not saying this is unexpected, just that the built RPMs should express this ;)

And that's a good point. @pirat89 what do you say, as our packaging expert?

pirat89:
@evgeni correct. it's bug of this and leapp pr. good catch!
@fernflower @evgeni is right. We did mistake when we merged these PRs so hasty. https://leapp.readthedocs.io/en/latest/infrastructure.html

PR #756, opened 2021-11-04
migratentp: Convert restrictions with IPv6 mask

mlichvar:
The ipaddress library doesn't support IPv6 networks specified with an
expanded netmask. Parse the mask as an IP address and count the bits to
specify the network with the prefix length instead.


vinzenz:
LGTM let's just check if it works with the e2e tests

vinzenz (dismissed):

fernflower:
@mlichvar Could you please rebase against master? LGTM, but I'd like to run regression tests on the change and merge it.

mlichvar:
Rebased on current master

fernflower (approved):

PR #757, opened 2021-11-08
Fetch more than a single commit in unit-tests workflow

fernflower:
The tricky leapp dependency logic in unit and e2e depend on
getting Depends-On information from the pr's commits messages.
The default github action checkout module fetches just 1 merge
commit which kind of break it.
This should bring back the old pr checkout behavior formerly used
in travis.


vojtechsokol (approved):
LGTM, will merge after status checks are green.

PR #758, opened 2021-11-08
Fix getting device attribute through udev on py3

Rezney:
On Python 3 pyudev Attribute object provides its own getter and missing
the __getitem__. Switch to use the "get" method which will work for both
py2 and py3.


pirat89:
@Rezney  is it connected any problem/bug with the original solution? If so, add please tests for the library.

Rezney:
@pirat89 Test added.

pirat89 (approved):
lgtm

fernflower (approved):
LGTM. AFAIU this doesn't require any new integration test/ QE attention and can be merged just now, right?

PR #759, opened 2021-11-18
unittests: Scenario based testing

vinzenz:
Previously we tested all repositories all together. We didn't really
care wether it was for el7toel8 or el8toel9, we even had el8toel9
being tested with python2.7 python until we deleted the el8toel9
repository within the GitHub action as a temporary measure.

This implements the feature of allowing el7toel8 and el8toel9 testing
scenarios to be implemented properly.
This means as of this patch there are 4 unit test runs:
- Python 2.7 with system_upgrade/el7toel8 and system_upgrade/common
  repos in a CentOS7 container
- Python 3.6 with system_upgrade/el7toel8 and system_upgrade/common
  repos in a CentOS8 container
- Python 3.6 with system_upgrade/el8toel9 and system_upgrade/common
  repos in a CentOS8 container
- Python 3.9 with system_upgrade/el8toel9 and system_upgrade/common
  repos in a CentOS8 container

Additionally some tests had to be altered, which either had some
problems during the execution OR were just plain wrong from the start
and had to be rewritten in order to make them being proper tests
(looking at you transactionworkarounds actor...)

On top of that pylint is now executed with -j 8 which should use 8
paralell processes to execute, which hopefully will result in faster
results.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,12 +9,19 @@ DEPS_PKGNAME=leapp-el7toel8-deps
 VERSION=`grep -m1 "^Version:" packaging/$(PKGNAME).spec | grep -om1 "[0-9].[0-9.]**"`
 DEPS_VERSION=`grep -m1 "^Version:" packaging/$(DEPS_PKGNAME).spec | grep -om1 "[0-9].[0-9.]**"`
 REPOS_PATH=repos
-ACTOR_PATH=
 LIBRARY_PATH=
 REPORT_ARG=
+space :=
+space += 
+comma := ,
+REPOSITORIES ?= $(shell ls repos/system_upgrade/ | xargs echo | tr ' ' ',')
+SYSUPG_TEST_PATHS=repos/system_upgrade/$(subst $(comma),$(space)repos/system_upgrade/,$(REPOSITORIES))
+TEST_PATHS:=repos/common
+TEST_PATHS += ${SYSUPG_TEST_PATHS}
+
 
 ifdef ACTOR
-	ACTOR_PATH=`python utils/actor_path.py $(ACTOR)`
+	TEST_PATHS=`python utils/actor_path.py $(ACTOR)`
 endif
 
 ifeq ($(TEST_LIBS),y)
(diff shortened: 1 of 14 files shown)


Makefile
```diff
@@ -242,18 +249,18 @@ install-deps-fedora:
 lint:
 	. $(VENVNAME)/bin/activate; \
 	echo "--- Linting ... ---" && \
-	SEARCH_PATH=$(REPOS_PATH) && \
+	SEARCH_PATH="$(TEST_PATHS)" && \
 	echo "Using search path '$${SEARCH_PATH}'" && \
 	echo "--- Running pylint ---" && \
-	bash -c "[[ ! -z $${SEARCH_PATH} ]] && find $${SEARCH_PATH} -name '*.py' | sort -u | xargs pylint" && \
+	bash -c "[[ ! -z '$${SEARCH_PATH}' ]] && find $${SEARCH_PATH} -name '*.py' | sort -u | xargs pylint -j 8" && \
```
pirat89, line 255: @vinzenz  what about to specify rather `0`, so pylint itself discover the number of processors that can be used? Seems to me better for machines that have just 4 processors.
```suggestion
	bash -c "[[ ! -z '$${SEARCH_PATH}' ]] && find $${SEARCH_PATH} -name '*.py' | sort -u | xargs pylint -j 0" && \
```

pirat89:
@vinzenz is it still WIP or it could be merged? :) Or are we moving it to #760  ?

vinzenz:
No, no moving to #760 - #760 is just based on this one. Still this needs a final approval.

repos/system_upgrade/common/actors/selinux/selinuxapplycustom/Makefile
```diff
@@ -1,2 +1,5 @@
 install-deps:
-	yum install -y policycoreutils policycoreutils-python
+	yum install -y policycoreutils
+	-yum install -y policycoreutils-python
+	-yum install -y python3-policycoreutils
```
pirat89, line 4: @vinzenz just thinking about nitpick, what about change it to:

```
yum install -y policycoreutils-python || yum install -y python3-policycoreutils
```
So python-3policycoreutils is tried to be installed only when policycoreutils-python has not been installed? As well, in this way we are sure that at least on of the really has been installed. In the current way, we could continue even when the expected package has not been installed eg. because of troubles with networking (I expect that we are speaking just about rare corner case, my point mainly about saving another couple of seconds for py2 systems). It's really nitpick, so I am ok if you want to keep it as it is.

pirat89: @vinzenz mybe even better, the whole block to:
```
yum install -y policycoreutils policycoreutils-python || yum install -y policycoreutils python3-policycoreutils
```

That will save even more time :)

vinzenz: Well, this is not skipping the next line, it continues with it and if would fail.
'-' just ignores the failure

vinzenz: Will apply in #760 as discussed offline

pirat89 (approved):

PR #760, opened 2021-11-18
UBI based containers during unit tests with podman

vinzenz:
This patch moves from Docker to podman and from centos images to UBI
base images.


vinzenz:
@pirat89 @fernflower @Rezney 
WDYT? 
This makes the tests right on the rhel universal base images, so we have a more proper environment with the right patch levels.

vinzenz:
> @pirat89 @fernflower @Rezney WDYT? This makes the tests right on the rhel universal base images, so we have a more proper environment with the right patch levels.

Just to make it clear i am mainly interested in your opinion on the last commit, which implements the change from CentOS base images to ubi images

fernflower (dismissed):
Fine by me.

Rezney:
> > @pirat89 @fernflower @Rezney WDYT? This makes the tests right on the rhel universal base images, so we have a more proper environment with the right patch levels.
> 
> Just to make it clear i am mainly interested in your opinion on the last commit, which implements the change from CentOS base images to ubi images

I like it...

utils/docker-tests/Dockerfile.centos7
```diff
@@ -12,11 +12,14 @@ RUN yum update -y && \
...
-RUN yum -y install epel-release && \
-    yum -y install python2-pip python3-pip && \
-    python2 -m pip install --upgrade pip==20.3.4 && \
-    python3 -m pip install --upgrade pip==20.3.4
+# This is some trickery: We install python27-python-pip from the scl, use the scl to bootstrap the python
+# module of pip version 20.3.0 and then make it update to 20.3.4 resulting the 'pip' command to be available.
+# The --target approach doesn't add it, but at least we now have pip 20.3.4 installed ;-)
+RUN yum -y install python27-python-pip && \
+    scl enable python27 -- pip install -U --target /usr/lib/python2.7/site-packages/ pip==20.3.0 virtualenv && \
+    python -m pip install --ignore-installed pip==20.3.4 ipaddress
```
pirat89, line 22: Nice trick!

vinzenz:
The problem is that we install things with --update which will try to update all the things, including the python-virtualenv package, which it cannot update. --ignore-installed solved that outside.

fernflower (approved):
Thanks a lot for solving commands installation problem!

fernflower (dismissed):

packaging/leapp-repository.spec
```diff
@@ -227,6 +227,7 @@ done;
 %dir %{repositorydir}
 %dir %{custom_repositorydir}
 %dir %{leapp_python_sitelib}/leapp/cli/commands
+%exclude %{leapp_python_sitelib}/leapp/cli/commands/tests
```
pirat89, line 230: @vinzenz @fernflower  this will result in possible problems as rpmbuild will report some "installed" files are not part of any rpm. Instead of this, put the removal of the directory into the `%install` section as we do for other tests directories. Currently it's not visible as the directory doesn't exist yet.

IOW, when the `%exclude` cmd is used, the files should be present in another RPM. So the removal from the installation dirs is the better way.
Putting this on the line 189 does the thing:
```
rm -rf %{buildroot}%{leapp_python_sitelib}/leapp/cli/commands/tests
```

pirat89 (changes requested):
Needed one change in the spec file. It's not blocking error anyway as it most likely it just creates warnings, but let's rather omit that warnings too.

packaging/leapp-repository.spec
vinzenz: updated, thanks!

pirat89 (approved):

pirat89 (approved):
added fix for some systems to be able to execute `make test` locally

PR #763, opened 2021-11-24
unit-tests: Turn off fail-fast

vinzenz:
Previously we used the fail-fast default value for GitHub Actions.
This leads to all still running unit tests to be cancelled when one
fails.
With this option one can see the results for all test scenarios, even if
one fails.


fernflower (approved):

pirat89 (approved):

vojtechsokol:
leapp-ci build

PR #764, opened 2021-11-24
Do not crash when the next EFI boot cannot be set (hotfix)

pirat89:
We want to set the EFI next-boot entry to be the current booted one
as the default boot entry could be a special "handler" for machines.
It could be used e.g. on machines that are part of testing
infrastructure, to service them (e.g. put them back to the pool of
available machines, ...).

However, for IPU 8 -> 9 we have reported error on Hyper-V, where we
are trying to set the next boot to a record that doesn't exist
anymore. Which is weird as we obtain the record from the efibootmgr
itself just before we try to set it. Possibly there is a bug in the
efibootmgr, not sure. For now, print warning when we cannot set the
next boot with additional information available, but do not crash
in such a case.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2024059

UPDATE: just for info, we need to bindmount specific /sys/... files into the container to fix the problem. so ignore this solution for now, I will fix the PR later this week (or someone else can open a new PR :))

### DIFF
--- a/repos/system_upgrade/common/libraries/efi_reboot_fix.py
+++ b/repos/system_upgrade/common/libraries/efi_reboot_fix.py
@@ -1,7 +1,7 @@
 from re import compile as regexp
 import os
 
-from leapp.libraries.stdlib import run
+from leapp.libraries.stdlib import api, run
 
 
 _current_boot_matcher = regexp(r'BootCurrent: (?P<boot_current>([0-9A-F]*))')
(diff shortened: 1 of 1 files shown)


pirat89:
We have some additional data and seems the proposed hotfix is not good hotfix to be honest, but just mask of the problem.
```
[  693.095850] upgrade[20169]: 2021-11-24 19:13:03.471 DEBUG    PID: 357 leapp.workflow.Finalization.efi_finalization_fix: External command has started: ['/sbin/efibootmgr', '-n', '0005']
[  693.183244] upgrade[20169]: Could not set BootNext: Read-only file system
[  693.224369] upgrade[20169]: 2021-11-24 19:13:03.672 DEBUG    PID: 357 leapp.workflow.Finalization.efi_finalization_fix: Command ['/sbin/efibootmgr', '-n', '0005'] failed with exit code 13.
[  693.328120] upgrade[20169]: 2021-11-24 19:13:03.759 DEBUG    PID: 357 leapp.workflow.Finalization.efi_finalization_fix: External command has finished: ['/sbin/efibootmgr', '-n', '0005']
```

Regarding that, I am not a fan of this hotfix anymore. We should rather investigate what's the problem with the FS.

fernflower:
Following the discussion in the bugzilla - you are going to switch that hotfix to a real fix (mounting efivars into the container), right?

pirat89:
@fernflower yep. or whoever who will get to it earlier than me :) feel free to do it. for sure, we need to test it with EFI on intel. I expect this is generic issue for all EFI right now - not just EFI on hyper-v

Rezney:
Closing, this will be handled in a different fashion.

Issue #20, opened 2018-12-11
Invalid SELinux model and provided information in case of disabled selinux


pirat89:
The `SELinux` model does not reflect correctly what information we can get about SElinux with various setup. E.g. in case SElinux is disabled, we cannot get information about runtime mode (just `None`). Policy cannot contain whatever string. It can contain only these values:
- **targeted** - Targeted processes are protected,
- **minimum** - Modification of targeted policy. Only selected processes are protected. 
- **mls** - Multi Level Security protection.

Regarding the `/etc/selinux/config` file. The model will be changed partially by the PR I will add here later just to fix broken tests. Proper fix will be still need.

- [x] resolve issue
- [x] write tests

pirat89:
Possible solution would be detection of disabled selinux in the [get_selinux_status](https://github.com/oamg/leapp-repository/blob/99049192172b1af21f7c2a7b4f9dc83cfd6ceda1/repos/system_upgrade/el7toel8/actors/systemfacts/libraries/systemfacts.py#L207) function and in such case return SELinux object with values reflecting the state.

radomil:
This issue is addressed in #27

radomil:
This issue has been fixed by above mentioned PR #27 merge and tests were covered by PR #56 merge.

PR #769, opened 2021-11-30
Apply a new way of disabling SELinux

Rezney:
On RHEL 9, the only way to completely disable SELinux will be through the kernel commandline using selinux=0 parameter. Disabling in the "/etc/selinux/config" is no longer supported.

### DIFF
--- a/repos/system_upgrade/common/actors/checkselinux/actor.py
+++ b/repos/system_upgrade/common/actors/checkselinux/actor.py
@@ -1,8 +1,7 @@
 from leapp.actors import Actor
+from leapp.libraries.actor import checkselinux
+from leapp.models import KernelCmdlineArg, Report, SELinuxFacts, SelinuxPermissiveDecision, SelinuxRelabelDecision
 from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
-from leapp.models import SELinuxFacts, SelinuxPermissiveDecision, SelinuxRelabelDecision
-from leapp.reporting import Report, create_report
-from leapp import reporting
 
 
 class CheckSelinux(Actor):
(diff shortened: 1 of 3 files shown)


fernflower:
This needs rebase.

repos/system_upgrade/common/actors/checkselinux/libraries/checkselinux.py
```diff
@@ -0,0 +1,69 @@
...
+                reporting.Severity(reporting.Severity.INFO),
+                reporting.Tags([reporting.Tags.SELINUX]),
+                reporting.RelatedResource('file', '/etc/selinux/config'),
+                reporting.ExternalLink(url=DOC_URL, title='Disabling SELinux'),
+            ])
+
+        if enabled:
+            reporting.create_report([
+                reporting.Title('SElinux disabled in configuration file but currently enabled'),
+                reporting.Summary('This message is to inform user about non-standard SElinux configuration.'),
```
pirat89, line 37: this is bad msg. the summary basically is not telling user why the situation happened and what they are expected to do.

repos/system_upgrade/common/actors/checkselinux/libraries/checkselinux.py
```diff
@@ -0,0 +1,69 @@
...
+                ),
+                reporting.Severity(reporting.Severity.INFO),
+                reporting.Tags([reporting.Tags.SELINUX]),
+                reporting.RelatedResource('file', '/etc/selinux/config'),
+                reporting.ExternalLink(url=DOC_URL, title='Disabling SELinux'),
+            ])
+
+        if enabled:
+            reporting.create_report([
+                reporting.Title('SElinux disabled in configuration file but currently enabled'),
```
pirat89, line 36: it sounds weird. what about:

```suggestion
                reporting.Title('SElinux should be disabled based on the configuration file but it is enabled'),
```

repos/system_upgrade/common/actors/checkselinux/libraries/checkselinux.py
```diff
@@ -0,0 +1,69 @@
...
+        reporting.create_report([
+            reporting.Title('SElinux disabled'),
+            reporting.Summary('SElinux disabled, continuing...'),
+            reporting.Tags([reporting.Tags.SELINUX, reporting.Tags.SECURITY])
+        ])
+        return
+
+    if conf_status in ('enforcing', 'permissive'):
+        api.produce(SelinuxRelabelDecision(
+            set_relabel=True))
```
pirat89, line 50:
```suggestion
        api.produce(SelinuxRelabelDecision(set_relabel=True))
```

repos/system_upgrade/common/actors/checkselinux/libraries/checkselinux.py
pirat89: btw, very same should be possibly detected regardint `enforcing=...` parameter in the kernel cmdline, as the parameter has higher prio than the configuration file as well. In such a case, the msg could be updated to be more generic.

repos/system_upgrade/common/actors/checkselinux/libraries/checkselinux.py
Rezney: This is not change of this PR. It was just refactored into library. Wondering whether we want to change a report message which is already out but, mmm, why not...

repos/system_upgrade/common/actors/checkselinux/libraries/checkselinux.py
Rezney: Same as my previous reaction.

repos/system_upgrade/common/actors/checkselinux/actor.py
```diff
@@ -15,50 +14,13 @@ class CheckSelinux(Actor):
 
     name = 'check_se_linux'
     consumes = (SELinuxFacts,)
-    produces = (Report, SelinuxPermissiveDecision, SelinuxRelabelDecision)
+    produces = (
+        Report,
+        SelinuxPermissiveDecision,
+        SelinuxRelabelDecision,
+        KernelCmdlineArg,
```
fernflower, line 21: Hmm just curious why didn't the linter catch that?

repos/system_upgrade/common/actors/checkselinux/tests/test_checkselinux.py
```diff
@@ -1,6 +1,11 @@
...
+from leapp.libraries.actor import checkselinux
+from leapp.libraries.common.testutils import create_report_mocked, CurrentActorMocked, produce_mocked
+from leapp.libraries.stdlib import api
+from leapp.models import Report, SELinuxFacts, SelinuxPermissiveDecision, SelinuxRelabelDecision
 from leapp.snactor.fixture import current_actor_context
-from leapp.models import (Report, SELinuxFacts, SelinuxPermissiveDecision,
-                          SelinuxRelabelDecision)
 
 
 # FIXME: fix the file properly regarding the fix of the issue:
```
fernflower, line 11: The mentioned issue is actually closed, is this comment still needed? https://github.com/oamg/leapp-repository/issues/20

Rezney:
@fernflower should be ok now...

fernflower (approved):
lgtm

PR #770, opened 2021-11-30
[WIP] [FOR VISIBILITY ONLY] Add flake8-isort and address existing violations

fernflower:
This should close the problem with import order once and for all.
Max line length has been set to the value from .pylintrc file.


repos/system_upgrade/common/actors/checkbootavailspace/actor.py
```diff
@@ -1,8 +1,5 @@
 from leapp.actors import Actor
-from leapp.libraries.actor.checkbootavailspace import (
-    check_avail_space_on_boot,
-    get_avail_bytes_on_boot,
-)
+from leapp.libraries.actor.checkbootavailspace import check_avail_space_on_boot, get_avail_bytes_on_boot
```
pirat89, line 2: in this case, the import is sorted already, why to do this change? I do not see so big benefit to change? I do not see this anyhow better.

repos/system_upgrade/common/actors/checkbootavailspace/tests/unit_test_checkbootavailspace.py
```diff
@@ -4,7 +4,7 @@
 from leapp.libraries.actor.checkbootavailspace import (
     MIN_AVAIL_BYTES_FOR_BOOT,
     check_avail_space_on_boot,
-    inhibit_upgrade,
+    inhibit_upgrade
```
pirat89, line 7: why this change? I prefer to skip any changes that does not do anything when speaking about alphabetic order of imports.

pirat89 (changes requested):

repos/system_upgrade/common/actors/checkbootavailspace/tests/unit_test_checkbootavailspace.py
fernflower: In general I'd prefer the trailing comma in imports tbh, but if the idea is to keep changes minimal we can tweak this rule for now.

fernflower: I am afraid it's either "make all multiline imports have a trailing comma" or "make all go without one" choice.

repos/system_upgrade/common/actors/checkbootavailspace/actor.py
fernflower: Per current configuration multi line imports are used only if single line import is not possible. I'll see if it's possible to relax the rule.

repos/system_upgrade/common/actors/cloud/checkhybridimage/tests/test_checkhybridimage.py
```diff
@@ -1,12 +1,11 @@
 import pytest
 
+from leapp import reporting
 from leapp.libraries.actor import checkhybridimage
-from leapp.libraries.common.testutils import produce_mocked, create_report_mocked, CurrentActorMocked
+from leapp.libraries.common.testutils import create_report_mocked, CurrentActorMocked, produce_mocked
 from leapp.libraries.stdlib import api
 from leapp.models import FirmwareFacts, InstalledRPM, RPM
```
fernflower, line 7: @pirat89 RPM ordering should be fine now, it's case-insensitive lexicographic importing now as you wanted.

repos/system_upgrade/common/actors/checkbootavailspace/actor.py
fernflower: Unfortunately, I could not figure out how to do this (keep both styles even when imports can be made on a single line without violating max line length setting).
Are you ok to keep the current logic if we go with a gradual rule introduction https://github.com/oamg/leapp-repository/pull/770/ ?

PR #772, opened 2021-12-02
Linters: fix isort check on master and HEAD..

pirat89:
When running "isort" check introduced in previous commit, it should
check only changed files. However, the check has been broken after
the merge in two possible ways:
  - all files has been checked on master (right now, it means always
    broken)
  - nothing has been discovered (HEAD or another branch) created
    without any change introduced against the master branch

Skip the isort check in such a cases

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -261,15 +261,17 @@ lint:
 		echo "--- Linting done. ---"; \
 	fi
 
-	. $(VENVNAME)/bin/activate; \
-	git diff $(MASTER_BRANCH) --name-only | xargs isort -c --diff || \
-	{ \
-		echo; \
-		echo "------------------------------------------------------------------------------"; \
-		echo "Hint: Apply the required changes."; \
-		echo "      Execute the following command to apply them automatically: make lint_fix"; \
-		exit 1; \
-	} && echo "--- isort check done. ---";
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -261,15 +261,17 @@ lint:
...
-	. $(VENVNAME)/bin/activate; \
-	git diff $(MASTER_BRANCH) --name-only | xargs isort -c --diff || \
-	{ \
-		echo; \
-		echo "------------------------------------------------------------------------------"; \
-		echo "Hint: Apply the required changes."; \
-		echo "      Execute the following command to apply them automatically: make lint_fix"; \
-		exit 1; \
-	} && echo "--- isort check done. ---";
+	if [[  "`git rev-parse --abbrev-ref HEAD`" != "master" ]] && [[ -n "`git diff $(MASTER_BRANCH) --name-only`" ]]; then \
```
fernflower, line 264: Do you need the second condition? 
`[[ -n "`git diff $(MASTER_BRANCH) --name-only`" ]]`

fernflower: Aha, this explains it https://github.com/oamg/leapp-repository/runs/4395200390?check_suite_focus=true

pirat89: yep. both are needed.

fernflower (approved):
lgtm, will test it with some custom pr and then merge

PR #773, opened 2021-12-02
Bindmount "/sys/firmware/efi/efivars/"

Rezney:
Bindmount "/sys/firmware/efi/efivars/" into the host container in order to avoid an issue with "efivars" special filesystem being
mounted as read-only.


fernflower:
Do you have a reproducer for the original issue? I'd like to have it covered with int. test

pirat89:
@fernflower any efi system installed is reproducer. Currently we do not have any efi int. test for 8 -> 9, but other teams have tests for it (or they found problems manually). Int. tests for EFI are goal for the next Q. We are ok to merge that without int. tests now (waiting for virt QE for the verification already; probably we will merge it today). Now let's focus on the currrent Q goals and opened PRs so we will have more free time to improve it the next Q.

lixuemin2016:
The patch is working. For details: the test build works well for generation 2 (UEFI) VM (minimum install) on Hyper-V host, it can upgrade from 8 to 9 successfully.  

Thank you so much.

fernflower (approved):

PR #777, opened 2021-12-10
actor: add inihibtor FirewalldCheckAllowZoneDrifting

erig0:
This config option does not exist in RHEL-9. So we must make sure it's
not being used.

Fixes: OAMG-4391

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/firewalldcheckallowzonedrifting/actor.py
+++ b/repos/system_upgrade/el8toel9/actors/firewalldcheckallowzonedrifting/actor.py
@@ -1,0 +1,51 @@
+from leapp import reporting
+from leapp.actors import Actor
+from leapp.models import FirewalldGlobalConfig, FirewallsFacts
+from leapp.reporting import create_report, Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class FirewalldCheckAllowZoneDrifting(Actor):
+    """
+    This actor will check if AllowZoneDrifiting=yes in firewalld.conf. This
+    option has been removed in RHEL-9 and behavior is as if
+    AllowZoneDrifiting=no.
+    """
+
+    name = 'firewalld_check_allow_zone_drifting'
+    consumes = (FirewallsFacts, FirewalldGlobalConfig)
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        # If firewalld is not enabled then don't bother the user about its
+        # configuration. This Report keys off a _default_ value and as such
+        # will trigger for all users that have not done one of the following:
+        #   - disabled firewalld
+        #   - manually set AllowZoneDrifting=no (as firewalld logs suggests)
+        #
+        for facts in self.consume(FirewallsFacts):
+            if not facts.firewalld.enabled:
+                return
+
+        for facts in self.consume(FirewalldGlobalConfig):
+            if not facts.allowzonedrifting:
+                return
+
+        create_report([
+            reporting.Title('Unsupported Firewalld Configuration In Use'),
+            reporting.Summary('Firewalld has enabled configuration option '
+                              '"{conf_key}" which has been removed in RHEL-9. '
+                              'New behavior is as if "{conf_key}" was set to "no".'.format(
+                                  conf_key='AllowZoneDrifiting')),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Tags([reporting.Tags.SANITY, reporting.Tags.FIREWALL]),
+            reporting.Flags([reporting.Flags.INHIBITOR]),
+            reporting.ExternalLink(
+                url='https://access.redhat.com/articles/4855631',
+                title='Changes in firewalld related to Zone Drifting'),
+            reporting.Remediation(
+                hint='Set AllowZoneDrifting=no in /etc/firewalld/firewalld.conf',
+                commands=[['sed -i "s/^AllowZoneDrifting=.*/AllowZoneDrifting=no/" '
+                           '/etc/firewalld/firewalld.conf']]),
+        ])
--- a/repos/system_upgrade/el8toel9/actors/firewalldcheckallowzonedrifting/tests/component_test_firewalldcheckallowzonedrifting.py
+++ b/repos/system_upgrade/el8toel9/actors/firewalldcheckallowzonedrifting/tests/component_test_firewalldcheckallowzonedrifting.py
@@ -1,0 +1,35 @@
+from leapp.models import FirewalldGlobalConfig, FirewallsFacts, FirewallStatus
+from leapp.reporting import Report
+
+
+def test_actor_firewalldcheckallowzonedrifting(current_actor_context):
+    status = FirewallStatus(enabled=True, active=True)
+    current_actor_context.feed(FirewallsFacts(firewalld=status,
+                                              iptables=status,
+                                              ip6tables=status))
+    current_actor_context.feed(FirewalldGlobalConfig(allowzonedrifting=True))
+
+    current_actor_context.run()
+    assert current_actor_context.consume(Report)
+
+
+def test_actor_firewalldcheckallowzonedrifting_negative(current_actor_context):
+    status = FirewallStatus(enabled=False, active=True)
+    current_actor_context.feed(FirewallsFacts(firewalld=status,
+                                              iptables=status,
+                                              ip6tables=status))
+    current_actor_context.feed(FirewalldGlobalConfig(allowzonedrifting=True))
+
+    current_actor_context.run()
+    assert not current_actor_context.consume(Report)
+
+
+def test_actor_firewalldcheckallowzonedrifting_negative2(current_actor_context):
+    status = FirewallStatus(enabled=True, active=True)
+    current_actor_context.feed(FirewallsFacts(firewalld=status,
+                                              iptables=status,
+                                              ip6tables=status))
+    current_actor_context.feed(FirewalldGlobalConfig(allowzonedrifting=False))
+
+    current_actor_context.run()
+    assert not current_actor_context.consume(Report)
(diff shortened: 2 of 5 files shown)


erig0:
review please

repos/system_upgrade/el8toel9/actors/firewalldcheckallowzonedrifting/actor.py
```diff
@@ -0,0 +1,52 @@
...
+    AllowZoneDrifiting=no.
+    """
+
+    name = 'firewalld_check_allow_zone_drifting'
+    consumes = (FirewallsFacts,)
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        conf_key = "AllowZoneDrifting"
```
vinzenz, line 22:
```suggestion
        conf_key = 'AllowZoneDrifting'
```

repos/system_upgrade/el8toel9/actors/firewalldcheckallowzonedrifting/actor.py
```diff
@@ -0,0 +1,52 @@
...
+
+        conf_dict = private.read_config(path)
+        if private.isAllowZoneDriftingInUse(conf_dict):
+            create_report([
+                reporting.Title("Unsupported configuration"),
+                reporting.Summary("Firewalld is using configuration option "
+                                  + "'%s' which has been removed in RHEL-9. "
+                                  % (conf_key)
+                                  + "New behavior is as if '%s' was set to 'no'."
+                                  % (conf_key)),
```
vinzenz, line 48: Please do not use % formatting, but use .format
So to get you an idea:
```python
'Starting with RHEL-9, FirewallD removed the support for the configuration option `{option}`.'.format(option=conf_key)
```

The variable '{option}' can be also used multiple times inside of the string, so no need to repeat the argument. That's the advantage of the named formatting strings.

vinzenz (changes requested):
Thank you very much for your PR, it looks quite good already, however there are a few things in general that need to be addressed.

* Please drop the usage of double quotes and replace them with single quotes.
* Please do not use % style string formatting
* The detection part, is by our convention to be done in the FactsCollection phase (FactsPhaseTag)
    * The ChecksPhaseTag should check just for message provided by the fact collection actor and generate the inhibitor based on that. 
    *  For that you will have to create a Model that you can pass that information, e.g. to make it usable for the future, it could be a list of deprecated/removed options that are used by the user, etc.

repos/system_upgrade/el8toel9/actors/firewalldcheckallowzonedrifting/tests/component_test_firewalldcheckallowzonedrifting.py
```diff
@@ -0,0 +1,58 @@
+from leapp.libraries.actor import private
```
fernflower, line 1: 'private' library name is to me a bit confusing. Maybe rename it to firewalldlib or just 'library' or something similar?

repos/system_upgrade/el8toel9/actors/firewalldcheckallowzonedrifting/actor.py
```diff
@@ -0,0 +1,52 @@
...
+                reporting.Severity(reporting.Severity.HIGH),
+                reporting.Tags([reporting.Tags.SANITY]),
```
fernflower, line 50: You can also add Tag.FIREWALL there

erig0:
> Thank you very much for your PR, it looks quite good already, however there are a few things in general that need to be addressed.
> 
>     * Please drop the usage of double quotes and replace them with single quotes.

ACK. Both are valid and there is no PEP suggesting either. That being said, if Leapp uses single quotes everywhere I will change them.

>     * Please do not use % style string formatting

ACK. Bad habit from python < 2.7.

>     * The detection part, is by our convention to be done in the FactsCollection phase (FactsPhaseTag)
>       
>       * The ChecksPhaseTag should check just for message provided by the fact collection actor and generate the inhibitor based on that.
>       * For that you will have to create a Model that you can pass that information, e.g. to make it usable for the future, it could be a list of deprecated/removed options that are used by the user, etc.

Knowledge about `AllowZoneDrifting` is only useful to this actor. I based this code off a similar actor from el7_to_el8. I suppose I can rework it to add a Model.

repos/system_upgrade/el8toel9/actors/firewalldcheckallowzonedrifting/tests/component_test_firewalldcheckallowzonedrifting.py
erig0: "private" is the convention used in the docs, but I'll change it.

repos/system_upgrade/el8toel9/actors/firewalldcheckallowzonedrifting/actor.py
erig0: For identifiers, that makes sense.

erig0:
review please.

Rezney:
@erig0 Thanks for PR. It is in a good shape now. I would just ask you to modify the report title a little. As for now it shows as:

```
    1. Inhibitor: Unsupported configuration
```

I suggest to incorporate "firewalld" into the title.

erig0:
Thanks. I folded and this change and pushed.

```python
diff --git a/repos/system_upgrade/el8toel9/actors/firewalldcheckallowzonedrifting/actor.py b/repos/system_upgrade/el8toel9/actors/firewalldcheckallowzonedrifting/actor.py
index 9f2a26a3706e..6db0b81c959d 100644
--- a/repos/system_upgrade/el8toel9/actors/firewalldcheckallowzonedrifting/actor.py
+++ b/repos/system_upgrade/el8toel9/actors/firewalldcheckallowzonedrifting/actor.py
@@ -33,7 +33,7 @@ class FirewalldCheckAllowZoneDrifting(Actor):
                 return
 
         create_report([
-            reporting.Title('Unsupported configuration'),
+            reporting.Title('Unsupported Firewalld Configuration'),
             reporting.Summary('Firewalld has enabled configuration option '
                               '"{conf_key}" which has been removed in RHEL-9. '
                               'New behavior is as if "{conf_key}" was set to "no".'.format(
```

Rezney:
@erig0 Thank you. Please do a rebase and it should be good to go.

erig0:
> @erig0 Thank you. Please do a rebase and it should be good to go.

Done.

erig0:
> > @erig0 Thank you. Please do a rebase and it should be good to go.
> 
> Done.

bump

fernflower (approved):

repos/system_upgrade/el8toel9/actors/firewalldcheckallowzonedrifting/actor.py
```diff
@@ -0,0 +1,44 @@
...
+from leapp.reporting import create_report, Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class FirewalldCheckAllowZoneDrifting(Actor):
+    '''
+    This actor will check if AllowZoneDrifiting=yes in firewalld.conf. This
+    option has been removed in RHEL-9 and behavior is as if
+    AllowZoneDrifiting=no.
+    '''
```
pirat89, line 13: @erig0 pls use triple double quotes for all docstrings
* https://leapp.readthedocs.io/en/latest/python-coding-guidelines.html#write-docstrings-that-work-well-with-tools
* https://www.python.org/dev/peps/pep-0257/

erig0: done

repos/system_upgrade/el8toel9/actors/firewalldcheckallowzonedrifting/actor.py
```diff
@@ -0,0 +1,51 @@
...
+                                  conf_key='AllowZoneDrifiting')),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Tags([reporting.Tags.SANITY, reporting.Tags.FIREWALL]),
+            reporting.Flags([reporting.Flags.INHIBITOR]),
+            reporting.ExternalLink(
+                url='https://access.redhat.com/articles/4855631',
+                title='Changes in firewalld related to Zone Drifting'),
+            reporting.Remediation(
+                hint='Set AllowZoneDrifting=no in /etc/firewalld/firewalld.conf',
+                commands=[['sh', '-c',
```
Rezney, line 49: As this is just a single command, please remove the "sh -c" and let's go with the sed command only. 

Now, in the report, it translates to:
`sh -c sed -i "s/^AllowZoneDrifting=.*/AllowZoneDrifting=no/" /etc/firewalld/firewalld.conf` which is broken without apostrophes around the sed command.
Thanks...

erig0: Oh. I thought this was doing shell escaping. I followed the `sh -c` pattern from this example: https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/actors/checkhacluster/libraries/checkhacluster.py

I'll change it to just call `sed` directly.

Is there a way to test this via snactor?

Rezney: I do not think we can test it via snactor but you can test it using Leapp cockpit plugin. I can share details with you offline.

Rezney: In any way, when the command is in a good shape and executable, we should be good to merge this.

erig0: done.

Rezney (approved):

erig0:
@Rezney, thanks for working through these with me!

PR #785, opened 2022-01-03
[WIP] add actor that reports removed rhel-9 packages

ZoltanFridrich:
These packages have been removed between RHEL-8 and RHEL-9:
compat-openssl10, fipscheck, openssh-cavs, openssh-ldap, pinentry-emacs, pinentry-gtk, pyopenssl
Add an actor that checks if these packages are installed on the system and inhibit the upgrade process in that case.

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/checkremovedpackages/actor.py
+++ b/repos/system_upgrade/el8toel9/actors/checkremovedpackages/actor.py
@@ -1,0 +1,31 @@
+from leapp.actors import Actor
+from leapp.libraries.common.rpms import get_installed_rpms
+from leapp.reporting import Report, create_report
+from leapp import reporting
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+class CheckRemovedPackages(Actor):
+    """
+    Check for packages that have been removed between RHEL-8 and RHEL-9.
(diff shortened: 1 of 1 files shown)


pirat89:
@ZoltanFridrich Hi, could we sync over vcall about this PR? I am little bit confused about the purpose. I mean, there are hundreds of packages removed between RHEL X systems. Why these packages should inhibit the upgrade? I have more points to discuss, but in case you have time for a vcall, i think it will be faster.

pirat89 (changes requested):
As discussed on our sync call, the inhibitor in this case does not make sense. As well, the actor does not seem to provide any benefit right now. Try to answer these questions when thinking about the solution:
* What is the user actually supposed to do after seeing this msg?
* What is impact on user regarding the "removed" packages after the upgrade?
* Regarding the PES db, there are many packages that are going to be removed during IPU 8 -> 9. What is the difference in the context of other XXX removed packages that are not reported, to report explicitly the removal of these packages?

pirat89:
@ZoltanFridrich  as the internal jira ticket has been closed, I guess this one is supposed to be closed as well. Feel free to reopen if not.

ZoltanFridrich:
I was just waiting if there will be any other comments on the jira ticket. As it got closed on jira it is ok to close this PR. Thank you.

PR #787, opened 2022-01-04
Pass COMPOSE_URL for 8to9 tests

fernflower:
In order to do the --no-rhsm properly you have to pass both
TARGET_RELEASE and COMPOSE_URL, otherwise generated
leapp_upgrade_repositories.repo file will contain rhel8 repos
instead of rhel9.


fernflower:
If tests get better (https://api.dev.testing-farm.io/v0.1/requests/e329f1ba-551a-4f9f-a8e7-9cce64044b43, results of /rerun with this patch applied on my fork of leapp-repository https://github.com/fernflower/leapp-repository/pull/3) then it's a real solution.

fernflower:
Proof it works - https://github.com/fernflower/leapp-repository/actions/runs/1655722290

fernflower:
It was decided not to set COMPOSE_URL, so the proper compose choosing logic will be moved to the testing framework.

PR #789, opened 2022-01-05
WIP: PesEventsScanner: Filter releases that happened before the IPU source ver.

MichalHe:
Currently the PESEventsScanner filters releases (and thus events) only based on the target system version. This patch tweaks this behavior, so that also events with historical events (events that happened before the IPU source version) are considered irrelevant.

To illustrate why this is not a desired behavior, consider upgrading from RHEL8.6 to RHEL9.0. Currently also events that happened before release 8.7 (eg. events with source_version=7.5, target_version=7.6) would be considered relevant. If the relevance criteria of such events are met the events will affect the performed RPM transaction, however the source OS already reflects those changes.

Jira ref.: OAMG-6332


MichalHe:
leapp-ci build

MichalHe:
leapp-ci build

MichalHe:
leapp-ci build

MichalHe:
leapp-ci build

repos/system_upgrade/common/actors/peseventsscanner/libraries/peseventsscanner.py
```diff
@@ -76,10 +76,8 @@ def pes_events_scanner(pes_json_directory, pes_json_filename):
     arch = api.current_actor().configuration.architecture
     events = get_events(pes_json_directory, pes_json_filename)
     releases = get_releases(events)
-    source = version._version_to_tuple(api.current_actor().configuration.version.source)
-    target = version._version_to_tuple(api.current_actor().configuration.version.target)
 
-    filtered_releases = filter_releases(releases, source, target)
+    filtered_releases = filter_irrelevant_releases(releases)
```
Rezney, line 80: strange, when I run 8 > 9 preupgrade, I got `(9, 0)` in filtered releases.

MichalHe: IMHO, that is the correct behavior. This information is used to filter out event that have their `to_release` in one of the relevant releases. Therefore, when upgrading 8.6->9.0, only events with target release `9.0` are relevant. If we kept them the way it is now, it means that also an event, say with `from_release=8.5` and `to_release=8.6` will be processed. However, the source rhel 8.6 system already reflects such an event, which might cause troubles.

vinzenz: It is correct that for 8->9 you should only get 9.0 events, I can't think of a reason why you would want any other ones.
Also right now I think that this implementation is correct. I can't see why it wouldn't be. 

Now when we have a target version 9.1, I would want to see 9.0 and 9.1 as target, but not 9.2.

Rezney: Ok, it looks like I got the naming wrong.

PR #790, opened 2022-01-06
Packit: stop merging with master for PRs builds

pirat89:
When a PR build is created by PackIt it contains different hash
of the commit than the latest in the PR. This is caused by the
default behaviour, when PackIt merge locally the PR changes to the
master.

As we like this idea to test stuff how it works when applied to the
master, it has several problems:
  - sometimes a PR could have conflicts with master and the build
    will not be created in such a case; but for some simple testing
    it could be a little bit annoying for someone
  - it's not possible to install a build to a specific commit using
    the commit short-hash; sometimes it's usefull when you want to
    really try the build from specific commit and it's hard to find
    the right one in such a case

So let's set `merge_pr_in_ci` to `false` so the created build
is identical to the one in the PR.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -6,6 +6,7 @@ specfile_path: packaging/leapp-repository.spec
 upstream_package_name: leapp-repository
 downstream_package_name: leapp-repository
 upstream_tag_template: 'v{version}'
+merge_pr_in_ci: false
 
 # This is just for the build from the CLI - all other builds for jobs use own
 # actions


pirat89:
Putting to the draft for not as current version of PackIt contains bug. When `merge_pr_in_ci` is `false`, the package schema is incorrect due to wrong local ref.

---
* Expected: leapp-repository-0.15.0-0.20220106125723353413.**pr790**.39.g07a7000...
* Current: leapp-repository-0.15.0-0.20220106125723353413.**pr.changes790**.39.g07a7000...

Reported issue: https://github.com/packit/packit/issues/1443
Thanks @TomasTomecek for fast investigation!

TomasTomecek:
@pirat89 thank you for checking last week, completely fell off my radar last week since I went for a PTO

We deployed the fix last week, Tuesday and I can see it indeed resolves the issue \o/

Mergy mergy :)

pirat89:
@TomasTomecek  thanks for awesome quick support from your side guys!! :)

vinzenz (approved):

PR #791, opened 2022-01-06
Fix invalid format string in tmt-tests workflow

fernflower:
This should make multiple copr build tests to be launced again with
/rerun LEAPP_PR_NUMBER command.


fernflower:
Proof it works - https://github.com/fernflower/leapp-repository/runs/4728664566?check_suite_focus=true

Rezney (approved):

pirat89 (approved):
seems good

PR #792, opened 2022-01-06
actor: add inhibitor FirewalldCheckServiceTftpClient

erig0:
Fixes: OAMG-6032

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/firewalldcheckservicetftpclient/actor.py
+++ b/repos/system_upgrade/el8toel9/actors/firewalldcheckservicetftpclient/actor.py
@@ -1,0 +1,51 @@
+from leapp import reporting
+from leapp.actors import Actor
+from leapp.models import FirewalldUsedObjectNames
+from leapp.reporting import create_report, Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class FirewalldCheckServiceTftpClient(Actor):
+    """
+    This actor will inhibit if firewalld's configuration is using service
+    'tftp-client'.
+    """
+
+    name = 'firewalld_check_service_tftp_client'
+    consumes = (FirewalldUsedObjectNames,)
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        tftp_client_service = 'tftp-client'
+        send_report = False
+
+        for facts in self.consume(FirewalldUsedObjectNames):
+            if tftp_client_service in facts.services:
+                send_report = True
+
+        if send_report:
+            create_report([
+                reporting.Title('Unsupported Firewalld Configuration In Use'),
+                reporting.Summary('Firewalld has service "{service}" enabled. '
+                                  'Service "{service}" has been removed in RHEL-9.'.format(
+                                      service=tftp_client_service)),
+                reporting.Severity(reporting.Severity.HIGH),
+                reporting.Tags([reporting.Tags.SANITY, reporting.Tags.FIREWALL]),
+                reporting.Flags([reporting.Flags.INHIBITOR]),
+                reporting.Remediation(
+                    hint=(
+                        'Remove all usage of service "{service}" from '
+                        'firewalld\'s permanent configuration. '
+                        'It may be in use by: zones, policies, or rich rules.\n'
+                        'Usage can be found by listing zone and policy '
+                        'configuration:\n'
+                        '  # firewall-cmd --permanent --list-all-zones\n'
+                        '  # firewall-cmd --permanent --list-all-policies\n'
+                        'Example to remove usage from a zone:\n'
+                        '  # firewall-cmd --permanent --zone public '
+                        ' --remove-service {service}\n'.format(
+                            service=tftp_client_service)
+                        )
+                    ),
+            ])
--- a/repos/system_upgrade/el8toel9/actors/firewalldcollectusedobjectnames/libraries/private_firewalldcollectusedobjectnames.py
+++ b/repos/system_upgrade/el8toel9/actors/firewalldcollectusedobjectnames/libraries/private_firewalldcollectusedobjectnames.py
@@ -1,0 +1,91 @@
+from leapp.models import FirewalldUsedObjectNames
+
+try:
+    from firewall.core.fw import Firewall
+except ImportError:
+    pass
+
+
+def is_zone_in_use(conf):
+    # This does not account for zone assignments given by NetworkManager.
+    if conf.get('interfaces', []) or conf.get('sources', []):
+        return True
+
+    return False
+
+
+def is_policy_in_use(conf, used_zones):
+    # A policy is in use if both ingress_zones and egress_zones contain at
+    # least one of following: an active zone, 'ANY', 'HOST'.
+    for zone in conf.get('ingress_zones', []):
+        if zone in ['ANY', 'HOST'] or zone in used_zones:
+            break
+    else:
+        return False
+    for zone in conf.get('egress_zones', []):
+        if zone in ['ANY', 'HOST'] or zone in used_zones:
+            return True
+    return False
+
+
+def get_used_services(conf, isZone):
+    used_services = set()
+
+    for service in conf.get('services', []):
+        used_services.add(service)
+
+    # Also need to look for 'service name="<service>"' in rich rules. The rule
+    # strings are normalized by firewalld. zone keyword is 'rules_str'. policy
+    # keyword is 'rich_rules'.
+    for rule in conf.get('rules_str', []) if isZone else conf.get('rich_rules', []):
+        try:
+            search = 'service name="'
+            start = rule.index(search) + len(search)
+            stop = rule[start:].index('"')
+            used_services.add(rule[start:start+stop])
+        except ValueError:
+            pass
+
+    return used_services
+
+
+def read_config():
+    try:
+        fw = Firewall(offline=True)
+    except NameError:
+        # import failure missing means firewalld is not installed. Just return
+        # the defaults.
+        return FirewalldUsedObjectNames()
+
+    # This does not actually start firewalld. It just loads the configuration a
+    # la firewall-offline-cmd.
+    fw.start()
+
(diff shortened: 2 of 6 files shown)


erig0:
review please

repos/system_upgrade/el8toel9/actors/firewalldcollectusedobjectnames/libraries/private_firewalldcollectusedobjectnames.py
```diff
@@ -0,0 +1,92 @@
...
+    # least one of following: an active zone, 'ANY', 'HOST'.
+    for zone in conf.get('ingress_zones', []):
+        if zone in ['ANY', 'HOST'] or \
+           zone in used_zones:
+            break
+    else:
+        return False
+    for zone in conf.get('egress_zones', []):
+        if zone in ['ANY', 'HOST'] or \
+           zone in used_zones:
```
Rezney, line 28: Nit: this line should easily fit to the previous one. Same for the line 22. However, this is probably matter of personal taste. It just looks like indentation is broken on first look.

Rezney:
@erig0 Please do a rebase.

repos/system_upgrade/el8toel9/actors/firewalldcollectusedobjectnames/libraries/private_firewalldcollectusedobjectnames.py
erig0: ACK. I'll fix it when I rebase. It's probably an artifact from code refactoring during development (i.e. original was indented more).

repos/system_upgrade/el8toel9/actors/firewalldcollectusedobjectnames/libraries/private_firewalldcollectusedobjectnames.py
```diff
@@ -0,0 +1,90 @@
...
+            pass
+
+    return used_services
+
+
+def read_config():
+    if 'Firewall' not in dir():
+        # import failure missing means firewalld is not installed. Just return
+        # the defaults.
+        return FirewalldUsedObjectNames()
```
erig0, line 56: @Rezney, this if statement is evaluating to True when running `leapp preupgrade --debug --no-rhsm`. It works as expected inside a repl.

The only reason the import check is like this is because a local import was denied by CI. I had to put the import at the top of the file.

Inside the if statement sys.path = `['/usr/bin', '/usr/lib6/python36.zip', '/usr/lib64/python3.6', '/usr/lib64/python3.6/lib-dynload', '/usr/lib64/python3.6/site-packages', '/usr/lib/python3.6/site-packages']

erig0: Since `dir()` is inside a function `Firewall` is in the global scope and not returned by `dir()`.

Solution is to try and use `Firewall()` and catch the exception.

repos/system_upgrade/el8toel9/actors/firewalldcollectusedobjectnames/libraries/private_firewalldcollectusedobjectnames.py
```diff
@@ -0,0 +1,90 @@
...
+    if conf.get('interfaces', []) or conf.get('sources', []):
+        return True
+
+    return False
+
+
+def is_policy_in_use(conf, used_zones):
+    # A policy is in use if both ingress_zones and egress_zones contain at
+    # least one of following: an active zone, 'ANY', 'HOST'.
+    for zone in conf.get('ingress_zones', []):
```
vinzenz, line 20: Sorry couldn't help myself
Wouldn't do this the same?
```python

ingress = set(conf.get('ingress_zones', []))
egress = set(conf.get('egress_zones', []))
used = set(used_zones + ['ANY', 'HOST'])
return ingress & used and egress & used
```

repos/system_upgrade/el8toel9/actors/firewalldcollectusedobjectnames/libraries/private_firewalldcollectusedobjectnames.py
erig0: Done. Please take a look.

Rezney:
@erig0 Thanks for the fix. Now, after issuing `firewall-cmd --permanent --zone public --remove-service tftp-client --permanent`, the actor works inhibits just fine.

erig0:
> @erig0 Thanks for the fix. Now, after issuing `firewall-cmd --permanent --zone public --remove-service tftp-client --permanent`, the actor works inhibits just fine.

bump

repos/system_upgrade/el8toel9/models/firewalldusedobjectnames.py
```diff
@@ -0,0 +1,14 @@
+from leapp.models import fields, Model
+from leapp.topics import SystemInfoTopic
+
+
+class FirewalldUsedObjectNames(Model):
+    '''
+    This model contains lists of firewalld object (e.g. zones, services) names
+    in use by the permanent firewalld configuration.
+    '''
```
pirat89, line 9: @erig0 use please double-quotes instead of single-quotes for docstrings (everywhere). e.g.
```suggestion
    """
    This model contains lists of firewalld object (e.g. zones, services) names
    in use by the permanent firewalld configuration.
    """
```

repos/system_upgrade/el8toel9/models/firewalldusedobjectnames.py
```diff
@@ -0,0 +1,14 @@
...
+    topic = SystemInfoTopic
+
+    services = fields.List(fields.String(), default=[])
```
pirat89, line 12: @erig0 if possible, add please description of the fields (example of some models with proper docs. https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/common/models/targetuserspace.py). but it's just nitpick, so keeping it on you.

repos/system_upgrade/el8toel9/actors/firewalldcheckservicetftpclient/actor.py
```diff
@@ -0,0 +1,35 @@
...
+        if send_report:
+            create_report([
+                reporting.Title('Unsupported Firewalld Configuration'),
+                reporting.Summary('Firewalld has enabled service '
+                                  '"{service}" which has been removed in RHEL-9.'.format(
+                                      service=tftp_client_service)),
+                reporting.Severity(reporting.Severity.HIGH),
+                reporting.Tags([reporting.Tags.SANITY, reporting.Tags.FIREWALL]),
+                reporting.Flags([reporting.Flags.INHIBITOR])
+            ])
```
pirat89, line 35: @erig0 Couple of tips when creating report based on our experience with customers, so they will not be opening bunch of customer tickets & possibly bugs for that. Not saying you have to change it, just giving you tips how you could improve it if you want to improve UX.

it's good to ask yourself generic questions like:
* what is user supposed to do after the reading the report?
* what I want to tell to the user by this report?
* is it something we have documented / we are going to document in the official RH docs so there is URL with more details? (`reporting.ExternalLink`)

In my case it seems obvious that in case I want to continue with the upgrade, I have to disable the service. But from our experience, some users will not have an idea what they are supposed to do. Adding the remediation instructions can reduce amount of opened support tickets. Remediation instructions:
* with the hint field, like: in case you want to proceed the upgrade and you are ok the service is missing on the target system, stop & disable the service before the upgrade
* it's possible to provide the command field as well, which can be used directly via satellite or cockpit, but it has to be written now in the way it's just ONE command; e.g. in case of stop & disable the service, something like `bash -c "cmd1 && cmd2;"` because of the way how the current satellite/foreman plugin is written.

For inspiration abour remediations instructions see e.g. https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el7toel8/actors/checkhacluster/libraries/checkhacluster.py

In this case you can also add `reporting.RelatedResource('service', tftp_client_service)`

repos/system_upgrade/el8toel9/actors/firewalldcheckservicetftpclient/actor.py
```diff
@@ -0,0 +1,35 @@
...
+        send_report = False
+
+        for facts in self.consume(FirewalldUsedObjectNames):
+            if tftp_client_service in facts.services:
+                send_report = True
+
+        if send_report:
+            create_report([
+                reporting.Title('Unsupported Firewalld Configuration'),
+                reporting.Summary('Firewalld has enabled service '
```
fernflower, line 29: Sorry for nitpicking, imho 'Firewalld has service {service} enabled..' reads better here

repos/system_upgrade/el8toel9/actors/firewalldcheckservicetftpclient/actor.py
fernflower: +1, it would be great to give user some pointer how to proceed. Ideally a remediation command that could be applied, but at least description\general guidance.

repos/system_upgrade/el8toel9/models/firewalldusedobjectnames.py
erig0: Done

repos/system_upgrade/el8toel9/models/firewalldusedobjectnames.py
erig0: Done

repos/system_upgrade/el8toel9/actors/firewalldcheckservicetftpclient/actor.py
erig0: I added a Remediation, but there is no generic command to resolve the situation. The commands to resolve depends on the users config.

repos/system_upgrade/el8toel9/actors/firewalldcheckservicetftpclient/actor.py
erig0: Done

Rezney (approved):

PR #795, opened 2022-01-07
Bump leapp-framework-dependencies to 5

pirat89:
Regarding the change in https://github.com/oamg/leapp/pull/749
Added dependency on python3: python3 should refer always to the
system python package, read:
  python 3.6 on RHEL 8
  python 3.9 on RHEL 9
  ...


pirat89:
Tested successfully on intel - see https://github.com/oamg/leapp/pull/749 for details.

Rezney (approved):

PR #799, opened 2022-01-13
modularity: Reset modules before upgrade

vinzenz:
Previously we have added the modularity support into leapp. However we
struggled with the part of resetting the modules. This patch will add
the missing piece for that feature.
It implements resetting modules for module streams that are not going to
be installed. If the same module gets installed, but a different stream,
the module needs to be reset as well. The only exception is when the
module:stream tuple is the same, then we should rather not reset it,
according to the DNF developers.

### DIFF
--- a/repos/system_upgrade/common/libraries/dnfplugin.py
+++ b/repos/system_upgrade/common/libraries/dnfplugin.py
@@ -89,7 +89,6 @@ def build_plugin_data(target_repoids, debug, test, tasks, on_aws):
             'to_remove': tasks.to_remove,
             'to_upgrade': tasks.to_upgrade,
             'modules_to_enable': ['{}:{}'.format(m.name, m.stream) for m in tasks.modules_to_enable],
-            'modules_to_reset': ['{}:{}'.format(m.name, m.stream) for m in tasks.modules_to_reset],
         },
         'dnf_conf': {
             'allow_erasing': True,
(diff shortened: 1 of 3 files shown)


repos/system_upgrade/common/libraries/dnfplugin.py
```diff
@@ -163,29 +162,51 @@ def _transaction(context, stage, target_repoids, tasks, plugin_info, test=False,
...
             env = {'SYSTEMD_SECCOMP': '0'}
+
+            # We need to reset modules twice, once before we check, and the second time before we actually perform
+            # the upgrade. Not more often as the modules will be reset already.
+            if stage in ('check', 'upgrade') and tasks.modules_to_reset:
+                # We shall only reset modules that are not going to be enabled
+                # This will make sure it is so
+                modules_to_reset = {(module.name, module.stream) for module in tasks.modules_to_reset}
+                modules_to_enable = {(module.name, module.stream) for module in tasks.modules_to_enable}
+                module_reest_list = [module[0] for module in modules_to_reset - modules_to_enable]
```
Rezney, line 187:
```suggestion
                module_reset_list = [module[0] for module in modules_to_reset - modules_to_enable]
```

repos/system_upgrade/common/libraries/dnfplugin.py
```diff
@@ -163,29 +162,51 @@ def _transaction(context, stage, target_repoids, tasks, plugin_info, test=False,
...
+                # Perform module reset
+                cmd = ['/usr/bin/dnf', 'module', 'reset', '--enabled', ] + module_reest_list
```
Rezney, line 189:
```suggestion
                cmd = ['/usr/bin/dnf', 'module', 'reset', '--enabled', ] + module_reset_list
```

Rezney (commented):
So far looks good to me (just two typos) but I still have to test it yet.

mmacura311:
tested and it solves the the yum errors after upgrade to el9.

Rezney (approved):
I tested it too. Works fine.

PR #801, opened 2022-01-16
dnfdryrun: Consume BootContent to run the actor after upgrade initramfs

pirat89:
The actor is supposed to check the "free space" for the upgrade
transaction after the upgrade initramfs is generated and before the
bootloader is updated. However in some cases we could see the actor
is executed before the upgrade initramfs is generated, which is
problematic in case there is not enough free space on the disk.

Consume the BootContent msg in the actor so we are sure it's
executed always after the upgrade initramfs is generated.


Rezney (dismissed):
Makes sense...

MichalHe:
The integration tests covering this PR has been merged. LGTM. Merging.

PR #802, opened 2022-01-17
actor: Add postgresqlcheck actor for el8toel9 upgrade

fila43:
actor: Add postgresqlcheck actor for el8toel9 upgrade
Since RHEL 8 supports multiple Postgresql streams(9.6, 10, 12, and 13)
and RHEL 9 contains only Postgresql 13 the actor checks presence of
postgresql-server and provides a link to upgrade documentation.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=2037850

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/postgresqlcheck/libraries/postgresqlcheck.py
+++ b/repos/system_upgrade/el8toel9/actors/postgresqlcheck/libraries/postgresqlcheck.py
@@ -1,0 +1,56 @@
+from leapp import reporting
+from leapp.libraries.common.rpms import has_package
+from leapp.libraries.stdlib import api
+from leapp.models import InstalledRedHatSignedRPM
+
+# Summary for postgresql-server report
+report_server_inst_summary = (
+    'PostgreSQL server component will be upgraded. Since RHEL-9 includes'
+    ' PostgreSQL server 13 by default, which is incompatible with 9.6, 10 and 12'
+    ' included in RHEL-8, it is necessary to proceed with additional steps'
+    ' for the complete upgrade of the PostgreSQL data.'
+)
+
+report_server_inst_hint = (
+    'Back up your data before proceeding with the upgrade'
+    ' and follow steps in the documentation section "Migrating to a RHEL 9 version of PostgreSQL"'
+    ' after the upgrade.'
+)
+
+# Link URL for postgresql-server report
+report_server_inst_link_url = 'https://access.redhat.com/articles/6654721'  # noqa: E501; pylint: disable=line-too-long
+
+
+def _report_server_installed():
+    """
+    Create report on postgresql-server package installation detection.
+
+    Should remind user about present PostgreSQL server package
+    installation, warn them about necessary additional steps, and
+    redirect them to online documentation for the upgrade process.
+    """
+    reporting.create_report([
+        reporting.Title('PostgreSQL (postgresql-server) has been detected on your system'),
+        reporting.Summary(report_server_inst_summary),
+        reporting.Severity(reporting.Severity.MEDIUM),
+        reporting.Tags([reporting.Tags.SERVICES]),
+        reporting.ExternalLink(title='Migrating to a RHEL 9 version of PostgreSQL',
+                               url=report_server_inst_link_url),
+        reporting.RelatedResource('package', 'postgresql-server'),
+        reporting.Remediation(hint=report_server_inst_hint),
+        ])
+
+
+def report_installed_packages(_context=api):
+    """
+    Create reports according to detected PostgreSQL packages.
+
+    Create the report if the postgresql-server rpm (RH signed) is installed.
+    Additionally, create another report if the postgresql-contrib rpm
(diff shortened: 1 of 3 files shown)


repos/system_upgrade/el8toel9/actors/postgresqlcheck/libraries/postgresqlcheck.py
```diff
@@ -0,0 +1,55 @@
...
+)
+
+report_server_inst_hint = (
+    'Back up your data before proceeding with the upgrade'
+    ' and follow steps in the documentation section "Migrating to a RHEL 9 version of PostgreSQL"'
+    ' after the upgrade.'
+)
+
+# Link URL for postgresql-server report
+report_server_inst_link_url = 'https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9-beta/html/configuring_and_using_database_servers/using-postgresql#migrating-to-a-rhel-9-version-of-postgresql_using-postgresql'  # noqa: E501; pylint: disable=line-too-long
```
hhorak, line 22: I think we should rather have a KB article here, as we're more flexible in changing it and it can later serve as a land-page for postgresql updates (so it would include the actual links to migration doc and other info).

fila43: thanks, definitely make sense

repos/system_upgrade/el8toel9/actors/postgresqlcheck/libraries/postgresqlcheck.py
```diff
@@ -0,0 +1,55 @@
+from leapp import reporting
+from leapp.libraries.common.rpms import has_package
+from leapp.libraries.stdlib import api
+from leapp.models import InstalledRedHatSignedRPM
+
+
+# Summary for postgresql-server report
+report_server_inst_summary = (
+    'PostgreSQL server component will be upgraded. Since RHEL-9 includes only'
```
hhorak, line 9: A slight nitpick -- thinking about this long-term, there will be more versions at some point and then this text could be misleading. I'm thinking whether we can figure out some wording that will not need to be changed when we add more alternative versions. Maybe just remove the "only" word is good start.

pirat89: @hhorak good point. I've been thinking about that couple of time and I have to agree I do not have better way than what you suggested. (e.g. explaining that properly requires much more text and it could be confusing for user probably). The issue is already resolved in the proposed way, so resolving conversation :)

pirat89 (changes requested):
@fila43  rebase please against the up-to-date master and fix pls linter errors. in case you are using fedora, you can do something like that to run linters and unit tests on your machine (but do the rebase first otherwise you see just bunch of errors):
```bash
# in the root leapp-repository directory
make install-deps-fedora
make fast_lint
make test_no_lint
```

also, in case of possible errors related to the import order, you can use `make lint_fix` that will automatically fix all imports as expected. The PR seems good otherwise. It can be merged after the linter & unit tests pass.

fila43:
[review please]

fila43:
leapp-ci build

fila43:
review please

fila43:
review please

repos/system_upgrade/el8toel9/actors/postgresqlcheck/libraries/postgresqlcheck.py
```diff
@@ -0,0 +1,56 @@
...
+    Create reports according to detected PostgreSQL packages.
+
+    Create the report if the postgresql-server rpm (RH signed) is installed.
+    Additionally, create another report if the postgresql-contrib rpm
+    is installed.
+    """
+    has_server = has_package(InstalledRedHatSignedRPM, 'postgresql-server', context=_context)
+
+    if has_server:
+        # postgresql-server
```
fernflower, line 55: Nit: the comment isn't saying much really. Maybe change or remove?

repos/system_upgrade/el8toel9/actors/postgresqlcheck/libraries/postgresqlcheck.py
```diff
@@ -0,0 +1,56 @@
...
+        reporting.Severity(reporting.Severity.MEDIUM),
+        reporting.Tags([reporting.Tags.SERVICES]),
+        reporting.ExternalLink(title='Migrating to a RHEL 9 version of PostgreSQL',
+                               url=report_server_inst_link_url),
+        reporting.RelatedResource('package', 'postgresql-server'),
+        reporting.Remediation(hint=report_server_inst_hint),
+        ])
+
+
+def report_installed_packages(_context=api):
```
fernflower, line 44: Nit: Technically you can go without `_context=api` as `stdlib.api` is the default for has_package function anyway. This should not affect your testing.

fernflower (approved):
Lgtm, some nits.

pirat89 (approved):

PR #803, opened 2022-01-18
scandasd: el8 => el9 requires s390utils-core package

vinzenz:
Since el9, the s390utils-core package is no longer installed via
dependency for s390 systems. This PR moves the actor to common, as it is
required also for el8 to el9 upgrades and makes sure that the
s390utils-core package will be installed.


pirat89 (dismissed):
lgtm. waiting for Sebastian's feedback about the test results before the merge but this should be it.

repos/system_upgrade/common/actors/scandasd/libraries/scandasd.py
```diff
@@ -14,7 +14,7 @@ def process():
         # the file has to be copied into the targetuserspace container first,
         # then it can be included into the initramfs ==> both messages are
         # needed to be produced
-        api.produce(TargetUserSpaceUpgradeTasks(copy_files=[CopyFile(src=DASD_CONF)]))
+        api.produce(TargetUserSpaceUpgradeTasks(copy_files=[CopyFile(src=DASD_CONF)], install_rpms=['s390utils-core']))
```
smitterl, line 17: Even if the system boots from a DASD disk the `DASD_CONF` doesn't have to exist. Would it make more sense to always install the `s390utils-core`?

vinzenz: that's what this does

vinzenz: oh now I see what you mean

vinzenz: @pirat89 that's a good point, no?

pirat89: good catch!

pirat89 (approved):

vinzenz:
leapp-ci build

PR #804, opened 2022-01-18
modularity: Handle missing modulestream_maps

vinzenz:
In the case that there's no modulestream_maps in the pes event, handle
the situation for now as if it would be demodularized. There's no other
good solution than this, as we cannot guess what will be the actual
target in case there would be multiple output modulestreams.

### DIFF
--- a/repos/system_upgrade/common/actors/peseventsscanner/libraries/peseventsscanner.py
+++ b/repos/system_upgrade/common/actors/peseventsscanner/libraries/peseventsscanner.py
@@ -296,7 +296,11 @@ def event_for_modulestram_mapping(from_to, event):
 
 def event_by_modulestream_mapping(mapping, event):
     if not mapping:
-        return [event_for_modulestram_mapping(None, event)]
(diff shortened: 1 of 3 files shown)


repos/system_upgrade/common/actors/peseventsscanner/libraries/peseventsscanner.py
```diff
@@ -384,8 +388,15 @@ def parse_action(action_id):
 
 
 def parse_modulestream(package):
+    modulestreams = package.get('modulestreams', (package.get('modulestream'),))
```
MichalHe, line 391: Is this line needed?

vinzenz: Yes it is

vinzenz: just read below

pirat89 (approved):

PR #805, opened 2022-01-18
Persistent net names: Move actors to common repository for IPU

pirat89:
The persistentnetnames* actors have been kept in el7toel8 leapp
repository during the big repository split until we find whether
they are needed for IPU 8 -> 9 as well. Right now it seems that
are needed for all IPUs in general, so moving them to the common
repository.


msekletar:
+1 from me. These actors are of course needed in all upgrade workflows.

pirat89:
@msekletar thanks!

vinzenz (approved):

PR #816, opened 2022-01-23
[WIP] Zipl: Check for entries sharing the same kernel image.

MichalHe:
Zipl sharing same kernel image entries are causing troubles as the script converting zipl configuration
to BLS uses machine_id + image version to create temporary files. Should a temporary file exist, the
script fails, and thus, the upgrade cannot proceed.


MichalHe:
leapp-ci build

MichalHe:
According to the logic in `zipl-switch-to-blscfg`, rescue entries are identified *only* if the word `rescue` is present in the name of the kernel image. 
# Testing performed
### Entries sharing the same kernel image
Configuration contains two entries sharing the exactly same kernel image. Testing was performed twice, second time with one of the entries having `rescue` in the title, but not in the image path.
- `zipl-switch-to-blscfg` executed manually crashes - `BLS file X already exists`
- leapp inhibits the upgrade:
```
Risk Factor: high (inhibitor)
Title: Boot entries sharing the same kernel version found.
Summary: The zipl configuration file /etc/zipl.conf contains boot entries sharing the same kernel version:
    - 3.10.0-1160.53.1.el7.s390x (found in entries: "linux", "linux0")
Remediation: [hint] Remove boot entries sharing the same kernel version from the configuration and leave just one.
```
### Entries sharing the same kernel version (different images)
This is a weird scenario where the user has boot entries with images `/boot/vmlinuz-<VER>` and `/boot/<VER>` (which due to internal logic in zipl-switch-to-blscfg` results in a crash). It is unlikely to happen, but it is better to have it covered.
- `zipl-switch-to-blscfg` executed manually crashes - `BLS file X already exists`
- leapp inhibits the upgrade:
```
Risk Factor: high (inhibitor)
Title: Boot entries sharing the same kernel version found.
Summary: The zipl configuration file /etc/zipl.conf contains boot entries sharing the same kernel version:
    - 3.10.0-1160.53.1.el7.s390x (found in entries: "linux", "linux0")
Remediation: [hint] Remove boot entries sharing the same kernel version from the configuration and leave just one.
```
### Multiple rescue entries are present
- `zipl-switch-to-blscfg` executed manually crashes - `BLS file X-rescue already exists`
- leapp inhibits the upgrade:
```
Risk Factor: high (inhibitor)
Title: Multiple rescue boot entries present in the bootloader configuration.
Summary: The zipl configuration file /etc/zipl.conf contains multiple rescue boot entries:
    - linux-0-rescue
    - linux-1-rescue
Remediation: [hint] Remove rescue boot entries from the configuration and leave just one.
```
### Multiple rescue entries & entries sharing images
- `zipl-switch-to-blscfg` executed manually crashes - `BLS file X-rescue already exists`
- leapp inhibits the upgrade:
```
isk Factor: high (inhibitor)
Title: Multiple rescue boot entries present in the bootloader configuration.
Summary: The zipl configuration file /etc/zipl.conf contains multiple rescue boot entries:
    - linux-0-rescue
    - linux-0-arescue
Remediation: [hint] Remove rescue boot entries from the configuration and leave just one.
Key: x
Risk Factor: high (inhibitor)
Title: Boot entries sharing the same kernel version found.
Summary: The zipl configuration file /etc/zipl.conf contains boot entries sharing the same kernel version:
    - 3.10.0-1160.53.1.el7.s390x (found in entries: "linux", "linux2")
Remediation: [hint] Remove boot entries sharing the same kernel version from the configuration and leave just one.
Key: x
```

repos/system_upgrade/el7toel8/actors/sourcebootloaderscanner/libraries/sourcebootloaderscanner.py
```diff
@@ -26,20 +49,29 @@ def scan_boot_entries():
...
-        # of the grubby output into sections according to the `index` lines
-        if output_line.startswith('title='):
-            boot_entry = output_line[6:]  # Remove the `title=` prefix
-
-            # On s390x grubby produces quotes only when needed (whitespace in
-            # the configuration values), on x86 the values are quoted either way
-            boot_entry = boot_entry.strip('\'"')
+        if output_line == 'non linux entry':
+            # Grubby does will not display info about non-linux entries
+            # Such entry is not problematic from our PoV, therefore, skip it
```
pirat89, line 56:
```suggestion
            # Grubby does not display info about non-linux entries
            # Such an entry is not problematic from our PoV, therefore, skip it
```

MichalHe: Thanks, fixed.

repos/system_upgrade/el7toel8/actors/sourcebootloaderscanner/libraries/sourcebootloaderscanner.py
```diff
@@ -1,11 +1,34 @@
...
+from leapp.libraries.stdlib import api, CalledProcessError, run
+from leapp.models import BootEntry, SourceBootLoaderConfiguration
 
 CMD_GRUBBY_INFO_ALL = ['grubby', '--info', 'ALL']
 
 
+def parse_grubby_output_line(line):
+    """
+    Parses a single output line of `grubby --info ALL` that has the property=value format and returns a tuple
+    (property, value).
```
pirat89, line 11:
```suggestion
    (property, value).

```

repos/system_upgrade/el7toel8/actors/ziplcheckbootentries/libraries/ziplcheckbootentries.py
```diff
@@ -0,0 +1,134 @@
...
+                    FMT_LIST_SEPARATOR,
+                    version,
+                    ', '.join(entry_titles)
+            )
+
+        reporting.create_report([
+            reporting.Title('Boot entries sharing the same kernel version found.'),
+            reporting.Summary(
+                'The zipl configuration file {0} contains boot entries sharing the same kernel version: {1}'
+                .format(ZIPL_CONFIG_PATH, problematic_entries_details)
```
pirat89, line 118: @MichalHe add please more info into the summary explaining why this is the problem. e.g. mention something like: It's not possible to migrate the configuration to the BLS format correctly in such a case, which causes issues with the bootloader after the upgrade.

repos/system_upgrade/el7toel8/actors/ziplcheckbootentries/libraries/ziplcheckbootentries.py
```diff
@@ -0,0 +1,134 @@
...
+            reporting.Title('Multiple rescue boot entries present in the bootloader configuration.'),
+            reporting.Summary(
+                'The zipl configuration file {0} contains multiple rescue boot entries:{1}'
+                .format(ZIPL_CONFIG_PATH, rescue_entries_text)
+            ),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Tags([reporting.Tags.BOOT]),
+            reporting.Remediation(hint='Remove rescue boot entries from the configuration and leave just one.'),
+            reporting.Flags([reporting.Flags.INHIBITOR])
+        ])
```
pirat89, line 54: @MichalHe  isn't this just a subcase of the problem that is this PR resolving? but I thought that in such a case we will see this & the other report in the same time always when `rescuse` entries are affected. Is it correct?

repos/system_upgrade/el7toel8/actors/ziplcheckbootentries/libraries/ziplcheckbootentries.py
```diff
@@ -0,0 +1,134 @@
...
+    images.
+
+    :param str kernel_img_path: The path to the kernel image.
+    :returns: Extracted kernel version from the given path
+    :rtype: str
+    """
+
+    if 'vmlinuz-' in kernel_img_path:
+        fragments = kernel_img_path.rsplit('/vmlinuz-', 1)
+        # Mimick bash substitution used in the conversion script
```
MichalHe, line 72: Add this link to comment: 
```
See.... https://github.com/ibm-s390-linux/s390-tools/blob/b5604850ab66f862850568a37404faa647b5c098/scripts/zipl-switch-to-blscfg#L168
```

repos/system_upgrade/el7toel8/actors/sourcebootloaderscanner/libraries/sourcebootloaderscanner.py
```diff
@@ -26,20 +49,27 @@ def scan_boot_entries():
...
+            # Such an entry is not problematic from our PoV, therefore, skip it
+            boot_entry_data = {}
+            continue
 
-            boot_entries.append(BootEntry(title=boot_entry))
+        prop, value = parse_grubby_output_line(output_line)
+        if prop == 'index':
+            # Start of a new boot entry section
+            if boot_entry_data is not None:
+                boot_entries.append(BootEntry(title=boot_entry_data.get('title', ''),
```
MichalHe, line 64: Seatbelt - theoretically grubby might not print entry title (according to grubby code) but it is highly unlikely.

MichalHe: Comment will be added in as part of the changes in the next force push.

pirat89:
@MichalHe pls squash commits :)

MichalHe:
@pirat89 All should be prepared as requested :)

pirat89 (dismissed):
thanks!!! good job :)

repos/system_upgrade/el7toel8/actors/ziplcheckbootentries/actor.py
```diff
@@ -1,21 +1,21 @@
...
 from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
 
 
-class ZiplCheckMultipleRescueEntries(Actor):
+class ZiplCheckBootEntries(Actor):
     """
     Inhibits the upgrade process if there are more than one rescue entries in
     the zipl configuration.
     """
```
smitterl, line 12: This comment looks outdated.

repos/system_upgrade/el7toel8/actors/ziplcheckbootentries/libraries/ziplcheckbootentries.py
```diff
@@ -0,0 +1,134 @@
...
+    fragments = kernel_img_path.rsplit('/', 1)
+    return fragments[1] if len(fragments) > 1 else fragments[0]
+
+
+def inhibit_if_entries_share_kernel_version(bootloader_config):
+    """
+    Inhibits the upgrade if there are boot entries sharing the same kernel image version.
+
+    The logic of identification whether the images are the same mimics the zipl-switch-to-blscfg, as it fails
+    to perform the conversion if there are entries with the same kernel image.
```
smitterl, line 85: By mimicking the script aren't we opening doors to duplicating that script's validation logic in the leapp upgrade scripts?

repos/system_upgrade/el7toel8/actors/sourcebootloaderscanner/libraries/sourcebootloaderscanner.py
```diff
@@ -26,20 +50,28 @@ def scan_boot_entries():
...
+            # Grubby does not display info about non-linux entries
+            # Such an entry is not problematic from our PoV, therefore, skip it
+            boot_entry_data = {}
+            continue
 
-            boot_entries.append(BootEntry(title=boot_entry))
+        prop, value = parse_grubby_output_line(output_line)
+        if prop == 'index':
+            # Start of a new boot entry section
+            if boot_entry_data is not None:
```
fernflower, line 64: Nit: you can have `if boot_entry_data:` here

repos/system_upgrade/el7toel8/actors/ziplcheckbootentries/libraries/ziplcheckbootentries.py
smitterl: The script allows for being run into a temporary bls folder

```
--bls-directory=DIR     path to generate BLS files, defaults to /boot/loader/entries
--config-file=FILE      path to zipl configuration file, defaults to /etc/zipl.conf
```
Would it make sense to try this and inhibit on failure passing on the scripts output? This way the validation logic would remain in the zipl-switch-to-blscfg script and any new case wouldn't have to be added to leapp's logic IMO.

repos/system_upgrade/el7toel8/actors/sourcebootloaderscanner/libraries/sourcebootloaderscanner.py
```diff
@@ -26,20 +50,28 @@ def scan_boot_entries():
...
+                boot_entries.append(
+                    BootEntry(title=boot_entry_data.get('title', ''),  # In theory, the title property can be missing
+                              kernel_image=boot_entry_data['kernel']))
```
fernflower, line 67: Hmm correct me if I am wrong please - given the way code is written now we can have KeyError at L67 when `boot_entry_data={}` and we are trying to retrieve `kernel`.

Theoretically, if we have been getting only 'non linux entry' in grubby_output['stdout'], then the first run with `prop == 'index'` at L62 will have `boot_entry_data={}` and that's when we are running into trouble with KeyError.

MichalHe: Brilliant observation, thanks! I'll fix it right away. 

I remember that I've retrofitted this the 'non linux entry' only after I've realized that testing farm VMs have such entries, so that might be why I did not realize this might be a problem.

repos/system_upgrade/el7toel8/actors/ziplcheckbootentries/libraries/ziplcheckbootentries.py
MichalHe: The script is quite bad at explaining to the user why did it fail - it terminates with `File already exists.` and does not provide any further explanation about what caused the file to exist, which can be considered as a painfully bad user experience (no instructions on what exactly went wrong or how to fix it).

repos/system_upgrade/el7toel8/actors/ziplcheckbootentries/actor.py
MichalHe: Thanks a lot, fixed it.

repos/system_upgrade/el7toel8/actors/sourcebootloaderscanner/libraries/sourcebootloaderscanner.py
MichalHe: I remember that there was a reason why I needed to explicitly want `None`, however I do not see it now. I've fixed it.

This also fixes the possible `KeyError` as an empty dictionary resulting from a non-linux entry will be treated as `False`, and therefore, this branch will not be taken.

repos/system_upgrade/el7toel8/actors/ziplcheckbootentries/actor.py
```diff
@@ -0,0 +1,24 @@
...
+from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class ZiplCheckBootEntries(Actor):
+    """
+    Inhibits the upgrade if a problematic Zipl configuration is detected on the system.
+
+    The configuration is considered problematic if it will cause troubles during its conversion to BLS.
+    Such troubles can be caused by either containing multiple rescue entries, or containing rescue entries
```
pirat89, line 13:
```suggestion
    Such troubles can be caused by either containing multiple rescue entries, or containing any boot entries
```

pirat89 (approved):

PR #818, opened 2022-01-24
Fix GRUB configuration with BLS on ppc64le

Rezney:
On "ppc64le" systems with BLS enabled, the GRUB configuration is not
properly converted after the upgrade and Leapp has to run "grub2-mkconfig"
-o /boot/grub2/grub.cfg command in order to fix an issue with booting into
the RHEL 8 kernel instead of RHEL 9.

### DIFF
--- a/repos/system_upgrade/common/actors/systemfacts/actor.py
+++ b/repos/system_upgrade/common/actors/systemfacts/actor.py
@@ -1,8 +1,20 @@
 from leapp.actors import Actor
 from leapp.libraries.actor import systemfacts
-from leapp.models import SysctlVariablesFacts, ActiveKernelModulesFacts, UsersFacts, GroupsFacts, RepositoriesFacts, \
-    SELinuxFacts, FirewallsFacts, FirmwareFacts
-from leapp.tags import IPUWorkflowTag, FactsPhaseTag
+from leapp.libraries.common.config import architecture
+from leapp.models import (
+    ActiveKernelModulesFacts,
+    DefaultGrubInfo,
+    FirewallsFacts,
+    FirmwareFacts,
+    GroupsFacts,
+    GrubCfgBios,
+    Report,
+    RepositoriesFacts,
+    SELinuxFacts,
+    SysctlVariablesFacts,
+    UsersFacts
+)
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
 
 
 class SystemFactsActor(Actor):
@@ -22,14 +34,19 @@ class SystemFactsActor(Actor):
 
     name = 'system_facts'
     consumes = ()
-    produces = (SysctlVariablesFacts,
-                ActiveKernelModulesFacts,
-                UsersFacts,
-                GroupsFacts,
-                RepositoriesFacts,
-                SELinuxFacts,
-                FirewallsFacts,
-                FirmwareFacts)
+    produces = (
+        SysctlVariablesFacts,
+        ActiveKernelModulesFacts,
+        UsersFacts,
+        GroupsFacts,
+        RepositoriesFacts,
+        SELinuxFacts,
+        FirewallsFacts,
+        FirmwareFacts,
+        DefaultGrubInfo,
+        GrubCfgBios,
+        Report
+    )
     tags = (IPUWorkflowTag, FactsPhaseTag,)
 
     def process(self):
@@ -41,3 +58,10 @@ class SystemFactsActor(Actor):
         self.produce(systemfacts.get_selinux_status())
         self.produce(systemfacts.get_firewalls_status())
         self.produce(systemfacts.get_firmware())
+
+        if not architecture.matches_architecture(architecture.ARCH_S390X):
+            self.produce(systemfacts.get_default_grub_conf())
+
+        bios_grubcfg_details = systemfacts.get_bios_grubcfg_details()
+        if bios_grubcfg_details:
+            self.produce(bios_grubcfg_details)
--- a/repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
+++ b/repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
@@ -8,16 +8,21 @@ import re
 
 import six
 
+from leapp import reporting
 from leapp.libraries.common import repofileutils
+from leapp.libraries.common.config import architecture
 from leapp.libraries.stdlib import api, CalledProcessError, run
 from leapp.models import (
     ActiveKernelModule,
     ActiveKernelModulesFacts,
+    DefaultGrub,
+    DefaultGrubInfo,
     FirewallsFacts,
     FirewallStatus,
     FirmwareFacts,
     Group,
     GroupsFacts,
+    GrubCfgBios,
     KernelModuleParameter,
     RepositoriesFacts,
     SELinuxFacts,
@@ -254,4 +259,53 @@ def get_firewalls_status():
 
 def get_firmware():
     firmware = 'efi' if os.path.isdir('/sys/firmware/efi') else 'bios'
+    if architecture.matches_architecture(architecture.ARCH_PPC64LE):
+        ppc64le_opal = bool(os.path.isdir('/sys/firmware/opal/'))
+        return FirmwareFacts(firmware=firmware, ppc64le_opal=ppc64le_opal)
     return FirmwareFacts(firmware=firmware)
+
+
+@aslist
+def _default_grub_info():
+    default_grb_fpath = '/etc/default/grub'
+    if not os.path.isfile(default_grb_fpath):
+        reporting.create_report([
+            reporting.Title('File "{}" does not exist!'.format(default_grb_fpath)),
+            reporting.Summary(
+                'Leapp detected "{}" does not exist. The file is essential for the in-place upgrade '
+                'to finish successfully. This scenario might have occured if the system was already '
+                'upgraded from RHEL 6. Please re-create the file manually.'.format(default_grb_fpath)
+            ),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Flags([reporting.Flags.INHIBITOR]),
+            reporting.Tags([reporting.Tags.BOOT]),
+            reporting.RelatedResource('file', default_grb_fpath),
+            reporting.ExternalLink(
+                url='https://access.redhat.com/solutions/3185891',
+                title='How to re-create the missing "{}" file in Red Hat Enterprise Linux 7?'.format(
+                    default_grb_fpath
+                )
+            ),
+        ])
+    else:
+        for line in run(['cat', default_grb_fpath], split=True)['stdout']:
+            name, value = tuple(map(type(line).strip, line.split('=', 1)))
+            yield DefaultGrub(
+                name=name,
+                value=value
+            )
+
+
+def get_default_grub_conf():
+    """ Get a list of GRUB parameters from /etc/default/grub """
+    return DefaultGrubInfo(default_grub_info=_default_grub_info())
+
+
+def get_bios_grubcfg_details():
+    """ Get BIOS (non-EFI) Grub config details """
+    if get_firmware().firmware == 'bios' and not architecture.matches_architecture(architecture.ARCH_S390X):
+        with open('/boot/grub2/grub.cfg') as fo:
+            content = fo.read()
+        insmod_bls = bool('insmod blscfg' in content)
+        return GrubCfgBios(insmod_bls=insmod_bls)
+    return None
--- a/repos/system_upgrade/common/libraries/grub.py
+++ b/repos/system_upgrade/common/libraries/grub.py
@@ -1,7 +1,7 @@
 import os
 
-from leapp.libraries.stdlib import run, api, CalledProcessError
 from leapp.exceptions import StopActorExecution
+from leapp.libraries.stdlib import api, CalledProcessError, run
 
 
 def has_grub(blk_dev):
@@ -70,3 +70,14 @@ def get_grub_device():
     api.current_logger().info('GRUB is installed on {}'.format(grub_dev))
     # if has_grub(grub_dev):
     return grub_dev if has_grub(grub_dev) else None
+
+
+def is_blscfg_enabled_in_defaultgrub(default_grub_msg):
+    """
+    Check if GRUB_ENABLE_BLSCFG is true in /etc/default/grub file
+    """
+    grub_options_lst = default_grub_msg.default_grub_info
+    default_grub_options = {
+        option.name: option.value.strip('"') for option in grub_options_lst
+    }
+    return bool(default_grub_options.get('GRUB_ENABLE_BLSCFG', '') == 'true')
--- a/repos/system_upgrade/common/models/defaultgrubinfo.py
+++ b/repos/system_upgrade/common/models/defaultgrubinfo.py
@@ -1,0 +1,25 @@
+from leapp.models import fields, Model
+from leapp.topics import SystemFactsTopic
+
+
+class DefaultGrub(Model):
+    """
+    A model with '/etc/default/grub' option as key / value
+    """
+    topic = SystemFactsTopic
+
+    name = fields.String()
+    """ Option name, e.g. GRUB_TIMEOUT """
+    value = fields.String()
+    """ Option value, e.g. '5' in case of GRUB_TIMEOUT=5 """
+
+
+class DefaultGrubInfo(Model):
+    """
+    A message with '/etc/default/grub' content
+    """
+
+    topic = SystemFactsTopic
+
+    default_grub_info = fields.List(fields.Model(DefaultGrub))
+    """ List of '/etc/default/grub' options as key / value """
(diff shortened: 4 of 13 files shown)


repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -242,3 +258,25 @@ def _get_firewall_status(service_name):
...
+def get_default_grub():
+    """ Get a list of GRUB parameters from /etc/default/grub """
+    return DefaultGrubInfo(default_grub_info=_default_grub_info())
+
+
+def get_grubcfg():
+    """ Get '/boot/grub2/grub.cfg' content """
+    with open('/boot/grub2/grub.cfg') as fo:
+        grub_cfg = fo.read()
+    return GrubCfg(grub_cfg=grub_cfg)
```
pirat89, line 282: is there a reason to store just the file instead of parsed / processed content? storing the text file and than keep everyone parsing it again and again. that's opposite what we tell people they should do when they create scanners and models. this is not good design and example

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -242,3 +258,25 @@ def _get_firewall_status(service_name):
...
+
+
+@aslist
+def _default_grub_info():
+    for line in run(['cat', '/etc/default/grub'], split=True)['stdout']:
+        name, value = tuple(map(type(line).strip, line.split('=', 1)))
+        yield DefaultGrub(
+            name=name,
+            value=value
+        )
```
pirat89, line 270: more in comments in the model file itself

repos/system_upgrade/common/libraries/grub.py
```diff
@@ -70,3 +70,23 @@ def get_grub_device():
...
+
+def is_blscfg_enabled(default_grub_msg):
+    """
+    Check if GRUB_ENABLE_BLSCFG is true
+    """
+    grub_options_lst = default_grub_msg.default_grub_info
+    default_grub_options = {
+        option.name: option.value.strip('"') for option in grub_options_lst
+    }
+    return True if default_grub_options.get('GRUB_ENABLE_BLSCFG', '') == 'true' else False
```
pirat89, line 92: we should work with the parsed content, and these two functions seem to be confusing in my POV . Based on their names and descriptions, I am not able to see what is difference between them unless I check the code.

repos/system_upgrade/common/models/defaultgrubinfo.py
```diff
@@ -0,0 +1,15 @@
...
+    topic = SystemFactsTopic
+
+    name = fields.String()
+    value = fields.String()
+
+
+class DefaultGrubInfo(Model):
+    topic = SystemFactsTopic
+
+    default_grub_info = fields.List(fields.Model(DefaultGrub))
```
pirat89, line 15: we should keep every new model pretty wel covered by docstring (includes docstring for the fields) + I think that these models could be designed differently. Let's discuss it as it's highly possible I missed something, but if we can do it better, I think we should.

repos/system_upgrade/common/models/grubcfg.py
```diff
@@ -0,0 +1,8 @@
+from leapp.models import fields, Model
+from leapp.topics import SystemFactsTopic
+
+
+class GrubCfg(Model):
+    topic = SystemFactsTopic
+
+    grub_cfg = fields.String()
```
pirat89, line 8: Similar to models defined above. In this case the content of the file is not even parsed, which is not good. We should introduce model that reflects the configuration. Let's look at what we could deliver. In case it's not a big deal, I think we should give it some love.

repos/system_upgrade/el8toel9/actors/grub2mkconfigonppc64/actor.py
```diff
@@ -0,0 +1,19 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import grub2mkconfigonppc64
+from leapp.models import DefaultGrubInfo
+from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
+
+
+class Grub2MkconfigOnPpc64(Actor):
+    """
+    TBA
+    """
```
pirat89, line 10: let it be added :)

repos/system_upgrade/common/actors/systemfacts/actor.py
```diff
@@ -41,3 +55,5 @@ def process(self):
         self.produce(systemfacts.get_selinux_status())
         self.produce(systemfacts.get_firewalls_status())
         self.produce(systemfacts.get_firmware())
+        self.produce(systemfacts.get_default_grub())
```
pirat89, line 58: what about `get_default_grub_configuration`? or `get_default_grub_conf`? As well, we should check the file exists. On some systems upgraded from RHEL 6 the file does not have to be present in case that users have not finished the migration to GRUB 2. we should check it to be able to give people proper hints what's wrong in such a case.

repos/system_upgrade/el8toel9/actors/checkblsgrubcfgonppc64/libraries/blsgrubcfgonppc64.py
```diff
@@ -0,0 +1,36 @@
...
+            reporting.Summary(
+                'On "ppc64le" systems with BLS enabled, the GRUB configuration is not '
+                'properly converted after the upgrade and Leapp has to run "grub2-mkconfig" '
+                '-o /boot/grub2/grub.cfg command in order to fix an issue with booting into '
+                'the RHEL 8 kernel instead of RHEL 9.'
+
+            ),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Tags([reporting.Tags.BOOT]),
+        ])
```
pirat89, line 36: from my understanding, is it possible to properly update the bootloader configuration on RHEL 8 system from user POV? I can imagine users could be afraid when they read the msg :) We could give them a remediation instructions how they can fix their bootloader configuration prior the upgrade on their own, so we will not do it. wdyt?

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
pirat89: as well, thinking about these scanners, they do not seem to be valid for all arches. Including that /boot/grub2/grub.cfg is not correct path on intel for EFI if I remember well. (let's check that)

pirat89 (changes requested):
let's sit on it as I found number of stuff that worries me.

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
pirat89: aah. sorry. I've exchanged the file with the /etc/default/grub :3 but instead of the file itself, let's share a usable info only. e.g. info about BLS in this case.

repos/system_upgrade/el8toel9/actors/checkblsgrubcfgonppc64/libraries/blsgrubcfgonppc64.py
pirat89: discussed, there is nothing what could be done before the upgrade because of bug in grub (the actions is not possible to do before the new grub is installed).

repos/system_upgrade/common/models/defaultgrubinfo.py
pirat89: we will just add docstrings here

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
pirat89: so checked, the file does not exist on EFI (and we are not sure how to correctly detect it as it could have various names, so the model should be created only on non-efi systems right now.

@smitterl I expect this file does not exist on s390x at all as there is used ZIPL instead of grub, correct?

repos/system_upgrade/common/models/grubcfg.py
pirat89: let's just produce info / facts we use instead of content of the file

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -242,3 +260,47 @@ def _get_firewall_status(service_name):
 def get_firmware():
     firmware = 'efi' if os.path.isdir('/sys/firmware/efi') else 'bios'
     return FirmwareFacts(firmware=firmware)
+
+
+@aslist
+def _default_grub_info():
+    default_grb_fpath = '/etc/default/grub'
+    if not os.path.isfile(default_grb_fpath):
```
pirat89, line 268: @Rezney aaah...I think I wanted to ask @smitterl  abou existence of this file on s390x... as this one is created during installation of the system. I expect the file does not exist on s390x. but better to check.

repos/system_upgrade/common/actors/systemfacts/actor.py
```diff
@@ -41,3 +57,8 @@ def process(self):
...
```
pirat89, line 60:
```suggestion
        self.produce(systemfacts.get_default_grub_conf())
```

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -242,3 +260,47 @@ def _get_firewall_status(service_name):
...
+                title='How to re-create the missing /etc/default/grub file in Red Hat Enterprise Linux 7?'
+            ),
+        ])
+
+    for line in run(['cat', '/etc/default/grub'], split=True)['stdout']:
+        name, value = tuple(map(type(line).strip, line.split('=', 1)))
+        yield DefaultGrub(
+            name=name,
+            value=value
+        )
```
pirat89, line 291:
```suggestion
        ])
    return []

    for line in run(['cat', '/etc/default/grub'], split=True)['stdout']:
        name, value = tuple(map(type(line).strip, line.split('=', 1)))
        yield DefaultGrub(
            name=name,
            value=value
        )
```

repos/system_upgrade/common/actors/systemfacts/actor.py
pirat89: and question is, in case the file doesn't exist on s390x, should be the msg produced? (someone should check s390x)

repos/system_upgrade/el8toel9/actors/checkblsgrubcfgonppc64/libraries/blsgrubcfgonppc64.py
```diff
@@ -0,0 +1,42 @@
+from leapp import reporting
+from leapp.libraries.common import grub
+from leapp.libraries.stdlib import api
+from leapp.models import DefaultGrubInfo, GrubCfgBios
+
+
+def _is_system_ppc64():
+
```
pirat89, line 8:
```suggestion
```

repos/system_upgrade/el8toel9/actors/checkblsgrubcfgonppc64/libraries/blsgrubcfgonppc64.py
```diff
@@ -0,0 +1,42 @@
...
+    arch = api.current_actor().configuration.architecture
+    return bool(arch == 'ppc64le')
+
+
+def process():
+
```
pirat89, line 14:
```suggestion
```

pirat89: pls remove all unneeded blank lines after the function header :)

repos/system_upgrade/el8toel9/actors/grub2mkconfigonppc64/actor.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class Grub2MkconfigOnPpc64(Actor):
+    """
+    Actor runs 'grub2-mkconfig' to regenerate Grub config after RHEL 8 to RHEL 9
+    rpm transaction in case system is running on ppc64le, GRUB_ENABLE_BLSCFG=true
+    and '/boot/grub2/grub.cfg' config file does not yet work with BLS entries.
+    For the system to successfully boot into el9 kernel, it is essential that the
+    grub config file is abvle to work with the BLS entries.
+    """
```
pirat89, line 14:
```
"""
title

description
"""

File: repos/system_upgrade/el8toel9/actors/grub2mkconfigonppc64/libraries/grub2mkconfigonppc64.py [hunk 15]
```diff
@@ -0,0 +1,36 @@
...
+from leapp.models import DefaultGrubInfo
+
+GRUB_CFG_PATH = '/boot/grub2/grub.cfg'
+
+
+def _is_system_ppc64():
+    arch = api.current_actor().configuration.architecture
+    return bool(arch == 'ppc64le')
+
+
```
pirat89, line 12: drop

```suggestion
```

repos/system_upgrade/common/actors/systemfacts/actor.py
Rezney: Nope, it should not. Good catch!

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
Rezney: It does not exist.

repos/system_upgrade/el8toel9/actors/grub2mkconfigonppc64/tests/test_grub2mkconfigonppc64.py
```diff
@@ -0,0 +1,78 @@
...
+
+CUR_DIR = os.path.dirname(os.path.abspath(__file__))
+
+
+class RunMocked(object):
+    def __init__(self):
+        self.called = 0
+
+    def __call__(self):
+        self.called += 1
```
pirat89, line 20: it's not used anywhere

pirat89 (commented):
drop pls the dead code

pirat89 (approved):
lgtm now \o/

pirat89:
waiting for test results from a QE team. it can be merged if it passes their testing

pirat89 (dismissed):

Rezney:
@pirat89 please review the latest changes. The inhibitor was discussed with @shaded-enmity

pirat89 (commented):
@Rezney lgtm. make just linters & unit-tests happy

pirat89 (approved):

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -254,4 +259,51 @@ def _get_firewall_status(service_name):
...
+
+
+@aslist
+def _default_grub_info():
+    default_grb_fpath = '/etc/default/grub'
+    if not os.path.isfile(default_grb_fpath):
+        reporting.create_report([
+            reporting.Title('File "{}" does not exist!'.format(default_grb_fpath)),
+            reporting.Summary(
+                'Leapp detected "{}" does not exist. The file is essential for the in-place upgrade '
```
fernflower, line 275: Hmmm isn't it missing `.format(default_grb_path)` ?

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -254,4 +259,51 @@ def _get_firewall_status(service_name):
...
+                'to finish successfully. This scenario might have occured if the system was already '
+                'upgraded from RHEL 6. Please re-create the file manually.'
+            ),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Flags([reporting.Flags.INHIBITOR]),
+            reporting.Tags([reporting.Tags.BOOT]),
+            reporting.RelatedResource('file', default_grb_fpath),
+            reporting.ExternalLink(
+                url='https://access.redhat.com/solutions/3185891',
+                title='How to re-create the missing /etc/default/grub file in Red Hat Enterprise Linux 7?'
```
fernflower, line 285: Nit: you can reuse default_grb_path here as you are doing above

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -254,4 +259,51 @@ def _get_firewall_status(service_name):
...
+            ),
+        ])
+    else:
+        for line in run(['cat', '/etc/default/grub'], split=True)['stdout']:
```
fernflower, line 289: Nit: same here

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
Rezney: nice catch

mmacura311:
tested this PR when comparison to master, this fixes the "old kernel" issue on virtual machines and the inhibitor correctly kicks in on baremetal ppc64le machines.

pirat89 (approved):

PR #827, opened 2022-02-01
Refactor utils/get_latest_copr_build

fernflower:
Recent attempt to refactor leapp's tmt-workflow to
enable both 7to8 and 8to9 upstream regression testing
has shown that we need some tooling to get most recent
leapp-repository's master copr build id.
So doing some necrocoding here.

### DIFF
--- a/utils/get_latest_copr_build
+++ b/utils/get_latest_copr_build
@@ -2,83 +2,92 @@
 
 import json
 import os
+import re
 import sys
 
 import copr.v3
 
+ENV_VARS = {
+    '_COPR_CONFIG': '~/.config/copr',  # Copr config file. Get it through https://<copr instance>/api/.
+    'COPR_OWNER': '',  # Owner of the Copr project
+    'COPR_PROJECT': '',  # The Copr project to search
+    'COPR_PACKAGE': '',  # Name of the package to look for. This is optional - if empty, any package in the
+                         # project is considered.
+    'PKG_RELEASE': '',  # The release part of the pkg NEVRA string, e.g.
+                        #  0.201906041623Z.f82f863.add_missing_deps.PR231
+    'COPR_REPO': '',  # An alternative to COPR_OWNER & COPR_PROJECT. This env var should hold "owner/project".
+    'REGEX': ''   # A more general way to search for release_id matches via regex. Generalization of PKG_RELEASE
+}
+# override defaults with environment variables
+for env, default in ENV_VARS.items():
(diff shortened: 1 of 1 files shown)


utils/get_latest_copr_build
```diff
@@ -2,83 +2,90 @@
...
 for env, default in ENV_VARS.items():
     ENV_VARS[env] = os.getenv(env, default)
-
 CONFIG_PATH = os.path.expandvars(ENV_VARS['_COPR_CONFIG'])
-client = copr.v3.Client(copr.v3.config_from_file(path=CONFIG_PATH))
-ownername = ENV_VARS['COPR_OWNER']
-projectname = ENV_VARS['COPR_PROJECT']
-
-if projectname and not ownername:
-    ownername = get_owner_from_config()
```
fernflower, line 43: I have on purpose removed ownername guessing from config as it is not used in the use case I need and is imho overcomplicating thing. username != owner generally speaking.

pirat89: @fernflower originally, the script has been executed with the config of the owner, which has been shared in our team. but I agree that this change seems better to me regarding nowadays expactations and switch to public repos.

utils/get_latest_copr_build
```diff
@@ -2,83 +2,89 @@
...
+            ownername, projectname = ENV_VARS['COPR_REPO'].split('/', 1)
+        else:
+            projectname = ENV_VARS['COPR_REPO']
+    # If after all those actions either owner or project is not defined - give up and fail
+    if not ownername or not projectname:
+        error = ('Error: Use either COPR_REPO env var in a format "owner/project" or '
+                 'COPR_OWNER & COPR_PROJECT env vars to specify the Copr repository to search in.')
+        _fail(error)
+    # Check that match criteria is defined
+    match_criteria = ENV_VARS['REGEX'] or ENV_VARS['PKG_RELEASE']
```
pirat89, line 77: @fernflower maybe the script should print at least a warning or somethine, in case the REGEX + PKB_RELEASE are set in the same time, as in such a case only the REGEX is used. wdyt?

fernflower: Agreed, should be like this now:
```
PKG_RELEASE='12321' REGEX='leapp-repository.*master.*' COPR_REPO='@oamg/leapp' python utils/get_latest_copr_build
Warning: Both REGEX and PKG_RELEASE were set - REGEX has higher priority
3289015
```

pirat89 (approved):
awesome!!

PR #828, opened 2022-02-07
WIP: Enable RHEL 8 to RHEL 9 upgrades using RHUI

Rezney:
- make RHUI_CLOUD_MAP aware of multiple major version paths
- make some RHUI_CLOUD_MAP entities upgrade path agnostic (src / target vs.
  el7 / el8)
- remove unsupported cloud architectures


repos/system_upgrade/common/libraries/module.py
```diff
@@ -16,7 +16,9 @@
 def _create_or_get_dnf_base(base=None):
     if not base:
         base = dnf.Base()
+        base.init_plugins()
         base.read_all_repos()
+        base.configure_plugins()
```
pirat89, line 21: @Rezney  To make it more clear for people, why the plugins are not configured immadiately after the initialisation.
```suggestion
        # configure plugins after the repositories are loaded
        # e.g. the amazon-id plugin requires loaded repositories
        # for the proper configuration.
        base.configure_plugins()
```

pirat89 (approved):
Discussed in person. The `RequiredTargetUserspacePackages` is deprecated and should not be used anymore, but the actor is expected to contain more changes in close future and the testing is already in progress, so let's keep other changes after the release. We are going with this change now.

alexxa:
Tested for AWS SAP images upgrading to 8.2 and 8.4. Evrth works. 
Thanks @Rezney

pirat89 (approved):

PR #830, opened 2022-02-11
Provide LEAPP_IPU_IN_PROGRESS envar

pirat89:
The environement variable contains string in format XtoY, where
*  X is major version of the current system
*  Y is major version of the target system

So contains:
*  `7to8` for IPU 7 -> 8,
*  `8to9` for IPU 8 -> 9

this envar is supposed to be used e.g. in RPM scriptlets in case
maintainers want to treat the upgrade correctly during the IPU
process between major releases of RHEL.

As a benefit, this allows people to handle situations when specific
action has to happen in the middle of the DNF transaction - as leapp
actors can do actions only before and after the transaction.

### DIFF
--- a/commands/upgrade/util.py
+++ b/commands/upgrade/util.py
@@ -204,6 +204,12 @@ def prepare_configuration(args):
     os.environ['LEAPP_UPGRADE_PATH_TARGET_RELEASE'] = target_version
     os.environ['LEAPP_UPGRADE_PATH_FLAVOUR'] = flavor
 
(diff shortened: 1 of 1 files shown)


Rezney:
So far tested it on 8 > 9 and everything looks ok. Will test it on 7 > 8 on Monday just to be sure.

Rezney:
Works fine and the env var is accessible in the container: 

```
Feb 14 08:45:17 localhost upgrade[615]: LEAPP_IPU_IN_PROGRESS=7to8
```

Rezney (approved):

PR #833, opened 2022-02-16
Move "multipathconfread" into common repository

Rezney:
"multipathconfread" actor is compatible with RHEL 8 so let's move
it into common repository. For the moment it will not be of any use but
will help us to fix issue with enabling multipath as we refactored some
parts of related functionality there.


xiachen-rh:
We were able to upgrade RHEL 8.6 to RHEL 9.0 using this build on the IBM Cloud PowerVS instance.

pirat89 (approved):

PR #838, opened 2022-02-22
targetinitramfsgenerator: remove extra quotes causing dracut install failures

MichalHe:
The presence of extra double quotes in the dracut --install and --add
parameters is causing the target initramfs generation to fail. This
patch removes the extra quotes.

## Steps to verify that the current command crashes
Place the problematic code to some of the actors executed early on (e.g. `SystemFactsActor`). This way, it is not necessary to wait until the target initramfs is being built. The code precisely mimics how the dracut command is prepared in the targetinitramfsgenerator.
```python3
        from leapp.libraries.stdlib import run
        files = ['/etc/os-release']
        dracut = ['dracut', '-f', '--install', '"{}"'.format(' '.join(files))]
        run(dracut)
```
Same for the `--add`:
```python3
        from leapp.libraries.stdlib import run
        files = ['dm']
        dracut = ['dracut', '-f', '--add', '"{}"'.format(' '.join(files))]
        run(dracut)
```


pirat89 (approved):
lgtm. waiting for results of integration tests. if they pass, we can merge it.

PR #839, opened 2022-02-23
el8toel9: Add actor that checks upgradability of ifcfg files

lkundrak:
RHEL 9 is not going to ship network-scripts. While NetworkManager
generally supports ifcfg files just fine, there are some corner cases
that are either unsupportable or explicitely configured not to work with
NetworkManager.

This adds an actor that checks for those cases and inhibits upgrade if
they are encountered.

https://bugzilla.redhat.com/show_bug.cgi?id=2052343

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/checkifcfg/actor.py
+++ b/repos/system_upgrade/el8toel9/actors/checkifcfg/actor.py
@@ -1,0 +1,24 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import checkifcfg_ifcfg as ifcfg
+from leapp.models import InstalledRPM, Report, RpmTransactionTasks
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+
+
+class CheckIfcfg(Actor):
+    """
+    Ensures that ifcfg files are compatible with NetworkManager
+
+    Checks whether the ifcfg files would work with NetworkManager's ifcfg-rh
+    compatibility module -- they are of known type, well-formed and not
+    explicitly disabled with NM_CONTROLLED=no.
+
+    Makes sure relevant NetworkManager modules end up getting installed.
+    """
+
+    name = "check_ifcfg"
+    consumes = (InstalledRPM,)
+    produces = (Report, RpmTransactionTasks,)
+    tags = (IPUWorkflowTag, FactsPhaseTag,)
+
+    def process(self):
+        ifcfg.process()
--- a/repos/system_upgrade/el8toel9/actors/checkifcfg/tests/unit_test_ifcfg.py
+++ b/repos/system_upgrade/el8toel9/actors/checkifcfg/tests/unit_test_ifcfg.py
@@ -1,0 +1,147 @@
+import mock
+import six
+
+from leapp import reporting
+from leapp.libraries.actor import checkifcfg_ifcfg as ifcfg
+from leapp.libraries.common.testutils import create_report_mocked, CurrentActorMocked, produce_mocked
+from leapp.libraries.stdlib import api
+from leapp.models import InstalledRPM, RPM, RpmTransactionTasks
+
+RH_PACKAGER = 'Red Hat, Inc. <http://bugzilla.redhat.com/bugzilla>'
+
+NETWORK_SCRIPTS_RPM = RPM(
+    name='network-scripts', version='10.00.17', release='1.el8', epoch='',
+    packager=RH_PACKAGER, arch='x86_64',
+    pgpsig='RSA/SHA256, Fri 04 Feb 2022 03:32:47 PM CET, Key ID 199e2f91fd431d51'
+)
+
+NETWORK_MANAGER_RPM = RPM(
+    name='NetworkManager', version='1.36.0', release='0.8.el8', epoch='1',
+    packager=RH_PACKAGER, arch='x86_64',
+    pgpsig='RSA/SHA256, Mon 14 Feb 2022 08:45:37 PM CET, Key ID 199e2f91fd431d51'
+)
+
+INITSCRIPTS_INSTALLED = CurrentActorMocked(
+    msgs=[InstalledRPM(items=[NETWORK_SCRIPTS_RPM])]
+)
+
+INITSCRIPTS_AND_NM_INSTALLED = CurrentActorMocked(
+    msgs=[InstalledRPM(items=[NETWORK_SCRIPTS_RPM, NETWORK_MANAGER_RPM])]
+)
+
+
+def test_ifcfg_none(monkeypatch):
+    """
+    No report and don't install anything if there are no ifcfg files.
+    """
+
+    monkeypatch.setattr(ifcfg.api, 'current_actor', INITSCRIPTS_AND_NM_INSTALLED)
+    monkeypatch.setattr(ifcfg.api, "produce", produce_mocked())
+    monkeypatch.setattr(ifcfg.os, 'listdir', lambda dummy: ('hello', 'world',))
+    monkeypatch.setattr(ifcfg.os.path, 'isfile', lambda dummy: True)
+    monkeypatch.setattr(reporting, "create_report", create_report_mocked())
+    ifcfg.process()
+    assert not reporting.create_report.called
+    assert not api.produce.called
+
+
+def test_ifcfg_rule_file(monkeypatch):
+    """
+    Install NetworkManager-dispatcher-routing-rules package if there's a
+    file with ip rules.
+    """
+
+    monkeypatch.setattr(ifcfg.api, 'current_actor', INITSCRIPTS_AND_NM_INSTALLED)
+    monkeypatch.setattr(ifcfg.api, "produce", produce_mocked())
+    monkeypatch.setattr(ifcfg.os, 'listdir', lambda dummy: ('hello', 'world', 'rule-eth0',))
+    monkeypatch.setattr(ifcfg.os.path, 'isfile', lambda dummy: True)
+    monkeypatch.setattr(reporting, "create_report", create_report_mocked())
+    ifcfg.process()
+    assert not reporting.create_report.called
+    assert api.produce.called
+    assert isinstance(api.produce.model_instances[0], RpmTransactionTasks)
+    assert api.produce.model_instances[0].to_install == ['NetworkManager-dispatcher-routing-rules']
+
+
+def test_ifcfg_good_type(monkeypatch):
+    """
+    No report if there's an ifcfg file that would work with NetworkManager.
+    Make sure NetworkManager itself is installed though.
+    """
+
+    mock_config = mock.mock_open(read_data="TYPE=Ethernet")
+    with mock.patch("builtins.open" if six.PY3 else "__builtin__.open", mock_config) as mock_ifcfg:
+        monkeypatch.setattr(ifcfg.api, 'current_actor', INITSCRIPTS_AND_NM_INSTALLED)
+        monkeypatch.setattr(ifcfg.api, "produce", produce_mocked())
+        monkeypatch.setattr(ifcfg.os, 'listdir', lambda dummy: ('hello', 'world', 'ifcfg-eth0', 'ifcfg-lo',))
+        monkeypatch.setattr(ifcfg.os.path, 'isfile', lambda dummy: True)
+        monkeypatch.setattr(reporting, "create_report", create_report_mocked())
+        ifcfg.process()
+        mock_ifcfg.assert_called_once_with('/etc/sysconfig/network-scripts/ifcfg-eth0')
+        assert not reporting.create_report.called
+        assert api.produce.called
+        assert isinstance(api.produce.model_instances[0], RpmTransactionTasks)
+        assert api.produce.model_instances[0].to_install == ['NetworkManager']
+
+
+def test_ifcfg_not_controlled(monkeypatch):
+    """
+    Report if there's a NM_CONTROLLED=no file.
+    """
+
+    mock_config = mock.mock_open(read_data="TYPE=Ethernet\nNM_CONTROLLED=no")
+    with mock.patch("builtins.open" if six.PY3 else "__builtin__.open", mock_config) as mock_ifcfg:
+        monkeypatch.setattr(ifcfg.api, 'current_actor', INITSCRIPTS_INSTALLED)
+        monkeypatch.setattr(ifcfg.api, "produce", produce_mocked())
+        monkeypatch.setattr(ifcfg.os, 'listdir', lambda dummy: ('hello', 'world', 'ifcfg-eth0',))
+        monkeypatch.setattr(ifcfg.os.path, 'isfile', lambda dummy: True)
+        monkeypatch.setattr(reporting, "create_report", create_report_mocked())
+        ifcfg.process()
+        mock_ifcfg.assert_called_once_with('/etc/sysconfig/network-scripts/ifcfg-eth0')
+        assert reporting.create_report.called
+        assert 'disabled NetworkManager' in reporting.create_report.report_fields['title']
+        assert api.produce.called
+
+
+def test_ifcfg_unknown_type(monkeypatch):
+    """
+    Report if there's configuration for a type we don't recognize.
+    """
+
+    mock_config = mock.mock_open(read_data="TYPE=AvianCarrier")
+    with mock.patch("builtins.open" if six.PY3 else "__builtin__.open", mock_config) as mock_ifcfg:
+        monkeypatch.setattr(ifcfg.api, 'current_actor', INITSCRIPTS_AND_NM_INSTALLED)
+        monkeypatch.setattr(ifcfg.api, "produce", produce_mocked())
+        monkeypatch.setattr(ifcfg.os, 'listdir', lambda dummy: ('hello', 'world', 'ifcfg-pigeon0',))
+        monkeypatch.setattr(ifcfg.os.path, 'isfile', lambda dummy: True)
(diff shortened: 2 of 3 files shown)


repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # Don't bother reporting the file for NM_CONTROLLED=no
+        # if its type is not supportable with NetworkManager anyway
+        if bad_type is True:
+            bad_type_files.append(f.path)
+        elif nm_controlled is False:
+            not_controlled_files.append(f.path)
+
+    if rule_files:
+        title = 'Rule file detected'
```
pirat89, line 76: This is generic too much. From the title, I cannot even imagine what is it about. Imagine you see by default just this line. Create the title that gives you an idea about what the report is.

pirat89: (this is not review yet. just a one thing that I see right now)

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
```diff
@@ -0,0 +1,145 @@
+import os
+
+from leapp import reporting
+
+
+def process():
+    TRUE_VALUES = ['yes', 'true', '1']
+    GOOD_TYPES = [
+        'ethernet', 'ctc', 'wireless', 'infiniband', 'bond',
```
tyll, line 9: wireless with WEP also needs to inhibit the upgrade

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
```diff
@@ -0,0 +1,145 @@
...
+            continue
+
+        if f.name.startswith('rule-') or f.name.startswith('rule6-'):
+            rule_files.append(f.path)
+            continue
+
+        if not f.name.startswith('ifcfg-'):
+            continue
+
+        if f.name == 'ifcfg-lo':
```
tyll, line 39: If this configures anything for lo like DNS options, should this inhibit the upgrade, too? Seems best to request to get this file migrated, too.

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
```diff
@@ -0,0 +1,145 @@
...
+            continue
+
+        for line in open(f.path):
+            try:
+                (key, value) = line.split('#')[0].strip().split('=')
+            except ValueError:
+                # We're not interested in lines that are not
+                # simple assignments. Play it safe.
+                continue
```
tyll, line 48: If there are complex structures that cannot be parsed, it seems it would be best to inhibit the upgrade, wouldn't it?

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
```diff
@@ -0,0 +1,145 @@
...
+
+    if not_controlled_files:
+        title = 'Network configuration with disabled NetworkManager support detected'
+        summary = ('RHEL 9 does not support the legacy network-scripts'
+                   ' package that was deprecated in RHEL 8 in favor of'
+                   ' NetworkManager. Configuration present in the system'
+                   ' prohibit NetworkManager from loading it.')
+        remediation = ('Ensure the ifcfg files comply with format described in'
+                       ' nm-settings-ifcfg-rh(5) manual and remove the'
+                       ' NM_CONTROLLED key from them.')
```
tyll, line 129: Best would be to recommend to migrate to keyfiles and if this is not an option, ensure that the file complies with the nm format.

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
tyll: team and wireless require plugins, how is this going to be handled?

liangwen12year:
It seems to me that you did not include a `unit_test_ifcfg.py`, it is required to have such file, right @pirat89  ?

repos/system_upgrade/el8toel9/actors/checkifcfg/actor.py
```diff
@@ -0,0 +1,21 @@
...
+from leapp.models import Report
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+
+
+class CheckCustomNetworkScripts(Actor):
+    """
+    Checks whether ifcfg files are compatible with NetworkManager
+
+    Checks whether the ifcfg files would work with NetworkManager's ifcfg-rh
+    compatibility module -- they are of know type, well-formed and not
```
liangwen12year, line 12: `known type` ?

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
```diff
@@ -0,0 +1,145 @@
...
+            reporting.RelatedResource('file', fname)
+            for fname in rule_files
+        ])
+
+    if bad_type_files:
+        title = 'Network configuration for unsupported device types detected'
+        summary = ('RHEL 9 does not support the legacy network-scripts'
+                   ' package that was deprecated in RHEL 8 in favor of'
+                   ' NetworkManager. Files for device types that are not'
+                   ' supported by NetworkManager are present in the system.')
```
liangwen12year, line 104: Maybe it is better to include the invalid device types that are present in the system.

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
```diff
@@ -0,0 +1,145 @@
...
+            reporting.Title(title),
+            reporting.Summary(summary),
+            reporting.Remediation(hint=remediation),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Tags([reporting.Tags.NETWORK, reporting.Tags.SERVICES]),
+            reporting.Flags([reporting.Flags.INHIBITOR]),
+            reporting.RelatedResource('package', 'NetworkManager'),
+        ] + [
+            reporting.RelatedResource('file', fname)
+            for fname in bad_type_files
```
liangwen12year, line 118: How about including the external link for `nm-settings-ifcfg-rh(5) manual` here?

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
lkundrak: > wireless with WEP also needs to inhibit the upgrade

This actor deals with ifcfg files. WEP is entirely out of scope here.

> team and wireless require plugins, how is this going to be handled?

They're presumably already installed in the system that's being upgraded. Why else would the configuration be there.

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
lkundrak: NetworkManager just ignores ifcfg-lo altogether. If you put DNS options into there, then it has no effect.

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
lkundrak: It's not necessarily complex structures. It's anything multi-line, comments, etc.
We quite intentionally don't implement a full-fledged parser here, because 1.) it's very difficult to get right 2.) we really only need to read TYPE and NM_CONTROLLED. We're not in the position to judge whether NM would be able to load the file here. Also, NM's parser also just ignores things it can't grok.

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
lkundrak: That would be a good idea if we had a migration tool, which we don't have.
Suggesting that people just somehow migrate the configuration without explaining how is just going to do more harm than good.
Also, RHEL 9 supports ifcfg files just fine, there's no need to do this.

This is likely going to come up in future RHEL version though.

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
lkundrak: Well, anything we don't recognize is invalid. The set of valid types is finite, set of invalid ones is potentially infinite.
In other words: with network-scripts someone could just use a TYPE=[pigeons](https://datatracker.ietf.org/doc/html/rfc1149)  and drop in an ifup-pigeons scripts and things would happily work.

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
```diff
@@ -0,0 +1,145 @@
...
+
+    if bad_type_files:
+        title = 'Network configuration for unsupported device types detected'
+        summary = ('RHEL 9 does not support the legacy network-scripts'
+                   ' package that was deprecated in RHEL 8 in favor of'
+                   ' NetworkManager. Files for device types that are not'
+                   ' supported by NetworkManager are present in the system.')
+        remediation = ('Consult the nm-settings-ifcfg-rh(5) manual for'
+                       ' valid types of ifcfg files. Remove configuration'
+                       ' files that can not be supported.')
```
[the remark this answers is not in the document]
lkundrak: Well, generally, no.

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
lkundrak: I intentionally omitted it because if there's some unsupported device type, then we can't guess what type it actually is or what is it supposed to do, and the ifcfg manual is unlikely to help.

It's difficult to guess what could be happening at that time, but it's likely that the only option the user has is to remove the file and do whatever they were trying to achieve differently (e.g. with a shell script or something).

repos/system_upgrade/el8toel9/actors/checkifcfg/actor.py
lkundrak: Fixed. Thanks!

lkundrak:
> It seems to me that you did not include a `unit_test_ifcfg.py`, it is required to have such file, right @pirat89 ?

Thanks for pointing this out. Added a test now.

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
lkundrak: Tried to improve this in an updated patch. I'm not sure if it's good enough because I don't have a very clear idea about where does the user see this message. If it's still not clear enough, I'd be very thankful for suggestions.

lkundrak:
Pushed an updated version. Most notable change is an addition of a test and getting rid of some Python3-isms that made the test fail.

lkundrak:
Reworked things quite a bit, because @tyll suggested we need to support the case where ```network-scripts``` are installed, but no ```NetworkManager``` initially. Now the actor pulls in the necessary packages and the test suite has been updated accordingly.

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
tyll: at this point the plugin knows which unsupported types are present, so the summary could mention those.

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
tyll: In this message the users are asked to manually check the ifcfg files for compliance with nm-settings-ifcfg-rh(5). The time spent in doing this seems to be better spent in understanding keyfiles.

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
tyll: Yes, if NetworkManager ignores this file but network-scripts does not, then this means the file needs to be migrated/the update needs to be blocked, otherwise the behavior might change/things might get broken.

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
tyll: I agree with your observations. The consequences are different for me, though. If there are indications that NM might ignore something, then the upgrade should be blocked IMHO.

repos/system_upgrade/el8toel9/actors/checkifcfg/tests/unit_test_ifcfg.py
```diff
@@ -0,0 +1,147 @@
...
+)
+
+INITSCRIPTS_AND_NM_INSTALLED = CurrentActorMocked(
+    msgs=[InstalledRPM(items=[NETWORK_SCRIPTS_RPM, NETWORK_MANAGER_RPM])]
+)
+
+
+def test_ifcfg_none(monkeypatch):
+    """
+    No report and don't install anythin if there are no ifcfg files.
```
tyll, line 35:
```suggestion
    No report and don't install anything if there are no ifcfg files.
```

liangwen12year (commented):
Typo in commit message, `that are either unsupportable or explicitely configured not to work with
NetworkManager.`, should be "explicitly".

liangwen12year:
Except for the typo above,  LGTM.

repos/system_upgrade/el8toel9/actors/checkifcfg/tests/unit_test_ifcfg.py
```diff
@@ -0,0 +1,147 @@
...
+def test_ifcfg_unknown_type(monkeypatch):
+    """
+    Report if there's configuration for a type we don't recognize.
+    """
+
+    mock_config = mock.mock_open(read_data="TYPE=AvianCarrier")
+    with mock.patch("builtins.open" if six.PY3 else "__builtin__.open", mock_config) as mock_ifcfg:
+        monkeypatch.setattr(ifcfg.api, 'current_actor', INITSCRIPTS_AND_NM_INSTALLED)
+        monkeypatch.setattr(ifcfg.api, "produce", produce_mocked())
+        monkeypatch.setattr(ifcfg.os, 'listdir', lambda _: ('hello', 'world', 'ifcfg-pigeon0',))
```
pirat89, line 115: please do not use `_` varname anywhere. Use `dummy` instead

tyll: Is it ok to use `unused` instead of `dummy`? `dummy` is not recommended because it can also be used as an insult.

repos/system_upgrade/el8toel9/actors/checkifcfg/libraries/ifcfg.py
```diff
@@ -0,0 +1,143 @@
...
+            reporting.RelatedResource('file', fname)
+            for fname in bad_type_files
+        ])
+
+    if not_controlled_files:
+        title = 'Network configuration with disabled NetworkManager support detected'
+        summary = ('RHEL 9 does not support the legacy network-scripts'
+                   ' package that was deprecated in RHEL 8 in favor of'
+                   ' NetworkManager. Configuration present in the system'
+                   ' prohibit NetworkManager from loading it.')
```
pirat89, line 116: @lkundrak I suggest to add the list of bad files into the summary as well. This report is good for users reading it in cockpit or satellite, but in case people read just the txt form, addititional fields like "resources" are not printed to user.

```suggestion
        FMT_LIST_SEPARATOR = '\n    - '
        summary = ('RHEL 9 does not support the legacy network-scripts'
                   ' package that was deprecated in RHEL 8 in favor of'
                   ' NetworkManager. Configuration present in the system'
                   ' prohibit NetworkManager from loading it. Files with the problematic configuration:{}'
        ).format(''.join(['{}{}'.format(FMT_LIST_SEPARATOR, bfile) for bfile in bad_files]))
```

repos/system_upgrade/el8toel9/actors/checkifcfg/actor.py
```diff
@@ -0,0 +1,24 @@
...
+    Checks whether the ifcfg files would work with NetworkManager's ifcfg-rh
+    compatibility module -- they are of known type, well-formed and not
+    explicitly disabled with NM_CONTROLLED=no.
+
+    Makes sure relevant NetworkManager modules end up getting installed.
+    """
+
+    name = "check_ifcfg"
+    produces = (Report, RpmTransactionTasks,)
+    consumes = (InstalledRPM,)
```
pirat89, line 20:
```suggestion
    consumes = (InstalledRPM,)
    produces = (Report, RpmTransactionTasks,)
```

repos/system_upgrade/el8toel9/actors/checkifcfg/tests/unit_test_ifcfg.py
pirat89: unused is good as well if it's accepted by linters

pirat89 (approved):
lgtm. Discussed in person the code is expected to be split into two actors in future, but regarding the time it's ok to keep it as it is for the upcoming release and refactore it for the another one without the stress. let's wait for the test result. if everything is ok, we can merge it

PR #841, opened 2022-02-24
Introduction of DNFWorkarounds

vinzenz:
Previously we have had only one workaround to handle before we were
running the DNF transaction. That was implemented by executing the
'handleyumconfig' script. This was applicable for el7 => el8 upgrades.
Now we are currently having a WIP PR for handling satellite upgrades,
and the need for such a workaround has emerged there as well. The
problem is just, that it cannot be properly conditionalized the way how
it is implemented at the moment without importing too much irrelevant
things to the dnfplugin part of the workflow.
To solve this, this patch will introduce now the DNFWorkaround model,
that will allow actors to register a workaround script to be registered
when the right conditions are met and all of the workarounds are
processed as received before dnf is called. (Just like it was the case
for the handleyumconfig script before).

The advantage of this implementation is that it will now make this whole
approach more modular and can be used by 3rd parties and customers a
like to workaround their problems by just creating an actor that
registers their own script.

As part of this patch, the changes that are necessary for
handleyumconfig have been already applied.


repos/system_upgrade/common/libraries/utils.py
```diff
@@ -46,31 +46,6 @@ def makedirs(path, mode=0o777, exists_ok=True):
...
-    except OSError as e:
-        raise StopActorExecutionError(
-            message='Failed to exceute script to apply yum adjustment. Message: {}'.format(str(e))
-        )
-    except CalledProcessError as e:
-        raise StopActorExecutionError(
-            message='Failed to apply yum adjustment. Message: {}'.format(str(e))
-        )
-
-
```
pirat89, line 73: must be kept and marked for deprecation

vinzenz: Maybe, didn't think so though - Because it doesn't make much sense to be used other than the current use cases.

pirat89: that's true, but keeping it and set it as deprecated does not hurt us. I prefer to keep our promise about the deprecation until we document / specify functions/models/.... are not supposed to be used by others.

PR #842, opened 2022-03-01
el8toel9: Add actor that scans for deprecated network settings

lkundrak:
RHEL 9 is going to deprecate certain network configurations. In
particular, it will ship without support for WEP networks.

Let's make sure the users are not caught by surprise that their
networks no longer connect.

Pick a generic name for the actor, as there are possibly more deprecations
coming in future.

https://bugzilla.redhat.com/show_bug.cgi?id=2056925


repos/system_upgrade/el8toel9/actors/networkdeprecations/libraries/networkdeprecations.py
```diff
@@ -0,0 +1,87 @@
...
+                wep_files.append(path)
+                continue
+
+    if wep_files:
+        title = 'Wireless networks using unsupported WEP encryption detected'
+        summary = ('The Wired Equivalent Privacy (WEP) algorithm used for'
+                   ' authenticating to wireless networks has been phased out'
+                   ' due to known security weaknesses. Configuration for networks'
+                   ' that use the phased out WEP algorithm is present in the system'
+                   ' and will not work agter the upgrade.')
```
pirat89, line 72:
```suggestion
                   ' and will not work after the upgrade.')
```

pirat89: and similar to ifcfg PR, I suggest to inform people about the problematic files in the summary (or remediation..) field.

pirat89:
@lkundrak  This should be split in two actors (scanner, checker). However, let's sync about that first via video call before you do any changes.

pirat89 (approved):

pirat89 (approved):
lgtm. as discussed, the code will be in future split into two actors (scanner, checker), but regarding the time we will keep it in this form for the upcoming release. let's make it more robust for the other one so there is plenty of time to do it. waiting for the test results. I will do also some manual testing before the merge, but so fat it seems good.

PR #847, opened 2022-03-07
packaging: use sha512 sig alg.

pirat89:


pirat89:
just a try...the problem is lvl above

PR #849, opened 2022-03-09
config: Fallback to /etc/leapp/repos.d if missing config option

vinzenz:
Previously if the repositories.repo_path configuration value was missing
from the leapp.conf, because someone deleted it, leapp fell back to to
the current working directory as root path which could have caused
unwanted side effects.
This patch will force /etc/leapp/repos.d to be the default path for
leapp answer/upgrade/preupgrade etc

### DIFF
--- a/commands/answer/__init__.py
+++ b/commands/answer/__init__.py
@@ -1,7 +1,7 @@
 import itertools
 import sys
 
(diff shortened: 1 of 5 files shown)


Rezney (approved):
Works fine.

PR #850, opened 2022-03-09
Bring back the same log level

fernflower:
The --target patch now calls some code before logging is
configured, which causes messages to be logged at WARN level
instead of DEBUG.
This patch should fix that.

### DIFF
--- a/commands/command_utils.py
+++ b/commands/command_utils.py
@@ -3,6 +3,7 @@ import os
 import re
 
 from leapp.exceptions import CommandError
+from leapp.logger import configure_logger
 from leapp.utils import path
 
 HANA_BASE_PATH = '/hana/shared'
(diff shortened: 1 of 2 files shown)


commands/command_utils.py
```diff
@@ -3,6 +3,7 @@
 import re
 
 from leapp.exceptions import CommandError
+from leapp.logger import configure_logger
```
vinzenz, line 6:
```suggestion
from leapp import logger
```

commands/command_utils.py
```diff
@@ -77,6 +78,10 @@ def get_upgrade_paths_config():
     # NOTE(ivasilev) Importing here not to have circular dependencies
     from leapp.cli.commands.upgrade import util  # noqa: C415; pylint: disable=import-outside-toplevel
 
+    # NOTE(ivasilev) This function is called on module load level to figure out choices for --target argument when
+    # logging is not set up yet, so let's configure it on demand.
+    configure_logger('leapp.log')
+
     repository = util.load_repositories_from('repo_path', '/etc/leapp/repo.d/', manager=None)
     with open(path.get_common_file_path(repository, LEAPP_UPGRADE_PATHS)) as f:
         upgrade_paths_map = json.loads(f.read())
```
vinzenz, line 87:
```suggestion
    try:
        logger.configure_logger('leapp.log')

        repository = util.load_repositories_from('repo_path', '/etc/leapp/repo.d/', manager=None)
        with open(path.get_common_file_path(repository, LEAPP_UPGRADE_PATHS)) as f:
            upgrade_paths_map = json.loads(f.read())
    finally:
        logger._logger = None
```

vinzenz:
Another alternative would be to introduce something like configure_logger(filename, force=True) and make configure_logger forcibly recreate it

fernflower:
I like the force configure idea, but I will be able to get back to it only on Friday. So leaving the original one with unifying logs under `leapp-upgrade.log`.
If you decide to implement a force configure fix tomorrow to squeeze in the patch before devel freeze, this PR can be safely closed. Thanks!

vinzenz:
Fixed in https://github.com/oamg/leapp/pull/764

PR #538, opened 2020-07-28
Add actor NisCheck

fila43:
This actor checks configuration of ypbind and nsswitch. Address of NIS server cannot be specified by host name if NIS is used to resolve host names.

PR #851, opened 2022-03-09
el8toel9: Add nis_check actor

mkulik-rh:
NIS components (ypserv, ypbind, yp-tools) will no longer be available in RHEL-9.
Check and notify user about possible solutions when any of the NIS packages is installed and configured.

- It check if any packages (ypbind, ypserver) is installed
- if above statement is true, it check for any not default configuration (depends on detected packages)
- if above statement is true, report is created with installed NIS packages

Related: https://bugzilla.redhat.com/show_bug.cgi?id=2060066

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/nischeck/actor.py
+++ b/repos/system_upgrade/el8toel9/actors/nischeck/actor.py
@@ -1,0 +1,19 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.nischeck import report_nis
+from leapp.models import InstalledRedHatSignedRPM, NISConfig, Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class NISCheck(Actor):
+    """
+    Checks if any of NIS components is installed and configured
+    on the system and warns users about discontinuation.
+    """
+
+    name = 'nis_check'
+    consumes = (InstalledRedHatSignedRPM, NISConfig)
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        report_nis()
--- a/repos/system_upgrade/el8toel9/actors/nischeck/libraries/nischeck.py
+++ b/repos/system_upgrade/el8toel9/actors/nischeck/libraries/nischeck.py
@@ -1,0 +1,72 @@
+from leapp import reporting
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.common.rpms import has_package
+from leapp.libraries.stdlib import api
+from leapp.models import InstalledRedHatSignedRPM, NISConfig
+
+report_summary = (
+    'The NIS components (ypserv, ypbind, and yp-tools) are no longer available in RHEL-9.'
+    ' The technology behind those packages is based an outdated design patterns, which are'
+    ' no longer considered as secure. There is no direct alternative with fully compatible'
+    ' features.'
+)
+
+report_hint = (
+    'The alternatives are LDAP and for some use cases Kerberos or migrating to IPA.'
+)
+
+report_link_url = 'https://access.redhat.com/solutions/5991271'
+
+
+def report_nis():
+    """
+    Create the report if any of NIS packages (RH signed)
+    is installed and configured.
+
+    Should notify user about present NIS compnent package
+    installation, warn them about discontinuation, and
+    redirect them to online documentation for possible
+    alternatives.
+    """
+
+    installed_packages = []
+    configured_rpms = []
+
+    # Get necessary models created by another actors
+    nis_confs = api.consume(NISConfig)
+    nis_conf = next(nis_confs, None)
+    if not nis_conf:
+        raise StopActorExecutionError(
+            'Could not obtain NIS RPM information',
+            details={'details': 'Actor did not receive NISConfig message.'}
+        )
+
+    if next(nis_confs, None):
+        api.current_logger().warning('Unexpectedly received more than one NISConfig message.')
+
+    configured_rpms = nis_conf.nis_not_default_conf
+
+    installed_packages = [package for package in (
+        'ypserv', 'ypbind') if has_package(InstalledRedHatSignedRPM, package)]
+
+    # Final list of NIS packages (configured and installed)
+    rpms_configured_installed = [x for x in installed_packages if x in configured_rpms]
+
+    if not rpms_configured_installed:
+        return
+
+    # Create report
+    report_content = [
+        reporting.Title('NIS component has been detected on your system'),
+        reporting.Summary(report_summary),
+        reporting.Severity(reporting.Severity.MEDIUM),
+        reporting.Tags([reporting.Tags.SERVICES]),
+        reporting.ExternalLink(title='RHEL 9 (NIS) discontinuation',
+                               url=report_link_url),
+        reporting.Remediation(hint=report_hint),
+    ]
+
+    related_resources = [reporting.RelatedResource('package', pkg) for pkg in rpms_configured_installed]
+    report_content += related_resources
+
+    reporting.create_report(report_content)
--- a/repos/system_upgrade/el8toel9/actors/nischeck/tests/test_nischeck.py
+++ b/repos/system_upgrade/el8toel9/actors/nischeck/tests/test_nischeck.py
@@ -1,0 +1,71 @@
+import functools
+
+import pytest
+
+from leapp import reporting
+from leapp.libraries.actor import nischeck
+from leapp.libraries.common.testutils import create_report_mocked, CurrentActorMocked
+from leapp.libraries.stdlib import api
+from leapp.models import InstalledRedHatSignedRPM, NISConfig, RPM
+
+_generate_rpm = functools.partial(RPM,
+                                  pgpsig='RSA/SHA256, Mon 01 Jan 1970 00:00:00 AM -03, Key ID 199e2f91fd431d51',
+                                  packager='Red Hat, Inc. <http://bugzilla.redhat.com/bugzilla>',
+                                  arch='noarch')
+
+
+@pytest.mark.parametrize('pkgs_installed', [
+    ('ypserv',),
+    ('ypbind',),
+    ('ypserv-not-ypbind',),
+    ('ypbind', 'ypserv'),
+])
+@pytest.mark.parametrize('pkgs_configured', [
+    (),
+    ('ypbind',),
+    ('ypserv',),
+    ('ypbind', 'ypserv'),
+])
+def test_actor_nis(monkeypatch, pkgs_installed, pkgs_configured):
+    """
+    Parametrized helper function for test_actor_* functions.
+
+    First generate list of RPM models based on set arguments. Then, run
+    the actor feeded with our RPM list and mocked functions. Finally, assert
+    Reports according to set arguments.
+
+    Parameters:
+        pkgs_installed  (touple): installed pkgs
+        fill_conf_file  (bool): not default ypbind config file
+        fill_ypserv_dir  (bool): not default ypserv dir content
+    """
+
+    # Generate few standard packages
+    rpms = [_generate_rpm(name='rclone', version='2.3', release='3', epoch='1'),
+            _generate_rpm(name='gdb', version='2.0.1', release='1', epoch='2')]
+
+    # Generate packages from 'pkgs_installed'
+    for pkg_name in pkgs_installed:
+        rpms += [_generate_rpm(name=pkg_name, version='2.0', release='3', epoch='1')]
+
+    # Generate NIS facts
+    nis_facts = NISConfig(nis_not_default_conf=pkgs_configured)
+
+    curr_actor_mocked = CurrentActorMocked(msgs=[InstalledRedHatSignedRPM(items=rpms), nis_facts])
+    monkeypatch.setattr(api, 'current_actor', curr_actor_mocked)
+    monkeypatch.setattr(reporting, "create_report", create_report_mocked())
+
+    # Executed actor feeded with out fake msgs
+    nischeck.report_nis()
+
+    # Iterate through installed packages
+    for pkg in pkgs_installed:
+        # Check if package is configured
+        if pkg in pkgs_configured:
+            # Don't waste time checking other conditions
+            # if one of pkgs is already found
+            assert reporting.create_report.called == 1
+            break
+    else:
+        # Assert for no NIS installed and configured packages
+        assert not reporting.create_report.called
--- a/repos/system_upgrade/el8toel9/actors/nisscanner/actor.py
+++ b/repos/system_upgrade/el8toel9/actors/nisscanner/actor.py
@@ -1,0 +1,18 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.nisscan import NISScanLibrary
+from leapp.models import NISConfig
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+
+
+class NISScanner(Actor):
+    """
+    Collect information about the NIS packages configuration.
+    """
+
+    name = 'nis_scanner'
+    consumes = ()
+    produces = (NISConfig,)
+    tags = (FactsPhaseTag, IPUWorkflowTag,)
+
+    def process(self):
+        NISScanLibrary().process()
--- a/repos/system_upgrade/el8toel9/actors/nisscanner/libraries/nisscan.py
+++ b/repos/system_upgrade/el8toel9/actors/nisscanner/libraries/nisscan.py
@@ -1,0 +1,58 @@
+import os
+
+from leapp.libraries.stdlib import api
+from leapp.models import NISConfig
+
+PACKAGES_NAMES = ('ypserv', 'ypbind')
+YPBIND_CONF_FILE = '/etc/yp.conf'
+YPSERV_DIR_PATH = '/var/yp'
+YPSERV_DEFAULT_FILES = ('binding', 'Makefile', 'nicknames')
+
+
+class NISScanLibrary:
+    """
+    Helper library for NISScan actor.
+    """
+
+    def client_has_non_default_configuration(self):
+        """
+        Check for any significant ypbind configuration lines in .conf file.
+        """
+        if not os.path.isfile(YPBIND_CONF_FILE):
+            return False
+
+        # Filter whitespaces and empty lines
+        with open(YPBIND_CONF_FILE) as f:
+            lines = [line.strip() for line in f.readlines() if line.strip()]
+
+        for line in lines:
+            # Cheks for any valid configuration entry
+            if not line.startswith('#'):
+                return True
+        return False
+
+    def server_has_non_default_configuration(self):
+        """
+        Check for any additional (not default) files in ypserv DIR.
+        """
+        if not os.path.isdir(YPSERV_DIR_PATH):
+            return False
+
+        return any(f not in YPSERV_DEFAULT_FILES for f in os.listdir(YPSERV_DIR_PATH))
+
+    def process(self):
+        """
+        Check NIS pkgs configuration for the following options:
+
+        - yp.conf custom configuration
+        - /var/yp not default entry
+        """
+        pkgs = []
+
+        if self.server_has_non_default_configuration():
+            pkgs.append('ypserv')
+
+        if self.client_has_non_default_configuration():
+            pkgs.append('ypbind')
+
+        api.produce(NISConfig(nis_not_default_conf=pkgs))
--- a/repos/system_upgrade/el8toel9/actors/nisscanner/tests/test_nisscan.py
+++ b/repos/system_upgrade/el8toel9/actors/nisscanner/tests/test_nisscan.py
@@ -1,0 +1,82 @@
+import mock
+import pytest
+import six
+
+from leapp.libraries.actor import nisscan
+from leapp.libraries.common.testutils import CurrentActorMocked, produce_mocked
+from leapp.libraries.stdlib import api
+from leapp.models import NISConfig
+
+# Examples of /etc/yp.conf configuration file (default and configured)
+YPBIND_DEFAULT_CONF = """# /etc/yp.conf - ypbind configuration file
+# Valid entries are
+#
+# domain NISDOMAIN server HOSTNAME
+#	Use server HOSTNAME for the domain NISDOMAIN"""
+
+YPBIND_CONFIGURED_CONF = """# /etc/yp.conf - ypbind configuration file
+domain whoisredhat.redhat server prod-db
+
+domain whatisredhat.redhat server prod-db"""
+
+
+@pytest.mark.parametrize('pkgs_installed', [
+    ('ypserv',),
+    ('ypbind',),
+    ('ypserv-not-ypbind',),
+    ('ypbind', 'ypserv'),
+])
+@pytest.mark.parametrize("fill_conf_file", [True, False])
+@pytest.mark.parametrize("fill_ypserv_dir", [True, False])
+def test_actor_nisscan(monkeypatch, pkgs_installed, fill_conf_file, fill_ypserv_dir):
+    """
+    Parametrized helper function for test_actor_* functions.
+
+    Run the actor feeded with our mocked functions and assert
+    produced messages according to set arguments.
+
+    Parameters:
+        pkgs_installed  (touple): installed pkgs
+        fill_conf_file  (bool): not default ypbind config file
+        fill_ypserv_dir  (bool): not default ypserv dir content
+    """
+
+    # Store final list of configured NIS packages
+    configured_pkgs = []
+
+    # Fill ypbind config
+    yp_conf_content = YPBIND_CONFIGURED_CONF if fill_conf_file else YPBIND_DEFAULT_CONF
+
+    # Fill ypserv dir files
+    yp_dir_content = (nisscan.YPSERV_DEFAULT_FILES + ('example.com',) if fill_ypserv_dir
+                      else nisscan.YPSERV_DEFAULT_FILES)
+
+    # Mock 'isfile' & 'isdir' based on installed pkgs
+    mocked_isfile = 'ypbind' in pkgs_installed
+    mocked_isdir = 'ypserv' in pkgs_installed
+
+    mock_config = mock.mock_open(read_data=yp_conf_content)
+    with mock.patch("builtins.open" if six.PY3 else "__builtin__.open", mock_config):
+        curr_actor_mocked = CurrentActorMocked()
+        monkeypatch.setattr(api, 'current_actor', curr_actor_mocked)
+        monkeypatch.setattr(api, "produce", produce_mocked())
+        monkeypatch.setattr(nisscan.os, 'listdir', lambda dummy: yp_dir_content)
+        monkeypatch.setattr(nisscan.os.path, 'isfile', lambda dummy: mocked_isfile)
+        monkeypatch.setattr(nisscan.os.path, 'isdir', lambda dummy: mocked_isdir)
+
+        # Executed actor feeded with mocked functions
+        nisscan.NISScanLibrary().process()
+
+    # Filter NIS pkgs
+    filtered_installed_pkgs = [x for x in pkgs_installed if x in nisscan.PACKAGES_NAMES]
+
+    # Create correct list of pkgs for assert check
+    for pkg in filtered_installed_pkgs:
+        if (pkg == 'ypserv' and fill_ypserv_dir) or (pkg == 'ypbind' and fill_conf_file):
+            configured_pkgs.append(pkg)
+
+    # Sort NISConfig objects
+    nisconf_template = set(NISConfig(nis_not_default_conf=configured_pkgs).nis_not_default_conf)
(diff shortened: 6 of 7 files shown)


repos/system_upgrade/el8toel9/actors/nischeck/libraries/nischeck.py
```diff
@@ -0,0 +1,97 @@
...
+        nis_configured = True if check_server_configuration() else nis_configured
+
+    if installed_packages and nis_configured:
+        report_nis_installed(installed_packages)
+
+def check_client_configuration():
+    """
+    Check for any significant ypbind configuration lines in .conf file.
+    """
+    if not os.path.isfile(YPBIND_CONF_FILE):
```
pirat89, line 78: in the check phase, actors must not communicate with the system. split it into 2 actors (aka scanner, checker) where the scanner gets all required data, propagates them via a msg, and the check actor does additoinal actions.

mkulik-rh:
Changes:

- Added new actor nisscan responsible for scanning NIS configuration files. Produces new NISConfig fact as a result.
- nischeck actor creates report by only checking results from NISConfig and InstalledRedHatSignedRPM
- Cover both actors with proper tests

repos/system_upgrade/el8toel9/actors/nischeck/libraries/nischeck.py
mkulik-rh: New nisscaner actor new provides this information for nischeck.

repos/system_upgrade/el8toel9/actors/nischeck/libraries/nischeck.py
```diff
@@ -0,0 +1,70 @@
...
+        return
+
+    # Create report
+    report_content = [
+        reporting.Title('NIS component has been detected on your system'),
+        reporting.Summary(report_summary),
+        reporting.Severity(reporting.Severity.MEDIUM),
+        reporting.Tags([reporting.Tags.SERVICES]),
+        reporting.ExternalLink(title='RHEL 9 (NIS) discontinuation',
+                                url=report_link_url),
```
fernflower, line 63: formatting a bit off, extra whitespace

repos/system_upgrade/el8toel9/actors/nischeck/tests/test_nischeck.py
```diff
@@ -0,0 +1,82 @@
+import pytest
+import mock
```
fernflower, line 2: This should trigger import order issue.
Could you run make_lint_fix?

repos/system_upgrade/el8toel9/actors/nisscanner/actor.py
```diff
@@ -0,0 +1,17 @@
...
+
+class NISScanner(Actor):
+    """
+    Collect information about the NIS packages configuration.
+    """
+
+    name = 'nis_scanner'
+    consumes = ()
+    produces = (NISConfig, )
+    tags = (FactsPhaseTag, IPUWorkflowTag, )
```
fernflower, line 14: extra whitespace

fernflower:
Aha, I see the workflow didn't run automatically. This explains why the linter didn't complain.
Aproved the workflow run, please fix the linting issues.

fernflower:
Also wondering - is this pr gonna obsolete https://github.com/oamg/leapp-repository/pull/538 ? Maybe we could close https://github.com/oamg/leapp-repository/pull/538 then?

mkulik-rh:
Thanks for review @fernflower. All issues should be fixed now.

#538 seems to check for something different. It also seems to be mixed with something else accidentally. I'll ask and let you know.

repos/system_upgrade/el8toel9/actors/nischeck/libraries/nischeck.py
```diff
@@ -0,0 +1,71 @@
...
+        reporting.Summary(report_summary),
+        reporting.Severity(reporting.Severity.MEDIUM),
+        reporting.Tags([reporting.Tags.SERVICES]),
+        reporting.ExternalLink(title='RHEL 9 (NIS) discontinuation',
+                               url=report_link_url),
+        reporting.Remediation(hint=report_hint),
+    ]
+
+    for package in rpms_configured_installed:
+        reporting.RelatedResource('package', package)
```
fernflower, line 69: I believe you would like to add related resources to the report, so this needs to be changed a bit.

Maybe:
```
related_resources = [reporting.RelatedResource('package', pkg) for pkg in rpms_configured_installed]
report_content += related_resources
```

repos/system_upgrade/el8toel9/actors/nisscanner/libraries/nisscan.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def check_client_configuration(self):
+        """
+        Check for any significant ypbind configuration lines in .conf file.
+        """
+        if not os.path.isfile(YPBIND_CONF_FILE):
+            return False
+
+        # Filter whitespaces and empty lines
+        lines = [line.strip() for line in open(YPBIND_CONF_FILE).readlines() if line.strip()]
```
fernflower, line 25: Please use open as context manager
```
with open(YPBIND_CONF_FILE) as f:
    lines = [line.strip() for line in f.readlines() if line.strip()]
```

repos/system_upgrade/el8toel9/actors/nisscanner/libraries/nisscan.py
```diff
@@ -0,0 +1,60 @@
...
+        return False
+
+    def check_server_configuration(self):
+        """
+        Check for any additional (not default) files in ypserv DIR.
+        """
+        if not os.path.isdir(YPSERV_DIR_PATH):
+            return False
+
+        for f in os.listdir(YPSERV_DIR_PATH):
```
fernflower, line 40: Nit:
L40-43 could be substituted for
```
return any(f not in YPSERV_DEFAULT_FILES for f in os.listdir(YPSERV_DIR_PATH))
```

repos/system_upgrade/el8toel9/actors/nisscanner/libraries/nisscan.py
```diff
@@ -0,0 +1,60 @@
...
+        # Filter whitespaces and empty lines
+        lines = [line.strip() for line in open(YPBIND_CONF_FILE).readlines() if line.strip()]
+
+        for line in lines:
+            # Cheks for any valid configuration entry
+            if not line.startswith('#'):
+                return True
+        return False
+
+    def check_server_configuration(self):
```
fernflower, line 33: check_server_configuration name implies some validation, but iiuc the idea is to check if any additional config files are present. Maybe consider renaming it to something more self-speaking, like `has_non_default_configuration` or `check_non_default_configuration`?

repos/system_upgrade/el8toel9/actors/nisscanner/tests/test_nisscan.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+
+    # Examples of /etc/yp.conf configuration file (default and configured)
+    YPBIND_DEFAULT_CONF = ''.join(("# /etc/yp.conf - ypbind configuration file\n",
+                                   "# Valid entries are\n",
+                                   "#\n",
+                                   "# domain NISDOMAIN server HOSTNAME\n",
+                                   "#	Use server HOSTNAME for the domain NISDOMAIN."))
+
+    YPBIND_CONFIGURED_CONF = ''.join(("# /etc/yp.conf - ypbind configuration file\n",
```
fernflower, line 39: Same here, multi line string can be used

repos/system_upgrade/el8toel9/actors/nisscanner/tests/test_nisscan.py
```diff
@@ -0,0 +1,79 @@
...
+
+    # Fill ypbind config
+    yp_conf_content = YPBIND_CONFIGURED_CONF if fill_conf_file else YPBIND_DEFAULT_CONF
+
+    # Fill ypserv dir files
+    yp_dir_content = (nisscan.YPSERV_DEFAULT_FILES + ('example.com',) if fill_ypserv_dir
+                      else nisscan.YPSERV_DEFAULT_FILES)
+
+    # Mock 'isfile' & 'isdir' based on installed pkgs
+    mocked_isfile = bool('ypbind' in pkgs_installed)
```
fernflower, line 55: Unnecessary bool
`mocked_isfile = 'ypbind' in pkgs_installed`

repos/system_upgrade/el8toel9/actors/nisscanner/tests/test_nisscan.py
```diff
@@ -0,0 +1,79 @@
...
+    mocked_isdir = bool('ypserv' in pkgs_installed)
```
fernflower, line 56: Unnecessary bool

fernflower (changes requested):
Left a couple of questions/suggestions.
Thanks for working on this PR! Could you please also rebase against latest master?

mkulik-rh:
Thanks for comments. Fixed and rebased against current master.

repos/system_upgrade/el8toel9/actors/nischeck/libraries/nischeck.py
```diff
@@ -0,0 +1,71 @@
...
+    model_nis_conf = api.consume(NISConfig)
+    model_rpms = api.consume(InstalledRedHatSignedRPM)
+
+    # Get and filter information from models
+    for model in model_nis_conf:
+        if isinstance(model, NISConfig):
+            configured_rpms = model.nis_not_default_conf
+
+    for model in model_rpms:
+        if not isinstance(model, InstalledRedHatSignedRPM):
```
fernflower, line 44: I'm not sure you really need that check, model_rpms should contain only InstalledRedHatSignedRPMs.
@pirat89 ?

repos/system_upgrade/el8toel9/actors/nischeck/libraries/nischeck.py
```diff
@@ -0,0 +1,71 @@
...
+            continue
+
+        for package in ('ypserv', 'ypbind'):
```
fernflower, line 47: Not sure you need that inside iteration over model_rpms list.. Maybe lines L42-L49 can be substituted for a single
```
installed_packages = [package for package in ('ypserv', 'ypbind') if has_package(InstalledRedHatSignedRPM, package)]
``` 
?

repos/system_upgrade/el8toel9/actors/nisscanner/tests/test_nisscan.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # Filter NIS pkgs
+    filtered_installed_pkgs = [x for x in pkgs_installed if x in nisscan.PACKAGES_NAMES]
+
+    # Create correct list of pkgs for assert check
+    for pkg in filtered_installed_pkgs:
+        if (pkg == 'ypserv' and fill_ypserv_dir) or (pkg == 'ypbind' and fill_conf_file):
+            configured_pkgs.append(pkg)
+
+    assert set(NISConfig(nis_not_default_conf=configured_pkgs).nis_not_default_conf) == \
```
fernflower, line 82: Please avoid using \ as separator for long lines (we don't have it in our guidance for some reason but our linter should surely complain).
Could you please refactor that without \?

repos/system_upgrade/el8toel9/actors/nisscanner/tests/test_nisscan.py
```diff
@@ -0,0 +1,83 @@
...
+
+    Parameters:
+        pkgs_installed  (touple): installed pkgs
+        fill_conf_file  (bool): not default ypbind config file
+        fill_ypserv_dir  (bool): not default ypserv dir content
+    """
+
+    # Examples of /etc/yp.conf configuration file (default and configured)
+    YPBIND_DEFAULT_CONF = """
+    # /etc/yp.conf - ypbind configuration file
```
fernflower, line 34: I believe you should remove whitespaces at the beginning of L34-39 ans L42-46

repos/system_upgrade/el8toel9/actors/nischeck/libraries/nischeck.py
mkulik-rh: Yes, fixed. I tried to merge them at one point and forgot to remove it.

repos/system_upgrade/el8toel9/actors/nisscanner/tests/test_nisscan.py
mkulik-rh: Refactored. I could use set() instead of list in NISConfig but no one seems to use it in a model and 'fields.Set' does not exists.

repos/system_upgrade/el8toel9/actors/nisscanner/tests/test_nisscan.py
mkulik-rh: Yes, sorry. I use join when I need to create multi line string because it is more readable most of the time.

mkulik-rh:
All comments addressed.

mkulik-rh:
@fernflower Hi, any progress with merging this PR ?

fernflower:
@mkulik-rh Could you please rebase against master? I'm putting this pr on top of my todo list.

mkulik-rh:
@fernflower Yes, branch updated with upstream. Thank you.

repos/system_upgrade/el8toel9/actors/nisscanner/libraries/nisscan.py
```diff
@@ -0,0 +1,58 @@
...
+from leapp.libraries.stdlib import api
+from leapp.models import NISConfig
+
+PACKAGES_NAMES = ('ypserv', 'ypbind')
+YPBIND_CONF_FILE = '/etc/yp.conf'
+YPSERV_DIR_PATH = '/var/yp'
+YPSERV_DEFAULT_FILES = ('binding', 'Makefile', 'nicknames')
+
+
+class NISScanLibrary(object):
```
vinzenz, line 12: Nit, starting with Python 3 one writes classes again like this:
```suggestion
class NISScanLibrary:
```

vinzenz: Also looking at this class, I don't see why this is a class at all. It doesn't use any shared variables. These could be just 3 free standing functions.

repos/system_upgrade/el8toel9/actors/nisscanner/libraries/nisscan.py
```diff
@@ -0,0 +1,58 @@
...
+            return False
+
+        # Filter whitespaces and empty lines
+        with open(YPBIND_CONF_FILE) as f:
+            lines = [line.strip() for line in f.readlines() if line.strip()]
+
+        for line in lines:
+            # Cheks for any valid configuration entry
+            if not line.startswith('#'):
+                return True
```
vinzenz, line 31: So a non default configuration contains comments, and there's nothing else that can be changed?
Wouldn't it better to ask rpm -V for the file to check if it is default?

repos/system_upgrade/el8toel9/actors/nischeck/libraries/nischeck.py
```diff
@@ -0,0 +1,64 @@
...
+
+    installed_packages = []
+    configured_rpms = []
+
+    # Get necessary models created by another actors
+    model_nis_conf = api.consume(NISConfig)
+
+    # Get and filter information from models
+    for model in model_nis_conf:
+        configured_rpms = model.nis_not_default_conf
```
MichalHe, line 39: This silently relies on receiving exactly one `NISConfig` message. IMHO, if receiving exactly one message is expected, it should be explicit from the code. 
```suggestion
    nis_confs = api.consume(NISConfig)
    nis_conf = next(nis_confs, None)
    if not nis_conf:
        # Not receiving any NISConfig is highly unlikely, and it is OK to terminate Leapp by raising StopActorExecution informing about no messages received
    if next(nis_confs, None):
        # Log a warning about receiving multiple NIS configs
   configured_rpms = nis_conf.nis_not_default_conf
```

repos/system_upgrade/el8toel9/actors/nisscanner/libraries/nisscan.py
```diff
@@ -0,0 +1,58 @@
+import os
+
+from leapp.libraries.stdlib import api
+from leapp.models import NISConfig
+
+PACKAGES_NAMES = ('ypserv', 'ypbind')
+YPBIND_CONF_FILE = '/etc/yp.conf'
+YPSERV_DIR_PATH = '/var/yp'
+YPSERV_DEFAULT_FILES = ('binding', 'Makefile', 'nicknames')
```
vinzenz, line 9: I mean I do not know anything about `ypserv` but there's a Makefile?

repos/system_upgrade/el8toel9/actors/nischeck/actor.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class NISCheck(Actor):
+    """
+    Checks if any of NIS components is installed and configured
+    on the system and warns users about discontinuation.
+    """
+
+    name = 'nis_check'
+    consumes = (InstalledRedHatSignedRPM, NISConfig,)
```
vinzenz, line 14: Here's an unnecessary trailing comma
```suggestion
    consumes = (InstalledRedHatSignedRPM, NISConfig)
```
If there's more than one entry it is already a tuple, a trailing comma is only needed when there's only one item, to indicate that it is a tuple and not a constrained expression

repos/system_upgrade/el8toel9/actors/nischeck/tests/test_nischeck.py
```diff
@@ -0,0 +1,83 @@
...
+    Returns:
+        rpm  (RPM): new RPM object with name parameter set
+    """
+    return RPM(name=name,
+               version=version,
+               release=release,
+               epoch=epoch,
+               pgpsig='RSA/SHA256, Mon 01 Jan 1970 00:00:00 AM -03, Key ID 199e2f91fd431d51',
+               packager='Red Hat, Inc. <http://bugzilla.redhat.com/bugzilla>',
+               arch='noarch')
```
MichalHe, line 29: A small nitpick: This function can be replaced by `functools.partial`:
```suggestion
import functools
_generate_rpm = functools.partial(RPM,
                                  pgpsig='RSA/SHA256, Mon 01 Jan 1970 00:00:00 AM -03, Key ID 199e2f91fd431d51',
                                  packager='Red Hat, Inc. <http://bugzilla.redhat.com/bugzilla>',
                                  arch='noarch')
```
And then used as follows:
```
_generate_rpm(name='<name>', version='<version>', release='<release>', epoch='<epoch>')
```
Using keyword arguments helps with readability.

repos/system_upgrade/el8toel9/actors/nischeck/tests/test_nischeck.py
```diff
@@ -0,0 +1,83 @@
...
+    nischeck.report_nis()
+
+    # Check for configured NIS package
+    for pkg in pkgs_configured:
+        if pkg in pkgs_installed:
+            assert reporting.create_report.called == 1
+            break
+    else:
+        # Assert for no NIS installed and configured package
+        assert not reporting.create_report.called
```
vinzenz, line 83: TBH this is quite kind confusing to read. 

If I understand the intention is that is only called once if a pkg is installed and configured and not called if not.

```suggestion
    # Check if there any relevant packages installed with a modified configuration
    if set(pkgs_configured) & set(pkgs_installed)):
        # Ensure that it is reported only once
        assert reporting.create_report.called == 1
    else:
        # Otherwise ensure that no report is generated
        assert not reporting.create_report.called    
```

repos/system_upgrade/el8toel9/actors/nischeck/libraries/nischeck.py
```diff
@@ -0,0 +1,64 @@
+from leapp import reporting
+from leapp.libraries.common.rpms import has_package
+from leapp.libraries.stdlib import api
+from leapp.models import InstalledRedHatSignedRPM, NISConfig
+
+report_summary = (
+    'NIS components (ypserv, ypbind, yp-tools) will no longer be available in RHEL-9.'
+    ' Technology behind those packages is based an outdated design patterns no longer'
+    ' considered secure. There is no direct alternative with fully compatible features.'
+)
```
vinzenz, line 10:
```suggestion
report_summary = (
    'The NIS components (ypserv, ypbind, and yp-tools) are no longer available in RHEL-9.'
    ' The technology behind those packages is based an outdated design patterns, which are'
    '  no longer considered as secure. There is no direct alternative with fully compatible'
    ' features.'
)
```

repos/system_upgrade/el8toel9/actors/nischeck/libraries/nischeck.py
```diff
@@ -0,0 +1,64 @@
...
+
+report_hint = (
+    'The alternatives are LDAP and for some use cases Kerberos. Migrating to IPA'
+    ' is also a solution.'
+)
```
vinzenz, line 15:
```suggestion
report_hint = (
    'The alternatives are LDAP and for some use cases Kerberos. An alternative solution, is to migrate to IPA.'
)
```

fernflower:
@mkulik-rh Asked @vinzenz and @MichalHe to take a closer view as I had to switch to another task. Nitpicks are up to you of course.

repos/system_upgrade/el8toel9/actors/nisscanner/libraries/nisscan.py
mkulik-rh: This what was recommend for me to do. Comment from March 12:

> in the check phase, actors must not communicate with the system. split it into 2 actors (aka scanner, checker) where the scanner gets all required data, propagates them via a msg, and the check actor does additoinal actions.

repos/system_upgrade/el8toel9/actors/nisscanner/libraries/nisscan.py
mkulik-rh: Yes, ypserv is an old package. Makefile is used, among other things, to propagate a new maps to slaves.

repos/system_upgrade/el8toel9/actors/nisscanner/libraries/nisscan.py
fernflower: I think what @vinzenz means is stylistic "there is no need for a class here, just use functions".

fernflower: @vinzenz @mkulik-rh I suggest leave it be (leave the object). Our linter needs to be taught to treat different actors (like py3 only for 8to9) differently first, now it applies generic rules to 7to8 and 8to9. 
It is a NIT anyway in my opinion.

mkulik-rh:
`make test_no_lint` stopped working.

```
  File "/usr/lib64/python2.7/pkgutil.py", line 243, in load_module
    mod = imp.load_module(fullname, self.file, self.filename, self.etc)
  File "/home/mkulik/Documents/docker-bind/leapp-repository/repos/system_upgrade/el8toel9/actors/dotnet/actor.py", line 24
    runtime_package = f'dotnet-runtime-{unsupported_version
```

It seems that tests defaults to python2 when we don't use virtual env.
```
# python version to run test with
_PYTHON_VENV=$${PYTHON_VENV:-python2.7}

```

repos/system_upgrade/el8toel9/actors/nisscanner/libraries/nisscan.py
vinzenz: It is, that's the first word in the original comment :-)

repos/system_upgrade/el8toel9/actors/nisscanner/libraries/nisscan.py
vinzenz: In this actor it's ok to do that, the scanner may do that.

repos/system_upgrade/el8toel9/actors/nisscanner/libraries/nisscan.py
vinzenz: Ok thanks for clarifying. It felt strange.

vinzenz:
Well since your code is for EL8 -> EL9 you should run it with python3.6 anyway, your code does not need to run with python 2.7 since it will be executed on EL8 where python3.6 is the "default" (aka platform python).

```
$ PYTHON_VENV=python3.6 make test_no_lint
```
and to speed up your testing:

```
$ export PYTHON_VENV=python3.6
$ # If you didn't already for python3.6
$ make install_deps
$ make test_no_lint ACTOR=nischeck
$ make test_no_lint ACTOR=nisscanner
```

Rezney:
@mkulik-rh will you have time to resume work on this please?

mkulik-rh:
Hi, every time I try to fix and rebase it someone else will try to review the code and spot different things. It would be 5th time that I need to fix it and rebase again.

What is also not really clear to me is inconsistency. Someone suggested some tweaks/changes but when I look at last merged PR (at that time) I can see the same code scheme that has been merged recently.

as an example I can give splitting code to two different actors because of checking content in configuration file. Reason that has been given for it is "in the check phase, actors must not communicate with the system", which seems fine and logical but then I go to recently merged actors like 'checkifcfg', where everything is put into one actor (even one method) and it also checking content in system files.

another small thing, generating fake RPM for tests. I took this idea from 'postgresqlcheck' to keep consistency. Here on 4th rebase someone suggested again to change it to use 'functools' for it which is probably good idea but again, consistency.

another thing: Someone tries to spot small changes but those exact changes caused new errors in make lint check.

another thing: When I tried to test recently fixed code after rebase I got errors but in completely different actor. It is explained above but is does not changes a fact that when this actor was merged with python3 specific code any default references to python2 should be removed.

I often try to use logic/schemes from previously (recently) merged code because I assume that code was checked and it is safe to use. It also allows to keep consistency in the code. Here this logic does not seems to play out well.

pirat89:
@mkulik-rh Hi Marek, thank you for the valid feedback, we appreciate that. We are aware that this is far from ideal state and we are discussing how to improve the process so we are not making this harder for contributors (e.g. assigning 1-2 reviewers reponsible for a PR to limit the ping-pong review process). Let's please focus on finishing the PR this time and hopefully your next experience will be much better in future.

About the splitting the "scan & check actors" that does not happen in couple of actors and we merged them anyway; well the problem was that we were quite close to deadlines and we knew that work had to be merged. So we did some exceptions with the promise the solution will be updated. And yes, it's bad that the update of actors still did not happen.

repos/system_upgrade/el8toel9/actors/nisscanner/libraries/nisscan.py
mkulik-rh: Changed.

repos/system_upgrade/el8toel9/actors/nisscanner/libraries/nisscan.py
mkulik-rh: I don't really like the idea of executing rpm command and I cannot find docs for rpm bindings. I think that my solution is slightly better. Someone can add empty line to config which is meaningless and that would result with outcome:

 "Hey someone change it so it's configured"

if we use rpm to check that file.

repos/system_upgrade/el8toel9/actors/nischeck/libraries/nischeck.py
mkulik-rh: It's also fine and might be more readable for someone. Changed.

repos/system_upgrade/el8toel9/actors/nischeck/actor.py
mkulik-rh: Changed.

repos/system_upgrade/el8toel9/actors/nischeck/tests/test_nischeck.py
mkulik-rh: Changed.

repos/system_upgrade/el8toel9/actors/nischeck/libraries/nischeck.py
mkulik-rh: Changed.

repos/system_upgrade/el8toel9/actors/nischeck/libraries/nischeck.py
mkulik-rh: Changed.

repos/system_upgrade/el8toel9/actors/nischeck/tests/test_nischeck.py
mkulik-rh: That's different logic proposed by you. We are 'faking' two variables here with fixtures. `pkgs_configured` and `pkgs_installed`. pytest will test all possible scenarios so we will get weird combination like: installed ypserv but configured ypbind.

I didn't make special condition for that in test code because there is no need. This is not possible scenario and we don't need to cover it.

What might be a bit confusing is that we start iteration with `pkgs_configured` and not `pkgs_installed`. It actually does not matter but I changed it and added comments.

fernflower (approved):
Thanks for your patience and feedback, we really appreciate it!
LGTM

PR #852, opened 2022-03-09
OpenSSH PermitRootLogin check: avoid false reports

Jakuje:
I thought this was implemented as all the bits were available. But it
turned out that this check was missing and was marked only with TODO for
all these years ...


vinzenz (approved):

vinzenz:
LGTM running tests when ok we will merged it.

PR #853, opened 2022-03-10
Fix VDO actor answer wording

Rezney:


Rezney:
Credit goes to @mportman12

pirat89 (approved):

PR #854, opened 2022-03-10
[IPU 8->9] Detect RPMs with RSA/SHA1 signatures

pirat89:
The SHA-1 hash algorithm is considered as unsafe and it is not longer
allowed to be used on RHEL 9 system using default crypto policies.
One of side effects is problem with installed packages with RSA/SHA1
signatures, as these packages could cause fail of the in-place upgrade
because such packages cannot be read, nor removed by dnf/rpm so the
tests of the transaction already fails in case we want to remove
(include replace) such packages. Additionally, even when we do not
touch such packages during the IPU, the final system will be still
malfunction as rpm will not be able to handle it by default.

Exception are packages signed by RSA/SHA1 keys that are not imported
in the system, so such signatures are not checked and are ignored
like in the case no signature is present at all.

To prevent unwanted situations regarding the installed packages
(packages that will be installed during the upgrade will be handled
in separate commit):
 * inhibit the upgrade when an RSA/SHA1 rpm is installed and
   current crypto policy is not LEGACY or ANY_POL:SHA-1 (e.g.
   DEFAULT:SHA-1).
 * just report presence of such packages with HIGH risk so users
   are informed about the changed policy on RHEL 9

The current crypto policy is obtained from
  /etc/crypto-policies/state/current


repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
```diff
@@ -0,0 +1,62 @@
...
+FMT_LIST_SEPARATOR = '\n    - '
+SUMMARY_FMT = (
+    'The SHA-1 hash algorithm is considered as unsafe and it is not longer'
+    ' allowed to be used on RHEL 9 system using default crypto policies.'
+    ' This causes issues when using DNF/RPM to handle packages with RSA/SHA1'
+    ' signatures as the signature cannot be checked with the default crypto'
+    ' policies. Any such packages cannot be usually installed, nor removed,'
+    ' nor replaced unless the check of signature is disabled in dnf/rpm'
+    ' or SHA-1 is enabled using non-default crypto-policies.'
+    ' The list of problematic packages:{}{}'
```
abadger, line 14:
```suggestion
    'The SHA-1 hash algorithm is considered unsafe and it is no longer'
    ' allowed to be used on RHEL 9 systems using the default crypto policy.'
    ' This causes issues when using DNF/RPM to handle packages with RSA/SHA1'
    ' signatures as the signature cannot be checked with the default crypto'
    ' policy. Any such packages cannot be installed, removed,'
    ' or replaced unless the signature check is disabled in dnf/rpm'
    ' or SHA-1 is enabled using non-default crypto-policies.'
    ' The list of problematic packages:{}{}'
```

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
```diff
@@ -0,0 +1,62 @@
...
+)
+REMEDY_HINT = (
+    'It is recommended to contact vendors of such packages and ask them for'
+    ' new builds with secure signatures and install the new packages before'
+    ' the upgrade. If This is not possible and want to upgrade the system,'
+    ' a) remove such packages b) or update crypto policies using the'
+    ' update-crypto-policies tool prior the upgrade'
+    ' to allow SHA-1 on the upgraded system.'
+    ' The crypto policy must be in such a case "LEGACY" or must be in form'
+    ' "POLICY:SHA-1" (e.g. "{}:SHA-1").'
```
abadger, line 24:
```suggestion
    'It is recommended to contact vendors of such packages to ask them for'
    ' new builds with secure signatures and install the new packages before'
    ' the upgrade. If this is not possible you may instead:'
    ' a) remove such packages or b) update the crypto policy using the'
    ' update-crypto-policies tool to allow SHA-1 usage on the upgraded'
    ' system prior to the upgrade.'
    ' The crypto policy must be "LEGACY" or in form'
    ' "POLICY:SHA-1" (e.g. "{}:SHA-1").'
```

pirat89:
@abadger  thanks!! applied.

abadger:
The code looks good!

I added some minor grammatical suggestions.  Other than those, I wonder if we should include something that tells the user to set the crypto policy back to default after the upgrade?  (Or tells them the pros and cons of leaving it on legacy vs setting it back to default).  If we don't encourage them to set it back after the upgrade, they probably will leave it on legacy which will circumvent one of rhel9's security enhancements.

pirat89:
@abadger I would like to add a link to a RH article that gives people more information about the impacts. But at least some summary we should add in the code as well. Of course, FIPS systems could not use anything like that, as they become non-FIPS when they do it. Their only solution could be to replace/remove the packages prior the upgrade. But let's keep that as good enough for the first release and improvement could be delivered with the next one.

Also I checked that we have to do additional stuff in the container really as I was afraid. It means, we will restrict the possible values just to `DEFAULT:SHA-1` and  `LEGACY` right now so we know we can apply "same" policies inside the container.

repos/system_upgrade/el8toel9/actors/targetuserspacecryptopolicies/libraries/targetuserspacecryptopolicies.py
```diff
@@ -0,0 +1,38 @@
+import os
+
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.common import mounting
+from leapp.libraries.stdlib import api, CalledProcessError
+from leapp.models import TargetUserSpaceInfo, CryptoPolicyInfo
+
+
+def _set_crypto_policy(context, current_policy):
+    if not os.path.exists(context.full_path('/usr/bin/update-crypto-policies')):
```
fernflower, line 10: Could you make this a constant?

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
```diff
@@ -0,0 +1,67 @@
...
+        return True
+    return False
+
+
+def process():
+    # TODO(pstodulk): add link to the official announce of the change in crypto policies
+    bad_rpms = _get_rpms_with_sha1_sig()
+    cpi = next(api.consume(CryptoPolicyInfo), None)
+    if bad_rpms:
+        bad_rpms_str = ''.join(['{}{} ({})'.format(FMT_LIST_SEPARATOR, pkg.name, pkg.pgpsig) for pkg in bad_rpms])
```
fernflower, line 50: Please use named parameters as there are more than 2

repos/system_upgrade/el8toel9/actors/targetuserspacecryptopolicies/libraries/targetuserspacecryptopolicies.py
pirat89: I am not fan of creation of constants that used just once in the whole code. but possibly we could use it for unit-tests.

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
pirat89: done

pirat89:
> Just please run make lint_fix

I did. however I can run it just with Py3 only nowadays on my system. We should add run of tests inside containers

pirat89:
If you want to test scenario with `DEFAULT:SHA1` crypto policy, create
`/usr/share/crypto-policies/policies/modules/SHA1.pmod`:


```
# This subpolicy adds SHA1 hash and signature support

hash = SHA1+

sign = ECDSA-SHA1+ RSA-PSS-SHA1+ RSA-SHA1+

sha1_in_certs = 1
```

abadger (dismissed):

pirat89:
The current implementation covers issue with already installed rpms. I've tested scenarios when we try to install rpms that are signed with RSA/SHA1 signatures - and when not bad rpms are present on the system, so `DEFAULT` crypto policies are set and we have these two scenarios:
* **installing bad RPM when the key is not imported on the system before the run of leapp** - in this case nothing happens because
  * our upgrade dnf plugin is executed with `gpgcheck = False`, so the keys are not imported and so not checked, so the whole IPU passes and no problems are present on the system 
* **installing bad RPM when the key is already imported on the system** - in such a case dnf/rpm checks packages and informs about bad signatures of packages
```
Error: Transaction test error:
  package my-signed-sha1-package-10.0.9-0.202203111513Z.8619719a.deprecated_sha1_alg.el9.noarch does not verify: Header V3 RSA/SHA1 Signature, key ID efc5967f: BAD
```

Under current circumstances, I think this is good enough solution

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
```diff
@@ -0,0 +1,77 @@
...
+)
+
+# TODO(pstodulk): SHA1 sub policy does not exist on RHEL 8 and need to be created manually.
+# Regarding that, we now suggest only the LEGACY policy.
+REMEDY_HINT = (
+    'It is recommended to contact vendors of such packages to ask them for'
+    ' new builds with secure signatures and install the new packages before'
+    ' the upgrade. If this is not possible you may instead:'
+    ' a) remove such packages or b) update the crypto policy using the'
+    ' update-crypto-policies tool to allow SHA-1 usage on the upgraded'
```
Rezney, line 24:
```suggestion
    ' update-crypto-policies tool to allow SHA-1 usage on the '
```

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
```diff
@@ -0,0 +1,77 @@
...
+    ' policy. Any such packages cannot be installed, removed,'
+    ' or replaced unless the signature check is disabled in dnf/rpm'
+    ' or SHA-1 is enabled using non-default crypto-policies.'
+    ' The list of problematic packages:{}'
+)
+
+# TODO(pstodulk): SHA1 sub policy does not exist on RHEL 8 and need to be created manually.
+# Regarding that, we now suggest only the LEGACY policy.
+REMEDY_HINT = (
+    'It is recommended to contact vendors of such packages to ask them for'
```
Rezney, line 20:
```suggestion
    'It is recommended that you contact your package vendor and ask them for'
```

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
```diff
@@ -0,0 +1,77 @@
...
+from leapp.libraries.stdlib import api
+from leapp.models import CryptoPolicyInfo, InstalledRPM, TargetUserSpacePreupgradeTasks
+
+FMT_LIST_SEPARATOR = '\n    - '
+SUMMARY_FMT = (
+    'The SHA-1 hash algorithm is considered unsafe and it is no longer'
+    ' allowed to be used on RHEL 9 systems using the default crypto policy.'
+    ' This causes issues when using DNF/RPM to handle packages with RSA/SHA1'
+    ' signatures as the signature cannot be checked with the default crypto'
+    ' policy. Any such packages cannot be installed, removed,'
```
Rezney, line 11:
```suggestion
    ' policy. Any such package cannot be installed, removed,'
```

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/actor.py
```diff
@@ -0,0 +1,44 @@
...
+# TODO(pstodulk): TargetUserSpacePreupgradeTasks regarding crypto policies should be set
+# in different actor. Let's keep it here for now, but we will generalize this with the next
+# iteration
+
+
+class CheckDeprecatedRPMSignature(Actor):
+    """
+    Check whether any packages signed by RSA/SHA1 are installed
+
+    Crypto policies on RHEL 9 disallows use of the SHA-1 hash algorithm.
```
Rezney, line 15:
```suggestion
    Crypto policies on RHEL 9 disallow use of the SHA-1 hash algorithm.
```

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
```diff
@@ -0,0 +1,77 @@
...
+    # userspace container
+    if current_policy == 'LEGACY' or "DEFAULT:SHA1" in current_policy:
+        return True
+    return False
+
+
+def process():
+    # TODO(pstodulk): add link to the official announce of the change in crypto policies
+    bad_rpms = _get_rpms_with_sha1_sig()
+    cpi = next(api.consume(CryptoPolicyInfo), None)
```
Rezney, line 55: We should not continue when "cpi" is None, cause we would fail when accessing "current_policy later.

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/tests/unit_test_checkdeprecatedrpmsignature.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+
+# from leapp import reporting
```
Rezney, line 3: Please remove the comments when you are ready...

repos/system_upgrade/el8toel9/actors/targetuserspacecryptopolicies/actor.py
```diff
@@ -0,0 +1,32 @@
...
+# be in future generalized for all IPUs. But right now, we have blocking
+# issues just for 8 -> 9.
+
+
+class TargetUserspaceCryptoPolicies(Actor):
+    """
+    Set crypto policies inside the target userspace container.
+
+    The crypto policies inside the container are "DEFAULT" by default, ignoring
+    the setup on the host system. This leads in such situations like when
```
Rezney, line 16:
```suggestion
    the setup on the host system. This leads to situations like when
```

repos/system_upgrade/el8toel9/actors/targetuserspacecryptopolicies/actor.py
```diff
@@ -0,0 +1,32 @@
...
+# TODO(pstodulk): move to common repo in future. This is something what should
+# be in future generalized for all IPUs. But right now, we have blocking
+# issues just for 8 -> 9.
+
+
+class TargetUserspaceCryptoPolicies(Actor):
+    """
+    Set crypto policies inside the target userspace container.
+
+    The crypto policies inside the container are "DEFAULT" by default, ignoring
```
Rezney, line 15:
```suggestion
    The crypto policies inside the container are "DEFAULT" by default and ignoring
```

repos/system_upgrade/el8toel9/actors/targetuserspacecryptopolicies/actor.py
```diff
@@ -0,0 +1,32 @@
...
+    the setup on the host system. This leads in such situations like when
+    we work with rpms differently than the host system after the upgrade.
```
Rezney, line 17: sorry, this does not seem to make sense, could you please rephrase it?

repos/system_upgrade/el8toel9/actors/targetuserspacecryptopolicies/libraries/targetuserspacecryptopolicies.py
```diff
@@ -0,0 +1,56 @@
...
+        context.call(['update-crypto-policies', '--set', current_policy])
+    except CalledProcessError as e:
+        # maybe in this case we could set an inhibitor instead?
+        details = {'details:': str(e), 'stderr': e.stderr}
+
+        # NOTE(pstodulk): In case the LEGACY crypto policy is set, I expect
+        # a different root cause..
+        if current_policy != 'LEGACY':
+            details['hint'] = (
+                'Possibly the custom crypto policies are used on the system.'
```
Rezney, line 30:
```suggestion
                'Custom crypto policies are likely used on the system.'
```

Rezney (changes requested):

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
pirat89: yep, but it exists always.

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/tests/unit_test_checkdeprecatedrpmsignature.py
pirat89: I would like to keep these as it is the first thing people will do when they starte to write tests for this actor.

repos/system_upgrade/el8toel9/actors/targetuserspacecryptopolicies/actor.py
pirat89: `and` does not make sense tbh.

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
pirat89: `such a package` or `any such packages`. In this case I would like to stick with plural.

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
```diff
@@ -0,0 +1,77 @@
...
+        ])
+        report = [
+            reporting.Title('Detected RPMs with RSA/SHA1 signature'),
+            reporting.Summary(SUMMARY_FMT.format(bad_rpms_str)),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Tags([reporting.Tags.SECURITY, reporting.Tags.SANITY]),
+        ]
+        if not _is_sha1_allowed(cpi.current_policy):
+            # TODO:
+            cmds = [['update-crypto-policies', '--set', 'LEGACY']]
```
t184256, line 69: What is the system that's affected with this and will it be reverted?

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
```diff
@@ -0,0 +1,77 @@
...
+# TODO(pstodulk): SHA1 sub policy does not exist on RHEL 8 and need to be created manually.
+# Regarding that, we now suggest only the LEGACY policy.
+REMEDY_HINT = (
+    'It is recommended to contact vendors of such packages to ask them for'
+    ' new builds with secure signatures and install the new packages before'
+    ' the upgrade. If this is not possible you may instead:'
+    ' a) remove such packages or b) update the crypto policy using the'
+    ' update-crypto-policies tool to allow SHA-1 usage on the upgraded'
+    ' system prior to the upgrade.'
+    ' In such a case the crypto policy must be set to "LEGACY".'
```
t184256, line 26:
```suggestion
    ' the upgrade. If this is not possible you may instead'
    ' remove the incompatible packages.'
```

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
```diff
@@ -0,0 +1,77 @@
+from leapp import reporting
+from leapp.libraries.stdlib import api
+from leapp.models import CryptoPolicyInfo, InstalledRPM, TargetUserSpacePreupgradeTasks
+
+FMT_LIST_SEPARATOR = '\n    - '
+SUMMARY_FMT = (
+    'The SHA-1 hash algorithm is considered unsafe and it is no longer'
+    ' allowed to be used on RHEL 9 systems using the default crypto policy.'
+    ' This causes issues when using DNF/RPM to handle packages with RSA/SHA1'
```
t184256, line 9:
```suggestion
    'Digital signatures using SHA-1 hash algorithm are no longer considered'
    ' secure and are not allowed to be used on RHEL 9 systems by default.'
    ' This causes issues when using DNF/RPM to handle packages with RSA/SHA1'
```

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
t184256, line 11:
```suggestion
    ' signatures as the signature cannot be checked with the default'
    ' cryptographic policy. Any such packages cannot be installed, removed,'
```

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
pirat89: @t184256  The host system (the system that is going to be upgraded). Regarding the changed remediation instruction, we will not print this CMD in the report. -- this command is not applied by us automatically. It's information for the user what command they could execute to deal with the issue. so as said, regarding the remediation instruction does not inform about legacy anymore, I will remove the cmd. DONE (with the next push, wait a minute..)

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
pirat89: added (and changed a little bit)

abadger (approved):

Rezney (approved):

PR #856, opened 2022-03-14
Refactor PESEventScanner

MichalHe:
The pes_events_scanner actor is a critical part of the upgrade process, however, its implementation degraded over time due to requirements changing and bugs being discovered to a point that it is hard to debug and to extend this component. This patch rewrites the old implementation using implicit invariants imbued in the old implementation, while trying to make them more explicit. The actor no longer tracks what particular operation should be performed with a package (install/keep/remove) in a manner making sure that the package is only in one of the install/keep/remove buckets. Instead, the actor starts with a set of installed packages and gradually changes this set - removes/adds packages - according to the semantics of events. This shift improves the readability of the code, making it easier to track the effect of an event to a set of packages. The refactor also splits the target package set computation away from event parsing, further improving readability.
    
I have compared the outputs produced on our rhel7 vagrant box and rhel8 vagrant box and they mostly match the old implementation. The only known divergence is that the old implementation will sometimes mark a package to be installed,
even if the package is already installed on a system. This situation happens, e.g., with the `python3-wheel-wheel`. The reason for this behavior is that in the old implementation, the input packages of an event and the output packages of an event are explicitly compared, and only new packages are set to be installed. This comparison is performed in a way that is sensitive to modular information. The new implementation does not care whether some packages of an event are already installed, because adding them to a set does not change the set contents. The packages to be installed are generated at the end based on the difference between source os packages and the computed target os packages.


repos/system_upgrade/common/actors/peseventsscanner/libraries/peseventsscanner.py
```diff
@@ -400,7 +404,7 @@ def parse_entry(entry):
     to_release = parse_release(entry.get('release'))
 
     architectures = entry.get('architectures') or []
-    invalid_archs = tuple(arch for arch in architecture if arch not in architecture.ARCH_ACCEPTED)
```
fernflower: Wow, how did that even work?

repos/system_upgrade/common/actors/peseventsscanner/tests/unit_test_peseventsscanner.py
```diff
@@ -277,6 +294,10 @@ def test_filter_out_pkgs_in_blacklisted_repos(monkeypatch, caplog):
 def test_resolve_conflicting_requests(monkeypatch):
     """
     Verifies that the algorithm correctly resolves conflicting pes events.
+
+    TODO(refactor): This asserts that when two split happening in same release are threated as if they
+                    are performed simultaneously. The packages are setup that such events are applied one after
+                    another the assertion fails.
```
MichalHe, line 300: TODO: Delete this

repos/system_upgrade/common/actors/peseventsscanner/libraries/peseventsscanner.py
```diff
@@ -72,20 +76,25 @@ def past(self):
 
 def pes_events_scanner(pes_json_directory, pes_json_filename):
     """Entrypoint to the library"""
+    # 1) Get input data - in pkgs, transaction configuration, pes events
+    # 2) Filter PES events so that only those relevant for the current IPU remain
+    # 3) Process events so that we know what pkgs to install, remove, keep
+    # 4) Produce output
+
```
MichalHe, line 83: TODO: Delete this comment

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
```diff
@@ -0,0 +1,462 @@
...
+    repoids_of_source_pkgs = {pkg.repository for pkg in source_pkgs}
+
+    # Apply events - compute what packages should the target system have
+    target_pkgs = compute_packages_on_target_system(source_pkgs, events, releases)
+
+    # Packages comming out of the events have PESID as their repository, however, we need real repoid
+    target_pkgs = replace_pesids_with_repoids_in_packages(target_pkgs, repoids_of_source_pkgs)
+
+    # Apply the desired repository blacklisting
+
```
Rezney, line 451: This empty line is redundant I guess...

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
```diff
@@ -0,0 +1,462 @@
...
+                target_pkgs = target_pkgs.union(event.in_pkgs)
+        else:
+            # All other packages have the same semantics - they remove their in_pkgs from the system with given
+            # from_release and add out_pkgs to the system matching to_release
+            are_all_in_pkgs_present = all(in_pkg in source_installed_pkgs for in_pkg in event.in_pkgs)
+            is_any_in_pkg_present = any(in_pkg in source_installed_pkgs for in_pkg in event.in_pkgs)
+
+            # For MERGE to be relevant it is sufficient for only one of its in_pkgs to be installed
+            if are_all_in_pkgs_present or (event.action == Action.MERGED and is_any_in_pkg_present):
+                # In pkgs are present, event can be applied
```
Rezney, line 190:
```suggestion
                # If pkgs are present, event can be applied
```

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
```diff
@@ -0,0 +1,462 @@
...
+    source_pkgs = apply_transaction_configuration(source_pkgs)
+
+    # Keep track of what repoids have the source packages to be able to determine what are the PESIDs of the computed
+    # packages of the target system, so we can distinguish what needs to be repomapped
+    repoids_of_source_pkgs = {pkg.repository for pkg in source_pkgs}
+
+    # Apply events - compute what packages should the target system have
+    target_pkgs = compute_packages_on_target_system(source_pkgs, events, releases)
+
+    # Packages comming out of the events have PESID as their repository, however, we need real repoid
```
Rezney, line 447:
```suggestion
    # Packages coming out of the events have PESID as their repository, however, we need real repoid
```

Rezney (changes requested):
I like the excess of commentary. It helps! Just one minor thing to fix and a few typos. I would like to also run a couple of tests before approving.

Rezney:
Testing results looks fine. @vinzenz could you please also take a look at the PR? Thanks...

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
vinzenz: That's not an If ... then do ..., it's a statement saying that they are present

vinzenz:
@Rezney first pass was OK - I will have a closer look tomorrow

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
MichalHe: @vinzenz is right, its a statement about what does the condition above the comment imply when the branch is taken.

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
```diff
@@ -0,0 +1,462 @@
...
+    'See the report message titled "Excluded target system repositories" '
+    'for details.\nThe list of these packages:'
+)
+
+
+TransactionConfiguration = namedtuple('TransactionConfiguration', ('to_install', 'to_remove', 'to_keep'))
+
+
+def get_cloud_provider_name(cloud_provider_variant):
+    for cloud_provider_prefix in ('aws', 'azure'):
```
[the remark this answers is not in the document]
MichalHe: Fixed.

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
MichalHe: Fixed.

Rezney (commented):
Looks good to me from my side. Let's wait for @vinzenz

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_event_parsing.py
```diff
@@ -0,0 +1,252 @@
...
+    try:
+        all_events = parse_pes_events(fetch.read_or_fetch(pes_json_filename, directory=pes_json_directory,
+                                                          allow_empty=True))
+        arch = api.current_actor().configuration.architecture
+        events_matching_arch = [e for e in all_events if not e.architectures or arch in e.architectures]
+        return events_matching_arch
+    except (ValueError, KeyError):
+        title = 'Missing/Invalid PES data file ({}/{})'.format(pes_json_directory, pes_json_filename)
+        summary = 'Read documentation at: https://access.redhat.com/articles/3664871 for more information ' \
+            'about how to retrieve the files'
```
vinzenz, line 69: I personally prefer this kind of formatting over backslash continuation
```suggestion
        summary = ('Read documentation at: https://access.redhat.com/articles/3664871 for more information '
                   'about how to retrieve the files')
```

Not necessary, but It feels more readable to me

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_event_parsing.py
```diff
@@ -0,0 +1,252 @@
...
+    to (Y, ms=MSY1).
+
+
+    :param mapping: Dictionary mapping input package modulestreams to output modulestreams.
+    :param event: PES event that should be expanded.
+    :returns: A list of generated PES events.
+    """
+    if not mapping:
+        # In case there is no mapping, assume all is going to non modular
+        mapping = {}
```
vinzenz, line 112: I know it's late to ask for this, but please start using a Dict[str, List[str]] instead of Dict[str, str] dictionary - I am fine with a follow up to add it, just please we need to do this, due to changes to PES

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
```diff
@@ -0,0 +1,461 @@
...
+    packages_without_known_repoid = {pkg for pkg in packages_with_pesid if pkg.repository not in pesid_to_repoid_map}
+
+    if packages_without_known_repoid:
+        report_skipped_packages(
+            title='Packages from unknown repositories may not be installed',
+            message='packages may not be installed or upgraded due to repositories unknown to leapp:',
+            skipped_pkgs=packages_without_known_repoid,
+            remediation=(
+                "Please file a bug in http://bugzilla.redhat.com/ for leapp-repository component of "
+                "the Red Hat Enterprise Linux product."
```
vinzenz, line 400: " => '

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_event_parsing.py
MichalHe: Resolved.

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
MichalHe: Resolved.

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
```diff
@@ -0,0 +1,461 @@
...
+        report_skipped_packages(
+            title='Packages available in excluded repositories will not be installed',
+            message=SKIPPED_PKGS_MSG,
+            skipped_pkgs=pkgs_from_blacklisted_repos,
+        )
+    return blacklisted_repoids, target_pkgs.difference(pkgs_from_blacklisted_repos)
+
+
+def get_blacklisted_repoids():
+    """Consume message and return a set of blacklisted repositories"""
```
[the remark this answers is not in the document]
MichalHe: I have realized, that more importantly than the documentation style employed is the fact that these single-line docs are useless. If what the function does can be put on a single line, then it shouldn't be hard to name it in a way that the docs are not needed. Thus I have removed all redundant (=function name tells the same as the docs) single-line docstrings.

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_event_parsing.py
MichalHe: I have changed the code to collect all `to_modulestreams` per `from_modulestream`,  @vinzenz please check that my understanding was correct, as I had to refresh what were the changes in PES about.

vinzenz (approved):

MichalHe:
I've tested for any regressions with the new changes, no regressions found.

PR #860, opened 2022-03-16
Add actor for updating OpenSSH configuration to RHEL9

Jakuje:
The new RHEL9 installation of OpenSSH server provide drop-in directory used by many existing applications (freeipa, crypto-policies, ...) in the OS already and the upgrade needs to make sure the main configuration contains the "Include" directive so the configuration expected by the applications is applied to the sshd server.

Unfortunately, modifying some configuration is not much documented in https://leapp.readthedocs.io/ so I tried to recycle some of the existing code so I would be happy for a review if the changes make sense, they are done correctly, in the right phase, if the testing coverage is enough or if there is anything else to do.


repos/system_upgrade/el8toel9/actors/opensshdropindirectory/actor.py
```diff
@@ -0,0 +1,29 @@
...
+         main sshd_config even in RHEL9 so there is no obvious upgrade path where it
+         could cause issues (unlike the Debian version).
+
+    [1] https://bugzilla.mindrot.org/show_bug.cgi?id=3236
+    """
+
+    name = 'open_ssh_drop_in_directory'
+    consumes = (OpenSshConfig, InstalledRedHatSignedRPM,)
+    produces = ()
+    tags = (IPUWorkflowTag, PreparationPhaseTag,)
```
Rezney, line 26: Do we need to do this before the RPM transaction? If not, please move it into the ApplicationsPhase.

Rezney (changes requested):
I am wondering whether we do not want to also inform customers about the future change to the configuration file in the CheckPhase?
@pirat89 WDYT?

repos/system_upgrade/el8toel9/actors/opensshdropindirectory/actor.py
Jakuje: not needed before RPM transaction. I did not find much documentation about this so I used the first phase I found suitable and late enough.

Note, that if this would be changed in the rhel8, the rhel8 sshd would not be able to start as it does not know the `Include` configuration option, but I hope this phase is late enough.

repos/system_upgrade/el8toel9/actors/opensshdropindirectory/libraries/opensshdropindirectory.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    lines = f.readlines()
+    for line in lines:
+        if line.lstrip().startswith(check_string):
+            # The directive is present
+            return
+
+    # prepend it otherwise, also with comment
+    f.seek(0)
+    f.write(''.join((content, '\n'.join(lines))))
```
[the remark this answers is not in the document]
Jakuje: You are right. I will fix it. I am wondering why the pytest did not capture this as it is clearly covered. Is there some gotcha in the the `assert f.content == expected` can ignore newlines? Is there better way to test this?

Jakuje:
> I am wondering whether we do not want to also inform customers about the future change to the configuration file in the CheckPhase?

I think it is a good idea. How should this report look like? Just low priority non-inhibiting report?

Rezney:
> > I am wondering whether we do not want to also inform customers about the future change to the configuration file in the CheckPhase?
> 
> I think it is a good idea. How should this report look like? Just low priority non-inhibiting report?

Yep, low prio report message please.

repos/system_upgrade/el8toel9/actors/opensshdropindirectorycheck/actor.py
```diff
@@ -0,0 +1,64 @@
...
+    RHEL9 packages are effective.
+
+    This might change the sshd behavior so it is advised to verify by the
+    customer that the updated system behaves as expected.
+    """
+
+    name = 'open_ssh_drop_in_directory_check'
+    consumes = (OpenSshConfig, InstalledRedHatSignedRPM,)
+    produces = (Report,)
+    tags = (IPUWorkflowTag, CheckPhaseTag,)
```
Rezney, line 23:
```suggestion
    tags = (IPUWorkflowTag, ChecksPhaseTag,)
```
It is `Checks`, otherwise it would blow up :)

repos/system_upgrade/el8toel9/actors/opensshdropindirectory/libraries/opensshdropindirectory.py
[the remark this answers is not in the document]
Jakuje: with pytest locally and in your ci. Fixed anyway.

repos/system_upgrade/el8toel9/actors/opensshdropindirectorycheck/actor.py
Jakuje: Changed also in the import line

fernflower:
Could you please rebase? This should fix linter errors

Rezney (approved):

PR #862, opened 2022-03-17
Remove inhibitor flag from checkbind

pemensik:
All detected issues could be fixed by bindupdate. Inhibition is not
required for any action yet, no data would be lost.


pirat89 (approved):
discussed in the chat, it is not supposed to inhibit the upgrade in this case

PR #863, opened 2022-03-18
Openssh sftp

Jakuje:
The first commit overlaps with a first commit from  #860.

Regarding our changes in RHEL9, changing from SCP to SFTP. The default configuration of sshd allows sftp server so the transition should be quite painless. But on upgraded systems, where the sftp subsystem was disabled in sshd config, it might cause existing scripts using scp to start failing. Our proposal is to warn user if there is a chance to hit this.

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/opensshsubsystemsftp/libraries/opensshsubsystemsftp.py
+++ b/repos/system_upgrade/el8toel9/actors/opensshsubsystemsftp/libraries/opensshsubsystemsftp.py
@@ -1,0 +1,47 @@
+from leapp import reporting
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.stdlib import api
+
+
+def process(openssh_messages):
+    config = next(openssh_messages, None)
+    if list(openssh_messages):
+        api.current_logger().warning('Unexpectedly received more than one OpenSshConfig message.')
+    if not config:
+        raise StopActorExecutionError(
+            'Could not check openssh configuration', details={'details': 'No OpenSshConfig facts found.'}
+        )
+
+    # not modified configuration will get updated by RPM automatically
+    if not config.modified:
+        return
+
+    if not config.subsystem_sftp:
+        resources = [
+            reporting.RelatedResource('package', 'openssh-server'),
+            reporting.RelatedResource('file', '/etc/ssh/sshd_config'),
+            reporting.ExternalLink(
+                title="SCP support in RHEL",
+                url="https://access.redhat.com/articles/5284081",
+            ),
+            # TODO provide a link to documentation or blog post
+        ]
+        reporting.create_report([
+            reporting.Title('OpenSSH configured without SFTP subsystem'),
+            reporting.Summary(
+                'The RHEL9 is changing the default SCP behaviour to use SFTP internally '
+                'so not having SFTP server enabled can prevent interoperability and break existing '
+                'scripts on other systems updated to RHEL9 to copy files to or from this machine.'
+            ),
+            reporting.Remediation(
+                hint='Add the following line to the /etc/ssh/sshd_config to enable SFTP server: '
+                     'Subsystem sftp /usr/libexec/openssh/sftp-server'
+            ),
+            reporting.Severity(reporting.Severity.MEDIUM),
+            reporting.Tags([
+                    reporting.Tags.AUTHENTICATION,
+                    reporting.Tags.SECURITY,
+                    reporting.Tags.NETWORK,
+                    reporting.Tags.SERVICES
+            ]),
+        ] + resources)
--- a/repos/system_upgrade/common/actors/opensshconfigscanner/actor.py
+++ b/repos/system_upgrade/common/actors/opensshconfigscanner/actor.py
@@ -15,6 +15,7 @@ class OpenSshConfigScanner(Actor):
      * Protocol
      * Ciphers
      * MACs
+     * Subsystem sftp
 
     """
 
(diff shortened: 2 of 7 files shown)


Rezney (changes requested):

repos/system_upgrade/el8toel9/actors/opensshsubsystemsftp/libraries/opensshsubsystemsftp.py
```diff
@@ -0,0 +1,39 @@
...
+        return
+
+    if not config.subsystem_sftp:
+        resources = [
+            reporting.RelatedResource('package', 'openssh-server'),
+            reporting.RelatedResource('file', '/etc/ssh/sshd_config'),
+            # TODO provide a link to documentation or blog post
+        ]
+        reporting.create_report([
+            reporting.Title('OpenSSH configured without SFTP subsystem'),
```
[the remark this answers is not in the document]
Jakuje: You mean on the SSHD server side? It is enabled by default so I assume there was no need for separate kcs. Add SFTP server is only about adding `Subsystem sftp /usr/share/libexec/sftp-server` into the `sshd_config`, but you are right, its probably worth mentioning it.

repos/system_upgrade/el8toel9/actors/opensshsubsystemsftp/libraries/opensshsubsystemsftp.py
```diff
@@ -0,0 +1,39 @@
...
+            reporting.Summary(
+                'The RHEL9 is changing the default SCP behaviour to use SFTP internally '
+                'so this configuration can prevent interoperability and break existing '
```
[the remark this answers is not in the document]
Jakuje: "this configuration" -> "not having SFTP server enabled"

repos/system_upgrade/el8toel9/actors/opensshsubsystemsftp/libraries/opensshsubsystemsftp.py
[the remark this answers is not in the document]
Jakuje: Found the following article describing the background for SCP deprecation: https://access.redhat.com/articles/5284081 Attached too

Rezney:
@Jakuje could you please rebase against the latest master? Thanks

Jakuje:
> @Jakuje could you please rebase against the latest master? Thanks

Done

> EDIT: also please fix the linting issue if it is still actual

I think it was in the previous master -- it does not point in any code that I would be changing.

repos/system_upgrade/el8toel9/actors/opensshsubsystemsftp/libraries/opensshsubsystemsftp.py
```diff
@@ -0,0 +1,47 @@
...
+        reporting.create_report([
+            reporting.Title('OpenSSH configured without SFTP subsystem'),
+            reporting.Summary(
+                'The RHEL9 is changing the default SCP behaviour to use SFTP internally '
+                'so not having SFTP server enabled can prevent interoperability and break existing '
+                'scripts on other systems updated to RHEL9 to copy files to or from this machine.'
+            ),
+            reporting.Remediation(
+                hint='Add the following line to the /etc/ssh/sshd_config to enable SFTP server: '
+                     'Subsystem sftp /usr/share/libexec/sftp-server'
```
Rezney, line 38: @Jakuje 
Should not this rather be `/usr/libexec/openssh/sftp-server` ?

Jakuje: Indeed. Thanks for checking. Fixed.

Rezney (approved):

repos/system_upgrade/el8toel9/actors/opensshsubsystemsftp/libraries/opensshsubsystemsftp.py
```diff
@@ -0,0 +1,47 @@
...
+
+    if not config.subsystem_sftp:
+        resources = [
+            reporting.RelatedResource('package', 'openssh-server'),
+            reporting.RelatedResource('file', '/etc/ssh/sshd_config'),
+            reporting.ExternalLink(
+                title="SCP support in RHEL",
+                url="https://access.redhat.com/articles/5284081",
+            ),
+            # TODO provide a link to documentation or blog post
```
pirat89, line 27: @Jakuje is it the TODO resolved by the link above?

Jakuje: no, the above is generic information about SCP. I wanted to add something more specific regarding to the deprecation. I will try to find the correct one.

Jakuje: @pirat89 This should be the one I had in my mind when writing the todo: https://www.redhat.com/en/blog/openssh-scp-deprecation-rhel-9-what-you-need-know
Should I add a separate PR or can you update it somehow on your own?

pirat89: @Jakuje  create please the separate PR. thanks

Jakuje: Filled as #893

PR #868, opened 2022-03-29
Enable building leapp-repository for specific chroots only

fernflower:
Instead of building for all project chroots use -r/--chroot
option to specify chroots.
Tmt-tests workflow and copr-build workflow will be passing proper
chroots now via COPR_CHROOT env var.
Also some refactoring has been done to the workflows to move all
magic constants to env section.

OAMG-6735

### DIFF
--- a/.github/workflows/copr-build.yml
+++ b/.github/workflows/copr-build.yml
@@ -32,5 +32,4 @@ jobs:
           EOF
 
           pip install copr-cli
-          make copr_build
-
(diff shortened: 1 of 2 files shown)


fernflower:
Looks viable - https://github.com/fernflower/leapp-repository/runs/5737376736?check_suite_focus=true 
Builds were created only for epel-7 and epel-8 - https://copr.fedorainfracloud.org/coprs/g/oamg/leapp/build/3911059/

pirat89:
ah, I see the comment now. Looks good. merging

pirat89 (approved):

PR #869, opened 2022-03-29
Fix linting violations

fernflower:
used-before-assignment is a nice check worth keeping, so let's
fix commands.upgrade.util code.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/cupsscanner/libraries/cupsscanner.py
+++ b/repos/system_upgrade/el7toel8/actors/cupsscanner/libraries/cupsscanner.py
@@ -93,21 +93,29 @@ def include_directive_check(read_func=_read_file):
     included_files = ['/etc/cups/cupsd.conf']
     error_list = []
 
-    for included_file in included_files:
+    vetted_included_files = []
+    while included_files:
+        # NOTE(ivasilev) Will be using stack to process last encountered include directives first
+        included_file = included_files.pop(-1)
         try:
             lines = read_func(included_file)
         except IOError:
             error_list.append('Error during reading file {}: file not'
                               ' found'.format(included_file))
-            included_files.remove(included_file)
             continue
-
+        # Append to the resulting list of vetted files if exception wasn't raised
+        vetted_included_files.append(included_file)
+        # Mark any other included file you find as need-to-be-validated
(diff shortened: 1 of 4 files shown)


repos/system_upgrade/el7toel8/actors/cupsscanner/libraries/cupsscanner.py
```diff
@@ -93,21 +93,30 @@ def include_directive_check(read_func=_read_file):
     included_files = ['/etc/cups/cupsd.conf']
     error_list = []
 
-    for included_file in included_files:
+    vetted_included_files = []
+    while included_files:
+        # NOTE(ivasilev) Will be using stack to process last encountered include directives first
+        included_file = included_files.pop(-1)
```
vinzenz, line 99:
```suggestion
        for included_file in included_files.copy():
```
?

fernflower: That is not the only place, another violation is in remove part. I felt pretty bad we missed that in original pr so I decided to refactor the implementation completely (as the functionality is decently covered by tests I can be sure my version has the same behaviour). Too much in your opinion? You'd rather leave it be?

vinzenz: I just realized you're right, I didn't get the whole context.

repos/system_upgrade/el7toel8/actors/cupsscanner/libraries/cupsscanner.py
```diff
@@ -93,21 +93,30 @@ def include_directive_check(read_func=_read_file):
...
+        includes_to_process = []
         for line in lines:
             value = get_directive_value('Include', line)
             if value:
-                included_files.append(value)
+                includes_to_process.append(value)
+        # NOTE(ivasilev) Add discovered Include directives to the stack in reversed order, so that they are processed
+        # in the same order they appeared in the file
+        includes_to_process.reverse()
+        included_files.extend(includes_to_process)
```
vinzenz, line 117:
```suggestion
        included_files.extend(reversed(includes_to_process))
```

pirat89 (approved):

PR #871, opened 2022-03-31
Fix the matrix of supported upgrade paths for rhel-alt

pirat89:
RHEL 7.6 is meant just for RHEL-alt and we do not expect to deliver
any changes for it anymore. Hence the reality should reflect the upgrade
is possible to 8.4 only.


pirat89:
As we do not deliver any more content to rhel-alt, no need to apply the fix in downstream

pirat89:
closing. we will do additional changes soon and we should also drop rhel-alt completely. so removing.

PR #874, opened 2022-04-01
pcidevicesscanner: Also match deprecation data against kernel modules

vinzenz:
Previously when the deprecation data got introduced the kernel drivers
reported to be used by lspci have not been checked.
This patch fixes this regression.


pirat89 (approved):
works as expected

pirat89:
ahh... I see that we have still problems with the machine on tft. Let's merge it. i tried that manually and saying sorry to 5+ h running tests + broken infra.

PR #875, opened 2022-04-04
call Satellite installer with --disable-system-checks if possible

evgeni:
The installer has a set of checks to verify whether the current system
is suitable for running Satellite. The administrator of the system can
choose to ignore those checks with `--disable-system-checks`.

As the installer invocation inside LEAPP is non-interactive, we should
err on the side of not running checks, so that the upgrade doesn't abort
in the case where the administrator has chosen to ignore the warnings.

This is in line with other non-interactive invocations of the installer
that other tools (like foreman-maintain) do.

The "if katello installer" logic is needed, as the checks and the cli
parameter is only present in Katello installations, not plain Foreman.

### DIFF
--- a/repos/system_upgrade/el7toel8/models/satellite.py
+++ b/repos/system_upgrade/el7toel8/models/satellite.py
@@ -22,5 +22,7 @@ class SatelliteFacts(Model):
 
     has_foreman = fields.Boolean(default=False)
     """Whether or not foreman is installed on this system"""
+    has_katello_installer = fields.Boolean(default=True)
+    """Whether or not the installer supports Katello additions"""
     postgresql = fields.Model(SatellitePostgresqlFacts)
     """ Foreman related PostgreSQL facts """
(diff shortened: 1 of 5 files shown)


evgeni:
@ehelms would you mind double-checking my reasoning here?

repos/system_upgrade/el7toel8/models/satellite.py
```diff
@@ -22,5 +22,7 @@ class SatelliteFacts(Model):
 
     has_foreman = fields.Boolean(default=False)
     """Whether or not foreman is installed on this system"""
+    has_katello_installer = fields.Boolean(default=True)
```
vinzenz, line 25: Hmm are you sure that you want this by default to be True?

vinzenz: To me it seems like rather something to be default False, but well then again you wouldn't send the message in the first place if you wouldn't have checked for it. 

This is just a question not a request for changes.

evgeni: I have no strong opinion about the default, as all further actions that would take this field into account are guarded by the fact that the `SatelliteFacts` message was sent at all *and* the `has_foreman` field is true. And if that happens, this field has been set explicitly by the `satellite_upgrade_facts` actor to either `True` or `False`.

vinzenz (approved):

ehelms (approved):

PR #876, opened 2022-04-05
make RedHatSignedRpmScanner distribution agnostic

evgeni:
The original detection covered only RHEL system, requiring rpms
to be signed by Red Hat (hardcoded). Also the model
InstalledRedHatSignedRPM didn't provide to much space for detection
of other distros.

The new solution checks RPMs signatures based on the detected
distribution ID (currently: rhel, centos). Fingerprints of GPG keys
and the packager string are stored under
  repos/system_upgrade/common/files/distro/<distro>/signatures.json
where <distro> is the distribution id.

RedHatSignedRPMScanner is deprecated, replaced by DistributionSignedRPM
message. The original RedHatSignedRPMScanner will contain till the
removal just packages signed by RH.

All actors have been updated to consume `DistributionsSignedRPM` msg instead of the deprecated one.

jira: OAMG-9824

----
update:
- rebased against the master and squashed commits.
- resolved conflicts and updated newly created actors to use the new msg
- make pylint happy (at least locally on my machine, waiting for the public test)

TODO:
- [x]  update paths to stored keys based on the voting: `files/distro/<distro>/... `
- [x] update some imports to be multiline again (`fast_lint` usually creates one line imports that are not so nice..) (stretch)
  - pstodulk: skipped: isort is annoying here...
- [x] update commit msg
- [x] passing tests
- [x]  ask for review


repos/system_upgrade/common/actors/redhatsignedrpmscanner/actor.py
```diff
@@ -3,25 +3,39 @@
...
+    """Provide data about installed RPM Packages signed by the vendor.
 
     After filtering the list of installed RPM packages by signature, a message
     with relevant data will be produced.
     """
 
-    name = 'red_hat_signed_rpm_scanner'
+    name = 'vendor_signed_rpm_scanner'
     consumes = (InstalledRPM,)
     produces = (InstalledRedHatSignedRPM, InstalledUnsignedRPM,)
```
evgeni, line 32: I didn't change those (yet), as this would make the diff huge, for no good reason.

repos/system_upgrade/common/actors/redhatsignedrpmscanner/actor.py
```diff
@@ -36,7 +50,7 @@ def process(self):
         ]
 
         def has_rhsig(pkg):
```
ekohl, line 52: Should this be renamed in favor of a vendor agnostic name?

Conan-Kudo:
This seems like a reasonable approach. Would it be possible to make the vendor sigs available through drop-in configuration files? That would make it easier to extend across all the various EL distributions...

evgeni:
You mean by a file in /etc/leapp/files? I guess so.

No idea how complicated the delivery of additional files via the Red Hat portal is, but probably others can weigh in here.

Conan-Kudo:
If it's delivered as an RPM, that shouldn't be an issue either way, right?

evgeni:
For Red Hat systems, those files are not shipped in an RPM but delivered by either the Portal API or a downloadable tarball.
For CentOS/Alma/etc it's an RPM, yeah.

Thinking more about this, we could just have a `vendor.json` and if that's missing fallback to the hard-coded RH values here?

ekohl:
> Thinking more about this, we could just have a `vendor.json` and if that's missing fallback to the hard-coded RH values here?

I wonder if it makes sense to have some `vendor.d/*.json` style. Then it's easy to drop in additional files which could even be done by distros in packaging.

Conan-Kudo:
Yeah, perhaps what we have here could be a fallback if there are no vendor json files.

evgeni:
> > Thinking more about this, we could just have a `vendor.json` and if that's missing fallback to the hard-coded RH values here?
> 
> I wonder if it makes sense to have some `vendor.d/*.json` style. Then it's easy to drop in additional files which could even be done by distros in packaging.

I wouldn't overcomplicate things. Multiple files would require merging and stuff, which can get hairy.

pirat89:
hi @evgeni, basically the RedHatSignedRPMs msgs have to always contain just packages signed by RH - otherwise the name does not make sense. We are thinking about something like `DistributionSignedRPMs` or something like that, where would be packages delivered by the vendor of the particular distribution. We haven't had a time to discuss what will be the wanted solution yet. Having the set of vendors & keys specified in one json is good idea. Currently I do not have an idea for use cases where we would need multiple json files to obtain this information.

ekohl:
> We are thinking about something like `DistributionSignedRPMs` or something like that, where would be packages delivered by the vendor of the particular distribution.

Does it make sense to have `InstalledRedHatSignedRPM` inherit from `DistributionSignedRPMs`?

pirat89:
> Does it make sense to have `InstalledRedHatSignedRPM` inherit from `DistributionSignedRPMs`?

It depends wdyt by that. If you mean something like
```python3
class DistributionSignedRPM(InstalledRPM):
    pass

class InstalledRedHatSignedRPM(DistributionsignedRPM):
    pass
```

but in fact it does not make so much difference when both classes have same fields and I guess the InstalledRedHatSignedRPM should be replaced in future by the new one. But as I told, we will have a discussion about this yet, to decide what is the desired way we would like to go. So it's not carved in the rock the `DistributionSignedRPM` name is the one. Or whether the desired solution will go in this way at all.

ekohl:
That is indeed what I meant, but you're right that it may not make sense to keep `InstalledRedHatSignedRPM` around at all.

pirat89:
```
class DistributionSignedRPM(InstalledRPM):
    pass

@deprecated("since=<date>",  message='Replaced by DistributionSignedRPM')
class InstalledRedHatSignedRPM(InstalledRPM):
    pass
```

pirat89:
@fernflower we want this, but it needs more changes. so postponing that, but keeping it opened.

pirat89:
We know we will not make it for the upcoming release. rolling the stone

evgeni:
Given the overall feedback was "we want this", I went ahead and:
* Renamed `RedHatSignedRpmScanner` to `DistributionSignedRpmScanner`
* Added `DistributionSignedRPM` model and deprecated `InstalledRedHatSignedRPM`
* Refactored the code to use `DistributionSignedRPM` only

Open questions:
* What should happen with `RedHatSignedRpmCheck`? I guess it should be also renamed to `DistributionSignedRpmCheck`?
* How should we (or distributions) deliver the "config" part (keys, packager name)?
    * `/etc/leapp/files/distribution.json`?
    * `/etc/leapp/files/distributions/<release_id>.json`? (my favourite)
    * Something completely else? (Hardcoding as it is today seems wrong in the long run)

Conan-Kudo:
For the open questions:

* I think `DistributionSignedRpmCheck` is a reasonable name for this.
* Can we have the files in `/usr` so that packaging doesn't have to drop the files in `/etc`? Other than that, a `distributions.d/<id>.json` file would work for this.

evgeni:
In the past (when the data was primarily hosted on RH CDN and only downloaded on demand), /etc/leapp/ was the place the toolkit would find it when no CDN connection was available.

I agree /usr would be better, but I'll have to look if there is a way to load it already in a nice way.

Another thing that I realized just now: we marked InstalledRedHatSignedRPM as deprecated, but nothing emits this message anymore, so it's kinda useless. Should we emit both, RH and DistributionSignedRPM for a while?

pirat89:
@evgeni 
> Another thing that I realized just now: we marked InstalledRedHatSignedRPM as deprecated, but nothing emits this message anymore, so it's kinda useless. Should we emit both, InstalledRedHatSignedRPM and DistributionSignedRPM for a while?

definitely. deprecated means, that there is no functional change around it. So it means the msg must be still generated and contain all data. in this case, it means that even when the msg is introduced with the support for other systems, the original one must remain to contain only RH signed packages as before - ignoring other operating systems.

the file with keys should be stored in the /usr as it's not supposed to be modified by user in any way. regarding that,
I suggest either system_upgrades/common/files or private actor ..../files/ directory. the first one seems better to me as it's easier for people to find it. the second one makes sense too as only the one particular actor will need to access it.

evgeni:
> definitely. deprecated means, that there is no functional change around it. So it means the msg must be still generated and contain all data.

done

> in this case, it means that even when the msg is introduced with the support for other systems, the original one must remain to contain only RH signed packages as before - ignoring other operating systems.

not sure how to do it. only add to `rh_signed_pkgs` when `self.configuration.os_release.release_id == 'rhel'`?

> unit tests should keep the cover for the deprecated msg too to be sure the original behaviour has not been modified

I have added unit tests (back) for the RH message, should they also assert the deprecation status? if so, no idea how :D

pirat89:
@evgeni let'S keep it as it is now. I will get to it later to help you to resolve some problems. Yes, our actors should not produce the msg about the use of deprecated stuff - I will handle this part.

> not sure how to do it. only add to rh_signed_pkgs when self.configuration.os_release.release_id == 'rhel'?

most likely we will do something like that. I will check it later.

repos/system_upgrade/common/actors/distributionsignedrpmscanner/tests/test_distributionsignedrpmscanner.py
```diff
@@ -57,12 +60,56 @@ def test_actor_execution_with_signed_unsigned_data(current_actor_context):
...
+            pgpsig='SOME_OTHER_SIG_X'),
+        RPM(name='sample09', version='0.1', release='1.sm01', epoch='1', packager=CENTOS_PACKAGER, arch='noarch',
+            pgpsig='RSA/SHA256, Mon 01 Jan 1970 00:00:00 AM -03, Key ID 45689c882fa658e0')]
+
+    current_actor_context.feed(InstalledRPM(items=installed_rpm))
+    current_actor_context.run(config_model=config)
+    assert current_actor_context.consume(DistributionSignedRPM)
+    assert len(current_actor_context.consume(DistributionSignedRPM)[0].items) == 3
+    assert current_actor_context.consume(InstalledRedHatSignedRPM)
+    # assert len(current_actor_context.consume(InstalledRedHatSignedRPM)[0].items) == 0
```
evgeni, line 108: this assert obviously fails right now, but that's what you wanted to have, right?

pirat89:
@evgeni I am going to start the team discussions today/tomorrow on this. Most likely, we will need to change paths to the json files. I am already aware of other use-cases for the need of various distribution dirs. for this reason I expect we will want to go with something like:
* files/distribution/<distro>/distro-signatures.json
* files/<distro>/distro-signatures.json
* files/..../<distro>-signatures.json
...
as we have already e.g. `files/rpm-gpg/*` etc. where it will be also needed to cover other gpg keys per distribution. So it make sense to me to split distribution specific files under a <distro> dir, where we could have the same hierarchy per each distribution. Instead of having several hierarchies, dealing with each problem on each distro differently.

@evgeni please wait with another changes for now until we have the final decision, as this PR affects many actors, so it will be easier to postpone the rebase until the very end :) I would like to deal with this PR this week. I've already added suppress for the deprecation warning for the run of leapp. (deprecation warning in unit-tests is not so big deal).

@oamg/developers  headsup ^

repos/system_upgrade/common/actors/distributionsignedrpmscanner/actor.py
```diff
@@ -0,0 +1,94 @@
...
+            return any(key in pkg.pgpsig for key in distribution_keys)
+
+        def is_gpg_pubkey(pkg):
+            """
+            Check if gpg-pubkey pkg exists or LEAPP_DEVEL_RPMS_ALL_SIGNED=1
+
+            gpg-pubkey is not signed as it would require another package
+            to verify its signature
+            """
+            return (    # pylint: disable-msg=consider-using-ternary
```
fernflower, line 63: Just curious: why disable linter check here?

fernflower (changes requested):

repos/system_upgrade/common/actors/distributionsignedrpmscanner/actor.py
```diff
@@ -0,0 +1,94 @@
...
+    produces = (DistributionSignedRPM, InstalledRedHatSignedRPM, InstalledUnsignedRPM,)
+    tags = (IPUWorkflowTag, FactsPhaseTag)
+
+    def process(self):
+        # TODO(pstodulk): refactor this function
+        # - move it to the private library
+        # - split it into several functions (so the main function stays small)
+        # FIXME(pstodulk): gpg-pubkey is handled wrong; it's not a real package
+        # and create FP report about unsigned RPMs. Keeping the fix for later.
+        distribution = self.configuration.os_release.release_id
```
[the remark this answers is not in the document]
pirat89: keeping it for a followup PR, as this is the way how the actor was implemented originally. having a few other lines in it, does not make it worse - and moving only this part to the library does not help neither. From my pov, this actor needs to be refactored completely and to reduce amount of changes and make it more clear for review, I kept this as it is. PR with the refactoring will be then more clean.

repos/system_upgrade/common/actors/distributionsignedrpmscanner/actor.py
pirat89: it's because the way it's written. Note this is not a new code, it's original code which is there last 3y. It could be dropped in a followup PR that will refactor this actor.

pirat89: to be clear, I am in favor to write the code in the way the pylint disabling is not needed. I expect to drop it once the code is rewritten.

fernflower: Nah, let's fix it right here and not in a follow up pr please :) We are touching the code already, let's make it right.

repos/system_upgrade/common/actors/distributionsignedrpmscanner/actor.py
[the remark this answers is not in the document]
fernflower: Are we in a rush to get this in?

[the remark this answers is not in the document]
pirat89: We discussed that on a mtg, preferred is to make it early than later. One of reasons why we want to do it now is that we are not in less hectic time (when I do not speak about number of investigations I had to work on in last weeks), and this PR touches 100files right now. Also we have small amount of PR opened - we can expect very different situations when people start to work on RHEL 10. Another reason is that I would create still the downstream build, and I prefer to do it rather tomorrow than on my last wday in this year (in such a case I would prefer to move it already to Jan). Also I am not sure what I will have to work on in upcoming days regarding. I was thinking to refactor it already in a new commit, but I really think it will be better to separate it.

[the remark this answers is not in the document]
pirat89: Of course, if upstream tests are operational again. working on the refactoring now.

repos/system_upgrade/common/actors/distributionsignedrpmscanner/actor.py
pirat89: @fernflower  fixed!

repos/system_upgrade/common/actors/distributionsignedrpmscanner/actor.py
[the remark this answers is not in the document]
pirat89: @fernflower done! On top of that, I've resolved also the fixme about the gpg-pubkey (I couldn't hold my fingers..). I was thinking whether to mark gpg-pubkey at all as signed rpms, as it's not a real rpm, but in the end, as we have it between InstalledRPMs already, it make sense to have them somewhere.

Not sure whether this approach is better, or whether we should keep sorting them between signed and unsigned packages to check for imported keys that are from the distro itself. In the second case, it would need to be fixed the actor with FP report. But somehow I am seeing this approach better. Let me know if you have a different idea about that change.

@oamg/developers the "fixme" note can be discussed widely. Checked with QE, it seems we do not have currently  any tests for that, so it should be ok. Only tests that are related are connected to check of GPG signatures, which is unaffected by this change.

pirat89:
@fernflower  re-review it please after the return from PTO

fernflower (approved):
Can't find anything wrong :) lgtm

Conan-Kudo (approved):

pirat89:
Hmm...I think this old PR needs some love in this moment: "It's a little leap for leapp, but a big leap for community" :yum:

evgeni:
Hah! Thanks a ton for finishing this one up and merging it!

pirat89:
@evgeni thanks for kicking this one out! :)

PR #878, opened 2022-04-07
Ensure the right repositories are enabled on Satellite Capsules

evgeni:


vinzenz:
@evgeni Is satellite only for x86_64? because I just realize that the PRs are only having those x86_64 repositories in that case this needs to be restricted to x86_64, if not, then you need to have for all the supported architectures repositories here.

evgeni:
I've posted https://github.com/oamg/leapp-repository/pull/879 for that, but it shouldn't really be a problem as the packages are available on x86_64 only and then the `has_foreman` step will disable the actors.

vinzenz (approved):

PR #881, opened 2022-04-14
pciscanner: Fix 2 issues in regards to pci address handling

vinzenz:
In a previous patch, the introduction of the new handling of deprecation
data, 2 problems slipped through.

1. The regex replacement for pci ids errornous adds an empty space
   instead of empty string
2. Drivers should be matched on lspci output against the driver
   deprecation data only if the pci_id is empty


pirat89 (approved):
lgtm. I will retest it later.

pirat89:
tested. seems it works as expected.

PR #882, opened 2022-04-25
[IPU 8 -> 9] Migrate blacklisted CAs

pirat89:


pirat89:
@rjrelyea ping

repos/system_upgrade/el8toel9/actors/migrateblacklistca/libraries/migrateblacklistca.py
```diff
@@ -0,0 +1,89 @@
...
+                )
+                failed_files.append(filename)
+        if not failed_files:
+            # the failed removal is not such a big issue here
+            # clean the dir if all files have been migrated successfully
+            shutil.rmtree(dirname, ignore_errors=True)
+    try:
+        run(['/usr/bin/update-ca-trust'])
+    except CalledProcessError as e:
+        api.current_logger().error(
```
Rezney, line 85: Just a comment/question unrelated to this PR: Apart from post-upgrade report, do you think that logs from error level up to critical might go to the breadcrumbs?

rjrelyea (dismissed):
This does the right thing from a ca-certificates perspective.

Rezney (approved):
Looks good to me. Basic scenario where a simple certificate is moved to a correct directory is working.

repos/system_upgrade/el8toel9/actors/migrateblacklistca/libraries/migrateblacklistca.py
pirat89: hmm....good question. tbh, I think that not. As we want to provide post-upgrade reports in future as well so I guess in future we will report start to update many actors to provide such reports to customers. Regarding that, I am not sure how much it could be helpful for us to have these msgs in breadcrumbs. But I would say that at least an indicator that some errors+ logs have been created, could be good. Not sure what state in breadcrumbs is reported now when actors do not fail but create an error logs. @oamg/developers fyi

pirat89:
Regarding my test, the `/usr/share/pki/ca-trust-source/blocklist/` has been migrated successfully but `/etc/pki/ca-trust/blacklisted` has not been touched. IOW, `/etc/pki/ca-trust/blocklisted` has not been touched as well. I am going to investigate this further. Tested the case with symlinks included

pirat89:
merging as this is going to the downstream now and I have problems to find someone who could ack that today again after the last fixes.

PR #883, opened 2022-04-27
Skip comment lines when parsing grub configuration file

pirat89:
Also added unit tests for the valid defautl grub file. Unit tests for the invalid grub file are postponed as we are running out of time. Also, in case of parsing issue the actors stops now with proper exception providing better info to used about the problem - including info about problematic line.

### DIFF
--- a/repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
+++ b/repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
@@ -9,6 +9,7 @@ import re
 import six
 
 from leapp import reporting
+from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.common import repofileutils
 from leapp.libraries.common.config import architecture
 from leapp.libraries.stdlib import api, CalledProcessError, run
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -289,7 +289,9 @@ def _default_grub_info():
         ])
     else:
         for line in run(['cat', default_grb_fpath], split=True)['stdout']:
-            if not line.strip():
+            line = line.strip()
+            if not line or line[0] == '#':
+                # skip comments and empty lines
                 continue
             name, value = tuple(map(type(line).strip, line.split('=', 1)))
```
vinzenz, line 296:
```suggestion
            line = line.strip()
            try:
                 name, value = tuple(map(type(line).strip, line.split('=', 1)))
            except ValueError:
                continue
```

vinzenz, line 296:
```suggestion
            line = line.strip()
            if not line or line[0] == '#':
                # skip comments and empty lines
                continue
            try:
                name, value = tuple(map(type(line).strip, line.split('=', 1)))
            except ValueError:
                continue 
```

pirat89: @vinzenz I do not think this is what we want. From the point that in case we are not able to parse the file, we are not sure it's valid (we hit already issues with invalid default grub file long time ago). From this point, I think we should not allow people to continue. that's why I choosed this solution. But I think we should raise correct error msg to point out what lines are problematic so we can decide whether the problem is on our or customers side when we get the report

pirat89: discussed over irc. we will go with the strict parsing right now. discussion is opened in future

vinzenz (approved):

PR #885, opened 2022-04-28
Drop the obsoleted copr-build job

pirat89:
This job is already obsoleted as nowadays the builds are delivered
by Packit.

Also this is only failing recent month. So it seems there is no need
for this. (look at master branch and history of commits since 7th April)


fernflower (dismissed):

vinzenz (approved):

vinzenz:
Please rebase - Because it packit never reported a status back - Kind of blocks the whole thing.

vinzenz (approved):

PR #888, opened 2022-05-10
CheckNFS actor should respect nfsd filesystem

fernflower:
Check filesystem type for full match with 'nfs'
otherwise false positive fires like in 'nfsd' may
occur.

OAMG-6355


vinzenz (approved):

vinzenz (changes requested):
Please add a test case for the nfsd case - Thanks.

Rezney (approved):

PR #889, opened 2022-05-13
Remove temporary leapp directory in /root

Rezney:
Leaving the directory behind will cause issues during upgrades to
another major version. E.g in case of RHEL 7 > RHEL 8 > RHEL 9
upgrades.

### DIFF
--- a/repos/system_upgrade/common/actors/preparepythonworkround/libraries/workaround.py
+++ b/repos/system_upgrade/common/actors/preparepythonworkround/libraries/workaround.py
@@ -1,7 +1,9 @@
 import os
+import shutil
 import sys
 
 from leapp.libraries.common.utils import makedirs
+from leapp.libraries.stdlib import api
 
 LEAPP_HOME = '/root/tmp_leapp_py3'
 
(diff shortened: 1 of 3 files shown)


repos/system_upgrade/common/actors/preparepythonworkround/libraries/workaround.py
```diff
@@ -18,6 +19,8 @@ def _get_orig_leapp_path():
 
 def apply_python3_workaround():
     py3_leapp = os.path.join(LEAPP_HOME, 'leapp3')
+    if os.path.exists(LEAPP_HOME):
+        shutil.rmtree(LEAPP_HOME, ignore_errors=True)
```
pirat89, line 23: hmm...it's a nitpick. but...I know we do not expect errors during the removal. I would add there a note or something, that in case we ever see an issue happening here, we should create a separate actor to handle it before the rpm transaction is performed. and possibly remove the "ignore_errors" in such a case. I mean, in case this will crash, most likely we will see crash when we try to run leapp after the rpm transaction...

repos/system_upgrade/common/actors/preparepythonworkround/libraries/workaround.py
```diff
@@ -18,6 +19,8 @@ def _get_orig_leapp_path():
...
     makedirs(LEAPP_HOME)
     leapp_lib_symlink_path = os.path.join(LEAPP_HOME, 'leapp')
     if not os.path.exists(leapp_lib_symlink_path):
```
pirat89, line 26: probably we could drop the condition when we are already removing the whole directory in the step before

Rezney: dropped

repos/system_upgrade/common/actors/preparepythonworkround/libraries/workaround.py
Rezney: added a logger

vinzenz (dismissed):

pirat89 (approved):

PR #892, opened 2022-05-18
Drop the checkcpu actor from the el8toel9 repo

pirat89:
The actor has been supposed to be removed a time ago already
as this functionality is handled in generic way by
    common/actors/checkdetecteddevicesanddrivers
As this actor still exists it breaks the expectation the CPU check
is driven by the `device_driver_deprecation_data.json` file. Regarding
that, we cannot dynamically react to changes around supported CPUs
on s390x (IBM Z) architecture for IPU 8 -> 9.

This relates to BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2087664
The bz is reported for IPU 7 -> 8, however the problem is on IPU 8 -> 9
as well and we cannot handle via data files due to this actor.
So dropping it.


pirat89:
@vinzenz the failed unit-tests could be ignored. It's caused by isort that's trying to check content in removed files :)

vinzenz (approved):

vinzenz (approved):

PR #894, opened 2022-05-19
Add upgrade path 8.7 -> 9.0

pirat89:
The next supported upgrade path for IPU 8 -> 9 is 8.7 -> 9.0.
As the development and testing are already in progress, enable
8.7 system for the upgrade for non-sap systems.

With that, we expect in future to drop 8.6 -> 9.0, but keeping it
for now as we want to be able to test the functionality with 8.6
as well in this phase.

Depends-On: #891


pirat89:
@oamg/developers  it's expected tests will crash until #891 is merged and this PR is rebased on top of it

Rezney (approved):

PR #895, opened 2022-05-19
Fix satellite actor due to some oversight of a missing parameter

vinzenz:


fernflower (approved):
Thanks a lot for a prompt fix!

PR #898, opened 2022-05-27
Handle custom crypto policies during upgrade

Jakuje:
Handle corner cases of upgrade of crypto policies:
 * warn if non-default crypto policy is used
 * prepare container and target system to have expected custom crypto policies files and modules

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/cryptopoliciescheck/libraries/cryptopoliciescheck.py
+++ b/repos/system_upgrade/el8toel9/actors/cryptopoliciescheck/libraries/cryptopoliciescheck.py
@@ -1,0 +1,50 @@
+from leapp import reporting
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.stdlib import api
+from leapp.models import CopyFile, TargetUserSpacePreupgradeTasks
+
+
+def _get_files_to_copy(cpi):
+    return [f.path for f in cpi.custom_policies + cpi.custom_modules]
+
+
+def process(cpi_messages):
+    cpi = next(cpi_messages, None)
+    if list(cpi_messages):
+        api.current_logger().warning('Unexpectedly received more than one CryptoPolicyInfo message.')
+    if not cpi:
+        raise StopActorExecutionError(
+            'Could not check crypto policies status',
+            details={'details': 'No CryptoPolicyInfo facts found.'}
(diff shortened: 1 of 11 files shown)


Rezney (changes requested):

Rezney:
Apart from the submitted, looks good to me. @pirat89 I recall you had some discussions about this PR, is everything clarified?

Jakuje:
Thank you for review. The wording was fixed in last commit. The CI failure does not look like related to my changes.

t184256:
I can't find anything wrong with the code.

Can't say I understand why we're separating subpolicies from policies and stripping paths to names when we could just `glob.glob()` 4 locations, copy the results over and thus only deal in absolute filepaths, but it should work that way anyway.

pirat89:
@Jakuje rebase against up-to-date master fixes CI

repos/system_upgrade/el8toel9/actors/cryptopoliciescheck/libraries/cryptopoliciescheck.py
```diff
@@ -0,0 +1,48 @@
...
+                ).format(cpi.current_policy)),
+            reporting.Severity(reporting.Severity.MEDIUM),
+            reporting.Tags([reporting.Tags.SECURITY, reporting.Tags.SANITY]),
+            reporting.Remediation(hint="Review the current policy in /etc/crypto-policies/state/CURRENT.pol"),
+            reporting.RelatedResource('package', 'crypto-policies'),
+            reporting.ExternalLink(
+                url='https://red.ht/rhel-9-security-considerations',
+                title='Security Considerations in adopting RHEL 9'
+            )
+        ])
```
pirat89, line 48: @Jakuje I think the TODO above is mentioned just for this report. Regarding the text of the report I think it make sense to do it
```python

....
        if cpi.current_policy in ('FIPS', 'FUTURE'):
            return
        reporting.create_report([ 
```

Jakuje: Fixed.

pirat89 (approved):
lgtm and the solution has been tested by @t184256 . Starting regressions tests

Rezney (approved):

PR #899, opened 2022-05-30
BZ#2089887 - enable the Satellite module when uprading

evgeni:
Also enable the Capsule module when upgrading capsules.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/satellite_upgrade_facts/tests/unit_test_satellite_upgrade_facts.py
+++ b/repos/system_upgrade/el7toel8/actors/satellite_upgrade_facts/tests/unit_test_satellite_upgrade_facts.py
@@ -25,6 +25,8 @@ FOREMAN_PROXY_RPM = fake_package('foreman-proxy')
 KATELLO_INSTALLER_RPM = fake_package('foreman-installer-katello')
 KATELLO_RPM = fake_package('katello')
 POSTGRESQL_RPM = fake_package('rh-postgresql12-postgresql-server')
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/el7toel8/actors/satellite_upgrade_facts/tests/unit_test_satellite_upgrade_facts.py
```diff
@@ -84,6 +86,20 @@ def test_enables_pki_modules(current_actor_context):
     assert Module(name='pki-deps', stream='10.6') in message.modules_to_enable
 
 
+def test_enables_satellite_module(current_actor_context):
+    current_actor_context.feed(InstalledRPM(items=[FOREMAN_RPM, SATELLITE_RPM]))
+    current_actor_context.run(config_model=mock_configs.CONFIG)
+    message = current_actor_context.consume(RpmTransactionTasks)[0]
+    assert Module(name='satellite', stream='el8') in message.modules_to_enable
```
vinzenz, line 93: I mean since they are mutually exclusive as per code at least, maybe you'd want to check the negative as well
```suggestion
    assert Module(name='satellite', stream='el8') in message.modules_to_enable
    assert Module(name='satellite-capsule', stream='el8') not in message.modules_to_enable
```

same just swapped for `test_enables_satellite_capsule_module`

evgeni: good point! updated

Rezney (approved):

PR #905, opened 2022-06-13
Use library function is_inhibitor to check for failures

fernflower:
This patch has to be merged together with leapp's PR677.

OAMG-1429


pirat89 (changes requested):
Needs the bump of the leapp-framework in the specfile.

fernflower:
Thanks to @MichalHe who helped pinpoint the problem with check_yum_plugins_enabled unit test. Turned out we have a flaw in implementation of create_report_mocked, which caused all BaseListPrimitive fields to be processed incorrectly.
This should be fixing the issue https://github.com/oamg/leapp-repository/pull/905/commits/4064daea230d2b6861654871417fe6d602b46b17

pirat89 (approved):
tested. the inhibitor works as expected again.

PR #908, opened 2022-06-14
storage: check for /var mounted with the noexec option

MichalHe:
When /var is mounted with noexec option leapp fails as it cannot execute
binaries from containers. This patch adds a generic actor checking for
mounting options, currently only checking for /var being mounted with
noexec.

JIRA ref: OAMG-3248
BZ: 1836179

### DIFF
--- a/repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
+++ b/repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
@@ -1,0 +1,75 @@
+from leapp import reporting
+from leapp.libraries.stdlib import api
+from leapp.models import StorageInfo
+
+
+def inhibit_upgrade_due_var_with_noexec(mountpoint, found_in_fstab=False):
+    summary = (
+        'Leapp detected that the {0} mountpoint is mounted with the "noexec" option, '
+        'which prevents binaries necessary for the upgrade from being executed. '
+        'The upgrade process cannot continue with {0} mounted using the "noexec" option.'
+    )
+
+    if found_in_fstab:
+        hint = (
+            'Temporarily remove the "noexec" option from {0} entry in /etc/fstab until the system is upgraded, '
+            'and remount the partition without the "noexec" option.'
+        )
+        related_resource = [reporting.RelatedResource('file', '/etc/fstab')]
+    else:
+        hint = (
+            'Remount {0} without the noexec option and make sure the change is persistent'
+            'during the entire in-place upgrade process.'
+        )
+        related_resource = []
+
+    reporting.create_report([
+        reporting.Title(
+            'Detected partitions mounted with the "noexec" option, preventing a successful in-place upgrade.'
+        ),
+        reporting.Summary(summary.format(mountpoint)),
+        reporting.Remediation(hint=hint.format(mountpoint)),
+        reporting.RelatedResource('file', '/etc/fstab'),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Tags([reporting.Tags.FILESYSTEM]),
+        reporting.Flags([reporting.Flags.INHIBITOR]),
+    ] + related_resource)
+
+
+def find_mount_entry_with_mountpoint(mount_entries, mountpoint):
+    for mount_entry in mount_entries:
+        if mount_entry.mount == mountpoint:
+            return mount_entry
+    return None
+
+
+def find_fstab_entry_with_mountpoint(fstab_entries, mountpoint):
+    for fstab_entry in fstab_entries:
+        if fstab_entry.fs_file == mountpoint:
+            return fstab_entry
+    return None
+
+
+def check_noexec_on_var(storage_info):
+    """Check for /var or /var/lib being mounted with noexec mount option."""
+
+    # Order of checking is important as mount options on /var/lib override those on /var
+    mountpoints_to_check = ('/var/lib/leapp', '/var/lib', '/var')
+    for mountpoint in mountpoints_to_check:
+        fstab_entry = find_fstab_entry_with_mountpoint(storage_info.fstab, mountpoint)
+        if fstab_entry and 'noexec' in fstab_entry.fs_mntops.split(','):
+            inhibit_upgrade_due_var_with_noexec(fstab_entry.fs_file, found_in_fstab=True)
+            return  # Do not check further as present mounts would likely reflect fstab
+
+    # Make sure present mountpoints don't contain noexec as well - user might have fixed noexec in fstab
+    # but did not remount the partition, or, less likely, mounted the partition without creating a fstab entry
+    for mountpoint in mountpoints_to_check:
+        mount_entry = find_mount_entry_with_mountpoint(storage_info.mount, mountpoint)
+        if mount_entry and 'noexec' in mount_entry.options.split(','):
+            inhibit_upgrade_due_var_with_noexec(mount_entry.mount, found_in_fstab=False)
+            return
+
+
+def check_mount_options():
+    for storage_info in api.consume(StorageInfo):
+        check_noexec_on_var(storage_info)
--- a/repos/system_upgrade/common/actors/checkmountoptions/actor.py
+++ b/repos/system_upgrade/common/actors/checkmountoptions/actor.py
@@ -1,0 +1,21 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.checkmountoptions import check_mount_options
+from leapp.models import StorageInfo
+from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckMountOptions(Actor):
+    """
+    Check for mount options preventing the upgrade.
+
+    Checks performed:
+        - /var is mounted with the noexec option
+    """
+    name = "check_mount_options"
+    consumes = (StorageInfo,)
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag,)
+
+    def process(self):
+        check_mount_options()
--- a/repos/system_upgrade/common/actors/checkmountoptions/tests/test_checkmountoptions.py
+++ b/repos/system_upgrade/common/actors/checkmountoptions/tests/test_checkmountoptions.py
@@ -1,0 +1,61 @@
+from collections import namedtuple
+from functools import partial
+
+import pytest
+
+from leapp import reporting
+from leapp.libraries.actor.checkmountoptions import check_mount_options
+from leapp.libraries.common.testutils import create_report_mocked, CurrentActorMocked
+from leapp.libraries.stdlib import api
+from leapp.models import FstabEntry, MountEntry, StorageInfo
+
+
+@pytest.mark.parametrize(
+    ('fstab_entries', 'mounts', 'should_inhibit'),
+    [
+        (
+            (('/var', 'default'), ),
+            (('/var', 'default'), ),
+            False
+        ),
+        (
+            (('/var', 'default'), ('/var/lib', 'default'), ),
+            (('/var', 'default'), ('/var/lib', 'default'), ),
+            False
+        ),
+        (
+            (('/var', 'default'), ('/var/lib/leapp', 'noexec')),
+            (('/var', 'default'), ('/var/lib/leapp', 'noexec')),
+            True
+        ),
+        (
+            (('/var', 'defaults'), ('/var/lib', 'noexec')),
+            (('/var', 'noexec'), ('/var/lib', 'noexec')),
+            True
+        ),
+        (
+            (('/var', 'noexec'), ('/var/lib', 'defaults')),
+            (('/var', 'noexec'), ('/var/lib', 'noexec')),
+            True
+        ),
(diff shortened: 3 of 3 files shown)


repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def check_noexec_on_var(storage_info):
+    """Check for /var being mounted with the noexec option."""
+
+    for mount_entry in storage_info.mount:
+        if mount_entry.mount == '/var' and 'noexec' in mount_entry.options.split(','):
+            reporting.create_report([
+                reporting.Title(
+                    'Mount options used to mount /var are preventing the upgrade'
```
Rezney, line 13:
```suggestion
                   'The /var is mounted with the "noexec option". Inhibiting the upgrade...'
```

repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
```diff
@@ -0,0 +1,27 @@
...
+                ),
+                reporting.Summary(
+                    'The /var is mounted with the noexec option, preventing the upgrade.'
```
Rezney, line 16:
```suggestion
                    'Leapp detected the /var filesystem is mounted with the "noexec" option which prevents binaries from being executed. The upgrade process cannot continue with this option being used.
```

repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
```diff
@@ -0,0 +1,27 @@
...
+                ),
+                reporting.Severity(reporting.Severity.HIGH),
+                reporting.Tags([reporting.Flags.FILESYSTEM]),
+                reporting.Flags([reporting.Flags.INHIBITOR]),
```
Rezney, line 20: Please add a "Remediation hint".

Rezney (changes requested):

repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
```diff
@@ -0,0 +1,27 @@
...
+        if mount_entry.mount == '/var' and 'noexec' in mount_entry.options.split(','):
+            reporting.create_report([
+                reporting.Title(
+                    'Mount options used to mount /var are preventing the upgrade'
+                ),
+                reporting.Summary(
+                    'The /var is mounted with the noexec option, preventing the upgrade.'
+                ),
+                reporting.Severity(reporting.Severity.HIGH),
+                reporting.Tags([reporting.Flags.FILESYSTEM]),
```
Rezney, line 19: `AttributeError: type object 'Flags' has no attribute 'FILESYSTEM'`

```suggestion
                reporting.Tags([reporting.Tags.FILESYSTEM]),
```

pirat89 (changes requested):

repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
```diff
@@ -0,0 +1,27 @@
+from leapp import reporting
+from leapp.libraries.stdlib import api
+from leapp.models import StorageInfo
+
+
+def check_noexec_on_var(storage_info):
+    """Check for /var being mounted with the noexec option."""
+
+    for mount_entry in storage_info.mount:
+        if mount_entry.mount == '/var' and 'noexec' in mount_entry.options.split(','):
```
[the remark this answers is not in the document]
MichalHe: I've included only `/var` as I did not think there would be system with that level of granularity when it comes to what is being mounted to `/var`. I will change the code to include `/var/lib` as well, however, I do not think someone would have a separate mount for _leapp_, therefore, I will left `/var/lib/leapp` out.

btw, thanks for reviewing :smile:

repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
MichalHe: Thanks, fixed.

repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
MichalHe: Remediation hint added.

repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
MichalHe: Summary changed.

repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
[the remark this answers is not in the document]
pirat89: @MichalHe actually, we know that `/var/lib/leapp` is happening to deal with problem of extra needed space when multiple XFS without fstype attributes are present / mounted. However as written above, I do not expect that people set `noexec` attirubte for that mountpoint :) if so, we will know about that in future :D currently if think that `/var` and `/var/lib` is good enough

repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
```diff
@@ -0,0 +1,77 @@
...
+            'and remount the partition without the "noexec" option.'
+        )
+        related_resource = [reporting.RelatedResource('file', '/etc/fstab')]
+    else:
+        hint = 'Remount {0} without the noexec option.'
+        related_resource = []
+
+    reporting.create_report([
+        reporting.Title(
+            'The {0} mountpoint is mounted with the "noexec" option.'.format(mountpoint)
```
pirat89, line 25: @MichalHe reading the solution, I prefer to make it general s we do not have different "reports" for /var & /var/lib. Let's keep the title really static and make just dynamic other data (summary, remediation, ...). So the title could be e.g.
```suggestion
            'The partitions mounted with the "noexec" option preventing the correct in-place upgrade..'.format(mountpoint)
```

the function could accept the list of mountpoints instead of one.

repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
```diff
@@ -0,0 +1,77 @@
...
+    var_lib_fstab_entry = find_fstab_entry_with_mountpoint(storage_info.fstab, '/var/lib')
+    if var_lib_fstab_entry:
+        if 'noexec' in var_lib_fstab_entry.fs_mntops.split(','):
+            inhibit_upgrade_due_var_with_noexec(var_lib_fstab_entry.fs_file, found_in_fstab=True)
+            return  # Do not check further as present mounts would likely reflect fstab
+    else:
+        var_fstab_entry = find_fstab_entry_with_mountpoint(storage_info.fstab, '/var')
+        if var_fstab_entry and 'noexec' in var_fstab_entry.fs_mntops.split(','):
+            inhibit_upgrade_due_var_with_noexec(var_fstab_entry.fs_file, found_in_fstab=True)
+            return  # Do not check further as present mounts would likely reflect fstab
```
pirat89, line 62: See my comment above. Regarding that, the solution here could look like this (pseudocode):
```python
    noexec_fstab_entries = []
    for mountpoint in ['/var', '/var/lib']:
        fstab_entry = find_fstab_entry_with_mountpoint(storage_info.fstab, '/var/lib')
        if fstab_entry and 'noexec' in fstab_entry.fs_mntops.split(','):
            noexec_fstab_entries.append(fstab_entry)
    if noexec_fstab_entries:
        inhibit_upgrade_due_var_with_noexec(noexec_fstab_entries, found_in_fstab=True)
        return  # Do not check further as present mounts would likely reflect fstab
```
similar for the block below.

repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+    if found_in_fstab:
+        hint = (
+            'Temporarily remove the "noexec" option from {0} entry in /etc/fstab until the system is upgraded, '
+            'and remount the partition without the "noexec" option.'
+        )
+        related_resource = [reporting.RelatedResource('file', '/etc/fstab')]
+    else:
+        hint = 'Remount {0} without the noexec option.'
```
pirat89, line 20: @MichalHe  thinking about that, in case we do not discover the noexec in /etc/fstab but the mountpoint is noexec, I am curious how this happen. Regarding upgrades, all mountpoins anywhere on the path `/var/lib/leapp` needs to be in fstab when doing the upgrade (otherwise additional issues are hit during the upgrade) and when we find different options than in fstab, it's very suspicious. Probably the msg should be extended e.g. like: `Remount mentioned partitions without the noexec option and ensure this change is persistent during whole in-place upgrade process.`


Btw, if we discover a mountpoint anywhere between `/` and `/var/lib/leapp` and it's not in fstab (we already investigated such a problem regarding /var/lib/leapp mountpoint), we should inhibit the upgrade as well. But I suggest to keep such an extension for a separate PR.

pirat89 (commented):
Thinking about the solution I found couple of things that could be probably changed:
* current solution could raise inhibitors several times and each time just one; consider this scenario:
  * run leapp preupgrade -> hit inhibitor /var/lib -> fix /var/lib -> rerun leapp -> hit /var
  * adressing the same issue by different reports (different title -> different key); the same problem should be adressed by the same report (with the same key)

repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
pirat89: btw, in case it's implemented like that, `/var/lib/leapp` could be added as well for free.

MichalHe: The code was written in a specific manner to account for mount option "shadowing" - for example:
- `/var` is mounted with `noexec`
- `/var/lib` is mounted *without* noexec
presents no problem for leapp (I've tried the described scenario), however, the suggested change would mark it as an inhibitor for the upgrade.

pirat89: you are right!! Important is just the last mountpoint on the path. Where I left my head? :-D Is it monday?

repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
```diff
@@ -0,0 +1,75 @@
...
+
+    # Order of checking is important as mount options on /var/lib override those on /var
+    mountpoints_to_check = ('/var/lib/leapp', '/var/lib', '/var')
+    for mountpoint in mountpoints_to_check:
+        fstab_entry = find_fstab_entry_with_mountpoint(storage_info.fstab, mountpoint)
+        if fstab_entry and 'noexec' in fstab_entry.fs_mntops.split(','):
+            inhibit_upgrade_due_var_with_noexec(fstab_entry.fs_file, found_in_fstab=True)
+            return  # Do not check further as present mounts would likely reflect fstab
+
+    # Make sure present mounts doesn't contain noexec as wells - likely user mounted a partition
```
pirat89, line 64: nitpitck:
```suggestion
    # Make sure present mountpointss don't contain noexec as well - likely user mounted a partition
```

repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
```diff
@@ -0,0 +1,75 @@
...
+    # and without adding it to fstab
+    for mountpoint in mountpoints_to_check:
+        mount_entry = find_mount_entry_with_mountpoint(storage_info.mount, mountpoint)
+        if mount_entry and 'noexec' in mount_entry.options.split(','):
+            inhibit_upgrade_due_var_with_noexec(mount_entry.mount, found_in_fstab=True)
```
pirat89, line 69: it should be `False` here
```suggestion
            inhibit_upgrade_due_var_with_noexec(mount_entry.mount, found_in_fstab=False)
```

pirat89 (changes requested):
one mini copy&paste error :) otherwise it seems good already :)

repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
MichalHe: Thank you, fixed.

repos/system_upgrade/common/actors/checkmountoptions/libraries/checkmountoptions.py
MichalHe: I've fixed the comment and extended it with the scenario you've described as it is likely to happen.

pirat89 (approved):
LGTM now! :) (untested yet)

pirat89 (approved):

PR #909, opened 2022-06-15
Detect kernel-core|kernel-rt-core on RHEL 8+

pirat89:
Since RHEL 8, the kernel|kernel-rt rpm is just a metapackage that even
does not have to be installed on the system.
The kernel-core|kernel-rt-core rpm is the one we care about instead.


MichalHe (approved):
Code looks OK, I like the explanatory comments :+1:. I've tested the patch on RHEL8 with the `kernel` metapackage removed - current master fails to recognize kernel being installed; build including this patch correctly detects that kernel is installed. I have not manually tested this patch with RHEL7, as a possible problem would be detected by CI.

holser:
I tested it in my env. LTGM

MichalHe:
Integration tests passed with the exception of two tests failed due to reasons not related to this PR. Merging.

PR #912, opened 2022-06-23
InterimPreparationPhase: change the error policy to FailImmediately

MichalHe:
Currently, the InterimPreparationPhase has its error policy set to
FailPhase, meaning that other actors are allowed to run even when an
actor fails with an error. This policy makes it possible to add a boot
entry when dnf dry run fails, as the AddUpgradeBootEntry actor does not
checks for whether the DNFDryRun really produced a message (it only
declares a dependency on its output). A possible solution would be to
handle not receiving the required message in AddUpgradeBootEntry,
however, due to the linear nature of the dependency of actors executed
during the InterimPreparationPhase, it is better to handle the
situations where such message-retrieval-checking logic is missing
globally by setting the error policy on the phase - implemented in this
patch.

### DIFF
--- a/repos/system_upgrade/common/workflows/inplace_upgrade.py
+++ b/repos/system_upgrade/common/workflows/inplace_upgrade.py
@@ -116,7 +116,7 @@ class IPUWorkflow(Workflow):
 
         name = 'InterimPreparation'
         filter = TagFilter(tags.InterimPreparationPhaseTag)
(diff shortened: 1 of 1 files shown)


fernflower (approved):
lgtm, regression tests also fine.

MichalHe:
@fernflower thanks for reviewing :relaxed:

PR #914, opened 2022-06-30
Proxy detection

PeterMocary:
The actor ``scan_pkg_manager`` now scans the configuration file of package manager for proxy, which helps to identify the reason why the installation of packages in ``target_userspace_creator`` actor failed and provide better error message. The actor ``yum_config_scanner`` is merged into the ``scan_pkg_manager`` actor due to their similarity.

Jira ref: OAMG-6180
RHBZ: 2014009 (RFE)


repos/system_upgrade/common/actors/scanpkgmanager/libraries/scanpkgmanager.py
```diff
@@ -28,5 +33,46 @@ def get_etc_releasever():
...
+        return None
+
+    # boundaries of the proxy address
+    proxy_address_start = proxy_match.end()
+    proxy_address_end = config.find('\n', proxy_address_start)
+
+    proxy_address = config[proxy_address_start:]
+    if proxy_address_end != -1:
+        # the proxy setting is not on the last line of the config file
+        proxy_address = proxy_address[:proxy_address_end]
```
MichalHe, line 66: Regular expression are very, however, they can be hard to maintain. The purpose of `re.search` is clear, however, the boundary calculation is a bit hard to think about. It is apparent, that the proxy configuration will be on one line - that's why you are searching for the closest `\n`. Therefore, the problem we are trying to solve can be stated as follows: _Find a line in the config that starts with `proxy<whitespaces>=`. If you have found such a line, return what stands on the right side of `=` (without possible whitespaces)_.

With this formulation, we could derive a simpler solution, one that does not require searching for newlines and indexing into the contents of the config file. Your solution at the moment works, which is the most important thing, however, simplicity (and thus maintainability) should be our aim, especially since this code will be likely there for a very long time.

repos/system_upgrade/common/actors/scanpkgmanager/tests/test_scanpkgmanager.py
```diff
@@ -25,6 +24,12 @@ def mocked_get_releasever_path():
     return os.path.join(CUR_DIR, 'files/releasever')
 
 
+def mock_get_config_contents(config_contents):
+    def mocked_get_config_contents(path):
+        return config_contents
+    return mocked_get_config_contents
```
MichalHe, line 30: This function can be defined locally, in the test that is using this function. Personally, I am against creating short functions that have less than `5` lines and are used only in one place.

The function could then be defined inside `test_get_proxy_if_set` as follows:
```python

@pytest.mark.parametrize('proxy_set', [True, False])
def test_get_proxy_if_set(monkeypatch, proxy_set):
    config_contents = 'Config go brrr'
    def mocked_get_config_contents(config_path):
        assert config_path == 'x'  # Check that the actor is getting contents of files you expect it to be
        return config_contents

    monkeypatch.setattr(scanpkgmanager, 'get_config_contents', mocked_get_config_contents) 
    

```

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -166,10 +167,16 @@ def prepare_target_userspace(context, userspace_dir, enabled_repos, packages):
...
-                details={'details': str(exc), 'stderr': exc.stderr}
-            )
+            message = 'Unable to install RHEL {} userspace packages.'.format(target_major_version)
+            details = {'details': str(exc), 'stderr': exc.stderr}
+
+            # If a proxy was set in dnf config, it should be the reason why dnf
+            # failed since leapp does not support updates behind proxy yet.
+            for manager_info in api.consume(PkgManagerInfo):
+                if manager_info.dnf_proxy:
+                    details['details'] = "Cannot download packages when behind proxy."
```
MichalHe, line 177: The message provided should be more user friendly, e.g. _DNF failed to install userspace packages, likely due to the proxy configuration detected in the YUM/DNF configuration._ This should be discussed with @pirat89 or @Rezney.

repos/system_upgrade/common/models/packagemanagerinfo.py
```diff
@@ -17,3 +17,8 @@ class PkgManagerInfo(Model):
     In case the value is empty string, it means the file exists but it is empty. In such a case the
     original configuration is obviously broken.
     """
+
+    dnf_proxy = fields.Nullable(fields.String())
+    yum_proxy = fields.Nullable(fields.String())
```
MichalHe, line 22: Is it important where does the proxy comes from? For the purpose of the changes introduced in this PR, IMHO it does not matter where the proxy was detected, it is important that the source system was using proxy to access new content (package updates), which could cause problems in the `target_userspace_creator`. If it the source of the proxy is important, then the checks in `target_userspace_creator` should check for both - the source system has used a means of reaching content through a proxy, and we do not posses those means which could cause the dnf error.

MichalHe (commented):
Cool patch - I love that some actors were merged and you took the effort to change all code/docs properly. I have left you some comments, aiming at building your mindset around how to approach problems, and what to consider when implementing a patch. From my past experience, it really helps as a newcomer to use these reviews to really think about what you could do better next time and to improve.

repos/system_upgrade/common/actors/scanpkgmanager/libraries/scanpkgmanager.py
```diff
@@ -28,5 +33,38 @@ def get_etc_releasever():
     return releasever
 
 
+def get_config_contents(config_path):
+    if os.path.isfile(config_path):
+        with open(config_path, 'r') as config:
+            return config.read()
+    else:
```
fernflower, line 40: Nit: else not needed

fernflower (commented):
Nice refactoring!

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -166,10 +167,17 @@ def prepare_target_userspace(context, userspace_dir, enabled_repos, packages):
...
-                details={'details': str(exc), 'stderr': exc.stderr}
-            )
+            message = 'Unable to install RHEL {} userspace packages.'.format(target_major_version)
+            details = {'details': str(exc), 'stderr': exc.stderr}
+
+            # If a proxy was set in dnf config, it should be the reason why dnf
+            # failed since leapp does not support updates behind proxy yet.
+            for manager_info in api.consume(PkgManagerInfo):
+                if manager_info.configured_proxies:
+                    details['details'] = ("DNF failed to install userspace packages, likely due to the proxy"
```
MichalHe, line 177:
```suggestion
                    details['details'] = ("DNF failed to install userspace packages, likely due to the proxy "
```

MichalHe:
Tested and works. However, proxies can be configured per repository, which could also be the source of the problem. However, this can be touched upon in a follow up task. Please fix the missing space and then we are good to go.

MichalHe:
I have started testing this patch, arriving at the following conclusions:
## Performed tests
### RHEL7
- if a proxy is configured and the `dnf` command in the `target_userspace_creator` fails, the message contains a mention of the detected proxy configuration
- if a proxy is not configured and the `dnf` command in the `target_userspace_creator` fails, the message does not contain a mention of the detected proxy configuration

I am suspending my testing of this patch, as agreed with @PeterMocary that the introduced changes should also include a scan for per-repository proxy configuration, so a follow up would not be needed then.

PeterMocary:
The target user space creator now also checks the repositories configuration files for a proxy, which could also be the source of the problem as @MichalHe pointed out.

repos/system_upgrade/common/actors/scanpkgmanager/libraries/scanpkgmanager.py
```diff
@@ -28,5 +33,53 @@ def get_etc_releasever():
     return releasever
 
 
+def get_config_contents(config_path):
```
fernflower, line 36: Nit: If it's a helper designed for a one time use in `get_configured_proxies` function I would rename the function to `_get_config_contents` to make it clear

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -166,10 +168,31 @@ def prepare_target_userspace(context, userspace_dir, enabled_repos, packages):
...
+                    raise StopActorExecutionError(
+                        message=message,
+                        details=("DNF failed to install userspace packages, likely due to the proxy "
+                                 "configuration detected in the YUM/DNF configuration file.")
+                    )
+
+            # Similarly if a proxy was set specifically for one of the repositories.
+            for repo_facts in api.consume(RepositoriesFacts):
+                for repo_file in repo_facts.repositories:
+                    for repo_data in repo_file.data:
```
fernflower, line 187: Nit: you could remove at least one cycle if you use `any`, like

```
for repo_facts in api.consume(RepositoriesFacts):
  for repo_file in repo_facts.repositories:
    if any(repo_data.proxy for repo_data in repo_file.data):
       raise StopActorExecutionError(...)
```

fernflower (commented):
Just some nit suggestions for your consideration.

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
PeterMocary: Thanks, that's a useful tip!

repos/system_upgrade/common/actors/scanpkgmanager/libraries/scanpkgmanager.py
```diff
@@ -28,5 +33,53 @@ def get_etc_releasever():
...
+    :rtype: List
+    """
+
+    configured_proxies = []
+    for config_path in (DNF_CONFIG_PATH, YUM_CONFIG_PATH):
+        proxy = _get_proxy_if_set(config_path)
+        if proxy and proxy not in configured_proxies:
+            configured_proxies.append(proxy)
+
+    return configured_proxies
```
MichalHe, line 76: Codewise okay. If you want to collect a unique set of proxies, then just use `set`. Sorting the set when returning results in a list with canonical form, in other words you can predict how will the list look, no matter of the order in what file (dnf/yum conf) was the proxy found. 
```suggestion
    configured_proxies = set()
    for config_path in (DNF_CONFIG_PATH, YUM_CONFIG_PATH):
        proxy = _get_proxy_if_set(config_path)
        if proxy:
            configured_proxies.add(proxy)

    return sorted(configured_proxies)
```

repos/system_upgrade/common/models/packagemanagerinfo.py
```diff
@@ -17,3 +17,10 @@ class PkgManagerInfo(Model):
     In case the value is empty string, it means the file exists but it is empty. In such a case the
     original configuration is obviously broken.
     """
+
+    configured_proxies = fields.List(fields.String(), default=[])
+    """
+    Contain a set of proxies from configuration files of yum and dnf.
```
MichalHe, line 23: If you decide to sort the proxies as suggested above.
```suggestion
    A sorted list of proxies present in yum and dnf configuration files.
```
If you are not going to sort it then:
```suggestion
    A list of proxies present in yum and dnf configuration files.
```

MichalHe (commented):
First pass of my review done, including some comments what needs some extra care :)

I will do a second pass tomorrow including testing, it should be good to go.

repos/system_upgrade/common/actors/scanpkgmanager/tests/test_scanpkgmanager.py
```diff
@@ -46,3 +49,47 @@ def test_get_etcreleasever(monkeypatch, etcrelease_exists):
...
+
+@pytest.mark.parametrize(
+    ('proxy_set_in_configs', 'expected_output'),
+    [
+      ((True, True), [PROXY_ADDRESS]),
+      ((True, False), [PROXY_ADDRESS]),
+      ((False, False), [])
+    ]
+)
+def test_get_configured_proxies(monkeypatch, proxy_set_in_configs, expected_output):
```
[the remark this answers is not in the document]
PeterMocary: You are right, the tuple is confusing and it should be changed. Thank you for providing such a detailed description of your point of view!

repos/system_upgrade/common/actors/targetuserspacecreator/actor.py
```diff
@@ -1,20 +1,22 @@
 from leapp.actors import Actor
 from leapp.libraries.actor import userspacegen
 from leapp.libraries.common.config import get_env, version
+from leapp.models import RequiredTargetUserspacePackages  # deprecated
+from leapp.models import TMPTargetRepositoriesFacts  # deprecated all the time
```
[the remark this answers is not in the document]
PeterMocary: The liter thinks it should be put separately.

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -166,10 +168,30 @@ def prepare_target_userspace(context, userspace_dir, enabled_repos, packages):
...
+            details = {'details': str(exc), 'stderr': exc.stderr}
+
+            # If a proxy was set in dnf config, it should be the reason why dnf
+            # failed since leapp does not support updates behind proxy yet.
+            for manager_info in api.consume(PkgManagerInfo):
+                if manager_info.configured_proxies:
+                    raise StopActorExecutionError(
+                        message=message,
+                        details=("DNF failed to install userspace packages, likely due to the proxy "
+                                 "configuration detected in the YUM/DNF configuration file.")
```
MichalHe, line 181: `details=` are supposed to be a dictionary - check how they are used above.

MichalHe:
I have tested the patch with the following results:
*RHEL7*:
- no proxy configured, the call to `dnf` in `target_userspace_creator` fails => error does not mention proxies
- proxy configured in `/etc/yum.conf`, the call to `dnf` in `target_userspace_creator` fails => error mentions proxy being likely cause of the error
- proxy configured in a repo, the call to `dnf` in `target_userspace_creator` fails => error mentions proxy being likely cause of the error
*RHEL8*:
- no proxy configured, the call to `dnf` in `target_userspace_creator` fails => error does not mention proxies
- proxy configured in `/etc/dnf/dnf.conf`, the call to `dnf` in `target_userspace_creator` fails => error mentions proxy being likely cause of the error
- proxy configured in a repo, the call to `dnf` in `target_userspace_creator` fails => error mentions proxy being likely cause of the error

The results are correct and the patch works as expected.

MichalHe (dismissed):
I have tested the patch, works as expected. Code-wise OK, thus LGTM.

Let's wait on integration tests, and if they are OK I will merge the patch.

pirat89 (dismissed):
it was discussed that there is not enough capacity to cover the testing for this release already, so it's postponed for the next one. It can be merged right after the current release is created.

# setting the required changes so it's not merged by mistake

pirat89:
@PeterMocary rebase please to resolve the conflicts

PeterMocary:
@pirat89 Rebased.

pirat89 (approved):
merging \o

PR #915, opened 2022-07-08
Fix warnings raised by shellcheck in the Coverity scan.

PeterMocary:
Warnings raised by shellcheck are fixed as recommended or suppressed whenever fix isn't necessary or possible. 

Jira ref.: OAMG-6721

### DIFF
--- a/repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/module-setup.sh
+++ b/repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/module-setup.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-LEAPPBIN=/usr/bin/leapp
-LEAPP_VERSION=0.8.10
+# Variable not used at the moment; would be needed in case of a remote upgrade
+#LEAPPBIN=/usr/bin/leapp
 
 check() {
     require_binaries base64 || return 1
@@ -30,11 +30,17 @@ install() {
     #dracut_install /etc/selinux/*/policy/*
     #dracut_install $(find /etc/selinux/*/contexts)
 
+    # shellcheck disable=SC2154 # Following variables are assigned by dracut
+    { # NOTE: use these new variables instead to avoid the warning
+    local _initdir=$initdir
+    local _moddir=$moddir
+    }
+
     # NOTE: rather remove it then keep it; we really don't want plymouth now
     #       (hopefully forever)
     # remove the plymouth text plugin so we get either graphics or details
-    rm -rf ${initdir}/$(plymouth --get-splash-plugin-path)text.so \
-          ${initdir}/usr/share/plymouth/themes/text/*
+    rm -rf "${_initdir}/$(plymouth --get-splash-plugin-path)text.so" \
+           "${_initdir}"/usr/share/plymouth/themes/text/*
 
 
     # stuff we use in upgrade hook(s)
@@ -87,8 +93,8 @@ install() {
     inst_binary grep
 
     # script to actually run the upgrader binary
-    inst_hook upgrade 49 "$moddir/mount_usr.sh"
-    inst_hook upgrade 50 "$moddir/do-upgrade.sh"
+    inst_hook upgrade 49 "$_moddir/mount_usr.sh"
+    inst_hook upgrade 50 "$_moddir/do-upgrade.sh"
 
     #NOTE: some clean up?.. ideally, everything should be inside the leapp*
     #NOTE: current *.service is changed so in case we would like to use the
--- a/repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/mount_usr.sh
+++ b/repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/mount_usr.sh
@@ -7,14 +7,13 @@ type info >/dev/null 2>&1 || . /lib/dracut-lib.sh
 export NEWROOT=${NEWROOT:-"/sysroot"}
 
 filtersubvol() {
-    local _oldifs
     _oldifs="$IFS"
     IFS=","
-    set $*
+    set "$@"
     IFS="$_oldifs"
     while [ $# -gt 0 ]; do
         case $1 in
-            subvol\=*) :;;
+            subvol=*) :;;
             *) printf '%s' "${1}," ;;
         esac
         shift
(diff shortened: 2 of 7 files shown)


repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/mount_usr.sh
```diff
@@ -38,8 +36,14 @@ mount_usr()
                     _dev="/dev/disk/by-uuid/${_dev#UUID=}"
                     ;;
             esac
+
+            # TODO: -ef replaced with inode number comparsion - is this enough?
+            # TODO: supresses all shellcheck warnings regarding declared, but
+            # not assigned variables in the whole if statement - is this the
+            # right way of dealing with this warning?
+            # shellcheck disable=SC2154 # Variable root is assigned by dracut
```
PeterMocary, line 44: SC2154 is suppressed (not only in this case), because a variable is assigned by dracut, however this could suppress warnings about other variables within the same command scope.

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/mount_usr.sh
```diff
@@ -38,8 +36,14 @@ mount_usr()
...
             if strstr "$_opts" "subvol=" && \
-                [ "${root#block:}" -ef $_dev ] && \
+                [ "`stat -c '%i' ${root#block:}`" = "`stat -c '%i' $_dev`" ] && \
```
PeterMocary, line 46: The -ef isn't defined in POSIX, so this should be equivalent way of checking, however might be false positive if the two files are on different filesystems but happen to have the same inode.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/files/generate-initram.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+# shellcheck disable=SC2164 # The directory changes are run as root
```
PeterMocary, line 2: This suppresses the SC2164 warning for the whole file. This warning requires each change of directory to handle its failure and I'm not certain if that is required or if it can be suppressed.

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/mount_usr.sh
MichalHe: If you have a look at what is the `read` reading you see that is reading some `/etc/fstab`. Therefore the `$_dev` variable should point to a device inside the `/dev` (devtmpfs filesystem). Similarly, `root` variable contains the name of partition containing the root filesystem, and therefore, should be also in `/dev`. (IIUC the `root` variable should contain the value of `root` cmdline argument). Therefore, both of these variable should point somewhere inside the same filesystem (`/dev`), meaning that inode number collisions are highly unlikely.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/files/generate-initram.sh
MichalHe: Maybe I am wrong, but there are only 2 pairs of `pushd` and `popd`, so it should not be that much code to check their result. It is unlikely that they fail, however, I would rather see a meaningful error message when they do rather than disabling the warning globally.

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/mount_usr.sh
MichalHe: You can always introduce a new local variable e.g. `_root="${root#block:}"` and disable the warning only on this assignment.

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/module-setup.sh
```diff
@@ -30,11 +30,17 @@ install() {
...
+    local _moddir=$moddir
+    }
+
     # NOTE: rather remove it then keep it; we really don't want plymouth now
     #       (hopefully forever)
     # remove the plymouth text plugin so we get either graphics or details
-    rm -rf ${initdir}/$(plymouth --get-splash-plugin-path)text.so \
-          ${initdir}/usr/share/plymouth/themes/text/*
+    rm -rf "${_initdir}/$(plymouth --get-splash-plugin-path)text.so" \
+           "${_initdir}/usr/share/plymouth/themes/text/*"
```
MichalHe, line 43: Are you sure that there will be only 1 file in that folder? Otherwise the added quotes will cause `rm` not to delete anything.

Example:
```
mkdir /tmp/test
cd /tmp/test
touch aa ba
mkdir dir
touch dir/cc
rm -rf "/tmp/test/*"
ls  # All files are still there, since "/tmp/test/*" expands to "/tmp/test/aa /tmp/test/ba /tmp/test/dir" as one string, but there is no such file
```

MichalHe (commented):
Tests have passed, however, there are other tips raise by shellcheck in these shell files. IMHO we should use this PR as an opportunity to harden our dracut scripts, e.g. by adding quoting when advised and so on. We should also make sure, that modules are posix compliant and have correct shebangs so that they are checked by shellcheck.

BTW don't worry about this being super critical part of an IPU, do what you believe is good, if anything will seem suspicious we will talk about it in a review :)

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/module-setup.sh
```diff
@@ -30,11 +30,17 @@ install() {
...
+    local _moddir=$moddir
+    }
+
     # NOTE: rather remove it then keep it; we really don't want plymouth now
     #       (hopefully forever)
     # remove the plymouth text plugin so we get either graphics or details
-    rm -rf ${initdir}/$(plymouth --get-splash-plugin-path)text.so \
-          ${initdir}/usr/share/plymouth/themes/text/*
+    rm -rf "${_initdir}/$(plymouth --get-splash-plugin-path)text.so" \
+           "${_initdir}/usr/share/plymouth/themes/text"
```
pirat89, line 43: This is changed behaviour. We do not want to remove the `text` directory. We want to remove only its content. I do not rember now the reason why not to remove the directory itself however. We could investigate that whether there is a reson for it, but right now I prefer no change that.

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/90sys-upgrade/module-setup.sh
```diff
@@ -35,16 +35,23 @@ install() {
...
-    inst_simple "$moddir/upgrade.target" "$unitdir/upgrade.target"
+    inst_simple "${_moddir}/upgrade.target" "${unitdir}/upgrade.target"
 
     mkdir -p "$upgrade_wantsdir"
     for s in $upgrade_hooks; do
-        inst_simple "$moddir/$s.service" "$unitdir/$s.service"
-        inst_script "$moddir/$s.sh"      "/bin/$s"
-        ln -sf "../$s.service" $upgrade_wantsdir
+        inst_simple "${_moddir}/${s}.service" "${unitdir}/${s}.service"
+        inst_script "${_moddir}/${s}.sh"      "/bin/$s"
```
MichalHe, line 53: I am not looking at the consistency across the entire file, however, lets be consistent at line level:
```suggestion
        inst_script "${_moddir}/${s}.sh"      "/bin/${s}"
```

MichalHe (approved):

PR #916, opened 2022-07-11
BZ#2104730 - correct PostgreSQL data ownership after move

evgeni:
When moving the PostgreSQL data between different disks, shutil.move()
does not properly transfer file ownership, so we need to fix that after
the fact.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/satellite_upgrade_data_migration/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/satellite_upgrade_data_migration/actor.py
@@ -8,6 +8,8 @@ from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
 
 POSTGRESQL_DATA_PATH = '/var/lib/pgsql/data/'
 POSTGRESQL_SCL_DATA_PATH = '/var/opt/rh/rh-postgresql12/lib/pgsql/data/'
+POSTGRESQL_USER = 'postgres'
+POSTGRESQL_GROUP = 'postgres'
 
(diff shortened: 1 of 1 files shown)


fernflower:
The change looks mostly harmless, our regression tests are also happy. Unfortunately, we don't have a setup to test the functionality directly (we are not testing satellite envs). @evgeni if you say it fixes the problem for you - I am ok to merge it.

fernflower (approved):

evgeni:
I'll talk to @lpramuk later today, to see what he thinks. But yes, in my tests it was fixing the issue.

evgeni:
Lukas is testing it right now. But I am also rather confident that it's OK ;-)

Rezney (approved):

repos/system_upgrade/el7toel8/actors/satellite_upgrade_data_migration/actor.py
```diff
@@ -41,3 +43,8 @@ def process(self):
                 os.rmdir(POSTGRESQL_DATA_PATH)
             # move PostgreSQL data to the new home
             shutil.move(POSTGRESQL_SCL_DATA_PATH, POSTGRESQL_DATA_PATH)
```
pirat89, line 45: @evgeni  moving the mountpoing leads to the crash of in-place upgrade

pirat89: the same issue happens when you try to remove mountpoint on line 43 if it exists

PR #918, opened 2022-07-12
Enable 8to9 ipu on azure vanilla

MichalHe:
Implements necessary changes to enable 8to9 upgrade on base RHEL8 running on Azure.

### DIFF
--- a/repos/system_upgrade/common/actors/redhatsignedrpmscanner/actor.py
+++ b/repos/system_upgrade/common/actors/redhatsignedrpmscanner/actor.py
@@ -55,7 +55,8 @@ class RedHatSignedRpmScanner(Actor):
             return pkg.name.startswith('katello-ca-consumer')
 
         upg_path = rhui.get_upg_path()
-        whitelisted_cloud_flavours = ('azure', 'azure-sap', 'google', 'google-sap')
+        # AWS RHUI packages do not have to be whitelisted because they are signed by RedHat
+        whitelisted_cloud_flavours = ('azure', 'azure-eus', 'azure-sap', 'google', 'google-sap')
         whitelisted_cloud_pkgs = {
             rhui.RHUI_CLOUD_MAP[upg_path].get(flavour, {}).get('src_pkg') for flavour in whitelisted_cloud_flavours
         }
--- a/repos/system_upgrade/common/libraries/module.py
+++ b/repos/system_upgrade/common/libraries/module.py
@@ -1,4 +1,7 @@
+import os
 import warnings
+
+from leapp.libraries.common.config.version import get_source_major_version
 
 try:
     import dnf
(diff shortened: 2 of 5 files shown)


repos/system_upgrade/common/libraries/module.py
```diff
@@ -16,6 +17,17 @@
 def _create_or_get_dnf_base(base=None):
     if not base:
         base = dnf.Base()
+
+        # The DNF command reads /etc/yum/vars/releasever, but the DNF library does not. It parses redhat-release
+        # package to retrieve system's major version which it then uses as $releasever. However, some systems might
+        # have repositories only for exact system version (including minor number). In case /etc/yum/vars/releasever
+        # is present, read its contents so that we can access repositores on such systems
+        releasever_path = '/etc/yum/vars/releasever'
```
MichalHe, line 25: On RHEL8 this path might not exist, check for `/etc/dnf/vars/releasever`

MichalHe: Fixed

repos/system_upgrade/common/actors/redhatsignedrpmscanner/actor.py
```diff
@@ -55,7 +55,7 @@ def has_katello_prefix(pkg):
             return pkg.name.startswith('katello-ca-consumer')
 
         upg_path = rhui.get_upg_path()
-        whitelisted_cloud_flavours = ('azure', 'azure-sap', 'google', 'google-sap')
+        whitelisted_cloud_flavours = ('azure', 'azure-eus', 'azure-sap', 'google', 'google-sap')
```
Rezney, line 58: We may add a comment that "AWS" is missing here because those packages are signed by RH.

repos/system_upgrade/common/libraries/rhui.py
```diff
@@ -146,6 +145,22 @@
                 ('leapp-azure.repo', YUM_REPOS_PATH)
             ],
         },
+        # FIXME(mhecko): This entry is identical to the azure one, since we have no EUS content yet, therefore, it
+        # #              serves only the purpose of containing the name of rhui client package to correctly detect
+        # #              cloud provider. Trying to work around this entry by specifying --channel, will result in
+        # #              failures - there is no repomapping for EUS content, and the name of target pkg differs on EUS.
+        'azure-eus': {
+            'src_pkg': 'rhui-azure-rhel8-eus',
+            'target_pkg': 'rhui-azure-rhel9',
```
Rezney, line 154: Let's comment that EUS mapping will be added later when the EUS image is released. In case such an image is available sooner than we release a new Leapp version, the switch can be done using official MS docs.

Rezney (changes requested):
Works like a charm. Please add the requested comments and we can merge it.

repos/system_upgrade/common/libraries/rhui.py
MichalHe: The switch cannot be done since I presume the target package will be different. We can either predict the name of target package, but make the EUS 8to9 upgrade unavailable, or we can keep it as it is, which would mean that we have to patch this piece of code when the EUS image is available.

repos/system_upgrade/common/actors/redhatsignedrpmscanner/actor.py
MichalHe: Fixed

repos/system_upgrade/common/libraries/rhui.py
Rezney: It can be done after the upgrade:
https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/redhat/redhat-rhui#switch-a-rhel-vm-8x-to-eus-version-lock-to-a-specific-minor-version

MichalHe: I have added the comment with the most recent push

Rezney (approved):

PR #919, opened 2022-07-13
Update LsblkEntry (StorageInfo) and fix the vdo scanner

pirat89:
The /usr/libexec/vdopreparelvm binary fails when try to check
devices/partitions smaller than 4 MiB as VDO does not count
with any use-case with such small partitions. To cover this,
skip all such small devices when checking for the converted
VDO devices. (rhbz#2090830)

Also it could happen on some systems that check fails when the
NAME and KERNEL NAME of a device is different as the actor
checked devices just on `/dev/{NAME}` path. This should be mostly
ok when checking `/dev/{KNAME}` instead. However after discussion
with @jshimkus-rh and @bmarzins we know there could be cases when
even this is not true. In such cases, report the valid error msg
so it's more clear what actually happened.
(rhbz#206159)

To simplify the solution mentioned above, the `LsblkEntry` model
has been extended to cover kernel name (`kname`) and size in
bytes (`bsize`) as well. Includes changes in the `storagescanner`
actor.

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=2090830
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=2096159

-----------
## TODO
- [x] update unit-tests
- [x] testing (pretesting done, excluding report of the missing media that I do not have idea how to reproduce as it's corner case; covered at least by unit-tests)

### DIFF
--- a/repos/system_upgrade/common/actors/inhibitwhenluks/tests/test_inhibitwhenluks.py
+++ b/repos/system_upgrade/common/actors/inhibitwhenluks/tests/test_inhibitwhenluks.py
@@ -5,9 +5,8 @@ from leapp.utils.report import is_inhibitor
 
 
 def test_actor_with_luks(current_actor_context):
-    with_luks = [LsblkEntry(name='luks-132', maj_min='253:0', rm='0',
-                            size='10G', ro='0', tp='crypt', mountpoint=''
-                            )]
+    with_luks = [LsblkEntry(name='luks-132', kname='kname1', maj_min='253:0', rm='0',
+                            size='10G', bsize=10*(1 << 39), ro='0', tp='crypt', mountpoint='')]
 
     current_actor_context.feed(StorageInfo(lsblk=with_luks))
     current_actor_context.run()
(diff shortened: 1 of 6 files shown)


pirat89:
@jshimkus-rh  ping - something like that

jshimkus-rh:
Looks reasonable though I want to understand why skipping the device if it doesn't exist via the kname path is safe.

pirat89:
@jshimkus-rh well, not sure whether it's safe. That's what I want to discuss with you. The actual point is that when we cannot find the device on the path, there is nothing to check and the vdo tooling is raising cryptic error msgs. If we do not believe that using `/dev/{kname}` path is safe, I think we could extend the `LsblkEntry` model by `dev_path` field which will be definitely safe -- in phases before the reboot - after the reboot the path could be different (which is one of reasons why I skipped adding that field into the model).

pirat89:
Discussed with Joe & Ben Marzinsky, we will drop the skipping of non-existing files for block devices to stay on the safe side. There are cases when the block devices does not have to necessary be registered under `/dev/<kname>` and we are not 100% sure whether such devices could be used for VDO or not.

pirat89:
@jshimkus-rh  @vinzenz  prepared for the final review

vinzenz (approved):
Code changes seems fine to me, waiting for E2E to be not affected.

jshimkus-rh (approved):

pirat89:
Thanks \o  Merging.

PR #920, opened 2022-07-13
Fix for RHEL 8.6 to 9 upgrades on AWS

thimslugga:
This change is a fix for RHEL 8.6 to 9 leapp upgrades on AWS. The src client package for AWS should be rh-amazon-rhui-client as rh-amazon-rhui-client-ha is the RHUI client package for the RHEL with HA add-on image.

I also tested the change and confirmed it resolves the issue.

Bugzilla: [rhbz#2106904](https://bugzilla.redhat.com/show_bug.cgi?id=2106904)

### DIFF
--- a/repos/system_upgrade/common/libraries/rhui.py
+++ b/repos/system_upgrade/common/libraries/rhui.py
@@ -105,7 +105,7 @@ RHUI_CLOUD_MAP = {
     },
     '8to9': {
         'aws': {
-            'src_pkg': 'rh-amazon-rhui-client-ha',
+            'src_pkg': 'rh-amazon-rhui-client',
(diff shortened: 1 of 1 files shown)


pirat89:
@thimslugga thank you for the PR, we will get to it in upcoming weeks at latest. Definitely we want to have this problem resolved with the next release.

fernflower:
@MichalHe There is some weird rhel-8.6-rhui failure http://artifacts.osci.redhat.com/testing-farm/c4026cd7-6b71-4374-8cb7-161e7379f2a7/ actually. Not yet sure if it's me screwing up tests refactor or really related to the PR

PR #921, opened 2022-07-20
Check if /var/lib/leapp is mounted in a persistent fashion

matejmatuska:
A new `check_persistent_mounts` actor checks if `/var/lib/leapp` is mounted
and in that case checks if there is an `/etc/fstab` entry with this
mountpoint. If there isn't, the upgrade is inhibited. This ensures
`/var/lib/leapp/` will be available after a reboot for leapp toaccess the
upgrade environment. access the upgrade environment.

Jira ref: OAMG-7070

Tested on RHEL7 by mounting `/var/lib/leapp' to a directory using bind mount. Example report:

```
Risk Factor: high (inhibitor)
Title: Detected partitions mounted in a non-persistent fashion, preventing a successful in-place upgrade.
Summary: Leapp detected that the /var/lib/leapp mountpoint is mounted, with no corresponding entry in /etc/fstab. The upgrade process cannot continue with /var/lib/leapp mounted in non-persistent fashion, as Leapp needs this mount to be available after a reboot.
Remediation: [hint] Add /var/lib/leapp mount entry to /etc/fstab
```

### DIFF
--- a/repos/system_upgrade/common/actors/checkpersistentmounts/libraries/checkpersistentmounts.py
+++ b/repos/system_upgrade/common/actors/checkpersistentmounts/libraries/checkpersistentmounts.py
@@ -1,0 +1,43 @@
+from leapp import reporting
+from leapp.libraries.stdlib import api
+from leapp.models import StorageInfo
+
+
+def inhibit_upgrade_due_non_persistent_mount(mountpoint):
+    summary = (
+        'Leapp detected that the {0} mountpoint is mounted, with no corresponding entry in /etc/fstab. '
+        'The upgrade process cannot continue with {0} mounted in non-persistent fashion, '
+        'as Leapp needs this mount to be available after a reboot.'
+    )
+
+    hint = (
+        'Add {0} mount entry to /etc/fstab'
+    )
+
+    reporting.create_report([
+        reporting.Title(
+            'Detected partitions mounted in a non-persistent fashion, preventing a successful in-place upgrade.'
+        ),
+        reporting.Summary(summary.format(mountpoint)),
+        reporting.Remediation(hint=hint.format(mountpoint)),
+        reporting.RelatedResource('file', '/etc/fstab'),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Tags([reporting.Tags.FILESYSTEM]),
+        reporting.Flags([reporting.Flags.INHIBITOR]),
+    ])
+
+
+def check_mount_is_persistent(storage_info, mountpoint):
+    """Check if mountpoint is mounted in persistent fashion"""
+
+    mount_entry_exists = any(me.mount == mountpoint for me in storage_info.mount)
+    fstab_entry_exists = any(fe.fs_file == mountpoint for fe in storage_info.fstab)
+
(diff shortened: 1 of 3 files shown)


repos/system_upgrade/common/actors/checkpersistentmounts/libraries/checkpersistentmounts.py
```diff
@@ -0,0 +1,56 @@
...
+    mount_entry_exists = mount_entry_with_mountpoint_exists(storage_info.mount, mountpoint)
+    fstab_entry_exists = fstab_entry_with_mountpoint_exists(storage_info.fstab, mountpoint)
+
+    if mount_entry_exists and not fstab_entry_exists:
+        inhibit_upgrade_due_non_persistent_mount(mountpoint)
+
+
+def check_persistent_mounts():
+    for storage_info in api.consume(StorageInfo):
+        check_mount_is_persistent(storage_info, '/var/lib/leapp')
```
vinzenz, line 56: We should check this only *once* and only for the first message received. The expectation is that there's only one StorageInfo message. Also we want to avoid multiple inhibiting reports for potentially multiple StorageInfo messages.
```suggestion
     storage_info = next(api.consume(StorageInfo), None)
     if storage_info:
        check_mount_is_persistent(storage_info, '/var/lib/leapp')
```

vinzenz (dismissed):
Very nice PR, besides the comment about multiple StorageInfo messages and the potential for multiple reports due to that.

repos/system_upgrade/common/actors/checkpersistentmounts/libraries/checkpersistentmounts.py
```diff
@@ -0,0 +1,57 @@
...
+        reporting.Remediation(hint=hint.format(mountpoint)),
+        reporting.RelatedResource('file', '/etc/fstab'),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Tags([reporting.Tags.FILESYSTEM]),
+        reporting.Flags([reporting.Flags.INHIBITOR]),
+    ])
+
+
+def mount_entry_with_mountpoint_exists(mount_entries, mountpoint):
+    for mount_entry in mount_entries:
```
fernflower, line 31: Nit: L31-L34 are equivalent to `return any(me.mount == mountpoint for me in mount_entries)`

repos/system_upgrade/common/actors/checkpersistentmounts/libraries/checkpersistentmounts.py
```diff
@@ -0,0 +1,57 @@
...
+        if mount_entry.mount == mountpoint:
+            return True
+    return False
+
+
+def fstab_entry_with_mountpoint_exists(fstab_entries, mountpoint):
+    for fstab_entry in fstab_entries:
```
fernflower, line 38: Nit: same here

Also if you decide to go with refactoring you might consider getting rid of those helpers and just having in-place checks at L47-L48.

fernflower (approved):
lgtm, all comments addressed, both from me and @vinzenz

vinzenz (approved):

PR #922, opened 2022-07-21
sssd: Fix sssd startup problem due to cache files

vinzenz:
In RHEL8 sssd uses a lower version for the sssd cache file format than
in RHEL7.9 which causes the SSSD service to fail on startup after the
upgrade.

This patch solves this problem by adding a workaround to remove the sssd
cache files during the upgrade.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2109309
Jira ref.: OAMG-7246

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/sssd_handle_cache_files/libraries/sssd_cache_files.py
+++ b/repos/system_upgrade/el7toel8/actors/sssd_handle_cache_files/libraries/sssd_cache_files.py
@@ -1,0 +1,19 @@
+import glob
+import os
+
+from leapp.libraries.stdlib import api
+
+
+def remove_sssd_cache_files(config):
+    """
+    Remove all SSSD cache files.
+    """
+    if not config:
+        return
+
+    for cache_file in glob.glob('/var/lib/sss/db/*.ldb'):
+        try:
+            api.current_logger().info('Pruning SSSD cache file {}'.format(cache_file))
+            os.remove(cache_file)
+        except OSError:
+            api.current_logger().warning('Failed to remove cache file: %s', cache_file)
--- a/repos/system_upgrade/el7toel8/actors/sssd_handle_cache_files/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/sssd_handle_cache_files/actor.py
@@ -1,0 +1,23 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.sssd_cache_files import remove_sssd_cache_files
+from leapp.models import SSSDConfig
+from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
+
+
+class SSSDHandleCacheFiles(Actor):
+    """
+    If an SSSD configuration has been detected, special handling for SSSD cache files will be scheduled.
+
+    After the upgrade from EL7 to EL8, SSSD isn't able to start anymore because the cache files have a newer
+    format in RHEL7.9 than in RHEL8. This is a workaround for this issue.
+    """
+
+    name = 'sssd_handle_cache_files'
+    consumes = (SSSDConfig,)
+    produces = ()
+    tags = (IPUWorkflowTag, ApplicationsPhaseTag)
+
+    def process(self):
+        remove_sssd_cache_files(
+            next(self.consume(SSSDConfig), None)
+        )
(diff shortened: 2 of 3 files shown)


evgeni:
Why did you chose the DNFWorkaround framework, and not a "normal" actor running in the ApplicationPhase that would do a `glob` and `os.unlink`?

vinzenz:
> Why did you chose the DNFWorkaround framework, and not a "normal" actor running in the ApplicationPhase that would do a `glob` and `os.unlink`?

TBH it was just a quick workaround to have something, but you're right an actor in the ApplicationPhase would be easier as a drop-in workaround for CEE & Solution Architects until we have some fix coming a long.

evgeni:
Yeah, both ways should work in the end.

Did you find someone to tell which files to delete? In my tests the one you have was sufficient, but other online docs that were talking about recovery from a too new DB we're talking about deleting all files in that directory.

vinzenz:
> Yeah, both ways should work in the end.
> 
> Did you find someone to tell which files to delete? In my tests the one you have was sufficient, but other online docs that were talking about recovery from a too new DB we're talking about deleting all files in that directory.

Yes - SSSD guys told me that deleting all files should be better.

alexey-tikhonov:
> In RHEL8 sssd uses a lower version for the sssd cache file format than in RHEL7.9 which causes the SSSD service to fail on startup after the upgrade.

No, that's not the reason.

I think reason is wrong / unexpected files ownership in '/var/lib/sss/db'
Since https://github.com/SSSD/sssd/pull/5782 it should be sssd:sssd, and it seems on RHEL7 machine it was root:root?

Need to confirm with @thalman

alexey-tikhonov:
> No, that's not the reason: in the logs upgrade is 0.22 -> 0.23

https://bugzilla.redhat.com/show_bug.cgi?id=2109309#c17

So I think cache removal is a way to go.

pirat89 (approved):
lgtm. I will do the manual test yet before the merge just for sure, but it seems clearly good. we will merge it this week for sure :)

repos/system_upgrade/el7toel8/actors/sssd_handle_cache_files/libraries/sssd_cache_files.py
```diff
@@ -0,0 +1,18 @@
...
+def remove_sssd_cache_files(config):
+    """
+    Remove all SSSD cache files.
+    """
+    if not config:
+        return
+
+    for cache_file in glob.glob('/var/lib/sss/db/*.ldb'):
+        try:
+            os.remove(cache_file)
```
pirat89, line 16:
```suggestion
            api.current_logger().info('Prune the cache file: %s', cache_file)
            os.remove(cache_file)
```

pirat89:
tested manually, it seems to be working as expected. 
```
##### orig system

[root@localhost ~]# systemctl status sssd
● sssd.service - System Security Services Daemon
   Loaded: loaded (/usr/lib/systemd/system/sssd.service; enabled; vendor preset: disabled)
   Active: active (running) since Fri 2022-08-19 13:54:40 CEST; 3min 14s ago
 Main PID: 657 (sssd)
   CGroup: /system.slice/sssd.service
           ├─657 /usr/sbin/sssd -i --logger=files
           ├─682 /usr/libexec/sssd/sssd_be --domain default --uid 0 --gid 0 --logger=files
           └─687 /usr/libexec/sssd/sssd_autofs --uid 0 --gid 0 --logger=files

.....

#### after the upgrade
# systemctl status sssd
● sssd.service - System Security Services Daemon
   Loaded: loaded (/usr/lib/systemd/system/sssd.service; enabled; vendor preset: enabled)
   Active: active (running) since Fri 2022-08-19 14:31:19 CEST; 1min 43s ago
 Main PID: 865 (sssd)
    Tasks: 4 (limit: 11374)
   Memory: 18.3M
   CGroup: /system.slice/sssd.service
           ├─865 /usr/sbin/sssd -i --logger=files
           ├─889 /usr/libexec/sssd/sssd_be --domain implicit_files --uid 0 --gid 0 --logger=files
           ├─890 /usr/libexec/sssd/sssd_be --domain default --uid 0 --gid 0 --logger=files
           └─891 /usr/libexec/sssd/sssd_autofs --uid 0 --gid 0 --logger=files
....

```

pirat89 (approved):

pirat89:
Internal infrastracture is probably down. I believe it works based on previous testing and recently added log does not break it. merging

PR #923, opened 2022-07-22
Add Makefile targets for running tests in containers

matejmatuska:
Running tests in containers brings multiple benefits:
 - avoids conflicts between multiple repositories (unsafe namespaces)
 - allows running test on host where tests are broken (e.g. Fedora 35)
 - avoids polluting the host with dependencies, the only dependency is `podman`

To run tests in container run `make test_container`.
To run justs test without linitinh run `make test_container_no_lint`.
By default `rhel8` container is used. To change the container set the
`TEST_CONTAINER` environment variable. For example
`TEST_CONTAINER="rhel7" make test_container`.

There are also targets for running tests in all available containers,
`test_container_all` and `test_container_no_lint`.

Since tests are already being run in parallel, the `test_container_all` and
`test_container_all_no_lint` targets don't run containers in parallel as
this would bring no benefit.

Available containers are `f34`, `rhel7` and `rhel8`.

On first run container image with dependencies installed is built. This
acts as a cache to avoid installing dependencies on every run.
Subsequent runs reuse the image. To delete these images (to force a
rebuild for example) run `make clean_containers`.

Cointainer images are built from Containerfiles in
`utils/container-tests/`. Existing `Containerfile.ubi*` container files
couldn't be reused, because extra dependencies are needed e.g. `rsync`
and because of having `ENTRYPOINT` set, which is undesirable as
container setup is needed before running the tests.

On some Python versions two IPUs are tested, because for example both
el7toeel8 and el8toel9 IPUs must work on python3.6 as this is the
version on RHEL8.

Tests are run in a copy of the repository, host is not modified. For
copying `rsync` is used to easily keep repo on the host in sync with the
copy in the container and exclude system specific files we don't want to
sync such as virtualenv. Since rsync is called only from inside the
containers it is not required to be installed on the host system.

Jira ref: OAMG-6965

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,9 @@ _COPR_REPO=$${COPR_REPO:-leapp}
 _COPR_REPO_TMP=$${COPR_REPO_TMP:-leapp-tmp}
 _COPR_CONFIG=$${COPR_CONFIG:-~/.config/copr_rh_oamg.conf}
 
+# container to run tests in
+_TEST_CONTAINER=$${TEST_CONTAINER:-rhel8}
+
 # In case just specific CHROOTs should be used for the COPR build, you can
 # set the multiple CHROOTs separated by comma in the COPR_CHROOT envar, e.g.
 # "epel-7-x86_64,epel-8-x86_64". But for the copr-cli utility, each of them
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -288,6 +296,60 @@ test_no_lint:
...
+test_no_lint_container:
+	_TEST_CONT_TARGET="test_no_lint" $(MAKE) test_container
+
+test_container_all:
+	for container in "f34" "rhel8"; do \
+		TEST_CONTAINER=$$container $(MAKE) test_container; \
+	done
+
+clean_containers:
+	podman rmi -i "leapp-repo-tests-f34" "leapp-repo-tests-rhel8"
```
pirat89, line 351: To be compatible with older versions of podman
```suggestion
	for i in "leapp-repo-tests-f34" "leapp-repo-tests-rhel8"; do \
               podman rmi -f $$i >/dev/null 2>&1 || : ; \
       done

```

pirat89 (commented):
after some testing we have found some additional issues that need to be addressed.
currently the last one I have in my mind still is problem with "still running containers". Running tests in containers, anytime they are interrupted, container still exists and it's running, so additional further executions of tests are crashing becaused of that. Not sure whether podman kill & rm are alled when tests do not pass. Currently I have a funny experience with my podman after some experiments so I need to fix my env before additional testing :D

pirat89 (commented):
lgtm and manually tested. Works as expected. Currently it's not working the run of tests for particular actors and testing of libs, but I prefer to merge it in the current shape and keep the rest for follow up PRs. Also implementation for running with docker is kept for a follow up PR.

Discussed with Matej, waiting yet for commit msg / PR descr. update and we can merge it.

pirat89 (approved):
thanks for the great work! merging. The failed CI is unrelated to the changes in this PR.

PR #925, opened 2022-07-26
selinux: Improve importing customizations

vmojzis:
Re-applying customizations in a single transaction may fail in case a
package already applied it during upgrade. Fall back to applying
customizations one by one.

Stop using a file for "semanage import", since "stdin" is much cleaner.

Do not import "delete all" commands to avoid removing customizations
done by package scripts during upgrade.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=2111074

Signed-off-by: Vit Mojzis <vmojzis@redhat.com>

### DIFF
--- a/repos/system_upgrade/common/actors/selinux/selinuxapplycustom/actor.py
+++ b/repos/system_upgrade/common/actors/selinux/selinuxapplycustom/actor.py
@@ -101,6 +101,8 @@ class SELinuxApplyCustom(Actor):
                 )
                 # Retry, but install each module separately
                 for module in semodules.modules:
+                    if module.name in installed_modules:
+                        continue
                     cil_filename = os.path.join(
                         WORKING_DIRECTORY, '{}.cil'.format(module.name)
                     )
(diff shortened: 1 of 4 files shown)


vmojzis:
Needs some more testing before merging.

vmojzis:
> Needs some more testing before merging.

The new functionality seems to work fine at least on rhel 8 (I couldn't install leapp-upgrade package on rhel 7 because of missing dependencies, so no testing there... but I'm not using any new selinux commands so it should be fine). Ready for review :)

pirat89:
@vmojzis wdym by missing dependencies on rhel7? you should be able to install it on rhel7 for sure

pirat89:
@vmojzis it seems good. let us know when the testing is finished so we could merge it

vmojzis:
> @vmojzis wdym by missing dependencies on rhel7? you should be able to install it on rhel7 for sure

Some dnf subpackage was not available and downloading it from brew only showed other missing dependencies... I guess you have some repository for that.

> @vmojzis it seems good. let us know when the testing is finished so we could merge it

Go ahead. It seems to work as intended.

pirat89:
> > @vmojzis wdym by missing dependencies on rhel7? you should be able to install it on rhel7 for sure
> > @pirat89
> > Some dnf subpackage was not available and downloading it from brew only showed other missing dependencies... I guess you have some repository for that.

Ah, you need to have enabled `rhel-7-server-extras-rpms` repository on RHEL 7 to see dnf.

> > @vmojzis it seems good. let us know when the testing is finished so we could merge it
> 
> Go ahead. It seems to work as intended.

Thanks for the info. I wil rerun the tests as there have been some issues in the infrastrusture and merge it.

pirat89 (approved):
lgtm & works. thanks \o

PR #926, opened 2022-07-28
RHUI: Add 7to8 upgrade path for SAP Apps on Azure

MichalHe:
At the moment Leapp supports upgrades of SAP HA RHEL7 systems on Azure, however, we have found that there is another variant of SAP RHEL7 systems which will require a special care in order for Leapp to be able to upgrade it correctly. This patch adds a support for 7to8 upgrade of SAP Apps systems on Azure.


alexxa:
LGTM. Thanks

Rezney (approved):

PR #927, opened 2022-07-28
Changes around root_scanner to accommodate with non-utf8 filenames

fernflower:
Once the issue if fixed the tests won't be failing
anymore.

OAMG-4306

### DIFF
--- a/packaging/leapp-repository.spec
+++ b/packaging/leapp-repository.spec
@@ -96,7 +96,7 @@ Requires:       leapp-repository-dependencies = %{leapp_repo_deps}
 
 # IMPORTANT: this is capability provided by the leapp framework rpm.
 # Check that 'version' instead of the real framework rpm version.
-Requires:       leapp-framework >= 3.0, leapp-framework < 4
+Requires:       leapp-framework >= 3.1, leapp-framework < 4
 
 # Since we provide sub-commands for the leapp utility, we expect the leapp
 # tool to be installed as well.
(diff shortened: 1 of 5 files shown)


Rezney:
Looks good to me. AFAIK, @pirat89 has had some concerns regarding this solution. Petr, could you please share your concerns and/or potentially suggest a different approach? Thanks...

pirat89:
@fernflower  @oamg/developers  I would like to open the discussion on this solution before it's merged. tbh, right now we expect to work with string everywhere and making this field bytestring means that if anyone would like to work with it, it should code/encode the field or any other fields working with the information. I am mostly speaking about the design question, whether we want to have bytestream type at all in fields. you know, even the simple check cmp of strings does not work when we mix bytestream and string. e.g. `b"str str" == "str str"` returns `False`.

vinzenz:
Sorry for chiming in so late. 

I needed a moment to realize what this is actually about and how the problems addressed is actually impacting the whole upgrade process.


There are multiple things that aren't a good approach in this PR and let me suggest some alternative approach for it.

1. You change the types of an existing model, even though similar it's not the same type, so this would actually warrant for our deprecation 'policy'
2. However the attempted solution has the issue that it kind of delays the problem at hand. Now the root scanner is failing but in the end we don't know what the encoding is of the data that is put into that 'ByteString' so if some code is trying to decode it as utf-8 it will fail at that stage, and that might be during the initram phase which can leave your system in an unbootable state. (That's if someone would use this information for anything e.g. third parties)


So the suggestions on how to approach this:
1. Do not change RootSubdirectory instead create a {Failed,Invalid}RootSubdirectory, or whatever more suitable name, type with the fields you proposed
2. Extend RootDirectory to include a list of failed names
3. Harden the scanner that it will allow to fail with those non utf-8 names bug log a warning about that with 'WARNING' level

Last, but not least, even irrelevant I noticed now how 'ByteString' is misleading in the name, I would like to ask you to rename it to Blob instead, that will make it clear that whatever is in there might not be a valid string. I'd expect a ByteString to be decodable, but it actually isn't.

Thanks.

Rezney (approved):

PR #928, opened 2022-07-29
Inhibit Satellite migrations if PG data is a mount

ekohl:
shutil.move fails if the source directory is a mountpoint. The short term solution is to inhibit the migration so the user doesn't end up in a broken state. Long term we may want to introduce a bind mount or otherwise.

https://bugzilla.redhat.com/show_bug.cgi?id=2111835

This is a draft PR to start the conversation.


ekohl:
https://github.com/oamg/leapp-repository/pull/934 was merged instead

PR #932, opened 2022-08-02
Replace use of reporting.(Tags|Flags) by reporting.Groups

pirat89:
The next version of the leapp framework deprecates reporting Tags
and Flags primitives. These are replaced by generic Groups.

Also apply mass refactoring for imports to make linters happy.
As we are already doing mass refactoring, let's do it properly :smiling_imp: 

## Related PRs:
* tba
* https://github.com/oamg/leapp/pull/788

## Instructions for easy refactoring in other PRs (after the merge of this one)
1. get up-to-date master branch
2. rebase your branch against the up-to-date master branch
3. go to `repos` directory
4. apply:
```bash
for i in $(grep -Er "reporting\.(Tags|Flags)" | cut -d ":" -f1 | sort | uniq); do   sed -ri "s/reporting\.(Flags|Tags)/reporting.Groups/g" $i; done
```
Note: in case you want to update your actors prior this PR, apply the command above just on your actors 

## TODO:
* [x] rebase against up-to-date master when all other PRs for the `7.9/8.7` milestone are merged
* [x] check that we are ok with tests; or how much work could be possibly needed to add this change
* [x] the isort is now applied just to set of actors; but it should be applied to all of them; I tried that but some fixes lead into less readable imports, so some import will need to be changed manually
* [x] probably some fixed import use less readable format now, check & fix them as well


pirat89:
@oamg/developers  running tests again. There was a change in internal services which broke majority of tests.

vinzenz:
Please rebase

pirat89:
@fernflower @MichalHe I have problem with `isort`. I've manually fixed some autoformatted outputs as from my pov it was mor readable before. However, isort wants strictly sometimes grid, sometimes one line, sometimes multiline. Which to be honest does not make sense at all. I've checked the configuration before and to be honest I do not see there a configuration that could satisfy me. So I am thinking to drop the check with `isort` at all.

MichalHe:
I am okay with dropping isort temporarily, since I prefer the original style. I have looked at what is isort doing under the hood, and it should not be a big deal for us to patch it according to what we really want.

MichalHe (approved):
Tests have passed, with a few failures not related to the changes introduced in this PR. I have manually grepped the codebase for `Flags` and `Tags`, not finding any remnants, thus we should be good to go. LGTM!

PR #933, opened 2022-08-02
Add possibility to use docker to run tests and build in containers

matejmatuska:
On some systems podman is unavailable. Also, some people prefer docker
over podman. Add possibility to use docker to run tests and build
packages.

Set the `CONTAINER_TOOL` environment variable to change the container
tool used for running tests and building packages, for example:
`CONTAINER_TOOL="docker" make test_container`.

Unless you are running the Docker daemon in rootless mode, you will have
to run as root. In such case files created in `packaging/` using `make
build_container` will be owned by root. To change the ownership back to
your user, you can for example run `chown -R $USER packaging/`.

Several podman commands and options such as `container exists` are
unavailable in docker, therefore a more ugly but compatible alternative
is used.

It is also impossible to mount a volume during `docker build`. The
Containerfile `COPY` directive works around that.

Jira ref: OAMG-7286

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -224,7 +224,6 @@ _build_local: source
 	@mv packaging/$(PKGNAME).spec.bak packaging/$(PKGNAME).spec
 
 build_container:
-	@[ "$(_CONTAINER_TOOL)" == "docker" -a "$$UID" -ne 0 ] && { echo "Container tool 'docker' must be run as root!"; exit 1; }; \
 	echo "--- Build RPM ${PKGNAME}-${VERSION}-${RELEASE}.el$(DIST_VERSION).rpm in container ---"; \
 	case "$(BUILD_CONTAINER)" in \
 		el7) \
@@ -381,7 +380,6 @@ _test_container_ipu:
 # because e.g. RHEL7 to RHEL8 IPU must work on python2.7 and python3.6
 # and RHEL8 to RHEL9 IPU must work on python3.6 and python3.9.
 test_container:
-	@[ "$(_CONTAINER_TOOL)" == "docker" -a "$$UID" -ne 0 ] && { echo "The 'docker' container tool must be run as root!"; exit 1; }; \
 	case $(_TEST_CONTAINER) in \
 	f34) \
 		export CONT_FILE="utils/container-tests/Containerfile.f34"; \
@@ -441,7 +439,6 @@ test_container_all_no_lint:
 
 # clean all testing and building containers and their images
 clean_containers:
-	@[ "$(_CONTAINER_TOOL)" == "docker" -a "$$UID" -ne 0 ] && { echo "Container tool 'docker' must be run as root!"; exit 1; }; \
 	for i in "leapp-repo-tests-f34" "leapp-repo-tests-rhel7" "leapp-repo-tests-rhel8" \
 	"leapp-repo-build-el7" "leapp-repo-build-el8"; do \
 		$(_CONTAINER_TOOL) kill "$$i-cont" || :; \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -347,9 +352,9 @@ test: lint test_no_lint
 _build_container_image:
 	@[ -z "$$CONT_FILE" ] && { echo "CONT_FILE must be set"; exit 1; } || \
 	[ -z "$$TEST_IMAGE" ] && { echo "TEST_IMAGE must be set"; exit 1; }; \
-	podman image exists "$$TEST_IMAGE" && exit 0; \
+	$(_CONTAINER_TOOL) image inspect "$$TEST_IMAGE" > /dev/null 2>&1 && exit 0; \
 	echo "=========== Building container test env image ==========="; \
-	podman build -f $$CONT_FILE --tag $$TEST_IMAGE -v $$PWD:/repo:Z
+	$(_CONTAINER_TOOL) build -f $$CONT_FILE --tag $$TEST_IMAGE .
```
pirat89, line 357: @matejmatuska why is the volume replaced bz `.` ?

Makefile
```diff
@@ -367,16 +372,17 @@ _test_container_ipu:
...
+	$(_CONTAINER_TOOL) exec -w /repocopy -e REPOSITORIES $$_CONT_NAME make $${_TEST_CONT_TARGET:-test}
 
 # Runs tests in a container
 # Builds testing image first if it doesn't exist
 # On some Python versions, we need to test both IPUs,
 # because e.g. RHEL7 to RHEL8 IPU must work on python2.7 and python3.6
 # and RHEL8 to RHEL9 IPU must work on python3.6 and python3.9.
 test_container:
-	@case $(_TEST_CONTAINER) in \
+	@[ "$(_CONTAINER_TOOL)" == "docker" -a "$$UID" -ne 0 ] && { echo "Container tool 'docker' must be run as root!"; exit 1; }; \
```
pirat89, line 384:
```suggestion
	@[ "$(_CONTAINER_TOOL)" == "docker" -a "$$UID" -ne 0 ] && { echo "The 'docker' container tool must be run as root!"; exit 1; }; \
```

Makefile
matejmatuska: The volume is no longer necessary, instead of copying using `cp` inside the containers during build, the `COPY` Dockerfile command is used. For that to work the last parameter, `context` must be set, hence the `.`.

pirat89: thanks for the clarification!

pirat89:
lgtm. tested with podman, works as expected. waiting for @vinzenz  for docker testing

Makefile
```diff
@@ -220,8 +224,9 @@ _build_local: source
 	@mv packaging/$(PKGNAME).spec.bak packaging/$(PKGNAME).spec
 
 build_container:
-	@echo "--- Build RPM ${PKGNAME}-${VERSION}-${RELEASE}.el$(DIST_VERSION).rpm in container ---"
-	@case "$(BUILD_CONTAINER)" in \
+	@[ "$(_CONTAINER_TOOL)" == "docker" -a "$$UID" -ne 0 ] && { echo "Container tool 'docker' must be run as root!"; exit 1; }; \
```
vinzenz, line 227: Actually this is not true for macos, I don't need to run it as root. Also docker afaik can be configured for non root access.

matejmatuska: Maybe the implementation on MacOS is different. 9n Linux rootless mode was introduced in docker version 20.10 which is pretty recent and not available on a lot of systems.

Do you suggest to just drop the check?

vinzenz (approved):

vinzenz:
LGTM & tested on macos & fedora

PR #934, opened 2022-08-08
BZ#2111835 - move PostgreSQL data content individually

evgeni:
Some users prefer to have the data directory on an own mountpoint (vs
the whole lib/pgsql directory). In such setups we cannot just move data
to the new location, as we can't move the mointpoint.
Instead, lets move the contents of the data directory individually,
which should work fine in this scenario.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/satellite_upgrade_data_migration/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/satellite_upgrade_data_migration/actor.py
@@ -35,16 +35,27 @@ class SatelliteUpgradeDataMigration(Actor):
         # disable services, will be re-enabled by the installer
         for service_name in SERVICES_TO_DISABLE:
             for service in glob.glob(os.path.join(SYSTEMD_WANTS_BASE, '{}.service'.format(service_name))):
-                os.unlink(service)
+                try:
+                    os.unlink(service)
+                except Exception as e:  # pylint: disable=broad-except
+                    self.log.warning('Failed disabling service {}: {}'.format(service, e))
 
         if facts.postgresql.local_postgresql and os.path.exists(POSTGRESQL_SCL_DATA_PATH):
-            # remove empty PostgreSQL data from the package
-            if os.path.exists(POSTGRESQL_DATA_PATH):
-                os.rmdir(POSTGRESQL_DATA_PATH)
+            # we can assume POSTGRESQL_DATA_PATH exists and is empty
             # move PostgreSQL data to the new home
-            shutil.move(POSTGRESQL_SCL_DATA_PATH, POSTGRESQL_DATA_PATH)
+            for item in glob.glob(os.path.join(POSTGRESQL_SCL_DATA_PATH, '*')):
+                try:
+                    shutil.move(item, POSTGRESQL_DATA_PATH)
+                except Exception as e:  # pylint: disable=broad-except
+                    self.log.warning('Failed moving PostgreSQL data: {}'.format(e))
+                    return
+
             if not facts.postgresql.same_partition:
                 for dirpath, _, filenames in os.walk(POSTGRESQL_DATA_PATH):
-                    shutil.chown(dirpath, POSTGRESQL_USER, POSTGRESQL_GROUP)
-                    for filename in filenames:
-                        shutil.chown(os.path.join(dirpath, filename), POSTGRESQL_USER, POSTGRESQL_GROUP)
+                    try:
+                        shutil.chown(dirpath, POSTGRESQL_USER, POSTGRESQL_GROUP)
+                        for filename in filenames:
+                            shutil.chown(os.path.join(dirpath, filename), POSTGRESQL_USER, POSTGRESQL_GROUP)
+                    except Exception as e:  # pylint: disable=broad-except
+                        self.log.warning('Failed fixing ownership of PostgreSQL data: {}'.format(e))
+                        return


repos/system_upgrade/el7toel8/actors/satellite_upgrade_data_migration/actor.py
```diff
@@ -38,11 +38,10 @@ def process(self):
                 os.unlink(service)
 
         if facts.postgresql.local_postgresql and os.path.exists(POSTGRESQL_SCL_DATA_PATH):
-            # remove empty PostgreSQL data from the package
-            if os.path.exists(POSTGRESQL_DATA_PATH):
-                os.rmdir(POSTGRESQL_DATA_PATH)
+            # we can assume POSTGRESQL_DATA_PATH exists and is empty
             # move PostgreSQL data to the new home
-            shutil.move(POSTGRESQL_SCL_DATA_PATH, POSTGRESQL_DATA_PATH)
+            for item in glob.glob(os.path.join(POSTGRESQL_SCL_DATA_PATH, '*')):
```
ekohl, line 43: Does this also cover hidden files?

ekohl (commented):
Is there a performance difference in this? Probably not noticeable.

repos/system_upgrade/el7toel8/actors/satellite_upgrade_data_migration/actor.py
evgeni: I don't think it does, but I'd also not expect any hidden files in that directory.

evgeni:
> Is there a performance difference in this? Probably not noticeable.

There are 29 "items" on my test installation, so I don't think it'd be too bad moving them one by one, but I also didn't benchmark it.

repos/system_upgrade/el7toel8/actors/satellite_upgrade_data_migration/actor.py
ekohl: I thought I had seen some files there in the past, like shell history files but couldn't find them now on my existing install. Then it would also have been a directory higher so :+1: from me.

repos/system_upgrade/el7toel8/actors/satellite_upgrade_data_migration/actor.py
```diff
@@ -38,11 +38,10 @@ def process(self):
                 os.unlink(service)
 
         if facts.postgresql.local_postgresql and os.path.exists(POSTGRESQL_SCL_DATA_PATH):
-            # remove empty PostgreSQL data from the package
-            if os.path.exists(POSTGRESQL_DATA_PATH):
-                os.rmdir(POSTGRESQL_DATA_PATH)
+            # we can assume POSTGRESQL_DATA_PATH exists and is empty
```
ekohl, line 41: Can we? What makes it so that we can assume it? A pre check?

evgeni:
```console
# rpm -qf /var/lib/pgsql/data/
postgresql-server-12.11-2.module_el8.6.0+1153+eb826827.x86_64
```

And we install the package.

As for empty:
https://github.com/oamg/leapp-repository/blob/65fec1995d791cee8cf59e4b2ec1cbccdc4e8757/repos/system_upgrade/el7toel8/actors/satellite_upgrade_check/libraries/satellite_upgrade_check.py#L8-L21

ekohl: Doesn't that sort of guarantee it does NOT exist? I'd expect a common solution for users will be `rm -rf /var/lib/pgsql/data` as that would remove the inhibitor.

repos/system_upgrade/el7toel8/actors/satellite_upgrade_data_migration/actor.py
evgeni: The home directory of the `postgres` user is `/var/lib/pgsql`, so if at all, things would end up there, not in `data`

repos/system_upgrade/el7toel8/actors/satellite_upgrade_data_migration/actor.py
evgeni: That's a good point!

evgeni: Actually, no, this is still true.

If the user has an EL7 install that started with PostgreSQL 9.2 in `/var/lib/pgsql` and then was migrated to PostgreSQL 12 in `/var/opt/rh/rh-postgresql12/lib/pgsql` and the user plain `rm -rf /var/lib/pgsql` or `rm -rf /var/lib/pgsql/data` to clean up the old cruft, the installation of PostgreSQL 12 from the EL8 module will re-recreate all these folders for us.

ekohl: Not a 100% reproducer, but close enough to confirm that as long as we (re)install `postgresql-server`, the directory should be present:
```console
# rpm -qf /var/lib/pgsql/data
postgresql-server-12.11-2.module_el8.6.0+1153+eb826827.x86_64
# rm -rf /var/lib/pgsql/data
# dnf reinstall postgresql-server -y
# ls -d /var/lib/pgsql/data
/var/lib/pgsql/data
```

vinzenz (changes requested):
Please do harden all the calls that can throw exceptions.
The main problem is that the upgrade is interrupted because the shutil.move code, for example raises an exception.

You must not leave exception leave the actor. Also do _NOT_ raise StopActorExecutionError, you can raise the StopActorExcecution exception to bail out without having the upgrade fail and end up stuck in the initramdisk.

Even if this fails, yeah the execution of the satellite installer might fail later, and postgres isn't starting - but it is much easier recoverable than having a system being stuck in an interrupted upgrade.

So please do guard all the os.rmdir, shutil.chown, shutil.move calls with try/except blocks.

repos/system_upgrade/el7toel8/actors/satellite_upgrade_data_migration/actor.py
```diff
@@ -35,16 +35,27 @@ def process(self):
         # disable services, will be re-enabled by the installer
         for service_name in SERVICES_TO_DISABLE:
             for service in glob.glob(os.path.join(SYSTEMD_WANTS_BASE, '{}.service'.format(service_name))):
-                os.unlink(service)
+                try:
+                    os.unlink(service)
+                except Exception as e:
+                    self.log.warning('Failed disabling service {}: {}'.format(service, e))
```
evgeni, line 41: I didn't `return` here, as failing to disable those services won't fail the upgrade (they will just try to start before the installer deployed the new config, failing and producing useless errors in logs)

pirat89 (commented):
waiting for the finished testing

lpramuk:
I finished the testing successfully, the patch is resolving the issue

pirat89 (approved):
merging

PR #936, opened 2022-08-08
Draft: Enable 8.7->9.1 upgrade

fernflower:
Versions mapping has to be updated in order
to allow 9.1 to be specified with --target.


fernflower:
@mm0ran this should unblock your testing

pirat89 (changes requested):
@mm0ran  @fernflower we should not include this change as upgrade to 9.1 will not be supported and the `--target` option should allow to specify only supported target releases. For the testing with unsupported target releases it's expected to use `LEAPP_DEVEL_TARGET_RELEASE` envar

PR #937, opened 2022-08-09
Fix swap page size during el8toel9 upgrade on aarch64

matejmatuska:
The kernel in RHEL8 uses 64k page size, but this is changed in RHEL9 to
4k. This breaks SWAP and it must be reinitialized.

The reinitialization is done in initramfs before first boot into the new
system as on some systems SWAP is essential to boot successfully.

The `swapon` command used for reinitialization internally uses
`mkswap` and both of these commands are unavailable in the dracut shell.
As a workaround we can use `$NEWROOT/usr/sbin/swapon` and
`$NEWROOT/usr/sbin/mkswap`. For `swapon` to find `mkswap` the `PATH` must be
temporarily adjusted.

Jira ref: OAMG-6382

### DIFF
--- a/repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
+++ b/repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
@@ -228,6 +228,20 @@ do_upgrade() {
         #PY_LEAPP_PATH=/usr/lib/python2.7/site-packages/leapp/
         #$NEWROOT/bin/systemd-nspawn $NSPAWN_OPTS -D $NEWROOT -E PYTHONPATH="${PYTHONPATH}:${PY_LEAPP_PATH}" /usr/bin/python3 $LEAPPBIN upgrade --resume $args
 
+        # on aarch64 systems during el8 to el9 upgrades the swap is broken due to change in page size (64K to 4k)
+        # adjust the page size before booting into the new system, as it is possible the swap is necessary for to boot
+        # `arch` command is not available in the dracut shell, using uname -m instead
+        [ "$(uname -m)" = "aarch64" -a "$RHEL_OS_MAJOR_RELEASE" = "9" ] && {
+            cp -aS ".leapp_bp" $NEWROOT/etc/fstab /etc/fstab
+            # swapon internally uses mkswap and both swapon and mkswap aren't available in dracut shell
+            # as a workaround we can use the one from $NEWROOT in $NEWROOT/usr/sbin
+            # for swapon to find mkswap we must temporarily adjust the PATH
+            # NOTE: we want to continue the upgrade even when the swapon command fails as users can fix it
+            # manually later. It's not a major blocker.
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -228,6 +228,18 @@ do_upgrade() {
...
 
+        # on aarch64 systems during el8 to el9 upgrades the swap is broken due to change in page size (64K to 4k)
+        # adjust the page size before booting into the new system, as it is possible the swap is necessary for to boot
+        # `arch` command is not available in the dracut shell, using uname -m instead
+        [ "$(uname -m)" = "aarch64" -a "$RHEL_OS_MAJOR_RELEASE" = "9" ] && {
+            cp -aS ".leapp_bp" $NEWROOT/etc/fstab /etc/fstab
+            # swapon internally uses mkswap and both swapon and mkswap aren't available in dracut shell
+            # as a workaround we can use the one from $NEWROOT in $NEWROOT/usr/sbin
+            # for swapon to find mkswap we must temporarily adjust the PATH
+            PATH="$PATH:${NEWROOT}/usr/sbin/" swapon -af || echo "Failed fixing swap page size"
```
pirat89, line 239:
```suggestion
            # NOTE: we want to continue the upgrade even when the swapon command fails as users can fix it
            # manually later. It's not a major blocker.
            PATH="$PATH:${NEWROOT}/usr/sbin/" swapon -af || echo >&2 "Error: Failed fixing swap page size. Manual action is required after the upgrade."
```

Rezney (approved):
Thanks for your patch. Works fine. There is an issue when users use file-based swap as when we copy the fstab into the initramdisk, the path to the file is already different. E.g /spawfile vs. $SYSROOT/swapfile. Anyway, we may deal with that later.

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -228,6 +228,20 @@ do_upgrade() {
...
+        # `arch` command is not available in the dracut shell, using uname -m instead
+        [ "$(uname -m)" = "aarch64" -a "$RHEL_OS_MAJOR_RELEASE" = "9" ] && {
+            cp -aS ".leapp_bp" $NEWROOT/etc/fstab /etc/fstab
+            # swapon internally uses mkswap and both swapon and mkswap aren't available in dracut shell
+            # as a workaround we can use the one from $NEWROOT in $NEWROOT/usr/sbin
+            # for swapon to find mkswap we must temporarily adjust the PATH
+            # NOTE: we want to continue the upgrade even when the swapon command fails as users can fix it
+            # manually later. It's not a major blocker.
+            PATH="$PATH:${NEWROOT}/usr/sbin/" swapon -af || echo >&2 "Error: Failed fixing the swap page size. Manual action is required after the upgrade."
+            mv /etc/fstab.leapp_bp /etc/leapp
```
matejmatuska, line 242: I just noticed a bug, the destination should be `/etc/fstab` not `/etc/leapp`. 
The whole command should look like this: `mv /etc/fstab.leapp_bp /etc/fstab`

PR #938, opened 2022-08-09
Stop debug kernelopt from getting to target

MichalHe:
This patch:
- cleans up the `kernelcmdlineconfig` actor
- adds a new model `TargetKernelCmdlineArgTasks` to allow clear specification of what args to add/remove to the target boot entry
- extends `kernelcmdlineconfig` to be able to also remove modules

Jira ref: OAMG-6918

### DIFF
--- a/repos/system_upgrade/common/actors/addupgradebootentry/libraries/addupgradebootentry.py
+++ b/repos/system_upgrade/common/actors/addupgradebootentry/libraries/addupgradebootentry.py
@@ -3,8 +3,8 @@ import re
 
 from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.common.config import architecture
-from leapp.libraries.stdlib import api, run, CalledProcessError
-from leapp.models import BootContent
+from leapp.libraries.stdlib import api, CalledProcessError, run
+from leapp.models import BootContent, KernelCmdlineArg, TargetKernelCmdlineArgTasks
 
 
 def add_boot_entry(configs=None):
@@ -33,6 +33,14 @@ def add_boot_entry(configs=None):
             # otherwise the new boot entry will not be set as default
             # See https://bugzilla.redhat.com/show_bug.cgi?id=1764306
             run(['/usr/sbin/zipl'])
+
+        if debug:
+            # The kernelopts for target kernel are generated based on the cmdline used in the upgrade initramfs,
+            # therefore, if we enabled debug above, and the original system did not have the debug kernelopt, we
+            # need to explicitly remove it from the target os boot entry.
+            # NOTE(mhecko): This will also unconditionally remove debug kernelopt if the source system used it.
+            api.produce(TargetKernelCmdlineArgTasks(to_remove=[KernelCmdlineArg(key='debug')]))
+
     except CalledProcessError as e:
         raise StopActorExecutionError(
            'Cannot configure bootloader.',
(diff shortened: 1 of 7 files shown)


Rezney:
@MichalHe Thanks for the work. I tested it in both Vagrant and Azure (as one of the related functionality was affected) environments and it seems to be working fine. I will leave the main review for @pirat89

Rezney:
Btw, is the "draft" state still necessary?

MichalHe:
@Rezney thanks, I have forgotten to mark the PR as ready to be reviewed.

repos/system_upgrade/common/actors/addupgradebootentry/libraries/addupgradebootentry.py
```diff
@@ -33,6 +33,12 @@ def add_boot_entry(configs=None):
             # otherwise the new boot entry will not be set as default
             # See https://bugzilla.redhat.com/show_bug.cgi?id=1764306
             run(['/usr/sbin/zipl'])
+
+        if debug:
+            # The kernelopts for target kernel are generated based on the cmdline used in the upgrade initramfs,
+            # therefore, if we enabled debug above, we need to explicitly remove it from the boot entry.
+            api.produce(TargetKernelCmdlineArgTasks(to_remove=[KernelCmdlineArg(key='debug')]))
```
pirat89, line 40:
```suggestion
            # NOTE: the current behaviour drops the debug option even in case it is set on the original system
            # already. We do not check original 'persistent' kernel args and when speaking about the debug
            # option, it would be pretty rare and not so reasonable to have debug option enabled on production
            # systems so we are ok to drop it always when leapp is executed with --debug.
            api.produce(TargetKernelCmdlineArgTasks(to_remove=[KernelCmdlineArg(key='debug')]))
```

pirat89 (commented):
I would add the one NOTE (comment above) to the code, but it looks good otherwise

MichalHe:
@pirat89 Thanks for pointing that out. Since we have the information about used `kernelopts`, I went ahead and made the solution more robust.

repos/system_upgrade/common/actors/addupgradebootentry/libraries/addupgradebootentry.py
```diff
@@ -33,6 +33,26 @@ def add_boot_entry(configs=None):
...
+                # Highly unlikely - something is wrong.
+                raise StopActorExecutionError('Actor did not receive any KernelCmdline messages.')
+            if next(source_kernel_cmdline_msgs, None):
+                api.current_logger().warning('Received more than one KernelCmdline message.')
+
+            has_source_debug_kernelopt = any(
+                (kernelopt.key == 'debug' for kernelopt in source_kernel_cmdline.parameters)
+            )
+
+            if not has_source_debug_kernelopt:
```
pirat89, line 50: @MichalHe  `KernelCmdline` contain just kernel args that have been set for the current system run, from `/proc/cmdline`. Someone could add the `debug` option during the boot and in such case I do not think it's expected in such a case preserve the option. For that, we would need to parse option for the current boot entry, not frmo the current boot. Which I do not think it's reasonable to do now. I think we should go with the previous solution and comment. wdyt?

pirat89 (approved):
lgtm. @Rezney is finishing the testing so I keep the merge on him.

PR #942, opened 2022-08-15
Move blacklistrepositories to common

MichalHe:
The optional repositories were not blacklisted since the repositories_blacklist actor resided only in el7toel8/. This patch moves it to common/, enabling the necessary repositories blacklisting also on the 8to9 upgrade path.


Rezney (commented):
I cannot point to a section that has not been changed but I suggest tuning this wording:

~~~
def _report_excluded_repos(repos):
    api.current_logger().info(
        "The optional repository is not enabled. Excluding %r "
        "from the upgrade",
        repos,
    )
~~~

Let's either add "The optional / CRB  repository is not enabled" string or a condition depending on the source version. Otherwise, the plural "optional repositories" should be ok.

MichalHe:
@Rezney Fixed by adding a condition.

pirat89 (dismissed):
lgtm. untested but I think it should work. keeping unmerged yet for the testing purposes.

pirat89:
@MichalHe @Rezney this is supposed to go into the upcoming release, right?

MichalHe:
@pirat89 As far as I know it definitely should go there. We have already seen a system where leapp tried to enable CRB, although it was disabled on the original system.  It wasn't like the system had something from CRB installed and then the repository was disabled, it was due to some split event, where one of the resulting packages resided in CRB.

MichalHe:
leapp-ci build

Rezney (approved):

PR #945, opened 2022-08-17
Detect missing new line in /etc/default/grub and fix it

PeterMocary:
The missing new line at the end of ``/etc/default/grub`` causes the IPU to fail on
reboot. Detecting and fixing this problem before the upgrade is just a temporary
solution and it will be addressed in grub2 package instead, later on.

JIRA ref: OAMG-4535

### DIFF
--- a/repos/system_upgrade/common/actors/detectmissingnewlineingrubcfg/actor.py
+++ b/repos/system_upgrade/common/actors/detectmissingnewlineingrubcfg/actor.py
@@ -1,0 +1,35 @@
+from leapp import reporting
+from leapp.actors import Actor
+from leapp.libraries.actor.detectmissingnewlineingrubcfg import is_grub_config_missing_final_newline
+from leapp.models import GrubConfigError
+from leapp.reporting import create_report, Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class DetectMissingNewlineInGrubCfg(Actor):
+    """
+    Check the grub configuration for a missing newline at its end.
+    """
+
+    name = 'detect_missing_newline_in_grub_cfg'
+    consumes = ()
+    produces = (Report, GrubConfigError)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        config = '/etc/default/grub'
+        if is_grub_config_missing_final_newline(config):
+            create_report([
+                reporting.Title('Detected a missing newline at the end of grub configuration file.'),
+                reporting.Summary(
+                    'The missing newline in /etc/default/grub causes booting issues when appending '
+                    'new entries to this file during the upgrade. Leapp will automatically fix this '
+                    'problem by appending the missing newline to the grub configuration file.'
+                ),
+                reporting.Severity(reporting.Severity.LOW),
+                reporting.Tags([reporting.Tags.BOOT]),
+                reporting.RelatedResource('file', config)
+            ])
+
+            config_error = GrubConfigError(error_detected=True, error_type='missing newline')
+            self.produce(config_error)
--- a/repos/system_upgrade/common/actors/detectmissingnewlineingrubcfg/libraries/detectmissingnewlineingrubcfg.py
+++ b/repos/system_upgrade/common/actors/detectmissingnewlineingrubcfg/libraries/detectmissingnewlineingrubcfg.py
@@ -1,0 +1,13 @@
+import os
+
+
+def _get_config_contents(config_path):
+    if os.path.isfile(config_path):
+        with open(config_path, 'r') as config:
+            return config.read()
+    return ''
+
+
+def is_grub_config_missing_final_newline(conf_file):
+    config_contents = _get_config_contents(conf_file)
+    return config_contents != '' and config_contents[-1] != '\n'
--- a/repos/system_upgrade/common/models/grubconfigerror.py
+++ b/repos/system_upgrade/common/models/grubconfigerror.py
@@ -1,4 +1,4 @@
-from leapp.models import Model, fields
+from leapp.models import fields, Model
 from leapp.topics import SystemFactsTopic
 
 
@@ -6,3 +6,4 @@ class GrubConfigError(Model):
     topic = SystemFactsTopic
 
     error_detected = fields.Boolean(default=False)
+    error_type = fields.StringEnum(['GRUB_CMDLINE_LINUX syntax', 'missing newline'])
--- a/repos/system_upgrade/el7toel8/actors/detectgrubconfigerror/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/detectgrubconfigerror/actor.py
@@ -1,9 +1,9 @@
+from leapp import reporting
 from leapp.actors import Actor
 from leapp.libraries.actor.scanner import detect_config_error
 from leapp.libraries.common.config import architecture
 from leapp.models import GrubConfigError
-from leapp.reporting import Report, create_report
-from leapp import reporting
+from leapp.reporting import create_report, Report
 from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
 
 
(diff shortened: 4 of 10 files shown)


repos/system_upgrade/common/actors/detectmissingnewlineingrubcfg/actor.py
```diff
@@ -0,0 +1,37 @@
...
+    def process(self):
+        config = '/etc/default/grub'
+        error_detected = detect_config_error(config)
+        if error_detected:
+            create_report([
+                reporting.Title('Detected a missing new line at the end of grub configuration file.'),
+                reporting.Summary(
+                    'The missing new line in /etc/default/grub causes booting issues when appending '
+                    'new entries to this file during the upgrade. '
+                    'Error is automatically fixed by add_upgrade_boot_entry actor.'
```
MichalHe, line 28: Users don't care, nor now, what actor will do it. 
```suggestion
                    'Leapp will automatically fix this problem by appending the missing newline to the grub configuration file.'
```

repos/system_upgrade/common/actors/detectmissingnewlineingrubcfg/actor.py
```diff
@@ -0,0 +1,37 @@
...
+                ),
+                reporting.Severity(reporting.Severity.LOW),
+                reporting.Tags([reporting.Tags.BOOT]),
+                reporting.RelatedResource('file', config)
+            ])
+
+        config_error = GrubConfigError(error_detected=error_detected,
+                                       error_description='missing LF')
+        self.produce(config_error)
```
MichalHe, line 37: The design of the model does not make much sense. If there is no problem with configuration, don't produce anything.
```suggestion
        if error_detected:
            config_error = GrubConfigError(error_detected=error_detected, error_description='missing LF')
            self.produce(config_error)
```

repos/system_upgrade/common/actors/detectmissingnewlineingrubcfg/libraries/detectmissingnewlineingrubcfg.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def _get_config_contents(config_path):
+    if os.path.isfile(config_path):
+        with open(config_path, 'r') as config:
+            return config.read()
+    return ''
+
+
+def detect_config_error(conf_file):
```
MichalHe, line 11: Function name tells almost nothing about what the error should be.
```suggestion
def is_grub_config_missing_final_newline(conf_file):
```
There is only a small chance that a 2 line function cannot be precisely described by its name. After renaming, the docstrings can be dropped:
- the return value (boolean) is hinted by the new function name - `is` implies that it is a boolean
- what the error is is written in the function name

repos/system_upgrade/common/actors/detectmissingnewlineingrubcfg/libraries/detectmissingnewlineingrubcfg.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    Check grub configuration for missing new line at its end.
+
+    :return: Function returns True if error was detected, otherwise False.
+    """
+
+    config = _get_config_contents(conf_file)
+    return config != '' and config[-1] != '\n'
```
MichalHe, line 19:
```suggestion
    config_contents = _get_config_contents(conf_file)
    return config_contents and config_contents[-1] != '\n'
```

repos/system_upgrade/common/models/grubconfigerror.py
```diff
@@ -1,8 +1,9 @@
-from leapp.models import Model, fields
+from leapp.models import fields, Model
 from leapp.topics import SystemFactsTopic
 
 
 class GrubConfigError(Model):
     topic = SystemFactsTopic
 
     error_detected = fields.Boolean(default=False)
+    error_description = fields.StringEnum(['GRUB_CMDLINE_LINUX syntax', 'missing LF'])
```
MichalHe, line 9: It is not description per say. I would expect description to provide more information describing the actual error and why it is an error. Using StringEnum was a correct choice :+1:, and it also hints that the field is trying to categorize the errors (e.g. tells what kind/type of error it is) rather than describe what the error is.
```suggestion
    error_type = fields.StringEnum(['GRUB_CMDLINE_LINUX syntax', 'missing LF'])
```

repos/system_upgrade/el7toel8/actors/detectgrubconfigerror/actor.py
```diff
@@ -37,4 +37,6 @@ def process(self):
                 reporting.RelatedResource('file', config)
             ])
 
-        self.produce(GrubConfigError(error_detected=error_detected))
+        config_error = GrubConfigError(error_detected=error_detected,
+                                       error_description='GRUB_CMDLINE_LINUX syntax')
+        self.produce(config_error)
```
MichalHe, line 42: I would advocate to produce this conditionally as well. Please, check that whoever consumes grub errors checks whether  the error_detected is True. If so, the message can be produced conditionally.

repos/system_upgrade/common/actors/detectmissingnewlineingrubcfg/actor.py
MichalHe: I've noticed that this message is copied from the other error detecting actor, however, my original points still stands. It is unlikely that users know how is leapp structured internally.

repos/system_upgrade/common/actors/detectmissingnewlineingrubcfg/actor.py
```diff
@@ -0,0 +1,37 @@
...
+from leapp.actors import Actor
+from leapp.libraries.actor.detectmissingnewlineingrubcfg import detect_config_error
+from leapp.models import GrubConfigError
+from leapp.reporting import create_report, Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class DetectMissingNewLineInGrubCfg(Actor):
+    """
+    Check grub configuration for missing new line at its end.
```
MichalHe, line 11:
```suggestion
    Check the grub configuration for a missing new line at its end.
```

MichalHe (changes requested):

repos/system_upgrade/common/actors/detectmissingnewlineingrubcfg/actor.py
PeterMocary: You are right, the boolean is redundant and should be removed.

repos/system_upgrade/common/models/grubconfigerror.py
```diff
@@ -1,8 +1,8 @@
-from leapp.models import Model, fields
+from leapp.models import fields, Model
 from leapp.topics import SystemFactsTopic
 
 
 class GrubConfigError(Model):
     topic = SystemFactsTopic
 
-    error_detected = fields.Boolean(default=False)
```
MichalHe, line 8: We cannot do this. Models are part of our stable API and they undergo deprecation process. We need to keep the original field, including its semantics, so if an user relied on this model, he will still find its original fields intact. I know that I have commented on the strange nature of this model and its meaning in our IRC communication, however, I did not mean to imply that you should fix the illogical design. Lets keep the original field intact, and still populate with True, whenever an error was encountered, so we do not break our API for other users.

That being said, I really, really like the pragmatic approach you have shown here - its a problem, so why not solve it right away since I am touching the code anyway.

MichalHe (changes requested):

repos/system_upgrade/common/models/grubconfigerror.py
PeterMocary: Thanks, for the explanation. I didn't realize this although it makes sense.

MichalHe (approved):

MichalHe:
I have tested that the patch correctly detects a missing newline and fixes the config, thus the upgrade does not crash when a newline is missing in the config on the source system.

PR #950, opened 2022-08-22
Dont propragate enforcing=0 kernel arg to target system

matejmatuska:
The upgrade initramdisk has `enforcing=0` kernel arg set to disable
SELinux. After the upgrade these options are used for the target
kernel cmdline. However we don't want `enforcing=0` set on the target system.

Jira ref: OAMG-7452


vinzenz:
LGTM

MichalHe (approved):
I have tested both `7>8` and `8>9`, the patch correctly prevents propagation of `enforcing=0` to the target system.

MichalHe:
I will wait for tests to finish, so there are objective data that the patch did not break anything, and then I will merge it.

PR #951, opened 2022-08-23
Add codespell GitHub actions workflow for spell checking code

matejmatuska:
Use [codespell](https://github.com/codespell-project/actions-codespell) for spell checking code in PRs and pushes to master.

All existing typos are fixed or ignored in the following way:
- Files without any words to check are skipped, as they would just
  create false positives.
- Words like `ro` and `fo` are ignored since they are pretty common in
  the codebase and we don't want to trouble contributors with excluding
  every line containing these from spell checking. It's up to
  maintainers to spot these false negatives.

Jira ref: OAMG-7458

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,23 @@
+name: Codespell
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  codespell:
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: codespell-project/actions-codespell@master
+        with:
+          ignore_words_list: ro,fo
+          skip: "./repos/system_upgrade/common/actors/storagescanner/tests/files/mounts,\
+            ./repos/system_upgrade/el7toel8/actors/networkmanagerreadconfig/tests/files/nm_cfg_file_error,\
+            ./repos/system_upgrade/common/actors/scancpu/tests/files/lscpu_s390x"
--- a/.github/workflows/pr-welcome-msg.yml
+++ b/.github/workflows/pr-welcome-msg.yml
@@ -19,7 +19,7 @@ jobs:
           issue-number: ${{ github.event.pull_request.number }}
           body: |
             ## **Thank you for contributing to the Leapp project!**
-            Please note that every PR needs to comply with the [Leapp Guidelines](https://leapp.readthedocs.io/en/latest/contributing.html#) and must pass all tests in order to be mergable.
+            Please note that every PR needs to comply with the [Leapp Guidelines](https://leapp.readthedocs.io/en/latest/contributing.html#) and must pass all tests in order to be mergeable.
             If you want to request a review or rebuild a package in copr, you can use following commands as a comment:
             - **review please** to notify leapp developers of review request
             - **/packit copr-build** to submit a public copr build using packit
--- a/.github/workflows/reuse-copr-build.yml
+++ b/.github/workflows/reuse-copr-build.yml
@@ -35,7 +35,7 @@ jobs:
           echo "::set-output name=pr_nr::${PR_URL##*/}"
 
       - name: Checkout
-        # TODO: The correct way to checkout would be to use simmilar approach as in get_commit_by_timestamp function of
+        # TODO: The correct way to checkout would be to use similar approach as in get_commit_by_timestamp function of
         #       the github gluetool module (i.e. do not use HEAD but the last commit before comment).
         id: checkout
         uses: actions/checkout@v2
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ endif
 # someone will call copr_build without additional parameters
 MASTER_BRANCH=master
 
-# In case the PR or MR is defined or in case build is not comming from the
+# In case the PR or MR is defined or in case build is not coming from the
 # MATER_BRANCH branch, N_REL=0; (so build is not update of the approved
 # upstream solution). For upstream builds N_REL=100;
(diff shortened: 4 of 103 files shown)


fernflower:
That's a nice idea! Do you have results of a sample run? Won't it complain about existing spelling errors on files not touched by the PR?

matejmatuska:
@fernflower I have sample run in [my fork]( https://github.com/matejmatuska/leapp-repository/runs/7971474436?check_suite_focus=true).

matejmatuska:
The action is currently blocked by the organization policy, so that needs to be sorted out first.

fernflower:
@matejmatuska Nice, not so many errors as I expected. But still, if you enable it as is (checking all code in master and not only files touched by the PR), it will be failing all the time unless we fix the violations right now (and fixing the violations requires changes to many actors and tests probably, the ones relying on specific output). Do we plan to do this?

fernflower:
@matejmatuska Please comment the linter run in Makefile temporarily (in a separate commit). We will discuss that tomorrow with @pirat89

.codespell-exclude-lines
```diff
@@ -0,0 +1,34 @@
+            elif getargbool 0 ro; then
```
fernflower, line 1: I am afraid that going this way instead of creating an ignore words list will create the need to edit .codespell-exclude-lines from time to time which from the POV of the contributors may be not worth the frustration and extra work :( I'd say codespell is here to help the maintainers and reviers (us), not the contributors, so we shouldn't put that on them.
I'd say we can live with occasional typos like 'ro' but not with crazy-from-the-contributors-pov workflow. How about we implement a more relaxed solution like https://github.com/protocolbuffers/protobuf/actions/runs/277120884/workflow?

matejmatuska: Wouldn't there will be the need to edit the `ignore_words_list` from time to time? If a contributor opens a PR with a valid word which codespell will recognize as an error, then they would have to adjust that. I created the `codespell-exclude-lines` file using a script, maybe that could be an option.

fernflower: `ignore_words_list` maintenance is leapp's team responsibility and it should be updated a lot less frequently as this solution will create false negatives for spell checking. Also it's an "add once - stays forever". Like with the `fo` example, you will need one rule to cover all further usages of `fo` as a file object.

`.codespell-exclude-lines` moves the responsibility to the contributor and it's an "add once - will fix only one exact occurence" solution. That's why you had to create 3 rules for using `fo`, and every new patch that has `fo` as a file object will need to add another one.

I am really in favor of cutting the contributors some slack here. But if you are unsure - we could definitely put that to the vote, that's what we have cabal meetings for :)

matejmatuska: Like with the `fo` example, if `fo` is in `ignore_words_list` and there is a `fo` typo in a PR, then it is back on reviewers to find that typo, which is why the spell checking is being introduced in the first place. 

However I agree that this would probably happen very rarely.

fernflower: Exactly, it's up to the reviewers - it's our job :) And we should not make it the contributors PITA. 
Also even if we as the reviewers miss it - it won't be the end of the world and nothing will really break.

Don't get me wrong - I absolutely love this patch, coming up with this idea was brilliant and it's a great addition to the existing checks we have. My concerns are only about the perfectness / contribution-friendly tradeoff.

fernflower:
@matejmatuska I think I have figured out a better alternative to commenting out isort check - https://github.com/oamg/leapp-repository/pull/954/
Once this is merged you can rebase your patch and get rid of "[Temporarily comment out linter in Makefile as requested](https://github.com/oamg/leapp-repository/pull/951/commits/da2540e92c98890198e0f4fdcd1075576d8e2a12)" commit.

.codespell-exclude-lines
matejmatuska: I think you are right. It is actually @pirat89's idea, so maybe he could give us his opinion?

fernflower:
PR954 is in, you can rebase now.

.codespell-exclude-lines
matejmatuska: Do you think ignoring the word `ect` because of this single line: `FILES = ['/file1', '/file2', '/dir/ect/ory/file3', '/file4', '/file5']` is a good idea? I agree with the `ro` and `fo` examples, but I am not really sure about this one. I would rather change the path (as it is just a test).

Also similar example on this line: `Flags:                           esan3 zarch stfle msa ldisp eimm dfp edat etf3eh highgprs te vx vxd vxe gs vxe2 vxp sort dflt sie`, the offending word is `sie` which is the only valid occurrence in the entire project.

I think a combination of exclude lines file and `ignore_words`, might be the best solution, what do you think?

fernflower: Yeah, I think having a combination of both is just perfect.

.github/workflows/codespell.yml
```diff
@@ -18,4 +18,6 @@ jobs:
       - uses: codespell-project/actions-codespell@master
         with:
           exclude_file: .codespell-exclude-lines
-          skip: ./repos/system_upgrade/common/actors/storagescanner/tests/files/mounts
+          ignore_words_list: ro,fo
+          skip: "./repos/system_upgrade/common/actors/storagescanner/tests/files/mounts,\
```
fernflower, line 22: Do we really need to skip some files? Or is it a temporary measure until you catch all issues and fixup into the proper comments?

matejmatuska: Yeah we do, for example `./repos/system_upgrade/common/actors/storagescanner/tests/files/mounts` contains just mount points and mount options so there aren't really any words which should be checked. They would just create unnecessary false positives.

fernflower: Thanks or clarification!

fernflower (commented):
I just love this patch, thank for working on this initiative! 
Please squash the commits and I will approve the pr.

fernflower (approved):
LGTM, merging, yay!

PR #954, opened 2022-08-24
Disable isort check for deprecated imports

fernflower:
isort works fine most of the time, but the way it handles
multiline imports with inline comments is not acceptable
to everyone in the team.
So before we implement a solution it was decided to leave
those imports just as the have been for ages, do let's mute
isort import check.


pirat89 (approved):
Thanks! That's probably best solution really. Lgtm.

matejmatuska (approved):
Nice solution! Lgtm

PR #956, opened 2022-08-25
Improve error message when more space is needed for the upgrade

PeterMocary:
When there was not enough space, leapp would output misleading error message propagated from dnf. This error message was replaced and includes a solution article.

JIRA ref: OAMG-7256
RHBZ: 2110045 (RFE)

### DIFF
--- a/repos/system_upgrade/common/libraries/dnfplugin.py
+++ b/repos/system_upgrade/common/libraries/dnfplugin.py
@@ -146,7 +146,8 @@ def backup_debug_data(context):
             api.current_logger().warning('Failed to copy debugdata. Message: {}'.format(str(e)), exc_info=True)
 
 
-def _transaction(context, stage, target_repoids, tasks, plugin_info, test=False, cmd_prefix=None, on_aws=False):
+def _transaction(context, stage, target_repoids, tasks, plugin_info, xfs_info,
+                 test=False, cmd_prefix=None, on_aws=False):
     """
     Perform the actual DNF rpm download via our DNF plugin
     """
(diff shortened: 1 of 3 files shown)


pirat89 (commented):
seems good to me. untested yet

fernflower:
lgtm

fernflower (dismissed):

repos/system_upgrade/common/libraries/dnfplugin.py
```diff
@@ -214,10 +215,25 @@ def _transaction(context, stage, target_repoids, tasks, plugin_info, test=False,
...
+                article_section = 'Generic case'
+                if xfs_info.present and xfs_info.without_ftype:
+                    article_section = 'XFS ftype=0 case'
+
+                message = ('There is not enough space on the file system hosting /var/lib/leapp directory '
+                           'to extract the packages.')
+                details = {'hint': "Please follow the instructions in the '{}' section of the article at: "
+                                   "link: https://access.redhat.com/solutions/5057391".format(article_section)}
+
+            raise StopActorExecutionError(message=message, details=details)
```
fernflower, line 236: It would be great to cover this with a unit test as well

fernflower: On the other hand - looking at the fact that we have zero unit tests for the _transaction and there is a great deal of mocking to do, I think we could live with manual testing + an integration test for this part. WDYT?

PeterMocary: Yes, as discussed on IRC, I agree with integration and manual testing due to the increased complexity of unit testing in this case.

fernflower:
Could you please rebase against master? LGTM, I am ok to merge it once this is done.

PeterMocary:
> Could you please rebase against master? LGTM, I am ok to merge it once this is done.
@fernflower Rebased.

fernflower (approved):

pirat89:
test results seems ok. merging

PR #958, opened 2022-08-26
Ignore external accounts in /etc/passwd

PeterMocary:
The ``/etc/passwd`` can contain special entries to selectively incorporate entries
from another service source such as NIS or LDAP. These entries don't need to
contain all the fields that are normally present in the ``/etc/passwd`` entry and
would cause the upgrade failure in facts phase.

JIRA ref: OAMG-6279
RHBZ: 2031098

### DIFF
--- a/repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
+++ b/repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
@@ -60,13 +60,26 @@ def anyhasprefix(value, prefixes):
 
 @aslist
 def _get_system_users():
+    skipped_user_names = []
     for p in pwd.getpwall():
-        yield User(
-            name=p.pw_name,
-            uid=p.pw_uid,
-            gid=p.pw_gid,
-            home=p.pw_dir
-        )
+        # The /etc/passwd can contain special entries from another service source such as NIS or LDAP. These entries
+        # start with + or - sign and might not contain all the mandatory fields, thus are skipped along with other
+        # invalid entries for now. The UID and GID fields are always defined by pwd to 0 even when not specifiead in
+        # /etc/passwd.
+        if p.pw_name != '' and not p.pw_name.startswith(('+', '-')) and p.pw_dir:
+            yield User(
+                name=p.pw_name,
+                uid=p.pw_uid,
+                gid=p.pw_gid,
+                home=p.pw_dir
+            )
+        else:
+            skipped_user_names.append(p.pw_name)
+
+    if skipped_user_names:
+        api.current_logger().debug("These users from /etc/passwd that are special entries for service "
+                                   "like NIS, or don't contain all mandatory fields won't be included "
+                                   "in UsersFacts: {}".format(skipped_user_names))
 
 
 def get_system_users_status():
@@ -76,12 +89,25 @@ def get_system_users_status():
 
 @aslist
 def _get_system_groups():
+    skipped_group_names = []
     for g in grp.getgrall():
-        yield Group(
-            name=g.gr_name,
-            gid=g.gr_gid,
-            members=g.gr_mem
-        )
+        # The /etc/group can contain special entries from another service source such as NIS or LDAP. These entries
+        # start with + or - sign and might not contain all the mandatory fields, thus are skipped along with other
+        # invalid entries for now. The GID field is always defined by pwd to 0 even when not specifiead in
+        # /etc/group.
+        if g.gr_name != '' and not g.gr_name.startswith(('+', '-')):
+            yield Group(
+                name=g.gr_name,
+                gid=g.gr_gid,
+                members=g.gr_mem
+            )
+        else:
+            skipped_group_names.append(g.gr_name)
+
+    if skipped_group_names:
+        api.current_logger().debug("These groups from /etc/group that are special entries for service "
+                                   "like NIS, or don't contain all mandatory fields won't be included "
+                                   "in GroupsFacts: {}".format(skipped_group_names))
 
 
 def get_system_groups_status():
--- a/repos/system_upgrade/common/actors/systemfacts/tests/test_systemfacts.py
+++ b/repos/system_upgrade/common/actors/systemfacts/tests/test_systemfacts.py
@@ -1,4 +1,11 @@
-from leapp.libraries.actor.systemfacts import anyendswith, anyhasprefix, aslist
+import grp
+import pwd
+
+import pytest
+
+from leapp.libraries.actor.systemfacts import _get_system_groups, _get_system_users, anyendswith, anyhasprefix, aslist
+from leapp.libraries.common.testutils import logger_mocked
+from leapp.libraries.stdlib import api
 from leapp.snactor.fixture import current_actor_libraries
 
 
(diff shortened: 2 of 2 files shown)


PeterMocary:
Should we note somewhere what entries have been skipped or don't skip them at all?

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -61,12 +61,13 @@ def anyhasprefix(value, prefixes):
 @aslist
 def _get_system_users():
     for p in pwd.getpwall():
-        yield User(
-            name=p.pw_name,
-            uid=p.pw_uid,
-            gid=p.pw_gid,
-            home=p.pw_dir
-        )
+        if all([p.pw_name, p.pw_uid, p.pw_gid, p.pw_dir]):
```
MichalHe, line 64: This needs a comment. Under what circumstances can it happen, that some of the fields are None/empty?

PeterMocary: I changed the condition as I found out only ``p.pw_dir`` should be checked. The UID and GID are set to 0 by ``pwd`` when not specified however, 0 is also a valid ID, thus it would cause false negatives. The entries in ``/etc/passwd`` are skipped by ``pwd`` if they don't contain username therefore it will be always defined.

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -60,13 +60,24 @@ def anyhasprefix(value, prefixes):
...
+        # as the name field as entries without it are skipped by pwd.
+        if p.pw_dir:
+            yield User(
+                name=p.pw_name,
+                uid=p.pw_uid,
+                gid=p.pw_gid,
+                home=p.pw_dir
+            )
+        else:
+            logger.debug('The %s user from /etc/passwd will not be included in the UsersFacts '
```
MichalHe, line 79:
```suggestion
            api.current_logger().debug('The %s user from /etc/passwd will not be included in the UsersFacts '
```

MichalHe (changes requested):

MichalHe:
@PeterMocary Hi, I have started reading about NIS-style changes to the usual format of `/etc/passwd` and your assumption that only `pwd` might be empty is not correct:
- https://www.oreilly.com/library/view/linux-network-administrators/1565924002/ch13s08.html (good article)
- https://www.man7.org/linux/man-pages/man5/nsswitch.conf.5.html - see section about compact mode, but I have read it all and it is quite interesting
This patch should handle all possible entries introduced by the compact mode, therefore, basically ignore anything starting with `+`

MichalHe (changes requested):
See my previous comment. Please, make sure that we also handle (ignore) a standalone `+`, which is a valid syntax - I am not sure how such entry will be parsed.

PeterMocary:
> Hi, I have started reading about NIS-style changes to the usual format of `/etc/passwd` and your assumption that only `pwd` might be empty is not correct.
Even though the `/etc/passwd` can contain entries which have empty fields, the `pwd` library, used for the extraction of these entries, assigns default values to **some** of the fields when they are empty in `/etc/passwd`. Based on behavior of this library I decided to limit the check to only to `pw_dir` or that the `home` field of our `User` model isn't filled with `None`. I tested the behavior by hand however, which isn't the best way thus I missed some cases where the `pw_name` can also be empty and the entry is extracted by `pwd` library. This means that **`pw_name` should be checked too**.

>This patch should handle all possible entries introduced by the compact mode, therefore, basically ignore anything starting with +
The issue here is about the fact that the entries in `User` model aren't nullable and therefore their values need to be checked before assignment.  This was just pointed out by the specific `+` or `-` entries mentioned in this issue, which will be skipped from now on whenever necessary fields don't contain a value or aren't assigned one by the `pwd` library. Do you think that all of these entries should be skipped?

The same problem as in with the `User` model could happen for the `Group` model as its fields aren't nullable again and the values assigned to them aren't checked.

MichalHe:
# Tested in the following scenarios
## /etc/passwd
- `+user`
  - preupgrade does not fail
  - These users from /etc/passwd that are special entries for service like NIS, or don't contain all mandatory fields won't be included in UsersFacts: ['+someuser']
- `+@usernetgroup`
  - preupgrade does not fail
  - These users from /etc/passwd that are special entries for service like NIS, or don't contain all mandatory fields won't be included in UsersFacts: ['+@usernetgroup']
- `-someuser`
  - preupgrade does not fail
  - These users from /etc/passwd that are special entries for service like NIS, or don't contain all mandatory fields won't be included in UsersFacts: ['-someuser']
- `+@someuser::::::/bin/bash`
  - preupgrade does not fail
  - corresponding entry in `leapp-preupgrade.log`: These users from /etc/passwd that are special entries for service like NIS, or don't contain all mandatory fields won't be included in UsersFacts: ['+@someuser']
- `+`
  - preupgrade does not fail
  - corresponding entry in `leapp-preupgrade.log`: These users from /etc/passwd that are special entries for service like NIS, or don't contain all mandatory fields won't be included in UsersFacts: ['+']

## /etc/group
- `+nisgroup`
  - preupgrade does not fail
  - corresponding entry in `leapp-preupgrade.log`: These groups from /etc/passwd that are special entries for service like NIS, or don't contain all mandatory fields won't be included in GroupsFacts: ['+nisgroup']
- `-nisgroup`
  - preupgrade does not fail
  - corresponding entry in `leapp-preupgrade.log`:  These groups from /etc/passwd that are special entries for service like NIS, or don't contain all mandatory fields won't be included in GroupsFacts: ['-nisgroup']
- `+`
  - preupgrade does not fail
  - corresponding entry in `leapp-preupgrade.log`: These groups from /etc/passwd that are special entries for service like NIS, or don't contain all mandatory fields won't be included in GroupsFacts: ['+']

Testing was successful.

MichalHe (changes requested):

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -60,13 +60,26 @@ def anyhasprefix(value, prefixes):
...
+                gid=p.pw_gid,
+                home=p.pw_dir
+            )
+        else:
+            skipped_user_names.append(p.pw_name)
+
+    if skipped_user_names:
+        api.current_logger().debug("These users from /etc/passwd that are special entries for service "
+                                   "like NIS, or don't contain all mandatory fields won't be included "
+                                   "in UsersFacts: {}".format(skipped_user_names))
```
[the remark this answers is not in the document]
PeterMocary: List encloses its items into single quotes by default, which helps identify empty user names or group names and given that most of the other debug entries use it as well, I don't think this change is necessary.

MichalHe (dismissed):

fernflower (commented):
LGTM, just one thing - could you add a unit test for _get_system_users, please?

PeterMocary:
> LGTM, just one thing - could you add a unit test for _get_system_users, please?

Just added the unit test for ``_get_system_users`` and also a similar unit test for ``_get_system_groups``.

fernflower (approved):
lgtm, will wait for tests and then merge. Great job!

PR #960, opened 2022-08-29
[revived] Leapp initram networking poc

fernflower:
Rebase of https://github.com/oamg/leapp-repository/pull/583/ against master

### DIFF
--- a/repos/system_upgrade/common/actors/addupgradebootentry/libraries/addupgradebootentry.py
+++ b/repos/system_upgrade/common/actors/addupgradebootentry/libraries/addupgradebootentry.py
@@ -9,7 +9,8 @@ from leapp.models import BootContent, KernelCmdlineArg, TargetKernelCmdlineArgTa
 
 def add_boot_entry(configs=None):
     debug = 'debug' if os.getenv('LEAPP_DEBUG', '0') == '1' else ''
-
+    enable_network = os.getenv('LEAPP_INITRAM_NETWORK') in ('network-manager', 'scripts')
+    ip_arg = ' ip=on' if enable_network else ''
     kernel_dst_path, initram_dst_path = get_boot_file_paths()
     _remove_old_upgrade_boot_entry(kernel_dst_path, configs=configs)
     try:
@@ -20,7 +21,7 @@ def add_boot_entry(configs=None):
             '--title', 'RHEL-Upgrade-Initramfs',
             '--copy-default',
             '--make-default',
-            '--args', '{DEBUG} enforcing=0 rd.plymouth=0 plymouth.enable=0'.format(DEBUG=debug)
+            '--args', '{DEBUG}{NET} enforcing=0 rd.plymouth=0 plymouth.enable=0'.format(DEBUG=debug, NET=ip_arg)
         ]
         if configs:
             for config in configs:
--- a/repos/system_upgrade/common/actors/addupgradebootentry/libraries/addupgradebootentry.py
+++ b/repos/system_upgrade/common/actors/addupgradebootentry/libraries/addupgradebootentry.py
@@ -9,8 +9,8 @@ from leapp.models import BootContent, KernelCmdlineArg, TargetKernelCmdlineArgTa
 
 def add_boot_entry(configs=None):
     debug = 'debug' if os.getenv('LEAPP_DEBUG', '0') == '1' else ''
-    enable_network = os.getenv('LEAPP_INITRAM_NETWORK') in ('network-manager', 'scripts')
-    ip_arg = ' ip=on' if enable_network else ''
+    enable_network = os.getenv('LEAPP_DEVEL_INITRAM_NETWORK') in ('network-manager', 'scripts')
+    ip_arg = ' ip=dhcp rd.neednet=1' if enable_network else ''
     kernel_dst_path, initram_dst_path = get_boot_file_paths()
     _remove_old_upgrade_boot_entry(kernel_dst_path, configs=configs)
     try:
--- a/repos/system_upgrade/common/actors/checknfs/actor.py
+++ b/repos/system_upgrade/common/actors/checknfs/actor.py
@@ -1,5 +1,6 @@
 from leapp import reporting
 from leapp.actors import Actor
+from leapp.libraries.common.config import get_env
 from leapp.models import StorageInfo
 from leapp.reporting import create_report, Report
 from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
@@ -18,6 +19,9 @@ class CheckNfs(Actor):
     tags = (ChecksPhaseTag, IPUWorkflowTag,)
 
     def process(self):
+        # if network in initramfs is enabled NFS inhibitors are redundant
+        if get_env('LEAPP_INITRAM_NETWORK', None):
+            return
         details = "NFS is currently not supported by the inplace upgrade.\n" \
                   "We have found NFS usage at the following locations:\n"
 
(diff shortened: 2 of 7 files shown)


fernflower:
So I squashed commits.  I have written a regression test that disables nfs removal workarounds and still gets the machine successfully upgraded, I'm pretty happy with the solution at the moment, so calling for review.

repos/system_upgrade/common/actors/addupgradebootentry/libraries/addupgradebootentry.py
```diff
@@ -9,7 +9,8 @@
 
 def add_boot_entry(configs=None):
     debug = 'debug' if os.getenv('LEAPP_DEBUG', '0') == '1' else ''
-
+    enable_network = os.getenv('LEAPP_INITRAM_NETWORK') in ('network-manager', 'scripts')
+    ip_arg = ' ip=on rd.neednet=1' if enable_network else ''
```
pvalena, line 13: I prefer / have seen more often: `ip=dhcp` (also is more explicit on what is supposed to happen = will not change behaviour).

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -194,6 +194,18 @@ ibdmp() {
     done
 }
 
+bring_up_network() {
+    if [ -f /etc/leapp-initram-network-manager ]; then
+        . /lib/dracut/hooks/cmdline/99-nm-config.sh
+        . /lib/dracut/hooks/initqueue/settled/99-nm-run.sh
```
pvalena, line 200: Please note that in general, you'd want these regenerated&executed with updated dracut, which also runs them - otherwise I'd rather use the ones from the updated package:

```
/lib/dracut/modules.d/35network-manager/nm-config.sh
/lib/dracut/modules.d/35network-manager/nm-run.sh
``` 

They also expect `/lib/nm-lib.sh` to exist (should be updated from the same directory).

But you should be able to bring up the system with either.

fernflower: So you'd prefer that we substitute L199-L200 with 
```
/lib/dracut/modules.d/35network-manager/nm-config.sh
/lib/dracut/modules.d/35network-manager/nm-run.sh
```
, right?

repos/system_upgrade/common/actors/addupgradebootentry/libraries/addupgradebootentry.py
fernflower: Thanks, I will change `ip=on` to `ip=dhcp`. Thanks!

fernflower:
@pvalena Hi! Sorry for taking some time, I have applied your suggestions, could you please take a look? If it looks fine I'd like to give it another round of thorough testing on rhel7 and rhel8 vagrant boxes.

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
pvalena: Yes, that's what I assume, I've not tried it exists in the initrd... but I suppose it does. Let me know if you encounter issues, I'll investigate.

fernflower: @pvalena hmmm I have tried that on rhel8 vagrant box and got an early fail in initram stage http://pastebin.test.redhat.com/1077800

pvalena: Right, let me check where the files live, when in the initrd.

pvalena: Well, the locations are indeed where the files are installed in the initrd.

IIUC:

 - This initrd is built specially for the upgrade (the dracut version/build is corresponding to the system you're running)
 - you don't have access to the rootfs (or don't have it upgraded yet)
 - dracut+initqueue was invoked already, and you've reached some `target`
 - the network is still not running (or not properly configured) and you need to start the interfaces (NetworkManager) manually
 - ` Initqueue settled` means: 
    ``` This hook (initqueue/settled) gets executed every time udev has settled.```
 
(I'm sorry If I've missed something, I still need to take a look at the reproducer.)

In that case it would make sense to run those paths you had originally.

pirat89: @pvalena not sure I understand it right, but, will it work on systems where the network manager is not used?

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/files/generate-initram.sh
```diff
@@ -67,6 +67,19 @@ build() {
...
+            DRACUT_MODULES_ADD="$DRACUT_MODULES_ADD --add network-manager"
+            touch /etc/leapp-initram-network-manager
+        ;;
+        scripts)
+            DRACUT_MODULES_ADD="$DRACUT_MODULES_ADD --add network";
+            touch /etc/leapp-initram-network-scripts
+        ;;
+        *)
+        ;;
+    esac
```
pirat89, line 81: @fernflower this should be rather implemented via `UpgradeInitramfsTasks`  and `TargetUserSpaceUpgradeTasks` instead of adding another logic into the script.

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
pvalena: Yes, the `ifup` should do that.... TBH I'm not all that familiar with legacy networking, so I hope it's sufficient :).

repos/system_upgrade/common/actors/checknfs/actor.py
```diff
@@ -18,6 +19,9 @@ class CheckNfs(Actor):
     tags = (ChecksPhaseTag, IPUWorkflowTag,)
 
     def process(self):
+        # if network in initramfs is enabled NFS inhibitors are redundant
+        if get_env('LEAPP_DEVEL_INITRAM_NETWORK', None):
```
pirat89, line 23: lgtm. just thinking, wouldn't be better to check specific values here? just thinking about a possibility that test could contain e.g. `LEAPP_DEVEL_INITRAM_NETWORK=0` , or typo, whatever.. I have no strong opinion here so keeping on you @fernflower

fernflower: TBH I'd rather not put any initram-network related logic into checknfs actor. Imho as far as checknfs actor is concerned if this variable is set to anything it literally says "The person calling knows what they are doing"

fernflower:
@pirat89 @vinzenz @Rezney @MichalHe 
Please review.
I have tested it on beaker/vagrant/in cloud and can verify that the network stays up during initramfs stage. The variable is made DEVEL so this should be a user-friendly gradual adoption, all other comments resolved.

fernflower:
Converted to draft as automated test from mmoran showed problems with upgrading system with NFS shares on rhel8.6

Rezney (approved):
Both "scripts" and "network-manager" options seem to be working fine. Thanks for the great job to all involved!

PR #961, opened 2022-08-29
CheckVDO: Ask user only faiulres and undetermined devices

jshimkus-rh:
The previous solution made possible to skip the VDO check answering
the user question (confirming no vdo devices are present) if the
vdo package is not installed (as the scan of the system could not
be performed). However as part of the bug 2096159 it was discovered
that some systems have very dynamic storage which could dissapear
in the very moment the check by the vdo tool is performed which lead
to the reported inhibitor. We have discovered that this could be real
blocker of the upgrade on such systems as it's pretty simple to hit
at least 1 of N devices to raise such an issue. (*)

To make the upgrade possible on such systems, the dialog has been
updated to be able to skip any problematic VDO checks:
 - undetermined block devices
 - failures during the vdo scan of a block device

In such a case, user must confirm that no VDO device non-managed
by LVM is present. The dialog is asking now for the `confirm` key
from user instead of `all_vdo_converted`. If any non-LVM managed VDO
devices are discovered, the upgrade is inhibited despite the answer
(this is supposed to happen only when user's answer is not right so
we are ok about that behaviour).

Also reports are updated, as previously it could happen that several
reports with the same title appear during one run of leapp, but each
of them has a different meaning. Set individual titles to all
reports. Also summaries or reports have been updated.

(*) This also includes situations when discovered list of devices
is not complete as some block devices could be loaded after the
initial scan of block devices (StorageInfo msg) is created. Which
means that such devices will not be checked at all as they will not
be known to other actors. We consider this ok as when a system with
dynamic storage is present, usually many block devices are
redundant. So usually user will have to answer the dialog anyway due
to other "unstable" block devices.

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=2096159
Jira: OAMG-7025

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/checkvdo/actor.py
+++ b/repos/system_upgrade/el8toel9/actors/checkvdo/actor.py
@@ -32,12 +32,24 @@ class CheckVdo(Actor):
     If the VdoConversionInfo model indicates unexpected errors occurred during
     scanning CheckVdo will produce appropriate inhibitory reports.
 
-    Lastly, if the VdoConversionInfo model indicates conditions exist where VDO
-    devices could exist but the necessary software to check was not installed
-    on the system CheckVdo will present a dialog to the user. This dialog will
-    ask the user to either install the required software if the user knows or
-    is unsure that VDO devices exist or to approve the continuation of the
-    upgrade if the user is certain that no VDO devices exist.
+    If the VdoConversionInfo model indicates conditions exist where VDO devices
+    could exist but the necessary software to check was not installed on the
+    system CheckVdo will present a dialog to the user. This dialog will ask the
+    user to either install the required software if the user knows or is unsure
+    that VDO devices exist or to approve the continuation of the upgrade if the
+    user is certain that either there are no VDO devices present or that all
+    VDO devices have been successfully converted.
+
+    To maximize safety CheckVdo operates against all block devices which
+    match the criteria for potential VDO devices.  Given the dynamic nature
+    of device presence within a system some devices which may have been present
+    during leapp discovery may not be present when CheckVdo runs.  As CheckVdo
+    defaults to producing inhibitory reports if a device cannot be checked
+    (for any reason) this dynamism may be problematic.  To prevent CheckVdo
+    producing an inhibitory report for devices which are dynamically no longer
+    present within the system the user may answer the previously mentioned
+    dialog in the affirmative when the user knows that all VDO devices have
+    been converted.  This will circumvent checks of block devices.
     """
 
     name = 'check_vdo'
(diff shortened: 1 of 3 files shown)


jshimkus-rh:
review please

jshimkus-rh:
@pirat89 @smitterl 
A friendly ping on maybe getting the review to happen?

MichalHe (commented):
Hi @jshimkus-rh, the patch looks nice code-wise, thank you. I will prioritize testing this patch, with the hope of approving and merging it tomorrow morning, if no problems will be found. When I am finished with the testing, I will ping you to squash the commits into one, or you can squash them in advance, so that I could merge the patch right away.

repos/system_upgrade/el8toel9/actors/checkvdo/libraries/checkvdo.py
```diff
@@ -111,36 +111,36 @@ def _process_undetermined_conversion_devices(devices):
...
+    if all_vdo_converted:
+        summary = ('User has asserted all VDO devices on the system have been '
+                   'successfully converted to LVM management.')
+
+        reporting.create_report([
+            _report_title,
+            reporting.Summary(summary),
+            reporting.Severity(reporting.Severity.INFO),
+            reporting.Groups([reporting.Groups.SERVICES, reporting.Groups.DRIVERS]),
+            reporting.Groups([])
```
MichalHe, line 141: Just noticed this line is redundant, however, it is not blocking this patch to be merged. Removing the line is up to you.
```suggestion
```

MichalHe (approved):

pirat89 (changes requested):
we should change messaging. more than 50% of people that will see the msg will not even know what the VDO is. Asking them whether all VDO devices has been converted can be confusing when no vdo devices are present. Also, providing the hint to install vdo rpm when vdo package is installed could be confusing. We should rather ask people for the confirmation of a statement that will be clear.

pirat89:
working on the fix right now

repos/system_upgrade/el8toel9/actors/checkvdo/actor.py
```diff
@@ -62,10 +62,10 @@ class CheckVdo(Actor):
             reason='Confirmation',
             components=(
                 BooleanComponent(
-                    key='all_vdo_converted',
-                    label='Are all VDO devices successfully converted to LVM '
-                          'management?',
-                    description='Enter True if all VDO devices on the system '
+                    key='confirm',
+                    label='Are all VDO devices. if any, successfully converted to LVM management?',
```
jshimkus-rh, line 66:
```suggestion
                    label='Are all VDO devices, if any, successfully converted to LVM management?',
```

repos/system_upgrade/el8toel9/actors/checkvdo/actor.py
```diff
@@ -62,10 +62,10 @@ class CheckVdo(Actor):
...
+                    description='Enter True if no VDO devices are present '
+                                'on the systen or all VDO devices on the system '
```
jshimkus-rh, line 68:
```suggestion
                                'on the system or all VDO devices on the system '
```

jshimkus-rh (commented):
A couple misspellings/punctuation.  Other than that, looks good.

repos/system_upgrade/el8toel9/actors/checkvdo/libraries/checkvdo.py
jshimkus-rh: Petr addressed this in his push.

pirat89:
@jshimkus @MichalHe one small update in the no-vdo-pkg-installed report to prevent situations like this:
```
Risk Factor: high (inhibitor)
Title: Cannot perform the VDO check of block devices
Summary: User has opted to inhibit the upgrade in regard to potential VDO devices requiring conversion to LVM management.
Remediation: [hint] Install the 'vdo' package and re-run upgrade to check for VDO devices requiring conversion.
Key: 429a99e13b19a7eebadbb8cb35233d8119bcf255
Risk Factor: high (inhibitor)
Title: Missing required answers in the answer file
Summary: One or more sections in answerfile are missing user choices: check_vdo.confirm
For more information consult https://leapp.readthedocs.io/en/latest/dialogs.html
Remediation: [hint] Please register user choices with leapp answer cli command or by manually editing the answerfile.
[command] leapp answer --section check_vdo.confirm=True
Key: 2f9802dc91315806c7cdc3c18d7b74f2a2383285

```

MichalHe:
I have retested the patch in the following scenarios:
- when no answer is provided, the upgrade is inhibited with the following entry
```
Risk Factor: high (inhibitor)
Title: Missing required answers in the answer file
Summary: One or more sections in answerfile are missing user choices: check_vdo.confirm
For more information consult https://leapp.readthedocs.io/en/latest/dialogs.html
Remediation: [hint] Please register user choices with leapp answer cli command or by manually editing the answerfile.
[command] leapp answer --section check_vdo.confirm=True
Key: 2f9802dc91315806c7cdc3c18d7b74f2a2383285
```
- when the answer is True, and there are problematic devices, the upgrade is not inhibited, and the following report entry is generated
```
Risk Factor: info
Title: Skipping the VDO check of block devices
Summary: User has asserted all VDO devices on the system have been successfully converted to LVM management or no VDO devices are present.
Key: 5c729d77d060b671f037a5d75defec3ec3ff03dd
```
- when the answer is False, and there are problems, the corresponding entries (depending on device state) are generated. Only suspicious report is generated when the 'VDO check of a block device fails. Also, I don't thing that the user know what VDO check is the entry referring to, thus I would change it to `Checking for VDO status/conversion of the following devices failed:`
```
Risk Factor: high (inhibitor)
Title: The VDO check of block devices failed
Summary: Unexpected result checking device
<failure reason>
Remediation: [hint] Resolve the conditions leading to the reported failure and re-run the upgrade.
Key: b313bb77c39a150d39b0286090da8a8294876f0e
```
- when the answer is False, and there are no problems, the upgrade is not inhibited, and no VDO-related entries are generated

From the functionality POV, the patch is OK.

jshimkus-rh:
@MichalHe Made the check failure title more explicit.

repos/system_upgrade/el8toel9/actors/checkvdo/libraries/checkvdo.py
```diff
@@ -116,11 +116,13 @@ def _process_undetermined_conversion_devices(devices):
...
-                   'potential VDO devices requiring conversion to LVM '
-                   'management.')
+        summary = (
+            'The check of block devices could not be performed as the \'vdo\' '
+            'package is not installed. All VDO devices must be converted to '
+            'LVM management prior the upgrade to prevent the loss of data.')
         remedy_hint = ('Install the \'vdo\' package and re-run upgrade to '
-                       'check for VDO devices requiring conversion.')
+                       'check for VDO devices requiring conversion or confirm '
+                       'that all VDO devices, if any, are manager by LVM.')
```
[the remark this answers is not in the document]
pirat89: Definitely! Who should be 'managed' if not 'manager' :-D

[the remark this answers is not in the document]
pirat89: thanks for the fix :)

pirat89:
Squashed commits. But discovered another problem with this solution. The userquestion is raised in 100% cases, even on machines without LVM installed. I am starting to work on the improvement. But would like to discuss it yet

repos/system_upgrade/el8toel9/actors/checkvdo/libraries/checkvdo.py
```diff
@@ -135,7 +141,11 @@ def _process_undetermined_conversion_devices(devices):
 
 
 def check_vdo(conversion_info):
-    all_vdo_converted = api.current_actor().get_all_vdo_converted_response()
+    if not _lvm_package_installed():
+        # nothing to do. VDO does not work without LVM
+        return
+
+    all_vdo_converted = api.current_actor().get_vdo_answer()
```
jshimkus-rh, line 148: For completeness...
`get_vdo_answer` isn't part of this push, but I know it's coming. :-)

pirat89: aa. yep :)

repos/system_upgrade/el8toel9/actors/checkvdo/libraries/checkvdo.py
```diff
@@ -113,42 +135,41 @@ def _process_undetermined_conversion_devices(devices):
...
-            reporting.Severity(reporting.Severity.INFO),
-            reporting.Groups([reporting.Groups.SERVICES, reporting.Groups.DRIVERS]),
-        ])
-    else:
-        _process_pre_conversion_vdos(conversion_info.pre_conversion)
-        _process_post_conversion_vdos(conversion_info.post_conversion)
-        _process_undetermined_conversion_devices(conversion_info.undetermined_conversion)
-        _process_failed_check_devices(conversion_info)
+    detected = _process_undetermined_conversion_devices(conversion_info.undetermined_conversion)
+    detected |= _process_failed_check_devices(conversion_info)
```
jshimkus-rh, line 171: I'm not a fan of using bitwise operators with logical values.  It's something that technically works but mixes, imo, different concepts.
That having been said, if it's consistent with leapp coding guidelines so be it.

pirat89: Results split into separate variables.

pirat89:
Tested manually cases without VDO devices present:
* no vdo pkg installed (without answer, True, False)
* vdo installed + all devices determined
* vdo installed + undetermined devices (without answer, True, False)

Note: skipping the test of the "check failure" as that one could be hit only rarely (and randomly) under special circumstances.

MichalHe (approved):
Code-wise LGTM. Did not test the PR myself, as I trust the extensive testing performed by @pirat89.

PR #957, opened 2022-08-26
ci(lint): Add differential-shellcheck GitHub action

jamacku:
Differential ShellCheck is a GitHub action that performs differential ShellCheck scans on files changed via PR and reports results directly in PR.

Since this repository contains some shell scripts I think you might find it useful. It is able to produce reports in SARIF format. GitHub understands this format and is able to display it nicely as a PR comment, and on the `Files Changed` tab, please see below.

image

image

Documentation is available at [@redhat-plumbers-in-action/differential-shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck#readme). Let me know If you are missing some feature or setting. I'm always happy to extend functionality.

PR #965, opened 2022-09-09
Make shellcheck happy again

pirat89:
Fixing:
```
85sys-upgrade-redhat/do-upgrade.sh:236:37: warning[SC2166]: Prefer [ p ] && [ q ] as [ p -a q ] is not well defined.
```

It's not a real issue as we do not care about the order of the evaluation, but making shellcheck happy.


pirat89:
This change does not need a note in the changelog.

jamacku:
Adding a shell linter could help avoid introducing new defects into your code base.
- #957

pirat89:
@jamacku thanks. I have it on my todolist for this week to get to that PR :) possibly I will made it today/tomorrow

fernflower:
Is that related to https://github.com/oamg/leapp-repository/pull/957 ? Like if we had enabled the checker, we would not have the need for this particular patch (as our CI would have caught that earlier), right?

pirat89:
@fernflower  yes

PeterMocary (approved):
LGTM. The change is equivalent and when tested on aarch64 works as intended. Thanks for keeping shellcheck happy!

PR #964, opened 2022-09-08
set_systemd_services_state: Add an actor which sets systemd services state

matejmatuska:
Introduces a new `set_systemd_services_state` actor, which enables/disables systemd services according to received SetSystemdServicesStateTask messages.

Some things I am still considering:
- also add an option to set mask/unmask/reenable/preset on a service
- generate a post upgrade report on error in addition to logging it
- `systemctl enable` has an `-f` option to force override an existing symlink, maybe that could be useful

Jira ref.: OAMG-1745

PR #966, opened 2022-09-14
migratentp: Don't raise StopActorExecutionError

mlichvar:
When a service cannot be enabled (e.g. due to masking) or when ntp2chrony fails, log an error message and create a report instead of failing the migration.

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=2089514

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/migratentp/libraries/migratentp.py
+++ b/repos/system_upgrade/el7toel8/actors/migratentp/libraries/migratentp.py
@@ -2,11 +2,8 @@ import base64
 import io
 import tarfile
 
-from leapp import reporting
 from leapp.exceptions import StopActorExecutionError
-from leapp.libraries.stdlib import CalledProcessError, run
-
-COMMON_REPORT_TAGS = [reporting.Groups.SERVICES, reporting.Groups.TIME_MANAGEMENT]
+from leapp.libraries.stdlib import api, CalledProcessError, run
 
 
 def extract_tgz64(s):
@@ -82,21 +79,7 @@ def migrate_ntp(migrate_services, config_tgz64):
 
     ignored_lines = ntp2chrony('/', ntp_conf, step_tickers)
 
-    config_resources = [reporting.RelatedResource('file', mc) for mc in migrate_configs + [ntp_conf]]
-    package_resources = [reporting.RelatedResource('package', p) for p in ['ntpd', 'chrony']]
-
-    if not ignored_lines:
-        reporting.create_report([
-            reporting.Title('{} configuration migrated to chrony'.format(' and '.join(migrate_configs))),
-            reporting.Summary('ntp2chrony executed successfully'),
-            reporting.Severity(reporting.Severity.INFO),
-            reporting.Groups(COMMON_REPORT_TAGS)
-        ] + config_resources + package_resources)
-
-    else:
-        reporting.create_report([
-            reporting.Title('{} configuration partially migrated to chrony'.format(' and '.join(migrate_configs))),
-            reporting.Summary('Some lines in /etc/ntp.conf were ignored in migration (check /etc/chrony.conf)'),
-            reporting.Severity(reporting.Severity.MEDIUM),
-            reporting.Groups(COMMON_REPORT_TAGS)
-        ] + config_resources + package_resources)
+    api.current_logger().info('Configuration files migrated to chrony: {}'.format(' '.join(migrate_configs)))
+    if ignored_lines:
+        api.current_logger().warning('Some lines in /etc/ntp.conf were ignored in migration'
+                                     ' (check /etc/chrony.conf)')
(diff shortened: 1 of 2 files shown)


pirat89:
wut? github just throw out all msgs I prepared as part of the review.... :skull:

pirat89 (changes requested):
current report must be fixed. (as github throw out everything I wrote last time, I am going to write it in separate comments). Next to other issues, all tests must pass. please remove all issues @mlichvar

repos/system_upgrade/el7toel8/actors/migratentp/libraries/migratentp.py
```diff
@@ -80,12 +85,28 @@ def migrate_ntp(migrate_services, config_tgz64):
...
+            reporting.Groups(COMMON_REPORT_TAGS)
+        ] + config_resources + package_resources)
+
+    elif failed_services:
+        reporting.create_report([
+            reporting.Title('{} not migrated to chrony'.format(' and '.join(migrate_configs))),
+            reporting.Summary('Failed to enable {} service'.format(' and '.join(failed_services))),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Groups(COMMON_REPORT_TAGS)
+        ] + config_resources + package_resources)
```
pirat89, line 107: Please fix the reports. This is not report but log.
* TItle is supposed to be statis. When there is a problem and the problem has different parameters (e.g. list of files is different), the problem is still same. So the title is expected to be same. I see that other reports in the file are wrong as well. These should be ideally fixed as well.
* Current summary is not summary but log. When writting summary, it's good to ask yourself questions like:
  * What is the thing I want user to take from this report?
  * How there report helps user to understand the problem and fix it?
  * ....

mlichvar: Do you want that to be addressed in a separate PR? This one is about not using StopActorExecutionError.
.

pirat89: It should be addressed in this one as the exception replacement introduces problematic reports. Or you can drop the reports and keep just the logging. There is still plenty of time till CTC1 (10w) so I think it's worthy to do it just once and right.

Btw, regarding systemd services we are going to introduce (today?) new funcionality that handles enabling/disabling of systemd services in general: https://github.com/oamg/leapp-repository/pull/964

pirat89: you could also drop the post-upgrade reports completely which will resolve this problem in this PR. do not forget to update other stuff so tests are passing.

mlichvar:
pylint now fails with python-2 due to unknown exceptions. Any suggestions?

repos/system_upgrade/el7toel8/actors/migratentp/libraries/migratentp.py
pirat89: Seems good to me. Thanks

repos/system_upgrade/el7toel8/actors/migratentp/libraries/migratentp.py
```diff
@@ -36,12 +32,13 @@ def ntp2chrony(root, ntp_conf, step_tickers):
         ntp_configuration = ntp2chrony.NtpConfiguration(root, ntp_conf, step_tickers)
         ntp_configuration.write_chrony_configuration('/etc/chrony.conf', '/etc/chrony.keys',
                                                      False, True)
-    except Exception as e:
-        raise StopActorExecutionError('ntp2chrony failed: {}'.format(e))
+    except (FileExistsError, FileNotFoundError, PermissionError) as e:
```
pirat89, line 35: > pylint now fails with python-2 due to unknown exceptions. Any suggestions?

Usually we go with `OSError` (all mentioned exceptions are made from `OSError`) which is suggested way when writting Py2 & Py3 compatible code.

```suggestion
    except OSError as e:
```

Another possibility is to add comments like below (and you hit problem with too long line which you need then handle as well..), but I suggest to go with the `OSError` as mentioned above. Showing this possibility just for completness:
```suggestion
    # NOTE: this code is executed by Python3 only where mentioned exceptions are defined. Ignore pylint for Py2 here
    except (FileExistsError, FileNotFoundError, PermissionError) as e:  # noqa: E0602; pylint: disable-undefined-variable
```

mlichvar: Thank you. Hopefully pylint is happy now.

pirat89: You are welcome \o seems good

pirat89 (approved):

mlichvar:
> @mlichvar as the related BZ has attached customer case, can you assign QE that will test the BZ before we merge it

I'm sorry, I'm outside of the team. I don't know who is doing QE for you.

pirat89:
Sanity tests results seems ok. Some visible failures have been caused by internal / test errors but looks good otherwise.

PR #969, opened 2022-09-26
Rework network configuration handling

lkundrak:
This splits off the ifcfg & NetworkManager profile reading parts from CheckIfCfg and CheckNetworkDeprecations into separate scanners; allowing the ifcfg code to be shared.

There's a bugfix [1] at the end, depending on the IfCfg parsing consolidation.

[1] Make IfCfgScanner accept simple quoted values <https://bugzilla.redhat.com/show_bug.cgi?id=2111691>

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/networkdeprecations/tests/unit_test_networkdeprecations.py
+++ b/repos/system_upgrade/el8toel9/actors/networkdeprecations/tests/unit_test_networkdeprecations.py
@@ -1,148 +1,124 @@
-import errno
-import textwrap
-
-import mock
-import six
-
-from leapp import reporting
-from leapp.libraries.actor import networkdeprecations
-from leapp.libraries.common.testutils import create_report_mocked, make_OSError
-
-
-def _listdir_nm_conn(path):
-    if path == networkdeprecations.NM_CONN_DIR:
-        return ['connection']
-    raise make_OSError(errno.ENOENT)
+from leapp.models import (
+    IfCfg,
+    IfCfgProperty,
+    NetworkManagerConnection,
+    NetworkManagerConnectionProperty,
+    NetworkManagerConnectionSetting
+)
+from leapp.reporting import Report
+from leapp.utils.report import is_inhibitor
 
 
-def _listdir_ifcfg(path):
-    if path == networkdeprecations.SYSCONFIG_DIR:
-        return ['ifcfg-wireless']
-    raise make_OSError(errno.ENOENT)
-
-
-def _listdir_keys(path):
-    if path == networkdeprecations.SYSCONFIG_DIR:
-        return ['keys-wireless']
-    raise make_OSError(errno.ENOENT)
-
-
-def test_no_conf(monkeypatch):
+def test_no_conf(current_actor_context):
     """
     No report if there are no networks.
     """
 
-    monkeypatch.setattr(networkdeprecations.os, 'listdir', lambda _: ())
-    monkeypatch.setattr(reporting, 'create_report', create_report_mocked())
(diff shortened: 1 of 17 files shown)


lkundrak:
review please

pirat89:
@lkundrak hi, is there a BZ for this? It seems to me familiar but not sure. if there is, please add the link into the description. we will get to the review today/tomorrow (putting on my todo)

lkundrak:
> @lkundrak hi, is there a BZ for this? It seems to me familiar but not sure.

Only for the last change. Instead of applying it in two different places, I ended up moving the code into a scanner actor, as (I think) you advised before.

> if there is, please add the link into the description.

Done.

pirat89:
@lkundrak thanks! I am sorry for the postponed review (urgent looong bug investigation). It seems we get to review in the end of this week.

pirat89 (changes requested):

repos/system_upgrade/el8toel9/actors/networkdeprecations/tests/unit_test_networkdeprecations.py
```diff
@@ -1,148 +1,125 @@
...
-        key-mgmt=none
-        wep-key-type=1
-        wep-key0=abcde
-    """)
+    static_wep_nm_conn = NetworkManagerConnection(filename='/NM/wlan0.nmconn', settings=(
+        NetworkManagerConnectionSetting(name='wifi-security', properties=(
+            NetworkManagerConnectionProperty(name='auth-alg', value='open'),
+            NetworkManagerConnectionProperty(name='key-mgmt', value='none'),
+            NetworkManagerConnectionProperty(name='wep-key-type', value='1'),
+            NetworkManagerConnectionProperty(name='wep-key0', value='abcde'),
```
pirat89, line 45: @lkundrak so we discussed the secrets solution and for now, implement an allowlist of keys for which you need (and store) the value and for the others set None. We will investigate a better solution around secrets for the future but right now let's go just in this way.

lkundrak:
Pushed new version for review.

The major change is the omission of the secrets. Easy for ifcfg (where secrets are in a separate file), but for keyfiles we need to utilize libnm via GObject introspection. That won't be there on python2 platforms (no biggie, we're concerned about el8 and later only) and on UBI (used for testing, not NetworkManager there at all).

pirat89:
review in progress

repos/system_upgrade/el8toel9/actors/networkmanagerconnectionscanner/libraries/networkmanagerconnectionscanner.py
```diff
@@ -0,0 +1,65 @@
...
+try:
+    import gi
+    try:
+        gi.require_version("NM", "1.0")
+        from gi.repository import GLib, NM
+        libnm_available = True
+    except ValueError:
+        err_details = 'NetworkManager-libnm package is not available'
+except ImportError:
+    err_details = 'python3-gobject-base package is not available'
```
pirat89, line 20: @lkundrak  I need to sync  with you about this as probably the package should have updated requirement. Let's sync on monday.

pirat89: @lkundrak unfortunately I do not have permissions to update your branch. So I pushed the changes to my fork. Please update your branch when you are online, so we can merge it: https://github.com/pirat89/leapp-repository/tree/lr/ifcfg-nm-rework

pirat89 (approved):
merging \o/ thanks for the contribution!

PR #971, opened 2022-09-30
firewalldcheckallowzonedrifting: Fix the remediation cmd

pirat89:
The remediation cmd was incorrect as the cmd is written as string instead of list, the fix:
  ['cmd param param'] -> ['cmd', 'paramm', 'param']

JIRA: OAMG-7694


mmacura311:
tested manually in cockpit, the corrected command works fine.

mmacura311 (approved):
LGTM

PR #976, opened 2022-10-18
[WIP] move checksaphana actor from 7to8 to common

alexxa:
Don't merge.

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/checksaphana/libraries/checksaphana.py
+++ b/repos/system_upgrade/el8toel9/actors/checksaphana/libraries/checksaphana.py
@@ -1,0 +1,222 @@
+from leapp import reporting
+from leapp.libraries.common.config import architecture
+from leapp.libraries.stdlib import api
+from leapp.models import SapHanaInfo
+
+# SAP HANA Compatibility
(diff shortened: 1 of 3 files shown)


repos/system_upgrade/el8toel9/actors/checksaphana/libraries/checksaphana.py
```diff
@@ -0,0 +1,218 @@
...
+            req_sp, req_rev, req_pl = requirements
+            if sp == req_sp:
+                rev = int(number.lstrip('0') or '0')
+                if rev < req_rev:
+                    continue
+                if rev == req_rev:
+                    patch_level = int(_manifest_get(instance.manifest, 'rev-patchlevel', '00').lstrip('0') or '0')
+                    if patch_level < req_pl:
+                        continue
+                return True
```
vinzenz, line 125: Please make this a separate function and not duplicate it

alexxa: @vinzenz I know it's terrible, but we tested it already and the deadline for the whole thing to make it in time for a release is today. Ideally, I'd like to move running check to common actors, and keep versions check in corresponding repos paths. May we keep it as it is for now, and I will change it a bit later? 
I can't figure out how to make it with unit tests to pass

PR #977, opened 2022-10-19
Enable device_cio_free.service systemd service on s390x

matejmatuska:
After an IPU the device_cio_free.service systemd service (exclusive to s390x) is disabled even though the vendor preset is set to disable. The new actor enables the service.

The service is used to enable devices not explicitly enabled on kernel command line.

The test is broken now until oamg/leapp#680 is merged.

Jira ref.: OAMG-6302
Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=2032953

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/enableddeviceciofreeservices390/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/enableddeviceciofreeservices390/actor.py
@@ -1,0 +1,21 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import enabledeviceciofreeservice
+from leapp.models import SystemdServicesTasks
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class EnableDeviceCioFreeService(Actor):
+    """
+    Enables device_cio_free.service systemd service on s390x
+
+    After an upgrade this service ends up disabled even though it's vendor preset is set to enabled.
+    The service is used to enable devices which are not explicitly enabled on the kernel command line.
+    """
+
+    name = 'enable_device_cio_free_service'
+    consumes = ()
+    produces = (SystemdServicesTasks,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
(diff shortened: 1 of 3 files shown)


matejmatuska:
Not sure about the naming, maybe something more abstract would be better

matejmatuska:
Tested in a VM, the service was successfully enabled after the upgrade.

repos/system_upgrade/el7toel8/actors/enableddeviceciofreeservices390/actor.py
```diff
@@ -0,0 +1,21 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import enabledeviceciofreeservice
+from leapp.models import SystemdServicesTasks
+from leapp.tags import FinalizationPhaseTag, IPUWorkflowTag
```
pirat89, line 4: @matejmatuska  this one is expected to be performed during the checksphase

repos/system_upgrade/el7toel8/actors/enableddeviceciofreeservices390/tests/test_enableddeviceciofreeservice.py
```diff
@@ -0,0 +1,37 @@
...
+from leapp.libraries.stdlib import api
+from leapp.models import SystemdServicesTasks
+
+
+def test_task_produced_on_s390(monkeypatch):
+
+    def matches_architecture_mocked(arch):
+        return True
+
+    monkeypatch.setattr(architecture, 'matches_architecture', matches_architecture_mocked)
```
pirat89, line 15: @matejmatuska mocking the architecture in this way is not ideal. it's better to specify the architecture you want to have when the test is running via the CurrentActorMocked. It's more easier and you can then simply check the behaviour even for all architectures

pirat89: @matejmatuska the code is updated. you can take a look what I meant :)

pirat89 (approved):

PR #981, opened 2022-11-03
Propagate TEST_PATHS to test_container targets

lkundrak:
Allows for quick containerized runs of specific checks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -371,8 +371,8 @@ _test_container_ipu:
 		echo "Only supported TEST_CONT_IPUs are el7toel8, el8toel9"; exit 1; \
 		;; \
 	esac && \
-	$(_CONTAINER_TOOL) exec -w /repocopy $$_CONT_NAME make clean && \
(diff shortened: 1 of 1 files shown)


fernflower (approved):
lgtm, doesn't break anything

matejmatuska:
@fernflower @lkundrak 

This actually broke the containerized tests for me similar to how tests without containers are broken if you don't set REPOSITORIES.

The reason why the TEST_PATHS weren't propagated is that we want to test just the REPOSITORIES corresponding to python version e.g. el7toel8 on python2.7 and python 3.6 and el8elto9 on python3.6 and 3.9. There are problems with the tests if the repositories aren't tested in isolation. This was the reason for introducing containerized test.

What was the reason you wanted to propagate `TEST_PATHS`? If it was to make the `ACTOR` variable work we can just propagate that.

fernflower:
@lkundrak Hi, so in the end we had to revert the change in https://github.com/oamg/leapp-repository/pull/1023. @matejmatuska has some nice ideas for possible resolution, please create another PR if you still need this change.

PR #983, opened 2022-11-07
s390x fixes: zfcp+zipl

rmetrich:
When having systems configured with ZFCP instead of DASD, the disks
are not seen while rebooting because `/etc/zfcp.conf` is missing
in the initramfs.

The conversion of ZIPL to BLS on IBM Z machines failed when
  **a)** the machine was configured using ZFCP instead of DASD
  **b)** /boot was not on a separate partition

**In case a)**, the zipl-switch-to-blscfg script failed as the /dev has
not been propagated to into the el8userspace container and also all required nodes has not been visible inside the upgrade initramfs due to missing `/etc/zfcp.conf` inside the initramfs. Regarding that the script failed the conversion.

With this fix, the /dev is bindmounted into the el8userspace container
using the (systemd-nspawn) `--bind` option. The direct bind mounting
via `leapp.libraries.common.mounting.BindMount` cannot be used in this
case as it blocks the correct start of the container.

Also, when the file exists, it's copied inside the userspace container
and installed in the upgrade initramfs, producing
   TargetUserSpaceUpgradeTasks, UpgradeInitramfsTasks
messages.

**In case b)**, the content of /boot has been removed during the upgrade
due to problems when using BindMount on normal directory (that is not
mountpoint). This has been possibly resolved by this commit also,
as the /boot has been propagated using the --bind (sysmd-nspawn)
option as well.
**NOTE:** The case b) has not been tested yet and this PR primarily targets to fix the case a). As the inhibitor for s390x systems with /boot merged in rootfs, even when the case b) is not tested successfuly, it does not prevent the merge of this PR.

See https://bugzilla.redhat.com/show_bug.cgi?id=2140563

### DIFF
--- a/repos/system_upgrade/common/actors/scandasd/tests/unit_test_scandasd.py
+++ b/repos/system_upgrade/common/actors/scandasd/tests/unit_test_scandasd.py
@@ -3,18 +3,18 @@ import os
 import pytest
 
 from leapp.libraries.actor import scandasd
-from leapp.libraries.common.config.architecture import ARCH_S390X
-from leapp.libraries.common.testutils import logger_mocked, produce_mocked
+from leapp.libraries.common.config import architecture
+from leapp.libraries.common.testutils import CurrentActorMocked, logger_mocked, produce_mocked
 from leapp.models import CopyFile, TargetUserSpaceUpgradeTasks, UpgradeInitramfsTasks
 
 
 def test_dasd_exists(monkeypatch):
-    monkeypatch.setattr(scandasd.architecture, 'matches_architecture', lambda dummy: True)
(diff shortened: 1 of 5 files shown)


pirat89:
looks good. I will do some updates in unit-tests

pirat89:
@rmetrich seems done. Pls check if you are ok with changes. The commit msgs have been updated too.

@smitterl Will you have a time to create for me the system with merged /boot as mentioned in the case b) of the description of this PR? I can test whether the current fix is really fixing this issue (still manual update of the code is needed).

rmetrich:
@pirat89 LGTM

pirat89 (approved):

PR #984, opened 2022-11-08
Fix the check of memory (RAM) limits

pirat89:
The checkmem actor was incorrect as the limits have been written in MiBs however the value obtained from /proc/meminfo is in KiBs. So the actual check has been incorrect.

Also the memory limits have been changed since the creation of the actor. Updating the values in KiBs based on the current table:
    https://access.redhat.com/articles/rhel-limits

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=2139907

### DIFF
--- a/repos/system_upgrade/common/actors/checkmemory/libraries/checkmemory.py
+++ b/repos/system_upgrade/common/actors/checkmemory/libraries/checkmemory.py
@@ -5,10 +5,10 @@ from leapp.libraries.stdlib import api
 from leapp.models import MemoryInfo
 
 min_req_memory = {
-    architecture.ARCH_X86_64: 1536,  # 1.5G
-    architecture.ARCH_ARM64: 2048,  # 2Gb
-    architecture.ARCH_PPC64LE: 2048,  # 2Gb
-    architecture.ARCH_S390X: 1024  # 1Gb
+    architecture.ARCH_X86_64: 1572864,  # 1.5G
+    architecture.ARCH_ARM64: 1572864,  # 1.5G
+    architecture.ARCH_PPC64LE: 3145728,  # 3G
+    architecture.ARCH_S390X: 1572864,  # 1.5G
 }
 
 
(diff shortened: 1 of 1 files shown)


MichalHe (approved):
I've check that the memory limits match the minimal memory requirements for RHEL systems according to the attached link, and I have verified that the conversion do GiB to KiB is correct, and it indeed is:
- `1.5GiB =  1024 * 1024 * 1.5 KiB = 1572864 KiB`
- `3.0GiB =  1024 * 1024 * 3.0 KiB = 3145728 KiB`

Also, the patch correctly takes into account that `/etc/meminfo` incorrectly uses `kB` while the actual value is in `KiB` ([described here](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/s2-proc-meminfo)), thus the calculations above are correct wrt. units.

I have tested the proposed PR on a RHEL7 VM with 1GB of memory with the following results:
- `master` - preupgrade succeeds, upgrade is not inhibited
- `6a5b95e5` (this PR) - preupgrade succeeds with an inhibitor - low memory

The produced inhibitor:
```
Risk Factor: high (inhibitor)
Title: Minimum memory requirements for RHEL 8 are not met
Summary: Memory detected: 1014516 KiB, required: 1572864 KiB
```

Note that to perform the testing I had to build the package locally, as copr seemed broken, e.g., the offered artifacts from this PR (see https://dashboard.packit.dev/results/copr-builds/518027).
`leapp-upgrade-el7toel8-0.17.0-0.el7.noarch leapp-upgrade-el7toel8-deps-0.17.0-0.el7.noarch`

LGTM.

pirat89:
@viniciusferrao it's too late for RHEL 8.7 (released already). Most likely it's going to be part of the next release (8.8 GA time)

evgeni:
Small question: was it considered to keep `min_req_memory` in MB (or even use GB?) instead and then do the calculation of KiB in the actor code? With today's machines, writing memory limits as KB is tedious and makes it much harder to compare with documentation (like https://access.redhat.com/articles/rhel-limits) which uses "sane" units (see this PR where Michal went and re-calculated the numbers, just to be sure).

Obviously, not blocking or anything, but a thing to consider?

pirat89:
> Small question: was it considered to keep `min_req_memory` in MB (or even use GB?) instead and then do the calculation of KiB in the actor code? With today's machines, writing memory limits as KB is tedious and makes it much harder to compare with documentation (like https://access.redhat.com/articles/rhel-limits) which uses "sane" units (see this PR where Michal went and re-calculated the numbers, just to be sure).
> 
> Obviously, not blocking or anything, but a thing to consider?

Good point. let me do it

pirat89:
@MichalHe I've switched units to MiB and disabled linter for those two division lines due to FP errors.

MichalHe (approved):
The upgrade is correctly inhibited wrt. the latest changes when running on a system with insufficient memory:
```
Risk Factor: high (inhibitor)
Title: Minimum memory requirements for RHEL 8 are not met
Summary: Memory detected: 990 MiB, required: 1536 MiB
```
The updated memory limits and their conversion to KiB did not change with the recent changes, thus, LGTM.

I will wait on the results of integration tests before proceeding with merge, but I doubt any problems will be encountered.

MichalHe:
I have successfully performed a 7to8 upgrade to make sure that nothing is broken, since a lot of intergration tests were failing - most likely an infra issue since the tests fail on repo download, which has nothing to do with this patch. Since the upgrade went fine, merging.

PR #986, opened 2022-11-10
Fix cephvolume actor

holser:
Change cephvolume behavior to return None when ceph-osd container is not found.

Fixes: rhbz#2141393

### DIFF
--- a/repos/system_upgrade/common/actors/cephvolumescan/libraries/cephvolumescan.py
+++ b/repos/system_upgrade/common/actors/cephvolumescan/libraries/cephvolumescan.py
@@ -12,7 +12,6 @@ CONTAINER = "ceph-osd"
 
 
 def select_osd_container(engine):
-    container_name = ""
     try:
         output = run([engine, 'ps'])
     except CalledProcessError as cpe:
@@ -24,7 +23,7 @@ def select_osd_container(engine):
         container_name = line.split()[-1]
         if re.match(CONTAINER, container_name):
             return container_name
-    return container_name
+    return None
 
 
 def get_ceph_lvm_list():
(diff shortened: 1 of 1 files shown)


holser:
review please

jbadiapa (changes requested):
I will remove line 15 [1] as the initialization of the var is not needed.

[1] https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/common/actors/cephvolumescan/libraries/cephvolumescan.py#L15

asm0deuz:
@holser Hi! I am thinking about another solution which would require less changes. I will test it and open another PR if it works.

guits (dismissed):

asm0deuz:
@holser Your solution is the correct one. I have the same comment as @jbadiapa .

holser:
> I will remove line 15 [1] as the initialization of the var is not needed.
> 
> [1] https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/common/actors/cephvolumescan/libraries/cephvolumescan.py#L15

Thanks. I have updated PR with proposed change

holser:
> @holser Your solution is the correct one. I have the same comment as @jbadiapa .

Thanks a lot for comments. I have updated PR with proposed change.

jbadiapa:
lgtm
thanks,

holser:
@asm0deuz @guits @pirat89 Please review when you have time

pirat89 (approved):
lgtm. waiting for qe ack in the BZ + confirmation from qe it works before the merge.

pirat89:
we hav ethe confirmation from the QE, so merging.

PR #989, opened 2022-11-18
Use "grub" library to find the GRUB device

Rezney:
We need to find the device with GRUB in order to update grub core (on BIOS systems). Previously, we used "grubdevname" actor to get the info and send a message but this solution was error prone due to device renaming.

This fixes only the "checkgrubcore" actor as the "updategrubcore" actor already uses the library.


repos/system_upgrade/common/actors/checkgrubcore/actor.py
```diff
@@ -29,9 +28,8 @@ def process(self):
 
         ff = next(self.consume(FirmwareFacts), None)
         if ff and ff.firmware == 'bios':
-            dev = next(self.consume(GrubDevice), None)
+            dev = grub.get_grub_device()
```
pirat89, line 31: @Rezney  you cannot use the grub library in the checkphase.

Rezney:
@pirat89 @MichalHe this should be ready. Please take a look.

repos/system_upgrade/common/models/grubinfo.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class GrubInfo(Model):
+    """
+    Information about Grub
+    """
+    topic = SystemFactsTopic
+
+    device_name = fields.Nullable(fields.String())
+    """ Name of the block device where Grub is located. """
```
pirat89, line 12: @Rezney  discussed, rather change name to `orig_device_name` and add info into the docstring this is the original name, that is valid only during the preupgrade phases and could be different after the reboot.

```suggestion
    orig_device_name = fields.Nullable(fields.String())
    """
    Original name of the block device where Grub is located.
    
    The name is persistent during the boot of the system so it's safe to be used during
    preupgrade phases. However the name could be different after the reboot, so
    it's recommended to use `leapp.libraries.common.grub.get_grub_device()` anywhere
    else.
    """
```

Rezney: Fixed.

pirat89 (approved):
lgtm. can be merged after tests pass

pirat89:
postponing the merge by 1-2d until the infrastructure issues are resolved.

PR #990, opened 2022-11-25
Enable disabling dnf plugins in the dnfcnfig library

Rezney:
When on AWS, we need to disable the "amazon-id" plugin during the upgrade stage as we do not have network up and running there yet. Moreover, even with the network up, we do already have all the data cached so further communication with its backend could invalidate the data.

### DIFF
--- a/repos/system_upgrade/common/libraries/dnfconfig.py
+++ b/repos/system_upgrade/common/libraries/dnfconfig.py
@@ -30,15 +30,21 @@ def _strip_split(data, sep, maxsplit=-1):
     return [item.strip() for item in data.split(sep, maxsplit)]
 
 
-def _get_main_dump(context):
+def _get_main_dump(context, disable_plugins):
     """
     Return the dnf configuration dump of main options for the given context.
 
     Returns the list of lines after the line with "[main]" section
     """
 
+    cmd = ['dnf', 'config-manager', '--dump']
+
+    if disable_plugins:
+        for plugin in disable_plugins:
+            cmd += ['--disableplugin', plugin]
+
     try:
-        data = context.call(['dnf', 'config-manager', '--dump'], split=True)['stdout']
+        data = context.call(cmd, split=True)['stdout']
     except CalledProcessError as e:
         api.current_logger().error('Cannot obtain the dnf configuration')
         raise StopActorExecutionError(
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/common/libraries/dnfconfig.py
```diff
@@ -73,25 +77,27 @@ def _get_main_dump(context):
...
+def _set_excluded_pkgs(context, pkglist, disable_plugins):
     """
     Configure DNF to exclude packages in the given list
 
     Raise the CalledProcessError on error.
     """
     exclude = 'exclude={}'.format(','.join(pkglist))
     cmd = ['dnf', 'config-manager', '--save', '--setopt', exclude]
+    if disable_plugins:
+        cmd.extend(disable_plugins)
```
pirat89, line 100:
```suggestion
    if disable_plugins:
        for plugin in disable_plugins:
            cmd += ['--disableplugin', plugin]
```

repos/system_upgrade/common/libraries/dnfconfig.py
```diff
@@ -30,15 +30,19 @@ def _strip_split(data, sep, maxsplit=-1):
...
+def _get_main_dump(context, disable_plugins):
     """
     Return the dnf configuration dump of main options for the given context.
 
     Returns the list of lines after the line with "[main]" section
     """
 
+    cmd = ['dnf', 'config-manager', '--dump']
+    if disable_plugins:
+        cmd.extend(disable_plugins)
```
pirat89, line 42:
```suggestion
    if disable_plugins:
        for plugin in disable_plugins:
            cmd += ['--disableplugin', plugin]
```

repos/system_upgrade/common/libraries/dnfplugin.py
```diff
@@ -337,22 +339,28 @@ def perform_transaction_install(target_userspace_info, storage_info, used_repos,
         # communicate with udev
         cmd_prefix = ['nsenter', '--ipc=/installroot/proc/1/ns/ipc']
 
+        disable_plugins = []
+        if plugin_info:
+            for info in plugin_info:
+                if stage in info.disable_in:
+                    disable_plugins += ['--disableplugin', info.name]
```
pirat89, line 346:
```suggestion
                    disable_plugins.append(info.name)
```

pirat89 (changes requested):
The `disable_plugins` parameter should just contain the list of plugins supposed to be disabled, instead of the list of options with arguments. The completion construction of the command should be done in functions calling dnf.

repos/system_upgrade/common/libraries/dnfplugin.py
```diff
@@ -337,22 +339,28 @@ def perform_transaction_install(target_userspace_info, storage_info, used_repos,
         # communicate with udev
         cmd_prefix = ['nsenter', '--ipc=/installroot/proc/1/ns/ipc']
 
+        disable_plugins = []
+        if plugin_info:
```
fernflower, line 343: Nit: this can be moved into a function to be reused at L395 and L427.

pirat89: yes, but let's keep it already as it is in this phase. waiting for the test results. I am in favor to merge this one when tests pass

pirat89 (approved):
w8ing for test results

PR #991, opened 2022-11-25
Enable upgrades on s390x when /boot is part of rootfs

pirat89:
Regarding the fix provided in commit a6445b39 we do not need to inhibit the IPU on the s390x architecture when /boot is not separated on its own partition, but it's part of the rootfs.


pirat89:
Keeping it as a draft now until it's confirmed the fix works. Before the switch back to the PR, I need to update also the commit msg yet.

pirat89:
just a rebuild so a build is present till the EOY

smitterl:
@pirat89 Wei Shi has tested this and it works for 8->9.

Makefile
```diff
@@ -319,9 +319,9 @@ lint:
 		echo "--- Linting done. ---"; \
 	fi
 
-	if [[  "`git rev-parse --abbrev-ref HEAD`" != "master" ]] && [[ -n "`git diff $(MASTER_BRANCH) --name-only`" ]]; then \
+	if [[  "`git rev-parse --abbrev-ref HEAD`" != "$(MATER_BRANCH)" ]] && [[ -n "`git diff $(MASTER_BRANCH) --name-only --diff-filter AMR`" ]]; then \
```
fernflower, line 322: Typo s/MATER_BRANCH/MASTER_BRANCH ?

pirat89: Good catch! Thanks :) Fixed.

fernflower (approved):
code lgtm.

pirat89:
@smitterl @fernflower  postponing the merge as I've got confusing feedback about the solution - with positive and negative results. Waiting for re-testing

PR #993, opened 2022-11-29
missinggpgkey: polish the report msg

pirat89:


pirat89:
I've missed to upgrade the report. Handled in https://github.com/oamg/leapp-repository/pull/993

pirat89:
Result:
```
Risk Factor: high (inhibitor)
Title: Detected unknown GPG keys for target system repositories
Summary: Some of the target repositories require GPG keys that are not installed in the current RPM DB or are not stored in the /etc/leapp/repos.d/system_upgrade/common/files/rpm-gpg/8 directory. Leapp is not able to guarantee validity of such gpg keys and manual review is required, so any spurious keys are not imported in the system during the in-place upgrade. The following additional gpg keys are required to be imported during the upgrade:
    - https://download.copr.fedorainfracloud.org/results/pstodulk/test-repo/pubkey.gpg
Remediation: [hint] Check the listed GPG keys they are valid and import them into the host RPM DB or store them inside the /etc/leapp/repos.d/system_upgrade/common/files/rpm-gpg/8 directory prior the upgrade. If you want to proceed the in-place upgrade without checking any RPM signatures, execute leapp with the `--nogpgcheck` option.
Key: e7e3f34a530e861b14f1cdd759cb62f92677066f

```

PR #994, opened 2022-12-01
ensure Satellite metapackages are installed after upgrade

evgeni:
When upgrading from EL7 to EL8, we theoretically can run into the situation where the `satellite` and `satellite-capsule` metapackages are removed during the upgrade due to dependency problems. While we are not aware of any actual occurences of this problem today, let's play safe and explicitly add those packages to the `to_install` set.


pirat89 (approved):
seems good

fernflower (approved):

PR #995, opened 2022-12-06
Include also Leapp RHUI special rpms in the whitelist

Rezney:


MichalHe:
Th e2e tests for RHUI systems pass, the rest of the e2e tests seem to fail on infra issues, thus merging.

PR #996, opened 2022-12-07
Prevent failed upgrade from restarting in initramfs

matejmatuska:
When the upgrade fails in the initramfs the dracut shell is entered.
Upon exiting the dracut shell, the upgrade.target is restarted which
causes the upgrade.service, which runs the leapp upgrade, to rerun as
well.

This commit fixes that by creating a "flag" file when the upgrade
fails, whose existence is checked before reruning the upgrade and the
upgrade is prevented in such case.

Also, a new `removeupgradeartifacts` actor is introduced to clean up leftover upgrade artifacts, including the upgrade failed flag file, at the beginning of the upgrade process.

Jira ref.: OAMG-4224

### DIFF
--- a/repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
+++ b/repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
@@ -46,6 +46,8 @@ fi
 export NSPAWN_OPTS="$NSPAWN_OPTS --keep-unit --register=no --timezone=off --resolv-conf=off"
 
 
+export LEAPP_FAILED_FLAG_FILE="/root/tmp_leapp_py3/.leapp_upgrade_failed"
+
 #
 # Temp for collecting and preparing tarball
 #
@@ -268,6 +270,15 @@ do_upgrade() {
         rv=$?
     fi
 
+    if [ "$rv" -ne 0 ]; then
+        # set the upgrade failed flag to prevent the upgrade from running again
+        # when the emergency shell exits and the upgrade.target is restarted
+        local dirname
+        dirname="$(dirname "$NEWROOT$LEAPP_FAILED_FLAG_FILE")"
+        [ -d "$dirname" ] || mkdir "$dirname"
+        "$NEWROOT/bin/touch" "$NEWROOT$LEAPP_FAILED_FLAG_FILE"
+    fi
+
     # Dump debug data in case something went wrong
     if want_inband_dump "$rv"; then
         collect_and_dump_debug_data
@@ -338,6 +349,15 @@ mount -o "remount,rw" "$NEWROOT"
 
 ##### do the upgrade #######
 (
+    # check if leapp previously failed in the initramfs, if it did return to the emergency shell
+    [ -f "$NEWROOT$LEAPP_FAILED_FLAG_FILE" ] && {
+        echo >&2 "Found file $NEWROOT$LEAPP_FAILED_FLAG_FILE"
+        echo >&2 "Error: Leapp previously failed and cannot continue, returning back to emergency shell"
+        echo >&2 "Please file a support case with $NEWROOT/var/log/leapp/leapp-upgrade.log attached"
+        echo >&2 "To rerun the upgrade upon exiting the dracut shell remove the $NEWROOT$LEAPP_FAILED_FLAG_FILE file"
+        exit 1
+    }
+
     [ ! -x "$NEWROOT$LEAPPBIN" ] && {
         warn "upgrade binary '$LEAPPBIN' missing!"
         exit 1
--- a/repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
+++ b/repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
@@ -274,7 +274,7 @@ do_upgrade() {
         # set the upgrade failed flag to prevent the upgrade from running again
         # when the emergency shell exits and the upgrade.target is restarted
         local dirname
-        dirname="$(dirname "$NEWROOT$LEAPP_FAILED_FLAG_FILE")"
+        dirname="$("$NEWROOT/bin/dirname" "$NEWROOT$LEAPP_FAILED_FLAG_FILE")"
         [ -d "$dirname" ] || mkdir "$dirname"
         "$NEWROOT/bin/touch" "$NEWROOT$LEAPP_FAILED_FLAG_FILE"
     fi
--- a/repos/system_upgrade/common/actors/removeupgradeartifacts/actor.py
+++ b/repos/system_upgrade/common/actors/removeupgradeartifacts/actor.py
@@ -1,0 +1,23 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import removeupgradeartifacts
+from leapp.tags import InterimPreparationPhaseTag, IPUWorkflowTag
+
+
+class RemoveUpgradeArtifacts(Actor):
+    """
+    Removes artifacts left over by previous leapp runs
+
+    After the upgrade process, there might be some leftover files, which need
+    to be cleaned up before running another upgrade.
+
+    Removed artifacts:
+    - /root/tmp_leapp_py3/ directory (includes ".leapp_upgrade_failed" flag file)
+    """
+
+    name = 'remove_upgrade_artifacts'
+    consumes = ()
+    produces = ()
+    tags = (InterimPreparationPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        removeupgradeartifacts.process()
--- a/repos/system_upgrade/common/actors/removeupgradeartifacts/libraries/removeupgradeartifacts.py
+++ b/repos/system_upgrade/common/actors/removeupgradeartifacts/libraries/removeupgradeartifacts.py
@@ -1,0 +1,17 @@
+import os
+
+from leapp.libraries.stdlib import api, CalledProcessError, run
+
+UPGRADE_ARTIFACTS_DIR = '/root/tmp_leapp_py3/'
+
+
+def process():
+    if os.path.exists(UPGRADE_ARTIFACTS_DIR):
+        api.current_logger().debug(
+                "Removing leftover upgrade artifacts dir: {} ".format(UPGRADE_ARTIFACTS_DIR))
+
+        try:
+            run(['rm', '-rf', UPGRADE_ARTIFACTS_DIR])
+        except (CalledProcessError, OSError) as e:
+            api.current_logger().debug(
+                    'Failed to remove leftover upgrade artifacts dir: {}'.format(e))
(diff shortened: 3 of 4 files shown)


repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -324,6 +332,12 @@ mount -o "remount,rw" "$NEWROOT"
 
 ##### do the upgrade #######
 (
+    # check if leapp previously failed in the initramfs, if it did return to the emergency shell
+    [ -f "$NEWROOT$LEAPP_FAILED_FLAG_FILE" ] && {
+        echo "Leapp previously failed and cannot continue, returning back to emergency shell"
```
matejmatuska, line 337: One issue with the message is there are so many logging messages after it is printed that it is scrolled off screen.

Rezney: That is expected. Users should be ready to re-read the logs when debugging.

repos/system_upgrade/common/actors/removeupgradefailedflag/actor.py
```diff
@@ -0,0 +1,21 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import removeupgradefailedflag
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class RemoveUpgradeFailedFlag(Actor):
+    """
+    Removes upgrade failed flag file
```
Rezney, line 8:
```suggestion
    Removes the upgrade failed flag file
```

repos/system_upgrade/common/actors/removeupgradefailedflag/actor.py
```diff
@@ -0,0 +1,21 @@
...
+
+    The upgrade failed flag file is set in the initramfs when the upgrade fails
```
Rezney, line 10:
```
In case the in-place upgrade fails in the initramdisk "phase", the Leapp process creates the "/var/log/leapp/.leapp_upgrade_failed" flag file to prevent re-running this part from scratch. We need to remove the flag file before the next run.
```

repos/system_upgrade/common/actors/removeupgradefailedflag/libraries/removeupgradefailedflag.py
```diff
@@ -0,0 +1,8 @@
+import os
+
+LEAPP_UPGRADE_FAILED_FLAG_FILE = '/var/log/leapp/.leapp_upgrade_failed'
+
+
+def process():
+    if os.path.exists(LEAPP_UPGRADE_FAILED_FLAG_FILE):
+        os.remove(LEAPP_UPGRADE_FAILED_FLAG_FILE)
```
Rezney, line 8: Let's log this action.

repos/system_upgrade/common/actors/removeupgradefailedflag/actor.py
```diff
@@ -0,0 +1,21 @@
...
+    there to prevent rerunning the initramfs part of the upgrade.
+    It needs to be cleaned up before next leapp run.
+    """
+
+    name = 'remove_upgrade_failed_flag'
+    consumes = ()
+    produces = ()
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
```
Rezney, line 18: We are not supposed to touch the system during the `ChecksPhaseTag` (even if it was  created by us). Let's move this to the `InterimPreparationPhase`.

Rezney (changes requested):

repos/system_upgrade/common/actors/removeupgradefailedflag/libraries/removeupgradefailedflag.py
```diff
@@ -0,0 +1,11 @@
+import os
+
+from leapp.libraries.stdlib import api
+
+LEAPP_UPGRADE_FAILED_FLAG_FILE = '/var/log/leapp/.leapp_upgrade_failed'
```
Rezney, line 5: Sorry for late comment. We decided with @pirat89 we would rather put the file into the "/root/tmp_leapp_py3/". Thanks.

matejmatuska: Changed, squashed and tested.

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -338,6 +346,14 @@ mount -o "remount,rw" "$NEWROOT"
 
 ##### do the upgrade #######
 (
+    # check if leapp previously failed in the initramfs, if it did return to the emergency shell
+    [ -f "$NEWROOT$LEAPP_FAILED_FLAG_FILE" ] && {
+        echo "Found file $NEWROOT$LEAPP_FAILED_FLAG_FILE"
+        echo "Leapp previously failed and cannot continue, returning back to emergency shell"
```
Rezney, line 352: Sorry, one more thing. I guess we should advice users to open a BZ and attaching $NEWROOT/var/log/leapp/leapp-upgrade.log (or journal) as they are already messing around in the intiramdisk.
@pirat89 ^^^ wdyt?
I am going to test it again in the meantime.

pirat89, line 352: error msg should be on `stderr` and should start with a prefix that clearly mark the msg as error one. see other error msgs in the module. so each `echo` should be e.g.:
```
echo >&2 "Error: ....."
```

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -268,6 +270,12 @@ do_upgrade() {
         rv=$?
     fi
 
+    if [ "$rv" -ne 0 ]; then
+        # set the upgrade flag to prevent the upgrade from running again when
+        # the emergency shell exits and the upgrade.target is restarted
+        "$NEWROOT/bin/touch" "$NEWROOT$LEAPP_FAILED_FLAG_FILE"
```
pirat89, line 276: ensure the directory exists:
```suggestion
        [ -d "$(dirname $NEWROOT$LEAPP_FAILED_FLAG_FILE)" ] || mkdir "$NEWROOT$LEAPP_FAILED_FLAG_FILE"
        "$NEWROOT/bin/touch" "$NEWROOT$LEAPP_FAILED_FLAG_FILE"
```

repos/system_upgrade/common/actors/removeupgradefailedflag/libraries/removeupgradefailedflag.py
```diff
@@ -0,0 +1,12 @@
+import os
+
+from leapp.libraries.stdlib import api
+
+LEAPP_UPGRADE_FAILED_FLAG_FILE = '/root/tmp_leapp_py3/.leapp_upgrade_failed'
```
pirat89, line 5: remove the whole directory if it exists as it should not be present there anyway. And rename the actor to something like: `removeprevupgradeartifacts`, `removepythonwoarkaounddir`,  .... keeping on you

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
pirat89: also add another echo msg to inform they should open the support support, as when this happens, we cannot provide any specific instructions what to do regarding the situation can be caused by whatever error.

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -268,6 +270,14 @@ do_upgrade() {
         rv=$?
     fi
 
+    if [ "$rv" -ne 0 ]; then
+        # set the upgrade failed flag to prevent the upgrade from running again
+        # when the emergency shell exits and the upgrade.target is restarted
+        [ -d "$(dirname "$NEWROOT$LEAPP_FAILED_FLAG_FILE")" ] || {
+            "$NEWROOT/bin/touch" "$NEWROOT$LEAPP_FAILED_FLAG_FILE"
+        }
```
pirat89, line 278: @matejmatuska now, the `touch` command will be called only if it can fail, which does not make sense. the check has been supposed to be used to trigger the creation of the directory when it does not exist. the current solution does not prevent the loop. check the original suggestion for the hint

matejmatuska: I somehow got it mixed up. It should be alright now. The original suggestion creates the flag file as a directory though, so I changed it to just create the dir.

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -338,6 +349,15 @@ mount -o "remount,rw" "$NEWROOT"
 
 ##### do the upgrade #######
 (
+    # check if leapp previously failed in the initramfs, if it did return to the emergency shell
+    [ -f "$NEWROOT$LEAPP_FAILED_FLAG_FILE" ] && {
+        echo >&2 "Found file $NEWROOT$LEAPP_FAILED_FLAG_FILE"
+        echo >&2 "Error: Leapp previously failed and cannot continue, returning back to emergency shell"
+        echo >&2 "Please file a ticket at http://bugzilla.redhat.com/ with $NEWROOT/var/log/leapp/leapp-upgrade.log attached"
```
Rezney, line 356:
```suggestion
        echo >&2 "Please file a support case with $NEWROOT/var/log/leapp/leapp-upgrade.log attached"
```

Rezney (approved):
Works as expected.

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -338,6 +349,15 @@ mount -o "remount,rw" "$NEWROOT"
 
 ##### do the upgrade #######
 (
+    # check if leapp previously failed in the initramfs, if it did return to the emergency shell
+    [ -f "$NEWROOT$LEAPP_FAILED_FLAG_FILE" ] && {
+        echo >&2 "Found file $NEWROOT$LEAPP_FAILED_FLAG_FILE"
+        echo >&2 "Error: Leapp previously failed and cannot continue, returning back to emergency shell"
+        echo >&2 "Please file a support case with $NEWROOT/var/log/leapp/leapp-upgrade.log attached"
```
pirat89, line 356: Just thinking about that, most likely the rdsosreport.txt file will be needed also. @Rezney  @matejmatuska  wdyt?

matejmatuska: I think rdsosreport.txt could be useful to have.

PR #997, opened 2022-12-07
Add leapp debug tools to initramfs upgrade module

matejmatuska:
Install a script with debug utilities to the Leapp upgrade dracut module.

### DIFF
--- a/repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/90sys-upgrade/leapp_debug_tools.sh
+++ b/repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/90sys-upgrade/leapp_debug_tools.sh
@@ -1,0 +1,38 @@
+# library containing some useful functions for debugging in initramfs
+
+# mounts the sysroot
+leapp_dbg_mount() {
+    systemctl start sysroot.mount
+    mount -o remount,rw /sysroot
+}
+
+# source programs from $NEWROOT, mount if not mounted
+leapp_dbg_source() {
+    systemctl is-active sysroot.mount --quiet || {
+        echo "sysroot not mounted, mounting...";
+        leapp_dbg_mount || return 1
+    }
+
+    for dir in /bin /sbin; do
+        export PATH="$PATH:${NEWROOT}$dir"
+    done
+
+    export LD_LIBRARY_PATH=/sysroot/lib64
+}
+
+# chroot into $NEWROOT
+leapp_dbg_chroot() {
+    systemctl is-active sysroot.mount --quiet || {
+        echo "sysroot not mounted, mounting...";
+        leapp_dbg_mount || return 1
+    }
+
+    for dir in /sys /run /proc /dev /dev/pts; do
+        mount --bind $dir "$NEWROOT$dir"
+    done || { echo "Failed to mount some directories" || return 1 }
+
+    chroot $NEWROOT sh -c "mount -a; /bin/bash"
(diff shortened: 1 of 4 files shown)


matejmatuska:
@vinzenz I believe this is what you had in mind :) ?

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/90sys-upgrade/leapp_debug_tools.sh
```diff
@@ -0,0 +1,34 @@
+# library containing some useful functions for debugging in initramfs
+
+# mounts the sysroot
+leapp_dbg_mount() {
+    systemctl start sysroot.mount
```
vinzenz, line 5: you'd probably want it to be remounted rw

`mount -o remount,rw /sysroot

matejmatuska: I was thinking about it. Okay I will make it rw

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/90sys-upgrade/leapp_debug_tools.sh
```diff
@@ -0,0 +1,34 @@
...
+}
+
+# source programs from $NEWROOT, mount if not mounted
+leapp_dbg_source() {
+    systemctl is-active sysroot.mount --quiet || {
+        echo "sysroot not mounted, mounting...";
+        leapp_dbg_mount || return 1
+    }
+    for dir in /usr/local/bin /usr/local/sbin /usr/bin /usr/sbin; do
+        export PATH="$PATH:${NEWROOT}$dir"
```
vinzenz, line 15: /sysroot/bin and /sysroot/sbin is enough, you won't need /usr/{s,}bin nor /usr/local/{s,}bin

But you probably also want: `export LD_LIBRARY_PATH=/sysroot/lib64`

matejmatuska:
One issue is that some programs added to `$PATH`, such as `file` won't work because they are missing shared libs.

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/90sys-upgrade/leapp_debug_tools.sh
```diff
@@ -0,0 +1,34 @@
...
+    done
+}
+
+# chroot into $NEWROOT
+leapp_dbg_chroot() {
```
vinzenz, line 20: +1

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/90sys-upgrade/leapp_debug_tools.sh
matejmatuska: Yeah that's why the shared libs don't work

matejmatuska:
Also I am thinking about whether the script should be sourced automatically?

Rezney:
> Also I am thinking about whether the script should be sourced automatically?

@matejmatuska as discussed offline, go for the auto-sourcing.

matejmatuska:
The tools script is now auto sourced. Tested in VM, works.

Rezney (approved):
Works like a charm. Well done!

PR #1001, opened 2022-12-21
Don't skip the test that covers the missing repos issue

fernflower:
If some of the repos that are required for the upgrade are missing upgrade should fail early with an inhibitor.

OAMG-6459


fernflower:
Okay, maybe having it as a unit test was not such a great idea after all. For rhel7 we are missing dnf module, for rhel8 we are hitting the issue with mocked context.

pirat89 (approved):
lgtm and seems to be working as expected. I haven't found now any issue

PR #1002, opened 2022-12-22
breadcrumbs: Save breadcrumbs results as RHSM facts

vinzenz:
Additionally to the migration-results file save the breadcrumb results also to the the /etc/rhsm/facts/leapp.facts file in a flattened form.

Jira Issue: OAMG-7773

### DIFF
--- a/commands/upgrade/breadcrumbs.py
+++ b/commands/upgrade/breadcrumbs.py
@@ -110,7 +110,7 @@ class _BreadCrumbs(object):
                 crumbs.truncate()
                 json.dump(doc, crumbs, indent=2, sort_keys=True)
                 crumbs.write('\n')
-                if os.environ.get('LEAPP_NO_RHSM_FACTS', '0') == '1':
+                if os.environ.get('LEAPP_NO_RHSM_FACTS', '0') != '1':
                     self._save_rhsm_facts(doc['activities'])
         except OSError:
             sys.stderr.write('WARNING: Could not write to /etc/migration-results\n')
--- a/commands/upgrade/breadcrumbs.py
+++ b/commands/upgrade/breadcrumbs.py
@@ -3,6 +3,7 @@ import json
 import os
 import sys
 from functools import wraps
+from itertools import chain
 
 from leapp import FULL_VERSION
 from leapp.libraries.stdlib.call import _call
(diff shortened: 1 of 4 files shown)


commands/upgrade/breadcrumbs.py
```diff
@@ -59,6 +110,8 @@ def save(self):
                 crumbs.truncate()
                 json.dump(doc, crumbs, indent=2, sort_keys=True)
                 crumbs.write('\n')
+                if os.environ.get('LEAPP_NO_RHSM_FACTS', '0') == '1':
+                    self._save_rhsm_facts(doc['activities'])
```
pirat89, line 114: It should be opposite. When defined "1", facts should not be shared. Now it means, that facts are shared when `LEAPP_NO_RHSM_FACTS=1`. 

Also we should introduce an option like `--no-rhsm-facts` and probably the use of `--no-rhsm` should involve skipping of the use of RHSM also as: "Use only custom repositories and skip actions with Red Hat Subscription Manager".

pirat89 (changes requested):
one thing. otherwise it seems good. untested yet.

commands/upgrade/breadcrumbs.py
```diff
@@ -59,6 +110,8 @@ def save(self):
                 crumbs.truncate()
                 json.dump(doc, crumbs, indent=2, sort_keys=True)
                 crumbs.write('\n')
+                if os.environ.get('LEAPP_NO_RHSM_FACTS', '0') == '0':
+                    self._save_rhsm_facts(doc['activities'])
```
pirat89, line 114: This is also affected in situation when `LEAPP_NO_RHSM_FACTS=`, it's better to check the valus is not equal `1` so the function is not called iff `LEAPP_NO_RHSM_FACTS=1` 
```suggestion
                if os.environ.get('LEAPP_NO_RHSM_FACTS', '0') != '1':
                    self._save_rhsm_facts(doc['activities'])
```

MichalHe: Suggestion applied :)

commands/upgrade/breadcrumbs.py
MichalHe: Both concerns were addressed with the latest commit.

pirat89:
rebased & fixed conflicts

pirat89 (approved):
seems to be working as expected. merging

PR #1003, opened 2022-12-23
assets: check for what leapp version are assets meant for

MichalHe:
## Introduction of versions to leapp data assets

This patch introduces a mechanism allowing specifying that certain versions of leapp data assets are meant for a particular version of leapp. The motivation is being able to pair assets with a corresponding leapp version that passed QA verification.
To provide this functionality, every asset file was extended with a new JSON field `"provided_data_assets": ["X.Y", "XX.YY"]`. Additional logic checking assets was also added accordingly (`load_asset`). Assets with invalid versions are reported together.

Ref: OAMG-7989

### DIFF
--- a/repos/system_upgrade/common/actors/checkconsumedassets/actor.py
+++ b/repos/system_upgrade/common/actors/checkconsumedassets/actor.py
@@ -1,0 +1,18 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import check_consumed_assets
+from leapp.models import ConsumedDataAsset, Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckConsumedAssets(Actor):
+    """
+    Check whether Leapp is using correct data assets.
+    """
+
+    name = 'check_consumed_assets'
+    consumes = (ConsumedDataAsset,)
+    produces = (Report,)
+    tags = (IPUWorkflowTag, ChecksPhaseTag)
+
+    def process(self):
+        check_consumed_assets.inhibit_if_assets_with_incorrect_version()
--- a/repos/system_upgrade/common/actors/checkconsumedassets/libraries/check_consumed_assets.py
+++ b/repos/system_upgrade/common/actors/checkconsumedassets/libraries/check_consumed_assets.py
@@ -1,0 +1,162 @@
+import re
+from collections import defaultdict, namedtuple
+
+from leapp import reporting
+from leapp.libraries.common.config import get_consumed_data_stream_id
+from leapp.libraries.common.fetch import ASSET_PROVIDED_DATA_STREAMS_FIELD
+from leapp.libraries.stdlib import api
+from leapp.models import ConsumedDataAsset
+
+
+def compose_summary_for_incompatible_assets(assets, incompatibility_reason):
+    if not assets:
+        return []
+
+    summary_lines = ['The following assets are {reason}'.format(reason=incompatibility_reason)]
+    for asset in assets:
+        if asset.provided_data_streams is None:  # Assets with missing streams are placed only in .outdated bucket
+            details = (' - The asset {what_asset} is missing information about provided data streams '
+                       'in its metadata header')
+            details = details.format(what_asset=asset.filename)
+        else:
+            article, multiple_suffix = ('the ', '') if len(asset.provided_data_streams) == 1 else ('', 's')
+            details = ' - The asset {what_asset} provides {article}data stream{mult_suffix} {provided_streams}'
+            details = details.format(what_asset=asset.filename,
+                                     provided_streams=', '.join(asset.provided_data_streams),
+                                     article=article, mult_suffix=multiple_suffix)
+        summary_lines.append(details)
+    return summary_lines
+
+
+def make_report_entries_with_unique_urls(docs_url_to_title_map):
+    report_urls = []
+    # Add every unique asset URL into the report
+    urls_with_multiple_titles = []
+    for url, titles in docs_url_to_title_map.items():
+        if len(titles) > 1:
+            urls_with_multiple_titles.append(url)
+
+        report_entry = reporting.ExternalLink(title=titles[0], url=url)
+        report_urls.append(report_entry)
+
+    if urls_with_multiple_titles:
+        msg = 'Docs URLs {urls} are used with inconsistent URL titles, picking one.'
+        api.current_logger().warning(msg.format(urls=', '.join(urls_with_multiple_titles)))
+
+    return report_urls
+
+
+def report_incompatible_assets(assets):
+    if not any((assets.outdated, assets.too_new, assets.unknown)):
+        return
+
+    title = 'Incompatible Leapp data assets are present'
+
+    docs_url_to_title_map = defaultdict(list)
+    required_data_stream = get_consumed_data_stream_id()
+    summary_prelude = ('The currently installed Leapp consumes data stream {consumed_data_stream}, but the '
+                       'following assets provide different streams:')
+    summary_lines = [summary_prelude.format(consumed_data_stream=required_data_stream)]
+
+    assets_with_shared_summary_entry = [
+        ('outdated', assets.outdated),
+        ('intended for a newer leapp', assets.too_new),
+        ('has an incorrect version', assets.unknown)
+    ]
+
+    doc_url_to_title = defaultdict(list)  # To make sure we do not spam the user with the same URLs
+    for reason, incompatible_assets in assets_with_shared_summary_entry:
+        summary_lines += compose_summary_for_incompatible_assets(incompatible_assets, reason)
+
+        for asset in incompatible_assets:
+            doc_url_to_title[asset.docs_url].append(asset.docs_title)
+
+    report_parts = [
+        reporting.Title(title),
+        reporting.Summary('\n'.join(summary_lines)),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.INHIBITOR, reporting.Groups.REPOSITORY]),
+    ]
+
+    report_parts += make_report_entries_with_unique_urls(docs_url_to_title_map)
+    reporting.create_report(report_parts)
+
+
+def report_malformed_assets(malformed_assets):
+    if not malformed_assets:
+        return
+
+    title = 'Detected malformed Leapp data assets'
+    summary_lines = ['The following assets are malformed:']
+
+    docs_url_to_title_map = defaultdict(list)
+    for asset in malformed_assets:
+        if not asset.provided_data_streams:
+            details = (' - The asset file {filename} contains no values in its "{provided_data_streams_field}" '
+                       'field, or the field does not contain a list')
+            details = details.format(filename=asset.filename,
+                                     provided_data_streams_field=ASSET_PROVIDED_DATA_STREAMS_FIELD)
+        else:
+            # The asset is malformed because we failed to convert its major versions to ints
+            details = ' - The asset file {filename} contains invalid value in its "{data_streams_field}"'
+            details = details.format(filename=asset.filename, data_streams_field=ASSET_PROVIDED_DATA_STREAMS_FIELD)
+        summary_lines.append(details)
+        docs_url_to_title_map[asset.docs_url].append(asset.docs_title)
+
+    report_parts = [
+        reporting.Title(title),
+        reporting.Summary('\n'.join(summary_lines)),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.INHIBITOR, reporting.Groups.REPOSITORY]),
+    ]
+
+    report_parts += make_report_entries_with_unique_urls(docs_url_to_title_map)
+    reporting.create_report(report_parts)
+
+
+def inhibit_if_assets_with_incorrect_version():
+    required_data_stream = get_consumed_data_stream_id()
+    required_data_stream_major = int(required_data_stream.split('.', 1)[0])
+
+    # The assets are collected according to why are they considered incompatible, so that a single report is created
+    # for every class
+    IncompatibleAssetsByType = namedtuple('IncompatibleAssets', ('outdated', 'too_new', 'malformed', 'unknown'))
+    incompatible_assets = IncompatibleAssetsByType(outdated=[], too_new=[], malformed=[], unknown=[])
+
+    datastream_version_re = re.compile(r'\d+\.\d+$')
+
+    for consumed_asset in api.consume(ConsumedDataAsset):
+        if consumed_asset.provided_data_streams is None:  # There is no provided_data_streams field
+            # Most likely an old file that predates the introduction of versioning to data assets
+            incompatible_assets.outdated.append(consumed_asset)
+            continue
+
+        # Ignore minor stream numbers and search only for a stream matching the same major number
+        if all((datastream_version_re.match(stream) for stream in consumed_asset.provided_data_streams)):
+            provided_major_data_streams = sorted(
+                int(stream.split('.', 1)[0]) for stream in consumed_asset.provided_data_streams
+            )
+        else:
+            incompatible_assets.malformed.append(consumed_asset)
+            continue
+
+        if required_data_stream_major in provided_major_data_streams:
+            continue
+
+        if not provided_major_data_streams:
+            # The field contained [], or something that was not a list, but it was corrected to [] to satisfy model
+            incompatible_assets.malformed.append(consumed_asset)
+            continue
+
+        if required_data_stream_major > max(provided_major_data_streams):
+            incompatible_assets.outdated.append(consumed_asset)
+        elif required_data_stream_major < min(provided_major_data_streams):
+            incompatible_assets.too_new.append(consumed_asset)
+        else:
+            # Since the `provided_data_vers` is a list of values, it is possible that the asset provide, e.g., 4.0
+            # and 6.0,  but the leapp consumes 5.0, thus we need to be careful when to say that an asset is too
+            # new/outdated/none.
+            incompatible_assets.unknown.append(consumed_asset)
+
+    report_incompatible_assets(incompatible_assets)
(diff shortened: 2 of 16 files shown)


MichalHe:
# Conducted testing (rhel8 dev vagrant box)
- [x] All assets have a version field, that matches the current leapp version.
- [x] A single asset does not have a version field
  - `Inhibitor: Repositories mapping data is outdated`
- [x] A single asset is outdated (`"leapp_version": "0.16.0"`, the installed leapp is `0.17.0`)
  - `Inhibitor: Repositories mapping is outdated`
- [x] A single asset has a newer version than the current leapp (`"leapp_version": "0.18.0"`, the installed leapp is `0.17.0`)
  -  `Inhibitor: PES events file is intended for a newer leapp`
- [x] All assets are missing, and thus, they need to be fetched from the online service (service serves assets with correct versions). A mock server running on the testing machine on port 8888 was set up. The server offered only one endpoint template `@app.route('/api/pes/<leapp_version>/<data_file>')`. Leapp was executed as follows: `LEAPP_SERVICE_HOST=http://0.0.0.0:8888 leapp preupgrade --debug`. The following endpoints wwere were probed by leapp: 
```
        GET /api/pes/0.17.0/repomap.json HTTP/1.1
        GET /api/pes/0.17.0/pes-events.json HTTP/1.1
	GET /api/pes/0.17.0/device_driver_deprecation_data.json HTTP/1.1
```

repos/system_upgrade/common/libraries/fetch.py
```diff
@@ -108,3 +121,59 @@ def read_or_fetch(filename, directory="/etc/leapp/files", service=None, allow_em
...
+    :param str asset_filename: The file name of the asset to load.
+    :param str asset_fulltext_name: A human readable asset name to display in error messages.
+    :param str docs_url: Docs url to provide if an asset is malformed or outdated.
+    :param str docs_title: Title of the documentation to where `docs_url` points to.
+    :returns: A dict with asset contents (a parsed JSON), or None if the asset was outdated.
+    """
+
+    # Check that the actor that is attempting to obtain the asset meets the contract to call this function
+    # @Note(mhecko): Assertion is fine, as the assertion enforces a 'static' contract between the callee and the caller
+    assert models.LeappRepositoryInfo in actor_requesting_asset.consumes
```
Rezney, line 146: @pirat89 what is our position on using asserts outside of tests?

repos/system_upgrade/common/actors/scanleappinfo/actor.py
```diff
@@ -0,0 +1,16 @@
...
+from leapp.models import InstalledRPM, LeappRepositoryInfo
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+
+
+class ScanLeappInfo(Actor):
+    """Scan what leapp framework and repository is installed."""
+
+    name = 'scann_leapp_info'
+    consumes = (InstalledRPM,)
+    produces = (LeappRepositoryInfo,)
```
Rezney, line 12:
```suggestion
    produces = (LeappRepositoryRPMInfo,)
```

repos/system_upgrade/common/actors/checkconsumedassets/actor.py
```diff
@@ -0,0 +1,18 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import check_consumed_assets
+from leapp.models import ConsumedDataAsset, LeappRepositoryInfo, Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckConsumedAssets(Actor):
+    """
+    Check that leapp used correct data assets.
```
Rezney, line 9:
```suggestion
    Check whether Leapp is using correct data assets.
```

repos/system_upgrade/common/actors/checkconsumedassets/libraries/check_consumed_assets.py
```diff
@@ -0,0 +1,64 @@
...
+def inhibit_if_assets_with_incorrect_version():
+    leapp_info = next(api.consume(LeappRepositoryInfo), None)
+    if not leapp_info:
+        raise StopActorExecutionError('Did not receive any LeappRepositoryInfo messages.')
+
+    version_re = '([0-9]*)\\.([0-9]*)\\.([0-9]*)'
+    for consumed_asset in api.consume(ConsumedDataAsset):
+        if not consumed_asset.target_leapp_version:
+            # Most likely an old file that predates the introduction of versioning to data assets
+            summary = 'The asset {0} is missing target leapp version in its metadata'.format(consumed_asset.filename)
```
Rezney, line 18:
```suggestion
            summary = 'The asset {0} is missing target Leapp version in its metadata header'.format(consumed_asset.filename)
```

repos/system_upgrade/common/actors/checkconsumedassets/libraries/check_consumed_assets.py
```diff
@@ -0,0 +1,64 @@
...
+        if version_matches_expected_form:
+            # We need to parse in order to tell user more than just the asset version is wrong
+            asset_leapp_version = tuple(int(group) for group in version_matches_expected_form.groups())
+        else:
+            summary = '{0} has invalid value specifying leapp version for which it is meant.'
+            reporting.create_report([
+                reporting.Title('The leapp data file {0} is malformed'.format(consumed_asset.filename)),
+                reporting.Summary(summary.format(consumed_asset.fulltext_name)),
+                reporting.ExternalLink(title=consumed_asset.docs_title, url=consumed_asset.docs_url),
+                reporting.Severity(reporting.Severity.MEDIUM),
```
Rezney, line 38: Inhibitor report entry should always have a "HIGH" priority.

repos/system_upgrade/common/actors/checkconsumedassets/libraries/check_consumed_assets.py
```diff
@@ -0,0 +1,64 @@
...
+                reporting.Groups([reporting.Groups.INHIBITOR, reporting.Groups.REPOSITORY]),
+            ])
+            continue
+
+        version_matches_expected_form = re.match(version_re, consumed_asset.target_leapp_version)
+        if version_matches_expected_form:
+            # We need to parse in order to tell user more than just the asset version is wrong
+            asset_leapp_version = tuple(int(group) for group in version_matches_expected_form.groups())
+        else:
+            summary = '{0} has invalid value specifying leapp version for which it is meant.'
```
Rezney, line 33:
```suggestion
            summary = 'Invalid value detected in the version header of {0}.'
```

Rezney (dismissed):

Rezney:
@MichalHe we will want to discuss this solution with @pirat89. Let's have a quick sync next week.

repos/system_upgrade/common/libraries/fetch.py
vinzenz: It is a general no, no

Asserts should not be used, stopactorexecutionerror is your friend here

MichalHe: It is not even needed from the library to _enforce_ that the actors using the library are consuming specific content, so the assertion can be dropped all together, especially when I consider that there will not be a lot of people writing custom actors which use data files. On the other hand, if the code _assumes_ something it should be explicit, and thus I would rather make StopActorExecution there.

My curiosity then stumbles upon the reasons against using assertions. Why shouldn't one use assertions in this particular scenario?

repos/system_upgrade/common/actors/checkconsumedassets/actor.py
MichalHe: Resolved with the latest push as suggested.

repos/system_upgrade/common/actors/checkconsumedassets/libraries/check_consumed_assets.py
MichalHe: Resolved with the latest push as suggested.

repos/system_upgrade/common/actors/checkconsumedassets/libraries/check_consumed_assets.py
MichalHe: Resolved with the latest push as suggested.

repos/system_upgrade/common/actors/checkconsumedassets/libraries/check_consumed_assets.py
MichalHe: Fixed.

repos/system_upgrade/common/libraries/fetch.py
```diff
@@ -108,3 +121,59 @@ def read_or_fetch(filename, directory="/etc/leapp/files", service=None, allow_em
...
     return response.content.decode(encoding)
+
+
+def load_data_asset(actor_requesting_asset,
+                    asset_filename,
+                    asset_fulltext_name,
+                    docs_url,
+                    docs_title):
+    """
+    Load the contents of the data asset with given asset_filename
```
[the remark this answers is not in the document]
MichalHe: Resolved with the latest push as suggested.

repos/system_upgrade/common/actors/scanleappinfo/actor.py
```diff
@@ -0,0 +1,16 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import scan_leapp_info as scan_leapp_info_lib
+from leapp.models import InstalledRPM, LeappRepositoryInfo
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+
+
+class ScanLeappInfo(Actor):
+    """Scan what leapp framework and repository is installed."""
```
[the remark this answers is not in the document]
MichalHe: Resolved with the latest push as suggested.

repos/system_upgrade/common/actors/scanleappinfo/actor.py
MichalHe: model renamed with the latest push.

repos/system_upgrade/common/libraries/fetch.py
pirat89: long story short:
when you want to use something what is reliable, do not use assert to stop the execution in any code; it'S good for experimenting, testing, ... not for the use in something what you expect to be reliable or have a defined behaviour on all machines.

short story long:
asserts should be really used just in tests. Using asserts in the application itself is bad practice - usually inherited from old programming styles / projects. When you specify assert, you even do not know whether it will be really executed, as you can run python (or whatever language) in the way that all asserts are stripped or ignored. This leads to the situations when the code execution could lead to undefined situations / various different behaviours, and it could be then funny to debug something on your system why is something happening, when someone years ago thought that the problem is clearly caught by an assert. It's also nice source of security issues :)

Some examples:
* https://python.plainenglish.io/if-you-are-using-assert-you-must-read-this-7256e46dc816
* https://towardsdatascience.com/2-caveats-to-avoid-when-using-python-assert-statement-8060e123c26a

MichalHe: Thanks for a nice explanation, TIL you can disable assertions globally.

I do not think that the purpose of the asserts in the snippet might lead to the problems described in the linked articles as they were not meant to validate user supplied data (or any kind of data for that matter). Their purpose was to provide a helping hand to the programmer trying to rely on the library, by making sure that the actor in which the library is used consumes and produces everything necessary. The assertions are not even needed -- if an actor does not consume what is needed, then the library will raise a StopActorException about not receiving any LeappRepositoryRPMInfo. Thus the programmer has to think why was the exception raised whereas now the library tells him the first time he tries executing the actor containing a call to that function that he might be  forgetting to specify something in actor's consumes/produces. Leapp end users should not achieve anything by disabling assertions, because it is there to help the developer rather than to make sure that some data are not as expected.

That being said, using whatever kind of exception is fine as long it provides the same value to the developer. I would be even fine with dropping the exception all together based on a sole argument of how many custom actors might want to load data assets. The assertions have been removed in favor of StopActorExecution with the latest push.

repos/system_upgrade/common/actors/checkconsumedassets/libraries/check_consumed_assets.py
```diff
@@ -0,0 +1,128 @@
...
+from leapp.libraries.common.fetch import ASSET_PROVIDED_DATA_STREAMS_FIELD
+from leapp.libraries.stdlib import api
+from leapp.models import ConsumedDataAsset
+
+
+def report_incompatible_asset_of_same_type(incompatibility_characterization, assets, are_assets_malformed=False):
+    if not assets:
+        return
+
+    title = 'The asset{multiple_suffix} {which_assets} {verb} {reason}'
```
pirat89, line 14: title should not be affected singular & plural forms. The difference between one asset or several assets are e.g. incompatible, is irrelevant. The issue is same, it has only a different parameters. Imagine having a script that is detecting whether this issue has been hit or not. User is supposed to detect the particular problem in the report checking the `key` value. Which is also affected by the `title`, because the title is supposed to be invariant as  the problem. Keeping the review for the next week, but the function seems to me too much generic just from a short look.

MichalHe: Thanks for the detailed explanation on static report titles, I've gained another piece of knowledge and perspective.

I've crystalized the generic function into two separate functions, hoping to make reporting less convoluted.

fernflower:
@pirat89 @MichalHe I am a bit confused - this PR ties leapp asset version to leapp version, but in the Friday discussion with @pirat89 it was stated that we are going to introduce a separate asset versioning in X.Y. format (starting with '1.0'). Is it me getting something wrong or this PR hasn't been updated yet?

MichalHe:
@fernflower Hi, sorry, I forgot to update the PR description. The code is using a separate versioning format - long story short, it check for the presence of `CONSUMED_DATA_STREAM_ID` constant the `provided_data_stream` of every loaded asset.

repos/system_upgrade/common/actors/checkconsumedassets/libraries/check_consumed_assets.py
pirat89: @MichalHe it seems really good now! thanks :)

fernflower:
Okay, having tried this with updated leapp-data I managed to crash the actor
```
        Check whether Leapp is using correct data assets.
Process Process-290:
    self.run()
  File "/usr/lib/python3.6/site-packages/leapp/repository/actor_definition.py", line 74, in _do_run
  File "/usr/lib/python3.6/site-packages/leapp/actors/__init__.py", line 289, in run
    self.process(*args)
  File "/etc/leapp/repos.d/system_upgrade/common/actors/checkconsumedassets/actor.py", line 18, in process
    check_consumed_assets.inhibit_if_assets_with_incorrect_version()
  File "/etc/leapp/repos.d/system_upgrade/common/actors/checkconsumedassets/libraries/check_consumed_assets.py", line 162, in inhibit_if_assets_with_incorrect_version
    report_malformed_assets(incompatible_assets.malformed)
  File "/etc/leapp/repos.d/system_upgrade/common/actors/checkconsumedassets/libraries/check_consumed_assets.py", line 113, in report_malformed_assets
    report_parts += make_report_entries_with_unique_urls(docs_url_to_title_map)
  File "/etc/leapp/repos.d/system_upgrade/common/actors/checkconsumedassets/libraries/check_consumed_assets.py", line 44, in make_report_entries_with_unique_urls
    api.current_logger().warning(msg.format(', '.join(urls_with_multiple_titles)))
KeyError: 'urls'


======================================================================================================
Actor check_consumed_assets unexpectedly terminated with exit code: 1 - Please check the above details
======================================================================================================

```

repos/system_upgrade/common/actors/checkconsumedassets/libraries/check_consumed_assets.py
```diff
@@ -0,0 +1,162 @@
...
+    for url, titles in docs_url_to_title_map.items():
+        if len(titles) > 1:
+            urls_with_multiple_titles.append(url)
+
+        report_entry = reporting.ExternalLink(title=titles[0], url=url)
+        report_urls.append(report_entry)
+
+    if urls_with_multiple_titles:
+        msg = 'Docs URLs {urls} are used with inconsistent URL titles, picking one.'
+        api.current_logger().warning(msg.format(', '.join(urls_with_multiple_titles)))
```
fernflower, line 44: Missing `urls=','.join(urls_with_multiple_titles)` in call to format.

repos/system_upgrade/common/actors/checkconsumedassets/libraries/check_consumed_assets.py
```diff
@@ -0,0 +1,162 @@
...
+
+    summary_lines = ['The following assets are {reason}'.format(reason=incompatibility_reason)]
+    for asset in assets:
+        if asset.provided_data_streams is None:  # Assets with missing streams are placed only in .outdated bucket
+            details = (' - The asset {what_asset} is missing information about provided data streams '
+                       'in its metadata header')
+            details = details.format(what_asset=asset.filename)
+        else:
+            article, multiple_suffix = ('the ', '') if len(asset.provided_data_streams) == 1 else ('', 's')
+            details = ' - The asset {what_asset} provides data {article}stream{mult_suffix} {provided_streams}'
```
fernflower, line 23: Doesn't provide a nice message in all cases, for example for a single from-the-future asset 42.42:
```
"summary": "The currently installed Leapp consumes data stream 1.0, but the following assets provide different streams:\nThe following assets are intended for a newer leapp\n - The asset pes-events.json provides data the stream 42.42"
```
Note the "The asset pes-events.json provides **data the stream 42.42**"

fernflower (dismissed):
A unit test needs to be added to cover the crash.
(I might add it myself, just taking a note)

repos/system_upgrade/common/actors/checkconsumedassets/libraries/check_consumed_assets.py
pirat89: fixed

fernflower (approved):

PR #848, opened 2022-03-09
packit: build SRPM in Copr

TomasTomecek:
TL;DR:
 * specify the list of deps needed to prepare SRPM in the .packit file
 * move leapp-el7toel8-deps.spec into other_specs subdir
 * update the make commands for the building of packages

Previously we have prepared SRPM "locally" and then uploaded it
to the COPR server for the building. However, since Jan 2023 we are
required to build SRPMs on the COPR server itself. As in this project
we have to manage 2 SPEC files for two buildings:
  - leapp-el7toel8-deps.spec
        Creating metapackage resolving issues with dependencies
        on the target system; created RPMs are archived and are used
        as one of sources for building of leapp-repository.
  - leapp-repository.spec
        The main spec file used for the entire project, which requires
        archive with the deps packages.
Currently it's not possible to tell COPR which specific spec file
should be used and if they are in the same directory, COPR fails with
the error message about multiple SPEC files. But COPR is ok having
multiple spec files in a project when they are in separate directories.
To fix that, we are moving the deps spec file into the separate directory.

Also, explicitly set the `_rpmfilename` macro. This is super important as
the COPR build servers are using Mock, which redefines the macro, so packages
are stored inside RPMS directory, instead of `RPMS/%{ARCH}`. The macro must be
defined with double '%'. Using just single %, the macro is expanded when
the specfile is loaded, but it is expected to be expanded during
the build process when particular subpackages (RPMs) are created, so
each RPM has the right name. Using the single %, all RPMs would have the
name of the SRPM - which means effectively that only one RPM per build
would be created. (hopefully the explanation is clear :))

PR #1005, opened 2023-01-02
Packit: specify srpm_build_deps to allow builds with Packit again

pirat89:
The option / list is mandatory now for Packit, however we can keep it empty for now as it looks like as we build the srpm via the Makefile still. The same is used now in the leapp project.

It's expected we update the list in upcoming months most likely, when we integrate this project better with the current Packit best practices.


pirat89:
Closing this one. Follow https://github.com/oamg/leapp-repository/pull/848 for more details

PR #1006, opened 2023-01-11
check grubenv length

bessonc:
A customer had a /boot/efi/EFI/redhat/grubenv corrupted by a trailing line feed ('\n' or 0x0a).
That leads to the below error:
```
leapp.workflow.InitRamStart.remove_upgrade_boot_entry: External command has started: ['/usr/sbin/grubby', '--remove-kernel=/boot/vmlinuz-upgrade.x86_64']
leapp.workflow.InitRamStart.remove_upgrade_boot_entry: grub2-editenv: error: environment block too small.
```

After this IPU 7->8, the system reboots on the el7 kernel.

The below patch checks whether grubenv is 1K-length or inhibits the upgrade with the following message:
```
Risk Factor: high (inhibitor)
Title: Detected a corrupted grubenv file.
Summary: The file /boot/grub2/grubenv is corrupted. Please delete it and then regenerate it using grub2-mkconfig.
```

Tested only on a legacy BIOS VM.

Resolves rhbz#2159288


repos/system_upgrade/common/actors/detectcorruptedgrubenv/actor.py
```diff
@@ -0,0 +1,43 @@
...
+        config = '/boot/grub2/grubenv'
+        for fact in self.consume(FirmwareFacts):
+            if fact.firmware == 'efi':
+                config = '/boot/efi/EFI/redhat/grubenv'
+                break
+        if is_grubenv_corrupted(config):
+            create_report([
+                reporting.Title('Detected a corrupted grubenv file.'),
+                reporting.Summary(
+                    'The file %s is missing or corrupted. '
```
rmetrich, line 32: IMHO not having the file at all should not be considered as an error (proper handling is made in `grub.cfg`).

repos/system_upgrade/common/actors/detectcorruptedgrubenv/libraries/detectcorruptedgrubenv.py
```diff
@@ -0,0 +1,15 @@
+import os
+
+
+def _get_config_contents(config_path):
+    if os.path.exists(config_path):
```
rmetrich, line 5: See above, IMHO it should not fail on missing file.

rmetrich (changes requested):

rmetrich (dismissed):

rmetrich:
@pirat89 ^^ :-)

bessonc:
review please

repos/system_upgrade/common/actors/detectcorruptedgrubenv/actor.py
```diff
@@ -0,0 +1,43 @@
...
+                config = '/boot/efi/EFI/redhat/grubenv'
+                break
+        if is_grubenv_corrupted(config):
+            create_report([
+                reporting.Title('Detected a corrupted grubenv file.'),
+                reporting.Summary(
+                    'The file %s is corrupted. '
+                    'Please delete it and then regenerate it using grub2-mkconfig.'
+                    % config
+                ),
```
pirat89, line 35: thanks for the contribution!!
we should provide a little bit better explanation here what the *corrupted* means (i do not mean to have a dynamic text in this case, but rather a generic explanation `(the file has to have exact size and cannot end with the newline)  The corruption could be caused e.g. by a manual modification of the file which is not recommended.` and adding the information that the file must be valid to pass the upgrade correctly. The rest of the text should be under `reporting.Remediation` (with the hint parameter), to explan people what they are expected to do (skipping the "Please" word), e.g. `Regenerate the file using the grub2-mkconfig tool.` In case it's expect the command to be always same and safe, the command could be mentioned covered by `commands` field.
```python
reporting.Remediation(hint="text", commands=[[cmd, 'param', ......]])
```

Writting in rush now, I/We will look at it properly during the next two weeks. I am not sure now whether this will work well for hybrid boots as the configuration files have different paths based not only on the thing whether the efi is or not detected.

bessonc:
@fernflower yes feel free to do so! I understand that refactoring

fernflower:
Thank you! :)

PR #1008, opened 2023-01-16
pes_events_scanner: prefilter problematic events and add back logging

MichalHe:
The simplistic approach taken by the current PES implementation has problems with with events that overlap, e.g. A=(from=8.0, to=8.8) and B=(from=8.1,  to=8.7). This patch addresses only addresses a scenario when the overlapping packages have the same from release, dropping the event with lower to_release. Addressing the problem in general leads to too much changes in the way the core algorithm works, however, the solution is very cool.

jira: OAMG-8182, OAMG-8221
fixes: bz#2158527

### DIFF
--- a/repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
+++ b/repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
@@ -1,4 +1,4 @@
-from collections import namedtuple
+from collections import defaultdict, namedtuple
 from functools import partial
 
 from leapp import reporting
@@ -126,6 +126,7 @@ def compute_pkg_changes_between_consequent_releases(source_installed_pkgs,
                                                     release,
                                                     seen_pkgs,
                                                     pkgs_to_demodularize):
+    logger = api.current_logger()
     # Start with the installed packages and modify the set according to release events
     target_pkgs = set(source_installed_pkgs)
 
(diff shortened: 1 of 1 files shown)


Rezney:
@MichalHe Please format the pkg name output as string and it should be good to go:

~~~
2023-01-19 11:03:09.166 DEBUG    PID: 15656 leapp.workflow.FactsCollection.pes_events_scanner: Applying event 5933 (Action.REPLACED): replacing packages set([libstoragemgmt-python-clibs:rhel7-base]) with set([python3-libstoragemgmt-clibs:rhel8-BaseOS])
~~~

Unless @pirat89 has anything else...

MichalHe:
@Rezney I added the requested string formatting. However, I avoided adding too much logic handling the debug message formatting, because it is just a debug message. IMHO, if it tells what decisions were made and why during the run in a sane manner it is fine.

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
```diff
@@ -163,6 +170,55 @@ def compute_pkg_changes_between_consequent_releases(source_installed_pkgs,
...
+            events_with_same_in_pkgs_and_from_release[(event.from_release, tuple(sorted(event.in_pkgs)))].append(event)
+
+    cleaned_events = []
+    for from_release_in_pkgs_pair, problematic_events in events_with_same_in_pkgs_and_from_release.items():
+        if len(problematic_events) == 1:
+            cleaned_events.append(problematic_events[0])  # There is no problem
+            continue
+
+        # E.g., one of the problematic events is to=8.6, other one to=8.7, keep only 8.7
+        from_release, dummy_in_pkgs = from_release_in_pkgs_pair
```
pirat89, line 199: @MichalHe  i am ok with it, but in this case I would not even create `dummy_in_pkgs` var as it's clear what is the second item in the tuple regarding the name, and usually this (`x, dummy_y = z`) way is better when you want to clearly specify what is the other item for a possible later use. so no problem with it at all, just wanted to give you a tip. keep it as it is now. most likely we are going to finish the merge in upcoming minutes :) i am jut finishing my "painting" of graphs

pirat89 (approved):
Seems good to me and I like the actual solution. Waiting for the test results before the merge.

PR #1009, opened 2023-01-17
Overshadowing mount points in /etc/fstab

PeterMocary:
Checks the order of entries in /etc/fstab based on their mount point and inhibits upgrade if overshadowing is detected.

Jira ref.: OAMG#5090
Bugzilla: [BZ#1972238](https://bugzilla.redhat.com/show_bug.cgi?id=1972238)

### DIFF
--- a/repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
+++ b/repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
@@ -4,6 +4,8 @@ from leapp import reporting
 from leapp.libraries.stdlib import api
 from leapp.models import StorageInfo
 
+FMT_LIST_SEPARATOR = '\n    - '
+
 
 def _get_common_path(path1, path2):
     """
@@ -31,11 +33,10 @@ def _get_overshadowing_mount_points(mount_points):
     """
     Retrieve set of overshadowing and overshadowed mount points.
 
-    :param list[str] mount_points: absolute paths to mount points
+    :param list[str] mount_points: absolute paths to mount points without trailing /
     :returns: set of unique mount points without trailing /
     """
     overshadowing = set()
-    mount_points = [mp.rstrip('/') for mp in mount_points]
     for i, mount_point in enumerate(mount_points):
         for overshadowing_mount_point in mount_points[i+1:]:
             if _get_common_path(mount_point, overshadowing_mount_point) == overshadowing_mount_point:
@@ -50,7 +51,14 @@ def check_fstab_mount_order():
     if not storage_info:
         return
 
-    mount_points = [fstab_entry.fs_file.rstrip('/') for fstab_entry in storage_info.fstab]
+    mount_points = []
+    for fstab_entry in storage_info.fstab:
+        mount_point = fstab_entry.fs_file
+        if mount_point != '/':
+            mount_point = mount_point.rstrip('/')
+        if os.path.isabs(mount_point):
+            mount_points.append(mount_point)
+
     overshadowing = _get_overshadowing_mount_points(mount_points)
     duplicates = {mp for mp in mount_points if mount_points.count(mp) > 1}
 
@@ -63,15 +71,15 @@ def check_fstab_mount_order():
     hint = 'To prevent the overshadowing:'
 
     if duplicates:
-        summary += '\nDetected mount points with duplicates:\n{}'.format(', '.join(duplicates))
-        hint += '\nRemove detected duplicates by using unique mount points.'
+        summary += '\nDetected mount points with duplicates: {}'.format(', '.join(duplicates))
+        hint += ' Remove detected duplicates by using unique mount points.'
 
     if overshadowing:
-        summary += '\nDetected order of overshadowing mount points:\n{}'.format(', '.join(overshadowing_in_order))
+        summary += '\nDetected order of overshadowing mount points: {}'.format(', '.join(overshadowing_in_order))
         hint += (
-            '\nReorder the detected overshadowing entries.'
-            '\nPossible order of all mount points without overshadowing:\n{}'
-        ).format(', '.join(overshadowing_fixed))
+            ' Reorder the detected overshadowing entries. Possible order of all mount '
+            'points without overshadowing:{}{}'
+        ).format(FMT_LIST_SEPARATOR, FMT_LIST_SEPARATOR.join(overshadowing_fixed))
 
     reporting.create_report([
         reporting.Title(
--- a/repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
+++ b/repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
@@ -1,0 +1,93 @@
+import os
+
+from leapp import reporting
+from leapp.libraries.stdlib import api
+from leapp.models import StorageInfo
+
+
+def _get_common_path(path1, path2):
+    """
+    Return the longest common absolute sub-path for pair of given absolute paths.
+
+    Note that this function implements similar functionality as os.path.commonpath(), however this function is not
+    available in python2.7, thus can't be used here.
+    """
+
+    if path1 == '' or path2 == '':
+        return ''
+
+    path1 = path1.strip('/').split('/')
+    path2 = path2.strip('/').split('/')
+
+    common_path = []
+    for path1_part, path2_part in zip(path1, path2):
+        if path1_part != path2_part:
+            break
+        common_path.append(path1_part)
+    return '/' + '/'.join(common_path)
+
+
+def _get_incorrectly_ordered_fstab_entries(fstab_entries):
+    """
+    Retrieve pairs of incorrectly ordered entries in /etc/fstab based on mount point overshadowing.
+
+    :param list[FstabEntry] fstab_entries: fstab entries from StorageInfo
+    :returns: pairs of fstab entries with incorrect order
+    """
+
+    for i, fstab_entry in enumerate(fstab_entries):
+        mount_point = os.path.abspath(fstab_entry.fs_file)
+
+        for overshadowing_fstab_entry in fstab_entries[i+1:]:
+            overshadowing_mount_point = os.path.abspath(overshadowing_fstab_entry.fs_file)
+
+            if _get_common_path(mount_point, overshadowing_mount_point) == overshadowing_mount_point:
+                yield fstab_entry, overshadowing_fstab_entry
+
+
+def check_fstab_mount_order():
+    storage_info = next(api.consume(StorageInfo), None)
+
+    if storage_info:
+        overshadowing_pairs = list(_get_incorrectly_ordered_fstab_entries(storage_info.fstab))
+        if overshadowing_pairs:
+            mount_points = [fstab_entry.fs_file for fstab_entry in storage_info.fstab]
+
+            duplicates = set()
+            overshadowing = set()
+            for overshadowed_entry, overshadowing_entry in overshadowing_pairs:
+                if overshadowed_entry == overshadowing_entry:
+                    duplicates.add(overshadowed_entry.fs_file)
+                overshadowing.add(overshadowed_entry.fs_file)
+                overshadowing.add(overshadowing_entry.fs_file)
+
+            overshadowing_in_order = []
+            for mount_point in mount_points:
+                if mount_point in overshadowing:
+                    overshadowing_in_order.append(mount_point)
+
+            summary = (
+                'Leapp detected incorrect order of entries in /etc/fstab that causes '
+                'overshadowing of mount points.\nDetected order of overshadowing mount points:\n{}'
+            ).format(', '.join(overshadowing_in_order))
+
+            if duplicates:
+                summary += ',\nDetected duplicate mount points:\n{}'.format(', '.join(duplicates))
+
+            hint = (
+                'To prevent the overshadowing reorder the entries in /etc/fstab and remove any duplicates. '
+                'Possible order of mount points without overshadowing:\n{}'
+            ).format(', '.join(sorted(overshadowing, key=len)))
+
+            reporting.create_report([
+                reporting.Title(
+                    'Detected incorrect order of entries or duplicate entries in /etc/fstab, preventing a successful '
+                    'in-place upgrade.'
+                ),
+                reporting.Summary(summary),
+                reporting.Remediation(hint=hint),
+                reporting.RelatedResource('file', '/etc/fstab'),
+                reporting.Severity(reporting.Severity.HIGH),
+                reporting.Groups([reporting.Groups.FILESYSTEM]),
+                reporting.Groups([reporting.Groups.INHIBITOR]),
+            ])
--- a/repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
+++ b/repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
@@ -1,5 +1,3 @@
-import os
-
 from leapp import reporting
 from leapp.libraries.stdlib import api
 from leapp.models import StorageInfo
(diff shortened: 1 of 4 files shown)


repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
```diff
@@ -0,0 +1,51 @@
...
+                    reporting.Title(
+                        'Detected incorrect order of entries in /etc/fstab, preventing a successful in-place upgrade.'
+                    ),
+                    reporting.Summary(summary.format(overshadowed_entry.fs_file, overshadowing_entry.fs_file)),
+                    reporting.Remediation(hint=hint.format(overshadowed_entry.fs_file, overshadowing_entry.fs_file)),
+                    reporting.RelatedResource('file', '/etc/fstab'),
+                    reporting.Severity(reporting.Severity.HIGH),
+                    reporting.Groups([reporting.Groups.FILESYSTEM]),
+                    reporting.Groups([reporting.Groups.INHIBITOR]),
+                ])
```
pirat89, line 51: This could lead to the situation when multiple reports are generated, which we do not want. The expectations it that only one report at maximum will inform about all problems with the order of entries in /etc/fstab. So having e.g.
```
/var/log/leapp
/var/log
/var
/
```
will generate one report, informing user about all entries in incorrect order, with the hint how to fix the order.

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
```diff
@@ -0,0 +1,51 @@
...
+        mount_point = fstab_entry.fs_file
+
+        for overshadowing_fstab_entry in fstab[i:]:
+            overshadowing_mount_point = overshadowing_fstab_entry.fs_file
+            if mount_point != overshadowing_mount_point and mount_point.startswith(overshadowing_mount_point):
+                yield fstab_entry, overshadowing_fstab_entry
+
+
+def check_fstab_mount_order():
+    critical_mount_points = ['/var/log', '/var/lib', '/var/lib/leapp']
```
pirat89, line 24: we should not care about critical mountpoints in this case. the very same problem we could find in other cases, e.g.
/usr/lib
/usr
...

and all other partitions that exists on the system. the actor is supposed to ensure, that all entries in fstab are in the correct order.

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
PeterMocary: Changed the behavior, so that there is a single report created with a list of overshadowing mount points in detected order and, in the hint, a possible order without overshadowing. I also added a list of duplicate mount points since these shouldn't be in the /etc/fstab as well.

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
PeterMocary: Thanks for clearing that up to me, it's fixed now.

PeterMocary:
Added a fixup commit with slight refactoring, which simplifies the actor a bit.

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
```diff
@@ -0,0 +1,102 @@
...
+
+def _get_common_path(path1, path2):
+    """
+    Return the longest common absolute sub-path for pair of given absolute paths.
+
+    Note that this function implements similar functionality as os.path.commonpath(), however this function is not
+    available in python2.7, thus can't be used here.
+    """
+
+    if path1 == '' or path2 == '':
```
fernflower, line 14: This `path == ''` comparison catches the eye tbh.. Can we change it to `if not path1 or not path2` without breaking the logic?

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    Retrieve set of overshadowing and overshadowed mount points.
+
+    :param list[str] mount_points: absolute paths to mount points
+    :returns: set of mount points
+    """
+
+    overshadowing = set()
+    for i, mount_point in enumerate(mount_points):
+        if mount_point[-1] == '/':
```
fernflower, line 38: If the idea is to get rid of trailing '/', then you can do that at L37:

`for i, mount_point in enumerate(mp.rstrip('/') for mp in mount_points)`

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
```diff
@@ -0,0 +1,102 @@
...
+            mount_point = mount_point[:-1]
+
+        for overshadowing_mount_point in mount_points[i+1:]:
+            if overshadowing_mount_point[-1] == '/':
```
fernflower, line 42: Same here iiuc

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
```diff
@@ -0,0 +1,102 @@
...
+    Return the longest common absolute sub-path for pair of given absolute paths.
+
+    Note that this function implements similar functionality as os.path.commonpath(), however this function is not
+    available in python2.7, thus can't be used here.
+    """
+
+    if path1 == '' or path2 == '':
+        return ''
+
+    path1 = path1.strip('/').split('/')
```
fernflower, line 17: Sorry for being overcautious, could you please use `os.path.sep` value as a seprator instead of specifying '/' directly?

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
```diff
@@ -0,0 +1,102 @@
...
+    path2 = path2.strip('/').split('/')
+
+    common_path = []
+    for path1_part, path2_part in zip(path1, path2):
+        if path1_part != path2_part:
+            break
+        common_path.append(path1_part)
+    return '/' + '/'.join(common_path)
```
fernflower, line 25: Please use os.path.join for path join operations

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+
+    if path1 == '' or path2 == '':
+        return ''
+
+    path1 = path1.strip('/').split('/')
+    path2 = path2.strip('/').split('/')
+
+    common_path = []
+    for path1_part, path2_part in zip(path1, path2):
```
fernflower, line 21: Just curious, is there any reason why you didn't use the `os.path.commonpath` method?

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
```diff
@@ -0,0 +1,102 @@
...
+
+def check_fstab_mount_order():
+    storage_info = next(api.consume(StorageInfo), None)
+
+    if storage_info:
+
+        mount_points = []
+        for fstab_entry in storage_info.fstab:
+            mount_point = fstab_entry.fs_file
+            if mount_point[-1] == '/':
```
fernflower, line 60: imho `rstrip(os.path.sep)` is more readable and correct

IIUC L57-62 can be changed to

`mount_points = [fstab_entry.fs_file.rstrip(os.path.sep) for fstab_entry in storage_info.fstab]`

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
```diff
@@ -0,0 +1,102 @@
...
+                mount_point = mount_point[:-1]
+            mount_points.append(mount_point)
+
+        overshadowing = _get_overshadowing_mount_points(mount_points)
+
+        if overshadowing:
+
+            overshadowing_in_order = []
+            for mount_point in mount_points:
```
fernflower, line 69: Nit: can be changed to list comprehension, your decision

fernflower (dismissed):
Thanks for the unit test! Some open questions / comments inline.

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
pirat89: @fernflower does not exist under Python2

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
```diff
@@ -0,0 +1,102 @@
...
+                overshadowing.add(overshadowing_mount_point)
+                overshadowing.add(mount_point)
+
+    return overshadowing
+
+
+def check_fstab_mount_order():
+    storage_info = next(api.consume(StorageInfo), None)
+
+    if storage_info:
```
pirat89, line 55: drop unacessary indentations, also having the empty line after the `if` statement is not human friendly :)

I suggest to wach the following video:
* https://www.youtube.com/watch?v=CFRhGnuXG-4&ab_channel=CodeAesthetic

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
pirat89: I have a different question:
* Why so many `if ... == '/'` checks?

I would expect to deal with the trailing slash just on one place and never need to think about it again in the code. So we see just one `rstrip('/')` in the code and never need to hear about check/rstrip anywhere else during the execution.

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
PeterMocary: Fixed.

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
PeterMocary: Yes, you are right, it could be done using single ``rstrip('/')``. The ``rstrip('/')`` in ``check_fstab_mount_order()`` filters out the trailing ``/`` and we can forget about it, however...

Looking at the function ``_get_overshadowing_mount_points()`` as a standalone sub-program, it should strip the trailing ``/`` either way. Given that it relies on unified format in the comparison (``_get_common_path`` always returns path without the last ``/``) and also when creating its output (set of unique mount points that are involved in overshadowing). Therefore I decided to enforce it inside the function as well, to ensure correct results even without removal of the trailing ``/`` from its input list before calling the function.

This doesn't change the fact that the previous implementation was odd though. I adjusted it and I think that it doesn't feel that off now.

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
PeterMocary: Discussed in other thread.

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
PeterMocary: Discussed in other thread.

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
PeterMocary: I'll leave it at ``/`` then.

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
PeterMocary: Looking at it now, list comprehension reads better on multiple places, so I converted a few more loops.

repos/system_upgrade/common/actors/checkfstabmountorder/libraries/checkfstabmountorder.py
PeterMocary: Fixed.

PeterMocary:
@pirat89 @fernflower I addressed the issues.

fernflower (dismissed):
All my comments resolved, thanks!

pirat89 (changes requested):
@PeterMocary  Following report is confusing:
```
Risk Factor: high (inhibitor)
Title: Detected incorrect order of entries or duplicate entries in /etc/fstab, preventing a successful in-place upgrade.
Summary: Leapp detected incorrect /etc/fstab format that causes overshadowing of mount points.
Detected order of overshadowing mount points:
/boot,
Remediation: [hint] To prevent the overshadowing:
Reorder the detected overshadowing entries.
Possible order of all mount points without overshadowing:
, none, /boot
Key: fb07b688331fe65635aec13c2a1cb4e86ca6f4dc
```

Another confusing report is:
```
Risk Factor: high (inhibitor)
Title: Detected incorrect order of entries or duplicate entries in /etc/fstab, preventing a successful in-place upgrade.
Summary: Leapp detected incorrect /etc/fstab format that causes overshadowing of mount points.
Detected order of overshadowing mount points:
/boot/efi/foo, /boot/efi, /boot
Remediation: [hint] To prevent the overshadowing:
Reorder the detected overshadowing entries.
Possible order of all mount points without overshadowing:
, none, /boot, /boot/efi, /boot/efi/foo
Key: fb07b688331fe65635aec13c2a1cb4e86ca6f4dc
```

To reproduce the first one, just switch order of lines with `/` and `/boot` mountpoints. Here is my sys configuration used for the second report:
```
root@localhost ~]# cat /etc/fstab

#
# /etc/fstab
# Created by anaconda on Tue Jan 25 12:00:08 2022
#
# Accessible filesystems, by reference, are maintained under '/dev/disk/'.
# See man pages fstab(5), findfs(8), mount(8) and/or blkid(8) for more info.
#
# After editing this file, run 'systemctl daemon-reload' to update systemd
# units generated from this file.
#
/dev/something /boot/efi/foo ext4 defaults 0 0
/dev/something /boot/efi ext4 defaults 0 0
UUID=72847225-0df9-419e-8aed-95f385628f02 /boot                   xfs     defaults        0 0
/dev/mapper/rhel-swap   none                    swap    defaults        0 0


[root@localhost ~]# lsblk
NAME          MAJ:MIN RM SIZE RO TYPE MOUNTPOINT
vda           252:0    0  20G  0 disk 
├─vda1        252:1    0   1G  0 part /boot
└─vda2        252:2    0  19G  0 part 
  ├─rhel-root 253:0    0  17G  0 lvm  /
  └─rhel-swap 253:1    0   2G  0 lvm  [SWAP]
```

Suggested changes:
*  The `rstrip` should be skipped in case of `/` (rootfs).
* The `none` output is represented by swap that is not mounted anowhere. in case the mountpoint is not an absolute path, it should be ignored.
* The intendation is not visible and it's kind of hard to read it.  Especially in case of the possible order it's problem. I suggest to use "semi-standardized" way that uses `FMT_LIST_SEPARATOR` like in `checksystemdbrokensymlinks` actor. So the report gets better readability. E.g. something like:
```
Risk Factor: high (inhibitor)
Title: Detected incorrect order of entries or duplicate entries in /etc/fstab, preventing a successful in-place upgrade.
Summary: Leapp detected incorrect /etc/fstab format that causes overshadowing of mount points.
Detected order of overshadowing mount points:
    - /boot/efi/foo
    - /boot/efi
    - /boot
Remediation: [hint] To prevent the overshadowing:
Reorder the detected overshadowing entries.
Possible order of all mount points without overshadowing:
    - /
    - /boot
    - /boot/efi
    - /boot/efi/foo
Key: fb07b688331fe65635aec13c2a1cb4e86ca6f4dc
```
or you could do simple something like this in case you want to have everything just on one line (but keep in mind that some systems have even tens partitions):
```
Detected order of overshadowing mount points: /boot/efi/foo, /boot/efi, /boot
```
in this case even the one line seems good. 

* It's visible the hint is not friendly even with the suggestion above. Currently two hints could be possibly produced and it not so helpful to see something like
```
Remediation: [hint] To prevent the overshadowing:
Remove detected duplicates by using unique mount points.
Reorder the detected overshadowing entries.
Possible order of all mount points without overshadowing:
    - /
    - /boot
    - /boot/efi
    - /boot/efi/foo
```
the output seems mixed in this case. What about doing this?
```
Remediation: [hint] Remove detected duplicates by using unique mount points. Reorder the detected overshadowing entries. Possible order of all mount points without overshadowing:
    - /
    - /boot
    - /boot/efi
    - /boot/efi/foo
```
in case no duplicates are present, it would look just like this:
```
Remediation: [hint] Reorder the detected overshadowing entries. Possible order of all mount points without overshadowing:
    - /
    - /boot
    - /boot/efi
    - /boot/efi/foo
```
both seems to me more readable to be honest. Also you could do something like:
```
Remediation: [hint] To prevent the overshadowing:
    * Remove detected duplicates by using unique mount points.
    * Reorder the detected overshadowing entries. Possible order of all mount points without overshadowing:
        - /
        - /boot
        - /boot/efi
        - /boot/efi/foo
```

Also think about the presentation of reports in web UI (cockpit, foreman/satellite), where the readability could be also different. So e.g. the last suggestion, with multiple indentations, could be possibly displayed ignoring leading whitespaces / indentation (not 100% sure about that). So the last one is possibly not so safe for the use.

PeterMocary:
Added new fixup commit addressing found issues @pirat89. Thanks for such in depth description!

pirat89 (approved):
Seems good to me know! Great job!

```
Risk Factor: high (inhibitor)
Title: Detected incorrect order of entries or duplicate entries in /etc/fstab, preventing a successful in-place upgrade.
Summary: Leapp detected incorrect /etc/fstab format that causes overshadowing of mount points.
Detected order of overshadowing mount points: /boot/efi/foo, /boot/efi, /boot, /
Remediation: [hint] To prevent the overshadowing: Reorder the detected overshadowing entries. Possible order of all mount points without overshadowing:
    - /
    - /boot
    - /boot/efi
    - /boot/efi/foo
Key: fb07b688331fe65635aec13c2a1cb4e86ca6f4dc
```

PR #1011, opened 2023-01-24
check ceph-osd package presence

bessonc:
A customer reported they got a high risk factor while trying to retrieve the Ceph volume list, but this is only a ceph client. The issue is very low as it does not inhibit the upgrade.

The actor in question should not be executed:
- there is no ceph-osd service on this machine (package missing and no container)
- the ceph-volume binary itself comes from this package

Version: leapp-upgrade-el7toel8-0.17.0-1.el7_9.

The below suggestion is to check the presence of the ceph-osd package on the host.
Resolves [rhbz#2163796](https://bugzilla.redhat.com/show_bug.cgi?id=2163796)

### DIFF
--- a/repos/system_upgrade/common/actors/cephvolumescan/libraries/cephvolumescan.py
+++ b/repos/system_upgrade/common/actors/cephvolumescan/libraries/cephvolumescan.py
@@ -4,7 +4,7 @@ import re
 
 from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.common.rpms import has_package
-from leapp.libraries.stdlib import CalledProcessError, run
+from leapp.libraries.stdlib import api, CalledProcessError, run
 from leapp.models import InstalledRPM
 
 CEPH_CONF = "/etc/ceph/ceph.conf"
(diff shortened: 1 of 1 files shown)


pirat89:
Hi Chris! Thanks for the contribution. The bug is supposed to be fixed already by #986 and it will be part of the next release. If you can, please try to reproduce the current bug with the upstream build - if the current upstream fixes the problem already, I suggest to close this PR. (or let me ask @holser for the confirmation)

@holser  Hi, can you confirm the issue is already fixed by your PR? Double-check to be sure I haven't overlooked something.

bessonc:
It's not exactly the same problem. If there is no container named ceph-osd, this actor tries to run the ceph-volume command on the host itself. The PR 986 does not fix that.

pirat89:
@bessonc I did a rebase + updated the commit to satisfy tests

holser:
I have tested in in environment where we have ceph-mon containers on one machine and ceph-osd on others. Firstly, it doesn't break the logic for containerized ceph. However, this PR is trying to resolve non containerized ceph which I didn't get into account when I was working on #986 as I didn't have such environments. So, my +1 for this PR. Thanks

pirat89 (approved):
merging! thanks @bessonc  @holser

PR #1012, opened 2023-01-25
Refactoring of grub actors using scanner

fernflower:
Some refactoring on top of the inspirational https://github.com/oamg/leapp-repository/pull/1006 patch.


MichalHe (commented):
Small nitpicks :)

repos/system_upgrade/common/actors/scangrubconfig/libraries/scanner.py
```diff
@@ -0,0 +1,74 @@
...
+def _get_config_contents(config_path):
+    if os.path.isfile(config_path):
+        with open(config_path, 'r') as config:
+            return config.read()
+    return ''
+
+
+def is_grub_config_missing_final_newline(conf_file):
+    config_contents = _get_config_contents(conf_file)
+    return config_contents != '' and config_contents[-1] != '\n'
```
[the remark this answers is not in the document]
fernflower: Yeah, thanks, so far I have just moved the original code from PR1006 to the new place :)

repos/system_upgrade/common/actors/detectcorruptedgrubenv/actor.py
```diff
@@ -0,0 +1,37 @@
+from leapp import reporting
+from leapp.actors import Actor
+from leapp.models import GrubConfigError
+from leapp.reporting import create_report, Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class DetectCorruptedGrubenv(Actor):
```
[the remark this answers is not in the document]
fernflower: Hmm which exact logic you suggest to put there?

repos/system_upgrade/common/actors/scangrubconfig/libraries/scanner.py
```diff
@@ -0,0 +1,74 @@
...
+    # Check for GRUB_CMDLINE_LINUX syntax errors
+    # XXX FIXME(ivasilev) Can we make this check a common one? For now let's limit it to rhel7->rhel8 only
+    if version.get_major_version() == '7':
+        if not architecture.matches_architecture(architecture.ARCH_S390X):
+            # For now, skip just s390x, that's only one that is failing now
+            # because ZIPL is used there
+            config = '/etc/default/grub'
+            if detect_config_error(config):
+                errors.append(GrubConfigError(error_detected=True,
+                                              error_type=GrubConfigError.ERROR_GRUB_CMDLINE_LINUX_SYNTAX))
```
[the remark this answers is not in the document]
fernflower: (just some related thoughts on the matter)
I am very intrigued why do we keep the error_detected field in this Model. TBH I can't imagine the use for GrubConfigError with error_detected=False.
Do you have any idea?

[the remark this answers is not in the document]
pirat89: hmm, e.g.
```
if not error.error_detected:
    raise StopActorExecution
```
also it's easy check that we haven't detected any error inside the grub configuration when investigating. yes, we argue that the same could be achieved when the msg is not created at all, but it already exists, so are going to keep it in this way. It's not a new field so I would keep it.

repos/system_upgrade/common/actors/scangrubconfig/libraries/scanner.py
```diff
@@ -0,0 +1,74 @@
...
+            return config.read()
+    return ''
+
+
+def is_grub_config_missing_final_newline(conf_file):
+    config_contents = _get_config_contents(conf_file)
+    return config_contents != '' and config_contents[-1] != '\n'
+
+
+def detect_config_error(conf_file):
```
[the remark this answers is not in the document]
fernflower: If it's ok I'd rather keep the original version.

repos/system_upgrade/common/actors/scangrubconfig/libraries/scanner.py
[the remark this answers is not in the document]
fernflower: I don't see any occurencies like you mentioned in the code base unfortunately :(
The reason I am asking is because I'd like to add a docstring explaining why this field is here and what value should be set.

However I did find this in [addupgradebootentry actor](https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/common/actors/addupgradebootentry/actor.py#L23):
```
if grub_config_error.error_detected:
                fix_grub_config_error('/etc/default/grub', grub_config_error.error_type)
```
fix_grub_config implementation [is here](https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/common/actors/addupgradebootentry/libraries/addupgradebootentry.py#L99)

How about I put the following docstring
```
If the error can be automatically corrected (ex. in add upgrade boot entry actor), please set the error_detected field to True
```
?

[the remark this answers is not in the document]
fernflower: Hmmm now thinking about it - if that is indeed the case, how about renaming `error_detected` to `error_resolvable`?

repos/system_upgrade/common/actors/detectcorruptedgrubenv/actor.py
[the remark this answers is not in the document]
fernflower: Figured it out, thanks for the idea!

repos/system_upgrade/common/actors/scangrubconfig/libraries/scanner.py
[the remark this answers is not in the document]
pirat89: we cannot just rename it, the same problem why we cannot just remove it. we have the deprecation process so in case we would like to do something like that, we need to introduce the new field and support the old one for the protected period. but it's unfortunate use indeed when seeing the code.

repos/system_upgrade/common/actors/detectgrubconfigerror/actor.py
```diff
@@ -0,0 +1,73 @@
...
+                ),
+                reporting.Remediation(
+                    hint='Delete {} file(s) and regenerate grubenv using the grub2-mkconfig tool'.format(
+                        ','.join(error.files))),
+                reporting.Severity(reporting.Severity.HIGH),
+                reporting.Groups([reporting.Groups.BOOT]),
+                reporting.Groups([reporting.Groups.INHIBITOR]),
+            ]
+            report_fields.extend([reporting.RelatedResource('file', config) for config in error.files])
+            create_report(report_fields)
```
pirat89, line 73: @fernflower  can you actually put it into the library and split it into functions? also it will be better to nest the `if` statements into the for-loop  to improve the readability. Also I would add `break` after each report (or return when you put it into functions) so it's clear that it's not possible to see multiple reports for the same thing

MichalHe: @pirat89 IMHO splitting this into functions brings a little in terms of value. It might be me getting old and grumpy, but unless the code is called from multiple places, or the logic of what is being done in a block is very complex, I prefer linear stream of code, because it is just easier to debug when you don't have to deal with super-deep call stacks. Also, splitting this into several functions does not increase testability, as the tests can be essentially the same if it is one big function of several smaller. (_If an error of type X was detected, is inhibitor generated?_ - only the tested function is changed)

MichalHe: But I agree that on that having one forloop with if-elif-else branches inside would be more readable in a sense that it is obvious that any error can be reported precisely once.

MichalHe (commented):
Except for the comment around the structure of reporting logic, LGTM. @fernflower I would love to hear your input on that as structuring code is something I am still attempting to formulate my firm opinion on.

repos/system_upgrade/common/models/grubconfigerror.py
```diff
@@ -3,7 +3,15 @@
 
 
 class GrubConfigError(Model):
+    ERROR_CORRUPTED_GRUBENV = 'corrupted grubenv'
+    ERROR_MISSING_NEWLINE = 'missing newline'
+    ERROR_GRUB_CMDLINE_LINUX_SYNTAX = 'GRUB_CMDLINE_LINUX syntax'
```
pirat89, line 8: that's nice actually. However I am not sure whether this is allowed tbh. It's the first time (IIRC) we set any attribute for a model classes and we strongly required to have present only fields + topic in the class. But I really like it. so I want to be sure, we all agree to have attributes allowed in models, so we do not enforce people to use a different design in such cases (@oamg/developers ping).

repos/system_upgrade/common/actors/detectgrubconfigerror/actor.py
pirat89: @fernflower  @MichalHe I do not have strong opinion here - definitely not blocker at all. This is more about preferences. If you guys find it better in this way, I am ok with it. I thought it would be better/more readable to decompose it into smaller pieces, and adding breaks could be also seatbelts, so we are sure it will not happen in future that some reports appear multiple times (e.g. if an additional grub error msg is produced in future that is not expected now). But I do not have problem to keep it is if others find it better in this way. Thanks both for reacting :-)

repos/system_upgrade/common/actors/detectgrubconfigerror/actor.py
```diff
@@ -0,0 +1,87 @@
...
+    name = 'detect_grub_config_error'
+    consumes = (GrubConfigError,)
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        # syntax error in GRUB_CMDLINE_LINUX, recoverable
+        for error in [err for err in self.consume(GrubConfigError)
+                      if err.error_type == GrubConfigError.ERROR_GRUB_CMDLINE_LINUX_SYNTAX]:
+            _create_grub_error_report(
```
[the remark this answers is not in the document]
fernflower: Okay, tests added. Please re-review :)

repos/system_upgrade/common/actors/scangrubconfig/libraries/scanner.py
```diff
@@ -0,0 +1,73 @@
...
+            # For now, skip just s390x, that's only one that is failing now
+            # because ZIPL is used there
+            if detect_config_error(config):
+                errors.append(GrubConfigError(error_detected=True, files=[config],
+                                              error_type=GrubConfigError.ERROR_GRUB_CMDLINE_LINUX_SYNTAX))
+
+    # Check for missing newline errors
+    if is_grub_config_missing_final_newline(config):
+        errors.append(GrubConfigError(error_detected=True, error_type=GrubConfigError.ERROR_MISSING_NEWLINE),
+                      files=[config])
```
matejmatuska, line 71: Little detail: Seems like the order of checks doesn't matter, so they could be swapped and then you could return early if on s390.

matejmatuska (commented):
Also, the tests look ok codewise.

repos/system_upgrade/common/actors/detectgrubconfigerror/tests/test_detectgrubconfigerror.py
```diff
@@ -0,0 +1,58 @@
...
+    messages = current_actor_context.consume(Report)
+    assert len(messages) == 1
+    message = messages[0]
+    assert 'Detected a missing newline at the end of grub configuration file' in message.report['title']
+    assert message.report['severity'] == 'low'
+    assert message.report['detail']['related_resources'][0]['title'] == '/etc/default/someothergrub.cfg'
+
+
+def test_corrupted_config(current_actor_context):
+    # Make sure that just 1 low priority report message is created with config files present
```
MichalHe, line 47:
```suggestion
    # Make sure that just 1 high priority report message is created with config files present
```

MichalHe: If I understand correctly, it is not expected that actor will receive multiple messages of the same error type. There might be a problem in the future with these tests that they expect a certain ordering on how the actor processes the messages. If the actor implementation slightly changes for whatever reason and instead of the first message of a given type, the last one will be used to generate the report, the person implementing the changes will have to change it.

That being said, these are just tests and the above comment is meant to maybe give you a perspective you might consider in the future. Maybe I would feed the actor only with one message, but otherwise the tests are fine. We have more important things to work on then to make absolutely flawless unit tests.

repos/system_upgrade/common/actors/detectgrubconfigerror/actor.py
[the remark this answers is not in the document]
MichalHe: Thank you, looks fine to me.

repos/system_upgrade/common/actors/scangrubconfig/libraries/scanner.py
MichalHe, line 71: This is why the integration tests are failing.
```suggestion
        errors.append(GrubConfigError(error_detected=True, error_type=GrubConfigError.ERROR_MISSING_NEWLINE,
                      files=[config]))
```

MichalHe: The unit tests did not hit this because they are testing whether the logic actor uses correctly recognizes a missing newline (`is_grub_config_missing_final_newline`) rather than whether an the message is produced. Could we add a single unit tests that would cause all error reasons to be present and check that all errors are generated?

MichalHe (changes requested):
There is a misplaced parenthesis causing integration tests to fail. Both integration tests regarding grub newlines are failing because the first integration test modifies the grub config, but leapp crashes without correcting the config. Therefore the next subsequent test also crashes on the exactly same codepath.

I have finally figured out why the sanity test is crashing on rhel8. It turns out when you have a sanity test that queries leapp.db for logged errors, you should make sure that there is no leapp.db with errors from previous runs, or to filter the log messages by leapp execution ID (context). The failed integration tests contain a leapp.db that you can use to verify that the asserted error is from different execution. (DNF) Timestamps in the log messages of the "failed" leapp run will guide you to clearly see the problem.

Thus when the misplaced parenthesis is fixed, and the requested unit test it added, I will merge it, hoping that I will not spent too much time debugging integration tests again.

repos/system_upgrade/common/actors/scangrubconfig/libraries/scanner.py
fernflower: I owe you a beer! Thanks for catching this, I will definitely cover it with a unit tests.

repos/system_upgrade/common/actors/detectgrubconfigerror/tests/test_detectgrubconfigerror.py
fernflower: The only expectation so far is that the actor produces not more than 1 message of a certain type, yeah. 
I added this "1 message only check" here specifically to check that our current convention "only 1 message" is being met. If for some reason it changes in the future then these tests will fail, right, but that is the idea - to make the convention visible.
If it's ok with you I'd rather keep things as they are now, I really like visibility.

fernflower:
@MichalHe Damn, I have just realized that I haven't resolved all your comments - I need to add this one test that generates all messages. Please don't spend any time on reviews yet.

MichalHe:
@fernflower lol, the integration tests are failing because they check for the report message `Detected a missing newline at the end of grub configuration file.` with the dot at the end. As silly as it might sound (especially since it is not consistent with the rest of the produced messages), please add the dot back so we will not be seeing red on other PRs as we are in a dev phase when we will likely need to merge a lot of them. Please, also squash the last two commits.

fernflower:
@MichalHe I kind of wanted to change the integration test to not check for the dot (as is the right way imho)..

MichalHe (approved):

MichalHe:
The core logic is covered with passing unit tests, all integrations tests are green. It's merging time!

PR #1023, opened 2023-01-30
Revert "Propagate TEST_PATHS to test_container targets"

fernflower:
This reverts commit 1859d1811d6331eda8c9684fac47b12ca2e796ae.

Revert done because of affected container-tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -383,8 +383,8 @@ _test_container_ipu:
 		echo "Only supported TEST_CONT_IPUs are el7toel8, el8toel9"; exit 1; \
 		;; \
 	esac && \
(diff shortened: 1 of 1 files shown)


fernflower:
https://github.com/oamg/leapp-repository/pull/981 This should resolve issues mentioned by @matejmatuska while we figure out the better solution.

matejmatuska:
> #981 This should resolve issues mentioned by @matejmatuska while we figure out the better solution.

To make the `ACTOR` variable work we can just propagate it the same way as `REPOSITORIES` i.e. `-e ACTOR`.

pirat89 (approved):
let's revert it. The possible fix can be delivered later

matejmatuska (commented):
LGTM (late but for some reason the page didn't refresh for me on merge)

PR #1026, opened 2023-02-01
pes_events_scanner: Fix regression in pes_event_parsing

matejmatuska:
When `/etc/leapp/files/pes-events.json` contains invalid data (e.g. empty, invalid syntax, garbage content) leapp tracebacks.

This commit fixes that and also introduces a test case to make sure we actually produce report and raise the correct `StopActorExecution` exception.

Jira ref.: OAMG-7637

### DIFF
--- a/repos/system_upgrade/common/actors/peseventsscanner/tests/test_event_parsing.py
+++ b/repos/system_upgrade/common/actors/peseventsscanner/tests/test_event_parsing.py
@@ -3,14 +3,20 @@ from collections import namedtuple
 
 import pytest
 
+from leapp import reporting
+from leapp.exceptions import StopActorExecution
 from leapp.libraries.actor.pes_event_parsing import (
     Action,
     Event,
+    get_pes_events,
     Package,
     parse_entry,
     parse_packageset,
     parse_pes_events
 )
+from leapp.libraries.common import fetch
+from leapp.libraries.common.testutils import create_report_mocked, CurrentActorMocked
+from leapp.libraries.stdlib import api
 
 CUR_DIR = os.path.dirname(os.path.abspath(__file__))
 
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/common/actors/peseventsscanner/tests/test_event_parsing.py
```diff
@@ -146,3 +152,18 @@ def test_parse_pes_events_with_modulestreams(current_actor_context):
...
+    def read_or_fetch_mocked(*args, **kwargs):
+        raise ValueError()
+
+    monkeypatch.setattr(fetch, 'read_or_fetch', read_or_fetch_mocked)
+    created_reports = create_report_mocked()
+    monkeypatch.setattr(reporting, "create_report", created_reports)
+    monkeypatch.setattr(api, "current_actor", CurrentActorMocked())
+
+    with pytest.raises(StopActorExecution):
+        get_pes_events('doesnt', 'matter')
```
fernflower, line 167: Nit: (sorry, can't help my OCR) s/doesnt/doesn't

fernflower (approved):
lgtm

MichalHe:
3 of the integration tests checking the behavior of leapp with invalid pes data are failing. These failures are expected, as the test is checking for the pes_events_scanner actor failing unexpectedly, which is an incorrect behavior of leapp, and therefore, the test should not expect leapp failing unexpectedly on any input.

Moreover, a sanity test checking for no errors being generated is failing due to error log statement being generated  which I cannot reproduce locally. Moreover, it is weird that an error is reported in one actor (`targetuserspacecryptopolicies`), however, the upgrade is not stopped by another actor in the same phase that consumes the same message and raises an exception if no message was consumed (`MissingGpgKeysInhibitor`).

pirat89 (approved):

PR #1027, opened 2023-02-01
Handle possible failures of xfs_info command

fernflower:
In case the fstab entry is not mounted, the xfs_info command will fail. By no reason this should cause leapp to exit with a traceback.

OAMG-8147
BZ#2155661 - leapp xfs_info_scanner fails when fstab xfs entry not mounted

### DIFF
--- a/repos/system_upgrade/common/actors/xfsinfoscanner/libraries/xfsinfoscanner.py
+++ b/repos/system_upgrade/common/actors/xfsinfoscanner/libraries/xfsinfoscanner.py
@@ -1,4 +1,6 @@
-from leapp.libraries.stdlib import api, run
+import os
+
+from leapp.libraries.stdlib import api, CalledProcessError, run
 from leapp.models import StorageInfo, XFSPresence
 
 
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/common/actors/xfsinfoscanner/libraries/xfsinfoscanner.py
```diff
@@ -21,7 +21,13 @@ def scan_xfs_mount(data):
 
 
 def is_xfs_without_ftype(mp):
-    for l in run(['/usr/sbin/xfs_info', '{}'.format(mp)], split=True)['stdout']:
```
pirat89, line 24: @fernflower could add the check that the `mp` is really mountpoint? and in case it's not a mountpoint, just log info msg that particular entry is not mounted si skipping the check

pirat89 (approved):
lgtm and Works as expected. 
```
# cat /etc/fstab
UUID=72847225-0df9-419e-8aed-95f385628f02 /boot                   xfs     defaults        0 0
/dev/mapper/rhel-swap   none                    swap    defaults        0 0
/dev/mapper/something   /okno                   xfs defaults,nofail,noauto 0 0

# leapp preupgrade --debug --no-rhsm
[snip]
2023-02-03 11:22:54.988 INFO     PID: 24720 leapp: Answerfile will be created at /var/log/leapp/answerfile

Debug output written to /var/log/leapp/leapp-preupgrade.log

                           REPORT                           

A report has been generated at /var/log/leapp/leapp-report.json
A report has been generated at /var/log/leapp/leapp-report.txt

                       END OF REPORT                        

```

PR #1033, opened 2023-02-13
Set RHSM release for non-ga & non-beta channels

pirat89:
The actor ogirinally set the target release only when the GA channel has been set for the target OS - as originally it has been possible to set only: ga, beta, htb.

However, nowadays we exect additional possibilities, like: eus, aus, e4s, ... and for those we should set the channel as well - especially for these!!

Updating the skip condition, to ignore just 'beta' channel ('htb' does not exist anymore).

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=2168953


fernflower (commented):
```
However, nowadays we exect additional possibilities, like: eus, aus,
e4s, ... and for those we should set the channel as well - especially
for these!!
```
I'd rephrase ``However, now additional possibilities exist, like eus, aus, e4s and ...

Otherwise LGTM

pirat89:
> 

Thanks. Done. New wording: `However, additional possibilities exist now, like ...`

fernflower (approved):
lgtm

PR #1034, opened 2023-02-13
Makefile: Add rule for running just lint in container

matejmatuska:
Currently there are rules just for running both linter and tests or just tests in container. This commit adds rule for running just lint.

Jira ref.: OAMG-6724


pirat89 (approved):

pirat89:
lgtm and works as expected.

PR #1035, opened 2023-02-13
Move CIFS check actor to common

fernflower:
The check should be the safe for 7to8 and 8to9 upgrade. Also add logic to skip the check in case
LEAPP_DEVEL_INITRAM_NETWORK env var is set.

OAMG-8490


matejmatuska (approved):
Tested on both 7->8 and 8->9 with `cifs` entry in `/etc/fstab` and the upgrade was properly inhibited with message:
```
    1. Inhibitor: Use of CIFS detected. Upgrade can't proceed
```
With `LEAPP_DEVEL_INITRAM_NETWORK` set, the upgrade was not inhibited, as expected.

LGTM

MichalHe (approved):
LGTM

MichalHe:
The only integration tests that are failing are those testing for supplying invalid `pes-events.json` file and expecting leapp to crash due to unhandled exception. However, leapp not gracefully handling an invalid file is a bug that has been fixed with [PR1026](https://github.com/oamg/leapp-repository/pull/1026). There is already a patch correcting the integration tests. #

PR #1041, opened 2023-02-15
fix migration of sendmail configs after the upgrade

PeterMocary:
The ``migrate_sendmail`` actor would cause a traceback if a file marked for migration wasn't found, which would be caused for example by removing the ``sendmail`` package during the upgrade. The actor now checks for existence of files that need to be migrated and also informs when no files have been migrated.

Tested by adding ``sendmail`` package to ``/etc/leapp/transaction/to_remove``.

Jira ref.: OAMG-8311
BZ: [2162703](https://bugzilla.redhat.com/show_bug.cgi?id=2162703)

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/migratesendmail/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/migratesendmail/actor.py
@@ -1,6 +1,9 @@
+import os
+
 from leapp import reporting
 from leapp.actors import Actor
 from leapp.libraries.actor import migratesendmail
+from leapp.libraries.stdlib import api
 from leapp.models import SendmailMigrationDecision
 from leapp.reporting import create_report, Report
 from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/el7toel8/actors/migratesendmail/actor.py
```diff
@@ -21,14 +23,20 @@ def process(self):
         if not decision or not decision.migrate_files:
             return
 
+        migrated = []
         for f in decision.migrate_files:
-            migratesendmail.migrate_file(f)
+            if os.path.exists(f):
+                migratesendmail.migrate_file(f)
+                migrated.append(f)
```
pirat89, line 30: @PeterMocary  thinking about that, add also error log when a file cannot be migrated, instead of skipping it only.

repos/system_upgrade/el7toel8/actors/migratesendmail/actor.py
```diff
@@ -21,14 +23,20 @@ def process(self):
...
+
         list_separator_fmt = '\n    - '
+        summary = 'Uncompressed IPv6 addresses in {}'.format(list_separator_fmt.join(migrated))
+        if not migrated:
+            summary = 'No files migrated, which might be caused by removal of sendmail package during the upgrade.'
```
pirat89, line 35: in this case, it's opposite what the title is saying. So the current report is in case of issue saying
* title: sendmail configuration files migrated
* summary: no files migrated, probably due to ..

which is not ideal. I suggest to create even the list of "skipped" files and have the logic:
* if not skipped, print the migrated report (severity low/info)
* else report sendmail configuration files have not been migrated (severity medium)

pirat89 (changes requested):

repos/system_upgrade/el7toel8/actors/migratesendmail/actor.py
PeterMocary: Added.

repos/system_upgrade/el7toel8/actors/migratesendmail/actor.py
PeterMocary: I agree, the report was confusing. It should be better now.

pirat89 (approved):

PR #1046, opened 2023-02-17
Include leapp data files in the RPM & repository

pirat89:
In the past it was needed to obtain the data by
* automatic download of data files from RH Insights
  (cloud.redhat.com)
  - which required access to the server & have the system registered
* manual download from the article:
    https://access.redhat.com/articles/3664871
  which required to login to the portal, download the archive
  and install its content manually on the system

Additional problem was with the syncing of data, as because of the separation of data from the code, people had ensure they are using the right combination of data and SW manually - in case the data has not been downloaded automatically from RH Insights.

Having the data in the RPM makes our lives easier so let's install them to `/etc/leapp/files/` via the package directly.
Set /etc/leapp/files/* as configuration files to ensure that modified files will be backed up with *.rpmsave suffix as
expected for configuration files. This could be important in case users manually updated e.g. repomap.json file to reflect their internal settings of the satellite server.

The complete table how the configuration files are handled during
rpm installation/upgrade/... is here:
*    https://www.cl.cam.ac.uk/~jw35/docs/rpm_config.html

Keeping original functionality still available, so if people remove the files from the system, data can be still downloaded from the server automatically. It seems it does not make sense, but we could still use the possibility to download more up-to-date data from RH Insights.

Configure codespell to ignore .etc/leapp/files

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -20,4 +20,8 @@ jobs:
           ignore_words_list: ro,fo,couldn,repositor
           skip: "./repos/system_upgrade/common/actors/storagescanner/tests/files/mounts,\
             ./repos/system_upgrade/el7toel8/actors/networkmanagerreadconfig/tests/files/nm_cfg_file_error,\
-            ./repos/system_upgrade/common/actors/scancpu/tests/files/lscpu_s390x"
+            ./repos/system_upgrade/common/actors/scancpu/tests/files/lscpu_s390x,\
+            ./etc/leapp/files/device_driver_deprecation_data.json,\
+            ./etc/leapp/files/pes-events.json,\
+            ./etc/leapp/files/repomap.json"
+
--- a/etc/leapp/files/device_driver_deprecation_data.json
+++ b/etc/leapp/files/device_driver_deprecation_data.json
@@ -1,0 +1,4774 @@
+{
+  "data": [
+    {
+      "available_in_rhel": [
+        7
+      ],
+      "deprecation_announced": "",
+      "device_id": "0x15B3:0x1002",
+      "device_name": "Mellanox Technologies: MT25400 Family [ConnectX-2 Virtual Function]",
+      "device_type": "pci",
+      "driver_name": "mlx4_core",
+      "maintained_in_rhel": [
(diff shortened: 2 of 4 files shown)


pirat89:
@vinzenz  fyi :) I will later consider actually use of LFS. keeping in draft for the weekend

pirat89:
@oamg/developers  The failed codespell test is false positive. The issue has been reported in upstream: https://github.com/codespell-project/codespell/issues/2786

pirat89:
@oamg/developers  @oamg/qe  Tested manually on el7 & el8.
* install new packages, no config files present on the system before
```console
[root@localhost ~]# find /etc/leapp/files
find: ‘/etc/leapp/files’: No such file or directory
[root@localhost ~]# dnf install "leapp-upgrade-*pr1046*" -y
....
Complete!
[root@localhost ~]# find /etc/leapp/files
/etc/leapp/files
/etc/leapp/files/device_driver_deprecation_data.json
/etc/leapp/files/pes-events.json
/etc/leapp/files/repomap.json
```

* install new packages, config files already present on the system
```console
[root@localhost files]# mkdir -p /etc/leapp/files
[root@localhost files]# get_data_files prod
...
[root@localhost files]# dnf install "leapp-upgrade-*pr1046*" -y
...
Complete!
[root@localhost files]# find /etc/leapp/files
/etc/leapp/files
/etc/leapp/files/device_driver_deprecation_data.json.rpmorig
/etc/leapp/files/device_driver_deprecation_data.json
/etc/leapp/files/pes-events.json.rpmorig
/etc/leapp/files/pes-events.json
/etc/leapp/files/repomap.json.rpmorig
/etc/leapp/files/repomap.json
```

I do not have now tests to show how the rpmsave files are created as it would require additional rpm with new data (tested before when I updated the data files, but as we want to proceed with these files anyway, I am not creating an extra build for it to show the behaviour in the comment. just believe the rpm functionality)

pirat89:
the changelog updated in advance

pirat89:
rebased against up-to-date master to see whether it resolves some of failed tests

Rezney (approved):
The failing tests are known.

PR #1053, opened 2023-03-09
Enable 8>9 upgrades with FIPS enabled

MichalHe:
# Add support for in-place 8>9 upgrades of FIPS-enabled systems
FIPS refers to a set of security standards governing many aspects of how information should be handled by computers and by people. FIPS in context of RHEL typically means FIPS 140, a single document defining rules for the use of encryption and cryptographic services. In essence, it defines requirements for cryptographic modules (e.g. what algorithms can be used, thus preventing the use of insecure ones) manipulating sensitive information.

From the point of view of upgrades there are 5 components that are certified under FIPS: kernel, OpenSSL, GnuTLS, NSS, and libgcrypt. As for the kernel, `fips=1` needs to be present on the cmdline in order to enable FIPS in the kernel. Kernel offers a `/proc/sys/crypto/fips_enabled` virtual file containing the information about whether the kernel has booted with FIPS enabled.

According to FIPS, the userspace components need to verify that they were not tampered with, and thus they have to have some sort of checksum either in a special section of a binary, or in a separate file. The components read `/proc/sys/crypto/fips_enabled` to know when to switch into the FIPS enabled mode. OpenSSL does things a bit differently, by not including support for FIPS directly but via a special `fips.so` module. Furthermore, for OpenSSL to check whether FIPS is enabled a configuration file must be present at `/etc/pki/tls/openssl.cnf`, because the code checking for FIPS is a part of configuration parsing.

As every userspace component has different implementation of FIPS-mode, and a different way for the user to check whether the component believes that the system it is running on has FIPS enabled, there is no straightforward way for us to make really sure that all of the components run in FIPS enabled inside the target userspace container.

## Working around a bug in dracut
As @neverpanic discovered during his initial investigation of 8>9 upgrades with FIPS enabled, dracut currently contains a [bug](https://bugzilla.redhat.com/show_bug.cgi?id=2176560) due to which it does not include some necessary files in the initramfs. Until the issue is resolved in dracut, a temporary workaround consisting of manually specifying files to be included in the initramfs is implemented.

Finally, this PR, including the majority of the above description (modulo possible mistakes introduced by @MichalHe), is based on the great work done by @neverpanic. Thanks!

ref: OAMG-7824


repos/system_upgrade/common/actors/initramfs/checkfipsenabled/actor.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def process(self):
+        fips_info = next(self.consume(FIPSInfo), None)
+        if not fips_info:
+            # Unexpected, FIPSInfo is produced unconditionally
+            raise StopActorExecutionError('Cannot check for the correct FIPS state in the upgrade initramfs',
+                                          details={'Problem': 'Did not receive any FIPSInfo message'})
+
+        if fips_info.is_enabled:
+            # @Todo: checking for fips=1 in cmdline is likely insufficient to determine whether the upgrade environment
```
neverpanic, line 26: I don't think there's a simple way to run the check we would need to run, since AFAIK all of these tests run *inside* of the `systemd-nspawn` container, right? That did already contain everything that was needed, and FIPS mode does already work correctly in this environment.

In this case, I think verifying that FIPS mode was enabled is good enough – it'll at least fail when fips=1 was not correctly passed into the kernel command line, although we could probably get the same assurance by adding a unit test for the `addupgradebootentry` actor that verifies that `fips=1` will be preserved.

In any case, if we keep this test, it should read `/proc/sys/crypo/fips_enabled` and check that it contains `1`, since that will tell us whether the kernel (a) supports FIPS mode, and (b) has successfully been switched into FIPS mode.

repos/system_upgrade/common/actors/checkfips/actor.py
```diff
@@ -33,3 +37,11 @@ def process(self):
                     reporting.Groups([reporting.Groups.SECURITY]),
                     reporting.Groups([reporting.Groups.INHIBITOR])
                 ])
+        else:
+            if fips_info.is_enabled:
+                fips_required_initramfs_files = [
+                    '/usr/lib64/ossl-modules/fips.so',
+                    '/etc/pki/tls/openssl.cnf',
```
neverpanic, line 44: The default `/etc/pkit/tls/openssl.cnf` includes `/etc/crypto-policies/back-ends/opensslcnf.config` and will fail if that file isn't present, so we'll also need that.

On the other hand, this is clearly a bug in dracut (see https://bugzilla.redhat.com/show_bug.cgi?id=2176560), so maybe we should just wait for dracut to fix this, rather than working around it here?

repos/system_upgrade/common/actors/checkfips/actor.py
```diff
@@ -33,3 +37,11 @@ def process(self):
                     reporting.Groups([reporting.Groups.SECURITY]),
                     reporting.Groups([reporting.Groups.INHIBITOR])
                 ])
+        else:
+            if fips_info.is_enabled:
+                fips_required_initramfs_files = [
+                    '/usr/lib64/ossl-modules/fips.so',
```
neverpanic, line 43: Is this always going to be `/usr/lib64` for all supported architectures, or do we support systems where this might be `/usr/lib`? Should we use `rpm --eval '%{_libdir}'` to determine this path?

repos/system_upgrade/common/actors/checkfips/actor.py
MichalHe: I agree that we should not be copying the files ourselves, and the code copying the files will be removed once the problem is fixed in dracut.  I've opted for implementing the workaround as it did not require too much (temporary) code and it would allow for further prototyping if need be.

repos/system_upgrade/common/actors/initramfs/checkfipsenabled/actor.py
MichalHe: You are correct that leapp runs inside the container at the time these tests are executed. Historically, we have encountered issues when using grubby to manipulate boot entries due to which I am being rather cautious as booting without `fips=1` would result in the target system without `fips=1` as well (not sure how bad the consequences are).

neverpanic: If the system boots without `fips=1`, that means FIPS mode was essentially disabled silently, so we should really make sure that doesn't happen.

However, `fips=1` on the kernel command line and `1` in `/proc/sys/crypto/fips_enabled` have the same meaning, and the proc file is simpler to parse, so I'd suggest to keep this test but check the contents of `/proc/sys/crypto/fips_enabled`.

repos/system_upgrade/common/actors/checkfips/actor.py
MichalHe: We support `x86_64`, `aarch64`,`ppc64le`, `s390x` (all are 64bit). Just to be sure I've checked that `fips.so` provided by the `openssl-libs` is present in the `/usr/lib64` on all of the supported archs.

repos/system_upgrade/common/actors/checkfips/actor.py
MichalHe: I've added the `opensslcnf.config` into the list of files we include in the initramfs.

repos/system_upgrade/common/actors/initramfs/checkfipsenabled/actor.py
MichalHe: Thank you for your suggestion to use `/proc/sys/crypto/fips_enabled` instead of parsing cmdline, I've changed the code accordingly.

repos/system_upgrade/common/actors/initramfs/checkfipsenabled/actor.py
```diff
@@ -0,0 +1,33 @@
...
+        fips_info = next(self.consume(FIPSInfo), None)
+        if not fips_info:
+            # Unexpected, FIPSInfo is produced unconditionally
+            raise StopActorExecutionError('Cannot check for the correct FIPS state in the upgrade initramfs',
+                                          details={'Problem': 'Did not receive any FIPSInfo message'})
+
+        if fips_info.is_enabled:
+            with open('/proc/sys/crypto/fips_enabled') as fips_status_handle:
+                fips_status = fips_status_handle.read().strip()
+                if fips_status == '1':
```
neverpanic, line 30: We want `/proc/sys/crypto/fips_enabled` to contain 1, that tells us that FIPS mode is enabled.

```suggestion
                if fips_status != '1':
```

neverpanic (commented):
One minor nitpick, the rest looks good to me.

repos/system_upgrade/common/actors/checkfips/actor.py
```diff
@@ -11,25 +12,36 @@ class CheckFips(Actor):
...
                     reporting.Severity(reporting.Severity.HIGH),
                     reporting.Groups([reporting.Groups.SECURITY]),
                     reporting.Groups([reporting.Groups.INHIBITOR])
                 ])
+        else:
+            if fips_info.is_enabled:
+                fips_required_initramfs_files = [
+                    '/etc/crypto-policies/back-ends/opensslcnf.config',
+                    '/etc/pki/tls/openssl.cnf',
+                    '/lib64/.gnutls.hmac',
```
neverpanic, line 44: For GnuTLS, see also https://bugzilla.redhat.com/show_bug.cgi?id=2148269. It seems they moved the file to a different location that dracut might already expect.

repos/system_upgrade/common/actors/initramfs/checkfipsenabled/actor.py
MichalHe: Thank you. Fixed with latest changes.

repos/system_upgrade/common/actors/checkfips/actor.py
MichalHe: I was able to get dracut to automatically pick up the hmac files (`lib64/.libgnutls.so.*.hmac`) by adding the `fips` dracut module when building the upgrade initramfs. Unfortunately, dracut's codebase seems not to mention any of the openssl stuff except for `fips.so`. `fips.so` will be installed with a recently introduced patch  [c5dca3d68915cef077fda2bc5292e12f82cf6dd6](https://github.com/dracutdevs/dracut/commit/c5dca3d68915cef077fda2bc5292e12f82cf6dd6), however, as a part of `crypt` module and there should not be dependency relation between `crypt` and `fips` modules. Therefore, there is nothing "cleaner" leapp can than to specify the openssl files ourselves.

neverpanic: Sounds good.

neverpanic (approved):

repos/system_upgrade/common/actors/checkfips/actor.py
```diff
@@ -11,25 +12,36 @@ class CheckFips(Actor):
...
+                                          details={'Problem': 'Did not receive a message with information about FIPS '
+                                                              'usage'})
+
+        if version.get_target_major_version() == '8':
+            if fips_info.is_enabled:
                 title = 'Cannot upgrade a system with FIPS mode enabled'
-                summary = 'Leapp has detected that FIPS is enabled on this system. ' \
-                          'In-place upgrade of systems in FIPS mode is currently unsupported.'
+                summary = ('Leapp has detected that FIPS is enabled on this system. '
+                           'In-place upgrade of systems in FIPS mode is currently unsupported.')
```
Rezney, line 31: Let's adjust this message a little bit with this occasion and state that only RHEL 7 to RHEL 8 upgrades with FIPS are not supported. For now the generic message may discourage users from future upgrades from RHEL 8 to RHEL 9 too.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
```diff
@@ -188,14 +207,24 @@ def copy_boot_files(context):
...
+        initram_path=os.path.join('/boot', initram),
+        kernel_hmac_path=kernel_hmac_path
     )
 
     context.copy_from(os.path.join('/artifacts', kernel), content.kernel_path)
     context.copy_from(os.path.join('/artifacts', initram), content.initram_path)
 
+    if fips_info.is_enabled:
+        kernel_hmac_path = context.full_path(os.path.join('/artifacts', kernel_hmac))
+        create_upgrade_hmac_from_target_hmac(kernel_hmac_path, content.kernel_hmac_path, kernel)
```
Rezney, line 226: @pirat89 Based on your experience from the 6 > 7 upgrades, could you please take a look if there might be any issue around here?

repos/system_upgrade/common/actors/checkfips/actor.py
```diff
@@ -11,25 +12,36 @@ class CheckFips(Actor):
...
                     reporting.Title(title),
                     reporting.Summary(summary),
                     reporting.Severity(reporting.Severity.HIGH),
                     reporting.Groups([reporting.Groups.SECURITY]),
                     reporting.Groups([reporting.Groups.INHIBITOR])
                 ])
+        else:
+            if fips_info.is_enabled:
+                fips_required_initramfs_files = [
+                    '/etc/crypto-policies/back-ends/opensslcnf.config',
```
Rezney, line 42: Could you please add a "FIXME" here and reference the Dracut bug in the comment?

Rezney (changes requested):
I am yet to finish my testing but so far looks good to me. I suggest squashing some comments (not all of them as a good part of them makes sense). Thanks o/

repos/system_upgrade/common/actors/checkfips/actor.py
Rezney: Yes! Perfect reminder. @MichalHe could you please include it?

repos/system_upgrade/common/actors/checkfips/actor.py
MichalHe: done

MichalHe:
@Rezney I have reordered and cleaned up the git history. I was keeping the history in a messy state in case further touches would be required during reviews. Furthermore, commit hash is included in the leapp built I have used during testing, so you can identify the PR state directly from the logs attaches to the ticket.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
pirat89: This seems ok to me :+1:

pirat89:
Discussed one potential issue in person (under investigation now). Except that, everything else is good to go - just the code review without manual testing.

MichalHe:
@pirat89 I have rerun the upgrade after my latest changes removing some of the unnecessary logic, the upgrade went OK.

Rezney (approved):

pirat89 (approved):

pirat89 (approved):

PR #1056, opened 2023-03-10
[WIP] checkfips: Allow FIPS upgrades in 8.8

neverpanic:
The `addupgradebootentry` actor does already preserve `fips=1` on the kernel command line, so the upgrade initramfs is correctly booted in FIPS mode.

The upgrade initramfs does not correctly enable FIPS mode for all libraries due to a bug in dracut that also affects all normal boot environments generated by dracut. See rhbz#2176560 for a report of this problem. Once this is fixed, we can enable leapp upgrades in FIPS mode.

The systemd-nspawn container in the upgrade initramfs does correctly pass `/proc` and has the required files (OpenSSL FIPS provider, OpenSSL configuration file, crypto-policies, GnuTLS HMAC integrity checksums) installed, so that part already behaves as expected in a FIPS uprgade scenario.

See: OAMG-7824

### DIFF
--- a/repos/system_upgrade/common/actors/checkfips/actor.py
+++ b/repos/system_upgrade/common/actors/checkfips/actor.py
@@ -1,6 +1,7 @@
 from leapp import reporting
 from leapp.actors import Actor
 from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.common.config import version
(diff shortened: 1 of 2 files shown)


neverpanic:
See #1053.

PR #1059, opened 2023-03-14
Inhibit unsupported x86-64 microarchitecture RHEL9

dkubek:
As per [x86-64-ABI][1] In addition to the AMD64 baseline architecture, several micro-architecture levels implemented by later CPU modules have been defined, starting at level ``x86-64-v2``.

RHEL9 has a higher CPU requirement than older versions, it now requires a CPU compatible with ``x86-64-v2`` instruction set or higher. Until now, there was no check for this and the upgrade crashed unexpectedly. This commit handles this issue and provides the user with a report explaining the problem.

The CPU Features are gathered using the ``lscpu`` command by way of using the ``ScanCPU`` actor. The ``ScanCPU`` actor had to be also modified to provide the required flags. The mapping of CPU Features to flags provided by ``lscpu`` has been determined by using the ``/arch/x86/include/asm/cpufeatures.h`` file from the linux kernel.

Jira ref.: OAMG-8619

[1]: https://gitlab.com/x86-psABIs/x86-64-ABI.git

### DIFF
--- a/repos/system_upgrade/common/actors/scancpu/libraries/scancpu.py
+++ b/repos/system_upgrade/common/actors/scancpu/libraries/scancpu.py
@@ -17,6 +17,11 @@ def _get_lscpu_output():
     return ''
 
 
+def _get_cpu_flags(lscpu):
+    flags = lscpu.get('Flags', '')
+    return flags.split()
+
+
 def _get_cpu_entries_for(arch_prefix):
     result = []
     for message in api.consume(DeviceDriverDeprecationData):
@@ -137,4 +142,10 @@ def process():
     api.produce(*_find_deprecation_data_entries(lscpu))
     # Backwards compatibility
     machine_type = lscpu.get('Machine type')
-    api.produce(CPUInfo(machine_type=int(machine_type) if machine_type else None))
+    flags = _get_cpu_flags(lscpu)
+    api.produce(
+            CPUInfo(
+                machine_type=int(machine_type) if machine_type else None,
+                flags=flags
+                )
+            )
--- a/repos/system_upgrade/common/actors/scancpu/tests/test_scancpu.py
+++ b/repos/system_upgrade/common/actors/scancpu/tests/test_scancpu.py
@@ -1,14 +1,59 @@
 import os
 
+import pytest
+
 from leapp.libraries.actor import scancpu
 from leapp.libraries.common import testutils
+from leapp.libraries.common.config.architecture import (
+    ARCH_ARM64,
+    ARCH_PPC64LE,
+    ARCH_S390X,
+    ARCH_SUPPORTED,
+    ARCH_X86_64
+)
 from leapp.libraries.stdlib import api
 from leapp.models import CPUInfo
 
 CUR_DIR = os.path.dirname(os.path.abspath(__file__))
 
+LSCPU = {
+    ARCH_ARM64: {
+        "machine_type": None,
+        "flags": ['fp', 'asimd', 'evtstrm', 'aes', 'pmull', 'sha1', 'sha2', 'crc32', 'cpuid'],
+    },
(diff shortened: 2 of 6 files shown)


repos/system_upgrade/common/actors/scancpu/libraries/scancpu.py
```diff
@@ -17,6 +17,11 @@ def _get_lscpu_output():
     return ''
 
 
+def _get_cpu_flags(lscpu):
+    flags = lscpu.get('Flags')
```
fernflower, line 21: Nit: 
```
flags = lscpu.get('Flags', '')
return flags.split()
```

repos/system_upgrade/common/actors/scancpu/libraries/scancpu.py
```diff
@@ -137,4 +142,10 @@ def process():
     api.produce(*_find_deprecation_data_entries(lscpu))
     # Backwards compatibility
     machine_type = lscpu.get('Machine type')
```
fernflower, line 144: Just wondering if we can end up with non-int here? If yes conversion at L148 could fail

fernflower (commented):
Nothing serious, just a couple of remarks / open questions. lgtm otherwise

repos/system_upgrade/common/actors/scancpu/libraries/scancpu.py
dkubek: I'm unsure. This is the way I've found it. Looking at the source of ``lscpu``, it seems to be getting the "Machine type" field from the "Type" field in the``/proc/sysinfo`` file [ref. 1][1] . In the kernel source for the s390x architecture, this field is printed and stored as a string [ref. 2][2], [ref. 3][3]. This information is generated from reading the result of the STSI instruction of s390x. However, I was not able to find a reference that would specify what values can be produced there. So I guess it is possible hmm

[1]: https://github.com/util-linux/util-linux/blob/50d216859f7ec6a5b200a8fb90c9a0870d12e018/sys-utils/lscpu-cputype.c#L741
[2]: https://github.com/torvalds/linux/blob/94710cac0ef4ee177a63b5227664b38c95bbf703/arch/s390/kernel/sysinfo.c#L91
[3]: https://github.com/torvalds/linux/blob/94710cac0ef4ee177a63b5227664b38c95bbf703/arch/s390/include/asm/sysinfo.h#L27

pirat89:
@dkubek  the comit msg and description says that RHEL 9 requires x86-64-v2 and higher, but it requires x86-64-v3 and higher

repos/system_upgrade/common/actors/scancpu/libraries/scancpu.py
dkubek: Sure!

repos/system_upgrade/common/actors/scancpu/libraries/scancpu.py
dkubek: No worries! I've been checking the lscpu source for the stuff around flags as well so it was pretty quick. I've made one last quick search for the description of the STSI instruction and found it described in [z/Architecture Principles of Operation, page 114][1] with

> Type: Word 12 contains the four-character (0-9) EBCDIC type number of the configuration. (This is called the machine-type number in the definition of STORE CPU ID.)

So the only problem is when the type starts with leading zeros, which of course will not be shown when converted to int, but I think this is not a problem and I'll keep it as is :)

[1]: https://geraldine.fjfi.cvut.cz/~oberhuber/data/mainframe/redbooks/pmf/z-architecture-principles-of-operations.pdf

fernflower (commented):
lgtm, will schedule tests before approval.

fernflower (approved):

PR #1060, opened 2023-03-15
Transition systemd service states during upgrade

matejmatuska:
Sometimes after the upgrade some services end up disabled even if they have been enabled on the source system.

There are already two separate actors that fix this for `device_cio_free.service` and `rsyncd.service`.

A new actor `transition-systemd-services-states` handles this generally for all services. A "desired" state is determined depending on state and vendor preset of both source and target system and a SystemdServicesTasks message is produced with each service that isn't already in the "desired" state.

Jira ref.: OAMG-1745


matejmatuska:
Tested with on 7->8 with:
- rsyncd.service enabled, with actor from #1043 removed
- And with dhcpd.service enabled (see #826) 

#### After the reboot:
Produced message:
```
root@leapp-20230314094848 ~]# leapp-inspector messages --type SystemdServicesTasks
                          PRODUCED MESSAGES
Stamp: 2023-03-14T21:42:24.318449Z
Actor: transition_systemd_services_states
Phase: Applications
Type: SystemdServicesTasks
Message_data:
{
    "to_disable": [],
    "to_enable": [
        "abrt-ccpp.service",
        "dhcpd.service",
        "rsyncd.service"
    ]
}
```

Services are correctly kept enabled
```
[root@leapp-20230314094848 ~]# systemctl is-enabled rsyncd
enabled
[root@leapp-20230314094848 ~]# systemctl is-enabled dhcpd
enabled
```

matejmatuska:
Ran the same tests after the refactor and addition of reports and the results are the same.

pirat89:
@matejmatuska finally on my todolist for tomorrow!! :)

pirat89:
> @matejmatuska finally on my todolist for tomorrow!! :)

bad @pirat89 !! :vampire_man:  jumping on the review now

pirat89:
@matejmatuska  rebase please

repos/system_upgrade/common/actors/systemd/transitionsystemdservicesstates/libraries/transitionsystemdservicesstates.py
```diff
@@ -0,0 +1,211 @@
...
+        " were kept enabled after the upgrade. "
+    )
+    if tasks:
+        summary += (
+            "The following services were originally disabled on the upgraded system"
+            " and Leapp attempted to enable them:{}{}"
+        ).format(FMT_LIST_SEPARATOR, FMT_LIST_SEPARATOR.join(sorted(tasks.to_enable)))
+        # TODO(mmatuska): When post-upgrade reports are implemented in
+        # `setsystemdservicesstates actor, add a note here to check the reports
+        # if the enabling failed
```
pirat89, line 138: @matejmatuska regarding the current implementation, I expect this will be always true. Add please unit tests that will check the correct summary is picked based on the input.

pirat89: Seems this is the only problem I found by the review of the code. It seems to me good otherwise. I will get to the testing during the week.

repos/system_upgrade/common/actors/systemd/transitionsystemdservicesstates/libraries/transitionsystemdservicesstates.py
```diff
@@ -0,0 +1,211 @@
...
+    return message
+
+
+def process():
+    services_source = _expect_message(SystemdServicesInfoSource).service_files
+    services_target = _expect_message(SystemdServicesInfoTarget).service_files
+    presets_source = _expect_message(SystemdServicesPresetInfoSource).presets
+    presets_target = _expect_message(SystemdServicesPresetInfoTarget).presets
+
+    services_source = dict((p.name, p.state) for p in services_source)
```
fernflower, line 196: Nit: won't dict comprehension be a bit easer to read?
```
services_source = {p.name: p.state for p in services_source}
```

repos/system_upgrade/common/actors/systemd/transitionsystemdservicesstates/libraries/transitionsystemdservicesstates.py
matejmatuska: > @matejmatuska regarding the current implementation, I expect this will be always true.

I suspect that there might be a system which doesn't meet this condition, so I made this conditional. Also I think it's more future-proof.

I will add the test.

pirat89: I see, it make sense. Thanks for the explanation.

pirat89:
Seems to me working as expected, tested 7 -> 8 and 8 -> 9, tested:
* preserve enabled services (rsyncd)
* preserved explicitely disabled services (preset: enabled, manually disabled)  (kdump)


@matejmatuska squash fix commits and we can merge it.

matejmatuska:
> Seems to me working as expected, tested 7 -> 8 and 8 -> 9, tested:
> * preserve enabled services (rsyncd)
> * preserved explicitely disabled services (preset: enabled, manually disabled)  (kdump)
> 
> 
> @matejmatuska squash please fix commits, then we can merge it.

Would you prefer to keep the "remove obsoleted rsyncd.." separate from the main commit?

matejmatuska:
> @matejmatuska you can keep it separated if you want

@pirat89 Done, sorry for the delay.

matejmatuska:
I am not sure what's wrong with the codespell and linter, the warnings seem to be unrelated to this PR.

pirat89 (approved):
Finally it's here \o/ merging. Thanks!!

Note: tests passed. these forever queued tests are optional and not expected now to be executed. the fix is in progress.

PR #1061, opened 2023-03-20
Use the updated Leapp output API - reports summary

matejmatuska:
Use the updated Leapp output API introduced in oamg/leapp#818.
Leapp output will now include a summary of reports and also a list of titles of HIGH and MEDIUM severity reports.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -7,6 +7,7 @@ disable=
  no-member,
  no-name-in-module,
  raising-bad-type,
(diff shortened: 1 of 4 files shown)


pirat89:
@matejmatuska  then it's also referring to bug in the leapp PR

pirat89 (approved):

pirat89:
Internal tests have been updated. merging

PR #1062, opened 2023-03-20
Fix dead link in checkipaserver actor

andywaltlova:
Investigating inhibitors and found this dead link.

### DIFF
--- a/repos/system_upgrade/common/actors/checkipaserver/libraries/checkipaserver.py
+++ b/repos/system_upgrade/common/actors/checkipaserver/libraries/checkipaserver.py
@@ -1,10 +1,8 @@
 from leapp import reporting
 from leapp.libraries.common.config.version import get_source_major_version
 
-MIGRATION_GUIDE_7 = (
-    "https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux"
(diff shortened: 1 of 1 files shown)


andywaltlova:
review please

andywaltlova:
review please

PeterMocary (dismissed):
LGTM. The new link points to the relevant article. Thank you for the fix!

fernflower (dismissed):

repos/system_upgrade/common/actors/checkipaserver/libraries/checkipaserver.py
```diff
@@ -3,7 +3,7 @@
 
 MIGRATION_GUIDE_7 = (
```
fernflower, line 4: Maybe let's use shortener for this link as well, so that it is easily testable?

pirat89: I agree with Ina. I would use the shortened URL so in case of another change, we can reflect the broken symlink without the need to modify the code.

andywaltlova: Alright, updated

fernflower (approved):
perfect, thanks!

PR #1064, opened 2023-03-24
checkhybridimage: Fix the produce of the report

pirat89:
The actor can produce the report, however the report is never stored / accepted by the leapp framework as the Report has not been set in the `produces` tuple. Fixing the actor.


matejmatuska:
LGTM

matejmatuska (approved):
LGTM

PR #1070, opened 2023-04-17
Add tag in packit.yaml to enable cost metrics collection

fernflower:
Now all tft tests run by packit should be marked accordingly with a sst-upgrades tag.

OAMG-8892

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -95,15 +95,15 @@ jobs:
       distros: [RHEL-7.9-ZStream]
   identifier: tests-7.9to8.6
   tmt_plan: "^(?!.*upgrade_plugin)(?!.*tier[2-3].*)(?!.*rhsm)(?!.*c2r)(?!.*sap)(?!.*8to9)(?!.*max_sst)"
-  tf_post_install_script: "#!/bin/sh\nsudo sed -i s/.*ssh-rsa/ssh-rsa/ /root/.ssh/authorized_keys"
   tf_extra_params:
     environments:
       - tmt:
           context:
             distro: "rhel-7.9"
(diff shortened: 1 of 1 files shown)


.packit.yaml
```diff
@@ -157,12 +167,17 @@ jobs:
...
-  tf_post_install_script: "#!/bin/sh\nsudo sed -i s/.*ssh-rsa/ssh-rsa/ /root/.ssh/authorized_keys; echo 42; yum-config-manager --enable rhel-7-server-rhui-optional-rpms"
   tf_extra_params:
     environments:
       - tmt:
           context:
             distro: "rhel-7.9"
+        # tag resources as sst_upgrades to enable cost metrics collection
+        settings:
+          provisioning:
+            post_install_script: "#!/bin/sh\nsudo sed -i s/.*ssh-rsa/ssh-rsa/ /root/.ssh/authorized_keys; echo 42; yum-config-manager --enable rhel-7-server-rhui-optional-rpms"
```
pirat89, line 178: @fernflower  why the `echo 42`? :) is it forgotten debug cmd or it has actually some additional value?

pirat89 (approved):
seems good to me. untested.

.packit.yaml
fernflower: We can safely remove that now I guess :) I will fix it in the next patch when I will be squashing 2 last commits anyway, thanks

fernflower:
@pirat89 squashed the commits, please re-approve. 2 tests are not doing well, I am following up with the packit team on this as they started failing after move to packit and it looks like it is not because we messed up with the regex or something.
But anyway it is not due to this change.

PR #1071, opened 2023-04-21
BZ#2188239 - better detect already migrated PostgreSQL data

evgeni:
Users can (and in some cases have to) manually move the PostgreSQL data to the new location. In the case when they would completely delete the old location, the facts code would previously error out:

        self.run()
      File "/usr/lib/python2.7/site-packages/leapp/actors/init.py", line 289, in run
        self.process(*args)
      File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/satellite_upgrade_facts/actor.py", line 96, in process
        scl_psql_stat = os.stat(POSTGRESQL_SCL_DATA_PATH)
    OSError: [Errno 2] No such file or directory: '/var/opt/rh/rh-postgresql12/lib/pgsql/data/'

However, if the user already has migrated the data, we don't have to do it in the actor and thus can skip all that detection code.

Also adjust the report to say the data looks migrated (but still needs a reindex).

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/satellite_upgrade_check/libraries/satellite_upgrade_check.py
+++ b/repos/system_upgrade/el7toel8/actors/satellite_upgrade_check/libraries/satellite_upgrade_check.py
@@ -23,12 +23,18 @@ def satellite_upgrade_check(facts):
         title = "Satellite PostgreSQL data migration"
         flags = []
         severity = reporting.Severity.MEDIUM
+        intro_msg = "PostgreSQL on RHEL 8 expects its data in /var/lib/pgsql/data."
         reindex_msg = textwrap.dedent("""
-        After the data has been moved to the new location, all databases will require a REINDEX.
-        This will happen automatically during the first boot of the system.
+        PostgreSQL on RHEL 8 requires a rebuild of all database indexes, when using data created on RHEL 7.
+        This REINDEX will happen automatically during the first boot of the system.
         """).strip()
 
-        if facts.postgresql.same_partition:
+        if not facts.postgresql.scl_pgsql_data:
+            migration_msg = """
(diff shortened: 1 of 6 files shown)


repos/system_upgrade/el7toel8/actors/satellite_upgrade_check/libraries/satellite_upgrade_check.py
```diff
@@ -28,7 +28,9 @@ def satellite_upgrade_check(facts):
         This will happen automatically during the first boot of the system.
         """).strip()
 
-        if facts.postgresql.same_partition:
+        if not facts.postgresql.scl_pgsql_data:
+            migration_msg = "Your PostgreSQL data seems to be already migrated."
```
fernflower, line 32: Nit: Will it make sense to add "No additional action will be taken" to provide more clarity to the user?

evgeni: Yes and no ;-)
An action *will* be taken: the reindex will happen. But you are right, no *movement* of the data will take place, which is a good thing to point out! Will add!

evgeni:
> ` Failed loading plugin "copr": module 'dnf' has no attribute 'cli'`

this looks not like an related error?

fernflower:
@evgeni It is unrelated, yeah, sorry about that. It has emerged recently and we are actively investigating it.
BTW you don't need to leave a '/rerun' comment anymore, now packit will schedule tests for you. I guess I need to update the PR bot message :)

pirat89:
@lpramuk Hi Lukas, can you (or someone) test it before the merge and confirm it works as expected?

pirat89 (approved):
lgtm. waiting for @lpramuk for the feedback about the testing for now.

PR #1073, opened 2023-04-25
Update pr-welcome-msg with packit tests info

fernflower:
Let's mention the big leap for leapp officially together with a command to (re-)trigger tests.

### DIFF
--- a/.github/workflows/pr-welcome-msg.yml
+++ b/.github/workflows/pr-welcome-msg.yml
@@ -34,4 +34,4 @@ jobs:
 
             Please [open ticket](https://url.corp.redhat.com/oamg-ci-issue) in case you experience technical problem with the CI. (RH internal only)
 
-            **Note:** In case there are problems with tests not being triggered automatically on new PR/commit or pending for a long time, please consider rerunning the CI by commenting **leapp-ci build** (might require several comments). If the problem persists, contact leapp-infra.
+            **Note:** In case there are problems with tests not being triggered automatically on new PR/commit or pending for a long time, please contact leapp-infra.
--- a/.github/workflows/pr-welcome-msg.yml
+++ b/.github/workflows/pr-welcome-msg.yml
@@ -21,15 +21,15 @@ jobs:
             ## **Thank you for contributing to the Leapp project!**
             Please note that every PR needs to comply with the [Leapp Guidelines](https://leapp.readthedocs.io/en/latest/contributing.html#) and must pass all tests in order to be mergeable.
             If you want to request a review or rebuild a package in copr, you can use following commands as a comment:
-            - **review please** to notify leapp developers of review request
+            - **review please @oamg/developers** to notify leapp developers of the review request
             - **/packit copr-build** to submit a public copr build using packit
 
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-welcome-msg.yml
```diff
@@ -24,6 +24,8 @@ jobs:
             - **review please** to notify leapp developers of review request
```
pirat89, line 24:
```suggestion
            - **review please <at>oamg/developers ** to notify leapp developers of review request if nobody answered for couple of days
```

.github/workflows/pr-welcome-msg.yml
```diff
@@ -24,6 +24,8 @@ jobs:
...
             - **/packit copr-build** to submit a public copr build using packit
 
+            Packit will automatically schedule regression tests for this PR's build and leapp\*master\*. If you need a different version of leapp from PR#42, use `/packit test oamg/leapp#42`
```
pirat89, line 27:
```suggestion
            Packit will automatically schedule regression tests for this PR's build and latest upstream (devel) build of leapp. If you need a different version of leapp from PR#42, use `/packit test oamg/leapp#42`
```

@fernflower   I prefer to rather skip the name of the master/main/upstream branch as we are anyway enforced to change it soon to something else.

fernflower: Ok, I will use `latest upstream build` instead, thanks

.github/workflows/pr-welcome-msg.yml
fernflower: Done

.github/workflows/pr-welcome-msg.yml
fernflower: Done everywhere

fernflower:
Hmmm I am pretty sure there has been no reaction to leapp-ci build command, so I will add another commit to remove the note as well. I guess we have moved on for good :)

.github/workflows/pr-welcome-msg.yml
```diff
@@ -21,15 +21,17 @@ jobs:
             ## **Thank you for contributing to the Leapp project!**
             Please note that every PR needs to comply with the [Leapp Guidelines](https://leapp.readthedocs.io/en/latest/contributing.html#) and must pass all tests in order to be mergeable.
             If you want to request a review or rebuild a package in copr, you can use following commands as a comment:
-            - **review please** to notify leapp developers of review request
+            - **review please @oamg/developers** to notify leapp developers of the review request
```
pirat89, line 24: @fernflower  I tried to replace the `@` by `<at>` to be sure that people are not pinged with each new welcome msg. Would this lead to such a situation?

fernflower: From what I see in a test run https://github.com/fernflower/leapp-repository/pull/17#issuecomment-1521570808 that's not the case. I'd leave a `@` here and submit a patch later if we are not happy.

pirat89: @fernflower thanks. you are right, it seems ok. let's merge!

pirat89 (approved):

PR #1076, opened 2023-04-26
Add el8toel9 actor to handle directory -> symlink with ruby IRB.

jackorp:
A symlink to directory conversion halts DNF update with a transaction error.

As this was not handled on the RPM component level, the simplest fix now is to either uninstall the package first and then, or simply upgrade delete the offending directory.

The directory `/usr/share/ruby/irb` can be freely deleted as it does not hold any sort of configuration or similar. It is simply only containing code provided by the package.

Related BZ bug: https://bugzilla.redhat.com/show_bug.cgi?id=2030627


jackorp:

fernflower (commented):
lgtm, will schedule tests

fernflower (approved):

pirat89:
@fernflower please do not forget to update the changelog :)

PR #1077, opened 2023-05-05
Change the upgrade paths for SAP HANA

alexxa:
- Drop 7.9 to 8.2
 - Add 7.9 to 8.8
 - Add 8.8 to 9.2
 - Drop SAP HANA version starting from the target releases >=8.8 and >=9.2
 - Correct actor.py docstring to support ppc64le for 8to9 upgrade


alexxa:
Please don't merge without ack from @berndfinger

repos/system_upgrade/common/files/upgrade_paths.json
```diff
@@ -7,9 +7,10 @@
         "8": ["9.2"]
     },
     "saphana": {
-        "7.9": ["8.2", "8.6"],
-        "7": ["8.2", "8.6"],
+        "7.9": ["8.8", "8.6"],
```
vinzenz, line 10: please sort the numbers from low to high

repos/system_upgrade/common/actors/checksaphana/libraries/checksaphana.py
```diff
@@ -160,12 +184,10 @@ def _sp_rev_patchlevel_check(instance, patchlevels):
 
 def _fullfills_hana_min_version(instance):
     """ Performs a check whether the version of SAP HANA fulfills the minimal requirements for the target RHEL """
-    if version.get_target_major_version() == '8':
+    if version.matches_target_version('8.6'):
```
vinzenz, line 187: if this gets called for 7.9 to 8.7+ (I know 8.7 isn't a thing) it would put in the 9.0 string always, is that what you want here?

repos/system_upgrade/common/actors/checksaphana/libraries/checksaphana.py
```diff
@@ -72,12 +99,10 @@ def _create_detected_instances_list(details):
 
 
 def _min_ver_string():
-    if version.get_target_major_version() == '8':
+    if version.matches_target_version('8.6'):
```
vinzenz, line 102: This will return SAP_HANA_RHEL90_MINIMAL_VERSION_STRING when called for >=8.7 and <8.6  target versions. Don't think you want that like this. This needs to be checked properly that this is not a problem. However reading the code out of context is confusing.

repos/system_upgrade/common/actors/checksaphana/actor.py
```diff
@@ -12,17 +14,52 @@ class CheckSapHana(Actor):
...
+    dialogs = (
+        Dialog(
+            scope='check_sap_hana',
+            reason='Confirmation',
+            components=(
+                BooleanComponent(
+                    key='confirm',
+                    label='Do you want the upgrade to proceed for '
+                          'the currently installed SAP HANA 2.0 version?',
+                    description='Enter True, otherwise the upgrade process '
```
vinzenz, line 42: One should not write 'True' but 'Yes'

https://github.com/oamg/leapp/blob/9eb9c463fc849005378a34aac22a44a9cb1a662c/leapp/dialogs/components.py#L88

repos/system_upgrade/common/actors/checksaphana/actor.py
```diff
@@ -12,17 +14,52 @@ class CheckSapHana(Actor):
     If the upgrade flavour is 'default' no checks are being executed.
 
     The following checks are executed:
-    - If this system is _NOT_ running on x86_64, the upgrade is inhibited.
-    - If SAP HANA 1 has been detected on the system the upgrade is inhibited since it is not supported on RHEL8.
+    - If the major target release is 8, and this system is _NOT_ running on x86_64, the upgrade is inhibited.
+    - If the major target release is 9, and this system is _NOT_ running on x86_64 or ppc64le,
```
vinzenz, line 18: I don't see a change for the support of ppc64le - Did you miss that? Because it's not changed here in this PR but you document it that way. And also you say it's also for RHEL9, which most definitely needs a code change where the architecture is checked for SAP Hana (If it is at all)

vinzenz (dismissed):

repos/system_upgrade/common/files/upgrade_paths.json
alexxa: the order is correct, since the latest in the list is the default upgrade path.

repos/system_upgrade/common/actors/checksaphana/actor.py
alexxa: ppc64le support was added by #PR1042 , but this docstring was not updated.

repos/system_upgrade/common/actors/checksaphana/libraries/checksaphana.py
alexxa: no, because we can line 185 only if target_version <=8.6, see lines 200-203

repos/system_upgrade/common/actors/checksaphana/libraries/checksaphana.py
alexxa: no, because we can line 101 only if target_version <=8.6, see lines 200-203
plus <8.6 is not allowed for HANA case.

repos/system_upgrade/common/actors/checksaphana/actor.py
alexxa: I will correct this one, as soon we agree on other comments @vinzenz Thanks!

pirat89: @vinzenz @alexxa the `True` could works also (we use it in other place already), but yes seems to be more natural answer to the question, so let's stay with `Yes`

pirat89: I suggest to set also the default, so the cmd will be generated in the report that could be executed by user

repos/system_upgrade/common/files/upgrade_paths.json
pirat89: discussed, 8.6 (and 9.0) is expected to keep as default for saphana

repos/system_upgrade/common/actors/checksaphana/actor.py
```diff
@@ -12,17 +14,51 @@ class CheckSapHana(Actor):
...
     - If SAP HANA is running the upgrade is inhibited.
     """
 
     name = 'check_sap_hana'
     consumes = (SapHanaInfo,)
     produces = (Report,)
     tags = (IPUWorkflowTag, ChecksPhaseTag)
+    dialogs = (
+        Dialog(
+            scope='check_sap_hana',
```
pirat89, line 35: @alexxa just thinking, the current solution will also mean that if users answer True for IPU 7 -> 8, it will be already answered also for 8 -> 9! Even when the required version for the target system will be different. Thinking out loud, possibly you want to create scope `check_sap_hana_el8` and `check_sap_hana_el9` - it would mean you would need to create 2 different dialogs, but it would prevent a problem of automatically answered question. Later it could be possibly changed to a generator like:
```python
    dialogs =(for i in _gen_dialogs())

...
    @staticmethod
    def _gen_dialogs():
        for major_release in ('8', '9'):
            yield Dialog(scope='check_sap_hana_el{}'.format(major_release),
                ...
```
The problem is that we used dialogs for specific cases that have been connected always to a specific upgrade path, but now we introduces a requirement for a dialog(s) for multiple upgrade paths, which is different from previous scenearious.

@vinzenz @oamg/developers  any thoughts on this?

alexxa: Petr, I don't follow why it's valid for 7to8 case when get_saphana_answer and dialog inside are called only if a target release >8.6 and >9.0:
https://github.com/oamg/leapp-repository/pull/1077/files/84b14e4a20b26af380963927b7e0d1a323f470cf#diff-388596db3aa9f19b3e3ba62b12d22a70c8af01075f0384a4526cbbd4326691b0R198

pirat89: @alexxa the upgrade from rhel 7.9 -> 8.8 means the target is 8.8 which triggers the question. user will answer it and it's answered. if they later upgrade 8.8 -> 9.2, they will not be asked anymore because they answered the questions already for the upgrade 7.9 -> 8.8. and we can continue with the upgrade 9.6 -> 10.0 with the same problem.

alexxa: ah, do you mean inside for loop for multiple instances if answered once? =(((( but anyway, even if there are multiple HANA instances, they belong to one RHEL and one upgrade path, right?

repos/system_upgrade/common/actors/checksaphana/libraries/checksaphana.py
```diff
@@ -175,8 +195,14 @@ def version2_check(info):
     for instance in info.instances:
         if _manifest_get(instance.manifest, 'release', None) == '1.00':
             continue
-        if not _fullfills_hana_min_version(instance):
-            _add_hana_details(found, instance)
+        if version.matches_target_version('> 8.6', '> 9.0'):
```
pirat89, line 198:
```suggestion
        if version.matches_target_version('> 8.6', '< 9.0') or version.matches_target_version('> 9.0', '< 10.0'):
```

repos/system_upgrade/common/actors/checksaphana/actor.py
pirat89: Discussed via video call. We have decided to get rid of the user question as it's more feasible solution and keep it on the report only. As this is not real blocker for the upgrades and users are quired to read the report & documentation. Currently we have already some reports explaining people they are supposed to do some manual steps before the upgrade and they are not inhibitors neither.

@oamg/developers anyway, I think we have a nice topic for the discussion about dialogs. I will create a discussion topic later for that.

repos/system_upgrade/common/actors/checksaphana/libraries/checksaphana.py
alexxa: any target release  >9.0 should return warning in the report, so it's safe to skip '<10.0', right?

repos/system_upgrade/common/libraries/config/version.py
```diff
@@ -14,7 +14,7 @@
 _SUPPORTED_VERSIONS = {
     # Note: 'rhel-alt' is detected when on 'rhel' with kernel 4.x
     '7': {'rhel': ['7.9'], 'rhel-alt': [], 'rhel-saphana': ['7.9']},
-    '8': {'rhel': ['8.6', '8.8'], 'rhel-saphana': ['8.6']},
+    '8': {'rhel': ['8.6', '8.8'], 'rhel-saphana': ['8.6', '8.8']},
```
pirat89, line 17: :heart:  exactly. I am sorry I missed that. Hopefully it didn't make you a headache guys.

pirat89 (approved):
lgtm!

pirat89:
@alexxa  let me know when you finish the testing :)

repos/system_upgrade/common/actors/checksaphana/libraries/checksaphana.py
pirat89: yes, it shouldyes, it should.

pirat89:
@andywaltlova  marking this by `report` flag also as once we start to support the upgrade too, it's something what user should definitely know/check before they upgrade (there is no user who would like to upgrade to newest rhel version with sap hana installed if it is not certified by sap yet)

alexxa:
@pirat89 what is "once we start to support the upgrade too" and the whole comment about please? 
we still support the earlier release by default if HANA is detected to be on safer side, nevertheless,  there is no any certification status check in the upgrade code, and we have it as a document warning, and whatever customers do, it's at their responsibility then

andywaltlova:
I think @pirat89 wrote it in a hurry, it means once we will start to have upgrade task on red hat insights publicly available, we will want to be even more cautious on which systems we will show eligibility for upgrade - there is a rule which is basically saying 'your system is on upgrade path supported by leapp, do you want to try it? , we of course also render report in insights UI after preupgrade task runs and say everywhere that they should read it before proceeding, but.. well users.. :D they don't have upgrade task in UI yet so they 'experience on insights' now ends with remediations, but once the upgrade is added we just want to keep it in mind

PR #1080, opened 2023-05-11
Set merge_pr_in_ci to false in packit config

fernflower:
As automerging PR into the target branch has been problematic with testing farm as of lately and we don't really need this feature, let's disable it.


danmyway (approved):
Looks good @fernflower !

PR #1086, opened 2023-06-05
update URLs to docs: fix links and use shortened URLs when needed

pirat89:
It happened several times that a link we used in reports became invalid over time, so it could not provide expected help to customers. The fix of such links require update of the code, which does not have to be possible on existing releases. Instead of that, we rather decided to use shortnened URLs when possible and when it makes sense to reduce the risk. The benefit is that we are able to change the target destination of such https://red.ht/... URLs so we can fix such issues without the need to update the code (and rpms).

### New shortened URLs:
  * https://red.ht/upgrading-rhel7-to-rhel8-main-official-doc
  * https://red.ht/upgrading-rhel8-to-rhel9-main-official-doc
  * https://red.ht/ipu-customisation-repos-known-issues
  * https://red.ht/leapp-dialogs
  * https://red.ht/file-systems-and-storage-removed-btrfs-rhel-8
  * https://red.ht/rhel-8-migrate-postgresql-server
  * https://red.ht/rhel-8-python
  * https://red.ht/rhel-8-configuring-routing-protocols
  * https://red.ht/nm-settings-ifcfg-rh
  * https://red.ht/rhel-9-intel-microarchitectures
  * https://red.ht/rhel-9-blog-openssh-scp-deprecation

### Additional changes:
  * checktargetrepos: replace "TBA" by valid URL for 8 -> 9 docs
  * dropped unneeded pylint skips

### Another info
In some cases (at least in targetuserspacecreator - and possibly only in this one case) we have just links to our rhel8 documentation instead of dynamically point people to el8 or el9 docs (if they exists). I am considering to keep this as part of a future changes. In this one PR I do not want to do any changes in behavior.

### DIFF
--- a/repos/system_upgrade/common/actors/localreposinhibit/actor.py
+++ b/repos/system_upgrade/common/actors/localreposinhibit/actor.py
@@ -82,8 +82,7 @@ class LocalReposInhibit(Actor):
                             "in-place upgrade"
                         ),
                         url=(
-                            "https://access.redhat.com/articles/4977891/"
-                            "#repos-known-issues"
+                            "https://red.ht/ipu-customisation-repos-known-issues"
                         ),
                     ),
                 ]
--- a/repos/system_upgrade/el7toel8/actors/postgresqlcheck/libraries/postgresqlcheck.py
+++ b/repos/system_upgrade/el7toel8/actors/postgresqlcheck/libraries/postgresqlcheck.py
@@ -18,7 +18,7 @@ report_server_inst_hint = (
 )
 
 # Link URL for postgresql-server report
-report_server_inst_link_url = 'https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/deploying_different_types_of_servers/index#migrating-to-a-rhel-8-version-of-postgresql_using-postgresql'  # noqa: E501; pylint: disable=line-too-long
+report_server_inst_link_url = 'https://red.ht/rhel-8-migrate-postgresql-server'
 
 # List of dropped extensions from postgresql-contrib package
 report_contrib_inst_dropext = ['dummy_seclabel', 'test_parser', 'tsearch2']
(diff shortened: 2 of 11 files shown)


pirat89:
@Rezney do you want this one merge before you do the check of problematic URLs and cover them all at once or do you want this one covered prior the merge?

Rezney:
> @Rezney do you want this one merge before you do the check of problematic URLs and cover them all at once or do you want this one covered prior the merge?

@pirat89 I am fine with merging this.

repos/system_upgrade/el7toel8/actors/postgresqlcheck/libraries/postgresqlcheck.py
```diff
@@ -18,7 +18,7 @@
 )
 
 # Link URL for postgresql-server report
-report_server_inst_link_url = 'https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/deploying_different_types_of_servers/index#migrating-to-a-rhel-8-version-of-postgresql_using-postgresql'  # noqa: E501; pylint: disable=line-too-long
+report_server_inst_link_url = 'https://red.ht/rhel-8-deploying-different-types-of-servers'
```
Rezney, line 21: Are we sure this one is really the correct replacement?

repos/system_upgrade/common/actors/localreposinhibit/actor.py
```diff
@@ -82,8 +82,7 @@ def process(self):
                             "in-place upgrade"
                         ),
                         url=(
-                            "https://access.redhat.com/articles/4977891/"
-                            "#repos-known-issues"
+                            "https://red.ht/ipu-customisation-repos-known-issues"
```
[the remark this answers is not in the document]
pirat89: the summary speaks about how to setup apache server etc. which is covered in the known-issues section

repos/system_upgrade/el7toel8/actors/postgresqlcheck/libraries/postgresqlcheck.py
pirat89: @Rezney  good catch! fixed

Rezney (approved):

pirat89:
@jesusserranost  thanks, the link has been updated.

PR #1088, opened 2023-06-12
support upgrade el7to8 on AlibabaCloud rhui

jinkkkang:
support  rhel 7  upgrade to rhel 8  on alibaba cloud

### DIFF
--- a/etc/leapp/files/repomap.json
+++ b/etc/leapp/files/repomap.json
@@ -1,6 +1,6 @@
 {
-    "datetime": "202306191741Z",
-    "version_format": "1.1.0",
+    "datetime": "202307241553Z",
+    "version_format": "1.2.0",
     "mapping": [
         {
             "source_major_version": "7",
@@ -104,6 +104,12 @@
                     "source": "rhel7-rhui-google-compute-engine",
                     "target": [
                         "rhel8-rhui-google-compute-engine-leapp"
+                    ]
+                },
+                {
+                    "source": "rhel7-rhui-custom-client-at-alibaba",
+                    "target": [
+                        "rhel8-rhui-custom-client-at-alibaba"
                     ]
                 }
             ]
@@ -403,6 +409,14 @@
                     "channel": "ga",
                     "repo_type": "rpm",
                     "rhui": "google"
+                },
+                {
+                    "major_version": "7",
+                    "repoid": "rhui-rhel-7-server-rhui-rpms",
+                    "arch": "x86_64",
+                    "channel": "ga",
+                    "repo_type": "rpm",
+                    "rhui": "alibaba"
                 }
             ]
         },
@@ -577,6 +591,14 @@
                     "channel": "ga",
                     "repo_type": "rpm",
                     "rhui": "google"
+                },
+                {
+                    "major_version": "7",
+                    "repoid": "rhui-rhel-7-server-rhui-optional-rpms",
+                    "arch": "x86_64",
+                    "channel": "ga",
+                    "repo_type": "rpm",
+                    "rhui": "alibaba"
                 }
             ]
         },
@@ -784,6 +806,14 @@
                     "channel": "ga",
                     "repo_type": "rpm",
                     "rhui": "google"
+                },
+                {
+                    "major_version": "7",
+                    "repoid": "rhui-rhel-7-server-rhui-extras-rpms",
+                    "arch": "x86_64",
+                    "channel": "ga",
+                    "repo_type": "rpm",
+                    "rhui": "alibaba"
                 }
             ]
         },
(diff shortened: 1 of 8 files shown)


repos/system_upgrade/common/libraries/rhui.py
```diff
@@ -112,6 +112,17 @@
                 ('leapp-google-sap.repo', YUM_REPOS_PATH)
             ],
         },
+        'alibaba': {
+            'src_pkg': 'client-rhel7',
+            'target_pkg': 'aliyun_rhui_rhel8.6',
```
Rezney, line 117: Could you please elaborate on this `aliyun_rhui_rhel8.6` package? For a correct functionality we require not versioned RHUI client package which contains repositories using $releasever. Otherwise we would have to always change target_pkg for very release and could not support multiple upgrade paths.

Rezney (commented):
Thank you very much for your contribution. It is nice to see you already got used to our including Leapp metadata in our repository. Good job! Would you be able to share also the "leapp-rhui-alibaba' package?

repos/system_upgrade/common/libraries/rhui.py
jinkkkang: This package name contains version information, which may lead to ambiguity as it only applies to rhel 8.6. In fact, it applies to all rhel 8 versions, and the source address used $releasever,so please ignore the target pkg name

repos/system_upgrade/common/actors/redhatsignedrpmscanner/actor.py
```diff
@@ -56,7 +56,7 @@ def has_katello_prefix(pkg):
 
         upg_path = rhui.get_upg_path()
         # AWS RHUI packages do not have to be whitelisted because they are signed by RedHat
-        whitelisted_cloud_flavours = ('azure', 'azure-eus', 'azure-sap-ha', 'azure-sap-apps', 'google', 'google-sap')
+        whitelisted_cloud_flavours = ('azure', 'azure-eus', 'azure-sap-ha', 'azure-sap-apps', 'google', 'google-sap','alibaba')
```
fernflower, line 59: missing whitespace after comma

etc/leapp/files/repomap.json
```diff
@@ -105,6 +105,13 @@
                     "target": [
                         "rhel8-rhui-google-compute-engine-leapp"
                     ]
+                },
+                {                   
+                    "source": "rhui-custom-rhui_client_at_alibaba",
```
fernflower, line 110: Not sure change to repomap.json file belongs here :(
Yes, we started shipping leapp data files with the package, but repomapping is automatically generated and changing it here will last only till the next version is shipped.
I believe we have to change the data files that are used to generate the `repomap.json`, not the file itself. The problem is - they are not in this repo, so I believe this should be done by us -> the repomap.json should be regenerated -> the change will be shipped in a different commit.
@pirat89 @mm0ran @mmacura311 @Rezney what's your opinion on this?

fernflower (dismissed):

fernflower:
Please fix linter errors like 
```
 --- Running pylint ---
************* Module actor
[C0301 line-too-long] Line too long (127/119) File: repos/system_upgrade/common/actors/redhatsignedrpmscanner/actor.py, line 59, in 
```

etc/leapp/files/repomap.json
pirat89: @fernflower it's not supposed to be merged yet - it's waiting for @Rezney  when he returns from the vacation. yes, once it's decided this is going to be merged, the new data must be generated - for that, we will need to get "source of truth" for alibaba repositories.

@jinkkkang can you please provide repofiles from alibaba machine which we could use for our internal script? as the repomap data is generated, without updating our internals repositories would get removed if we do not update the script. regarding additional internal we use in the script, we cannot make it public, only the data

pirat89:
Related update to update the json schema:
* https://github.com/oamg/schema-test/pull/11

pirat89:
@jinkkkang Hi \o can you please check the changes in the repomap.json file? If it's correct, rebase please agaist up-to-date upstream master branch (or I can do it if you are ok that I change your commits). Also I want to ensure that `rhui-custom-rhui_client_at_alibaba` is the expected repoid on RHEL 7. To me it seems to be wrong, but that's what I found regarding the repo files you shared with me.

jinkkkang:
> @jinkkkang Hi \o can you please check the changes in the repomap.json file I did in the last commit? If it's correct, rebase please agaist up-to-date upstream master branch (or I can do it if you are ok that I change your commits). Also I want to ensure that `rhui-custom-rhui_client_at_alibaba` is the expected repoid on RHEL 7. To me it seems to be wrong, but that's what I found regarding the repo files you shared with me.

@pirat89   we have verified that the update to repomap file are OK, it's ok to change  repomap.json in my commit, just do it,  and the  `rhui-custom-rhui_client_at_alibaba`  name is ok,  the name on the rhel7 system is indeed  this, we name `rhui-custom-rhui_client_at_alibaba-rhel-8` in rhel 8 repo.

etc/leapp/files/repomap.json
pirat89: @fernflower repomap file has been updated and some other changes in the code applied. see my 2nd commit.

pirat89 (approved):
merging. thanks @jinkkkang  !!

PR #1089, opened 2023-06-14
Use correct flag and ENV var to disable insights registration

cdwertmann:


pirat89 (approved):
@cdwertmann thanks for the correction!

fernflower:
Could you please add some longer description to the commit message? Sorry for nitpicking, something straightforward like "previously a wrong --no-insights-autoregister flag was used instead of the correct --no-insights-register one" would do.

pirat89:
@cdwertmann or if you are ok with that, I can update the commit msg via the squash&merge:

```
Doc: Fix doc for disabling registration to RH Insights

The original document speaks about `LEAPP_NO_INSIGHTS_AUTOREGISTER` and the  `--no-insights-autoregister` option.
However the correct envar is `LEAPP_NO_INSIGHTS_REGISTER` and the option is `--no-insights-register`
```

pirat89:
@cdwertmann  thanks for the cnotribution \o

PR #1090, opened 2023-06-15
Preupgrade fails with a devtmpfs entry in fstab

bessonc:
systemd-nspawn fails when a devtmpfs entry is present in /etc/fstab
   mknod(/var/lib/leapp/scratch/mounts/root_/system_overlay/dev/null) failed: File exists

Ignoring such vfstype fixes the issue rhbz#2215027

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=2215027
JIRA: OAMG-9299


pirat89 (approved):
Thanks @bessonc  for the contribution!! :rocket: 

Make sense and LGTM. Waiting with the merge to discuss it with QE.

PR #1092, opened 2023-06-19
Update leapp data files, provide streams: 1.1, 2.0

pirat89:
- Added mapping for missing Azure repos (NFV, supplementary, ...)
- Updated PES data
- Repomap data version 1.1.0 to reflect added provides_version field
- CONSUMED_DATA_STREAM_ID = 2.0

The new streams are not 1.1 and 2.0 to differenciate the change from the previous data files. Expecting to drop 1.x stream completely prior the release.


pirat89:
@oamg/developers be aware that if you test with RHSM, you have to manually choose the `--target` release as with this change the default is switched to `9.3` and `8.9` to meet the currently expected behaviour. We will discuss yet whether the default will be set in future to EUS or latest minor release always.

PR #1093, opened 2023-06-22
Update Grub on component devices if /boot is on mdraid device

matejmatuska:
On BIOS systems, previously, if /boot was on md device such as RAID
consisting of multiple partitions on different MBR/GPT partitioned
drives, the part of Grub residing in the 512 Mb after MBR was only
updated for one of the drives. Similar situation occurred on GPT
partitioned drives and the BIOS boot partition. This resulted in
outdated GRUB on the remaining drives which could cause the system to be
unbootable.

Now, Grub is updated on all the component devices of an md array if Grub
was already installed on them before the upgrade.

Jira: OAMG-7835
BZ#2219544
BZ#2140011

### DIFF
--- a/repos/system_upgrade/common/actors/scangrubdevice/actor.py
+++ b/repos/system_upgrade/common/actors/scangrubdevice/actor.py
@@ -7,7 +7,7 @@ from leapp.tags import FactsPhaseTag, IPUWorkflowTag
 
 class ScanGrubDeviceName(Actor):
     """
-    Find the name of the block device where GRUB is located
+    Find the name of the block devices where GRUB is located
     """
 
     name = 'scan_grub_device_name'
@@ -19,8 +19,7 @@ class ScanGrubDeviceName(Actor):
         if architecture.matches_architecture(architecture.ARCH_S390X):
             return
 
-        device_name = grub.get_grub_device()
-        if device_name:
-            self.produce(GrubInfo(orig_device_name=device_name))
-        else:
-            self.produce(GrubInfo())
+        devices = grub.get_grub_devices()
+        grub_info = GrubInfo(orig_devices=devices)
+        grub_info.orig_device_name = devices[0] if len(devices) == 1 else None
+        self.produce(grub_info)
--- a/repos/system_upgrade/common/actors/updategrubcore/libraries/updategrubcore.py
+++ b/repos/system_upgrade/common/actors/updategrubcore/libraries/updategrubcore.py
@@ -1,35 +1,43 @@
 from leapp import reporting
-from leapp.exceptions import StopActorExecution
 from leapp.libraries.stdlib import api, CalledProcessError, config, run
 
 
-def update_grub_core(grub_dev):
+def update_grub_core(grub_devs):
     """
     Update GRUB core after upgrade from RHEL7 to RHEL8
 
     On legacy systems, GRUB core does not get automatically updated when GRUB packages
     are updated.
     """
-    cmd = ['grub2-install', grub_dev]
-    if config.is_debug():
-        cmd += ['-v']
-    try:
-        run(cmd)
-    except CalledProcessError as err:
-        reporting.create_report([
-            reporting.Title('GRUB core update failed'),
-            reporting.Summary(str(err)),
-            reporting.Groups([reporting.Groups.BOOT]),
-            reporting.Severity(reporting.Severity.HIGH),
-            reporting.Remediation(
-                hint='Please run "grub2-install <GRUB_DEVICE>" manually after upgrade'
-            )
-        ])
-        api.current_logger().warning('GRUB core update on {} failed'.format(grub_dev))
-        raise StopActorExecution()
+
+    successful = []
+    failed = []
+    for dev in grub_devs:
+        cmd = ['grub2-install', dev]
+        if config.is_debug():
+            cmd += ['-v']
+        try:
+            run(cmd)
+        except CalledProcessError as err:
+            api.current_logger().warning('GRUB core update on {} failed: {}'.format(dev, err))
+            failed.append(dev)
+            continue
+
+        successful.append(dev)
+
+    reporting.create_report([
+        reporting.Title('GRUB core update failed'),
+        reporting.Summary('Leapp failed to update GRUB on {}'.format(', '.join(failed))),
+        reporting.Groups([reporting.Groups.BOOT]),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Remediation(
+            hint='Please run "grub2-install <GRUB_DEVICE>" manually after upgrade'
+        )
+    ])
+
     reporting.create_report([
         reporting.Title('GRUB core successfully updated'),
-        reporting.Summary('GRUB core on {} was successfully updated'.format(grub_dev)),
+        reporting.Summary('GRUB core on {} was successfully updated'.format(', '.join(successful))),
         reporting.Groups([reporting.Groups.BOOT]),
         reporting.Severity(reporting.Severity.INFO)
     ])
--- a/repos/system_upgrade/common/libraries/grub.py
+++ b/repos/system_upgrade/common/libraries/grub.py
@@ -1,7 +1,9 @@
 import os
 
 from leapp.exceptions import StopActorExecution
+from leapp.libraries.common import mdraid
 from leapp.libraries.stdlib import api, CalledProcessError, run
+from leapp.utils.deprecation import deprecated
 
 
 def has_grub(blk_dev):
@@ -59,6 +61,32 @@ def get_boot_partition():
     return boot_partition
 
 
+def get_grub_devices():
+    """
+    Get block devices where GRUB is located. We assume GRUB is on the same device
+    as /boot partition is. In case that device is an md (Multiple Device) device, all
+    of the component devices of such a device are considered.
+
+    :return: Devices where GRUB is located
+    :rtype: list
+    """
+    boot_device = get_boot_partition()
+    devices = []
+    if mdraid.is_mdraid_dev(boot_device):
+        component_devs = mdraid.get_component_devices(boot_device)
+        blk_devs = [blk_dev_from_partition(dev) for dev in component_devs]
+        # remove duplicates as there might be raid on partitions on the same drive
+        # even if that's very unusual
+        devices = sorted(list(set(blk_devs)))
+    else:
+        devices.append(blk_dev_from_partition(boot_device))
+
+    have_grub = [dev for dev in devices if has_grub(dev)]
+    api.current_logger().info('GRUB is installed on {}'.format(",".join(have_grub)))
+    return have_grub
+
+
+@deprecated(since='2023-06-23', message='This function has been replaced by get_grub_devices')
 def get_grub_device():
     """
     Get block device where GRUB is located. We assume GRUB is on the same device
--- a/repos/system_upgrade/common/libraries/mdraid.py
+++ b/repos/system_upgrade/common/libraries/mdraid.py
@@ -1,0 +1,48 @@
+from leapp.libraries.stdlib import api, CalledProcessError, run
+
+
+def is_mdraid_dev(dev):
+    """
+    Check if a given device is an md (Multiple Device) device
+
+    It is expected that the "mdadm" command is available,
+    if it's not it is assumed the device is not an md device.
+
+    :return: True if the device is an md device, False otherwise
+    :raises CalledProcessError: If an error occurred
+    """
+    fail_msg = 'Could not check if device "{}" is an md device: {}'
+    try:
+        result = run(['mdadm', '--query', dev])
+    except OSError as err:
+        api.current_logger().warning(fail_msg.format(dev, err))
+        return False
+    except CalledProcessError as err:
+        err.message = fail_msg.format(dev, err)
+        raise  # let the calling actor handle the exception
+
+    return '--detail' in result['stdout']
+
+
+def get_component_devices(raid_dev):
+    """
+    Get list of component devices in an md (Multiple Device) array
+
+    :return: The list of component devices or None in case of error
+    :raises ValueError: If the device is not an mdraid device
+    """
+    try:
+        # using both --verbose and --brief for medium verbosity
+        result = run(['mdadm', '--detail', '--verbose', '--brief', raid_dev])
+    except (OSError, CalledProcessError) as err:
+        api.current_logger().warning(
+            'Could not get md array component devices: {}'.format(err)
+        )
+        return None
+    # example output:
+    # ARRAY /dev/md0 level=raid1 num-devices=2 metadata=1.2 name=localhost.localdomain:0 UUID=c4acea6e:d56e1598:91822e3f:fb26832c # noqa: E501; pylint: disable=line-too-long
+    #     devices=/dev/vda1,/dev/vdb1
+    if 'does not appear to be an md device' in result['stdout']:
+        raise ValueError("Expected md device, but got: {}".format(raid_dev))
+
+    return sorted(result['stdout'].rsplit('=', 2)[-1].strip().split(','))
--- a/repos/system_upgrade/common/libraries/mdraid.py
+++ b/repos/system_upgrade/common/libraries/mdraid.py
@@ -1,3 +1,5 @@
+import os
+
 from leapp.libraries.stdlib import api, CalledProcessError, run
 
 
@@ -12,11 +14,13 @@ def is_mdraid_dev(dev):
     :raises CalledProcessError: If an error occurred
     """
     fail_msg = 'Could not check if device "{}" is an md device: {}'
+    if not os.path.exists('/usr/sbin/mdadm'):
+        api.current_logger().warning(fail_msg.format(
+            dev, '/usr/sbin/mdadm is not installed.'
+        ))
+        return False
     try:
         result = run(['mdadm', '--query', dev])
-    except OSError as err:
-        api.current_logger().warning(fail_msg.format(dev, err))
-        return False
     except CalledProcessError as err:
         err.message = fail_msg.format(dev, err)
         raise  # let the calling actor handle the exception
--- a/repos/system_upgrade/common/actors/checkgrubcore/actor.py
+++ b/repos/system_upgrade/common/actors/checkgrubcore/actor.py
@@ -32,7 +32,7 @@ class CheckGrubCore(Actor):
             grub_info = next(self.consume(GrubInfo), None)
             if not grub_info:
                 raise StopActorExecutionError('Actor did not receive any GrubInfo message.')
-            if grub_info.orig_device_name:
+            if grub_info.orig_devices:
                 create_report([
                     reporting.Title(
                         'GRUB2 core will be automatically updated during the upgrade'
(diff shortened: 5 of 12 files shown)


repos/system_upgrade/common/actors/scangrubdevice/actor.py
```diff
@@ -19,8 +19,8 @@ def process(self):
         if architecture.matches_architecture(architecture.ARCH_S390X):
             return
 
-        device_name = grub.get_grub_device()
-        if device_name:
-            self.produce(GrubInfo(orig_device_name=device_name))
+        devices = grub.get_grub_devices()
```
fernflower, line 22: IIUC orig_devices is an empty list by default, so L22-L26 should be equivalent to
```
self.produce(GrubInfo(orig_devices=devices))
```

repos/system_upgrade/common/libraries/mdraid.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def get_component_devices(raid_dev):
+    try:
+        result = run(['mdadm', '--detail', '--verbose', '--brief', raid_dev])
+    except CalledProcessError as err:
+        api.current_logger().warning(
+            'Could not get md array component devices: {}'.format(err)
+        )
+        raise StopActorExecution()
```
fernflower, line 22: Would not it be more informative to add exception details to StopActorExecution? Smth like
`raise StopActorExecution('Could not get md array component devices', details={'details': err})` maybe?

repos/system_upgrade/common/libraries/mdraid.py
```diff
@@ -0,0 +1,26 @@
...
+from leapp.libraries.stdlib import api, CalledProcessError, run
+
+def is_mdraid_dev(dev):
+    try:
+        result = run(['mdadm', '--query', dev])
+    except CalledProcessError:
+        api.current_logger().warning(
+            'Could not check if device is an md device'
+        )
+        raise StopActorExecution()
```
fernflower, line 11: Same as below: maybe it's worth adding details to StopActorExecution?

repos/system_upgrade/common/libraries/grub.py
```diff
@@ -59,6 +60,22 @@ def get_boot_partition():
     return boot_partition
 
 
+def get_grub_devices():
+    boot_device = get_boot_partition()
+    devices = []
+    if mdraid.is_mdraid_dev(boot_device):
+        component_devs = mdraid.get_component_devices(boot_device)
+        blk_devs = [blk_dev_from_partition(dev) for dev in component_devs]
```
fernflower, line 68: Nit: IIUC you can assign directly to `devices`, removing the need for `blk_devs` var completely
```
devices = [blk_dev_from_partition(dev) for dev in component_devs]
```

repos/system_upgrade/common/actors/updategrubcore/libraries/updategrubcore.py
```diff
@@ -3,33 +3,39 @@
...
+                reporting.Title('GRUB core update failed'),
+                reporting.Summary(str(err)),
+                reporting.Groups([reporting.Groups.BOOT]),
+                reporting.Severity(reporting.Severity.HIGH),
+                reporting.Remediation(
+                    hint='Please run "grub2-install <GRUB_DEVICE>" manually after upgrade'
+                )
+            ])
+            api.current_logger().warning('GRUB core update on {} failed'.format(dev))
+            raise StopActorExecution()
```
fernflower, line 32: Same question about passing details to StopActorExecution exception as in mdraid library

repos/system_upgrade/common/actors/updategrubcore/libraries/updategrubcore.py
```diff
@@ -3,33 +3,39 @@
...
-        raise StopActorExecution()
+
+    successful = []
+    for dev in grub_devs:
+        cmd = ['grub2-install', dev]
+        if config.is_debug():
+            cmd += ['-v']
+        try:
+            run(cmd)
+        except CalledProcessError as err:
```
fernflower, line 21: IIUC the logic of grub devices processing/report creation here, then the first failing entry will end up being reported. Is it intended to report just the first error? (just asking)

fernflower (commented):
Please see some questions/remarks inline.
BTW, would it be possible to add some unit tests? It seems like a significant change that definitely deserves some :)

repos/system_upgrade/common/actors/updategrubcore/libraries/updategrubcore.py
matejmatuska: Still not sure what would be the right thing to do here. Maybe we could try to update it on as much devices as possible?

repos/system_upgrade/common/actors/updategrubcore/libraries/updategrubcore.py
matejmatuska: StopActorExecution != StopActorExecutionError, the former just terminates the actor, allowing the leapp execution to continue. I believe you can't even add the message there.

The same goes for the other places you mentioned.

repos/system_upgrade/common/libraries/mdraid.py
matejmatuska: Here I changed the return type to None, I think that makes more sense.

matejmatuska:
Tested on RHEL 7.9->8.8 and 8.8->9.2
With the following disk configuration:
```
# lsblk -so "NAME,SIZE,TYPE,FSTYPE,MOUNTPOINT"
NAME     SIZE TYPE  FSTYPE            MOUNTPOINT
md126     12G raid5 xfs               /
├─vda2     4G part  linux_raid_member
│ └─vda    5G disk
├─vdb2     4G part  linux_raid_member
│ └─vdb    5G disk
├─vdc2     4G part  linux_raid_member
│ └─vdc    5G disk
└─vdd2     4G part  linux_raid_member
  └─vdd    5G disk
md127   1023M raid1 xfs               /boot
├─vda1     1G part  linux_raid_member
│ └─vda    5G disk
├─vdb1     1G part  linux_raid_member
│ └─vdb    5G disk
├─vdc1     1G part  linux_raid_member
│ └─vdc    5G disk
└─vdd1     1G part  linux_raid_member
  └─vdd    5G disk
sr0     1024M rom
```
(4 drives, /boot on RAID1, / on RAID5)

Without this patch the upgrades failed to boot into the system after initramfs
phase. GRUB jumps into rescue with `symbol ... not found error`.

With this patch applied, this problem is fixed on both RHEL versions. In the scan phase the
scangrubdevice actor correctly identifies all boot devices:
`leapp.workflow.FactsCollection.scan_grub_device_name: GRUB is installed on /dev/vdc,/dev/vdb,/dev/vda,/dev/vdd`

And in the RPMUpgrade phase grub2-install is correctly called on each of those
devices:
```
leapp.workflow.RPMUpgrade.update_grub_core: External command has started: ['grub2-install', u'/dev/vdc', '-v']
...
leapp.workflow.RPMUpgrade.update_grub_core: External command has finished: ['grub2-install', u'/dev/vdc', '-v']
leapp.workflow.RPMUpgrade.update_grub_core: External command has started: ['grub2-install', u'/dev/vdc', '-v']
...
leapp.workflow.RPMUpgrade.update_grub_core: External command has finished: ['grub2-install', u'/dev/vdc', '-v']
leapp.workflow.RPMUpgrade.update_grub_core: External command has started: ['grub2-install', u'/dev/vdc', '-v']
...
leapp.workflow.RPMUpgrade.update_grub_core: External command has finished: ['grub2-install', u'/dev/vdc', '-v']
leapp.workflow.RPMUpgrade.update_grub_core: External command has started: ['grub2-install', u'/dev/vdc', '-v']
...
leapp.workflow.RPMUpgrade.update_grub_core: External command has finished: ['grub2-install', u'/dev/vdc', '-v']
```
The system then properly boots and the upgrade continues smoothly.

repos/system_upgrade/common/actors/updategrubcore/libraries/updategrubcore.py
fernflower: Yeah, you are right, it stops without reporting anything https://github.com/oamg/leapp/blob/master/leapp/actors/__init__.py#L290
Sorry, it was an early Friday review, disregard my comments then!

repos/system_upgrade/common/actors/updategrubcore/libraries/updategrubcore.py
fernflower: Yeah, I would probably update all and in case any error is reported - raise an inhibitor with a full report.

matejmatuska:
> Tested on RHEL 7.9->8.8 and 8.8->9.2 With the following disk configuration:
> 
> ```
> # lsblk -so "NAME,SIZE,TYPE,FSTYPE,MOUNTPOINT"
> NAME     SIZE TYPE  FSTYPE            MOUNTPOINT
> md126     12G raid5 xfs               /
> ├─vda2     4G part  linux_raid_member
> │ └─vda    5G disk
> ├─vdb2     4G part  linux_raid_member
> │ └─vdb    5G disk
> ├─vdc2     4G part  linux_raid_member
> │ └─vdc    5G disk
> └─vdd2     4G part  linux_raid_member
>   └─vdd    5G disk
> md127   1023M raid1 xfs               /boot
> ├─vda1     1G part  linux_raid_member
> │ └─vda    5G disk
> ├─vdb1     1G part  linux_raid_member
> │ └─vdb    5G disk
> ├─vdc1     1G part  linux_raid_member
> │ └─vdc    5G disk
> └─vdd1     1G part  linux_raid_member
>   └─vdd    5G disk
> sr0     1024M rom
> ```
> 
> (4 drives, /boot on RAID1, / on RAID5)
> 
> Without this patch the upgrades failed to boot into the system after initramfs phase. GRUB jumps into rescue with `symbol ... not found error`.
> 
> With this patch applied, this problem is fixed on both RHEL versions. In the scan phase the scangrubdevice actor correctly identifies all boot devices: `leapp.workflow.FactsCollection.scan_grub_device_name: GRUB is installed on /dev/vdc,/dev/vdb,/dev/vda,/dev/vdd`
> 
> And in the RPMUpgrade phase grub2-install is correctly called on each of those devices:
> 
> ```
> leapp.workflow.RPMUpgrade.update_grub_core: External command has started: ['grub2-install', u'/dev/vdc', '-v']
> ...
> leapp.workflow.RPMUpgrade.update_grub_core: External command has finished: ['grub2-install', u'/dev/vdc', '-v']
> leapp.workflow.RPMUpgrade.update_grub_core: External command has started: ['grub2-install', u'/dev/vdc', '-v']
> ...
> leapp.workflow.RPMUpgrade.update_grub_core: External command has finished: ['grub2-install', u'/dev/vdc', '-v']
> leapp.workflow.RPMUpgrade.update_grub_core: External command has started: ['grub2-install', u'/dev/vdc', '-v']
> ...
> leapp.workflow.RPMUpgrade.update_grub_core: External command has finished: ['grub2-install', u'/dev/vdc', '-v']
> leapp.workflow.RPMUpgrade.update_grub_core: External command has started: ['grub2-install', u'/dev/vdc', '-v']
> ...
> leapp.workflow.RPMUpgrade.update_grub_core: External command has finished: ['grub2-install', u'/dev/vdc', '-v']
> ```
> 
> The system then properly boots and the upgrade continues smoothly.

This should still be up to date.

pirat89 (changes requested):
@matejmatuska I have discussed the PR with @pholica and we have realized that actually it fixes different things than to which refers and commit msg actually should be also different. let's sync about it later.

pholica:
Going through the code quickly, it looks like this should be fixing situation when /boot is on RAID1 on BIOS system with MBR partition table, however as @pirat89 said, that's not situation of https://en.wikipedia.org/wiki/BIOS_boot_partition (referenced Jira task by this PR) neither /boot and /boot/efi being on RAID1 on UEFI (which are other parts of "SW RAID support" needed to be implemented to fix the referenced bug.

It sounds to me like the only problematic thing (please correct me if I'm wrong) when it comes to this PR is just bug/issue references and lack of mention of BIOS and MBR in the description.

matejmatuska:
> Going through the code quickly, it looks like this should be fixing situation when /boot is on RAID1 on BIOS system with MBR partition table, however as @pirat89 said, that's not situation of https://en.wikipedia.org/wiki/BIOS_boot_partition (referenced Jira task by this PR) neither /boot and /boot/efi being on RAID1 on UEFI (which are other parts of "SW RAID support" needed to be implemented to fix the referenced bug.
> 
> It sounds to me like the only problematic thing (please correct me if I'm wrong) when it comes to this PR is just bug/issue references and lack of mention of BIOS and MBR in the description.

I see the porblem now, for a moment I missed the fact that there are GPT disks in the BZ. Yes, this should solve the problem for MBR partitoned disks on BIOS, I will update the description. However seems like the BZ reporter fixed the bug in a similar fashion, so I will do some testing with GPT disks also and see if we can fix that here too.

repos/system_upgrade/common/libraries/mdraid.py
```diff
@@ -0,0 +1,41 @@
...
+
+    :raises: StopActorExecution in case of error
+    """
+    try:
+        result = run(['mdadm', '--query', dev])
+    except CalledProcessError as err:
+        api.current_logger().warning(
+                'Could not check if device is an md device: {}'.format(err)
+        )
+        raise StopActorExecution()
```
pirat89, line 17: @matejmatuska this is wrong. this function is in public library, which means it can be called by whatever actor. And right now, it's expected the function return True when it is raid device, or false when is not. In systems without the raid, it's possible typical that `mdadm` is not installed, in such a case the function should return False. Stopping an actor execution should be done inside the actor - not inside the public library function in this case (unless we speak about hard error, which this is not).

pirat89: probably best solution would be to check whether mdadm is installed. in case it is installed and the call crashes, than probably would make sense to actually expose the calledprocesserror and keep the handlin on an actor.

pirat89:
> > Going through the code quickly, it looks like this should be fixing situation when /boot is on RAID1 on BIOS system with MBR partition table, however as @pirat89 said, that's not situation of https://en.wikipedia.org/wiki/BIOS_boot_partition (referenced Jira task by this PR) neither /boot and /boot/efi being on RAID1 on UEFI (which are other parts of "SW RAID support" needed to be implemented to fix the referenced bug.
> > It sounds to me like the only problematic thing (please correct me if I'm wrong) when it comes to this PR is just bug/issue references and lack of mention of BIOS and MBR in the description.
> 
> @pholica @pirat89 I see the porblem now, for a moment I missed the fact that there are GPT disks in the BZ. Yes, this should solve the problem for MBR partitoned disks on BIOS, I will update the description. However seems like the BZ reporter fixed the bug in a similar fashion, so I will do some testing with GPT disks also and see if we can fix that here too.

for the GPT, it would need more work as currently we do not have even detection of the grub for GPT partitioned disks nat all.

matejmatuska:
@pirat89 can you please address the TODO in PR description, not sure what the process is there?

repos/system_upgrade/common/libraries/mdraid.py
matejmatuska: Should be resolved now

repos/system_upgrade/common/actors/scangrubdevice/actor.py
```diff
@@ -19,8 +19,5 @@ def process(self):
         if architecture.matches_architecture(architecture.ARCH_S390X):
             return
 
-        device_name = grub.get_grub_device()
-        if device_name:
-            self.produce(GrubInfo(orig_device_name=device_name))
-        else:
-            self.produce(GrubInfo())
+        devices = grub.get_grub_devices()
```
pirat89, line 22: @matejmatuska actually, as the `orig_device_name` is deprecated now, you should fill it also when only one grub device is set, otherwise it's not deprecation but removal with keeping dead fields present

repos/system_upgrade/common/actors/updategrubcore/libraries/updategrubcore.py
```diff
@@ -1,35 +1,42 @@
...
+    failed = []
+    for dev in grub_devs:
+        cmd = ['grub2-install', dev]
+        if config.is_debug():
+            cmd += ['-v']
+        try:
+            run(cmd)
+        except CalledProcessError as err:
+            api.current_logger().warning('GRUB core update on {} failed: {}'.format(dev, err))
+            failed.append(dev)
```
pirat89, line 23: @matejmatuska  missing the `continue` statement. in this case, you will add the device the both: `failed` and `successful` lists. seems to me that unit-tests did not cover this situation correctly in such a case

repos/system_upgrade/common/actors/scangrubdevice/actor.py
matejmatuska: That raises a question, what should I put into orig_device_name if there is more than one device?

matejmatuska: Should I preserve the original behavior in such case? @pirat89

pirat89: In such a case i would keep it None as it was already incorrect. But in case of the one device, I would fill it.

repos/system_upgrade/common/libraries/grub.py
```diff
@@ -59,6 +61,32 @@ def get_boot_partition():
...
+    :rtype: list
+    """
+    boot_device = get_boot_partition()
+    devices = []
+    if mdraid.is_mdraid_dev(boot_device):
+        component_devs = mdraid.get_component_devices(boot_device)
+        blk_devs = [blk_dev_from_partition(dev) for dev in component_devs]
+        # remove duplicates as there might be raid on partitions on the same drive
+        # even if that's very unusual
+        devices = list(set(blk_devs))
```
pirat89, line 80: I prefer to have it sorted so the order is not so random.
```suggestion
        devices = sorted(list(set(blk_devs)))
```

repos/system_upgrade/common/libraries/mdraid.py
```diff
@@ -0,0 +1,48 @@
...
+            'Could not get md array component devices: {}'.format(err)
+        )
+        return None
+    # example output:
+    # ARRAY /dev/md0 level=raid1 num-devices=2 metadata=1.2 name=localhost.localdomain:0 UUID=c4acea6e:d56e1598:91822e3f:fb26832c # noqa: E501; pylint: disable=line-too-long
+    #     devices=/dev/vda1,/dev/vdb1
+    if 'does not appear to be an md device' in result['stdout']:
+        raise ValueError("Expected md device, but got: {}".format(raid_dev))
+
+    return result['stdout'].rsplit('=', 2)[-1].strip().split(',')
```
pirat89, line 48: @matejmatuska  possibly this cuold be sorted too

pirat89:
Manual testing for https://bugzilla.redhat.com/show_bug.cgi?id=2219544:

* orig:
```
Stamp: 2023-07-17T09:56:02.523801Z
Actor: scan_grub_device_name
Phase: FactsCollection
Type: GrubInfo
Message_data:
{
    "orig_device_name": "/dev/vdb"
}

```
* new msg:
```
Actor: scan_grub_device_name
Phase: FactsCollection
Type: GrubInfo
Message_data:
{
    "orig_device_name": null, 
    "orig_devices": [
        "/dev/vdb", 
        "/dev/vda"
    ]
}
```

repos/system_upgrade/common/actors/updategrubcore/libraries/updategrubcore.py
matejmatuska: Added the missing continue and covered by tests.

pirat89:
@matejmatuska on normal systems it reproduces now 2 GrubInfo msgs, which is wrong. There must not be more than one:
```
                          PRODUCED MESSAGES                           
Stamp: 2023-07-17T11:00:21.381393Z
Actor: scan_grub_device_name
Phase: FactsCollection
Type: GrubInfo
Message_data:
{
    "orig_device_name": "/dev/vda", 
    "orig_devices": [
        "/dev/vda"
    ]
}
Stamp: 2023-07-17T11:00:21.355611Z
Actor: scan_grub_device_name
Phase: FactsCollection
Type: GrubInfo
Message_data:
{
    "orig_device_name": "/dev/vda", 
    "orig_devices": [
        "/dev/vda"
    ]
}

```

pirat89:
Tested manually
* system without raid - w/o mdadm installed
* system with raid (7 -> 8 -> 9)

works as expected. merging.

pirat89 (approved):
ltgm and works (see my comments)

PR #1097, opened 2023-07-10
Overlay redesign

pirat89:
The in-place upgrade itself requires to do some changes on the system to be able to perform the in-place upgrade itself - or even to be able to evaluate if the system is possible to upgrade. However, we do not want to (and must not) change the original system until we pass beyond the point of not return.

For that purposes we have to create a layer above the real host file system, where we can safely perform all operations without affecting the system setup, rpm database, etc. Currently overlay (OVL) technology showed it is capable to handle our requirements good enough - with some limitations.

However, the original design we used to compose overlay layer above the host system had number of problems:
* buggy calculation of the required free space for the upgrade RPM transaction
* consumed too much space to handle partitions formatted with XFS without ftype attributes (even tens GBs)
* bad UX as people had to manually adjust size of OVL disk images
* ...and couple of additional issues derivated from problems listed above

The new solution prepares a disk image (represented by sparse-file) and an overlay image for each mountpoint configured in /etc/fstab, excluding those with FS types noted in the `OVERLAY_DO_NOT_MOUNT` set. Such prepared OVL images are then composed together to reflect the real host filesystem. In the end everything is cleaned.

The composition could look like this:
```
  orig mountpoint -> disk img -> overlay img -> new mountpoint
  -------------------------------------------------------------
    /        -> root_        -> root_/ovl        -> root_/ovl/
    /boot    -> root_boot    -> root_boot/ovl    -> root_/ovl/boot
    /var     -> root_var     -> root_var/ovl     -> root_/ovl/var
    /var/lib -> root_var_lib -> root_var_lib/ovl -> root_/ovl/var/lib
    ...
```

The new solution can be now problematic for system with too many partitions and loop devices, as each disk image is loop mounted (that's same as before, but number of disk images will be bigger in total number).
For such systems we keep for now the possibility of the fallback to an old solution, which has number of issues mentioned above, but it's a trade of. To fallback to the old solution, set envar:
    LEAPP_OVL_LEGACY=1

Disk images created for OVL are formatted with XFS by default. In case of problems, it's possible to switch to Ext4 FS using:
    LEAPP_OVL_IMG_FS_EXT4=1
XFS is better optimized for our use cases (faster initialisation consuming less space). However we have reported several issues related to overlay images, that happened so far only on XFS filesystems. We are not sure about root causes, but having the possibility to switch to Ext4 seems to be wise. In case of issues, we can simple ask users to try the switch and see if the problem is fixed or still present.

# Still missing
- [x] Create sparse files, where each one reflects the free space of the partition it represents (can be always little bit lower)
  - [x] in case of partition hosting the scratch dir (where disk images are created), reflect the real storage will not contain these disk images in time of the transaction execution
- [x] Update various error messages to not provide misleading instructions (like LEAPP_OVL_SIZE...)
  - except the corner case described here https://github.com/oamg/leapp-repository/pull/1097#discussion_r1264530149
- [x] Implement heuristics to check there is enough disk space on the partition hosting scratch dirs & leapp friends
  - [x] Optimize the required disk space - taking into account already existing target userspace container, ....
    - [x] optimized for IPU 7 -> 8 (first iteration quality)
    - [x] optimize for IPU 8 -> 9 (first iteration quality).
    - [x] handle missing disk space for the installation inside the container when preparing for upgrade initramfs deps
    - [x] improve handling of persistant cache which can easily consume all the space right now
  - [x] Cleanup disk images (just realized they are not cleaned at all)
    - previously created images consume a space and affects also guessing the requried free space
- [ ] Handle bind mounts so we do not consume extra space for them
- [x] Update existing unit-tests
- [ ] Create additional unit-tests to cover new functionality 

## Various outputs visible
Note: texts could be changed in the final version. I am posting outputs continuously as it will be easier to understand all changes and we will have at least some output giving a hint it actually works (some of these scenarious could be hard to hit without without changes of magical constants, just showind these cases too for better visibility).

* newly covered issues with the missing space when installing new pkgs+deps for initramfs inside the container (originally uncaught exception that provided misleading information what partition needs more space):
```
2023-07-14 16:05:28.245942 [ERROR] Actor: upgrade_initramfs_generator
Message: There is not enough space on the file system hosting /var/lib/leapp.
Summary:
    Hint: Increase the free space on the filesystem hosting /var/lib/leapp by 534MB at minimum. It is suggested to provide reasonably more space to be able to perform all planned actions (e.g. when 200MB is missing, add 1700MB or more).
```

* Protect the space consumption during the generation of the upgrade initramfs. On RHEL 7, generating RHEL 8 initramfs requires ~ 400 MB. Added check that at least 500MB is free prior the creation of the upgrade initramfs (note this requires extra preparation to be able to hit it):
```
2023-07-14 18:22:25.683470 [ERROR] Actor: upgrade_initramfs_generator
Message: There is not enough space on the file system hosting /var/lib/leapp.
Summary:
    Detail: Remaining free space is lower than 500MB which is not enough to be able to generate the upgrade initramfs.
    Hint: Increase the free space on the filesystem hosting /var/lib/leapp by 500MB at minimum (suggested 1500MB).
          
          It is also a good practice to create dedicated partition for for /var/lib/leapp when more space is needed, which can be dropped after the system upgrade is fully completed. For more info, see: https://access.redhat.com/solutions/7011704
```

* the default initial check of the calculated required free space to compose source OVL (the requirement already counts with the space consumed by disk imgs + container + basic set of packages downloaded for the in-place upgrade transaction + initramfs creation stuff):
```
2023-07-15 18:40:15.704243 [ERROR] Actor: target_userspace_creator
Message: Not enough space available on /var/lib/leapp/scratch: Needed at least 898 MiB.
Summary:
    Detail: The file system hosting the /var/lib/leapp/scratch directory does not contain enough free space to proceed all parts of the in-place upgrade. Note the calculated required free space is the minimum derived from upgrades of minimal systems and the actual needed free space could be higher.
            Needed at least: 898 MiB.
            Suggested free space: 1898 MiB (or more).
```

* when DNF discover there is not enough space on the system to perform the upgrade transaction (I mean, after the download of pkgs, or dnf_dry_run):
```
2023-07-15 19:42:12.114097 [ERROR] Actor: dnf_package_download
Message: There is not enough space on some file systems to perform the upgrade transaction.
Summary:
    Disk requirements: At least 1116MB more space needed on the /usr filesystem.
                       At least 278MB more space needed on the /var filesystem.
    Hint: Increase the free space on listed filesystems. Presented values are required minimum calculated by RPM and it is suggested to provide reasonably more free space (e.g. when 200 MB is missing on /usr, add 1200MB or more).
```

### DIFF
--- a/repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
+++ b/repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
@@ -20,6 +20,7 @@ from leapp.utils.deprecation import suppress_deprecation
 
 INITRAM_GEN_SCRIPT_NAME = 'generate-initram.sh'
 DRACUT_DIR = '/dracut'
+DEDICATED_LEAPP_PART_URL = 'https://access.redhat.com/solutions/7011704'
 
 
 def _get_target_kernel_version(context):
@@ -231,6 +232,77 @@ def prepare_userspace_for_initram(context):
     _copy_files(context, files)
 
 
+def _get_fspace(path, convert_to_mibs=False, coefficient=1):
+    """
+    Return the free disk space on given path.
+
+    The default is in bytes, but if convert_to_mibs is True, return MiBs instead.
+
+    Raises OSError if nothing exists on the given `path`.
+
+    :param path: Path to an existing file or directory
+    :type path: str
+    :param convert_to_mibs: If True, convert the value to MiBs
+    :type convert_to_mibs: bool
+    :param coefficient: Coefficient to multiply the free space (e.g. 0.9 to have it 10% lower). Max: 1
+    :type coefficient: float
+    :rtype: int
+    """
+    # TODO(pstodulk): discuss the function params
+    # NOTE(pstodulk): This func is copied from the overlaygen.py lib
+    # probably it would make sense to make it public in the utils.py lib,
+    # but for now, let's keep it private
+    stat = os.statvfs(path)
+
+    coefficient = min(coefficient, 1)
+    fspace_bytes = int(stat.f_frsize * stat.f_bavail * coefficient)
+    if convert_to_mibs:
+        return int(fspace_bytes / 1024 / 1024)  # noqa: W1619; pylint: disable=old-division
+    return fspace_bytes
+
+
+def _check_free_space(context):
+    """
+    Raise StopActorExecutionError if there is less than 500MB of free space available.
+
+    If there is not enough free space in the context, the initramfs will not be
+    generated successfully and it's hard to discover what was the issue. Also
+    the missing space is able to kill the leapp itself - trying to write to the
+    leapp.db when the FS hosting /var/lib/leapp is full, kills the framework
+    and the actor execution too - so there is no gentle way to handle such
+    exceptions when it happens. From this point, let's rather check the available
+    space in advance and stop the execution when it happens.
+
+    It is not expected to hit this issue, but I was successful and I know
+    it's still possible even with all other changes (just it's much harder
+    now to hit it). So adding this seatbelt, that is not 100% bulletproof,
+    but I call it good enough.
+
+    Currently protecting last 500MB. In case of problems, we can increase
+    the value.
+    """
+    message = 'There is not enough space on the file system hosting /var/lib/leapp.'
+    hint = (
+        'Increase the free space on the filesystem hosting'
+        ' /var/lib/leapp by 500MB at minimum (suggested 1500MB).\n\n'
+        'It is also a good practice to create dedicated partition'
+        ' for /var/lib/leapp when more space is needed, which can be'
+        ' dropped after the system upgrade is fully completed.'
+        ' For more info, see: {}'
+        .format(DEDICATED_LEAPP_PART_URL)
+    )
+    detail = (
+        'Remaining free space is lower than 500MB which is not enough to'
+        ' be able to generate the upgrade initramfs. '
+    )
+
+    if _get_fspace(context.base_dir, convert_to_mibs=True) < 500:
+        raise StopActorExecutionError(
+            message=message,
+            details={'hint': hint, 'detail': detail}
+        )
+
+
 def generate_initram_disk(context):
     """
     Function to actually execute the init ramdisk creation.
@@ -238,6 +310,7 @@ def generate_initram_disk(context):
     Includes handling of specified dracut and kernel modules from the host when
     needed. The check for the 'conflicting' modules is in a separate actor.
     """
+    _check_free_space(context)
     env = {}
     if get_target_major_version() == '9':
         env = {'SYSTEMD_SECCOMP': '0'}
--- a/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
+++ b/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
@@ -1,5 +1,6 @@
 import itertools
 import os
+import shutil
 
 from leapp import reporting
 from leapp.exceptions import StopActorExecution, StopActorExecutionError
@@ -121,9 +122,12 @@ class _InputData(object):
 
 def _restore_persistent_package_cache(userspace_dir):
     if get_env('LEAPP_DEVEL_USE_PERSISTENT_PACKAGE_CACHE', None) == '1':
-        if os.path.exists(PERSISTENT_PACKAGE_CACHE_DIR):
-            with mounting.NspawnActions(base_dir=userspace_dir) as target_context:
-                target_context.copytree_to(PERSISTENT_PACKAGE_CACHE_DIR, '/var/cache/dnf')
+        if not os.path.exists(PERSISTENT_PACKAGE_CACHE_DIR):
+            return
+        dst_cache = os.path.join(userspace_dir, 'var', 'cache', 'dnf')
+        if os.path.exists(dst_cache):
+            run(['rm', '-rf', dst_cache])
+        shutil.move(PERSISTENT_PACKAGE_CACHE_DIR, dst_cache)
     # We always want to remove the persistent cache here to unclutter the system
     run(['rm', '-rf', PERSISTENT_PACKAGE_CACHE_DIR])
 
@@ -132,9 +136,9 @@ def _backup_to_persistent_package_cache(userspace_dir):
     if get_env('LEAPP_DEVEL_USE_PERSISTENT_PACKAGE_CACHE', None) == '1':
         # Clean up any dead bodies, just in case
         run(['rm', '-rf', PERSISTENT_PACKAGE_CACHE_DIR])
-        if os.path.exists(os.path.join(userspace_dir, 'var', 'cache', 'dnf')):
-            with mounting.NspawnActions(base_dir=userspace_dir) as target_context:
-                target_context.copytree_from('/var/cache/dnf', PERSISTENT_PACKAGE_CACHE_DIR)
+        src_cache = os.path.join(userspace_dir, 'var', 'cache', 'dnf')
+        if os.path.exists(src_cache):
+            shutil.move(src_cache, PERSISTENT_PACKAGE_CACHE_DIR)
 
 
 def _the_nogpgcheck_option_used():
--- a/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
+++ b/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
@@ -766,11 +766,13 @@ def perform():
 
     indata = _InputData()
     prod_cert_path = _get_product_certificate_path()
+    reserve_space = overlaygen.get_recommended_leapp_free_space(_get_target_userspace())
     with overlaygen.create_source_overlay(
             mounts_dir=constants.MOUNTS_DIR,
             scratch_dir=constants.SCRATCH_DIR,
             storage_info=indata.storage_info,
-            xfs_info=indata.xfs_info) as overlay:
+            xfs_info=indata.xfs_info,
+            scratch_reserve=reserve_space) as overlay:
         with overlay.nspawn() as context:
             # Mount the ISO into the scratch container
             target_iso = next(api.consume(TargetOSInstallationImage), None)
--- a/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
+++ b/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
@@ -1,5 +1,6 @@
 import itertools
 import os
+import re
 import shutil
 
 from leapp import reporting
@@ -55,6 +56,7 @@ from leapp.utils.deprecation import suppress_deprecation
 PROD_CERTS_FOLDER = 'prod-certs'
 GPG_CERTS_FOLDER = 'rpm-gpg'
 PERSISTENT_PACKAGE_CACHE_DIR = '/var/lib/leapp/persistent_package_cache'
+DEDICATED_LEAPP_PART_URL = 'https://access.redhat.com/solutions/7011704'
 
 
 def _check_deprecated_rhsm_skip():
@@ -172,6 +174,53 @@ def _import_gpg_keys(context, install_root_dir, target_major_version):
         )
 
 
+def _handle_transaction_err_msg_size_old(err):
+    # NOTE(pstodulk): This is going to be removed in future!
+
+    article_section = 'Generic case'
+    xfs_info = next(api.consume(XFSPresence), XFSPresence())
+    if xfs_info.present and xfs_info.without_ftype:
+        article_section = 'XFS ftype=0 case'
+
+    message = ('There is not enough space on the file system hosting /var/lib/leapp directory '
+               'to extract the packages.')
+    details = {'hint': "Please follow the instructions in the '{}' section of the article at: "
+                       "link: https://access.redhat.com/solutions/5057391".format(article_section)}
+
+    raise StopActorExecutionError(message=message, details=details)
+
+
+def _handle_transaction_err_msg_size(err):
+    if get_env('LEAPP_OVL_LEGACY', '0') == '1':
+        _handle_transaction_err_msg_size_old(err)
+        return  # not needed actually as the above function raises error, but for visibility
+    NO_SPACE_STR = 'more space needed on the'
+
+    # Disk Requirements:
+    #   At least <size> more space needed on the <path> filesystem.
+    #
+    missing_space = [line.strip() for line in err.stderr.split('\n') if NO_SPACE_STR in line]
+    size_str = re.match(r'At least (.*) more space needed', missing_space[0]).group(1)
+    message = 'There is not enough space on the file system hosting /var/lib/leapp.'
+    hint = (
+        'Increase the free space on the filesystem hosting'
+        ' /var/lib/leapp by {} at minimum. It is suggested to provide'
+        ' reasonably more space to be able to perform all planned actions'
+        ' (e.g. when 200MB is missing, add 1700MB or more).\n\n'
+        'It is also a good practice to create dedicated partition'
+        ' for /var/lib/leapp when more space is needed, which can be'
+        ' dropped after the system upgrade is fully completed'
+        ' For more info, see: {}'
+        .format(size_str, DEDICATED_LEAPP_PART_URL)
+    )
+    # we do not want to confuse customers by the orig msg speaking about
+    # missing space on '/'. Skip the Disk Requirements section.
+    # The information is part of the hint.
+    details = {'hint': hint}
+
+    raise StopActorExecutionError(message=message, details=details)
+
+
 def prepare_target_userspace(context, userspace_dir, enabled_repos, packages):
     """
     Implement the creation of the target userspace.
@@ -210,21 +259,11 @@ def prepare_target_userspace(context, userspace_dir, enabled_repos, packages):
             message = 'Unable to install RHEL {} userspace packages.'.format(target_major_version)
             details = {'details': str(exc), 'stderr': exc.stderr}
 
-            xfs_info = next(api.consume(XFSPresence), XFSPresence())
             if 'more space needed on the' in exc.stderr:
                 # The stderr contains this error summary:
                 # Disk Requirements:
                 #   At least <size> more space needed on the <path> filesystem.
-
-                article_section = 'Generic case'
-                if xfs_info.present and xfs_info.without_ftype:
-                    article_section = 'XFS ftype=0 case'
-
-                message = ('There is not enough space on the file system hosting /var/lib/leapp directory '
-                           'to extract the packages.')
-                details = {'hint': "Please follow the instructions in the '{}' section of the article at: "
-                                   "link: https://access.redhat.com/solutions/5057391".format(article_section)}
-                raise StopActorExecutionError(message=message, details=details)
+                _handle_transaction_err_msg_size(exc)
 
             # If a proxy was set in dnf config, it should be the reason why dnf
             # failed since leapp does not support updates behind proxy yet.
--- a/repos/system_upgrade/common/libraries/overlaygen.py
+++ b/repos/system_upgrade/common/libraries/overlaygen.py
@@ -5,6 +5,7 @@ from collections import namedtuple
 
 from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.common import mounting, utils
+from leapp.libraries.common.config import get_env
 from leapp.libraries.stdlib import api, CalledProcessError, run
 
 OVERLAY_DO_NOT_MOUNT = ('tmpfs', 'devpts', 'sysfs', 'proc', 'cramfs', 'sysv', 'vfat')
@@ -150,7 +151,7 @@ def _overlay_disk_size_old():
     Convenient function to retrieve the overlay disk size
     """
     try:
-        env_size = os.getenv('LEAPP_OVL_SIZE', default='2048')
+        env_size = get_env('LEAPP_OVL_SIZE', '2048')
         disk_size = int(env_size)
     except ValueError:
         disk_size = 2048
--- a/repos/system_upgrade/common/libraries/overlaygen.py
+++ b/repos/system_upgrade/common/libraries/overlaygen.py
@@ -13,15 +13,6 @@ OVERLAY_DO_NOT_MOUNT = ('tmpfs', 'devpts', 'sysfs', 'proc', 'cramfs', 'sysv', 'v
 MountPoints = namedtuple('MountPoints', ['fs_file', 'fs_vfstype'])
 
 
-def _ensure_enough_diskimage_space(space_needed, directory):
-    stat = os.statvfs(directory)
-    if (stat.f_frsize * stat.f_bavail) < (space_needed * 1024 * 1024):
-        message = ('Not enough space available for creating required disk images in {directory}. ' +
-                   'Needed: {space_needed} MiB').format(space_needed=space_needed, directory=directory)
-        api.current_logger().error(message)
-        raise StopActorExecutionError(message)
-
-
 def _get_mountpoints(storage_info):
     mount_points = set()
     for entry in storage_info.fstab:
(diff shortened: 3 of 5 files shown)


repos/system_upgrade/common/libraries/overlaygen.py
```diff
@@ -42,39 +137,74 @@ def _mount_name(mountpoint):
...
+    of partition hosting /var/lib/leapp/* calculate the free space value
+    taking `scratch_reserve` into account, as during the run of the tooling,
+    we will be consuming the space on the partition and we want to be more
+    sure that we do not consume all the space on the partition during the
+    execution - so we reduce the risk we affect run of other applications
+    due to missing space.
+
+    Note: the partition hosting the scratch dir is expected to be the same
+    partition that is hosting the target userspace container, but it does not
+    have to be true if the code changes. Right now, let's live with that.
```
Rezney, line 198: Just out of curiosity, do we have any plans for the code changes?

repos/system_upgrade/common/libraries/overlaygen.py
```diff
@@ -42,39 +137,74 @@ def _mount_name(mountpoint):
...
+    :type scratch_reserve: Optional[int]
+    """
+    mount_points = _get_mountpoints(storage_info)
+    scratch_mp = _get_mountpoint(mount_points, scratch_dir)
     disk_images_directory = os.path.join(scratch_dir, 'diskimages')
 
     # Ensure we cleanup old disk images before we check for space constraints.
+    # NOTE(pstodulk): Could we improve the process so we create imgs & calculate
+    # the required disk space just once during each leapp (pre)upgrade run?
     run(['rm', '-rf', disk_images_directory])
```
Rezney, line 218: What about adding some exception handling here?

repos/system_upgrade/common/libraries/overlaygen.py
```diff
@@ -128,29 +250,98 @@ def cleanup_scratch(scratch_dir, mounts_dir):
...
-        cmd=['/bin/dd', 'if=/dev/zero', 'of={}'.format(diskimage_path), 'bs=1M', 'count={}'.format(disk_size)],
-        hint='Please ensure that there is enough diskspace in {} at least {} MiB are needed'.format(
-            diskimage_path, disk_size)
-    )
-
     api.current_logger().debug('Creating ext4 filesystem in disk image at %s', diskimage_path)
+    cmd = [
+        '/sbin/mkfs.ext4',
+        '-J', 'size=32',
+        '-E', 'lazy_itable_init=0,lazy_journal_init=0',
```
Rezney, line 303: Could you please elaborate on why these options were used?

repos/system_upgrade/common/libraries/overlaygen.py
```diff
@@ -5,19 +5,146 @@
...
+_MAGICAL_CONSTANT_OVL_SIZE = 128
+"""
+Average size of created disk space images.
+
+The size can be lower or higher - usually lower. The value is higher as we want
+to rather prevent future actions in advance instead of resolving later issues
+with the missing space.
+"""
+
+_MAGICAL_CONSTANT_MIN_CONTAINER_SIZE = 2200
```
Rezney, line 26: I understand that we probably do not want to get back to affecting the size numbers like in the "LEAPP_OVL_SIZE" but could not some way to affect the numbers help in some edge cases?

repos/system_upgrade/common/libraries/overlaygen.py
pirat89: Question is not whether yes or no. Important is to be aware that we have some functionality implemented in the way that if a change happens, this will be broken with the current implementation. No, we do not plan changes here. But who knows in years?

repos/system_upgrade/common/libraries/overlaygen.py
```diff
@@ -128,29 +284,98 @@ def cleanup_scratch(scratch_dir, mounts_dir):
...
+        '-J', 'size=32',
+        '-E', 'lazy_itable_init=0,lazy_journal_init=0',
+        '-F', diskimage_path
+    ]
     try:
-        utils.call_with_oserror_handled(cmd=['/sbin/mkfs.ext4', '-F', diskimage_path])
+        utils.call_with_oserror_handled(cmd=cmd)
     except CalledProcessError as e:
+        # FIXME(pstodulk): taken from original, but %s seems to me invalid here
         api.current_logger().error('Failed to create ext4 filesystem %s', exc_info=True)
```
MichalHe, line 310: Agree. That `%s` should not be there, or the `diskimage_path` should be passed to the logging call.

repos/system_upgrade/common/libraries/overlaygen.py
pirat89: Thinking about that and around since the morning. Will see

repos/system_upgrade/common/libraries/overlaygen.py
pirat89: It's explained in the docstring.

repos/system_upgrade/common/libraries/overlaygen.py
pirat89: From my checks, this seems really like the reasonable minimum. Better questions is whether we could allow people to set higher value on their own. I am thinking about that. But not sure how good the idea is. Imagine how to document that variable in the documentation for users without explanation what we are doing - as it would need a documentation :/

repos/system_upgrade/common/libraries/overlaygen.py
```diff
@@ -42,39 +171,74 @@ def _mount_name(mountpoint):
 def _mount_dir(mounts_dir, mountpoint):
     return os.path.join(mounts_dir, _mount_name(mountpoint))
 
+def _get_mountpoint(mount_points, dir_path):
+    for mp in sorted(mount_points, reverse=True):
+        mod_mp = mp if mp[-1] == '/' else '{}/'.format(mp)
+        if dir_path.startswith(mod_mp):
+            # longest first, so the first one we find, is the last mp on the path
+            return mp
```
MichalHe, line 179: If the mountpoint semantics of this function are the same as for the `df` command, then this function returns incorrect mountpoint if called as `_get_mountpoint(['/var/lib', '/var'], '/var/lib')`

MichalHe (commented):
Thinking about the problem at hand, I've come to the conclusion the overall approach taken in this PR is legitimate and logical. The only part that might be interesting to discuss is the heuristic - it makes me wonder how well predictable is the container size. It would be super exciting to come up with a model that would be able to predict target userspace size dynamically based on the source system.

repos/system_upgrade/common/libraries/overlaygen.py
MichalHe: Reads fine to me.

repos/system_upgrade/common/libraries/overlaygen.py
MichalHe: I would add env vars that would allow modifying this particular constant, but I would argue not to put it into documentation anywhere. If we gather empirical data that there are people that are hitting this problem, then we can put it into documentation and deal with the testing situation.

repos/system_upgrade/common/libraries/overlaygen.py
```diff
@@ -206,15 +431,59 @@ def _mount_dnf_cache(overlay_target):
...
+    This function prepares a disk image and an overlay layer for each
+    mountpoint configured in /etc/fstab, excluding those with FS type noted
+    in the OVERLAY_DO_NOT_MOUNT set. Such prepared OVL images are then composed
+    together to reflect the real host filesystem. In the end everything is cleaned.
+
+    The new solution can be now problematic for system with too many partitions
+    and loop devices. For such systems we keep for now the possibility of the
+    fallback to an old solution, which has however number of issues that are
+    fixed by the new design. To fallback to the old solution, set envar:
+        LEAPP_OVL_FALLBACK=1
```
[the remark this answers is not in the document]
pirat89: That's true. But we expect the use of that variable by user when they realize the new solution does not work. Just it is not executed automatically in case of problems. Any suggestion for the name of the envar? Adding some cadidates below:

* LEAPP_OVL_FALLBACK
* LEAPP_OVL_OLD_SOLUTION
* LEAPP_OVL_LEGACY
* ??

repos/system_upgrade/common/libraries/overlaygen.py
pirat89: Good catch! I will check it

pirat89: Ah. Now I've got the idea why i did it in this way. It was expected that dir_path will be always e.g. scratch_dir or something like that, which will have always mountpoint above. But I changed the name of the function and it's not clear from that now. So either the funcionality should be updated or the name/docstring to make it clear. Also `_get_mountpoint` / `_get_mountpoints` is a little bit not ideal as these functions are very close to each other. So I vote for the name change.

pirat89: @MichalHe  renamed to `_get_scratch_mountpoint`

repos/system_upgrade/common/libraries/overlaygen.py
```diff
@@ -128,29 +284,98 @@ def cleanup_scratch(scratch_dir, mounts_dir):
...
+    The created disk image is formatted with XFS (default) or Ext4 FS
+    and it's supposed to be used for write directories of an overlayfs built
+    above it.
+
+    The disk image is formatted with Ext4 if (envar) `LEAPP_OVL_IMG_FS_EXT4=1`.
+
+    :param disk_images_directory: Path to the directory where disk images should be stored.
+    :type disk_images_directory: str
+    :param path:
+    :type pyth: str
```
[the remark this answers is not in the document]
pirat89: fixed

repos/system_upgrade/common/libraries/overlaygen.py
pirat89: let's keep it for now as it is. We have similar action on multiple places.

pirat89:
So I've discovered that my previous information was wrong. It seems that the minimal container size is
* IPU 7 -> 8:  ~2.9GiB (+/- 100MiB) 
* IPU 8 -> 9: ~1.8GiB (+/- 100MiB)

additionally I've discovered confusing errrors when trying to install rpms inside the container to be able to generate initramfs as such operation could end with information about missing disk space on `/` as it's generate inside the container. So this needs to be covered also as we really do not want to another confusion about what directories need to be enlarged. Added to the list of missing items...

repos/system_upgrade/common/libraries/overlaygen.py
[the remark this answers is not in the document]
fernflower: I am with @MichalHe here - fallback implies automatic action taken upon another one failing.
`LEAPP_OVL_LEGACY` is the best one from the list imho.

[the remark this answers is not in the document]
matejmatuska: I think that with the context you provided both FALLBACK and LEGACY make sense.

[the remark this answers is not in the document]
MichalHe: I've spent some time with thesaurus looking for good-sounding words that would express the purpose of the env var, but I have found none. Let's go with `LEGACY ` as that is something already used to express something very similar in the field, e.g., GRUB2 vs GRUB legacy.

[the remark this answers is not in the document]
Rezney: +1 to LEAPP_OVL_LEGACY

repos/system_upgrade/common/libraries/overlaygen.py
```diff
@@ -5,19 +5,150 @@
...
 MountPoints = namedtuple('MountPoints', ['fs_file', 'fs_vfstype'])
 
 
+def get_recommended_leapp_free_space(userspace_path=None):
+    """
+    Return recommended free space for the target container (+ pkg downloads)
+
+    If the path to the container is set, the returned value is updated to
+    reflect already consumed space by the installed container. In case the
+    container is bigger than the minimal protected size, always
```
PeterMocary, line 61: Unfinished sentence.

repos/system_upgrade/common/libraries/overlaygen.py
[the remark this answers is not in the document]
PeterMocary: LEAPP_OVL_LEGACY sounds the best to me also.

repos/system_upgrade/common/libraries/overlaygen.py
```diff
@@ -228,3 +505,119 @@ def create_source_overlay(mounts_dir, scratch_dir, xfs_info, storage_info, mount
...
         cleanup_scratch(scratch_dir, mounts_dir)
         raise
+
+
+# #############################################################################
+# Deprecated OVL solution ...
+# This is going to be removed in future as the whole functionality is going to
+# be replaced by new one. The problem is that the new solution can potentially
+# negatively affect systems with many loop mountpoints, so let's keep this
+# as a workaround for now. I am separating the old and new code in this way
```
PeterMocary, line 515: I'd just add that it is an optional workaround triggered by the LEAPP_OVL_LEGACY env. variable. Just for the comment to contain the complete information if it stays there longer.

repos/system_upgrade/common/libraries/overlaygen.py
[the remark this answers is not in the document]
pirat89: So LEAPP_OVL_LEGACY be it! thanks!

repos/system_upgrade/common/libraries/overlaygen.py
pirat89: the size has been changed to 3200 for now. most likely it will be lowered at least for IPU 8 -> 9. Just for now, keepin like that to stay safe

repos/system_upgrade/common/libraries/overlaygen.py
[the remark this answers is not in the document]
pirat89: done

pirat89:
Current space consumption and checks of the free space tested several times close to what I call close to minimal reasonable free space on /var - ending with 945MB of the free space for the upgrade - which is nearly boundary of the minimal required free space to be able to perform the upgrade transaction yet.

pirat89:
The estimated required disk space is optimized for IPU 8 -> 9 (reduced by 1GiB - based on experiments).

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -168,6 +173,53 @@ def _import_gpg_keys(context, install_root_dir, target_major_version):
...
+        ' dropped after the system upgrade is fully completed'
+        ' For more info, see: {}'
+        .format(size_str, DEDICATED_URL)
+    )
+    # we do not want to confuse customers by the orig msg speaking about
+    # missing space on '/'. Skip the Disk Requirements section.
+    # The information is part of the hint.
+    details = {'hint': hint}
+
+    raise StopActorExecutionError(message=message, details=details)
```
pirat89, line 220: I am not sure whether this is actually correct. I am afraid that I maybe missed a corner case here. Imagine this partitioning:
```
# mount point: free space
/: 10 GB
/var: 700MB (e.g.)
/var/lib/leapp: 10 GB
```
In this case, the disk img for /var is located under the leapp dir, but it's apparent size will be only ~ 665MB. Currently we need to download ~160M data (metadata + packages). And some space is consumed running dnf/rpm. This all is stored inside the OVL layer.  Imagine we need to download more pkgs (/var cannot have less than 500M otherwise it crashes during the scan phase when we call yum :)), in this only one situation, I think we will have problem with the msg, as it will be actually affected by the size of /var and not /var/lib/leapp. But who would actually try to upgrade when having only such small space on /var ??? So it's from my POV really total corner case that should not happen. Also it needs to be differenciated, whether the problem is caused during pkgs download or calculation whether there is enough disk space for the installation of the downloaded packages (both cases has different answer where the space is missing).

Opposite problem could appear if /var/lib/leapp is not big enough and we consume for the pkg download all remaining free space. But regarding the implemented checks, I do not believe this situation will happen (it would mean that someone would like to download couple of GBs pkgs to install super big container right now. and that does not seem meaningful.

repos/system_upgrade/common/libraries/overlaygen.py
pirat89: covered. thanks!

repos/system_upgrade/common/libraries/overlaygen.py
pirat89: covered

repos/system_upgrade/common/libraries/overlaygen.py
pirat89: used different values for each RHEL.

repos/system_upgrade/common/libraries/overlaygen.py
```diff
@@ -5,19 +5,112 @@
...
-                   'Needed: {space_needed} MiB').format(space_needed=space_needed, directory=directory)
+    if _get_fspace(directory) < (space_needed * 1024 * 1024):
+        message = (
+            'Not enough space available for creating required disk images in {directory}. '
+            'Needed at least: {space_needed} MiB.'
+            .format(space_needed=space_needed, directory=directory)
+        )
+        if get_env('LEAPP_OVL_SIZE', None):
+            # LEAPP_OVL_SIZE has not effect as we use sparse files now.
+            messages = '{} Note that LEAPP_OVL_SIZE has no effect anymore.'.format(messages)
```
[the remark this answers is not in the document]
pirat89: changed. the note is part of details now.

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
pirat89: @oamg/developers  ^ ping. do not miss this comment

MichalHe: The only scenario that I can image this to happen is that someone has small /var. The first leapp execution crashes, lacking required disk space. The user then deals with this problem by creating a separate /var/lib/leapp partition without increasing the size of /var. The wording of this error message should not create the impression that a separate /var/lib/leapp will not deal with the case when the /var partition is too small.

I've also checked the internet for recommended partitioning. [RHEL8 partitioning guide](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/performing_an_advanced_rhel_8_installation/partitioning-reference_installing-rhel-as-an-experienced-user) recommends that `5gb` is enough for `/var`. [RHEL7 guide](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/installation_guide/sect-disk-partitioning-setup-x86#sect-recommended-partitioning-scheme-x86) does not mention a precise number, but says it should be "sufficiently large", whatever that means. Other advice on the internet states that `/var` should be at least 2gb in size.

I've also went trough vanilla RHEL images on Azure and they all have `/var` 5gb+ in size (I remember encountering some issues around /var space there before).

Therefore, I do not think that this corner case will be encountered.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
```diff
@@ -149,13 +150,84 @@ def _update_files(copy_files):
     _copy_files(context, files)
 
 
+def _get_fspace(path, convert_to_mibs=False, coefficient=1):
```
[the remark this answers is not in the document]
MichalHe: Also the `convert_to_mibs` argument is kindof pointless, or? The only time the function is called without convert_to_mibs=True is when the other number that is  being compared against the free space is converted from mbs to bytes.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
```diff
@@ -149,13 +150,84 @@ def _update_files(copy_files):
...
+    exceptions when it happens. From this point, let's rather check the available
+    space in advance and stop the execution when it happens.
+
+    It is not expected to hit this issue, but I was successful and I know
+    it's still possible even with all other changes (just it's much harder
+    now to hit it). So adding this seatbelt, that is not 100% bulletproof,
+    but I call it good enough.
+
+    Currently protecting last 500MB. In case of problems, we can increase
+    the value.
```
MichalHe, line 199: I like saving additional context of the code in the comments, but I am not sure how to feel about putting it in documentation strings.

MichalHe (commented):
LGTM. I will approve when I try it out manually.

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -168,6 +174,53 @@ def _import_gpg_keys(context, install_root_dir, target_major_version):
...
+    missing_space = [line.strip() for line in err.stderr.split('\n') if NO_SPACE_STR in line]
+    size_str = re.match(r'At least (.*) more space needed', missing_space[0]).group(1)
+    message = 'There is not enough space on the file system hosting /var/lib/leapp.'
+    hint = (
+        'Increase the free space on the filesystem hosting'
+        ' /var/lib/leapp by {} at minimum. It is suggested to provide'
+        ' reasonably more space to be able to perform all planned actions'
+        ' (e.g. when 200MB is missing, add 1700MB or more).\n\n'
+        'It is also a good practice to create dedicated partition for'
+        ' for /var/lib/leapp when more space is needed, which can be'
```
matejmatuska, line 211:
```suggestion
        'It is also a good practice to create dedicated partition'
        ' for /var/lib/leapp when more space is needed, which can be'
```
There are two fors.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
```diff
@@ -19,6 +19,7 @@
 
 INITRAM_GEN_SCRIPT_NAME = 'generate-initram.sh'
 DRACUT_DIR = '/dracut'
+DEDICATED_URL = 'https://access.redhat.com/solutions/7011704'
```
matejmatuska, line 22: The name `DEDICATED_URL` is not very helpful, I think something like `DEDICATED_LEAPP_PART_URL` would be better.

The same goes for `userspacegen.py`

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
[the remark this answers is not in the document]
matejmatuska: Agree on this.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
```diff
@@ -149,13 +150,84 @@ def _update_files(copy_files):
...
+
+    Currently protecting last 500MB. In case of problems, we can increase
+    the value.
+    """
+    message = 'There is not enough space on the file system hosting /var/lib/leapp.'
+    hint = (
+        'Increase the free space on the filesystem hosting'
+        ' /var/lib/leapp by 500MB at minimum (suggested 1500MB).\n\n'
+        'It is also a good practice to create dedicated partition for'
+        ' for /var/lib/leapp when more space is needed, which can be'
```
matejmatuska, line 206:
```suggestion
        'It is also a good practice to create dedicated partition'
        ' for /var/lib/leapp when more space is needed, which can be'
```
Same as in `overlaygen.py`.

MichalHe:
I've performed a 7>8 upgrade using internal vagrant box, everything went fine.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
pirat89: good catch! it's not 3 places actually. all of them are fixed now

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
pirat89: updated in both files.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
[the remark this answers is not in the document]
pirat89: in this case, let's rather add TODO and discuss it later - but before the upstream release

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
pirat89: good point. let's discuss it later - but before the release. we can still change it before the mid Aug.

pirat89:
@MichalHe @Rezney  @matejmatuska  thanks you guys for the review. let's merge it now and resolve the missing stuff in the following period.

pirat89:
Merging regarding the time pressure and bunch of manual tests that have been done. Unless we speak about corner cases and additional possible optimisations, it should be most working (shame on me I haven't delivered this earlier for the proper testing prior the merge)

PR #1100, opened 2023-07-13
Experimental (unsupported) enable of upgrades for encrypted devices

pirat89:
Upgrades with encrypted devices / partitions are not supported now. Based on the feedback from @Naranthiran, it is possible to upgrade with encrypted devices (using TPM) with the following changes.

We have no idea in what all situations this will work correctly, nor detection of used technologies is implemented to say whether the decryption of encrypted devices will happen automatically correctly or not, etc. From this point, allowing the upgrade in unsupported mode when following devel envar is set:
```
    LEAPP_DEVEL_ALLOW_DISK_ENCRYPTION=1
```

For more details, see:
  https://github.com/oamg/leapp-repository/discussions/1098

Authored / Co-authored-by @Naranthiran

### DIFF
--- a/repos/system_upgrade/common/actors/inhibitwhenluks/actor.py
+++ b/repos/system_upgrade/common/actors/inhibitwhenluks/actor.py
@@ -1,8 +1,9 @@
 from leapp import reporting
 from leapp.actors import Actor
+from leapp.libraries.common.config import get_env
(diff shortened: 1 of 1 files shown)


pirat89:
Closing in favor of #1200

PR #1104, opened 2023-08-01
makefile: add dev_test_no_lint target

MichalHe:
Add a target for testing individual actors with almost-instant
execution time. Testing individual actors currently involves
a process in which every actor is instantiated in a separate
process, the created instance reports actor information such as actor's
name and then exits. As many processes are created, this process is
time consuming (cca 7s) which disrupts developer's workflow and causes
attention shift.

A newly added target `dev_test_no_lint` uses an introduced script
`find_actors`. To achieve the similar level of framework protection
as spawning actors in a separate process, the `find_actors` script
does not execute actors at all, and instead works on their ASTs.
Specifically, the script looks for all files named `actor.py`, finds
all classes that (explicitely) subclass Actor, and reads its `name`
attribute.

Usage example:
```
    ACTOR=check_target_iso make dev_test_no_lint
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,9 +16,12 @@ REPOSITORIES ?= $(shell ls $(_SYSUPG_REPOS) | xargs echo | tr " " ",")
 SYSUPG_TEST_PATHS=$(shell echo $(REPOSITORIES) | sed -r "s|(,\\|^)| $(_SYSUPG_REPOS)/|g")
 TEST_PATHS:=commands repos/common $(SYSUPG_TEST_PATHS)
 
+# python version to run test with
+_PYTHON_VENV=$${PYTHON_VENV:-python2.7}
 
 ifdef ACTOR
-	TEST_PATHS=`python utils/actor_path.py $(ACTOR)`
+	TEST_PATHS=`$(_PYTHON_VENV) utils/actor_path.py $(ACTOR)`
+	APPROX_TEST_PATHS=$(shell $(_PYTHON_VENV) utils/find_actors.py -C repos $(ACTOR))  # Dev only
 endif
 
 ifeq ($(TEST_LIBS),y)
@@ -31,9 +34,6 @@ endif
 
 # needed only in case the Python2 should be used
 _USE_PYTHON_INTERPRETER=$${_PYTHON_INTERPRETER}
-
-# python version to run test with
-_PYTHON_VENV=$${PYTHON_VENV:-python2.7}
 
 # by default use values you can see below, but in case the COPR_* var is defined
 # use it instead of the default
@@ -127,6 +127,7 @@ help:
 	@echo "                              - can be changed by setting TEST_CONTAINER env"
 	@echo "  test_container_all          run lint and tests in all available containers"
 	@echo "  test_container_no_lint      run tests without linting in container, see test_container"
+	@echo "  dev_test_no_lint            (advanced users) run only tests of a single actor specified by the ACTOR variable"
 	@echo "  test_container_all_no_lint  run tests without linting in all available containers"
 	@echo "  clean_containers            clean all testing and building container images (to force a rebuild for example)"
 	@echo ""
@@ -486,6 +487,10 @@ fast_lint:
 		echo "No files to lint."; \
 	fi
 
+dev_test_no_lint:
+	. $(VENVNAME)/bin/activate; \
+	$(_PYTHON_VENV) -m pytest $(REPORT_ARG) $(APPROX_TEST_PATHS) $(LIBRARY_PATH)
+
 dashboard_data:
 	. $(VENVNAME)/bin/activate; \
 	snactor repo find --path repos/; \
@@ -494,4 +499,4 @@ dashboard_data:
 	popd
 
 .PHONY: help build clean prepare source srpm copr_build _build_local build_container print_release register install-deps install-deps-fedora  lint test_no_lint test dashboard_data fast_lint
-.PHONY: test_container test_container_no_lint test_container_all test_container_all_no_lint clean_containers _build_container_image _test_container_ipu
+.PHONY: test_container test_container_no_lint test_container_all test_container_all_no_lint clean_containers _build_container_image _test_container_ipu dev_test_no_lint
--- a/utils/find_actors.py
+++ b/utils/find_actors.py
@@ -1,0 +1,81 @@
+import argparse
+import ast
+import os
+import sys
+
+
+def is_direct_actor_def(ast_node):
+    if not isinstance(ast_node, ast.ClassDef):
+        return False
+
+    direcly_named_bases = (base for base in ast_node.bases if isinstance(base, ast.Name))
+    for class_base in direcly_named_bases:
+        # We are looking for direct name 'Actor'
+        if class_base.id == 'Actor':
+            return True
+
+    return False
+
+
+def extract_actor_name_from_def(actor_class_def):
+    assignment_value_class = ast.Str if sys.version_info < (3,8) else ast.Constant
+    assignment_value_attrib = 's' if sys.version_info < (3,8) else 'value'
+
+    actor_name = None
+    class_level_assignments = (child for child in actor_class_def.body if isinstance(child, ast.Assign))
+    # Search for class-level assignment specifying actor's name: `name = 'name'`
+    for child in class_level_assignments:
+        assignment = child
+        for target in assignment.targets:
+            assignment_adds_name_attrib = isinstance(target, ast.Name) and target.id == 'name'
+            assignment_uses_a_constant_string = isinstance(assignment.value, assignment_value_class)
+            if assignment_adds_name_attrib and assignment_uses_a_constant_string:
+                rhs = assignment.value  # <lhs> = <rhs>
+                actor_name = getattr(rhs, assignment_value_attrib)
+                break
+        if actor_name is not None:
+            break
+    return actor_name
+
+
+def get_actor_names(actor_path):
+    with open(actor_path) as actor_file:
+        try:
+            actor_def = ast.parse(actor_file.read())
+        except SyntaxError:
+            error = ('Failed to parse {0}. The actor might contain syntax errors, or perhaps it '
+                     'is written with Python3-specific syntax?\n')
+            sys.stderr.write(error.format(actor_path))
+            return []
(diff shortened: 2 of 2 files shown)


MichalHe:
@oamg/developers this can be imho merged (I've used it while developing rhui simplification), but please try it couple of times to be sure that there is not some obvious problem.

MichalHe:
@oamg/developers this can be imho merged (I've used it while developing rhui simplification), but please try it couple of times to be sure that there is not some obvious problem.

fernflower:
@MichalHe Could you please rebase against master? This should fix packit tests

MichalHe:
@fernflower done.

utils/find_actors.py
```diff
@@ -0,0 +1,72 @@
...
+    return has_actor_base
+
+
+def extract_actor_name_from_def(actor_class_def):
+    actor_name = None
+    for child in actor_class_def.body:
+        if not isinstance(child, ast.Assign):  # We are looking for `name = 'name'`
+            continue
+        assignment = child
+        for target in assignment.targets:
```
fernflower, line 29: I would probably kill `lhs` and `assignment` variables and just use `child.value.value` and `child` instead

utils/find_actors.py
```diff
@@ -0,0 +1,72 @@
...
+    with open(actor_path) as actor_file:
+        actor_def = ast.parse(actor_file.read())
+        actor_defs = [ast_node for ast_node in actor_def.body if is_direct_actor_def(ast_node)]
+        actors = [extract_actor_name_from_def(actor_def) for actor_def in actor_defs]
+    return actors
+
+
+def make_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('actor_names', nargs='+', help='Actor names to look for')
```
fernflower, line 49: Could you please add a reference, that actor name means the `name` attribute of an actor object?

fernflower (commented):
I love the `find_actors` module, it works like a charm as a standalone module, thanks!
Could you please add a comment about how it should be supposed to run via `make`? However hard I tried I could not figure out how to run dev tests for specific actors only..

utils/find_actors.py
```diff
@@ -0,0 +1,72 @@
...
+        if class_base.id == 'Actor':
+            has_actor_base = True
+            break
+
+    return has_actor_base
+
+
+def extract_actor_name_from_def(actor_class_def):
+    actor_name = None
+    for child in actor_class_def.body:
```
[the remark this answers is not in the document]
MichalHe: Honestly, if I see more than one `for` in an expression it becomes hard for me to read it, maybe because I am not not used to seeing these in the wild. But I went ahead and I pulled the 'filtering for assignments' into a named generator so the loop body is simpler while not making maintainers to read nested inline for loops.

utils/find_actors.py
MichalHe: Hmm, the problem is that there is a typo and it should be `rhs` and not `lhs` :D

I would like to definitely keep the `rhs` as it provides you with more information about what is actually assigned as actor_name (it is the value of the right-hand side of the assignment).

utils/find_actors.py
MichalHe: Okayo, done.

Makefile
```diff
@@ -486,6 +487,10 @@ fast_lint:
 		echo "No files to lint."; \
 	fi
 
+dev_test_no_lint:
+	. $(VENVNAME)/bin/activate; \
+	$(_PYTHON_VENV) -m pytest $(REPORT_ARG) $(APPROX_TEST_PATHS) $(LIBRARY_PATH)
+
```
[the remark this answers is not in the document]
MichalHe: Great point! Done.

utils/find_actors.py
```diff
@@ -0,0 +1,72 @@
+import argparse
+import ast
+import os
+
+
+def is_direct_actor_def(ast_node):
+    if not isinstance(ast_node, ast.ClassDef):
+        return False
+
+    has_actor_base = False
```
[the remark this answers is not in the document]
MichalHe: Thanks. I've applied your suggestion.

MichalHe:
@fernflower Thank you for your review. I've applied your suggestions and written a proper commit message that includes instructions on how to execute the new target.

fernflower:
Hmm maybe it's my setup but I am getting a strange Permission Denied when specifying `ACTOR` env var
```
[ivasilev@localhost leapp-repository]$ ACTOR=check_target_iso make dev_test_no_lint
/bin/bash: line 1: utils/find_actors.py: Permission denied
. tut/bin/activate; \
```
After this I also see all unit tests being executed and not only those related to the actor.
Could you please confirm that is not the case for you, so I could start deep dive debugging of my setup?

MichalHe:
Hmm, I believe it should be OK now - I've used `PYTHON_VENV` instead of `_PYTHON_VENV` and I did not encounter any problems because I always define `PYTHON_VENV`. @fernflower can you try now?

fernflower:
Hmm I still got all unit tests executed
```
================================================ 15 failed, 2019 passed, 19 skipped, 3 warnings in 66.26 seconds =================================================
make: *** [Makefile:491: dev_test_no_lint] Error 1
```

matejmatuska:
> Hmm I still got all unit tests executed for `ACTOR=check_target_iso make dev_test_no_lint `
> 
> ```
> ================================================ 15 failed, 2019 passed, 19 skipped, 3 warnings in 66.26 seconds =================================================
> make: *** [Makefile:491: dev_test_no_lint] Error 1
> ```
> 
> At the same time when I define `PYTHON_VENV`, I get the expected result `PYTHON_VENV=python3.9 ACTOR=check_target_iso make dev_test_no_lint`
> 
> ```
> (only check_target_iso tests)
> =================================================================== 16 passed in 0.17 seconds ====================================================================
> ```

I have the same results, when I `unset PYTHON_VENV` it tries to execute all t

matejmatuska:
I am sorry I missclicked and accidentally closed this, I reopened it immediately.

Makefile
```diff
@@ -18,7 +18,8 @@ TEST_PATHS:=commands repos/common $(SYSUPG_TEST_PATHS)
 
 
 ifdef ACTOR
-	TEST_PATHS=`python utils/actor_path.py $(ACTOR)`
+	TEST_PATHS=`$(_PYTHON_VENV) utils/actor_path.py $(ACTOR)`
+	APPROX_TEST_PATHS=$(shell $(_PYTHON_VENV) utils/find_actors.py -C repos $(ACTOR))  # Dev only
```
matejmatuska, line 22: The `_PYTHON_VENV` var is used before it is defined, definition is on line 37. However after I moved the definition it still tries to run all the tests if PYTHON_VENV isn't set.

matejmatuska (changes requested):

Makefile
matejmatuska: I've done some further tests and for me on python2.7 and python3.6 it attempts to run all the tests, at the beginning of the output I have:
```
. tut/bin/activate; \
${PYTHON_VENV:-python2.7} -m pytest
===================================== test session starts =====================================
platform linux -- Python 3.6.15, pytest-4.6.11, py-1.11.0, pluggy-0.13.1 -- /home/matej/projects/leapp-repo/tut/bin/python3.6
``` 
whereas on python3.9, which works as expected I have:
```
. tut/bin/activate; \
${PYTHON_VENV:-python2.7} -m pytest  /home/matej/projects/leapp-repo/repos/system_upgrade/common/actors/checktargetiso
===================================== test session starts =====================================
platform linux -- Python 3.6.15, pytest-4.6.11, py-1.11.0, pluggy-0.13.1 -- /home/matej/projects/leapp-repo/tut/bin/python3.9
```

Is it possible that the lookup script is failing on 2.7 and 3.6?

MichalHe: You are absolutely right about the definition of `_PYTHON_VENV`. Moreover, your observations about python versions led me to realize that there are some breaking changes in the `ast` since python3.8. Nevertheless, I have modified the script to work with all python versions:
```
(tut) [root@13b46a877de5 leapp-repository]# for py in python2.7 python3.6 python3.11; do ${py} utils/find_actors.py checkrhui -C repos/ ; done
Failed to parse /work/leapp-repository/repos/system_upgrade/el8toel9/actors/dotnet/actor.py. The actor might contain syntax errors, or perhaps it is written with Python3-specific syntax?
/work/leapp-repository/repos/system_upgrade/common/actors/cloud/checkrhui
/work/leapp-repository/repos/system_upgrade/common/actors/cloud/checkrhui
/work/leapp-repository/repos/system_upgrade/common/actors/cloud/checkrhui
```
The dotnet actor is using f-strings, which are python3 specific, and therefore, the is no surprise that the ast module fails to parse it. However, the original `test_no_lint` would fail as well if used with python2.7. I believe that the target should work now, can you please try @matejmatuska @fernflower?

matejmatuska: I can confirm that it works correctly on python 2.7, 3.6, 3.9 now.

matejmatuska (changes requested):
Can you please add an entry to `make help` on how to use this?

Makefile
fernflower: Yep, works like a charm!

fernflower:
Could you please also rebased against recent master? Tests should go green.

MichalHe:
@matejmatuska I will add a documentation help entry, but I will not add examples. Although I believe that the code covers majority of leapp use cases, it is  fragile, e.g. not using the Actor base explicitly would cause the actor to not be found. The target is intended for leapp developers who have a deeper understanding of the codebase and can tell where a problem might be. 
```
MyActorAlias = Actor
class MyActor(MyActorAlias):
    pass
```

fernflower (approved):
LGTM, thank you!

matejmatuska:
> @matejmatuska I will add a documentation help entry, but I will not add examples. Although I believe that the code covers majority of leapp use cases, it is fragile, e.g. not using the Actor base explicitly would cause the actor to not be found. The target is intended for leapp developers who have a deeper understanding of the codebase and can tell where a problem might be.
> 
> ```
> from leapp.actors import Actor
> MyActorAlias = Actor
> class MyActor(MyActorAlias):
>     pass
> ```

Okay that's reasonable.

matejmatuska (approved):

fernflower (approved):

PR #1106, opened 2023-08-02
Add support for upgrade with custom https repos

dkubek:
Currently, upgrades with custom repositories hosted on private https servers are not functional. The reason is that the server certificates are unavailable in the upgrade container as they are not copied from the source system; therefore, the upgrade fails when trying to access the custom repository.

This change fixes the issue by copying all files from the ``/etc/pki`` folder of the source system, while respecting any new files installed by packages into the container. If a conflict arises a file installed by a package into a container is preferred. Any broken symlinks are skipped and ignored.

Jira (internal): [OAMG-6388](https://issues.redhat.com/browse/OAMG-6388)
BZ (internal): [RHBZ: 2040706](https://bugzilla.redhat.com/show_bug.cgi?id=2040706)

### DIFF
--- a/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
+++ b/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
@@ -303,7 +303,7 @@ def _get_all_rhui_pkgs():
     return pkgs
 
 
-def _get_files_owned_by_rpms(context, dirpath, pkgs=None):
+def _get_files_owned_by_rpms(context, dirpath, pkgs=None, recursive=False):
     """
     Return the list of file names inside dirpath owned by RPMs.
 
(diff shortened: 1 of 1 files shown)


fernflower:
@dkubek if you rebase against master you should get meaningful packit test run (without Pipeline timeout exceeded)

matejmatuska:
## Tests
### RHUI
Since the patch touches RHUI code, I reproduced the issue when the `_copy_certificates` function is unconditional and I verified that with this patch on RHUI on Azure 8->9 IPU it's working as expected, the RHUI certs are no longer rewritten.

### General
I also did a test in the internal vagrant box with IPU 7->8 and verified that:
- new content from target userspace packages, such as `/etc/pki/product-default/479.pem` is there.
- if there is a file that is provided by both source and target system packages, such as RPM GPG keys in `/etc/pki/rpm-gpg/`, the file from target system package is prefered.
- the files from source system that aren't in the target system packages, such as custom certificates, are also present in the target userspace. 

### Integration
The patch also passes all the testcases in the yet unmerged integration tests (see Jira task for those).

matejmatuska:
> ## Tests
> ### RHUI
> Since the patch touches RHUI code, I reproduced the issue when the `_copy_certificates` function is unconditional and I verified that with this patch on RHUI on Azure 8->9 IPU it's working as expected, the RHUI certs are no longer rewritten.
> 
> ### General
> I also did a test in the internal vagrant box with IPU 7->8 and verified that:
> 
> * new content from target userspace packages, such as `/etc/pki/product-default/479.pem` is there.
> * if there is a file that is provided by both source and target system packages, such as RPM GPG keys in `/etc/pki/rpm-gpg/`, the file from target system package is prefered.
> * the files from source system that aren't in the target system packages, such as custom certificates, are also present in the target userspace.
> 
> ### Integration
> The patch also passes all the testcases in the yet unmerged integration tests (see Jira task for those).

I reran the prepared test with the lastest patch and all cases passed.

matejmatuska (approved):
LGTM

PR #1109, opened 2023-08-09
Removing the config state for the data files at /etc/leapp/files/*.json

PicklesMcDougan:
With the config option applied, these files are not replaced when packages are reinstalled.

### DIFF
--- a/packaging/leapp-repository.spec
+++ b/packaging/leapp-repository.spec
@@ -262,7 +262,7 @@ done;
 %dir %{repositorydir}
 %dir %{custom_repositorydir}
 %dir %{leapp_python_sitelib}/leapp/cli/commands
-%config %{_sysconfdir}/leapp/files/*
+%{_sysconfdir}/leapp/files/*
 %{_sysconfdir}/leapp/repos.d/*
 %{_sysconfdir}/leapp/transaction/*
 %{repositorydir}/*


pirat89:
Hi @PicklesMcDougan  \o, thank you for the contribution, but we want these files to be tracked as configuration files. This is intended, not a mistake. Some users even want/have to update them manually if they have own very specific setups, e.g.
 * when they change official repository IDs on the satellite site and require their own repoids instead of official ones
 * if they want to fix something without waiting for an official update of data
 * if they want to automatize use of some repositories during the upgrade that are currently not officially covered
...

Files are automatically updated when newer package is installed, if we have changed the data files in the new package (typical scenario..). If original files have been modified before new package is installed, they are stored in such a case with .rpmsave suffix. So customer can easily check new changes or they can apply just their previous changes back if they want. If we make them data files, it will mean that any customisation is dropped without a backup during update of the package.

We had some discussions about that in the past, where we wanted to make them datafiles strictly and we agreed to make them configuration files. (no, we do no want to make them data files under the /etc/leapp/files/ directory - if we had done such a step, we would have also changed the path where they are stored.)

Current behaviour is the expected one and we do not want to change it. From our POV it's standard and in case customer change them by their mistake, they can still remove them and reinstall the package to recover them.

PR #1113, opened 2023-08-15
grub2mkconfigonppc64: Do not touch grubenv when generating new grub2 config

pirat89:
When we call grub2-mkconfig for upgrades to RHEL 9.3+, we need to add also the --no-grubenv-update option also so we do not overwrite already updated grubenv file again. This lead to the situation that the crashkernel kernel cmdline option could contain invalid arguments originally used on RHEL 8.

* The comand is called only on virtualized Power machines.
* The use of the option is ok also on RHEL 9.2 (most likely also on RHEL 9.0, as it is present there too)

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=2231343
Related bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=2221543


pirat89:
postponing for the spring release - we've got feedback the full impact of this change is not clear

pirat89:
closed. the bug has been fixed in the GRUB2 bootloader.

PR #1115, opened 2023-08-16
updategrubcore: Fix both failure and success always reported

matejmatuska:
Both the "Grub core update failed" and "Grub core successfully updated" reports were printed unconditionally.

This introduces the missing conditions that check if there are any failed/successful updates.

### DIFF
--- a/repos/system_upgrade/common/actors/updategrubcore/libraries/updategrubcore.py
+++ b/repos/system_upgrade/common/actors/updategrubcore/libraries/updategrubcore.py
@@ -25,19 +25,29 @@ def update_grub_core(grub_devs):
 
         successful.append(dev)
 
-    reporting.create_report([
-        reporting.Title('GRUB core update failed'),
-        reporting.Summary('Leapp failed to update GRUB on {}'.format(', '.join(failed))),
-        reporting.Groups([reporting.Groups.BOOT]),
-        reporting.Severity(reporting.Severity.HIGH),
-        reporting.Remediation(
-            hint='Please run "grub2-install <GRUB_DEVICE>" manually after upgrade'
-        )
-    ])
+    if failed:
+        if successful:
+            # partial failure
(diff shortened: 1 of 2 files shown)


matejmatuska:
## TEST
I tested the case where just some, but not all, of the grub devices fail to update with a hacked in device, because this scenario is difficult to reproduce on the system.

The system only has `/dev/vda`:
```
[root@leapp-20230816145245 ~]# lsblk
NAME                                      MAJ:MIN RM SIZE RO TYPE MOUNTPOINT
vda                                       252:0    0  40G  0 disk
├─vda1                                    252:1    0   1G  0 part /boot
└─vda2                                    252:2    0  39G  0 part
  ├─rhel_ibm--p8--kvm--03--guest--02-root 253:0    0  38G  0 lvm  /
  └─rhel_ibm--p8--kvm--03--guest--02-swap 253:1    0   1G  0 lvm  [SWAP]
```

I patched the `/usr/share/leapp-repository/repositories/system_upgrade/common/actors/updategrubcore/actor.py` as follows:
```
def process(self):
    ff = next(self.consume(FirmwareFacts), None)
    if ff and ff.firmware == 'bios':
        grub_devs = grub.get_grub_devices()
        if grub_devs:
            grub_devs.append('/dev/vdb') # <-------------------------- PATCHED THIS IN
            update_grub_core(grub_devs)
        else:
            api.current_logger().warning('Leapp could not detect GRUB devices')
```

The failure report after the upgrade: 
```
Risk Factor: high
Title: GRUB core update failed
Summary: GRUB was successfully updated on the following devices: /dev/vda,
however GRUB update failed on the following devices: /dev/vdb
Remediation: [hint] Please run "grub2-install <GRUB_DEVICE>" manually after upgrade
Key: 54ba8c51e17b2088df653f3df0f1ae490f9ead88
```

PR #1116, opened 2023-08-16
Use new leapp output apis

matejmatuska:
Leapp introduced new output apis (and removed some other).
See https://github.com/oamg/leapp/pull/840 for details.

The calls to `generate_report_file()` in `commands/upgrade/utils.py` are
swapped to workaround a bug in the framework where the json report
generation modifies messages passed to it.

Jira: OAMG-9663


matejmatuska:
Unit test are failing because of the change required in framework

dkubek (dismissed):
lgtm

pirat89 (approved):
tested manually, everything works as expected so far, see my comment in https://github.com/oamg/leapp/pull/840 for more details

PR #1121, opened 2023-08-23
load data files: update hints in error messages and prepare for the download drop

pirat89:
As the leapp upgrade data files are nowadays part of the install rpm, there is no need to download them anymore. Also, we plan to drop the service providing the data files online in future.

For that reason, update all texts and related error messages so people are not instructed to visit obsoleted article and do not try to apply invalid (obsoleted) data files anymore.

Right now, we are keeping the functionality for the data files download, but the fetch functino is already updated and prepared to stop trying to download the files if not present. As we have the functionality already present, I think we should keep a possibility of the download for additional custom data files (not provided by us) in custom actors, but for the official data files we will require them in future to be present locally only.

NOTE: regarding another planned changes soon, skipping update of unit-tests

### DIFF
--- a/repos/system_upgrade/common/libraries/fetch.py
+++ b/repos/system_upgrade/common/libraries/fetch.py
@@ -7,6 +7,7 @@ import requests
 from leapp import models
 from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.common.config import get_consumed_data_stream_id, get_env
+from leapp.libraries.common.config.version import get_source_major_version, get_target_major_version
 from leapp.libraries.stdlib import api
 
 SERVICE_HOST_DEFAULT = "https://cert.cloud.redhat.com"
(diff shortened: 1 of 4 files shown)


repos/system_upgrade/common/libraries/fetch.py
```diff
@@ -15,15 +16,25 @@
 ASSET_PROVIDED_DATA_STREAMS_FIELD = 'provided_data_streams'
 
 
+def _get_hint():
+    rpmname = 'leapp-upgrade-el{}toel{}'.format(get_source_major_version(), get_target_major_version())
+    hint = (
+        'All official data files are nowadays part of the installed rpms.'
+        ' This issue is usually visible when data files are incorrectly customized, replaced, or removed'
```
MichalHe, line 23:
```suggestion
        ' This issue is usually encountered when the data files are incorrectly customized, replaced, or removed'
```

repos/system_upgrade/common/libraries/fetch.py
```diff
@@ -75,7 +88,9 @@ def read_or_fetch(filename,
 
     # try to get the data locally
     if not os.path.exists(local_path):
-        logger.warning("File {lp} does not exist, falling back to online service".format(lp=local_path))
+        if not allow_download:
+            _raise_error(local_path, "File {lp} does not exist.".format(lp=local_path))
+        logger.warning("File {lp} does not exist, falling back to online servic)".format(lp=local_path))
```
MichalHe, line 93:
```suggestion
        logger.warning("File {lp} does not exist, falling back to online service)".format(lp=local_path))
```

MichalHe (commented):
Tiny topos, otherwise cool.

MichalHe:
Rewriting `allow_download=True` and removing one of the data files results in:
```
Message: Data file /etc/leapp/files/device_driver_deprecation_data.json is missing or invalid.
Summary:
    Details: File /etc/leapp/files/device_driver_deprecation_data.json does not exist.
    Hint: All official data files are nowadays part of the installed rpms. This issue is usually visible when data files are incorrectly customized, replaced, or removed (e.g. by custom scripts). In case you want to recover the original file, remove it (if still exists) and reinstall the leapp-upgrade-el8toel9 rpm.
```
With all data files present, everything works without problems.

MichalHe (approved):

pirat89:
@MichalHe  fixed additional occurence of the link to the article outside of the fetch library

MichalHe (approved):
I ran a grep to check whether there are more occurrences of the URL, there should be none.

matejmatuska (approved):

pirat89:
Thanks guys! Merging

PR #1110, opened 2023-08-10
Further narrow down packit tests

fernflower:
- Get rid of the sad uefi_upgrade test for now;
- Reduce time consuming partitioning tests to 3.

in demand /rerun command-scheduled tests will still be running the full destructive test set (no max_sst though).

PR #1124, opened 2023-08-24
Fix the issue of going out of bounds in the isccfg parser.

SandakovMM:
This problem can occur when attempting to parse an empty file.

### DIFF
--- a/repos/system_upgrade/el7toel8/libraries/isccfg.py
+++ b/repos/system_upgrade/el7toel8/libraries/isccfg.py
@@ -688,8 +688,11 @@ class IscConfigParser(object):
 
         while index != -1:
             keystart = index
-            while istr[index] in self.CHAR_KEYWORD and index < end_index:
+            while index < end_index and istr[index] in self.CHAR_KEYWORD:
                 index += 1
+
+            if index >= end_index:
+                break
 
             if keystart < index <= end_index and istr[index] not in self.CHAR_KEYWORD:
                 # key has been found
--- a/repos/system_upgrade/el7toel8/libraries/tests/test_isccfg.py
+++ b/repos/system_upgrade/el7toel8/libraries/tests/test_isccfg.py
@@ -116,6 +116,10 @@ view "v2" {
 };
 """)
 
+config_empty = isccfg.MockConfig('')
+
+config_empty_include = isccfg.MockConfig('options { include "/dev/null"; };')
+
 
 def check_in_section(parser, section, key, value):
     """ Helper to check some section was found
@@ -341,6 +345,34 @@ def test_walk():
     assert 'options' in state
     assert 'dnssec-lookaside' in state
     assert 'dnssec-validation' not in state
+
+
+def test_empty_config():
+    """ Test empty configuration """
+
+    callbacks = {}
+
+    parser = isccfg.IscConfigParser(config_empty)
+    assert len(parser.FILES_TO_CHECK) == 1
+    cfg = parser.FILES_TO_CHECK[0]
+    parser.walk(cfg.root_section(), callbacks)
+    assert cfg.buffer == ''
+
+
+def test_empty_include_config():
+    """ Test empty configuration """
+
+    callbacks = {}
+
+    parser = isccfg.IscConfigParser(config_empty_include)
+    assert len(parser.FILES_TO_CHECK) == 2
+    cfg = parser.FILES_TO_CHECK[0]
+    parser.walk(cfg.root_section(), callbacks)
+    assert cfg.buffer == 'options { include "/dev/null"; };'
+
+    null_cfg = parser.FILES_TO_CHECK[1]
+    parser.walk(null_cfg.root_section(), callbacks)
+    assert null_cfg.buffer == ''
 
 
 if __name__ == '__main__':


pirat89:
@SandakovMM thanks for the  contribitioun. LGTM.  @pemensik are you ok with the change?

fernflower (changes requested):
Thanks for the fix! Could you please add a unit test?

SandakovMM:
> Could you please add a unit test?
Sure

SandakovMM:
The test currently checks if the code will not crash. I'm not sure what else we could add to the check at this point. Please share your ideas if you have any.

repos/system_upgrade/el7toel8/libraries/isccfg.py
```diff
@@ -686,9 +686,12 @@ def find_next_key(self, cfg, index=0, end_index=-1, end_report=False):
 
         while index != -1:
             keystart = index
-            while istr[index] in self.CHAR_KEYWORD and index < end_index:
+            while index < end_index and istr[index] in self.CHAR_KEYWORD:
                 index += 1
 
+            if index >= end_index:
```
pemensik, line 692: These changes seem correct, fixes the problem for me

repos/system_upgrade/el7toel8/libraries/tests/test_isccfg.py
```diff
@@ -116,6 +116,8 @@
 };
 """)
 
+views_empty = isccfg.MockConfig("")
+
```
pemensik, line 120: I would suggest also ``MockConfig('options { include "/dev/null"; };')`` testing, so we test also empty include file. I would rename it to ``config_empty`` variable, it contains no view.

repos/system_upgrade/el7toel8/libraries/tests/test_isccfg.py
```diff
@@ -343,5 +345,22 @@ def test_walk():
     assert 'dnssec-validation' not in state
 
 
+def test_empty_config():
+    """ Test empty configuration """
+
+    callbacks = {
```
pemensik, line 351: for empty file no callbacks make sense or are needed. Use just ``callbacks = {}``.

repos/system_upgrade/el7toel8/libraries/tests/test_isccfg.py
```diff
@@ -343,5 +345,22 @@ def test_walk():
...
+        'options': cb_state,
+        'dnssec-lookaside': cb_state,
+        'dnssec-validation': cb_state,
+    }
+    state = {}
```
pemensik, line 356: Remove state usage, make it default to None. Unlike views test it is not needed here.

repos/system_upgrade/el7toel8/libraries/tests/test_isccfg.py
```diff
@@ -343,5 +345,22 @@ def test_walk():
...
+
+    parser = isccfg.IscConfigParser(views_empty)
+    assert len(parser.FILES_TO_CHECK) == 1
+    cfg = parser.FILES_TO_CHECK[0]
+    parser.walk(cfg.root_section(), callbacks, state)
```
pemensik, line 361: Omit 3rd parameter, not needed.

pemensik (changes requested):
Core change is good, I would simplify the new test function a bit. And add additional include "/dev/null"; to test also included file check.

repos/system_upgrade/el7toel8/libraries/tests/test_isccfg.py
SandakovMM: We can probably remove the `state` variable since it can be omitted from the `walk` call

fernflower:
@SandakovMM I could be mistaken, but I see that some comments from @pemensik have not been addressed (like https://github.com/oamg/leapp-repository/pull/1124/files/0f5b3da07b0ba3af50d88703e8620f54cd9af9be#diff-3755f648a6265f55301fda90f4333a3289285caa586c8bd8321c90d7746ae32bR361).

fernflower:
My bad, all addressed, I will schedule tests and merge it when they are done. Thanks!

fernflower:
Just one thing - could you please rebase?

pirat89:
> Certainly, would you prefer to consolidate all the modifications into a single commit?

@SandakovMM  yes please :)  thanks

fernflower:
Could you please rebase? I'd like to run regression tests for this, but I need https://github.com/oamg/leapp-repository/commit/93c6fd4f150229a01ba43ce74214043cffaf7dce (and ~10 later commits that were delivered by https://github.com/oamg/leapp-repository/pull/1110)

SandakovMM:
Sure

fernflower:
Either that is my bad and Friday speaking, or your branch https://github.com/SandakovMM/leapp-repository/tree/master is still missing the commits I need to enable testing :(
You need to rebase your branch against leapp-repository's master.

SandakovMM:
My bad, seem like I've rebased agains https://github.com/AlmaLinux/leapp-repository, not https://github.com/oamg/leapp-repository. Should be fine now.

fernflower (approved):
lgtm, when the tests pass I'll merge it.

fernflower:
@pirat89 but I already did that 2 days ago..

pirat89:
@fernflower aah. so you just forgot to set the label :)

PR #1127, opened 2023-09-14
checknfs: do not check systemd mounts

MichalHe:
Optional, keeping this code breaks nothing, however, the inhibitor cannot be reproduced.

Systemd mounts contain only *block* devices. Therefore, the list can never contain NFS shares at all and the check is redundant. This is apparent if one reads storagescanner/libraries/storagescanner.py:L251. This patch, therefore, removes the check for systemd mount alltogether.

@pirat89 @fernflower @matejmatuska

### DIFF
--- a/repos/system_upgrade/common/actors/checknfs/actor.py
+++ b/repos/system_upgrade/common/actors/checknfs/actor.py
@@ -10,7 +10,7 @@ class CheckNfs(Actor):
     """
     Check if NFS filesystem is in use. If yes, inhibit the upgrade process.
 
-    Actor looks for NFS in the following sources: /ets/fstab, mount and systemd-mount.
+    Actor looks for NFS in the following sources: /ets/fstab and mount.
     If there is NFS in any of the mentioned sources, actors inhibits the upgrade.
     """
     name = "check_nfs"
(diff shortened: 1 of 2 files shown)


fernflower:
I have used your pr's build for some unrelated upgrade testing and ended up with a crash
```
2023-09-15 07:48:24.63  INFO     PID: 4727 leapp.workflow.Checks: Executing actor check_nfs 
Process Process-370:
    self.run()
  File "/usr/lib/python3.6/site-packages/leapp/repository/actor_definition.py", line 74, in _do_run
  File "/usr/lib/python3.6/site-packages/leapp/actors/__init__.py", line 289, in run
    self.process(*args)
  File "/etc/leapp/repos.d/system_upgrade/common/actors/checknfs/actor.py", line 44, in process
    if any(fstab_nfs_mounts, nfs_mounts):
TypeError: any() takes exactly one argument (2 given)
2023-09-15 07:48:24.183 ERROR    PID: 4727 leapp.workflow.Checks: Actor check_nfs has crashed: Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/leapp/repository/actor_definition.py", line 74, in _do_run
  File "/usr/lib/python3.6/site-packages/leapp/actors/__init__.py", line 289, in run
    self.process(*args)
  File "/etc/leapp/repos.d/system_upgrade/common/actors/checknfs/actor.py", line 44, in process
    if any(fstab_nfs_mounts, nfs_mounts):
TypeError: any() takes exactly one argument (2 given)
```
I believe the rest of the tests failed for this particular reason as well

fernflower (changes requested):
Please fix the traceback

MichalHe:
@fernflower Thanks, should be fixed.

MichalHe:
@fernflower It isn't that you cannot mount a NFS share using systemd-mount, its that the `systemd_mounts` entry contains only block devices, see [storagescanner](https://github.com/oamg/leapp-repository/blob/4d8ad1c0363fc21f5d8a557f3319a6efacac9f2a/repos/system_upgrade/common/actors/storagescanner/libraries/storagescanner.py#L241), see https://stackoverflow.com/questions/58914555/nfs-mount-point-as-a-disk-device-linux

pirat89 (approved):
It make sense. code lgtm. waiting for tests (restarted as they stucked)

pirat89:
Rebased. hopefully tests will pass now.

PR #1130, opened 2023-10-10
Remove TUV from supported target channels

mkluson:
TUS (mispelled as TUV) is not suported channel for inplace upgrade, removed from the code.

Jira: OAMG-7288


pirat89:
@mkluson  thanks!! rebase it please against up-to-date upstream so you have fix for linters and spellcheckers

PR #1131, opened 2023-10-20
Openssl and ibmca

pirat89:
* The openssl-ibmca needs to be reconfigured manually after the upgrade.
  Report it to the user if the package is installed.

* The openssl configuration file (/etc/pki/tls/openssl.cnf) is not
  100% compatible between major verions of RHEL due to different
  versions of OpenSSL. Also the configuration is supposed to be
  done via system wide crypto policies instead, so it's expected
  to not modify this file anymore. If the content of the file has
  been modified, report to user what will happen during the upgrade
  and what they should do after it.

* If the openssl config file is modified (rpm -Vf <file>) and
  *.rpmnew file exists, back up the file with .leappsave suffix
  and replace it by the *.rpmsave one.

## Introduce TrackedFilesInfoSource
I see here some pattern, when it's needed to check just whether a file has been changed, etc.. Regarding that, I'Ve introduced `TrackedFilesInfoSource` msg and a generic actor that creates it. So in case someone needs similar information in future, they can add the file path to the input list defined in the actor without the need to create additional extra actor for it.

## Tickets
JIRA: OAMG-9960
JIRA: RHEL-11193

Missing unit-tests for `scansourcefiles` actor. Expecting to deliver them in a followup PR. Tracked by JIRA ticket: OAMG-10367

### DIFF
--- a/repos/system_upgrade/common/actors/openssl/migrateopensslconf/actor.py
+++ b/repos/system_upgrade/common/actors/openssl/migrateopensslconf/actor.py
@@ -1,0 +1,26 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import migrateopensslconf
+from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
+
+
+class MigrateOpenSslConf(Actor):
+    """
+    Enforce the target default configuration file to be used.
+
+    If the /etc/pki/tls/openssl.cnf has been modified and openssl.cnf.rpmnew
+    file is created, backup the original one and replace it by the new default.
+
+    tl;dr: (simplified)
+    if the file is modified; then
+      mv /etc/pki/tls/openssl.cnf{,.leappsave}
+      mv /etc/pki/tls/openssl.cnf{.rpmnew,}
+    fi
+    """
+
+    name = 'migrate_openssl_conf'
+    consumes = ()
+    produces = ()
+    tags = (IPUWorkflowTag, ApplicationsPhaseTag)
+
+    def process(self):
+        migrateopensslconf.process()
--- a/repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
+++ b/repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
@@ -1,0 +1,5 @@
+def test_scansourcefiles():
+    # TODO(pstodulk): keeping unit tests for later after I check the idea
+    # of this actor with the team.
+    # JIRA: OAMG-10367
+    pass
--- a/repos/system_upgrade/common/models/trackedfiles.py
+++ b/repos/system_upgrade/common/models/trackedfiles.py
@@ -1,0 +1,60 @@
+from leapp.models import fields, Model
+from leapp.topics import SystemInfoTopic
+
+
+class FileInfo(Model):
+    """
+    Various data about a file.
+
+    This model is not supposed to be used as a message directly.
+    See e.g. :class:`TrackedSourceFilesInfo` instead.
+    """
+    topic = SystemInfoTopic
+
+    path = fields.String()
+    """
+    Canonical path to the file.
+    """
+
+    exists = fields.Boolean()
+    """
+    True if the file is present on the system.
+    """
+
+    rpm_name = fields.String(default="")
+    """
+    Name of the rpm that owns the file. Otherwise empty string if not owned
+    by any rpm.
+    """
+
+    # NOTE(pstodulk): I have been thinking about the "state"/"modified" field
+    # instead. Which could contain enum list, where could be specified what has
+    # been changed (checksum, type, owner, ...). But currently we do not have
+    # use cases for that and do not want to implement it now. So starting simply
+    # with this one.
+    is_modified = fields.Boolean()
+    """
+    True if the checksum of the file has been changed (includes the missing state).
+
+    The field is valid only for a file tracked by rpm - excluding ghost files.
+    In such a case the value is always false.
+    """
+
+
+class TrackedFilesInfoSource(Model):
+    """
+    Provide information about files on the source system explicitly defined
+    in the actor to be tracked.
+
+    Search an actor producing this message to discover the list where you
+    could add the file into the list to be tracked.
+
+    This particular message is expected to be produced only once by the
+    specific actor. Do not produce multiple messages of this model.
+    """
+    topic = SystemInfoTopic
+
+    files = fields.List(fields.Model(FileInfo), default=[])
+    """
+    List of :class:`FileInfo`.
+    """
--- a/repos/system_upgrade/common/actors/openssl/checkopensslconf/actor.py
+++ b/repos/system_upgrade/common/actors/openssl/checkopensslconf/actor.py
@@ -1,0 +1,33 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import checkopensslconf
+from leapp.models import DistributionSignedRPM, Report, TrackedFilesInfoSource
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckOpenSSLConf(Actor):
+    """
+    Check whether the openssl configuration and openssl-IBMCA.
+
+    See the report messages for more details. The summary is that since RHEL 8
+    it's expected to configure OpenSSL via crypto policies. Also, OpenSSL has
+    different versions between major versions of RHEL:
+      * RHEL 7: 1.0,
+      * RHEL 8: 1.1,
+      * RHEL 9: 3.0
+    So OpenSSL configuration from older system does not have to be 100%
+    compatible with the new system. In some cases, the old configuration could
+    make the system inaccessible remotely. So new approach is to ensure the
+    upgraded system will use always new default /etc/pki/tls/openssl.cnf
+    configuration file (the original one will be backed up if modified by user).
+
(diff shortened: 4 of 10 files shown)


repos/system_upgrade/common/actors/openssl/migrateopensslconf/actor.py
```diff
@@ -0,0 +1,26 @@
...
+
+class MigrateOpenSslConf(Actor):
+    """
+    Enforce the target default configuration file to be used.
+
+    If the /etc/pki/tls/openssl.cnf has been modified and openssl.cnf.rpmnew
+    file is created, backup the original one and replace it by the new default.
+
+    tl;dr:
+    if the file is modified; then
```
kkaarreell, line 14: This is simple but incorrect way how to approach it. `rpmsave` and `rpmnew` files are created during `rpm` transaction, which means that if a config file has been updated AFTER the latest update, these files won't exist.
I am afraid there is no simple way how to restore the original config file (I mean other than extracting it from RPM file) but better to consult RPM folks.

pirat89: @kkaarreell you are right, and the code treats that - the tl;dr; is not exact. If I understand correctly your concern, it should be ok as:
* if the file has been modified before the transaction, it rpmnew file will exist
* if the file has been modified after the transaction (so the rpmnew is not created during the transaction), we consider the change as intentional and we do nothing

In this part, we already cannot rely on having the rpm locally (and we cannot rely on internet), so taking it from the package is not a point here. We could think about having the default file stored as a part of the actor, however that will not cover cases when the file is changed (z-stream, differences  between minor versions when nobody track this, etc.). We need to take into account, that nothing prevents customer to change the file anyway after this piece of code is executed.

pirat89: Discussed over video call. Current approach is ok from the point of possible missing *.rpmnew file.

repos/system_upgrade/common/models/trackedfiles.py
```diff
@@ -0,0 +1,60 @@
...
+    is_modified = fields.Boolean()
+    """
+    True if the checksum of the file has been changed (includes the missing state).
+
+    The field is valid only for a file tracked by rpm - excluding ghost files.
+    In such a case the value is always false.
+    """
+
+
+class TrackedSourceFilesInfo(Model):
```
pirat89, line 44: @oamg/developers  we have discussed the name is confusing/amgiguous/no so clear. Let's put into this thread proposals about a different name.

For a context, providing reasoning behind the current (`TrackedSourceFilesInfo`) name:
* Tracked - it's targetting only files we explecitely marked (now in the code..) to track
* Source - it needs to be clear that information here is about the `source` system (that's terminology we use for the original system, used also in other existing models)
  * To be clear that info here does not have to be valid e.g. ager the upgrade transaction.
  * Make our way clear in case we would like to introduce in future "Target" version of this model
* FilesInfo - ...

My first additional proposal into the plenum (not so original... :D):
* TrackedFilesInfoSource

dkubek: I think maybe the original ``TrackedSourceFilesInfo`` sounds the best and makes the most sense after all.  Or maybe after stretching the grammar a little ...

- What are we interested in? => ``FilesInfo``
- File info from what files? -> only tracked! => ``TrackedFilesInfo``
- But where is being tracked from? -> source! => ``SourceTrackedFilesInfo`` (this sounds like the source is the one doing the tracking though) or more explicit ``TrackedFilesInfoFromSource`` (``TrackedFilesInfoOfSource`` ?)

pirat89: @dkubek thanks for the thinking about this. Basically we have very similar opinion on this. I prefer to keep the `Source` as a suffix in this case to be more compatible with @matejmatuska`s modesl for systemd services. However, I want to omit `From` and `Of` substrings in the name to keep it shorter and also due to consistency, as we do not use them neither.

We should write some best practices, tips .... for model names. It could help people with orientation in the set of models.

pirat89:
@oamg/developers  I will most likely create a houseleeping ticket to create unit-tests for scansourcefiles actor later. The rest of the code is covered now.

fernflower (commented):
The unit tests for scansourcefiles actor are missing. I am not really comfortable merging the code without them, as standard upstream tests we run (even tier1 ones) might not trigger that actor execution.
Otherwise lgtm

repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
```diff
@@ -0,0 +1,4 @@
+def test_scansourcefiles():
+    # TODO(pstodulk): keeping unit tests for later after I check the idea
```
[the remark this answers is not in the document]
pirat89: Added Jira ticket to the comments: OAMG-10367. Regarding the time frame, I expect to deliver tests the next week. But most likely I will not be able to finish them today.

Rezney (approved):

fernflower:
Checked manually that scanner doesn't crash.

fernflower (approved):

pirat89:
Doing some additional manual sanity testing. For systems with changed openssl.cnf config file:
```
Risk Factor: high
Title: The /etc/pki/tls/openssl.cnf file is modified and will be replaced during the upgrade.
Summary: The OpenSSL configuration file (/etc/pki/tls/openssl.cnf) has been modified on the system. RHEL 8 (and newer) systems provide a crypto-policies mechanism ensuring usage of system-wide secure cryptography algorithms. Also the target system uses newer version of OpenSSL that is not fully compatible with the current one. To ensure the upgraded system uses crypto-policies as expected, the new version of the openssl configuration file must be installed during the upgrade. This will be done automatically. The original configuration file will be saved as "/etc/pki/tls/openssl.cnf.leappsave".

Note this can affect the ability to connect to the system after the upgrade if it depends on the current OpenSSL configuration. Such a problem may be caused by using a particular OpenSSL engine, as OpenSSL engines built for the RHEL 7 system are not compatible with RHEL 8.
Related links:
    - Using system-wide cryptographic policies.: https://red.ht/rhel-9-system-wide-crypto-policies
Remediation: [hint] Check that your ability to login to the system does not depend on the OpenSSL configuration. After the upgrade, review the system configuration and configure the system as needed. Please, be aware that it is not recommended to configure the system default {fpath}. Instead, it is recommended to copy the file and use this copy to configure particular applications. The default OpenSSL configuration file should be modified only when it is really necessary.
Key: 20e32da940e6b410dc82fed685233c9033066b74
```

During the upgrade:
```
Actor: migrate_openssl_conf
Executed: True
Phase: Applications
Started: 2024-01-12T11:31:08.737660Z
Produced messages:
    ----
Executed shell commands:
    - ['rpm', '-Vf', '/etc/pki/tls/openssl.cnf']
    - ['mv', '/etc/pki/tls/openssl.cnf', '/etc/pki/tls/openssl.cnf.leappsave']
    - ['mv', '/etc/pki/tls/openssl.cnf.rpmnew', '/etc/pki/tls/openssl.cnf']
Logs:
    2024-01-12 12:31:08.771 DEBUG    PID: 1293 leapp.workflow.Applications.migrate_openssl_conf: External command has started: ['rpm', '-Vf', '/etc/pki/tls/openssl.cnf']
    2024-01-12 12:31:08.794 DEBUG    PID: 1293 leapp.workflow.Applications.migrate_openssl_conf: S.5....T.  c /etc/pki/tls/openssl.cnf
    2024-01-12 12:31:08.803 DEBUG    PID: 1293 leapp.workflow.Applications.migrate_openssl_conf: Command ['rpm', '-Vf', '/etc/pki/tls/openssl.cnf'] failed with exit code 1.
    2024-01-12 12:31:08.807 DEBUG    PID: 1293 leapp.workflow.Applications.migrate_openssl_conf: External command has finished: ['rpm', '-Vf', '/etc/pki/tls/openssl.cnf']
    2024-01-12 12:31:08.809 DEBUG    PID: 1293 leapp.workflow.Applications.migrate_openssl_conf: External command has started: ['mv', '/etc/pki/tls/openssl.cnf', '/etc/pki/tls/openssl.cnf.leappsave']
    2024-01-12 12:31:08.816 DEBUG    PID: 1293 leapp.workflow.Applications.migrate_openssl_conf: External command has finished: ['mv', '/etc/pki/tls/openssl.cnf', '/etc/pki/tls/openssl.cnf.leappsave']
    2024-01-12 12:31:08.817 DEBUG    PID: 1293 leapp.workflow.Applications.migrate_openssl_conf: External command has started: ['mv', '/etc/pki/tls/openssl.cnf.rpmnew', '/etc/pki/tls/openssl.cnf']
    2024-01-12 12:31:08.823 DEBUG    PID: 1293 leapp.workflow.Applications.migrate_openssl_conf: External command has finished: ['mv', '/etc/pki/tls/openssl.cnf.rpmnew', '/etc/pki/tls/openssl.cnf']
```

On the upgraded system:
```
root@localhost ~]# ls -l /etc/pki/tls/openssl*
-rw-r--r--. 1 root root 11227 Nov 30 14:02 /etc/pki/tls/openssl.cnf
-rw-r--r--. 1 root root 10925 Jan 12 12:07 /etc/pki/tls/openssl.cnf.leappsave
[root@localhost ~]# rpm -V `rpm -qf /etc/pki/tls/openssl.cnf`
[root@localhost ~]# 
```

PR #1132, opened 2023-10-20
call correct mkdir when trying to create /etc/rhsm/facts

evgeni:
os.path has no mkdir, but os does.

traceback without the patch:

      File "/bin/leapp", line 11, in <module>
        load_entry_point('leapp==0.16.0', 'console_scripts', 'leapp')()
      File "/usr/lib/python3.6/site-packages/leapp/cli/__init__.py", line 45, in main
        cli.command.execute('leapp version {}'.format(VERSION))
      File "/usr/lib/python3.6/site-packages/leapp/utils/clicmd.py", line 111, in execute
        args.func(args)
      File "/usr/lib/python3.6/site-packages/leapp/utils/clicmd.py", line 133, in called
        self.target(args)
      File "/usr/lib/python3.6/site-packages/leapp/cli/commands/upgrade/breadcrumbs.py", line 170, in wrapper
        breadcrumbs.save()
      File "/usr/lib/python3.6/site-packages/leapp/cli/commands/upgrade/breadcrumbs.py", line 116, in save
        self._save_rhsm_facts(doc['activities'])
      File "/usr/lib/python3.6/site-packages/leapp/cli/commands/upgrade/breadcrumbs.py", line 64, in _save_rhsm_facts
        os.path.mkdir('/etc/rhsm/facts')
    AttributeError: module 'posixpath' has no attribute 'mkdir'

While at it, also catch OSError with errno 17, to safeguard against race
conditions if anything has created the directory between us checking for
it and us trying to create it.

### DIFF
--- a/commands/upgrade/breadcrumbs.py
+++ b/commands/upgrade/breadcrumbs.py
@@ -61,7 +61,12 @@ class _BreadCrumbs(object):
             if not os.path.exists('/etc/rhsm'):
                 # If there's no /etc/rhsm folder just skip it
                 return
-            os.path.mkdir('/etc/rhsm/facts')
+            try:
+                os.mkdir('/etc/rhsm/facts')
+            except OSError as e:
+                if e.errno == 17:
+                    # The directory already exists which is all we need.
+                    pass
         try:
             with open('/etc/rhsm/facts/leapp.facts', 'w') as f:
                 json.dump(_flattened({


evgeni:

vinzenz (dismissed):

abadger (dismissed):
Approving because this is a bug and the change fixes it.

Possible tangent: would we want to make this more robust as well? Something like this would remove an (unlikely) race condition:
``` python
    def _save_rhsm_facts(self, activities):
        if not os.path.exists('/etc/rhsm'):
            # If there's no /etc/rhsm folder just skip saving rhsm_facts
            return
        try:
            os.mkdir('/etc/rhsm/facts')
        except OSError as e:
            if e.errno == 17:
                # The directory already exists which is all we need.
                pass
```

evgeni:
We could also use `os.makedirs(…, exist_ok=True)` instead of catching the `OSError`?

I don't have a strong preference, as long as the bug gets fixed :)

evgeni:
> We could also use `os.makedirs(…, exist_ok=True)` instead of catching the `OSError`?
Huh, `exist_ok` doesn't exist in Python2, which we probably steel need to keep compat with?

pirat89:
yep, so far we need to be still python2 compatible yet

pirat89:
@evgeni  do you plan to address also @abadger  suggestion? for us, it's ok to merge it as it is, but if you want to address the change, we can wait. no rush :)

evgeni:
There you go, addressed.

abadger (approved):
Thanks.  Looks great!

PR #1134, opened 2023-10-21
load all substitutions from etc

evgeni:
On some distributions (like CentOS Stream and Oracle Linux), we need more substitutions to be able to load repositories properly.

DNF has a helper for that: conf.substitutions.update_from_etc.

On pure DNF distributions, calling this should be sufficient. On EL7, where the primary tool is YUM, DNF does not load vars from /etc/yum, only from /etc/dnf, so we have to help it a bit and explicitly try to load releasever from /etc/yum.
(DNF since 4.2.15 *does* also load substitutions from /etc/yum, but EL7 ships with 4.0.x)

### DIFF
--- a/repos/system_upgrade/common/libraries/module.py
+++ b/repos/system_upgrade/common/libraries/module.py
@@ -1,4 +1,3 @@
-import os
 import warnings
 
 from leapp.libraries.common.config.version import get_source_major_version
@@ -23,14 +22,20 @@ def _create_or_get_dnf_base(base=None):
         # have repositories only for the exact system version (including the minor number). In a case when
         # /etc/yum/vars/releasever is present, read its contents so that we can access repositores on such systems.
         conf = dnf.conf.Conf()
-        pkg_manager = 'yum' if get_source_major_version() == '7' else 'dnf'
-        releasever_path = '/etc/{0}/vars/releasever'.format(pkg_manager)
-        if os.path.exists(releasever_path):
-            with open(releasever_path) as releasever_file:
-                releasever = releasever_file.read().strip()
-                conf.substitutions['releasever'] = releasever
-        else:
-            conf.substitutions['releasever'] = get_source_major_version()
+
+        # preload releasever from what we know, this will be our fallback
+        conf.substitutions['releasever'] = get_source_major_version()
+
+        # dnf on EL7 doesn't load vars from /etc/yum, so we need to help it a bit
+        if get_source_major_version() == '7':
+            try:
+                with open('/etc/yum/vars/releasever') as releasever_file:
+                    conf.substitutions['releasever'] = releasever_file.read().strip()
+            except IOError:
+                pass
+
+        # load all substitutions from etc
+        conf.substitutions.update_from_etc('/')
 
         base = dnf.Base(conf=conf)
         base.init_plugins()


evgeni:

repos/system_upgrade/common/libraries/module.py
```diff
@@ -23,14 +23,16 @@ def _create_or_get_dnf_base(base=None):
...
         # /etc/yum/vars/releasever is present, read its contents so that we can access repositores on such systems.
         conf = dnf.conf.Conf()
-        pkg_manager = 'yum' if get_source_major_version() == '7' else 'dnf'
-        releasever_path = '/etc/{0}/vars/releasever'.format(pkg_manager)
-        if os.path.exists(releasever_path):
+
+        # dnf on EL7 doesn't load vars from /etc/yum, so we need to help it a bit
+        releasever_path = '/etc/yum/vars/releasever'
+        if get_source_major_version() == '7' and os.path.exists(releasever_path):
             with open(releasever_path) as releasever_file:
```
abadger, line 30: I would put the releasever_path setting inside of the condition to show that it is only a part of the code needed to handle EL7:
```suggestion
        # dnf on EL7 doesn't load vars from /etc/yum, so we need to help it a bit
        if get_source_major_version() == '7' and os.path.exists(releasever_path):
            releasever_path = '/etc/yum/vars/releasever'
            with open(releasever_path) as releasever_file:
```

A general note: using a `try: except:` around `open()` and passing if the file does not exist is resistant to race conditions whereas checking file existence first and then operating on the file is not so I generally prefer the `try: except:` strategy.  However, I see (1) that we use the file existence check strategy all over the code right now and (2) this specific case is unlikely to have the file deleted out from under us. So, something to keep in mind but I won't require that change to be made.

abadger (approved):
I don't know the dnf and yum apis intimately but the reasoning here makes sense and I've looked at the relevant code in `dnf.conf.substitutions.update_from_etc()` to check that it takes care of this case in EL8.

The only changes I asked for are cosmetic so I'm approving this.  You can add my suggestion or let me know you'd rather not and then this can be merged.

repos/system_upgrade/common/libraries/module.py
evgeni: I can't call `os.path.exists(releasever_path)` before assigning `releasever_path = '/etc/yum/vars/releasever'` ;-)

The logic I wanted to keep is "if on EL7 *and* the file exists, load it, in *all* other cases use `get_source_major_version`.

I could rewrite this to:
```
conf.substitutions['releasever'] = get_source_major_version()
if get_source_major_version() == '7':
  releasever_path = '/etc/yum/vars/releasever'
  if os.path.exists(releasever_path):
    with …:
      conf.substitutions['releasever'] = releasever_file.read().strip()
```

But it doesn't make the code much more readable, IMHO.

evgeni: Or even
```python
conf.substitutions['releasever'] = get_source_major_version()

if get_source_major_version() == '7':
  try:
    with open('/etc/yum/vars/releasever') as releasever_file:
      conf.substitutions['releasever'] = releasever_file.read().strip()
  except:
    pass

# load all substitutions from etc
conf.substitutions.update_from_etc('/')
```

evgeni: Done this now. Let me know what you think.

abadger: +1

abadger: And yeah, I got ahead of myself thinking about both changing to `try: except:` and moving the assignment but then only making one of those a suggestion.

abadger (approved):
Looks good

abadger:
Thanks @evgeni !

PR #1135, opened 2023-10-26
Draft: Fix and enhance invalid symlink handling when installing certificates to target container

dkubek:
In response to the identified flaws in the originally delivered fix, for feature enabling http repositories, this commit addresses the following issues:

1. Previously, files installed via RPMs that were originally symlinks were being switched to standard files. This issue has been resolved by preserving symlinks within the /etc/pki directory. Any symlink pointing to a file within the /etc/pki directory (whether present in the source system or installed by a package in the container) will be present in the container, ensuring changes to certificates are properly propagated.

2. Lists of trusted CAs were not being updated, as the update-ca-trust call was missing inside the container. This commit now includes the necessary update-ca-trust call.

The solution specification has been modified as follows:

 - Certificate _files_ in /etc/pki (excluding symlinks) are copied to
      the container as in the original solution.
 - Files installed by packages within the container are preserved and
      given higher priority.
 - Handling of symlinks is enhanced, ensuring that symlinks within
      the /etc/pki directory are preserved, while any symlink pointing
      outside the /etc/pki directory will be copied as a file.
 - Certificates are updated using `update-ca-trust`.

JIRA: [OAMG-6388](https://issues.redhat.com/browse/OAMG-6388)

Note: Also fixes [OAMG-5073](https://issues.redhat.com/browse/OAMG-5073)


abadger:
We need to make sure that this fixes this: https://github.com/oamg/leapp-repository/pull/1136#issuecomment-1789528335   and that the original issue there doesn't regress with this change. Perhaps a unitest checking that a dangling symlink in /etc/pki is copied over as a dangling symlink to the target container?

abadger:
Right.  I'm not sure if ignoring dangling symlinks or not breaking when we do so is the correct thing to do.  But since pstodulka solved that particular issue by not creating a dangling symlink, it's not something we have to decide on now.

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -345,12 +345,73 @@ def _get_files_owned_by_rpms(context, dirpath, pkgs=None, recursive=False):
...
+            pointee = None
+            if os.path.islink(source_filepath):
+                pointee = os.readlink(source_filepath)
+
+            # If source file is a symlink within `src` then preserve it,
+            # otherwise resolve and copy it as a file it points to
+            if pointee is not None and not pointee.startswith(srcdir):
+                # Follow the path until we hit a file or get back to /etc/pki
+                while not pointee.startswith(srcdir) and os.path.islink(pointee):
+                    pointee = os.readlink(pointee)
```
abadger, line 381: This misses one case: if you have a set of circular symlinks in /etc/pki.  To solve that, I think I would make the symlink portion into a helper function that can be called recursively, passing in the list of links in `/etc/pki` that have already been found (while dereferencing this particular link).  If the recursive function detects a link in `/etc/pki` that is already found, then it would decide that all of the links in the chain are broken symlinks and proceed accordingly.

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -345,12 +345,73 @@ def _get_files_owned_by_rpms(context, dirpath, pkgs=None, recursive=False):
...
+                    pass
+
+            # Ensure parent directory exists
+            parent_dir = os.path.dirname(target_filepath)
+            run(['mkdir', '-p', parent_dir])
+
+            if os.path.islink(source_filepath):
+                # Preserve the owner and permissions of the original symlink
+                run(['ln', '-s', pointee, target_filepath])
+                run(['chmod', '--reference={}'.format(source_filepath), target_filepath])
```
abadger, line 398: Just a note for you: on other platforms, symlinks can have their own permissions.  So if we had to be portable to non-Linux systems, you might have to use the real files rather than the symlinks here.

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -360,36 +421,52 @@ def _copy_certificates(context, target_userspace):
...
 
+    # Backup container /etc/pki
     run(['mv', target_pki, backup_pki])
-    context.copytree_from('/etc/pki', target_pki)
 
+    # Copy source /etc/pki to the container
+    _copy_decouple('/etc/pki', target_pki)
+
+    # Assertion: no broken symlinks exist in /etc/pki in the container
+    # So any broken symlinks created will be by the installed packages.
```
abadger, line 431:
```suggestion
    # Assertion: after running _copy_decouple(), no broken symlinks exist in /etc/pki in the container
    # So any broken symlinks created will be by the installed packages.
```

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -360,36 +421,52 @@ def _copy_certificates(context, target_userspace):
...
+
+    # Recover installed packages as they always get precedence
     for filepath in files_owned_by_rpms:
```
abadger, line 434: What is the difference in what the following code is attempting to achieve and the code in _copy_decouple?  It feels like the link detection and possible dereferencing and copying would be the same and the main difference is just that we're looping over different files?  The implementations are a bit different so I'm not sure if I'm understanding it, but if the purpose is the same, we should look at making it a helper function that we call from both places.

repos/system_upgrade/common/actors/targetuserspacecreator/tests/unit_test_targetuserspacecreator.py
```diff
@@ -48,6 +50,183 @@ def __exit__(self, exception_type, exception_value, traceback):
...
+#
+# 2. Depending on the value for a key in a dict, each key in the dictionary
+#    denotes the name of either a:
+#        a) directory -- if value is dict
+#        b) regular file -- if value is None
+#        c) symlink -- if a value is str
+#
+# 3. The value of a symlink entry is a absolute path to a file in the context of
+#    the structure.
+#
```
abadger, line 106: We also need A => B, B => C, C => A (expectation: broken symlink so none of the three files is present)

repos/system_upgrade/common/actors/targetuserspacecreator/tests/unit_test_targetuserspacecreator.py
```diff
@@ -1,5 +1,7 @@
 import os
```
abadger, line 1: The division without future message from pylint is kind of a false positive in this case (it won't affect the division operator used with pathlib) but it's easy enough to fix:

```suggestion
from __future__ import division

import os
```

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -401,6 +478,10 @@ def _prep_repository_access(context, target_userspace):
     backup_yum_repos_d = os.path.join(target_etc, 'yum.repos.d.backup')
 
     _copy_certificates(context, target_userspace)
+    # NOTE(dkubek): context.call(['update-ca-trust']) seems to not be working.
+    #               I am not really sure why. The changes to files are not
+    #               being written to disk.
```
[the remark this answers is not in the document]
pirat89: I think this is not suitable for unit-tests (in the way that unit-tests should uncover this problem) as we want to omit unit-tests that are actually somehow messing with the system (it would require an installation of a container somewhere, ....). This is something what should be covered by integration tests.

One difference between the `run` and `context.call` is, that the run operation executes `chroot`, in which the command is performed. Whereas the `context.call` performs `systemd-nspawn`. Also env & some sys/run/... files could be different. I am curious what is the real reason behind that difference. We should definitely check that. @Jakuje any thought?

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -345,12 +345,73 @@ def _get_files_owned_by_rpms(context, dirpath, pkgs=None, recursive=False):
...
 
 
+def _copy_decouple(srcdir, dstdir):
+    """
+    Copy `srcdir` to `dstdir` while decoupling symlinks.
+
+    What we mean by decoupling the `srcdir` is that any symlinks pointing
+    outside the directory will be copied as regular files. This means that the
+    directory will become independent from its surroundings with respect to
+    symlinks. Any symlink (or symlinks chains) within the directory will be
```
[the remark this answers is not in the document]
dkubek: Oops! Fixing, thanks!

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -345,12 +345,73 @@ def _get_files_owned_by_rpms(context, dirpath, pkgs=None, recursive=False):
...
+    preserved.
+
+    """
+
+    for root, _, files in os.walk(srcdir):
```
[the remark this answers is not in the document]
dkubek: dummy_me. Fixing!

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -345,12 +345,73 @@ def _get_files_owned_by_rpms(context, dirpath, pkgs=None, recursive=False):
...
+        for filename in files:
+            relpath = os.path.relpath(root, srcdir)
+            source_filepath = os.path.join(root, filename)
+            target_filepath = os.path.join(dstdir, relpath, filename)
+
+            # Skip and report broken symlinks
+            if not os.path.exists(source_filepath):
+                api.current_logger().warning('File {} is a broken symlink!'.format(source_filepath))
```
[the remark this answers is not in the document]
dkubek: Right, this is much more informative of what is really happening to the user (or well to everyone...). Thanks! Adding!

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
dkubek: Thanks! However, using the real files was the original solution which we had to adjust because changes to  certificates were not propagated it the container (as what were previously symlinks are now regular files which are not updated). So I generally need to preserve the symlink here. Not sure if this is what you meant though. Just out of curiosity, how would you solve this on non-Linux systems?

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
dkubek: Right, makes it much more clear. Adding!

repos/system_upgrade/common/actors/targetuserspacecreator/tests/unit_test_targetuserspacecreator.py
dkubek: Thanks! Good thing I did not rewrite the whole thing using os.join-s as I intended originally...

repos/system_upgrade/common/actors/targetuserspacecreator/tests/unit_test_targetuserspacecreator.py
dkubek: Sure, totally forgot about this possibility. Adding!

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
dkubek: I think this is already handled by ``os.path.exists``. On a toy example a circular symlinks returns ``False`` and so it will be marked as a broken symlink before and not considered at all.

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
dkubek: Yes I agree that the implementation is quite similar and I have considered to abstract it to to a common function, however there is a subtle difference in checking whether the symlink is broken. When checking a symlink ``/etc/pki/A -> /etc/pki/B`` on the source system, it is easy to dereference it to file ``/etc/pki/B``, however consider a symlink installed in the container ``$target_container/etc/pki/A -> /etc/pki/B`` - every time we dereference a link we have to manually update the path to point to a file within the the context of the container. Another small difference is that we have made the decision to respect the files installed by RPMs. So whenever they point to a file outside ``/etc/pki`` we keep this symlink there, even though when copying certificates from the source system we would replace this link with the file. I am not sure whether this happens or not or makes a significant difference. With these small differences I felt justified in keeping the two procedures separate (even though very similar).  However if you still feel differently I will figure something out.

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -345,12 +345,73 @@ def _get_files_owned_by_rpms(context, dirpath, pkgs=None, recursive=False):
...
+                continue
+
+            # Copy symlinks to the target userspace
+            pointee = None
+            if os.path.islink(source_filepath):
```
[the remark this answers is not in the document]
dkubek: As per your sugestion I have modified the function slightly, do you think it is more readable now?

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
abadger: <nod> I was a little unclear, I think.  On other platforms (some of the BSDs for sure) symlinks themselves have file permissions.  So if we needed to be portable to those, we'd need to make sure that the chmod was getting its permissions from the actual source file rather than a symlink and we might have to set those file permissions on all of the possible target_filepaths (the actual target file and the symlink chain leading to it.)  Symlinks with permissions is not a POSIX standard, though, so we'd probably need to test on each platform we cared about to see exactly how their chmod treated the permissions.

pirat89 (approved):
lgtm - discovered just one small corner case, that is ok to be merged if not fixed - see my comment.

I've also tested the behaviour when TLS configuration is broken after the `update-ca-trust` execution (can be reproduced when upgrade RHEL 6 -> 7 and use an older releases)

pirat89:
@dkubek let's continue with the merge, and in case of another changes, let's do them in another iteration, so we unblock the testing for CTC1.

@abadger it seems that all required changes have been applied. let us know in case you find something is missing, so it could be addressed in the next iteration.

Thank you guys for the good work!

abadger:
+1.  Merging is good with me.  I had some thoughts on code structure but I don't know if they'll pan out.  Since I'm on PTO and then a US-only holiday, I'll see if I can make any of that work for fun and submit a PR for @dkubek to review if so :-)

PR #1137, opened 2023-10-30
support rhel 8 upgrade to 9 on alibaba cloud

jinkkkang:
1 we add repomap in repomap.json including rhel 8 map to rhel 9 (x86 & arm).
2 we add alibaba rhui map in rhui.py.
3 rhel 8 repo rpm was signed with sha1,   there will have a check on checkdeprecatedrpmsignature.py for deprecated signature,  so we add a  whitelist for alibaba cloud,  repo rpm on rhel 9 have been updated  signed  with sha256， the new signature package will be installed in the new rhel 9  image in the future.

### DIFF
--- a/repos/system_upgrade/common/libraries/rhui.py
+++ b/repos/system_upgrade/common/libraries/rhui.py
@@ -481,6 +481,17 @@ RHUI_CLOUD_MAP = {
                 ('leapp-google-sap.repo', YUM_REPOS_PATH)
             ],
         },
+        'alibaba': {
+            'src_pkg': 'aliyun_rhui_rhel8',
+            'target_pkg': 'aliyun_rhui_rhel9',
+            'leapp_pkg': 'leapp-rhui-alibaba',
+            'leapp_pkg_repo': 'leapp-alibaba.repo',
+            'files_map': [
+                ('content.crt', RHUI_PKI_PRODUCT_DIR),
+                ('key.pem', RHUI_PKI_DIR),
+                ('leapp-alibaba.repo', YUM_REPOS_PATH)
+            ],
+        },
     },
 }
 


repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
```diff
@@ -59,6 +60,19 @@ def _is_sha1_allowed(current_policy):
...
+    whitelisted_cloud_flavours = (
+        'alibaba',
+    )
+    whitelisted_cloud_pkgs = {
+        rhui.RHUI_CLOUD_MAP[upg_path].get(flavour, {}).get('src_pkg') for flavour in whitelisted_cloud_flavours
+    }
+    for bad_rpm in bad_rpms:
+        if bad_rpm.name in whitelisted_cloud_pkgs:
+            bad_rpms.remove(bad_rpm)
+
```
pirat89, line 75: this is not how it works. such packages can break the rpm transaction due to the old signature, so whitelist is not an option. it's the same as for google cloud. there are 2 options only:
* alibaba update the signature using SHA256
* user manually switch system wide crypto policies to legacy mode or DEFAULT:SHA1

pirat89 (changes requested):

repos/system_upgrade/el8toel9/actors/checkdeprecatedrpmsignature/libraries/checkdeprecatedrpmsignature.py
jinkkkang: 1 we cannot update   signature using SHA256 for rhel 8 repo rpm,  it will cause the upgrade to fail for rhel 7 to 8
2 this  old signature repo rpm will not  break the rpm transaction, it will be removed on upgrade, (added to remove list in system_upgrade/common/actors/cloud/checkrhui/actor.py), we test the rpm transaction is ok and upgrade successful

pirat89: @jinkkkang the problem is, that RHEL 9 rpm does not manipulate with RPM packages with SHA1 signature. if you try to install, update, or remove rpm with SHA1, openssl will not allow the use of SHA1 algorithm and rpm than fail to do anything with the rpm if the key is installed. that why I am saying that the crypto policies needs to be updated otherwise.

Why the signing of packages with SHA256 should block the upgrade 7 -> 8? The only reason I can see here is that the gpg key (in the leapp-rhui package?) needs to be updated also and than there is no reason why the upgrade should not work. All RHEL packages since RHEL 7 are signed with RSA/SHA256, so there is no technical blocker for upgrades.

jinkkkang: we will update  the signature of rhui repo packages on RHEL 8 using SHA256,  we will also update the leapp-rhui-alibaba package for rhel 7 upgrade  to add the sha256 gpgkey to support upgrade (rhel 7 upgrade to 8 will failed without this),   I will update here if there is any new progress, thanks

jinkkkang: @pirat89  hello, we have updated  the signature of  rhui repo rpm for rhel 8   from SHA1 to SHA256, we also removed the section in the code that added a whitelist to rhel 8, please help review

pirat89: Hi @jinkkkang, regarding the repomapping, we will do the change on our side later as the last time (please keep the current changes in this PR as they are in this file for now, we will update them directly manually later - we need to regenerate the file using our scripts to ensure the mapping does not dissapear in future). It seems good otherwise. I am keeping the status of required changes until we deliver the updated repomap json file (Dec / first week in Jan). Please send us the *.repo file from /etc/yum.repos.d/ for RHEL 9 on alibaba cloud (e.g via our mail communication) as the last time.

pirat89: @jinkkkang changes in repomap.json included in PR #1165 . Once it's merged, this one needs to be rebased to cover up-to-date changes and can be merged then as well. I will do it right after the change.

pirat89:
@jinkkkang checking new repomap.json file based on data you we've got from you, I see differences between our set and your original one:
```
The following repos have been removed: 
 - Repo(pesid='rhel8-BaseOS', major_version='8', repoid='rhui-rhel-8-for-aarch64-baseos-rhui-rpms', repo_type='rpm', channel='ga', arch='aarch64', rhui='alibaba')
 - Repo(pesid='rhel8-AppStream', major_version='8', repoid='rhui-rhel-8-for-aarch64-appstream-rhui-rpms', repo_type='rpm', channel='ga', arch='aarch64', rhui='alibaba')
 - Repo(pesid='rhel8-rhui-custom-client-at-alibaba', major_version='8', repoid='rhui-custom-rhui_client_at_alibaba-rhel-8', repo_type='rpm', channel='ga', arch='aarch64', rhui='alibaba')
The following repos have been added: 
 - Repo(pesid='rhel9-Supplementary', major_version='9', repoid='rhui-rhel-9-for-x86_64-supplementary-rhui-rpms', repo_type='rpm', channel='ga', arch='x86_64', rhui='alibaba')
 - Repo(pesid='rhel9-CRB', major_version='9', repoid='rhui-codeready-builder-for-rhel-9-aarch64-rhui-rpms', repo_type='rpm', channel='ga', arch='aarch64', rhui='alibaba')
 - Repo(pesid='rhel9-Supplementary', major_version='9', repoid='rhui-rhel-9-for-aarch64-supplementary-rhui-rpms', repo_type='rpm', channel='ga', arch='aarch64', rhui='alibaba')
 - Repo(pesid='rhel9-HighAvailability', major_version='9', repoid='rhui-rhel-9-for-x86_64-highavailability-rhui-rpms', repo_type='rpm', channel='ga', arch='x86_64', rhui='alibaba')
 - Repo(pesid='rhel9-CRB', major_version='9', repoid='rhui-codeready-builder-for-rhel-9-x86_64-rhui-rpms', repo_type='rpm', channel='ga', arch='x86_64', rhui='alibaba')
```

Checking repofiles we have, I do not see any source of data for RHEL 8 aarch64, so the mapping for aarch64 does not exist for IPU 8 -> 9. Regarding the state, I decided to go in the way to merge this PR now. We will update the repomap.json file later prior the release if you share with us repo files for RHEL 8 RHUI on aarch64.

pirat89 (approved):
see my previous comment. merging

PR #1138, opened 2023-11-02
Do not create dangling symlinks for containerized RHSM

pirat89:
When setting RHSM into the container mode, we are creating symlinks to /etc/rhsm and /etc/pki/entitlement directories. However, this creates dangling symlinks if RHSM is not installed or user manually removes one of these dirs.

If any of these directories is missing, skip other actions and log the warning. Usually it means that RHSM is not actually used or installed at all, so in these cases we can do the skip. The only corner case when system could use RHSM without /etc/pki/entitlement is when RHSM is configured to put these certificate on a different path, and we do not support nor cover such a scenario as we are not scanning the RHSM configuration at all.

This also solves the problems on systems that does not have RHSM available at all.

Fixes #1136

### DIFF
--- a/repos/system_upgrade/common/libraries/rhsm.py
+++ b/repos/system_upgrade/common/libraries/rhsm.py
@@ -325,6 +325,11 @@ def set_container_mode(context):
     could be affected and the generated repo file in the container could be
     affected as well (e.g. when the release is set, using rhsm, on the host).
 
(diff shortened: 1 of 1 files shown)


pirat89:
teste manually:
* rhel 7 & 8 with rhsm
* rhel 7 & 8 without rhsm

evgeni:
Thank you! My CentOS boxen are much happier again :)

PR #1139, opened 2023-11-04
convert dot-less CentOS versions to X.999

evgeni:
CentOS Stream doesn't do dot releases, so it always has the major version in the version_id field of os-release.
Our code, in multiple places, expects versions in the form X.Y and doesn't cope well when given only X.
The value "999" is taken arbitrarily, with the sole assumption that it will be always higher than any RHEL release, as CentOS Stream is ahead of RHEL.

See https://issues.redhat.com/browse/CS-1757 for further discussion and alternative numbering options.


repos/system_upgrade/common/actors/ipuworkflowconfig/libraries/ipuworkflowconfig.py
```diff
@@ -50,12 +50,16 @@ def get_os_release(path):
     try:
         with open(path) as f:
             data = dict(l.strip().split('=', 1) for l in f.readlines() if '=' in l)
+            release_id = data.get('ID', '').strip('"')
+            version_id = data.get('VERSION_ID', '').strip('"')
+            if release_id == 'centos' and '.' not in version_id:
+                version_id = "{}.999".format(version_id)
```
fernflower, line 56: I am not too happy about 999 magic constant tbh :( 
Is there a universe where version_id is '' and leapp is still happy?

pirat89 (changes requested):
Honestly, I do not think this is the right solution for this. This should not be resolved by magical/crafted values that have nothing to do with the real system. This particular problem will need to be refink, together with the target os version - where this magic even cannot be used as it has effect on baseurls of target repositories if `$releasever` is present.

Please note that we do not have a capacity right now to think about the right solution, so it's possible we will need to postpone this for a time. But when we have a time, we could invite you to our team tech. mtg to discuss the solution if you are interested.

evgeni:
Yeah, I agree, this is not a nice solution (but it works :upside_down_face:)
When I tried making it "the other way" (= making the code accept a major-only version) I ended up with the fact that `(8,)` is sorted lower than `(8,0)` (or any other `(8, X)` and that confused other parts of the code that pick which packages are to be replaced etc.

I'll happily join the mtg to discuss this!

andrewlukoshko:
Just in case here are the commits we made in ELevate project to fix this CentOS version issue:
https://github.com/AlmaLinux/leapp-repository/commit/f95a819d6b3a773eea627f01693ee77fb091be17
https://github.com/AlmaLinux/leapp-repository/commit/8ea0f2b0d7beb39f50cbcd94c4b186a0700bcc20
https://github.com/AlmaLinux/leapp-repository/commit/96dbd35128163c738e819e632c359c85b907358f

But this PR looks cleaner.

pirat89:
It has been covered recently by #1363 . Closing

PR #1141, opened 2023-11-07
Introduce the checkmptctl actor

bessonc:
Raise an inhibitor if the mptctl is loaded and in use, and reports the userspace processes which use it.
The suggested patch is not ideal and could be refactored as it adds another inhibitor for the same problem reported by the CheckDetectedDevicesAndDrivers actor. This is essentially to share the logic in pure python, without any additional requirement.

Fixes Red Hat issue #RHEL-15894

Outputs:
```
    1. Inhibitor: Leapp detected loaded kernel drivers which have been removed in RHEL 8. Upgrade cannot proceed.
    2. Inhibitor: Detected processes that are using mptctl.
```
```
Risk Factor: high (inhibitor)
Title: Detected processes that are using mptctl.
Summary: Leapp detected that the following processes are using the mptctl kernel module which will be removed in RHEL 8:
20914: python -c f=open('/dev/mptctl','r');import time;time.sleep(300) 
14348: python -c f=open('/dev/mptctl','r');import time;time.sleep(300) 
Remediation: [hint] Stop this application or the service that runs this process.
Key: 66e1a0da39c67d3495587ed2d1bd8e810e41288d
```

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkmptctl/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkmptctl/actor.py
@@ -1,0 +1,19 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.checkmptctl import check_mptctl
+from leapp.models import ActiveKernelModulesFacts
+from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckMptctl(Actor):
+    """
+    Check if a process uses the mptctl kernel module that will be removed.
+    It could prevent the unloading of this module which does not handle HW.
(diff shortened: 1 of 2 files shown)


pirat89 (changes requested):
This change is connected to RHEL bug ticket, so for now I set the blocker for merge as it cannobe merged until we have agreement with QE that this change could be added or not to the release. Also not sure about the solution, we will get to it later. This ticket has not been reviewed yet.

mkluson:
Hi there, as we agreed in the bug (https://issues.redhat.com/browse/RHEL-15894) this issue will be addressed by KCS. Closing this PR

PR #1142, opened 2023-11-07
Add RHUI configuration

abadger:
Add configuration options for leapp's RHUI handling and RPM transactions.

## RHUI Configuration
Example of RHUI configuration:
```yaml
rhui:
  use_config : True
  source_clients: [ rh-amazon-rhui-client ]
  target_clients: [ rh-amazon-rhui-client ]
  cloud_provider: aws
  upgrade_files:
    /usr/share/leapp-repository/repositories/system_upgrade/common/files/rhui/aws/leapp-aws.repo : /etc/yum.repos.d/
    /usr/share/leapp-repository/repositories/system_upgrade/common/files/rhui/aws/rhui-client-config-server-9.crt : /etc/pki/rhui/product/
    /usr/share/leapp-repository/repositories/system_upgrade/common/files/rhui/aws/rhui-client-config-server-9.key : /etc/pki/rhui/
    /usr/share/leapp-repository/repositories/system_upgrade/common/files/rhui/aws/content-rhel9.crt : /etc/pki/rhui/product/
    /usr/share/leapp-repository/repositories/system_upgrade/common/files/rhui/aws/content-rhel9.key : /etc/pki/rhui/
    /usr/share/leapp-repository/repositories/system_upgrade/common/files/rhui/aws/cdn.redhat.com-chain.crt : /etc/pki/rhui/product/
  rhui_target_repositories_to_use:
    - rhel-9-appstream-rhui-rpms
    - rhel-9-baseos-rhui-rpms

```
Leapp reads the configuration, checking the `use_config` field to know whether the values from config should be used, instead of leapp's decision routines. If `use_config=True`, leapp will not do *any* automatic decisions about RHUI whatsoever. Therefore, the configured values determine what leapp will do.

The following configuration options are added:
| Configuration field | Description |
| ----    | ---- |
| `source_clients` |  RHUI clients to uninstall during the upgrade | 
| `target_clients` |  RHUI clients to install during the upgrade  | 
| `cloud_provider` |  so that leapp knows it is running on cloud  | 
| `upgrade_files` |  mapping of paths of files on the source system to the corresponding destinations in the scratch container.  Once the target container is set up, the files provided by the target client are used and the files from `upgrade_files` are removed.
| `enabled_target_repositories` |  a list of repositories to enable. Functions the same way as using `leapp --enablerepo`  | 

#### Known limitations
Using a repofile withing `upgrade_files` that uses different repoids than the target client(s) in combination with `enable_target_repositories` causes DNF transaction check to fail.

Jira refs:
   - RHUI: https://issues.redhat.com/browse/RHEL-56251

Acceptance criteria for RHUI configuration:
- [x] It is possible to use config to upgrade unknown regions
- [x] Should make HA upgrades on AWS possible (achieved by a `leapp-rhui-aws-ha` that does not require)
- [x] Make `rhui_repositories` a required field (or rename to `target_repoids_to_enable`)
- [x] Make sure that repomapping does not error on unknown cloud providers.
  - Using `cloud_provider = 'some_random_provider` has been tried (with `enabled_target_repositories`), there were no problems - no code modifications are required.
- [x] Remove defaults - require user to supply everything

### DIFF
--- a/commands/upgrade/__init__.py
+++ b/commands/upgrade/__init__.py
@@ -93,6 +93,9 @@ def upgrade(args, breadcrumbs):
     command_utils.set_resource_limits()
 
     workflow = repositories.lookup_workflow('IPUWorkflow')(auto_reboot=args.reboot)
+
+    command_utils.load_actor_configs_and_store_it_in_db(context, repositories, cfg)
+
     util.process_whitelist_experimental(repositories, workflow, configuration, logger)
     util.warn_if_unsupported(configuration)
     with beautify_actor_exception():
--- a/repos/system_upgrade/common/configs/rhui.py
+++ b/repos/system_upgrade/common/configs/rhui.py
@@ -1,0 +1,127 @@
+"""
+Configuration keys for RHUI.
+
+In case of RHUI in private regions it usual that publicly known RHUI data
+is not valid. In such cases it's possible to provide the correct expected
+RHUI data to correct the in-place upgrade process.
+"""
+
+from leapp.actors.config import Config
+from leapp.models import fields
+
+RHUI_CONFIG_SECTION = 'rhui'
+
+
+# @Note(mhecko): We use to distinguish config instantiated from default values that we should ignore
+# #              Maybe we could make all config values None and detect it that way, but then we cannot
+# #              give the user an example how the config should look like.
+class RhuiUseConfig(Config):
+    section = RHUI_CONFIG_SECTION
+    name = "use_config"
+    type_ = fields.Boolean()
+    default = False
+    description = """
+        Use values provided in the configuration file to override leapp's decisions.
+    """
+
+
+class RhuiSourcePkgs(Config):
+    section = RHUI_CONFIG_SECTION
+    name = "source_clients"
+    type_ = fields.List(fields.String())
+    default = []
+    description = """
+        The name of the source RHUI client RPMs (to be removed from the system).
+    """
+
+
+class RhuiTargetPkgs(Config):
+    section = RHUI_CONFIG_SECTION
+    name = "target_clients"
+    type_ = fields.List(fields.String())
+    default = []
+    description = """
+        The name of the target RHUI client RPM (to be installed on the system).
+    """
+
+
+class RhuiCloudProvider(Config):
+    section = RHUI_CONFIG_SECTION
+    name = "cloud_provider"
+    type_ = fields.String()
+    default = ""
+    description = """
+        Cloud provider name that should be used internally by leapp.
+
+        Leapp recognizes the following cloud providers:
+            - azure
+            - aws
+            - google
+
+        Cloud provider information is used for triggering some provider-specific modifications. The value also
+        influences how leapp determines target repositories to enable.
+    """
+
+
+# @Note(mhecko): We likely don't need this. We need the variant primarily to grab files from a correct directory
+# in leapp-rhui-<provider> folders.
+class RhuiCloudVariant(Config):
+    section = RHUI_CONFIG_SECTION
+    name = "image_variant"
+    type_ = fields.String()
+    default = "ordinary"
+    description = """
+        RHEL variant of the source system - is the source system SAP-specific image?
+
+        Leapp recognizes the following cloud providers:
+            - ordinary    # The source system has not been deployed from a RHEL with SAP image
+            - sap         # RHEL SAP images
+            - sap-apps    # RHEL SAP Apps images (Azure only)
+            - sap-ha      # RHEL HA Apps images (HA only)
+
+        Cloud provider information is used for triggering some provider-specific modifications. The value also
+        influences how leapp determines target repositories to enable.
+
+        Default:
+            "ordinary"
+    """
+
+
+class RhuiUpgradeFiles(Config):
+    section = RHUI_CONFIG_SECTION
+    name = "upgrade_files"
+    type_ = fields.StringMap(fields.String())
+    default = dict()
+    description = """
+        A mapping from source file paths to the destination where should they be
+        placed in the upgrade container.
+
+        Typically, these files should be provided by leapp-rhui-<PROVIDER> packages.
+
+        These files are needed to facilitate access to target repositories. Typical examples are: repofile(s),
+        certificates and keys.
+    """
+
+
+class RhuiTargetRepositoriesToUse(Config):
+    section = RHUI_CONFIG_SECTION
+    name = "rhui_target_repositories_to_use"
+    type_ = fields.List(fields.String())
+    description = """
+        List of target repositories enabled during the upgrade. Similar to executing leapp with --enablerepo.
+
+        The repositories to be enabled need to be either in the repofiles listed in the `upgrade_files` field,
+        or in repofiles present on the source system.
+    """
+    default = list()
+
+
+all_rhui_cfg = (
+    RhuiTargetPkgs,
+    RhuiUpgradeFiles,
+    RhuiTargetRepositoriesToUse,
+    RhuiCloudProvider,
+    RhuiCloudVariant,
+    RhuiSourcePkgs,
+    RhuiUseConfig
+)
--- a/commands/command_utils.py
+++ b/commands/command_utils.py
@@ -1,10 +1,12 @@
+import hashlib
 import json
 import os
 import re
 import resource
 
+from leapp.actors import config as actor_config
 from leapp.exceptions import CommandError
-from leapp.utils import path
+from leapp.utils import audit, path
 
 HANA_BASE_PATH = '/hana/shared'
 HANA_SAPCONTROL_PATH_X86_64 = 'exe/linuxx86_64/hdb/sapcontrol'
@@ -178,3 +180,31 @@ def set_resource_limits():
 
     if soft_fsize != fsize_limit:
         set_resource_limit(resource.RLIMIT_FSIZE, fsize_limit, fsize_limit)
+
+
+def load_actor_configs_and_store_it_in_db(context, repositories, framework_cfg):
+    """
+    Load actor configuration so that actor's can access it and store it into leapp db.
+
+    :param context: Current execution context
+    :param repositories: Discovered repositories
+    :param framework_cfg: Leapp's configuration
+    """
+    # Read the Actor Config and validate it against the schemas saved in the
+    # configuration.
+
+    actor_config_schemas = tuple(actor.config_schemas for actor in repositories.actors)
+    actor_config_schemas = actor_config.normalize_schemas(actor_config_schemas)
+    actor_config_path = framework_cfg.get('actor_config', 'path')
+
+    # Note: actor_config.load() stores the loaded actor config into a global
+    # variable which can then be accessed by functions in that file.  Is this
+    # the right way to store that information?
+    actor_cfg = actor_config.load(actor_config_path, actor_config_schemas)
+
+    # Dump the collected configuration, checksum it and store it inside the DB
+    config_text = json.dumps(actor_cfg)
+    config_text_hash = hashlib.sha256(config_text.encode('utf-8')).hexdigest()
+    config_data = audit.ActorConfigData(config=config_text, hash_id=config_text_hash)
+    db_config = audit.ActorConfig(config=config_data, context=context)
+    db_config.store()
--- a/commands/preupgrade/__init__.py
+++ b/commands/preupgrade/__init__.py
@@ -62,6 +62,9 @@ def preupgrade(args, breadcrumbs):
     command_utils.set_resource_limits()
 
     workflow = repositories.lookup_workflow('IPUWorkflow')()
+
+    command_utils.load_actor_configs_and_store_it_in_db(context, repositories, cfg)
+
     util.warn_if_unsupported(configuration)
     util.process_whitelist_experimental(repositories, workflow, configuration, logger)
     with beautify_actor_exception():
--- a/packaging/leapp-repository.spec
+++ b/packaging/leapp-repository.spec
@@ -250,6 +250,11 @@ install -m 0755 -d %{buildroot}%{_sysconfdir}/leapp/files/
 install -m 0644 etc/leapp/transaction/* %{buildroot}%{_sysconfdir}/leapp/transaction
 install -m 0644 etc/leapp/files/* %{buildroot}%{_sysconfdir}/leapp/files
 
+# Actor configuration dir
+install -m 0755 -d %{buildroot}%{_sysconfdir}/leapp/actor_conf.d/
+# uncomment to install existing configs
+#install -m 0644 etc/leapp/actor_conf.d/* %%{buildroot}%%{_sysconfdir}/leapp/actor_conf.d
+
 # install CLI commands for the leapp utility on the expected path
 install -m 0755 -d %{buildroot}%{leapp_python_sitelib}/leapp/cli/
 cp -r commands %{buildroot}%{leapp_python_sitelib}/leapp/cli/
@@ -295,6 +300,8 @@ done;
 %dir %{custom_repositorydir}
 %dir %{leapp_python_sitelib}/leapp/cli/commands
 %config %{_sysconfdir}/leapp/files/*
+# uncomment to package installed configs
+#%%config %%{_sysconfdir}/leapp/actor_conf.d/*
 %{_sysconfdir}/leapp/repos.d/*
 %{_sysconfdir}/leapp/transaction/*
 %{repositorydir}/*
--- a/packaging/leapp-repository.spec
+++ b/packaging/leapp-repository.spec
@@ -120,7 +120,7 @@ Requires:       leapp-repository-dependencies = %{leapp_repo_deps}
 
 # IMPORTANT: this is capability provided by the leapp framework rpm.
 # Check that 'version' instead of the real framework rpm version.
-Requires:       leapp-framework >= 5.0, leapp-framework < 6
+Requires:       leapp-framework >= 6.0, leapp-framework < 7
 
 # Since we provide sub-commands for the leapp utility, we expect the leapp
 # tool to be installed as well.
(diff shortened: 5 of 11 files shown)


repos/system_upgrade/common/configs/rpmtransactionconfigtaskscollector.py
```diff
@@ -0,0 +1,44 @@
...
+    type_ = fields.List(fields.String(), default=[
+        "leapp",
+        "python2-leapp",
+        "python3-leapp",
+        "leapp-repository",
+        "snactor",
+    ])
+    description = """
+        List of packages to be kept in the upgrade transaction.
+    """
```
pirat89, line 32: Just trying to think why it cannot be same as with models? I mean:
```suggestion
class Transaction_ToKeep(Config):
    section = "transaction"
    to_keep = fields.List(fields.String(), default=[
        "leapp",
        "python2-leapp",
        "python3-leapp",
        "leapp-repository",
        "snactor",
    ])
    """
        List of packages to be kept in the upgrade transaction.
    """
```

abadger: Since we've settled on yaml as the actual file system format, I think that loading more information into fields will be fine (default and type_).  to_keep and the docstring treatment were things I didn't know about.  I think they'll be fine too.

One ramification of relying on fields (I think in any incarnation) is that we should allow mapping values.  I heard that those were specifically excluded from the present fields because they were considered too complex and unnecessary for Models.  I do not think those are true for config values.  Yaml has dictionaries as basic data types and in cases where you want to group record-style information, yaml mappings make the most sense.
``` yaml
hosts:
    balancer.example.com: 192.168.100.2
    git.example.com: 192.168.100.3
```

I don't know if we'll want to keep field.Mapping (or field.Dict)  from being used in models or not care now that we have a need for it for a different purpose.

pirat89: @abadger  I was thinking about that during weekend, as I think I misunderstood you originally after reading this your comment and the example. I thought that this problem is addressed by [fields.JSON](https://github.com/oamg/leapp/blob/master/leapp/models/fields/__init__.py#L509) which we can allow for configs to not be so strict. However, thinking about that, I think you provided totally valid point about having something like a `Map`/`Mapping`/`Dict` field, which could have variable keys (like hosts, where we do not know number of such items, nor specific keys itself). But we could on the other hand still define what will be values like in case of other fields. So the use for the example above, I could imagine to allow something like:

```
class MyConfig(Config):
    hosts = fields.Mapping(fields.String)  # or e.g. fields.Mapping(value_type=fields.String)
```
Where we can define what will be values, hence the key is expected to be always string (no sure whether we would need anything else for keys than string). For the completness, I can imagine we could resolve it also differently, but usually it would lead to some magical tweaks around serialisation/deserialisation, or little bit worse UX, e.g. to require user to write config like:

```
hosts:
    - hostname: balancer.example.com
    - ip: 192.168.100.3
```

Which is doable and still kind of reasonable, but thinking about the trade-off, it make sense to me to be less strict here.

@abadger  Do I understand correctly your idea?

---
Just to provide context why we usually do not allow [fields.JSON](https://github.com/oamg/leapp/blob/master/leapp/models/fields/__init__.py#L509) in Models inside our official repositories. It's because the JSON field cannot be verified for the content. From the point, we can verify the structure is JSON, but we cannot verify it contains particular fields / content inside, so the structure can vary and we couldn't guarantee that part of this structure will be always a specific key/item (due to bugs, or changes in actor that produce a particular msg, so some other actors - e.g. custom ones - could start to be broken without clean visibility of the problem). Having strict structure in models, provides clear definition of the content and people can easily see it.

commands/upgrade/__init__.py
```diff
@@ -90,6 +91,17 @@ def upgrade(args, breadcrumbs):
...
+
+    # Read the Actor Config and validate it against the schemas saved in the
+    # configuration.
+    actor_config_schemas = tuple(actor.config_schemas for actor in repositories.actors)
+    actor_config_schemas = actor_config.normalize_schemas(actor_config_schemas)
+    actor_config_path = cfg.get('actor_config', 'path')
+    # Note: actor_config.load() stores the loaded actor config into a global
+    # variable which can then be accessed by functions in that file.  Is this
+    # the right way to store that information?
+    actor_config.load(actor_config_path, actor_config_schemas)
```
dkubek, line 103: I understand that we are doing it this way because we want to load the configs before anything else is done. However, this leads to the problem that whether config is loaded is not within the power of the framework. Instead of having it here in the command, wouldn't it be better  to have ``load()`` in some early stage of ``workflow.run()``? That way we can be sure that configs are loaded from the point of view of the framework

dkubek: Workflow should collect all discovered actors during its initialization (https://github.com/oamg/leapp/blob/7565b5c8cd5a098522565ff0ed5278eb2b04b7a8/leapp/workflows/__init__.py#L167). I have used this before and thought it might be helpful again here.

To me, loading the configuration just after workflow starts and before any actor is processed is a place where I would put it personally. However, there might be reasons not to do this if it has not come up in a discussion. Have you thought about this? @pirat89 Or am I missing something

pirat89: Not sure - I mean, we looked for a place where we could put it and that seemed good - especially in case we would like to have in future repository-shared configs (configs dir defined in the repository level). We already have such a handling in case of answerfiles, but you are right that having the possibility to keep it implemented just inside the framework side (without the need to know additional magic that should be specified in CLI commands on the repositoriy side) sounds much better to me. Also, remember on snactor! We need a way how to execute an actor itself with snactor without the need to run the workflow. (it can be done later, but it should be done still for the upcoming version)

abadger: @dkubek The code you're pointing to looks like it just looks at the actors available in all the phases that the workflow runs rather than all of the actors in the repositories?

Although I'm not sure if we need all of them or not.
* If we get all the actors, we can warn the user that there is unrecognized configuration.   This will catch typos and configuration of old versions which is no longer valid.
* If we only get the actors used by the workflow, then we wll have false positives if we emit warnings for that (so I would probably not emit a warning on unrecognized config entries).

abadger: (Also, I agree with you that if we can put this into framework code, it is the better place).

Rezney:
Got this when testing:

```python
2024-10-04 15:12:10.758 DEBUG    PID: 31702 leapp.repository.system_upgrade_common: Starting actor discovery in actors/cloud/checkhybridimage
2024-10-04 15:12:10.775 DEBUG    PID: 31702 leapp.repository.system_upgrade_common: Starting actor discovery in actors/cloud/checkrhui
Process Process-37:
    self.run()
  File "/usr/lib/python3.6/site-packages/leapp/repository/actor_definition.py", line 32, in inspect_actor
    definition.load()
  File "/usr/lib/python3.6/site-packages/leapp/repository/actor_definition.py", line 199, in load
    self._module = load_module(importer, name)
  File "/usr/lib/python3.6/site-packages/leapp/compat.py", line 86, in load_module
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/etc/leapp/repos.d/system_upgrade/common/actors/cloud/checkrhui/actor.py", line 2, in <module>
    from leapp.configs.common.rhui import all_rhui_cfg
ModuleNotFoundError: No module named 'leapp.configs.common'
2024-10-04 15:12:10.788 ERROR    PID: 31702 leapp.repository.system_upgrade_common: Process inspecting actor in actors/cloud/checkrhui failed with 1
```

etc/leapp/actor_conf.d/rhui.yml
```diff
@@ -0,0 +1,12 @@
+rhui:
+  use_config : True
+  source_clients: [ rh-amazon-rhui-client ]
+  target_clients: [ rh-amazon-rhui-client ]
+  cloud_provider: aws
+  upgrade_files:
```
bparry02, line 6: This example is looking good! I also know of some use cases that need to specify the CA certificate used for the dnf calls to the RHUI repo, but perhaps that configuration is encapsulated in the leapp-aws.repo file...?

MichalHe: Do you mean the `sslcacert`, `sslclientcert` and `sslclientkey` fields in repository definition? If yes, then yes, this is doable with the proposed configuration. You just specify all the `ssl*` fields in a repofile provided in `upgrade_files` as you would normally do. Then you need to list the certificates/keys also in `upgrade_files`, together with an information on where to place them. For example, your repofile `/root/my-upgrade-repos.repo` would contain a repository with:
```
sslclientcert=/etc/pki/rhui/product/content-rhel8.crt
```
Then you need tell leapp that it should grab the `content-rhel8.crt` somewhere from the source system and place it at `/etc/pki/rhui/product/content-rhel8.crt`.

The configuration would then look like:
```yaml
rhui:
  use_config : True
  source_clients: [  rhel8-client ]   # what client to remove from the source system
  target_clients: [ rhel9-client ]  # what client should be installed during the upgrade
  cloud_provider: aws  # for leapp to use internally 
  upgrade_files:
    /root/my-upgrade-repos.repo : /etc/yum.repos.d  # during upgrade, place this repofile into /etc/yum.repos.d
    /root/content-rhel8.crt : /etc/pki/rhui/product/  # during the upgrade, place this key here. Same path as in sslclientcert in the repo
   #  .... other keys/certs mentioned in the repofile, or otherwise important for leapp to access target OS content
```
This should enable basically anyone running a RHUI system to upgrade, assuming that provider's rhui clients do not use some eccentric python dependencies (although that should be fine too, I just haven't tested these setups).

You get all the necessary `upgrade_files` by spawning a RHEL VM running the target OS, and copying the files, keys and certs provided by the RHUI client. There are also other ways, but this is likely the simplest one. If you are upgrading an entire fleet, you naturally need to spawn the target machine only once.

Also note that if you rename the repositories in the repofile for some reason, and then use these new names in `enabled_target_repositories`, the DNF transaction check leapp does might fail. Therefore, use the same repoids as the target client uses and everything should be fine. If you would like to test this, please ping me with any questions. It is always nice to get feedback before we merge this :grinning:.

Rezney:
Got:

~~~python
2024-10-10 06:10:11.407 INFO     PID: 52881 leapp.workflow.configuration_phase: Starting stage Before of phase configuration_phase
2024-10-10 06:10:11.408 INFO     PID: 52881 leapp.workflow.configuration_phase: Starting stage Main of phase configuration_phase
2024-10-10 06:10:11.409 INFO     PID: 52881 leapp.workflow.configuration_phase: Executing actor ipu_workflow_config 
Process Process-202:
    self.run()
  File "/usr/lib/python3.6/site-packages/leapp/repository/actor_definition.py", line 73, in _do_run
    skip_dialogs=skip_dialogs)
  File "/usr/lib/python3.6/site-packages/leapp/actors/__init__.py", line 131, in __init__
    self.config = retrieve_config(self.config_schemas)
NameError: name 'retrieve_config' is not defined
2024-10-10 06:10:11.453 ERROR    PID: 52881 leapp.workflow.configuration_phase: Actor ipu_workflow_config has crashed: None
~~~

Tested with:
~~~
leapp-0.18.0-0.20241008065723373633.pr870.16.gb993556.el8.noarch
leapp-rhui-azure-1.0.0-14.el8.noarch
leapp-upgrade-el8toel9-0.21.0-0.20241009173513741982.pr1142.16.g13508f65.el8.noarch
~~~

@MichalHe @dkubek

dkubek:
@Rezney My bad. Should be fixed now in leapp

matejmatuska:
I performed an upgrade from RHEL 8.10 SAP to RHEL 9.4 SAP e4s on AWS and everything went okay, however the the custom modifications detaction actor reports that config has been modified, not sure if you are aware of this :):
```
Risk Factor: high
Title: Detected modified files of the in-place upgrade tooling.
Summary: We have detected that some files of the tooling processing the in-place upgrade have been modified. Note that such modifications can be allowed only after consultation with Red Hat - e.g. when support suggests the change to resolve discovered problem. If these changes have not been approved by Red Hat, the in-place upgrade is unsupported.
Following files have been modified:
    - /etc/leapp/actor_conf.d/rhui.yml
Remediation: [hint] To restore original files reinstall related packages.
Key: 5532a4fe27dc0b05de1e9e77bda407ea47ad6971
```

bparry02:
I was able to upgrade RHEL 8 to 9 x86 successfully using a personal RHUI and custom source and target client RPM names. Well done team!

```
leapp-0.18.0-0.20241020152857414426.pr870.2.ge2f13b4.el8.noarch
leapp-deps-el9-5.0.9-0.202410231928Z.356a5b37.pr_1142.el9.noarch
leapp-repository-deps-el9-5.0.9-0.202410231928Z.356a5b37.pr_1142.el9.noarch
leapp-upgrade-el8toel9-0.21.0-0.20241023192831897106.pr1142.34.g356a5b37.el8.noarch
python3-leapp-0.18.0-0.20241020152857414426.pr870.2.ge2f13b4.el8.noarch
```

I did try RHEL 7to8 as well, but ran into the glob error mentioned in the description of leapp PR 870.

pirat89:
added a squash commit modifying %files section covering correct mark of configuration files in the actor_conf.d directory and making the directory part of the rpm, to fix:
```
# rpm -qf /etc/leapp/actor_conf.d/
file /etc/leapp/actor_conf.d is not owned by any package
```

repos/system_upgrade/common/configs/rhui.py
```diff
@@ -0,0 +1,126 @@
...
+
+class RhuiSourcePkgs(Config):
+    section = RHUI_CONFIG_SECTION
+    name = "source_clients"
+    type_ = fields.List(fields.String())
+    default = []
+    description = """
+        The name of the source RHUI client RPMs (to be removed from the system).
+        Default: rhui
+    """
```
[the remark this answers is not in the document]
MichalHe: Thank you for noticing, should be resolved in the latest version.

repos/system_upgrade/common/configs/rhui.py
```diff
@@ -0,0 +1,126 @@
...
+        The name of the target RHUI client RPM (to be installed on the system).
+        Default: rhui
+    """
+
+
+class RhuiCloudProvider(Config):
+    section = RHUI_CONFIG_SECTION
+    name = "cloud_provider"
+    type_ = fields.String(default="rhui")
+    default = ""
```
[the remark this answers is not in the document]
MichalHe: Should be resolved.

repos/system_upgrade/common/configs/rhui.py
```diff
@@ -0,0 +1,126 @@
...
+    description = """
+        Cloud provider name that should be used internally by leapp.
+
+        Leapp recognizes the following cloud providers:
+            - azure
+            - aws
+            - google
+
+        Cloud provider information is used for triggering some provider-specific modifications. The value also
+        influences how leapp determines target repositories to enable.
```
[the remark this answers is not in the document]
MichalHe: In the case of RHUI config, there are no defaults - everything should be empty by default so that we can detect and require all of the config values to be provided by the user. The reason for no defaults is tha we cannot provide any reasonable ones - almost everything is different from provider to provider.

pirat89:
tested manually for IPU 7 -> 8 and 8 -> 9 on RHUI AWS.

Note, config for IPU 7 -> 8 on AWS looks like this:
```yaml
rhui:
  use_config : True
  source_clients: [ rh-amazon-rhui-client ]
  target_clients: [ rh-amazon-rhui-client ]
  cloud_provider: aws
  upgrade_files:
    /usr/share/leapp-repository/repositories/system_upgrade/common/files/rhui/aws/amazon-id.py: /usr/lib/python2.7/site-packages/dnf-plugins/
    /usr/share/leapp-repository/repositories/system_upgrade/common/files/rhui/aws/leapp-aws.repo : /etc/yum.repos.d/
    /usr/share/leapp-repository/repositories/system_upgrade/common/files/rhui/aws/cdn.redhat.com-chain.crt : /etc/pki/rhui/product/
    /usr/share/leapp-repository/repositories/system_upgrade/common/files/rhui/aws/content-rhel8.crt : /etc/pki/rhui/product/
    /usr/share/leapp-repository/repositories/system_upgrade/common/files/rhui/aws/content-rhel8.key : /etc/pki/rhui
    /usr/share/leapp-repository/repositories/system_upgrade/common/files/rhui/aws/rhui-client-config-server-8.crt : /etc/pki/rhui/product/
    /usr/share/leapp-repository/repositories/system_upgrade/common/files/rhui/aws/rhui-client-config-server-8.key : /etc/pki/rhui
  rhui_target_repositories_to_use:
    - rhel-8-appstream-rhui-rpms
    - rhel-8-baseos-rhui-rpms
```

repos/system_upgrade/common/configs/rhui.py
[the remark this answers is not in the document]
pirat89: @MichalHe that makes sense.

pirat89:
@MichalHe  I've drop the tmp commits. Also I made one small change. I'Ve added etc/leapp/actor_confid.d dir to the repo (with `.gitkeep` file) and ensure that if someone adds a file into the repo e.g. because we decide to have the config part of the rpm (livemode one?) it's automatically part of the rpm. (iow keeping the change from one of removed commits).

Also added a commit that ensures that all .gitkeep files are deleted when building rpm. so they are not installed then.

pirat89 (dismissed):
lgtm. I made some small changes (see my comment above). It's prepared for the merge, but let's wait for finished tests. Note that one of 8 -> 9 tests seems to be broken, so if the "rhsm" related test is the only one that's failing, it's ok to be merged.

pirat89 (approved):

etc/leapp/actor_conf.d/rhui.yml
pirat89: @bparry02 just a headsup for this small change https://github.com/oamg/leapp/pull/892 - we decided to simplify it and we will require `.yaml` extension in future.

PR #1144, opened 2023-11-10
Switch from plan name regex to filter by tags

fernflower:
Necessary work to adapt upstream tests to big refactoring changes brought by MR303.


pirat89 (approved):
merging as tests are passing.

PR #1145, opened 2023-11-14
Add 79to88 and 79to89 aws upgrade paths + 79to810 and 810to94

fernflower:
Thanks to the detailed downstream review by mmoran I have realised that upstream upgrade paths have to be revised and updated as well.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -123,7 +123,7 @@ jobs:
   targets:
     epel-7-x86_64:
       distros: [RHEL-7.9-rhui]
-  identifier: sanity-7to8-aws-e2e
+  identifier: sanity-7.9to8.6-aws-e2e
   # NOTE(ivasilev) Unfortunately to use yaml templates we need to rewrite the whole tf_extra_params dict
   # to use plan_filter (can't just specify one section test.tmt.plan_filter, need to specify environments.* as well)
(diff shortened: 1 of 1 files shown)


fernflower:
I guess this one will also be blocked on infra issues :(

.packit.yaml
```diff
@@ -487,7 +505,7 @@ jobs:
   targets:
     epel-8-x86_64:
       distros: [RHEL-8.6-rhui]
-  identifier: sanity-8to9-aws-e2e
+  identifier: sanity-8.6to9.0-aws-e2e
```
pirat89, line 508: @fernflower  @mm0ran  I guess it should be set 8.8to9.2 instead, as 8.6 to 9.0 is going to be dropped (I already wanted to drop it) from the enabled upgrade paths and we do not consider any other development related to this upgrade path, hence no testing.

The same about the `sanity-79to86-aws`, use rather 79to88. Thinking about that, 8.9 and 9.3 are outside. We want to drop these also, but I can keep them till Feb if wanted.

pirat89: Update: we are for now keeping also 8.6 & 9.0 upgrade paths until tests are updated.

pirat89 (approved):

PR #1149, opened 2023-11-16
adjustlocalrepos: suppress unwanted deprecation report

pirat89:
SSIA


pirat89:
Merging. I've tested it manually.

PR #1150, opened 2023-11-20
BZ#2250254 - force removal of tomcat during the upgrade

evgeni:
We need pki-servlet-engine, which we depend on, but tomcat conflicts with.

jira:  SAT-21362


pirat89:
lgtm. waiting for info about results of  manual testing.

evgeni:
@pirat89 thanks! I've verified this now to fix the problem described in the BZ/Jira. Feel free to merge if you're happy with it :)

pirat89 (approved):

pirat89:
merging \o thanks!

PR #1151, opened 2023-11-20
Utilize get_target_major_version in no enabled target repositories report

andywaltlova:
Added the link to shortener, now we have both:

- https://red.ht/preparing-for-upgrade-to-rhel8
- https://red.ht/preparing-for-upgrade-to-rhel9

Disclaimer: I did not test it works, just kinda wanted to improve it when I saw the todo, feel free to close it if the fix required is more complex :)

### DIFF
--- a/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
+++ b/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
@@ -678,8 +678,7 @@ def _get_rhsm_available_repoids(context):
 
             ).format(target_major_version)),
             reporting.ExternalLink(
-                # TODO: How to handle different documentation links for each version?
-                url='https://red.ht/preparing-for-upgrade-to-rhel8',
+                url='https://red.ht/preparing-for-upgrade-to-rhel{}'.format(target_major_version),
                 title='Preparing for the upgrade')
             ])
         raise StopActorExecution()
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -678,8 +678,7 @@ def _get_rhsm_available_repoids(context):
 
             ).format(target_major_version)),
             reporting.ExternalLink(
-                # TODO: How to handle different documentation links for each version?
-                url='https://red.ht/preparing-for-upgrade-to-rhel8',
+                url='https://red.ht/preparing-for-upgrade-to-rhel{}'.format(target_major_version),
```
pirat89, line 681: @andywaltlova  thanks for the contribution! it makes sense :) Just thinking that having formatted URL string is a little bit problematic as we need from time to time grep/search where we are using particular URLs. Could you update the solution in the way we use one or the other URL? like
```
URL_PREP_EL8=...
URL_PREP_EL9=...

def _get_prep_url():
  if target_major_version == '8':
    return URL_PREP_EL8
  else:
....
...
# and than e.g.
url=_get_prep_url()
....
```

I am already thinking about URL for upgrades to rhel 10, so `url=url8 if target... == '8' else url9` would be not so practical.

I would be ok also with solution to note expended URLs just in comments and keep the code as it is.

andywaltlova: I think the comment sounds better, and even for grep, I think it makes sense to keep formatted url, because then new major doesn't require code change, just creation of new link with same shortenedformat :) .. when I was thinking about it I though someone would just grep it without the version number or thats what I would do with such url.

Anyway, I added rhel8,9 and 10 specific urls to the comments above both uses of the formatted url!

pirat89:
@andywaltlova thanks! seems good to me.merging

pirat89 (approved):

PR #1153, opened 2023-11-23
pes_events_scanner: Ignore Leapp related PES events

matejmatuska:
When PES events are added for all the Leapp related packages, we need to make sure to ignore them in `pes_events_scanner` to make sure they are *not* taken into account during the RPM upgrade transaction as we don't want to upgrade them or want to handle their upgrade in a different way.

Jira: OAMG-5645

### DIFF
--- a/repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
+++ b/repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
@@ -5,6 +5,7 @@ from leapp import reporting
 from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.actor import peseventsscanner_repomap
 from leapp.libraries.actor.pes_event_parsing import Action, get_pes_events, Package
+from leapp.libraries.common import rpms
 from leapp.libraries.common.config import version
 from leapp.libraries.stdlib import api
 from leapp.libraries.stdlib.config import is_verbose
@@ -481,11 +482,9 @@ def apply_transaction_configuration(source_pkgs):
 
 
 def remove_leapp_related_events(events):
-    leapp_pkgs = [
-        'leapp', 'leapp-deps', 'leapp-upgrade-el7toel8', 'leapp-upgrade-el8toel9',
-        'leapp-upgrade-el7toel8-deps', 'leapp-upgrade-el8toel9-deps', 'python2-leapp',
-        'python3-leapp', 'snactor'
-    ]
+    # NOTE(ivasilev) Need to revisit this once rhel9->rhel10 upgrades become a thing
+    leapp_pkgs = rpms.get_leapp_dep_packages(
+            major_version=['7', '8']) + rpms.get_leapp_packages(major_version=['7', '8'])
     res = []
     for event in events:
         if not any(pkg.name in leapp_pkgs for pkg in event.in_pkgs):
--- a/repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
+++ b/repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
@@ -480,6 +480,21 @@ def apply_transaction_configuration(source_pkgs):
     return source_pkgs_with_conf_applied
 
 
+def remove_leapp_related_events(events):
+    leapp_pkgs = [
+        'leapp', 'leapp-deps', 'leapp-upgrade-el7toel8', 'leapp-upgrade-el8toel9',
+        'leapp-upgrade-el7toel8-deps', 'leapp-upgrade-el8toel9-deps', 'python2-leapp',
+        'python3-leapp', 'snactor'
+    ]
+    res = []
+    for event in events:
+        if not any(pkg.name in leapp_pkgs for pkg in event.in_pkgs):
+            res.append(event)
+        else:
+            api.current_logger().debug('Filtered out leapp related event, event id: {}'.format(event.id))
+    return res
+
+
 def process():
     # Retrieve data - installed_pkgs, transaction configuration, pes events
     events = get_pes_events('/etc/leapp/files', 'pes-events.json')
@@ -494,6 +509,7 @@ def process():
     # packages of the target system, so we can distinguish what needs to be repomapped
     repoids_of_source_pkgs = {pkg.repository for pkg in source_pkgs}
 
+    events = remove_leapp_related_events(events)
     events = remove_undesired_events(events, releases)
 
     # Apply events - compute what packages should the target system have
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
```diff
@@ -480,6 +480,21 @@ def apply_transaction_configuration(source_pkgs):
     return source_pkgs_with_conf_applied
 
 
+def remove_leapp_related_events(events):
+    leapp_pkgs = [
+        'leapp', 'leapp-deps', 'leapp-upgrade-el7toel8', 'leapp-upgrade-el8toel9',
+        'leapp-upgrade-el7toel8-deps', 'leapp-upgrade-el8toel9-deps', 'python2-leapp',
+        'python3-leapp', 'snactor'
+    ]
```
pirat89, line 488: https://github.com/oamg/leapp-repository/blob/d9af1f2a19ec3352a4eff596bcb13e7ad073d763/repos/system_upgrade/common/libraries/dnfconfig.py#L6 - we have this function in the shared libraries. I think it's better to use this function to get the list of packages so it's not defined on multiple places.  However, the set of packages is superset of the one got from the function, and the whole set is not expected to be obtained from that one function. So I am thinking about to introduce a new one, like `get_all_leapp_packages()` or a parameter to be able to obtain the whole set. The point is that we have already 3 situations when we need a list of our packages (usually not the full list). And seems to me that it would be great to have just a one place where it's defined. @fernflower  is dealing with it now also in PR #1148 .

@oamg/developers  any ideas?

MichalHe: Having a single place where leapp package names are defined is a way to go. Implementation-wise I would go with
```python3
def get_leapp_packages(kind: LeappPackageKind):
    if kind == LeappPackageKind.EVERY_IN_UNIVERSE:
        pass
    elif kind == LeappPackageKind.ALL_FOR_THIS_UPG:
        pass
```
where `LeappPackageKind` is an enum. This way it forces the caller to think about what leapp packages are needed, while the approach is extensible - the enum can always be extended with new "classes" of packages that are discovered to be needed in a situation. Having 2 separate functions is, imho, error-prone.

matejmatuska: We've already had an internal discussion about this on Cabal, there will be a follow up email.

pirat89: @MichalHe  in case you would like to take a look before I send the email (in the morning), check the cabal doc - mtg minutes. We operates now with something like `get_leapp_packages(..., <kind>=[...]`. We are discussing kind of the taxonomy, but enum class is the final way I would expect to go. just we do not know what will be particular entries - there will be voting on that on Thu in the cabal doc. proposals can be added during whole Wed :)

fernflower: Love the enum idea, I'd only use `package_type` instead of `kind`.

fernflower:
Rebase against master should solve test faiulres

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
```diff
@@ -481,11 +482,8 @@ def apply_transaction_configuration(source_pkgs):
 
 
 def remove_leapp_related_events(events):
-    leapp_pkgs = [
-        'leapp', 'leapp-deps', 'leapp-upgrade-el7toel8', 'leapp-upgrade-el8toel9',
-        'leapp-upgrade-el7toel8-deps', 'leapp-upgrade-el8toel9-deps', 'python2-leapp',
-        'python3-leapp', 'snactor'
-    ]
+    leapp_pkgs = rpms.get_leapp_dep_packages(
+            major_version=['7', '8']) + rpms.get_leapp_packages(major_version=['7', '8'])
```
pirat89, line 486: @fernflower  add here please TODO note that we should change 7, 8 major versions to source & target  and the get_leapp_packages needs to by extended by 9 & 10 then - we will do it when we start to work on 9 -> 10 upgrades. people will be then better aware we need to change that then.

pirat89: @fernflower  thanks!

pirat89 (approved):
LGTM. I am going to do some manual testing yet before the merge, but marking it approved already.

pirat89:
Done some manual testing on RHEL 7 (skipping testing on RHEL 8 now):

```
# leapp preupgrade --debug --no-rhsm
....
2024-01-12 11:50:58.866 DEBUG    PID: 32559 leapp.workflow.FactsCollection.pes_events_scanner: Filtered out leapp related event, event id: 50000
2024-01-12 11:50:58.867 DEBUG    PID: 32559 leapp.workflow.FactsCollection.pes_events_scanner: Filtered out leapp related event, event id: 50001
2024-01-12 11:50:58.868 DEBUG    PID: 32559 leapp.workflow.FactsCollection.pes_events_scanner: Filtered out leapp related event, event id: 50002
...
```

Modification in pes-event.json file:
```
{
    "action": 1, "id": 50000, "architectures": ["x86_64", "aarch64", "ppc64le", "s390x"],
    "in_packageset": {
        "package": [{
            "name": "leapp",
            "repository": "rhel7-extras",
            "modulestreams": [null]
        }],
        "set_id": 12
    },
    "initial_release": {
        "major_version": 7,
        "minor_version": 7,
        "os_name": "RHEL", "tag": null, "z_stream": null
    },
    "modulestream_maps": [],
    "out_packageset": null,
    "release": {
        "major_version": 8,
        "minor_version": 0,
        "os_name": "RHEL", "tag": null, "z_stream": null
    }
},

{
    "action": 3, "id": 50001,
    "architectures": ["x86_64", "aarch64", "ppc64le", "s390x"],
    "in_packageset": {
        "package": [{
            "name": "python2-leapp",
            "repository": "rhel7-extras",
            "modulestreams": [null]
        }],
        "set_id": 17
    },
    "initial_release": {
        "major_version": 7,
        "minor_version": 9,
        "os_name": "RHEL", "tag": null, "z_stream": null
    },
    "modulestream_maps": [{"in_modulestream": null, "out_modulestream": null}],
    "out_packageset": {
        "package": [{
            "name": "python3-leapp",
            "repository": "rhel8-AppStream",
            "modulestreams": [null]
        }],
        "set_id": 18
    },
    "release": {
        "major_version": 8,
        "minor_version": 6,
        "os_name": "RHEL", "tag": null, "z_stream": null
    }
},

{
    "action": 3, "id": 50002,
    "architectures": ["x86_64", "aarch64", "ppc64le", "s390x"],
    "in_packageset": {
        "package": [{
            "name": "leapp-upgrade-el7toel8",
            "repository": "rhel7-extras",
            "modulestreams": [null]
        }],
        "set_id": 17
    },
    "initial_release": {
        "major_version": 7,
        "minor_version": 9,
        "os_name": "RHEL", "tag": null, "z_stream": null
    },
    "modulestream_maps": [{"in_modulestream": null, "out_modulestream": null}],
    "out_packageset": {
        "package": [{
            "name": "leapp-upgrade-el8toel9",
            "repository": "rhel8-AppStream",
            "modulestreams": [null]
        }],
        "set_id": 18
    },
    "release": {
        "major_version": 8,
        "minor_version": 6,
        "os_name": "RHEL", "tag": null, "z_stream": null
    }
},
```

PR #1154, opened 2023-12-05
Update dependencies: require xfsprogs and e2fsprogs

pirat89:
To be able to format our OVL disk images with XFS or Ext4, we need the required tools present on the system. However, on systems with XFS file systems only, it's not needed to have tools for ext4, and vice versa. So on such systems, users can remove these packages manually. In that cases, we get into a problems, especially when XFS is the default FS in our case.

To resolve that, we add dependencies for xfsprogs and e2fsprogs rpms into the spec file, so we are sure these are always present on the system.

In case of Ext4 is a little bit "redundant" - as use of Ext4 is optional. However we expect actually that many people will do it (many == not a small amount of people -> not uncommon use..). So keeping this the least effort, let's add the requirement for both as the actual installation stack is not big.

Packaging:
    * Requires xfsprogs, e2fsprogs
    * Bump leapp-repository-dependencies to 10

jira: RHEL-10847

* Update: Note: we have agreement with QE on this.

### DIFF
--- a/packaging/leapp-repository.spec
+++ b/packaging/leapp-repository.spec
@@ -2,7 +2,7 @@
 %global repositorydir %{leapp_datadir}/repositories
 %global custom_repositorydir %{leapp_datadir}/custom-repositories
 
-%define leapp_repo_deps  9
+%define leapp_repo_deps  10
 
 %if 0%{?rhel} == 7
     %define leapp_python_sitelib %{python2_sitelib}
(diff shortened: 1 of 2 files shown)


dkubek (commented):
lgtm; tested by by installing leapp on a rhel7 machine with xfsprogs and e2fsprogs remove and they are present afterwards

pirat89:
rebased to fix issues with testing farm. also switched to the draft, to ensure we do not make the change until we have ack from qe

fernflower (dismissed):

fernflower (approved):

PR #1155, opened 2023-12-12
Workaround tft issue with listing disabled plans

fernflower:
Until TFT-2298 is resolved a mandatory enabled:true tests filtering won't hurt as we do have some tests that are disabled for particular distros.

OAMG-10177


pirat89 (approved):
lgtm and it works! thanks

PR #1156, opened 2023-12-14
Introduce rpms.get_leapp_packages library function

fernflower:
This function can be used as leapp-installtion-packages-getter. Refactoring of other actors using it will be done later.

### DIFF
--- a/repos/system_upgrade/common/libraries/rpms.py
+++ b/repos/system_upgrade/common/libraries/rpms.py
@@ -1,5 +1,45 @@
 from leapp.libraries import stdlib
+from leapp.libraries.common.config.version import get_source_major_version
 from leapp.models import InstalledRPM
+
+
+class LeappComponents(object):
+    """
+    Supported component values to be used with get_packages_function:
+    * FRAMEWORK - the core of the leapp project: the leapp executable and
+      associated leapp libraries
+    * REPOSITORY - the leapp-repository project
+    * COCKPIT - the cockpit-leapp project
+    * TOOLS - miscellaneous tooling like snactor
+    """
+    FRAMEWORK = 'framework'
+    REPOSITORY = 'repository'
+    COCKPIT = 'cockpit'
+    TOOLS = 'tools'
+
+
+_LEAPP_PACKAGES_MAP = {
+        LeappComponents.FRAMEWORK: {'7': {'pkgs': ['leapp', 'python2-leapp'],
+                                          'deps': ['leapp-deps']},
+                                    '8': {'pkgs': ['leapp', 'python3-leapp'],
+                                          'deps': ['leapp-deps']}
+                                    },
+        LeappComponents.REPOSITORY: {'7': {'pkgs': ['leapp-upgrade-el7toel8'],
+                                           'deps': ['leapp-upgrade-el7toel8-deps']},
+                                     '8': {'pkgs': ['leapp-upgrade-el8toel9'],
+                                           'deps': ['leapp-upgrade-el8toel9-deps']}
+                                     },
+        LeappComponents.COCKPIT: {'7': {'pkgs': ['cockpit-leapp']},
+                                  '8': {'pkgs': ['cockpit-leapp']}
+                                  },
+        LeappComponents.TOOLS: {'7': {'pkgs': ['snactor']},
+                                '8': {'pkgs': ['snactor']}
+                                }
+        }
+
+GET_LEAPP_PACKAGES_DEFAULT_COMPONENTS = frozenset((LeappComponents.FRAMEWORK,
+                                                   LeappComponents.REPOSITORY,
+                                                   LeappComponents.TOOLS))
 
 
 def get_installed_rpms():
@@ -114,3 +154,102 @@ def check_file_modification(config):
     """
     output = _read_rpm_modifications(config)
     return _parse_config_modification(output, config)
+
+
+def _get_leapp_packages_of_type(major_version, component, type_='pkgs'):
+    """
+    Private implementation of get_leapp_packages() and get_leapp_deps_packages().
+
+    :param major_version: Same as for :func:`get_leapp_packages` and
+        :func:`get_leapp_deps_packages`
+    :param component: Same as for :func:`get_leapp_packages` and :func:`get_leapp_deps_packages`
+    :param type_: Either "pkgs" or "deps".  Determines which set of packages we're looking for.
+        Corresponds to the keys in the `_LEAPP_PACKAGES_MAP`.
+
+    Retrieving the set of leapp and leapp-deps packages only differs in which key is used to
+    retrieve the packages from _LEAPP_PACKAGES_MAP.  This function abstracts that difference.
+    """
+    res = set()
+
+    major_versions = [major_version] if isinstance(major_version, str) else major_version
+    if not major_versions:
+        # No major_version of interest specified -> treat as if only current source system version
+        # requested
+        major_versions = [get_source_major_version()]
+
+    components = [component] if isinstance(component, str) else component
+    if not components:
+        error_msg = ("At least one component must be specified when calling this"
+                     " function, available choices are {choices}".format(
+                         choices=sorted(_LEAPP_PACKAGES_MAP.keys()))
+                     )
+        raise ValueError(error_msg)
+
+    for comp in components:
+        for a_major_version in major_versions:
+            if comp not in _LEAPP_PACKAGES_MAP:
+                error_msg = "The requested component {comp} is unknown, available choices are {choices}".format(
+                        comp=component, choices=sorted(_LEAPP_PACKAGES_MAP.keys()))
+                raise ValueError(error_msg)
+
+            if a_major_version not in _LEAPP_PACKAGES_MAP[comp]:
+                error_msg = "The requested major_version {ver} is unknown, available choices are {choices}".format(
+                        ver=a_major_version, choices=sorted(_LEAPP_PACKAGES_MAP[comp].keys()))
+                raise ValueError(error_msg)
+
+            # All went well otherwise, get the data
+            res.update(_LEAPP_PACKAGES_MAP[comp][a_major_version].get(type_, []))
+
+    return sorted(res)
+
+
+def get_leapp_packages(major_version=None, component=GET_LEAPP_PACKAGES_DEFAULT_COMPONENTS):
+    """
+    Get list of leapp packages.
+
+    :param major_version: a list or string specifying major_versions. If not defined then current
+        system_version will be used.
+    :param component: a list or a single enum value specifying leapp components
+        (use enum :class: LeappComponents) If defined then only packages related to the specific
+        component(s) will be returned.
+        The default set of components is in `GET_LEAPP_PACKAGES_DEFAULT_COMPONENTS` and
+        simple modifications of the default can be achieved with code like:
+
+        .. code-block:: python
+            get_leapp_packages(
+                component=GET_LEAPP_PACKAGES_DEFAULT_COMPONENTS.difference(
+                    [LeappComponents.TOOLS]
+            ))
+
+    :raises ValueError: if a requested component or major_version doesn't exist.
+
+    .. note::
+        Call :func:`get_leapp_dep_packages` as well if you also need the deps metapackages.
+        Those packages determine which RPMs need to be installed for leapp to function.
+        They aren't just Requires on the base leapp and leapp-repository RPMs because they
+        need to be switched from the old system_version's to the new ones at a different
+        point in the upgrade than the base RPMs.
+    """
+    return _get_leapp_packages_of_type(major_version, component, type_="pkgs")
+
+
+def get_leapp_dep_packages(major_version=None, component=GET_LEAPP_PACKAGES_DEFAULT_COMPONENTS):
+    """
+    Get list of leapp dep metapackages.
+
+    :param major_version: a list or string specifying major_versions. If not defined then current
+        system_version will be used.
+    :param component: a list or a single enum value specifying leapp components
+        (use enum :class: LeappComponents) If defined then only packages related to the specific
+        component(s) will be returned.
+        The default set of components is in `GET_LEAPP_PACKAGES_DEFAULT_COMPONENTS` and
+        simple modifications of the default can be achieved with code like:
+
+        .. code-block:: python
+            get_leapp_packages(
+                component=GET_LEAPP_PACKAGES_DEFAULT_COMPONENTS.difference(
+                    [LeappComponents.TOOLS]
+            ))
+    :raises ValueError: if a requested component or major_version doesn't exist.
+    """
+    return _get_leapp_packages_of_type(major_version, component, type_="deps")
--- a/repos/system_upgrade/common/libraries/tests/test_rpms.py
+++ b/repos/system_upgrade/common/libraries/tests/test_rpms.py
@@ -1,4 +1,8 @@
-from leapp.libraries.common.rpms import _parse_config_modification
+import pytest
+
+from leapp.libraries.common.rpms import _parse_config_modification, get_leapp_dep_packages, get_leapp_packages
+from leapp.libraries.common.testutils import CurrentActorMocked
+from leapp.libraries.stdlib import api
 
 
 def test_parse_config_modification():
@@ -30,3 +34,64 @@ def test_parse_config_modification():
         "S.5....T.  c /etc/ssh/sshd_config",
     ]
     assert _parse_config_modification(data, "/etc/ssh/sshd_config")
+
+
+@pytest.mark.parametrize('major_version,component,result', [
+   (None, None, ['leapp', 'python3-leapp', 'leapp-upgrade-el8toel9', 'snactor']),
+   ('7', None, ['leapp', 'python2-leapp', 'leapp-upgrade-el7toel8', 'snactor']),
+   (['7', '8'], None, ['leapp', 'python2-leapp', 'leapp-upgrade-el7toel8',
+                       'python3-leapp', 'leapp-upgrade-el8toel9', 'snactor']),
+   ('8', 'framework', ['leapp', 'python3-leapp']),
+   ])
+def test_get_leapp_packages(major_version, component, result, monkeypatch):
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(arch='x86_64', src_ver='8.9', dst_ver='9.3'))
+
+    kwargs = {}
+    if major_version:
+        kwargs["major_version"] = major_version
+    if component:
+        kwargs["component"] = component
+
+    assert set(get_leapp_packages(** kwargs)) == set(result)
+
+
+@pytest.mark.parametrize('major_version,component,result', [
+   ('8', 'nosuchcomponent',
+    (ValueError,
+     r"component nosuchcomponent is unknown, available choices are \['cockpit', 'framework', 'repository', 'tools']")
+    ),
+   ('nosuchversion', "framework",
+    (ValueError, r"major_version nosuchversion is unknown, available choices are \['7', '8']")),
+   ('nosuchversion', False,
+    (ValueError, r"At least one component must be specified when calling this function,"
+     r" available choices are \['cockpit', 'framework', 'repository', 'tools']")),
+])
+def test_get_leapp_packages_errors(major_version, component, result, monkeypatch):
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(arch='x86_64', src_ver='8.9', dst_ver='9.3'))
+
+    kwargs = {}
+    if major_version:
+        kwargs["major_version"] = major_version
+    if component is not None:
+        kwargs["component"] = component
+
+    exc_type, exc_msg = result
+    with pytest.raises(exc_type, match=exc_msg):
+        get_leapp_packages(**kwargs)
+
+
+@pytest.mark.parametrize('major_version,component,result', [
+    (None, None, ['leapp-deps', 'leapp-upgrade-el8toel9-deps']),
+    ('8', 'framework', ['leapp-deps']),
+    ('7', 'tools', []),
+])
+def test_get_leapp_dep_packages(major_version, component, result, monkeypatch):
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(arch='x86_64', src_ver='8.9', dst_ver='9.3'))
+
+    kwargs = {}
+    if major_version:
+        kwargs["major_version"] = major_version
+    if component:
+        kwargs["component"] = component
+
+    assert frozenset(get_leapp_dep_packages(**kwargs)) == frozenset(result)


repos/system_upgrade/common/libraries/rpms.py
```diff
@@ -114,3 +115,36 @@ def check_file_modification(config):
...
+
+def get_leapp_packages(source_version=None, component=None, include_deps=False):
+    """
+    Get list of leapp packages.
+
+    :param source_version: a list or string specifying source_versions. If not defined then all packages
+        for current source_target_version will be returned.
+    :param component: a list or string specifying leapp components. If defined then only packages related
+        to the specific component will be returned.
+    :param include_deps: a flag to control whether deps meta packages are to be filtered out.
```
abadger, line 128: First pass review: An api comment:  I think it would be better to add deps as either separate components or via a separate function.  Using `include_deps` here adds a new concept to understand and a matrix of  combinations to think about. If someone doesn't know what it does under the hood, just where it's used in the code, they might have these questions:

- Why does `get_leapp_packages(component=snactor, include_deps=True)` just return `snactor`?
- All the other parameters are to add additional information to the return, why does this one take things away? (Filter in documentation, not giving additional packages in the snactor case above)

Accessing them as components means you have two more component names (`framework-deps` and `repository-deps`) but you don't have to learn a new concept.

Writing a separate function lets you treat deps as wholly different concepts so you only encounter them in the case where you need to care about the deps packages too.  My impression is only one use case needs deps packages so this might make sense. However, for that use case, you then have to realize you need two use both functions to get what you want so it has a drawback too.  (You could also have a separate function to filter the deps but I think that's worse if the common case is to want results without deps).

Using separate components is what I personally would choose.

repos/system_upgrade/common/libraries/rpms.py
```diff
@@ -114,3 +115,36 @@ def check_file_modification(config):
...
+    for comp in components:
+        for src_version in source_versions:
+            try:
+                res.update(component_map[comp][src_version])
+            except KeyError:
+                continue
+    if not include_deps:
+        # filter out deps packages
+        return sorted([pkg for pkg in res if not pkg.endswith('-deps')])
+    return sorted(res)
```
abadger, line 150: Two things you could do with this code:
```suggestion
    if not include_deps:
        # filter out deps packages
        res = (pkg for pkg in res if not pkg.endswith('-deps'))

    return sorted(res)
```
* [Cosmetic] Generator expression instead of a list comprehension for less cluttered syntax (note this is slower on python2. On python3 it should be the same speed.  It will save memory on both.  But none of this optimization matters for such a short list).

* I'm not always against multiple return statements but this one seems better structured when written this way.  It's a processing of the result before returning.  If a second processing step was added, you would want `res` to pass through both before being returned. These are some general criteria I use for appropriate use of `return`:

  * The common path should exit at the bottom of the function.
  * Exits that can be determined early are fine (examples: cached results, no-ops due to specific parameters).
  * Exits due to errors are fine (some of these are exceptions but others are warnings plus returning from the code). 
  * Other early exits may have their use cases (examples: exiting from a loop, exiting when inside of a large conditional block far from the bottom of the function) but should be looked at on a case by case basis.

abadger (changes requested):
Second pass review.  Some things here are minor and somethings won't apply if we change the api.

pirat89 (changes requested):
tbh, for me it currently does not respect voting results (which we will discusse later; I am not saying the chosen form was the best one and possibly it would be better to go with the PR POC since the beginning, but I would appreciate to try it instead of ignoring it and go in own way).

the main problem I have with the current approach is that:
* it is incomplete - does not cover cockpit-leapp as we discussed
* it's adding snactor separately, which seems to me no conceptual. how we will deal with  leapp-inspector in future? the current logic suggest to  add "inspector" value, which seems almost to me like we are already listing each one rpm by one value. this is getting more far from having it simple.


Also I prefer to have defined a "LeappComponent" class (just a work name) which defines particular values. So using the function, people will be required to put there one of defined values instead of strings. e.g.
```
get_leapp_packages(component=[LeappComponent.Repository])
```

fernflower:
@pirat89 Sorry for taking the initiative, but I stand with what I said on cabal - pure implementation details should never be a matter of internal voting process in an opensource project as that's another layer of non-transparent unnecessary complexity we don't want. 
You are totally right about a dedicated enum, I will definitely do that (along with addressing @abadger implementation-related comments which I totally agree with),

repos/system_upgrade/common/libraries/rpms.py
```diff
@@ -114,3 +115,36 @@ def check_file_modification(config):
...
+                                 '8': ['snactor']}}
+    source_versions = [source_version] if isinstance(source_version, str) else (source_version or
+                                                                                [get_source_major_version()])
+    components = [component] if isinstance(component, str) else component or list(component_map.keys())
+    for comp in components:
+        for src_version in source_versions:
+            try:
+                res.update(component_map[comp][src_version])
+            except KeyError:
+                continue
```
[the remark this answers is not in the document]
fernflower: Do we really want to raise an exception in case user requests wrong version or component? Imho an empty list of packages would make it pretty obvious that something is wrong.

repos/system_upgrade/common/libraries/rpms.py
```diff
@@ -1,7 +1,38 @@
...
+                                     '8': {'pkgs': ['leapp-upgrade-el8toel9'],
+                                           'deps': ['leapp-upgrade-el8toel9-deps']}
+                                     },
+        LeappComponents.COCKPIT: {'7': {'pkgs': ['cockpit-leapp']},
+                                  '8': {'pkgs': ['cockpit-leapp']}
+                                  },
+        LeappComponents.TOOLS: {'7': {'pkgs': ['snactor']},
+                                '8': {'pkgs': ['snactor']}
+                                }
+        }
```
pirat89, line 30: I do not like this to be public. I prefer to keep this private (part of the function or use `_` prefix. There is no code outside of this library that should access this directly. Exposing more stuff than needed creates for us bigger maintainance burden.

repos/system_upgrade/common/libraries/rpms.py
```diff
@@ -114,3 +145,37 @@ def check_file_modification(config):
...
+    res = set()
+    source_versions = [source_version] if isinstance(source_version, str) else source_version
+    if not source_versions:
+        # No source_version of interest specified -> treat as if only current source system version
+        # requested
+        source_versions = [get_source_major_version()]
+    components = [component] if isinstance(component, str) else component
+    if not component:
+        # No component of interest specified -> treat as if packages for all components are requested
+        components = LEAPP_DEFAULT_COMPONENTS
```
pirat89, line 169: The comment and code are actually in conflict. Currently it does not return list of all packages or all components, but default components only. Cockpit-leapp is not included now. Hence the name of the constant is to me confusing.

repos/system_upgrade/common/libraries/rpms.py
fernflower: I can call it `_LEAPP_PACKAGES_MAP`, no problem.

repos/system_upgrade/common/libraries/rpms.py
fernflower: `LEAPP_DEFAULT_COMPONENTS` mean exactly what default components should - leapp components that are returned by default unless any specific ones have been requested.
I'll update the comment and rename the constant to `_LEAPP_DEFAULT_COMPONENTS`, no problem.

repos/system_upgrade/common/libraries/rpms.py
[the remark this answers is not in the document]
fernflower: Alright, let's make it an exception then.

repos/system_upgrade/common/libraries/rpms.py
pirat89: I agree with the suggested solution (update comment + make it private entity).

I do not agree  that "default" is clear, because the current implementation does not corerspond to exactly what default means. Currently, the default contains also `snactor` (tools), which does not have to be installed in the system to do the upgrade. Default from my understanding would be set of packages that are installed when user run `dnf install leapp-upgrade` - iow mandatory packages. Snactor is not part of it. Having it defined publicly, it's not saying it's related to the function - for that, it would require docstring saying that - which in case of public/shared stuff, docstring should be always present. Having it private resolves this problem.

repos/system_upgrade/common/libraries/rpms.py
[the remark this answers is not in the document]
abadger: My criteria for an exception vs something that would pass silently would be something like:

* If there is an error then an exception should be raised.

* If there is a cornercase that is not an error, then an "empty" value like empty list can (and likely should) be returned.

  An example that shows both of the above criteria in action would be:
  ``` python
    def make_hrefs((list_of_urls):
        hrefs = []
        for url in list_of_urls:
            if not validate_url(url):
                raise ValueError(f"{url} was not a valid url")
            hrefs.append(f"<a href={url}>{url}</a>")
        return hrefs
  ```
In the code in this PR, we have a small set of valid user inputs so I think it should be an error if the programmer calling this function sends an invalid input.  That way they are quickly notified that they have a bug that they need to fix.  If it's not an error then a programmer could make a typo when calling the function and either not notice there was a problem or it could cause an error far away from where the typo was made (leading to a more difficult time debugging the problem).

* Once in a while I see or write code that returns a value that isn't a possible, valid value (usually `None`) when it encounters a cornercase.  This probably happens for things that are deemed not to be errors when the return value is normally a scalar (example: an int) Since any integer returned might be valid, another value needs to be chosen.  I think these should be evaluated very closely since they can lead to errors in caller's code.  For instance, imagine if `strcmp()` was written to return `None` if the input value was not a string.  Then this code would be buggy but might not appear so in testing:
  ``` python
    value1 = str(100)
    value2 = 99  # Programmer forgot to convert this value
    # Programmer knows strcmp returns 0 when strings are equal and 1 or -1 if they aren't
    # but they forget that it returns None when bad values are given.
    if not strcmp(value1, value2):
        do_thing_when_strings_are_equal()
  ```
  I think returning a sentinel value is common in code influenced by C since C doesn't have exceptions but it should be avoided when possible in Python code.

abadger:
> pure implementation details should never be a matter of internal voting process in an opensource project

I agree with this sentiment but I wanted to think for a while about why I believe that before replying.  I think in open source projects, we should be striving for consensus decision making whenever possible.  That means that people will raise problems with a proposal (possibly giving a potential solution) and then the proposal should be modified to address the problem (sometimes with the potential solution and sometimes with something different).  We reach consensus when all the problems have addressed.

Sometimes we will disagree on whether something is actually a problem or whether other constraints (time, unable to see another way that has less drawbacks due to other factors, etc) make the solution worse than the problem raised.  For those types of disagreements, voting becomes a way to resolve the impasse.

Consensus decision making does require a few behaviours to work well:

* It requires moving away from expressing mere opinions about code (I think we should have two function to do this task) to being able to enunciate problems that our proposal would address (One parameter is asking to select more data and another parameter asking to filter some of them out.  It's easier to read code that places addition and subtraction into separate functions) so that we all understand what the base problems someone is pointing at are.

* We have to be flexible in allowing other ways of doing something to occur as long as it addresses the base problems (What if instead of filtering, we make the caller select that data as well?)

* It's good to be proactive about pointing out problems in our proposals that we think others might want to propose alternatives to: (I made this a constant holding a frozenset of strings but I know some people like to use Enums more than me.  I think it's okay to do this because it's not user visible but does someone have a different evaluation?)

Hope this is helpful.

pirat89:
> > pure implementation details should never be a matter of internal voting process in an opensource project
> 
> I agree with this sentiment but I wanted to think for a while about why I believe that before replying. I think in open source projects, we should be striving for consensus decision making whenever possible. That means that people will raise problems with a proposal (possibly giving a potential solution) and then the proposal should be modified to address the problem (sometimes with the potential solution and sometimes with something different). We reach consensus when all the problems have addressed.
> 
> Sometimes we will disagree on whether something is actually a problem or whether other constraints (time, unable to see another way that has less drawbacks due to other factors, etc) make the solution worse than the problem raised. For those types of disagreements, voting becomes a way to resolve the impasse.
> 
> Consensus decision making does require a few behaviours to work well:
> 
> * It requires moving away from expressing mere opinions about code (I think we should have two function to do this task) to being able to enunciate problems that our proposal would address (One parameter is asking to select more data and another parameter asking to filter some of them out.  It's easier to read code that places addition and subtraction into separate functions) so that we all understand what the base problems someone is pointing at are.
> * We have to be flexible in allowing other ways of doing something to occur as long as it addresses the base problems (What if instead of filtering, we make the caller select that data as well?)
> * It's good to be proactive about pointing out problems in our proposals that we think others might want to propose alternatives to: (I made this a constant holding a frozenset of strings but I know some people like to use Enums more than me.  I think it's okay to do this because it's not user visible but does someone have a different evaluation?)
> 
> Hope this is helpful.

I basically agree on that. To be clear, I do not have problem to do things differently in future. The only problem I've seen here was that we had 2 different places for discussions & decisions in this case, where some people discussed & voted in the first one (the one we started with), and another part of the team started different approach here. Which is inconsistent and does not meet the original expectation. Again, I am admitting the original way I chose (inspired by the usual one we used over years) in this case from the start was not the best one and this (PR) would be probably better approach. But it started in that way and it set already some expectations. I am in favor to do things differently in future (added topic on retro to discuss it), but I am urging to stay consistent, so people in team know what to expect and they are not surprised & confused later, that reality is different from what has been told - and also, we have more people involved in the process, the split resulted actually in situation when we haven't achieved the main reason why I wanted some voting around -- to involve as many people in the team as possible. I am keeping the discussion on the retro. What happened, it happened. I do not want to make feel anyone bad about that as everyone is making mistakes (including me, especially in this case). Instead, let's finish it and discuss how we will approach such situations (e.g. when voting is wanted) in the future.

repos/system_upgrade/common/libraries/rpms.py
```diff
@@ -114,3 +145,41 @@ def check_file_modification(config):
...
+    for comp in components:
+        for src_version in source_versions:
+            if comp not in _LEAPP_PACKAGES_MAP:
+                error_msg = "The requested component {comp} is unknown, available choices are {choices}".format(
+                        comp=component, choices=sorted(_LEAPP_PACKAGES_MAP.keys()))
+                raise ValueError(error_msg)
+            if src_version not in _LEAPP_PACKAGES_MAP[comp]:
+                error_msg = "The requested source_version {ver} is unknown, available choices are {choices}".format(
+                        ver=src_version, choices=sorted(_LEAPP_PACKAGES_MAP[comp].keys()))
+                raise ValueError(error_msg)
```
pirat89, line 180: I would like to know your opinion(s) on that guys. I am thinking in this way:
* we are responsible that the src_version is defined same for all `Leapp.Components`. 
* If some new component is added, or some existing dropped, it should return empty list.
* If I create a generic actor, that will use this function for all upgrade paths, and we realize it will crash under specific conditions for a new major upgrade path, it could be still too late (even unit-tests will not count with that as they will be written based on the situation that exists during writting the code calling this function).

The implementation does not need to be changed, but in case we have same understanding / opinion on the scenariou I mentioned above, it would make sense to put this check before the loop, to check just the generic list of possible values. In this case `['7', '8']` that will be soon extended by `'9'`. WDYT? @oamg/developers 

To be clear, this one is not blocker, but it's better to point this out so people think about that.

repos/system_upgrade/common/libraries/rpms.py
```diff
@@ -114,3 +145,41 @@ def check_file_modification(config):
     """
     output = _read_rpm_modifications(config)
     return _parse_config_modification(output, config)
+
+
+def get_leapp_packages(source_version=None, component=None, include_deps=False):
```
pirat89, line 150: As noted in the cabal document, `source_*version` is not good name as:
* it's not clear it is actually major version. just version means X.Y in case of RHEL, which we do not expect.
* There are use-cases when it's needed to get name of package for the different version of system, and than the source_version is confusing as we are interested also about the target major version.

Rename it to `major_version`.

```suggestion
def get_leapp_packages(major_version=None, component=None, include_deps=False):
```
- needs updated rest of the code and docstrings

pirat89 (dismissed):
I finished my review (sorry I did it on several times, but I didn't want to work during my PTO so I finished the whole pre-final review now. I am still unsure about some stuff but it needs to be discussed via vcall as I do not see how we could progress there soon enough as we have time pressure here. So hopefully we will have discussion about that at least in a group of people during our canban mtg.

The fingerprint of the function voted originally is following:
```
leapp.libraries.common.rpms.get_leapp_packages(
    major_release: str|list,
    project: list [framework, repository, cockpit-leapp],
    include_deps: bool
)
```

In case of the `project` name, I am ok to go with `component` instead. The rest should be still discussed.

repos/system_upgrade/common/libraries/rpms.py
fernflower: will do

pirat89:
@oamg/developers  adding comments from the discussion we have with @fernflower  and @Rezney:
* we agreed to stay with the current set of components (framework, reposutory, cockpit, tools)
  * docstring will explain what to expect from each of them
* we have decided to with the other earlier discussed proposal, to get deps packages from a different function. in this case another function (e.g. `get_leapp_deps_packages` or sth like that) will be created, that will have `major_version` and `component` parameters.

So `include_deps` parameter is going to be dropped.

Correct me if I missed something. So new functions have fingerprints.

repos/system_upgrade/common/libraries/rpms.py
fernflower: Sorrt, but I am against overcomplicating the logic at this point and committing ourselves to the Leapp.Components support cycle. The goal of this pr is to introduce a library helper - nothing more and nothing less.

pirat89: once this is part of the library, it's already supported by us.

fernflower: Sorry, I am completely missing what you are suggesting here, I see only many ifs and fear of something failing that is not even there yet.

If something crashes we fix it. The api doesn't change either way. What are we discussing here?

pirat89: Sorry, I tried to provide it the example in the description of this thread. But I will try it in a different way. So e.g.:
```
_MY_LEAPP_PACKAGES_MAP = {
    "framework": {'7': {'pkgs': ['leapp', 'python2-leapp'],
                                          'deps': ['leapp-deps']},
                                    '8': {'pkgs': ['leapp', 'python3-leapp'],
                                          'deps': ['leapp-deps']}
                                    },
    "newcomponent":  {'8': {'pkgs': ['leapp', 'python3-leapp'],
                                          'deps': ['leapp-deps']}
                                    },
```
calling `get_leapp_packages("7", "newcomponent")`  ends with an error now. As mentioned in the start of this thread, I will point out one more time key things:
* I think we should in such a case return empty list
* valid values in such a scenario are "7", "8".
* such a scenario should not happen, and when the map is changed in any way, all releases should be always defined. so in case of "newcomponent" should be always defined "7" & "8" in this scenario
  * it's simple to cover such a scenario in the code, where having undefined release lead to an empty list

Why I started the thread is, that if we agree on that, we can modify the implementation a little bit, to make it more clear and safe. Which would mean putting one if statement out of the loop, and in a case we would like to keep some releases "undefined", we could add one `if` into the loop to automatically act like an empty list is defined. pseudocode:
```
def get_leapp_packages(...):
    ...
    if major_release and major_release not in ["7", "8", ..]:
       raise ValueError
   ...
   for comp in components:
       if release not in _MAP[comp]:
           # nothing to do - release is valid, but undefined for the particular component
           log.warning(...) 
           continue
```

fernflower: Why should the behavior be different for `get_leapp_component('7', 'nosuchcomponent')` from raising ValueError as we agreed before? IIRC we had this changed specifically because we wanted to provide clear fast failure for anyone who uses it wrong, and the empty list doesn't help much here.

pirat89: @fernflower  because for me it's different whether I put call `get_leapp_component("1")` - with completely nonsense value for a release that we do not cover anywhere in the code / that is unsupported at all, `and sitaution when `get_leapp_component("7")` is valid but our mistake that the map is not full. As I wrote, we can take full responsibility, so with any change, the author and reviewer will ensure the map will be always complete. But I raise that to think about that, that we could prevent us from issues. As I wrote, it's not blocker,  I just started the public discussion to get people' opinions

fernflower: Okay, my position stands - I have no strong feelings about ValueError (crash early in case of wrong usage) or empty list (not getting a value and possibly messing up reports), but I am strictly against a mix of both, as it has the worst of both worlds unfortunately.

pirat89: Thanks for your opinion. As I mentioned, this is not blocker and it can be changed in future in case we will want as now there is not yet an existing scenario in the code for the behaviour we are discussing. Fix please the other stuff as we discsused so we could finish this PR. Keeping this thread opened for others.

repos/system_upgrade/common/libraries/rpms.py
```diff
@@ -114,3 +145,41 @@ def check_file_modification(config):
...
+    """
+    Get list of leapp packages.
+
+    :param source_version: a list or string specifying source_versions. If not defined then all packages
+        for current source_target_version will be returned.
+    :param component: a list or string specifying leapp components. If defined then only packages related
+        to the specific component will be returned.
```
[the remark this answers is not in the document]
fernflower: done

repos/system_upgrade/common/libraries/rpms.py
```diff
@@ -114,3 +145,41 @@ def check_file_modification(config):
...
     output = _read_rpm_modifications(config)
     return _parse_config_modification(output, config)
+
+
+def get_leapp_packages(source_version=None, component=None, include_deps=False):
+    """
+    Get list of leapp packages.
+
+    :param source_version: a list or string specifying source_versions. If not defined then all packages
+        for current source_target_version will be returned.
```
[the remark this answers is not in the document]
fernflower: done

repos/system_upgrade/common/libraries/rpms.py
```diff
@@ -1,7 +1,38 @@
 from leapp.libraries import stdlib
+from leapp.libraries.common.config.version import get_source_major_version
 from leapp.models import InstalledRPM
 
 
+class LeappComponents(object):
+    FRAMEWORK = 'framework'
+    REPOSITORY = 'repository'
+    COCKPIT = 'cockpit'
+    TOOLS = 'tools'
```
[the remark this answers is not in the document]
fernflower: done

fernflower:
Resolved all comments but splitting the function into 2 (one returns deps packages, one doesn't). Having slept on this I'd like to gather some more input from other folks, especially @matejmatuska who needs that and @dkubek .

repos/system_upgrade/common/libraries/rpms.py
```diff
@@ -114,3 +152,42 @@ def check_file_modification(config):
...
+
+def get_leapp_packages(major_version=None, component=None, include_deps=False):
+    """
+    Get list of leapp packages.
+
+    :param major_version: a list or string specifying major_versions. If not defined then current
+        system_version will be used.
+    :param component: a list of or a single enum value specifying leapp components
+        (use enum :class: LeappComponents) If defined then only packages related to the specific
+        component(s) will be returned.
```
pirat89, line 165: typo

abadger:
```suggestion
    :param component: a list or a single enum value specifying leapp components
        (use enum :class: LeappComponents) If defined then only packages related to the specific
        component(s) will be returned.
```

abadger, line 165: SUGGESTION_TAG: easier_component_modification

```suggestion
        (use enum :class: LeappComponents) If defined then only packages related to the specific
        component(s) will be returned. The default set of components is in
        `LEAPP_DEFAULT_COMPONENTS` and simple modifications of the default can be achieved
        with code like:
        .. code-block:: python
            get_leapp_packages(
                component=LEAPP_DEFAULT_COMPONENTS.difference(
                    [LeappComponents.TOOLS]
            ))
```

abadger (commented):
Looks good.  I have a set of suggestions that would make it easier for callers to make simple modifications to the default components which are requested.  What do you think of those?

repos/system_upgrade/common/libraries/rpms.py
abadger: When I read `DEFAULT` my understanding is that it is the default for the component argument to the `get_leapp_packages()` function.  But I agree that there's more than one meaning of default which the user could be conscious of here.  Maybe `LEAPP_*COMPONENTS`  fails to link `DEFAULT` to the function?  So something like `GET_LEAPP_PACKAGES_COMPONENT_DEFAULT` is better?  (But really long).

Note: I think the discussion about what to name the parameter also leads to this confusion: we chose to name it `component` because there is a pre-existing LEAPP concept of `component`.  But now, when we name this variable, it is only about the parameter name, not about the larger concept that the parameter name refers to.

repos/system_upgrade/common/libraries/rpms.py
abadger: I think we should raise `ValueError` in case there is an unknown component, but your point about making it easier for people to write generic actors is valid.  I think we could solve it in another way.  However, I can also see two scenarios: One in favour of making it easier and one in favour of forcing the caller to update their code:

### In favour of making it easier:
* `leapp-cockpit` needs to be factored into the RHEL 7 and 8 upgrade.
* `leapp-cockpit` must not be factored into the RHEL 9 upgrade (perhaps the package is not shipped on that distro version).
* Calling code had: `get_leapp_packages(components=(LeappComponents.FRAMEWORK, LeappComponents.COCKPIT))`

In this case, if we did not error due to the use of `LeappComponents.COCKPIT`, there would be no problem and the caller would appreciate not having to modify their code to handle both 7&8 upgrades and 9 upgrades.

### In favour of forcing an update:
* `leapp-cockpit` for the RHEL 9 upgrade, `leapp-cockpit` and `snactor` must be installed together.  Therefore, `leapp-cockpit`  is moved into the `TOOLS` component. 
* Calling code had: `get_leapp_packages(components=(LeappComponents.FRAMEWORK, LeappComponents.COCKPIT))`

In this case, we need to force the caller to update their code.  If they do not get an error, they will not get both `leapp-cockpit` and `snactor` returned as leapp_packages which will be a bug in their code.

### Solution that handles both (but may be overly complex):

* For the first use case, `LEAPP_DEFAULT_COMPONENTS` can continue to have an entry for `COCKPIT` but it will be an empty list.  Then, there will be no error in the calling code.
* For the second use case, `COCKPIT` must be removed.  That way callers will be forced to update their code to do the right thing.

The danger in having two different ways of handling this as just an entry or lack of entry in the mapping is that someone might not realize that a particular change should follow one or the other strategy and introduce a bug by doing the wrong thing (this could even be a regression if someone thought they were fixing a bug in another PR but in reality, it was doing the right thing by raising an exception/not raising an exception initially).

### How to make sure that testing catches errors?

We can write a unittest that will catch whether the code will throw errors or not throw errors when we don't expect it by looping through the components listed in the mapping and making sure that ones we do not want to throw errors exist in all releases.  However, as noted in the previous section, this test is only as good as we are in making sure the test is correct.  If the test failure doesn't explain the issue well enough to the developer, they will just change the test instead of realizing their code change is a mistake.

repos/system_upgrade/common/libraries/rpms.py
```diff
@@ -1,7 +1,45 @@
...
+    Supported component values to be used with get_packages_function:
+    * FRAMEWORK - the leapp project
+    * REPOSITORY - the leapp-repository project
+    * COCKPIT - the cockpit-leapp project
+    * TOOLS - miscellaneous tooling like snactor
+    """
+    FRAMEWORK = 'framework'
+    REPOSITORY = 'repository'
+    COCKPIT = 'cockpit'
+    TOOLS = 'tools'
```
pirat89, line 17: The docstring is incomplete. It still does not clarify correctly what to expect mainly in framework and tools components. The framework should clearly specify it does not include snactor (which is part of the framework project) and it includes also the leapp tool. Hence, the tools component does not contain the leapp tool.

abadger:
```suggestion
    """
    Supported component values to be used with get_packages_function:
    * FRAMEWORK - the core of the leapp project: the leapp executable and
      associated leapp libraries
    * REPOSITORY - the leapp-repository project
    * COCKPIT - the cockpit-leapp project
    * TOOLS - miscellaneous tooling like snactor
    """
    FRAMEWORK = 'framework'
    REPOSITORY = 'repository'
    COCKPIT = 'cockpit'
    TOOLS = 'tools'
```

repos/system_upgrade/common/libraries/rpms.py
```diff
@@ -1,7 +1,45 @@
...
+                                  '8': {'pkgs': ['cockpit-leapp']}
+                                  },
+        LeappComponents.TOOLS: {'7': {'pkgs': ['snactor']},
+                                '8': {'pkgs': ['snactor']}
+                                }
+        }
+_LEAPP_DEFAULT_COMPONENTS = [LeappComponents.FRAMEWORK,
+                             LeappComponents.REPOSITORY,
+                             LeappComponents.TOOLS]
+
```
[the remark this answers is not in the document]
abadger: Implemented this except for making the variable public (will add that as a separate commit).

repos/system_upgrade/common/libraries/rpms.py
```diff
@@ -114,3 +152,42 @@ def check_file_modification(config):
...
+    res = set()
+    major_versions = [major_version] if isinstance(major_version, str) else major_version
+    if not major_versions:
+        # No major_version of interest specified -> treat as if only current source system version
+        # requested
+        major_versions = [get_source_major_version()]
+    components = [component] if isinstance(component, str) else component
+    if not component:
+        # No component of interest specified -> return default leapp components
+        components = _LEAPP_DEFAULT_COMPONENTS
```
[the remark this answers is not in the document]
abadger: Implemented this

repos/system_upgrade/common/libraries/rpms.py
```diff
@@ -114,3 +152,42 @@ def check_file_modification(config):
     """
     output = _read_rpm_modifications(config)
     return _parse_config_modification(output, config)
+
+
+def get_leapp_packages(major_version=None, component=None, include_deps=False):
```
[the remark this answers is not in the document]
abadger: Implemented this except for making the variable public (will add that as a separate commit).

repos/system_upgrade/common/libraries/rpms.py
abadger: This (and the code to go along with it) has been added as https://github.com/oamg/leapp-repository/pull/1156/commits/e61b212aa845af70c30439498629fed74eb1be11

@pirat89 @fernflower If you like the change in that commit, we can squash everything together.  If you don't like it, we can revert just that one commit and discuss it in a different PR.

fernflower (commented):
@abadger thank you, looks perfect to me.
@pirat89 review please

repos/system_upgrade/common/libraries/rpms.py
abadger: We've moved include_deps to a separate function so this should now be resolved.

pirat89 (approved):
Looks great to me. `GET_LEAPP_PACKAGES_DEFAULT_COMPONENTS` is a little bit longer, but I believe it's not a problem. The new name is clear for me. Merging.

PR #1161, opened 2024-01-11
Several enhancements to the Makefile

abadger:
* Allow arbitrary user supplied arguments for pytest, pylint, and flake8.  This can be used, for instance, to select specific tests in pytest (PYTEST_ARGS="-k 'perform_ok'"), or to "disable" a single linter: (PYLINT_ARGS='--version').
* Better document how to determine the proper value for ACTOR=<actor>.


pirat89 (approved):
Nice change! :)

PR #1163, opened 2024-01-12
Update the data files (pes, repomap, dddd): CTC2-0

pirat89:
There will be another update of data duing the next week(s), so keeping expected provided_data_stream set on 2.0. However the data are already marked to provide 2.0 & 3.0.


abadger (approved):
`provided_data_streams` is correct and the explanation of keeping both `2.0` and `3.0` for now makes sense.

Approved.

pirat89:
Tests passed, merging.

PR #1164, opened 2024-01-17
[DO NOT MERGE] SELinux: Work around "semanage import bug" RHEL-3295

vmojzis:
https://issues.redhat.com/browse/RHEL-3295


vmojzis:
I'd like to test this change properly (i.e. add policy customizations that would trigger the bug), but https://gitlab.cee.redhat.com/leapp/oamg-rhel8-vagrant does not work for me (ERROR! couldn't resolve module/action 'redhat_subscription'). Are there any other ways?

vmojzis:
Could you please point me to some instructions on how to run the tests with "DESTRUCTIVE_TESTING=True" ? I don't know of a way to test this without possibly modifying the underlying system.

fernflower:
Hi @vmojzis ! As a team we have moved away from writing and running destructive unit tests in favor of dealing with anything destructing in our tmt/tft based integration testing framework. So the `DESTRUCTIVE_TESTING` marked tests are a leftover that is not being run at the moment.
But if you'd like to run this test anyway, a bit hackish approach comes to mind - to prepare a containerized environment (our Makefile is capable of that) and run the test there.
This should do the trick
1. Prepare the containerized environment (rhel8 by default, can be controlled by setting `_TEST_CONTAINER` env var). You can kill all tests execution right after container setup has been finished.
```
make test_container_no_lint
```
2. Run specifically the destructive test. `ACTOR` helps to limit the tests to the specific actor only, `DESTRUCTIVE_TESTING` env var will enable the skipped-by-default-destructive test.
```
podman exec leapp-repo-tests-rhel8-cont bash -c 'cd /repocopy; ACTOR=selinuxprepare DESTRUCTIVE_TESTING=True make test_no_lint'
```

Hope it helps!

fernflower:
But to be honest I believe we should add a proper integration test:
1. Add customizations
2. Attempt an upgrade
3. Check for errors and confirm that report matches what's expected

Although I managed to run the existing destructive selinux test I am getting an ERROR
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
repos/system_upgrade/common/actors/selinux/selinuxprepare/tests/component_test_selinuxprepare.py:35: in _run_cmd
    return run(cmd, split=split).get('stdout', '')
tut/lib/python3.6/site-packages/leapp/libraries/stdlib/__init__.py:213: in run
    'process-result', {'id': _id, 'parameters': args, 'result': audit_result, 'env': env}
tut/lib/python3.6/site-packages/leapp/utils/audit/__init__.py:286: in create_audit_entry
    'data': data
tut/lib/python3.6/site-packages/leapp/utils/audit/__init__.py:87: in store
    with get_connection(db) as connection:
tut/lib/python3.6/site-packages/leapp/utils/audit/__init__.py:73: in get_connection
    return create_connection(cfg.get('database', 'path'))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
```

fernflower:
Could not let it go - looks like container was missing several selinux related packages, after installing `policycoreutils-python-utils`, `policycoreutils` things started to look much-much better :) 

So this is a viable workaround.

vmojzis:
> But to be honest I believe we should add a proper integration test:
> 
> 1. Add customizations
> 2. Attempt an upgrade
> 3. Check for errors and confirm that report matches what's expected

That sounds great. Could you please point me to a simple existing integration test I could use as a template? The "destructive" tests we currently have for selinux actors should already contain all the pieces, so I just need to put it together.

fernflower:
@vmojzis Could you please describe the test scenario? I will help to convert it into leapp integration test in our downstream framework.

repos/system_upgrade/common/actors/selinux/selinuxapplycustom/libraries/selinuxapplycustom.py
```diff
@@ -70,3 +71,15 @@ def back_up_failed(module_path):
     except OSError:
         api.current_logger().warning('Failed to back-up: {}!'.format(module_path))
         return
+
+
+# Work around a "semanage import bug" by replacing "-a" (add) with -m (modify)
+def modify_instead_of_add(command):
+    com = command.split()
+    if len(com) < 2:
+        return None
```
fernflower, line 80: A plain `return` should do.

repos/system_upgrade/common/actors/selinux/selinuxapplycustom/libraries/selinuxapplycustom.py
```diff
@@ -70,3 +71,15 @@ def back_up_failed(module_path):
...
+    if com[0] in SEMANAGE_MODIFY_BUG and com[1] == "-a":
+        com[1] = "-m"
+        return " ".join(com)
+
+    return None
```
fernflower, line 85: Same here

repos/system_upgrade/common/actors/selinux/selinuxapplycustom/libraries/selinuxapplycustom.py
fernflower: Nope, linter will be unhappy.

fernflower (approved):
lgtm, tests passed

PR #1168, opened 2024-01-22
device driver deprecation data: print nice error msg on malformed data

pirat89:
In case of malformed device_driver_deprecation_data.json user could originally see raw traceback without having too much information what it actually means or how to fix it. That usually happens only when the file is manually modified on the machine. So in this case we inform user what file is problematic and how to restore the original file installed by our package.

In case of upstream development, this msg could be seen also when new data is provided if:
 * data file is malformed
 * data file has a new format of data (still json expected)
 * etc. These issues however will be discovered prior the merge as the running tests will fail, so such a problematic file should never get part of the upstream. From that point, we will be expect that user has malformed / customized data file. So no need to handle all possible errors differently in this case.

Jira: OAMG-7549

### DIFF
--- a/repos/system_upgrade/common/actors/loaddevicedriverdeprecationdata/libraries/deviceanddriverdeprecationdataload.py
+++ b/repos/system_upgrade/common/actors/loaddevicedriverdeprecationdata/libraries/deviceanddriverdeprecationdataload.py
@@ -1,6 +1,9 @@
+from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.common import fetch
+from leapp.libraries.common.rpms import get_leapp_packages, LeappComponents
 from leapp.libraries.stdlib import api
 from leapp.models import DeviceDriverDeprecationData, DeviceDriverDeprecationEntry
+from leapp.models.fields import ModelViolationError
 
 
 def process():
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/common/actors/loaddevicedriverdeprecationdata/libraries/deviceanddriverdeprecationdataload.py
```diff
@@ -22,12 +25,29 @@ def process():
...
+                    if entry.get('device_type') in supported_device_types
+                ]
+            )
         )
-    )
+    except (ModelViolationError, ValueError, KeyError, AttributeError, TypeError) as err:
+        # All listed errors, we expect this to happen only when data is malformed
+        # or manually updated. Corrupted data in the upstream is discovered
+        # prior the merge thanks to the testing. So suggest just the restoration
+        # of the file.
```
abadger, line 42:
```suggestion
        # For the listed errors, we expect this to happen only when data is malformed
        # or manually updated. Corrupted data in the upstream is discovered
        # prior the merge thanks to testing. So just suggest the restoration
        # of the file.
```

abadger (approved):
Looks good.  I made a cosmetic wording suggestion for a comment.  Entriely up to you whether to add that or not.

Rezney (approved):

PR #1173, opened 2024-02-02
 Make the reboot required text more visible in the console output

matejmatuska:
The "A reboot is required to continue. Please reboot your system." message is printed before the reports summary and thus is easily overlooked by users.

This patch adds a second such message after the report summary to improve this.

Example (slightly outdated, the extra comma is already removed):
image

Jira: RHEL-22736

### DIFF
--- a/commands/upgrade/__init__.py
+++ b/commands/upgrade/__init__.py
@@ -113,6 +113,11 @@ def upgrade(args, breadcrumbs):
 
     if workflow.failure:
         sys.exit(1)
+    elif not args.resume:
+        sys.stdout.write(
+            'Reboot the system to continue with the upgrade.'
+            ' This might take a while depending on the system configuration.\n'
+        )
 
 
 def register(base_command):


matejmatuska:
The exact message/wording is up for discussion of course, but I think it should also mention that the reboot can/will take more time than normal reboot.
The text could also be bold/colored, but that might be too much in my opinion.

commands/upgrade/__init__.py
```diff
@@ -113,6 +113,11 @@ def upgrade(args, breadcrumbs):
 
     if workflow.failure:
         sys.exit(1)
+    elif not args.resume:
+        sys.stdout.write(
+            'To continue with the upgrade reboot the system.'
```
fernflower, line 118: Nit: I'd rephrase it `Please reboot the system to continue with the upgrade` to emphasize the reboot action necessity.

pirat89:
@matejmatuska add please jira ticket number to the commit msg and PR description.

fernflower:
Rebase against master should fix the tests

commands/upgrade/__init__.py
pirat89: I like the suggested wording, I would just drop the please, so

```suggestion
            'Reboot the system to continue with the upgrade.'
```

but I am also ok with the current one. I will ask Chris about his opinion.

pirat89: @matejmatuska FYI,  I've updated the first sentence as above + rebased.

commands/upgrade/__init__.py
```diff
@@ -113,6 +113,11 @@ def upgrade(args, breadcrumbs):
 
     if workflow.failure:
         sys.exit(1)
+    elif not args.resume:
+        sys.stdout.write(
+            'Reboot the system to continue with the upgrade.'
+            ' This might take a while depending on the system configuration.\n'
```
fernflower, line 119: Nit: Could we please not start the next sentence with a whitespace, but rather finish the previous one with it? Other messages cling to that style and imho it's easier to read through.

`Reboot the system to continue with the upgrade. '`
`'This might take a while depending on the system configuration.\n'`

fernflower (commented):
One nit that is triggering my ocd :)

pirat89 (approved):
```
                   END OF REPORT OVERVIEW                   

Answerfile has been generated at /var/log/leapp/answerfile
Reboot the system to continue with the upgrade. This might take a while depending on the system configuration.
```

Tested, the msg is printed only when the system should be rebooted
* `preupgrade` - no msg
* `upgrade` with inhibitor or error - no msg
* `upgrade --resume (after the reboot) - no msg
* `upgrade` without inhibitor or error - msg printed

PR #1174, opened 2024-02-07
transitionsystemdservicesstates: Do not handle libvirt services on 8->9+

matejmatuska:
Beginning with RHEL8 there are two possible configurations (monolithic and modular) for the libvirt and virt* (virtqemud, virtlockd, virtproxyd, ...) services. The generic strategy implemented in the transitionsystemdservicesstates actor cannot handle such setup and leaves the upgraded system with a an incorrect combination of these configurations. Therefore these services are ignored by the actor and their state transition will be done by the services (and other related units) themselves.

Jira: RHEL-22597


andreabolognani:
@matejmatuska I've tested this locally and it worked exactly as advertised. Good stuff :)

matejmatuska:
I just squashed the commits

pirat89 (approved):

pirat89:
re-approved. the failed AWS tests are not related to this change and can be ignored.

PR #1175, opened 2024-02-09
Update upgrade paths

pirat89:
This is followup on announced changed from commit https://github.com/pirat89/leapp-repository/commit/5a3bded4be67e6a99ba739e15c3b9d533134f35b.
The support for noted RHEL releases ends on May 2024, hence the
upgrade paths related to these releases are dropped.

Upstream tests are going to be covered by https://github.com/oamg/leapp-repository/pull/1176. Unless they start to fail due to dropped upgrade paths, we will keep them as they are for now until we can resolve them nicely.

Jira: OAMG-10452


pirat89:
Note that AWS tests are broken due to a different problem (ignoring these). Tests for dropped upgrade paths started to fail. It's a question whether to wait for a change in packit or whether to do another hacks of tests in the meanwhile. I would honestly wait and ignore the failing tests in case we would be able to use the new structure I created in less than 2w.

Rezney (approved):

PR #1180, opened 2024-02-16
[ceph][luks] Fix ceph cephvolumescan for cephadm

xbezdick:
For cephadm the containers are named ceph-<hash>-osd... while ceph-ansible still uses the ceph-osd-...

Other issue is that OSDs can have multiple volumes in them so filtering just for the first one is wrong and we need to check each volume for the encryption.

Resolves: rhbz#2264543
jira: RHEL-25838

* https://issues.redhat.com/browse/RHEL-25838

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,0 +1,8 @@
+{
+    "extends": [
+        "config:base"
+    ],
+    "enabledManagers": [
+        "github-actions"
+    ]
+}


jbadiapa:
It looks good to me.

pirat89:
@xbezdick rebase please against up-to-date master (fixes crashing tests for dropped upgrade paths)

repos/system_upgrade/common/actors/cephvolumescan/libraries/cephvolumescan.py
```diff
@@ -63,7 +63,11 @@ def encrypted_osds_list():
         output = get_ceph_lvm_list()
         if output is not None:
             try:
-                result = [output[key][0]['lv_uuid'] for key in output if output[key][0]['tags']['ceph.encrypted']]
+                result = []
+                for key in output:
```
fernflower, line 67: `result=[]` at L66 looks like redefinition from L61.
Linter is unhappy at the moment about too many nested blocks, maybe something like this will do?

```
if not output:
    return []
 
for key, elements in output.items():
    for element in elements:
        try:
            if element['tags']['ceph.encrypted']:
                result.append(element['lv_uuid'])
        except KeyError:
            api.current_logger().warning('ceph-osd is installed but no encrypted osd has been found')
            break
return result         
```

fernflower:
Please make our linter happy, added a suggestion in the comment.
 https://github.com/oamg/leapp-repository/actions/runs/7931869922/job/21763410499?pr=1180#step:4:1241

fernflower (commented):
LGTM. One thing - given that the behavior has changed and this actor already has unit tests - could you please consider adding a unit test?

pirat89:
@fernflower Note this PR is bounded to RHEL ticket, so it cannot be merged before the ticket is approved for the release.

repos/system_upgrade/common/actors/cephvolumescan/libraries/cephvolumescan.py
```diff
@@ -8,7 +8,7 @@
 from leapp.models import InstalledRPM
 
 CEPH_CONF = "/etc/ceph/ceph.conf"
-CONTAINER = "ceph-osd"
+CONTAINER = "ceph-.*osd"
```
asm0deuz, line 11: This looks a bit greedy to me, I would rather use:
```suggestion

CONTAINER = r"ceph-(\S*-)?osd-"
```

holser:
LGTM

pirat89:
I am not able to reopen it. one of issues related to using `master` branch for opened PR... for some reason, I cannot see here the pushed commit. so moving to another pr..

PR #1181, opened 2024-02-19
Foreman/Satellite el8toel9 upgrade

evgeni:
* move common Satellite upgrade code to `common` so it can be used by `7to8` and `8to9` upgrades
* add `8to9` specific code to allow Foreman/Satellite upgrades from EL8 to EL9
* refresh collation version if pulp-ansible (and its custom collation) is present
* refactor satellite_upgrade_services to use SystemdServicesTasks


repos/system_upgrade/common/models/satellite.py
```diff
@@ -22,7 +22,7 @@ class SatelliteFacts(Model):
 
     has_foreman = fields.Boolean(default=False)
     """Whether or not foreman is installed on this system"""
-    has_katello_installer = fields.Boolean(default=True)
-    """Whether or not the installer supports Katello additions"""
+    installer_has_systemchecks = fields.Boolean(default=True)
+    """Whether or not the installer supports --disable-system-checks flag"""
```
evgeni, line 26: Is it OK to rename the field in the model, when I know we're the only consumer?
We could also add a new field (or keep the old, not-to-accurate, name).

repos/system_upgrade/common/actors/satellite_upgrade_services/actor.py
```diff
@@ -0,0 +1,35 @@
...
+from leapp.models import SatelliteFacts
+from leapp.tags import ApplicationsPhaseTag, IPUWorkflowTag
+
+SYSTEMD_WANTS_BASE = '/etc/systemd/system/multi-user.target.wants/'
+SERVICES_TO_DISABLE = ['dynflow-sidekiq@*', 'foreman', 'foreman-proxy',
+                       'httpd', 'postgresql', 'pulpcore-api', 'pulpcore-content',
+                       'pulpcore-worker@*', 'tomcat']
+
+
+class SatelliteUpgradeServices(Actor):
```
evgeni, line 14: this is just moved code from `satellite_upgrade_data_migration/actor.py` as those steps are el-version-agnostic, while the code remaining in `satellite_upgrade_data_migration/actor.py` is only needed for 7to8 due to the SCL usage

repos/system_upgrade/el8toel9/actors/satellite_upgrade_facts/actor.py
```diff
@@ -0,0 +1,75 @@
...
+from leapp.models import (
+    InstalledRPM,
+    RepositoriesSetupTasks,
+    RpmTransactionTasks,
+    SatelliteFacts,
+    SatellitePostgresqlFacts
+)
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+
+SATELLITE_VERSION = '6.99'
```
evgeni, line 13: obviously will need to be updated to the right Satellite version, once we know what it will be.

repos/system_upgrade/common/models/satellite.py
pirat89: no, you should  go with the deprecation process, which means 6month protection that the original field will exists and will have still the same functionality as before. with the point that upstream documentation is updated to mention this one field became deprecated. docstring should be updated as well.

evgeni: I guess I'll just stick with the old name then ;)

repos/system_upgrade/common/actors/satellite_upgrade_services/actor.py
```diff
@@ -0,0 +1,35 @@
...
+        facts = next(self.consume(SatelliteFacts), None)
+        if not facts or not facts.has_foreman:
+            return
+
+        # disable services, will be re-enabled by the installer
+        for service_name in SERVICES_TO_DISABLE:
+            for service in glob.glob(os.path.join(SYSTEMD_WANTS_BASE, '{}.service'.format(service_name))):
+                try:
+                    os.unlink(service)
+                except Exception as e:  # pylint: disable=broad-except
```
fernflower, line 34: Do we really need silencing the linter here? I'd say `FileNotFoundError` and `OSError` should cover everything

fernflower (commented):
I see that this is not a ready to go pr (at least given the fact SATELLITE_VERSIONs are not set in stone yet). I will move it to draft so that no one on a merging spree gets the wrong idea :)

repos/system_upgrade/common/actors/satellite_upgrade_services/actor.py
evgeni: Nah, it was just blindly copied from the old code.

Given `FileNotFoundError` is an `OSError`, I think the latter is sufficient?

evgeni:
> I see that this is not a ready to go pr (at least given the fact SATELLITE_VERSIONs are not set in stone yet). I will move it to draft so that no one on a merging spree gets the wrong idea :)

While not ready for Satellite consumption, this can today upgrade upstream Foreman just fine, so I'd be interested to get it in as-is, if you'd be OK with that.

repos/system_upgrade/common/actors/satellite_upgrade_services/actor.py
fernflower: Sure, OSError should be enough if you don't care about the reason

fernflower:
Got it, removing draft :)

repos/system_upgrade/el8toel9/actors/satellite_upgrade_facts/actor.py
fernflower: Could you please add a TODO here then? If this can get in before the right version is there?

evgeni: Actually, we can do better, we can compute the correct name for the el9 repo based on the list of enabled el8 repos!
Will do on Monday.
(Ideas you have when doing dishes and listening to 90ies pop music 😅)

.pylintrc
```diff
@@ -57,6 +57,7 @@ disable=
  redundant-u-string-prefix,  # still have py2 to support
  logging-format-interpolation,
  logging-not-lazy,
+ use-yield-from,  # whatever, new pylint 3.1 thing
```
evgeni, line 60: probably should be an own PR with nicer explanation, but I just wanted to silence it here for now

abadger: The comment could be

> Yield from cannot be used until we require python3.3 or greater

(I'm not sure we'd want to use yield from everywhere possible then but ot will do for now)

evgeni: Thanks, updated and opened https://github.com/oamg/leapp-repository/pull/1187 with this commit only

fernflower (approved):
lgtm codewise.
But looks like the final decision on mergeability is on @mkluson and/or after we get some proof that is works from the Satellite team.

mkluson:
> lgtm codewise. But looks like the final decision on mergeability is on @mkluson and/or after we get some proof that this works as expected from the Satellite team.

Ack, I am in touch with Satellite team and I will coordinate the merge. Thanks

pirat89 (changes requested):
seems there are some issues about the solution based on the internal communication. so I am setting required changes to prevent accidental merge. Feel free to drop it once all problems are resolved and the solution is confirmed to be working as expected.

@evgeni  fill please also the PR description.

evgeni:
@pirat89 FWIW, upgrades work when `foreman-protector` is disabled (as it is in upstream Foreman, or can be done manually in Satellite), so I wouldn't necessarily block this PR on that. Nevertheless, as discussed, I have a patch against `foreman-protector` to self-disable when it detects a Leapp upgrade: https://github.com/theforeman/foreman_maintain/pull/828

Description updated.

evgeni:
@pirat89 you asked me to re-check with the latest changes in `master` and things look good on my end :)

pirat89:
> @pirat89 you asked me to re-check with the latest changes in `master` and things look good on my end :)

Thanks for the check.

pirat89:
@evgeni ha! one more rebase needed due to major change in tests! Introduced by #1209

evgeni:
No worries, rebases are cheap :)

lpramuk:
I have successfully tested this PR for both Satellite and Capsule using:
`leapp-0.17.0-100.20240430083919859552.master.3.gb21450a.el8`
`leapp-repository-0.20.0-0.20240426090422205877.pr1181.30.gc00d93a3.el8`

fernflower (dismissed):
lgtm

fernflower:
@evgeni Could you please squash commits into logical blocks (should not be one commit) and add description to each of those squashed commits' messages?
Happy to merge it afterwards.

evgeni:
Will do.
Also just realized new PES data was updated, so I'll check if we still need the workaround for RHEL-32759 (my expectation is not)

evgeni:
It's not needed \o/

evgeni:
Thanks everyone!

evgeni:
1. Yes, utils is not for Satellite/Capsule systems. It's for systems that need Satellite cli tools (hammer etc), but are not Satellites.
2. This actor is only for Satellite/Capsule systems. Clients don't need special handling (besides repo-enablement, and that should happen via PES)
3. Sat on EL8 uses PostgreSQL 13, Sat on EL9 uses PostgreSQL 13. So we only need to run `reindexdb` to ensure consistent indexes, no PostgreSQL-upgrade is required.

PR #1185, opened 2024-02-21
[wip] Correct scanning of non-PCI network devices

pirat89:
Non-PCI network devices (present usually on IBM Z architecture) does not have ID_VENDOR_ID attribute and also there is nothing we could put into Interface.pci_info about them.

For such devices, set empty string for Interface.vendor field and None value for pci_info.

The Interface model has been updated, allowing None value for pci_info field.

### DIFF
--- a/repos/system_upgrade/common/libraries/persistentnetnames.py
+++ b/repos/system_upgrade/common/libraries/persistentnetnames.py
@@ -42,9 +42,13 @@ def interfaces():
             attrs['name'] = dev.sys_name
             attrs['devpath'] = dev.device_path
             attrs['driver'] = dev['ID_NET_DRIVER']
-            attrs['vendor'] = dev['ID_VENDOR_ID']
-            attrs['pci_info'] = PCIAddress(**pci_info(dev['ID_PATH']))
             attrs['mac'] = dev.attributes.get('address')
+            attrs['vendor'] = dev.get('ID_VENDOR_ID', '')
+            if not attrs['vendor']:
+                attrs['pci_info'] = PCIAddress(**pci_info(dev['ID_PATH']))
+            else:
+                # non-PCI
+                attrs['pci_info'] = None
             if isinstance(attrs['mac'], bytes):
                 attrs['mac'] = attrs['mac'].decode()
         except Exception as e:  # pylint: disable=broad-except
--- a/repos/system_upgrade/common/models/persistentnetnamesfacts.py
+++ b/repos/system_upgrade/common/models/persistentnetnamesfacts.py
@@ -24,7 +24,7 @@ class Interface(Model):
     devpath = fields.String()
     driver = fields.String()
     vendor = fields.String()
-    pci_info = fields.Model(PCIAddress)
+    pci_info = fields.Nullable(fields.Model(PCIAddress))
     mac = fields.String()
 
 


repos/system_upgrade/common/libraries/persistentnetnames.py
```diff
@@ -42,9 +42,13 @@ def interfaces():
             attrs['name'] = dev.sys_name
             attrs['devpath'] = dev.device_path
             attrs['driver'] = dev['ID_NET_DRIVER']
-            attrs['vendor'] = dev['ID_VENDOR_ID']
-            attrs['pci_info'] = PCIAddress(**pci_info(dev['ID_PATH']))
             attrs['mac'] = dev.attributes.get('address')
+            attrs['vendor'] = dev.get('ID_VENDOR_ID', '')
+            if not attrs['vendor']:
```
MichalHe, line 47: From PR description: _Non-PCI network devices (present usually on IBM Z architecture) does not have ID_VENDOR_ID attribute and also there is nothing we could put into Interface.pci_info about them._

This leads me to believe that the condition is negated.
```suggestion
            if attrs['vendor']:
```

pirat89: @MichalHe  hi \o you are right. I forgot to switch the condition when I changed the statement.

repos/system_upgrade/common/libraries/persistentnetnames.py
```diff
@@ -42,9 +42,13 @@ def interfaces():
             attrs['name'] = dev.sys_name
             attrs['devpath'] = dev.device_path
             attrs['driver'] = dev['ID_NET_DRIVER']
```
pirat89, line 44: note it's has been discovered another situation, when ID_NET_DRIVER is not defined and it'S not related to PCI/non-PCI issue.

docs/source/configuring-ipu/envars.md
```diff
@@ -26,6 +31,12 @@ If set to `1`, Leapp does not register the system into Red Hat Lightspeed automa
 #### LEAPP_NO_NETWORK_RENAMING
 If set to `1`, the actor responsible to handle NICs names ends without doing anything. The actor usually creates UDEV rules to preserve original NICs in case they are changed. However, in some cases it‘s not wanted and it leads in malfunction network configuration (e.g. in case the bonding is configured on the system). It‘s expected that NICs have to be handled manually if needed.
 
+```{warning}
+The variable is deprecated as it is connected to a legacy solution for handling
+of NIC names during the upgrade. The only supported solution in future counts with
+the use of `net.naming-scheme`.
```
PeterMocary: Nitpick: Some wording changes.
```suggestion
The variable is deprecated as it is a part of the legacy solution for handling
NIC names during the upgrade. Current supported solution uses the `net.naming-scheme`.
```

repos/system_upgrade/common/libraries/persistentnetnames.py
```diff
@@ -36,21 +36,50 @@ def interfaces():
...
+            attrs['mac'] = dev.attributes.get('address', '')
             if isinstance(attrs['mac'], bytes):
                 attrs['mac'] = attrs['mac'].decode()
+
+            # pci info is not provided for cards that do not use PCI bus,
+            # also it can be missing if the interface is not "managed" by udev.
+            # Cards that do not use PCI usually do not have specified vendor as
+            # well.
+            attrs['vendor'] = dev.get('ID_VENDOR_ID', '')
+            if attrs['vendor'] and dev.get('ID_PATH', '').startswith('pci-'):
```
PeterMocary: I don't know if we really want to extract the ID_PATH only if the vendor is defined. Wouldn't it just shadow the pci_info even if the ID_PATH was defined starting with 'pci-'? Regardless of what is expected, I don't see any reason to not extract it even if the vendor is undefined.

repos/system_upgrade/common/actors/persistentnetnamesdisable/libraries/persistentnetnamesdisable.py
```diff
@@ -0,0 +1,82 @@
...
+                url='https://access.redhat.com/solutions/7009239'
+            )
+        )
+
+    create_report(report_entries)
+
+
+def process():
+    interfaces = next(api.consume(PersistentNetNamesFacts)).interfaces
+
```
PeterMocary: Nitpick:
I've been suggested this before, shouldn't we put such guard in place every time?
```suggestion
    persistent_net_names_facts = next(api.consume(PersistentNetNamesFacts), None)
    if not persistent_net_names_facts:
        raise StopActorExecutionError("Did not receive PersistentNetNamesFacts message.")
    interfaces = persistent_net_names_facts.interfaces
```

repos/system_upgrade/common/libraries/persistentnetnames.py
```diff
@@ -36,21 +36,50 @@ def interfaces():
...
+            # Other values that are usually expected but can be missing
+            # in some situations are covered and do not raise this exception.
             continue
 
+        if not all([attrs['driver'], attrs['mac']]):
+            attrs['is_complete'] = False
+            api.current_logger().warning(
+                'Information in udev about %s network interface is incomplete.',
+                attrs['name']
+            )
```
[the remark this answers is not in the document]
pirat89: Actually, it states for information that should be colected but it could not. E.g. in case of the vendor field, it's expected that it will not be always specified - as vendor field seems to be directly connected to PCI. so interface that does not use PCI will have it most likely always empty. Is the information about the interface complete then? yes, it is. Because we have all information that you could get and it's not an error that the information does not exist at all. The flag started with more fields involved and when looking in more and more stuff, it resulted just in these two things. I will need to discuss it yet with udev guys.

repos/system_upgrade/common/actors/persistentnetnamesdisable/libraries/persistentnetnamesdisable.py
```diff
@@ -0,0 +1,131 @@
...
+
+def report_ethX_ifaces():
+    hint_text = 'Rename all ethX network interfaces following the attached KB solution article.'
+    report_external_links = [
+        reporting.ExternalLink(
+            title='How to Perform an In-Place Upgrade when Using Kernel-Assigned NIC Names',
+            url='https://access.redhat.com/solutions/4067471'
+        )
+    ]
```
PeterMocary: There is multiple links and might be two solution articles. Better be specific in the hint text.
```suggestion
    hint_text = 'Rename all ethX network interfaces following the "How to Perform an In-Place Upgrade when Using Kernel-Assigned NIC Names" solution article.'
    report_external_links = [
        reporting.ExternalLink(
            title='How to Perform an In-Place Upgrade when Using Kernel-Assigned NIC Names',
            url='https://access.redhat.com/solutions/4067471'
        )
    ]
```

PR #1186, opened 2024-02-26
Add test for udevadminfo actor

tomasfratrik:
Test if produced message has correctly parsed database exported from command 'udevamd info -e', using mocked database output in files directory.

Jira: OAMG-1277


repos/system_upgrade/common/actors/udev/udevadminfo/tests/test_udevadminfo.py
```diff
@@ -0,0 +1,21 @@
...
+
+    with open(os.path.join(CUR_DIR, 'files', 'udevadm_database'), 'r') as fp:
+        mocked_data = fp.read()
+    monkeypatch.setattr(api, 'produce', testutils.produce_mocked())
+    monkeypatch.setattr(udevadminfo, 'run', lambda *args: {'stdout': mocked_data})
+    udevadminfo.process()
+
+    assert api.produce.called == 1
+    assert isinstance(api.produce.model_instances[0], UdevAdmInfoData)
+    assert api.produce.model_instances[0].db == UdevAdmInfoData(db=mocked_data).db
```
MichalHe, line 21: It is a bit weird to create an `UdevAdmInfoData` instance in this case. 
```suggestion
    assert api.produce.model_instances[0].db == mocked_data
```

repos/system_upgrade/common/actors/udev/udevadminfo/libraries/udevadminfo.py
```diff
@@ -0,0 +1,7 @@
+from leapp.libraries.stdlib import api, run
+from leapp.models import UdevAdmInfoData
+
+
+def process():
+    out = run(['udevadm', 'info', '-e'])['stdout']
```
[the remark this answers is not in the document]
pirat89: yes, udevadm is always available. it's part of systemd-udev library. Some error catching can be added, however, considering this command mast always work otherwise the system is broken basically.

repos/system_upgrade/common/actors/udev/udevadminfo/libraries/udevadminfo.py
```diff
@@ -0,0 +1,11 @@
+from leapp.libraries.stdlib import api, CalledProcessError, run
+from leapp.models import UdevAdmInfoData
+
+
+def process():
+    try:
+        out = run(['udevadm', 'info', '-e'])['stdout']
+    except CalledProcessError:
+        api.current_logger().error("Failed to run 'udevadm info -e'")
+        out = ''
```
pirat89, line 10: @tomasfratrik  that's actually invalid. It changes the behaviour like in the way "nothing wrong happened, continue". try-except is useful, but it's expected to actually raise StopActorExecutionError with the proper message that we cannot obtain required information about devices from udev. In details, you could add the original error msg.

pirat89 (changes requested):

repos/system_upgrade/common/actors/udev/udevadminfo/libraries/udevadminfo.py
tomasfratrik: Okay I see, I'll try to fix it

repos/system_upgrade/common/actors/udev/udevadminfo/tests/test_udevadminfo.py
```diff
@@ -0,0 +1,21 @@
...
+from leapp.libraries.common import testutils
+from leapp.libraries.stdlib import api, run
+from leapp.models import UdevAdmInfoData
+
+CUR_DIR = os.path.dirname(os.path.abspath(__file__))
+
+
+def test_udevadminfo(monkeypatch):
+
+    with open(os.path.join(CUR_DIR, 'files', 'udevadm_database'), 'r') as fp:
```
fernflower, line 13: Nit: Now that `udevadm`  correct execution is checked you could also add a unit test that StopActorExecution is raised.

tomasfratrik: Good Idea, thank you

repos/system_upgrade/common/actors/udev/udevadminfo/libraries/udevadminfo.py
```diff
@@ -0,0 +1,19 @@
...
+from leapp.models import UdevAdmInfoData
+
+
+def process():
+    try:
+        out = run(['udevadm', 'info', '-e'])['stdout']
+    except CalledProcessError as err:
+        raise StopActorExecutionError(
+            message=(
+                "We were unable to gather information about the system devices"
```
pirat89, line 12: It's better to stay with passive voice.
```suggestion
                "Unable to gather information about the system devices"
```

repos/system_upgrade/common/actors/udev/udevadminfo/libraries/udevadminfo.py
```diff
@@ -0,0 +1,19 @@
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.stdlib import api, CalledProcessError, run
+from leapp.models import UdevAdmInfoData
+
+
+def process():
+    try:
+        out = run(['udevadm', 'info', '-e'])['stdout']
+    except CalledProcessError as err:
```
pirat89, line 9: @tomasfratrik  btw, totally correct would be
```suggestion
    except CalledProcessError, OSError as err:
```
as in case the command cannot be executed / executable does not exist, `OSError` is raised instead. But in this case, we are user that `udevadm` will be always present and executable on the system, so `CalledProcessError` is enough in this case. Just saying, so you are aware of the difference.

tomasfratrik: Done.

repos/system_upgrade/common/actors/udev/udevadminfo/libraries/udevadminfo.py
tomasfratrik: Done.

pirat89 (approved):
lgtm! thanks. :fireworks:  first one merged! - just giving it a time to finish tests that are stuck on COPR.

fernflower (approved):
lgtm!

PR #1188, opened 2024-03-08
cephvolumescan: Ceph container names change prevents upgrade due to inhibitwhenluks

asm0deuz:
Container names created by cephadm are now named ceph-{fsid}-osd-XX while in the past it was just
ceph-osd-XX. As the ceph-volume command cannot be run in an OSD container, it cannot detect if only
ceph OSDs are encrypted.

Fixes: https://issues.redhat.com/browse/RHEL-28647


pirat89:
@asm0deuz hi, thank you for the contribution, however it already duplicates #1180. Follow please #1180 for more info.

PR #1190, opened 2024-03-11
Add unit tests for scansourcefiles actor

tomasfratrik:
Test if files are correctly tracked by actor. Create mocked existent/non-existent, modified/non-modified, rpm owned/not-owned files. Also test non-existent version files.

Jira: OAMG-10367

### DIFF
--- a/repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
+++ b/repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
@@ -1,5 +1,142 @@
-def test_scansourcefiles():
-    # TODO(pstodulk): keeping unit tests for later after I check the idea
-    # of this actor with the team.
-    # JIRA: OAMG-10367
-    pass
+import os
+
+import pytest
+
+from leapp.libraries.actor import scansourcefiles
+from leapp.libraries.common import testutils
+from leapp.libraries.stdlib import api, CalledProcessError
+from leapp.models import FileInfo, TrackedFilesInfoSource
+
+
+@pytest.mark.parametrize(
+    ('run_output', 'expected_output_is_modified'),
+    (
+        ({'exit_code': 0}, False),
+        ({'exit_code': 1, 'stdout': 'missing     /boot/efi/EFI (Permission denied)'}, True),
+        ({'exit_code': 1, 'stdout': 'S.5......  c /etc/openldap/ldap.conf'}, True),
+        ({'exit_code': 1, 'stdout': '..?......  c /etc/libaudit.conf'}, False),
+        ({'exit_code': 1, 'stdout': '.....UG..  g /var/run/avahi-daemon'}, False),
+    )
+)
+def test_is_modified(monkeypatch, run_output, expected_output_is_modified):
+    input_file = '/file'
+
+    def mocked_run(cmd, *args, **kwargs):
+        assert cmd == ['rpm', '-Vf', '--nomtime', input_file]
+        return run_output
+
+    monkeypatch.setattr(scansourcefiles, 'run', mocked_run)
+    assert scansourcefiles.is_modified(input_file) == expected_output_is_modified
+
+
+@pytest.mark.parametrize(
+    'run_output',
+    [
+        {'stdout': ['']},
+        {'stdout': ['rpm']},
+        {'stdout': ['rpm1', 'rpm2']},
+    ]
+)
+def test_get_rpm_name(monkeypatch, run_output):
+    input_file = '/file'
+
+    def mocked_run(cmd, *args, **kwargs):
+        assert cmd == ['rpm', '-qf', '--queryformat', r'%{NAME}\n', input_file]
+        return run_output
+
+    monkeypatch.setattr(scansourcefiles, 'run', mocked_run)
+    monkeypatch.setattr(api, 'current_logger', testutils.logger_mocked())
+    assert scansourcefiles._get_rpm_name(input_file) == run_output['stdout'][0]
+
+    if len(run_output['stdout']) > 1:
+        expected_warnmsg = ('The {} file is owned by multiple rpms: {}.'
+                            .format(input_file, ', '.join(run_output['stdout'])))
+        assert api.current_logger.warnmsg == [expected_warnmsg]
+
+
+def test_get_rpm_name_error(monkeypatch):
+    input_file = '/file'
+
+    def mocked_run(cmd, *args, **kwargs):
+        assert cmd == ['rpm', '-qf', '--queryformat', r'%{NAME}\n', input_file]
+        raise CalledProcessError("mocked error", cmd, "result")
+
+    monkeypatch.setattr(scansourcefiles, 'run', mocked_run)
+    assert scansourcefiles._get_rpm_name(input_file) == ''
+
+
+@pytest.mark.parametrize(
+    ('input_file', 'exists', 'rpm_name', 'is_modified'),
+    (
+        ('/not_existing_file', False, '', False),
+        ('/not_existing_file_rpm_owned', False, 'rpm', False),
+        ('/not_existing_file_rpm_owned_modified', False, 'rpm', True),
+        ('/existing_file_not_modified', True, '', False),
+        ('/existing_file_owned_by_rpm_not_modified', True, 'rpm', False),
+        ('/existing_file_owned_by_rpm_modified', True, 'rpm', True),
+    )
+)
+def test_scan_file(monkeypatch, input_file, exists, rpm_name, is_modified):
+    monkeypatch.setattr(scansourcefiles, 'is_modified', lambda _: is_modified)
+    monkeypatch.setattr(scansourcefiles, '_get_rpm_name', lambda _: rpm_name)
+    monkeypatch.setattr(os.path, 'exists', lambda _: exists)
+
+    expected_model_output = FileInfo(path=input_file, exists=exists, rpm_name=rpm_name, is_modified=is_modified)
+    assert scansourcefiles.scan_file(input_file) == expected_model_output
+
+
+@pytest.mark.parametrize(
+    ('input_files'),
+    (
+        ([]),
+        (['/file1']),
+        (['/file1', '/file2']),
+    )
+)
+def test_scan_files(monkeypatch, input_files):
+    base_data = {
+        'exists': False,
+        'rpm_name': '',
+        'is_modified': False
+    }
+
+    def scan_file_mocked(input_file):
+        return FileInfo(path=input_file, **base_data)
+
+    monkeypatch.setattr(scansourcefiles, 'scan_file', scan_file_mocked)
+    expected_output_list = [FileInfo(path=input_file, **base_data) for input_file in input_files]
+    assert scansourcefiles.scan_files(input_files) == expected_output_list
+
+
+@pytest.mark.parametrize(
+    'rhel_major_version', ['8', '9']
+)
+def test_tracked_files(monkeypatch, rhel_major_version):
+    TRACKED_FILES_MOCKED = {
+        'common': [
+            '/file1',
+        ],
+        '8': [
+            '/file2',
+        ],
+        '9': [
+            '/file3',
+        ],
+    }
+
+    def scan_files_mocked(files):
+        return [FileInfo(path=file_path, exists=False, rpm_name='', is_modified=False) for file_path in files]
+
+    monkeypatch.setattr(api, 'produce', testutils.produce_mocked())
+    monkeypatch.setattr(scansourcefiles, 'TRACKED_FILES', TRACKED_FILES_MOCKED)
+    monkeypatch.setattr(scansourcefiles, 'get_source_major_version', lambda: rhel_major_version)
+    monkeypatch.setattr(scansourcefiles, 'scan_files', scan_files_mocked)
+
+    scansourcefiles.process()
+
+    tracked_files_model = api.produce.model_instances[0]
+    assert api.produce.called == 1
+    assert isinstance(tracked_files_model, TrackedFilesInfoSource)
+    # assert only 1 common and 1 version file were scanned
+    assert len(tracked_files_model.files) == 2
+    assert all(isinstance(files_list_item, FileInfo) for files_list_item in tracked_files_model.files)


repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
```diff
@@ -1,5 +1,130 @@
...
+    # assert only files from correct major version were scanned, so common files + 1 (version file)
+    assert len(api.produce.model_instances[0].files) == len(VERSION_FILES_MOCKED['common']) + 1
+
+    assert isinstance(api.produce.model_instances[0].files[0], FileInfo)
+    assert api.produce.model_instances[0].files[0].path == '{}/file_not_rpm_owned'.format(FILES_DIR)
+    assert api.produce.model_instances[0].files[0].exists is True
+    assert api.produce.model_instances[0].files[0].rpm_name == ''
+    assert api.produce.model_instances[0].files[0].is_modified is False
+
+    INDEX_OF_VERSION_FILE = len(VERSION_FILES_MOCKED['common'])
```
matejmatuska, line 49: You could store `api.produce.model_instances[0].files[INDEX_OF_VERSION_FILE]` inside a variable instread to avoid a lot of the repetition below.

matejmatuska (changes requested):
It is a common practice to first test the lower level functions as units and then mock them when testing functions that call them.

So in all of these 3 tests, you would just test the functionality on  `scansourcefiles.scan_file()`. Then you would add a test that tests `scansourcefiles.process()` and mock `scan_file()` in it, because that function is already covered. 

This would reduce the amount of mocks in the lower level tests and also allow you to run the test once per possible input (in this case a file). Then you could use `@pytest.mark.parametrize()` to run the test for multiple inputs (files). You can also specify the expected results there to avoid having specific asserts for each one of them.

You can take a look at https://github.com/oamg/leapp-repository/blob/master/repos/system_upgrade/el8toel9/actors/rocescanner/tests/unit_test_rocescanner.py for an example of such tests. First `get_roce_nics_lines` is properly tested and then it's mocked when testing `process`.

repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
```diff
@@ -1,5 +1,130 @@
...
+from leapp.models import FileInfo, TrackedFilesInfoSource
+
+CUR_DIR = os.path.dirname(os.path.abspath(__file__))
+FILES_DIR = os.path.join(CUR_DIR, 'files')
+
+# values in list should correspond with the return values of get_source_major_version()
+RHEL_MAJOR_VERSIONS_LIST = list(range(8, 9+1))
+
+TRACKED_FILES_BY_VERSIONS = {
+    version: ['{}/file_major_version_{}'.format(FILES_DIR, version)]
```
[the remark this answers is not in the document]
tomasfratrik: I was using `os.path.join` from start, but thought It would probably be more slower (because code uses many instances of it). But you are right, it is overally better, readibility, cross-platform ...

repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
tomasfratrik: Was hesitant about this, because didn't see other actors doing this (didn't check enough of them). But yes, there is no reason for this repetition, which makes it worse

matejmatuska: It isn't really an issue, but since you already have variable for the index you might as well store the object since you don't need the index anywhere else.

repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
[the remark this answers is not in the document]
matejmatuska: I think it would require a lot more calls to notice any performance hit, anyway the test already run for quite a while and in this case correctness > performance.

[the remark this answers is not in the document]
pirat89: @tomasfratrik I recommend you to watch the following video to understand when it's good and when it's premature to optimise something: https://youtu.be/tKbV6BpH-C8 

In short, optimisation that is faster by 0.01s does not make sense when the execution time is in minutes.

repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
```diff
@@ -1,5 +1,163 @@
...
+@pytest.mark.parametrize('input_file,rpm_name,modified,expected_output', (
+    (os.path.join(FILES_DIR, 'file_not_existant'), 'rpm', True,
+     FileInfo(path=os.path.join(FILES_DIR, 'file_not_existant'), exists=False, rpm_name='rpm', is_modified=True)),
+    (os.path.join(FILES_DIR, 'file_not_rpm_owned'), '', False,
+     FileInfo(path=os.path.join(FILES_DIR, 'file_not_rpm_owned'), exists=True, rpm_name='', is_modified=False))
+))
+def test_scan_file(monkeypatch, input_file, rpm_name, modified, expected_output):
+    monkeypatch.setattr(scansourcefiles, 'is_modified', lambda _: modified)
+    monkeypatch.setattr(scansourcefiles, '_get_rpm_name', lambda _: rpm_name)
+
```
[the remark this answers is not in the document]
matejmatuska: Also please add a case for rpm owned file both modified and unmodified.

repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
```diff
@@ -1,5 +1,163 @@
...
+@pytest.mark.parametrize('input_file,rpm_name,expected_output', (
+    (os.path.join(FILES_DIR, 'file_rpm_owned'), 'rpm',
+     FileInfo(path=os.path.join(FILES_DIR, 'file_rpm_owned'), exists=True, rpm_name='rpm', is_modified=False)),
+))
+def test_rpm_owned_files(monkeypatch, input_file, rpm_name, expected_output):
+    monkeypatch.setattr(api, 'produce', testutils.produce_mocked())
+    monkeypatch.setattr(scansourcefiles, 'get_source_major_version', lambda: 9)
+    monkeypatch.setattr(scansourcefiles, '_get_rpm_name', lambda _: rpm_name)
+    monkeypatch.setattr(scansourcefiles, 'is_modified', lambda _: False)
```
matejmatuska, line 119: Mock the scan_files function directly and return what it is expected to return. You can do that because that one is already covered by the test above.

Also, you shouldn't need to mock the TRACKED_FILES, since all calls operating directly on files can be mocked, e.g. the os.path.exists call (as explained in one of my other comments).

repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
```diff
@@ -1,5 +1,163 @@
...
+              for major_version in RHEL_MAJOR_VERSIONS_LIST]
+
+
+@pytest.mark.parametrize('major_version,expected_output,common_expected_output', param_list)
+def test_version_file_with_existant_common_file(monkeypatch, major_version, expected_output, common_expected_output):
+    monkeypatch.setattr(api, 'produce', testutils.produce_mocked())
+    monkeypatch.setattr(api, 'current_logger', testutils.logger_mocked())
+    monkeypatch.setattr(scansourcefiles, 'get_source_major_version', lambda: major_version)
+    monkeypatch.setattr(scansourcefiles, '_get_rpm_name', lambda _: '')
+    monkeypatch.setattr(scansourcefiles, 'is_modified', lambda _: False)
```
matejmatuska, line 148: The same applies as described in the comment in `test_rpm_owned_files`. In this case it is valid to mock TRACKED_FILES as there are no files by version in the actor library, but again it is not required for the `file_not_rpm_owned` to physically exists (see the other comments).

repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
```diff
@@ -1,5 +1,163 @@
...
+    assert isinstance(tracked_files, TrackedFilesInfoSource)
+    file_info = tracked_files.files[0]
+    assert isinstance(file_info, FileInfo)
+
+    assert api.produce.called == 1
+    assert len(tracked_files.files) == 1
+    assert file_info == expected_output
+
+
+param_list = [(major_version,
```
matejmatuska, line 134: This name is not very descriptive, please choose a more descriptive name. But there is nothing wrong with not extracting this into a variable if it's properly formatted in the parametrize.

repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
```diff
@@ -1,5 +1,163 @@
...
+    ({'exit_code': 1, 'stdout': '5 missing'}, True),
+    ({'exit_code': 1, 'stdout': 'fi5ve'}, True),
+    ({'exit_code': 1, 'stdout': 'NoFiveNoMissing'}, False),
+))
+def test_is_modified(monkeypatch, run_output, expected_output):
+    def mocked_run(cmd, *args, **kwargs):
+        assert cmd == ['rpm', '-Vf', '--nomtime', input_file]
+        return run_output
+
+    input_file = os.path.join(FILES_DIR, 'mocked_file')
```
matejmatuska, line 43: Nothing really wrong with this, but you could use basically any path such as `/use/somefile` as long as it's valid. There is not need for it to be at the path you used.

matejmatuska (changes requested):
It's now looking a lot better. However some additional changes are requested.

repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
```diff
@@ -1,5 +1,175 @@
...
+
+@pytest.mark.parametrize('input_files,rpm_name,exists,modified,expected_output', (
+    ([], '', False, False, []),
+
+    (['/file1'], '', False, False,
+     [FileInfo(path='/file1', exists=False, rpm_name='', is_modified=False)]),
+
+    (['/file1', '/file2'], '', False, False,
+     [FileInfo(path='/file1', exists=False, rpm_name='', is_modified=False),
+      FileInfo(path='/file2', exists=False, rpm_name='', is_modified=False)]),
```
pirat89, line 102: that's kind of duplicate input data and also prevents testing combination of files that exists or not, that are owned by rpm or not. The first line specifies params that must be found in all later defined files + names of files.  You can obtain the same result with params defined on the first line, or just with the last param. The second way is what I would do, as It could also introduce testing for various combinations. Another benefit is that input data are more readable.

repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
```diff
@@ -1,5 +1,175 @@
...
+
+import pytest
+
+from leapp.libraries.actor import scansourcefiles
+from leapp.libraries.common import testutils
+from leapp.libraries.stdlib import api, CalledProcessError
+from leapp.models import FileInfo, TrackedFilesInfoSource
+
+# values in list should correspond with the return values of get_source_major_version()
+RHEL_MAJOR_VERSIONS_LIST = list(range(8, 9+1))
```
pirat89, line 11: why? what is value in comparison to do just this?

```suggestion
RHEL_MAJOR_VERSIONS_LIST = [8, 9]
```

Also, it's incorrect as `get_source_major_version` returns string, not integer. so correctly it would be:
```suggestion
RHEL_MAJOR_VERSIONS_LIST = ['8', '9']
```

repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
```diff
@@ -1,5 +1,175 @@
...
+
+    (['/file_rpm_modified'], 'rpm', False, True,
+     [FileInfo(path='/file_rpm_modified', exists=False, rpm_name='rpm', is_modified=True)]),
+
+    (['/file_rpm_not_modified'], 'rpm', False, False,
+     [FileInfo(path='/file_rpm_not_modified', exists=False, rpm_name='rpm', is_modified=False)]),
+))
+def test_scan_files(monkeypatch, input_files, rpm_name, exists, modified, expected_output):
+    def scan_file_mocked(input_file):
+        return FileInfo(path=input_file, exists=exists, rpm_name=rpm_name, is_modified=modified)
```
pirat89, line 112: this is nice example that there is no reason for duplicated input data

pirat89 (changes requested):

tomasfratrik:
My notes:

* I separated some values in pytest parametrize, because I thinkit is now more readable (i can see it faster what is what, and it is not too much longer).
* I split test of function `_get_rpm_name` into 2 tests, 1 containing normal tests, and other one  error. Now it could considered as 'redundant' (even though the code still differs), but I think it is more readable.
* My thought process over test test_tracked_files, which tests process, was that, I will give 'you' (process function), tracked files, and major version, and you do the rest. Process should just take files from 'common' and 'major_version' but I "don't know that", so I let it do whatever it wants to, and then compare. And mainly compare produced message.

I will squash commits, and write proper commit message after approved, or when it is finalized.

repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
tomasfratrik: Done.

repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
tomasfratrik: Done.

repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
tomasfratrik: Yes I see that now, done.

repos/system_upgrade/common/actors/scansourcefiles/tests/unit_test_scansourcefiles.py
```diff
@@ -1,5 +1,175 @@
...
+    ({'exit_code': 1, 'stdout': 'S.5......  c /etc/openldap/ldap.conf'}, True),
+    ({'exit_code': 1, 'stdout': '..?......  c /etc/libaudit.conf'}, False),
+    ({'exit_code': 1, 'stdout': '.....UG..  g /var/run/avahi-daemon'}, False),
+))
+def test_is_modified(monkeypatch, run_output, expected_output):
+    def mocked_run(cmd, *args, **kwargs):
+        assert cmd == ['rpm', '-Vf', '--nomtime', input_file]
+        return run_output
+
+    input_file = '/file'
```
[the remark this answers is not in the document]
tomasfratrik: Done.

pirat89 (approved):
that's much better! lgtm!

tomasfratrik:
When splitting `_get_rpm_name` I forgot to check warnmsg, should be fixed now.

pirat89:
@fernflower seems that automatic building is broken by general.

pirat89 (approved):

matejmatuska (approved):

PR #1191, opened 2024-03-11
rocescanner: Actually call process() in test_roce_notibmz test

matejmatuska:


pirat89 (approved):
@matejmatuska good catch!

PR #1194, opened 2024-04-02
pes_events_scanner: override repositories when applying an event

MichalHe:
The Package class has custom `__hash__` and `__eq__` methods in order to achieve a straightforward presentation via set manipulation. However, this causes problems, e.g., when applying split events. For example:
 Applying the event `Split event: in={(A, repo1)}, out={(A, repo2), (B, repo2)}` to the package state `{(A, repo1), (B, repo1)}` results in the following:
 ` {(A, repo1), (B, repo1)} --apply--> {(A, repo2), (B, repo1)}`
which is undesired as repo1 is a source system repository. Such a package will get reported to the user as potentially removed during the upgrade. This patch addresses this unwanted behavior.

### DIFF
--- a/repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
+++ b/repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
@@ -139,8 +139,11 @@ def compute_pkg_changes_between_consequent_releases(source_installed_pkgs,
         if event.action == Action.PRESENT:
             for pkg in event.in_pkgs:
                 if pkg in seen_pkgs:
+                    # First remove the package with the old repository and add it back, but now with the new
+                    # repository. As the Package class has a custom __hash__ and __eq__ comparing only name
+                    # and modulestream, the pkg.repository field is ignore and therefore the add() call
+                    # does not update the entry.
                     if pkg in target_pkgs:
-                        # Remove the package with the old repository, add the one with the new one
                         target_pkgs.remove(pkg)
                     target_pkgs.add(pkg)
         elif event.action == Action.DEPRECATED:
@@ -163,7 +166,11 @@ def compute_pkg_changes_between_consequent_releases(source_installed_pkgs,
                              event.id, event.action, removed_pkgs_str, added_pkgs_str)
 
                 # In pkgs are present, event can be applied
+                # Note: We do a .difference(event.out_packages) followed by an .union(event.out_packages) to overwrite
+                # #     repositories of the packages (Package has overwritten __hash__ and __eq__, ignoring
+                # #     the repository field)
                 target_pkgs = target_pkgs.difference(event.in_pkgs)
+                target_pkgs = target_pkgs.difference(event.out_pkgs)
                 target_pkgs = target_pkgs.union(event.out_pkgs)
 
         pkgs_to_demodularize = pkgs_to_demodularize.difference(event.in_pkgs)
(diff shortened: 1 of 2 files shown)


fernflower:
While the comments give some details about the mechanism - would it be possible to come up with a unit test that covers the scenario "the package is no longer among packages-to-be-removed"? Or is it too much work and integration tests are a better fit?

MichalHe:
@fernflower I have added the described unit test.

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
```diff
@@ -139,8 +139,9 @@ def compute_pkg_changes_between_consequent_releases(source_installed_pkgs,
         if event.action == Action.PRESENT:
             for pkg in event.in_pkgs:
                 if pkg in seen_pkgs:
+                    # Remove the package with the old repository, add the one with the new one. The Package class has
+                    # a custom __hash__ and __eq__ comparing only name and modulestream - pkg.repository is ignored.
```
dkubek, line 143:
```suggestion
                    # First remove the package with the old repository and add it back, but now with the new repository. 
                    # As the Package class has a custom __hash__ and __eq__ comparing only name and modulestream,
                    # the pkg.repository field is ignore and therefore the add() call does not update the entry.
```

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
```diff
@@ -163,7 +164,10 @@ def compute_pkg_changes_between_consequent_releases(source_installed_pkgs,
                              event.id, event.action, removed_pkgs_str, added_pkgs_str)
 
                 # In pkgs are present, event can be applied
+                # Note: We do a .difference(event.out_packages) followed by an .union(event.out_packages) to overwrite
+                # #     repositories of the packages (Package has overwritten __hash__ and __eq__)
```
dkubek, line 168:
```suggestion
                # Note: We do a .difference(event.out_packages) followed by an .union(event.out_packages) to overwrite
                # #     repositories of the packages (Package has overwritten __hash__ and __eq__, ignoring the repository field)
```

dkubek (commented):
lgtm! Just suggested some changes to comments that make it more transparent to me what is happening, but it's non blocking.

pirat89 (dismissed):
tested manually. seems to be working like as expected now.

* before the fix:
```
Risk Factor: high
Title: Packages from unknown repositories may not be installed
Summary: 1 packages may not be installed or upgraded due to repositories unknown to leapp:
- evolution-data-server-ui (repoid: rhel8-AppStream)
Remediation: [hint] In case the listed repositories are mirrors of official repositories for RHEL (provided by Red Hat on CDN) and their repositories IDs has been customized, you can change the configuration to use the official IDs instead of fixing the problem. You can also review the projected DNF upgrade transaction result in the logs to see what is going to happen, as this does not necessarily mean that the listed packages will not be upgraded. You can also install any missing packages after the in-place upgrade manually.
Key: 9a2b05abf8f45fd7915e52542887bb334bb218ea
```
* after the fix, no report packages from unknown repositories has been generated

Postponing the merge for the possible change of comments resolving the above conversation.

MichalHe:
@pirat89 I have applied @dkubek's suggestions, and, therefore, the patch is ready to be merged.

pirat89:
@MichalHe  :D line too long. pls take a look

MichalHe:
@pirat89 huh, sorry. I did notice when I applied the changes. Should be fine now, though.

pirat89:
@MichalHe np :) it's happening to me as well.

PR #1195, opened 2024-04-02
boot: check first partition offset on GRUB devices

MichalHe:
Check that the first partition starts at least at 1MiB (2048 cylinders), as too small first-partition offsets lead to failures when doing grub2-install. The limit (1MiB) has been chosen as it is a common value set by the disk formatting tools nowadays.

#### Tips to reproduce
1. Use `fdisk -c=dos -u=cylinders </dev/drive>` to force fdisk to allow you to format the target disk in a way that the first partition starts at an offset <= 1MiB
2. Install rhel on the disk (tested with rhel7.9)

jira: https://issues.redhat.com/browse/RHEL-3341

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
+++ b/repos/system_upgrade/el7toel8/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
@@ -1,0 +1,52 @@
+from leapp import reporting
+from leapp.libraries.common.config import architecture
+from leapp.libraries.stdlib import api
+from leapp.models import FirmwareFacts, GRUBDevicePartitionLayout
+
+SAFE_OFFSET_BYTES = 1024*1024  # 1MiB
+
+
+def check_first_partition_offset():
+    if architecture.matches_architecture(architecture.ARCH_S390X):
+        return
+
+    for fact in api.consume(FirmwareFacts):
+        if fact.firmware == 'efi':
+            return  # Skip EFI system
+
+    problematic_devices = []
+    for grub_dev in api.consume(GRUBDevicePartitionLayout):
+        first_partition = min(grub_dev.partitions, key=lambda partition: partition.start_offset)
+        if first_partition.start_offset < SAFE_OFFSET_BYTES:
+            problematic_devices.append(grub_dev.device)
+
+    if problematic_devices:
+        summary = (
+            'On the system booting by using BIOS, the in-place upgrade fails '
+            'when upgrading the GRUB2 bootloader if the boot disk\'s embedding area '
+            'does not contain enough space for the core image installation. '
+            'This results in a broken system, and can occur when the disk has been '
+            'partitioned manually, for example using the RHEL 6 fdisk utility.\n\n'
+
+            'The list of devices with small embedding area:\n'
+            '{0}.'
+        )
+        problematic_devices_fmt = ['- {0}'.format(dev) for dev in problematic_devices]
+
+        hint = (
+            'We recommend to perform a fresh installation of the RHEL 8 system '
+            'instead of performing the in-place upgrade.\n'
+            'Another possibility is to reformat the devices so that there is '
+            'at least {0} kiB space before the first partition. '
+            'Note that this operation is not supported and does not have to be '
+            'always possible.'
+        )
+
+        reporting.create_report([
+            reporting.Title('Found GRUB devices with too little space reserved before the first partition'),
+            reporting.Summary(summary.format('\n'.join(problematic_devices_fmt))),
+            reporting.Remediation(hint=hint.format(SAFE_OFFSET_BYTES // 1024)),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Groups([reporting.Groups.BOOT]),
+            reporting.Groups([reporting.Groups.INHIBITOR]),
+        ])
--- a/repos/system_upgrade/el7toel8/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
+++ b/repos/system_upgrade/el7toel8/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
@@ -16,6 +16,13 @@ def check_first_partition_offset():
 
     problematic_devices = []
     for grub_dev in api.consume(GRUBDevicePartitionLayout):
+        if not grub_dev.partitions:
+            # NOTE(pstodulk): In case of empty partition list we have nothing to do.
+            # This can could happen when the fdisk output is different then expected.
+            # E.g. when GPT partition table is used on the disk. We are right now
+            # interested strictly about MBR only, so ignoring these cases.
+            # This is seatbelt, as the msg should not be produced for GPT at all.
+            continue
         first_partition = min(grub_dev.partitions, key=lambda partition: partition.start_offset)
         if first_partition.start_offset < SAFE_OFFSET_BYTES:
             problematic_devices.append(grub_dev.device)
--- a/repos/system_upgrade/el7toel8/actors/scangrubdevpartitionlayout/libraries/scan_layout.py
+++ b/repos/system_upgrade/el7toel8/actors/scangrubdevpartitionlayout/libraries/scan_layout.py
@@ -31,6 +31,34 @@ def get_partition_layout(device):
         unit = int(unit.split(' ')[0].strip())
         break  # First line of the partition table header
 
+    # Discover disk label type: dos | gpt
+    for line in table_iter:
+        line = line.strip()
+        if not line.startswith('Disk label type'):
+            continue
+        disk_type = line.split(':')[1].strip()
+        break
+
+    if disk_type == 'gpt':
+        api.current_logger().info(
+            'Detected GPT partition table. Skipping produce of GRUBDevicePartitionLayout message.'
+        )
+        # NOTE(pstodulk): The GPT table has a different output format than
+        # expected below, example (ignore start/end lines):
+        # --------------------------- start ----------------------------------
+        # #         Start          End    Size  Type            Name
+        # 1         2048         4095      1M  BIOS boot
+        # 2         4096      2101247      1G  Microsoft basic
+        # 3      2101248     41940991     19G  Linux LVM
+        # ---------------------------- end -----------------------------------
+        # But mainly, in case of GPT, we have nothing to actually check as
+        # we are gathering this data now mainly to get information about the
+        # actual size of embedding area (MBR gap). In case of GPT, there is
+        # bios boot / prep boot partition, which has always 1 MiB and fulfill
+        # our expectations. So skip in this case another processing and generation
+        # of the msg. Let's improve it in future if we find a reason for it.
+        return None
+
     for line in table_iter:
         line = line.strip()
         if not line.startswith('Device'):
--- a/repos/system_upgrade/el7toel8/actors/scangrubdevpartitionlayout/libraries/scan_layout.py
+++ b/repos/system_upgrade/el7toel8/actors/scangrubdevpartitionlayout/libraries/scan_layout.py
@@ -1,0 +1,64 @@
+from leapp.libraries.stdlib import api, CalledProcessError, run
+from leapp.models import GRUBDevicePartitionLayout, GrubInfo, PartitionInfo
+
+SAFE_OFFSET_BYTES = 1024*1024  # 1MiB
+
+
+def split_on_space_segments(line):
+    fragments = (fragment.strip() for fragment in line.split(' '))
+    return [fragment for fragment in fragments if fragment]
+
+
+def get_partition_layout(device):
+    try:
+        partition_table = run(['fdisk', '-l', '-u=sectors', device], split=True)['stdout']
+    except CalledProcessError as err:
+        # Unlikely - if the disk has no partition table, `fdisk` terminates with 0 (no err). Fdisk exits with an err
+        # when the device does not exists, or if it is too small to contain a partition table.
+
+        err_msg = 'Failed to run `fdisk` to obtain the partition table of the device {0}. Full error: \'{1}\''
+        api.current_logger().error(err_msg.format(device, str(err)))
+        return None
+
+    table_iter = iter(partition_table)
+
+    for line in table_iter:
+        if not line.startswith('Units'):
+            # We are still reading general device information and not the table itself
+            continue
+
+        unit = line.split('=')[2].strip()  # Contains '512 bytes'
+        unit = int(unit.split(' ')[0].strip())
+        break  # First line of the partition table header
+
+    for line in table_iter:
+        line = line.strip()
+        if not line.startswith('Device'):
+            continue
+
+        part_all_attrs = split_on_space_segments(line)
+        break
+
+    partitions = []
+    for partition_line in table_iter:
+        # Fields:               Device     Boot   Start      End  Sectors Size Id Type
+        # The line looks like: `/dev/vda1  *       2048  2099199  2097152   1G 83 Linux`
+        part_info = split_on_space_segments(partition_line)
+
+        # If the partition is not bootable, the Boot column might be empty
+        part_device = part_info[0]
+        part_start = int(part_info[2]) if len(part_info) == len(part_all_attrs) else int(part_info[1])
+        partitions.append(PartitionInfo(part_device=part_device, start_offset=part_start*unit))
+
+    return GRUBDevicePartitionLayout(device=device, partitions=partitions)
+
+
+def scan_grub_device_partition_layout():
+    grub_devices = next(api.consume(GrubInfo), None)
+    if not grub_devices:
+        return
+
+    for device in grub_devices.orig_devices:
+        dev_info = get_partition_layout(device)
+        if dev_info:
+            api.produce(dev_info)
--- a/repos/system_upgrade/el7toel8/actors/checkfirstpartitionoffset/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/checkfirstpartitionoffset/actor.py
@@ -1,0 +1,24 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import check_first_partition_offset
+from leapp.models import FirmwareFacts, GRUBDevicePartitionLayout
+from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckFirstPartitionOffset(Actor):
+    """
+    Check whether the first partition starts at the offset >=1MiB.
+
+    The alignment of the first partition plays role in disk access speeds. Older tools placed the start of the first
+    partition at cylinder 63 (due to historical reasons connected to the INT13h BIOS API). However, grub core
+    binary is placed before the start of the first partition, meaning that not enough space causes bootloader
+    installation to fail. Modern partitioning tools place the first partition at >= 1MiB (cylinder 2048+).
+    """
+
+    name = 'check_first_partition_offset'
+    consumes = (FirmwareFacts, GRUBDevicePartitionLayout,)
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag,)
+
+    def process(self):
+        check_first_partition_offset.check_first_partition_offset()
--- a/repos/system_upgrade/el7toel8/actors/checkfirstpartitionoffset/tests/test_check_first_partition_offset.py
+++ b/repos/system_upgrade/el7toel8/actors/checkfirstpartitionoffset/tests/test_check_first_partition_offset.py
@@ -1,0 +1,51 @@
+import pytest
+
+from leapp import reporting
+from leapp.libraries.actor import check_first_partition_offset
+from leapp.libraries.common import grub
+from leapp.libraries.common.testutils import create_report_mocked, CurrentActorMocked
+from leapp.libraries.stdlib import api
+from leapp.models import FirmwareFacts, GRUBDevicePartitionLayout, PartitionInfo
+from leapp.reporting import Report
+from leapp.utils.report import is_inhibitor
+
+
+@pytest.mark.parametrize(
+    ('devices', 'should_report'),
+    [
+        (
+            [
+                GRUBDevicePartitionLayout(device='/dev/vda',
+                                          partitions=[PartitionInfo(part_device='/dev/vda1', start_offset=32256)])
+            ],
+            True
+        ),
+        (
+            [
+                GRUBDevicePartitionLayout(device='/dev/vda',
+                                          partitions=[PartitionInfo(part_device='/dev/vda1', start_offset=1024*1025)])
+            ],
+            False
+        ),
+        (
+            [
+                GRUBDevicePartitionLayout(device='/dev/vda',
+                                          partitions=[PartitionInfo(part_device='/dev/vda1', start_offset=1024*1024)])
+            ],
+            False
+        )
+    ]
+)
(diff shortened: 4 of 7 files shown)


repos/system_upgrade/common/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
```diff
@@ -0,0 +1,64 @@
+from leapp import reporting
+from leapp.libraries.common import grub
+from leapp.libraries.stdlib import api, CalledProcessError, run
+
+SAFE_OFFSET_kiB = 1024  # 1MiB
+
+
+def is_first_partition_offset_ok(device):
+    try:
+        partition_table = run(['parted', device, 'unit', 'kiB', 'print'], split=True)['stdout']
```
dkubek, line 10: I would be for running ``parted`` here with the ``-m`` flag so we have a bit more guarantee on the output format. It does not simplify the parsing by much, but feels a bit safer at least to me.

repos/system_upgrade/common/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
```diff
@@ -0,0 +1,64 @@
...
+        if not is_first_partition_offset_ok(device):
+            problematic_devices.append(device)
+
+    if problematic_devices:
+        summary = ('GRUB devices with too little space reserved before the first partition were found. '
+                   'Such devices can present a problem when updating the boot loader. Problematic devices: \n'
+                   '{0}')
+        problematic_devices_fmt = ['- {0}'.format(dev) for dev in problematic_devices]
+
+        hint = 'Reformat the devices with so that there is at least {0} kiB space before the first partition.'
```
dkubek, line 55:
```suggestion
        hint = 'Reformat the devices so that there is at least {0} kiB space before the first partition.'
```

dkubek (commented):
I did a quick look over the code with no manual testing so far

repos/system_upgrade/common/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def is_first_partition_offset_ok(device):
+    try:
+        partition_table = run(['parted', device, 'unit', 'kiB', 'print'], split=True)['stdout']
+    except CalledProcessError:
+        # Unlikely - if the disk has no partition table, `parted` terminates with 0 (no err). Parted exits with an err
+        # when the device does not exists, or if it is too small to contain a partition table.
+        api.current_logger().error('Failed to run `parted` to obtain partition table of the device.')
+        return True  # There is no first partition, so the predicate holds vacuously
```
abadger, line 15: Two things herE:
* Add more information to the log,  For instance, the string representation of the exception and the name of the device.  If we encounter a case where we think parted should have found a partition table but it did not, we will want more information about what happened.
* Don't think we should `return True` here.  You list two cases that `parted` could exit with an error:
  * In the case the device does not exist, the caller should make the decision of what to do (It has to think about whether the devices it is passing in could be invalid or not and act appropriately)  So this case should raise an exception.
  * In the case the device is too small to contain a partition table, grub might also be too big to fit on the device, so I think this case should do one of the following:
    * `return False` (if the function is deciding whether there is space to place grub)
    * do some additional checking to see if there is room for grub before choosing whether to return True or False
    * or raise an exception for the caller to decide.

repos/system_upgrade/common/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
```diff
@@ -0,0 +1,64 @@
...
+
+    table_iter = iter(partition_table)
```
abadger, line 17: This is a neat idiom!  I hadn't seen this before but I can see the advantages to using the iterator.  Nice work.

repos/system_upgrade/common/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
```diff
@@ -0,0 +1,64 @@
...
+
+    for line in table_iter:
+        if not line.startswith('Number'):
+            # We are still reading general device information and not the table itself
+            continue
+        break  # First line of the partition table header
+
+    for partition_line in table_iter:
+        line_info = (fragment.strip() for fragment in partition_line.split('  '))
+        line_info = [fragment for fragment in line_info if fragment]
```
abadger, line 27: :+1: For proper use of a generator expression.

repos/system_upgrade/common/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
MichalHe: Thank you for your kind advice, however, I switched the patch to use `fdisk` instead of `parted` for the sake of not introducing needless dependencies.

repos/system_upgrade/common/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
MichalHe: Thanks! Applied.

repos/system_upgrade/common/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
MichalHe: Thank you for your kind words :bow:

MichalHe:
@pirat89 I have rewritten the code to use `fdisk` instead. I am against introducing needless dependencies and taking shortcuts. 

#### Beware that I have not tested the new solution yet.

repos/system_upgrade/common/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
MichalHe: Thank you :bow:

repos/system_upgrade/common/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
MichalHe: I've decided to raise exception that will eventually lead to leapp terminating with `StopActorExecutionError`. The reason is that the root cause of the problem -- the `get_grub_devices` function -- is also used later to update the grub core. If the function is (unintentionally) lying, then it is better to halt the upgrade and be aware of the problem, rather than attempting to install a boot loader to a device that, e.g., does not exist.

MichalHe:
@pirat89 I've manually tested the patch on RHEL7. On a VM with a disk formatted in a way so that the offset of the first partition is 63 sectors (the old way causing problems) the upgrade is inhibited due to the offset of the first partition being too small. With our vagrant box the upgrade was not inhibited (the offset of the first partition is 2048sectors (=1M)).

@oamg/developers This code is ready for a review, imho.

repos/system_upgrade/common/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
```diff
@@ -0,0 +1,96 @@
...
+    problematic_devices = []
+    for device in devices_to_check:
+        try:
+            is_ok = is_first_partition_offset_ok(device)
+            if not is_ok:
+                problematic_devices.append(device)
+        except CalledProcessError:
+            # The path does not point to an actual device, or the disk does not exists - grub.get_grub_devices lied?
+            msg = 'Failed to check the offset of the first partition of the device {0}'
+            raise StopActorExecutionError(msg.format(device))
```
abadger, line 79: Do we need to catch `ValueError` here as well?

repos/system_upgrade/common/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
```diff
@@ -0,0 +1,96 @@
...
+
+    if problematic_devices:
+        summary = ('GRUB devices with too little space reserved before the first partition were found. '
+                   'Such devices can present a problem when updating the boot loader. Problematic devices: \n'
+                   '{0}')
+        problematic_devices_fmt = ['- {0}'.format(dev) for dev in problematic_devices]
+
+        hint = 'Reformat the devices so that there is at least {0} kiB space before the first partition.'
```
pirat89, line 87: In case of summary, it would be nice to be specific, that failed update of the bootloader could lead to unbootable system, so they are not asking for the explanation what could happen. Also, we should specify clearly, that for in-place upgades it's required to start the first partition on sector 2048 or higher, to keep 1MB space before the first partition.

We will need to update this to be more specific. Especially the hint, as
* such an operation is not officially supported and customers are doing it on their own risk
* in some cases it does not have to be possible to do such an operation at all (e.g. if they used XFS for all partitions,  small amount of space on /boot already and shift of several partitions is problematic..., etc)
* also for number of our customers it's out of their skills to do something like that

So in this case, we should suggest also new clean installation of the target RHEL system with new partition layout.

pirat89:
The solution works, tested on my upgraded 6 ->  7 system:
```
                      REPORT OVERVIEW                       

    ...
    3. Found GRUB devices with too little space reserved before the first partition
    ...
...


### report:
Risk Factor: high (inhibitor)
Title: Found GRUB devices with too little space reserved before the first partition
Summary: GRUB devices with too little space reserved before the first partition were found. Such devices can present a problem when updating the boot loader. Problematic devices:
- /dev/vda
Remediation: [hint] Reformat the devices so that there is at least 1024 kiB space before the first partition.
Key: 98eb0d6d72263cb21ac172fc6a612e27e48a9e91

    
```

repos/system_upgrade/common/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
pirat89: todo(@pirat89 ): check whether KCS exists, in such a case we should add the link here too

repos/system_upgrade/common/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
pirat89: note: Something to think about are systems with GPT instead of MBR. I am thinking now whether it's "safe" to raise the error when we canot do the check, as I am little bit concerned of possible false positives that could block the upgrade and increase number of customer cases. So possibly log.warning that we cannot check whether system has enough free space in MBR gap for the upgrade (or maybe even nonblocking high sev. report? informing user to verify it on their own in such a case??) would be safer solution.

repos/system_upgrade/common/actors/checkfirstpartitionoffset/actor.py
```diff
@@ -0,0 +1,23 @@
...
+    The alignment of the first partition plays role in disk access speeds. Older tools placed the start of the first
+    partition at cylinder 63 (due to historical reasons connected to the INT13h BIOS API). However, grub core
+    binary is placed before the start of the first partition, meaning that not enough space causes bootloader
+    installation to fail. Modern partitioning tools place the first partition at >= 1MiB (cylinder 2048+).
+    """
+
+    name = 'check_first_partition_offset'
+    consumes = ()
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag,)
```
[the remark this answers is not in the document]
pirat89: btw, the problem with MBR gap is now expected to be happening just during IPU 7 -> 8 due to unfortunate fdisk default on RHEL 6. we expect that catching this for IPU  7 -> 8 cover these issues and nobody used a different value for the first partition that could be problematic (so they pass successfully 7 -> 8, but hit the problem for 8 -> 9). So it should be enough to check this just for IPU 7 -> 8 in el7toel8 repository. I am ok with common too, but also testing expect scenariou in 7 -> 8 only. So moving it to el7toel8 repo seems as better option. we can move it to back common if we realize related issues for 8 -> 9 too.

repos/system_upgrade/el7toel8/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
```diff
@@ -0,0 +1,30 @@
...
+        if first_partition.start_offset <= SAFE_OFFSET_BYTES:
+            problematic_devices.append(grub_dev.device)
+
+    if problematic_devices:
+        summary = ('GRUB devices with too little space reserved before the first partition were found. '
+                   'Such devices can present a problem when updating the boot loader. Problematic devices: \n'
+                   '{0}')
+        problematic_devices_fmt = ['- {0}'.format(dev) for dev in problematic_devices]
+
+        hint = 'Reformat the devices so that there is at least {0} kiB space before the first partition.'
```
pirat89, line 21:
```suggestion
        summary = (                                                            
            'On the system booting by using BIOS, the in-place upgrade fails ' 
            'when upgrading the GRUB2 bootloader if the boot disk\'s embedding area '
            'does not contain enough space for the core image installation. '  
            'This results in a broken system, and can occur when the disk has been '
            'partitioned manually, for example using the RHEL 6 fdisk utility.\n\n'
                                                                  
            'The list of devices with small embedding area:\n'    
            '{0}.'    
        )    
        problematic_devices_fmt = ['- {0}'.format(dev) for dev in problematic_devices]
    
        hint = (    
            'We recommend to perform a fresh installation of the RHEL 8 system '
            'instead of performing the in-place upgrade.\n'    
            'Another possibility is to reformat the devices so that there is ' 
            'at least {0} kiB space before the first partition. '    
            'Note that this operation is not supported and does not have to be '
            'always possible.'    
        )    
```

repos/system_upgrade/el7toel8/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
```diff
@@ -0,0 +1,30 @@
+from leapp import reporting
+from leapp.libraries.stdlib import api
+from leapp.models import GRUBDevicePartitionLayout
+
+SAFE_OFFSET_BYTES = 1024*1024  # 1MiB
+
+
+def check_first_partition_offset():
```
pirat89, line 8: suggestion:
* skip s390x
* skip if EFI

repos/system_upgrade/el7toel8/actors/scangrubdevpartitionlayout/libraries/scan_layout.py
```diff
@@ -0,0 +1,61 @@
...
+        part_device = part_info[0]
+        part_start = int(part_info[2]) if len(part_info) == len(part_all_attrs) else int(part_info[1])
+        partitions.append(PartitionInfo(part_device=part_device, start_offset=part_start*unit))
+
+    return GRUBDevicePartitionLayout(device=device, partitions=partitions)
+
+
+def scan_grub_device_partition_layout():
+    grub_devices = api.consume(GrubInfo)
+    for device in grub_devices.orig_devices:
```
pirat89, line 58: @MichalHe  The `GrubInfo` msg is not generated on s390x systems. Hence on these systems I expect that line 58 will try to invoke `orig_devices` on `None` type object. I suggest to apply the same skip mechanism as in case of the check actor, or at least skip on s390x arch. Another option would be e.g.
```suggestion
    grub_devices = api.consume(GrubInfo)
    if not grub_devices:
        return
    for device in grub_devices.orig_devices:
```

pirat89: Interesting, I wouldn't expect my mind to be so focused so late hour :D :
```
    for device in grub_devices.orig_devices:
AttributeError: 'generator' object has no attribute 'orig_devices'
```

MichalHe: Nice catch :+1: . I've wrapped it with `next` as suggested

repos/system_upgrade/common/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
pirat89: done. The `ValueError` is not raised anymore and this the code has been refactored.

pirat89:
@MichalHe  I have bad karma I guess. It seems that CI caught FP inhibitor on AWS. we should check that

repos/system_upgrade/el7toel8/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py
```diff
@@ -0,0 +1,52 @@
...
+        return
+
+    for fact in api.consume(FirmwareFacts):
+        if fact.firmware == 'efi':
+            return  # Skip EFI system
+
+    problematic_devices = []
+    for grub_dev in api.consume(GRUBDevicePartitionLayout):
+        first_partition = min(grub_dev.partitions, key=lambda partition: partition.start_offset)
+        if first_partition.start_offset <= SAFE_OFFSET_BYTES:
```
pirat89, line 20: aah!
```
if 1048576 <= 1048576:
```

pirat89: fixed. waiting for finished tests.

pirat89:
The last one? :) Actually, the GPT is really the trouble maker in the end.

```
2024-04-23 20:01:03.901 INFO     PID: 11755 leapp.workflow.Checks: Executing actor check_first_partition_offset 
Process Process-386:
    self.run()
  File "/usr/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 74, in _do_run
  File "/usr/lib/python2.7/site-packages/leapp/actors/__init__.py", line 289, in run
    self.process(*args)
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/checkfirstpartitionoffset/actor.py", line 24, in process
    check_first_partition_offset.check_first_partition_offset()
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py", line 19, in check_first_partition_offset
    first_partition = min(grub_dev.partitions, key=lambda partition: partition.start_offset)
ValueError: min() arg is an empty sequence
2024-04-23 20:01:03.952 ERROR    PID: 11755 leapp.workflow.Checks: Actor check_first_partition_offset has crashed: Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 74, in _do_run
  File "/usr/lib/python2.7/site-packages/leapp/actors/__init__.py", line 289, in run
    self.process(*args)
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/checkfirstpartitionoffset/actor.py", line 24, in process
    check_first_partition_offset.check_first_partition_offset()
  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/checkfirstpartitionoffset/libraries/check_first_partition_offset.py", line 19, in check_first_partition_offset
    first_partition = min(grub_dev.partitions, key=lambda partition: partition.start_offset)
ValueError: min() arg is an empty sequence
```

The msg has no partitions at all:
```
Stamp: 2024-04-23T20:00:59.709122Z
Actor: scan_grub_device_partition_layout
Phase: FactsCollection
Type: GRUBDevicePartitionLayout
Message_data:
{
    "device": "/dev/xvda",
    "partitions": []
}
```

In the actor we can see that it's caused by using GPT (and prep boot partition) instead of MBR for BIOS boot.
```
    2024-04-23 20:00:59.682 DEBUG    PID: 15396 leapp.workflow.FactsCollection.scan_grub_device_partition_layout: External command has started: ['fdisk', '-l', '-u=sectors', u'/dev/xvda']
    2024-04-23 20:00:59.689 DEBUG    PID: 15396 leapp.workflow.FactsCollection.scan_grub_device_partition_layout: WARNING: fdisk GPT support is currently new, and therefore in an experimental phase. Use at your own discretion.
    2024-04-23 20:00:59.691 DEBUG    PID: 15396 leapp.workflow.FactsCollection.scan_grub_device_partition_layout: 
    2024-04-23 20:00:59.692 DEBUG    PID: 15396 leapp.workflow.FactsCollection.scan_grub_device_partition_layout: Disk /dev/xvda: 268.4 GB, 268435456000 bytes, 524288000 sectors
    2024-04-23 20:00:59.693 DEBUG    PID: 15396 leapp.workflow.FactsCollection.scan_grub_device_partition_layout: Units = sectors of 1 * 512 = 512 bytes
    2024-04-23 20:00:59.695 DEBUG    PID: 15396 leapp.workflow.FactsCollection.scan_grub_device_partition_layout: Sector size (logical/physical): 512 bytes / 512 bytes
    2024-04-23 20:00:59.696 DEBUG    PID: 15396 leapp.workflow.FactsCollection.scan_grub_device_partition_layout: I/O size (minimum/optimal): 512 bytes / 512 bytes
    2024-04-23 20:00:59.697 DEBUG    PID: 15396 leapp.workflow.FactsCollection.scan_grub_device_partition_layout: Disk label type: gpt
    2024-04-23 20:00:59.699 DEBUG    PID: 15396 leapp.workflow.FactsCollection.scan_grub_device_partition_layout: Disk identifier: 2D3B54F1-2833-4F75-9BB7-C3A39B53EE7C
    2024-04-23 20:00:59.699 DEBUG    PID: 15396 leapp.workflow.FactsCollection.scan_grub_device_partition_layout: 
    2024-04-23 20:00:59.700 DEBUG    PID: 15396 leapp.workflow.FactsCollection.scan_grub_device_partition_layout: 
    2024-04-23 20:00:59.701 DEBUG    PID: 15396 leapp.workflow.FactsCollection.scan_grub_device_partition_layout: #         Start          End    Size  Type            Name
    2024-04-23 20:00:59.703 DEBUG    PID: 15396 leapp.workflow.FactsCollection.scan_grub_device_partition_layout:  1         2048         4095      1M  BIOS boot       
    2024-04-23 20:00:59.704 DEBUG    PID: 15396 leapp.workflow.FactsCollection.scan_grub_device_partition_layout:  2         4096    524287965    250G  Linux filesyste 
    2024-04-23 20:00:59.707 DEBUG    PID: 15396 leapp.workflow.FactsCollection.scan_grub_device_partition_layout: External command has finished: ['fdisk', '-l', '-u=sectors', u'/dev/xvda']
```

pirat89:
Created separate commit to handle the previousl described issue. The solution has 2 parts:
* scanner: skip disks with GPT (do not produce the msg for them; just log info)
* check: skip check of grub devices with empty list of partitions (seatbelt)

As described here:
* https://serverfault.com/questions/1121408/why-is-disk-label-type-dos-but-the-partition-system-is-gpt

Btw, disks with GPT could be also reported in some cases as "dos" instead:
* https://serverfault.com/questions/1121408/why-is-disk-label-type-dos-but-the-partition-system-is-gpt
The output format is then equal to the real DOS partition as it looks like so the parsing should not be affected in such a case.

I have also exteded a little bit original test cases, but the current new solution is not completely covered by unit-tests. I am testing this manually on my setups right now + CI set. So I believe we could call it good enough if all tests will pass

abadger (approved):

pirat89:
My manual testing results after the last changes:
* upgraded RHEL 6 machine with small MBR gap: detected (see report example above, msg has not been changed)
* RHEL 7 BIOS boot with GPT: passed
* RHEL 7 BIOS boot with MBR (and standard first partition on 2048 sector): passed
* RHEL 7 EFI: passed
* RHEL 7 on s390x: passed

PR #1196, opened 2024-04-04
check_microarch: require x86_64-v3 for RHEL10

MichalHe:
This patch extends the microarchitecture checks to require `x86_64-v3` CPU features when upgrading to RHEL10. The list of flags has been compiled based on the `x86_64-v3` ABI description found [here](https://gitlab.com/x86-psABIs/x86-64-ABI). The ABI CPU features were then translated to the corresponding names used by Linux kernel as found in [cpufeatures.h](https://github.com/torvalds/linux/blob/master/arch/x86/include/asm/cpufeatures.h).

I have verified that the list of should be correct by making sure that the upgrade will not be inhibited on my relatively new CPU (i.e. my CPU has all the required flags).


repos/system_upgrade/common/actors/checkmicroarchitecture/libraries/checkmicroarchitecture.py
```diff
@@ -0,0 +1,77 @@
...
+    if not microarch_info:
+        api.current_logger().info('No known microarchitecture requirements are known for target RHEL%s.',
+                                  get_target_major_version())
+        return
+
+    missing_flags = [flag for flag in microarch_info.required_flags if flag not in cpuinfo.flags]
+    api.current_logger().debug('Required flags missing: %s', missing_flags)
+    if missing_flags:
+        _inhibit_upgrade(missing_flags,
+                         'RHEL{0}'.format(get_target_major_version()),
```
pirat89, line 75: nitpick `RHEL9` -> `RHEL 9`
```suggestion
                         'RHEL {0}'.format(get_target_major_version()),
```

pirat89: @MichalHe  wait a sec with the change.

repos/system_upgrade/common/actors/checkmicroarchitecture/libraries/checkmicroarchitecture.py
```diff
@@ -0,0 +1,77 @@
...
+    )
+
+    rhel_major_to_microarch_reqs = {
+        '9': MicroarchInfo(microarch_ver='x86-64-v2',
+                           required_flags=(X86_64_BASELINE_FLAGS + X86_64_V2_FLAGS),
+                           extra_report_fields=[rhel9_microarch_article]),
+        '10': MicroarchInfo(microarch_ver='x86-64-v3',
+                            required_flags=(X86_64_BASELINE_FLAGS + X86_64_V2_FLAGS + X86_64_V3_FLAGS),
+                            extra_report_fields=[]),
+    }
```
pirat89, line 63: I've added a note to the OAMG-11275 ticket to not forget to implement the shortened URL when a documentation is created.

pirat89 (approved):
lgtm. keeping unmerged for now to test it manually yet.

pirat89:
works as expected. tested 7 -> 8 -> 9 with IvyBridge-IBRS:

```
Risk Factor: high (inhibitor)
Title: Current x86-64 microarchitecture is unsupported in RHEL10
Summary: RHEL10 has a higher CPU requirement than older versions, it now requires a CPU compatible with x86-64-v3 instruction set or higher.

Missings flags detected are: avx2, bmi1, bmi2, fma, abm, movbe

Remediation: [hint] If a case of using virtualization, virtualization platforms often allow configuring a minimum denominator CPU model for compatibility when migrating between different CPU models. Ensure that minimum requirements are not below that of RHEL10

Key: d11105984cabbcb2e199eb0499a9b97035d44c99
```

PR #1200, opened 2024-04-16
Add LUKS support via Clevis TPM 2 token

danzatt:
tldr; Add a `LuksScanner` actor which scans all crypt devices using `cryptsetup luksDump`. Don't inhibit, when all devices are LUKS2 with clevs TPM2 token.

    So far, upgrades with encrypted drives were not supported. Encrypted
    drives require interactively typing unlock passphrases, which is not
    suitable for automatic upgrades using Leapp. We add a feature, where
    systems with all drives configured with automatic unlock method can be
    upgraded.
    
    Currently, we only support drives configured with Clevis/TPM2 token,
    because networking is not configured during Leapp upgrade (excluding
    NBDE).
    
    We consume LuksDumps message to decide whether the upgrade process
    should be inhibited. If there is at least one LUKS2 device without
    Clevis TPM2 binding, we inhibit the upgrade because we cannot tell if
    the device is not a part of a more complex storage stack and the failure
    to unlock the device migt cause boot problem.
    
    Note that IPU 7 -> 8 with encrypted devices stays unsupported / inhibited.
    The new solution is applied only for upgrades from RHEL 8+.

jira: RHEL-3294

This PR introduces new shortened URLs:
* https://red.ht/clevis-tpm2-luks-auto-unlock-rhel8
* https://red.ht/clevis-tpm2-luks-auto-unlock-rhel9
* https://red.ht/convert-to-luks2-rhel8
* https://red.ht/convert-to-luks2-rhel9

/cc @pirat89


repos/system_upgrade/common/actors/inhibitwhenluks/actor.py
```diff
@@ -1,9 +1,15 @@
...
+CLEVIS_RHEL9_DOC_URL = 'https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/security_hardening/configuring-automated-unlocking-of-encrypted-volumes-using-policy-based-decryption_security-hardening#configuring-manual-enrollment-of-volumes-using-tpm2_configuring-automated-unlocking-of-encrypted-volumes-using-policy-based-decryption'  # noqa: E501; pylint: disable=line-too-long
+LUKS2_CONVERT_RHEL8_DOC_URL = 'https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/security_hardening/encrypting-block-devices-using-luks_security-hardening#luks-versions-in-rhel_encrypting-block-devices-using-luks'  # noqa: E501; pylint: disable=line-too-long
+LUKS2_CONVERT_RHEL9_DOC_URL = 'https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/security_hardening/encrypting-block-devices-using-luks_security-hardening#luks-versions-in-rhel_encrypting-block-devices-using-luks'  # noqa: E501; pylint: disable=line-too-long
+
```
pirat89, line 12: candidates for shortened URLs

* shortened URLs allow us to react on possible changes in access.redhat.com which could result in 404 errors, etc. you do not need to care about shortened URLs now, we will create them later when wanted.

pirat89:
@danzatt  Hi Dan \o most likely I will get to the review during early May or later June. We are dealing now with additional stuff.

repos/system_upgrade/common/models/luksdump.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+
+    token_type = fields.String()
+    """
+    Type of the token. For "clevis" type the concrete subtype (determined using
+    clevis luks list) is appended e.g. clevis-tpm2. clevis-tang, ...
+    """
+
+
+class LuksDump(Model):
```
pirat89, line 28: Checking the logic, I suggest to create one msg that will envelope all created LuksDump (list(Model(LuksDump)). It has some benefits:
* you can check that only one msg exists at all (e.g. no other actors, like custom/third-party, are produced the msg too)
* kind of easier debugging when checking leapp.db, as you find the msg and just go through the data; however for multiple msgs it's kind of more noise - you have more reading actually

It's not blocker, but I suggest to consider the change. You can find both principles used in the project, so no blocker at all. But from my experience having just one msg for "system facts" is little bit more comfortable. Keeping it on you what you is your preference.

pirat89 (changes requested):

pirat89:
```
# leapp preupgrade --debug --channel ga --no-rhsm
...
# sqlite3 /var/lib/leapp/leapp.db
sqlite> select * from execution;
...
7|1bba4374-2a9b-4bd0-a917-3178474d3a63|2024-05-22T09:55:13.396744Z|{"cmd": ["/usr/bin/leapp", "preupgrade", "--debug", "--channel", "ga", "--no-rhsm"], "debug": "1", "environment": {"LEAPP_DEBUG": "1", "LEAPP_DEVEL_DATABASE_SYNC_OFF": "1", "LEAPP_DEVEL_USE_PERSISTENT_PACKAGE_CACHE": "1", "LEAPP_EXPERIMENTAL": "0", "LEAPP_HOSTNAME": "localhost.localdomain", "LEAPP_IPU_IN_PROGRESS": "7to8", "LEAPP_NO_RHSM": "1", "LEAPP_NO_RHSM_FACTS": "1", "LEAPP_TARGET_PRODUCT_CHANNEL": "ga", "LEAPP_UNSUPPORTED": "1", "LEAPP_UPGRADE_PATH_FLAVOUR": "default", "LEAPP_UPGRADE_PATH_TARGET_RELEASE": "8.10", "LEAPP_VERBOSE": "1"}, "verbose": "1", "whitelist_experimental": []}|preupgrade
```

danzatt:
Hello @pirat89 thanks for the review!  I've hopefully addressed all your remarks now.

repos/system_upgrade/common/actors/inhibitwhenluks/libraries/inhibitwhenluks.py
```diff
@@ -0,0 +1,119 @@
...
+        luks2_convert_doc_url = LUKS2_CONVERT_RHEL9_DOC_URL
+    else:
+        create_report([
+            reporting.Title('LUKS encrypted partition detected'),
+            reporting.Summary('Upgrading system with encrypted partitions is not supported'),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Groups([reporting.Groups.BOOT, reporting.Groups.ENCRYPTION]),
+            reporting.Groups([reporting.Groups.INHIBITOR]),
+        ])
+        return
```
pirat89, line 34: @danzatt  is there such a difference for 9 -> 10 upgrades? just curious whether the current solution would be broken there or not.  Maybe we discussed that but I do not remember it now, so I am rather asking.

repos/system_upgrade/common/actors/luksscanner/libraries/luksscanner.py
```diff
@@ -0,0 +1,124 @@
...
+                   'to determine clevis token type.')
+        details = {'hint': 'Use dnf to install the "clevis" and "clevis-luks" package.'}
+        raise StopActorExecutionError(message=message, details=details)
+    except stdlib.CalledProcessError as e:
+        api.current_logger().debug("clevis list command failed with an error code: {}".format(e.exit_code))
+
+        message = ('The "clevis luks list" command failed. This'
+                   'might be because the clevis-luks package is'
+                   'missing on your system.')
+        details = {'hint': 'Use dnf to install the "clevis" and "clevis-luks" package.'}
```
pirat89, line 37: Seems that in this case should be missing only the clevis-luks package.
```suggestion
        details = {'hint': 'Use dnf to install the "clevis-luks" package.'}
```

repos/system_upgrade/common/actors/luksscanner/libraries/luksscanner.py
```diff
@@ -0,0 +1,124 @@
...
+    Assuming the device is initialized using clevis, determine the type of
+    clevis token associated to the specified keyslot.
+    """
+    try:
+        result = stdlib.run(["clevis", "luks", "list", "-d", device_path, "-s", str(keyslot)])
+    except OSError:
+        message = ('A LUKS drive with clevis token was discovered, but there is '
+                   'no clevis package installed. The clevis command is required '
+                   'to determine clevis token type.')
+        details = {'hint': 'Use dnf to install the "clevis" and "clevis-luks" package.'}
```
pirat89, line 29:
```suggestion
        details = {'hint': 'Use dnf to install the "clevis" and "clevis-luks" packages.'}
```

pirat89 (changes requested):
It's mostly ok. Just some minor changes should be done. Let's sync during the day to discuss it.

repos/system_upgrade/common/actors/inhibitwhenluks/libraries/inhibitwhenluks.py
```diff
@@ -0,0 +1,119 @@
...
+CLEVIS_RHEL9_DOC_URL = 'https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/security_hardening/configuring-automated-unlocking-of-encrypted-volumes-using-policy-based-decryption_security-hardening#configuring-manual-enrollment-of-volumes-using-tpm2_configuring-automated-unlocking-of-encrypted-volumes-using-policy-based-decryption'  # noqa: E501; pylint: disable=line-too-long
+LUKS2_CONVERT_RHEL8_DOC_URL = 'https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/security_hardening/encrypting-block-devices-using-luks_security-hardening#luks-versions-in-rhel_encrypting-block-devices-using-luks'  # noqa: E501; pylint: disable=line-too-long
+LUKS2_CONVERT_RHEL9_DOC_URL = 'https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/security_hardening/encrypting-block-devices-using-luks_security-hardening#luks-versions-in-rhel_encrypting-block-devices-using-luks'  # noqa: E501; pylint: disable=line-too-long
```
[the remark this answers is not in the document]
pirat89: discussed and resolved

repos/system_upgrade/common/actors/inhibitwhenluks/libraries/inhibitwhenluks.py
```diff
@@ -0,0 +1,119 @@
...
+
+def _at_least_one_tpm_token(luks_dump):
+    return any([token.token_type == "clevis-tpm2" for token in luks_dump.tokens])
+
+
+def check_invalid_luks_devices():
+    target_major_version = get_target_major_version()
+    if target_major_version == '8':
+        clevis_doc_url = CLEVIS_RHEL8_DOC_URL
+        luks2_convert_doc_url = LUKS2_CONVERT_RHEL8_DOC_URL
```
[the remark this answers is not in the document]
pirat89: discussed and resolved.

pirat89 (dismissed):
I will wait for another co-review before merging as I helped with some parts of the code but for now I consider it ready to merge from my side. I will possibly do some additional testing before the merge too.

repos/system_upgrade/common/models/luksdump.py
```diff
@@ -0,0 +1,73 @@
...
+
+    tokens = fields.List(fields.Model(LuksToken), default=[])
+    """
+    List of LUKS2 tokens
+    """
+
+
+class LuksDumps(Model):
+    """
+    Information about a all LUKS-encrypted devices on the system.
```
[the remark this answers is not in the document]
pirat89: thanks. fixed.

pirat89 (dismissed):

repos/system_upgrade/common/actors/inhibitwhenluks/libraries/inhibitwhenluks.py
```diff
@@ -0,0 +1,165 @@
...
+    if no_tpm2_partitions:
+        summary += (
+            '\n\nCurrently we require the process to be non-interactive and'
+            ' offline. For this reason we require automatic unlock of'
+            ' encrypted devices during the upgrade process.'
+            ' Currently we support automatic unlocking during the upgrade only'
+            ' for volumes bound to Clevis TPM2 token.'
+            ' The following LUKS2 devices without Clevis TPM2 token '
+            ' have been discovered on your system: {}'
+            .format(_formatted_list_output(no_tpm2_partitions))
```
dkubek, line 97: The current output looks something like this:
```
Currently we require the process to be non-interactive and offline. For this reason we require automatic unlock of encrypted devices during the upgrade process. Currently we support automatic unlocking during the upgrade only for volumes bound to Clevis TPM2 token. The following LUKS2 devices without Clevis TPM2 token  have been discovered on your system: ['\n    - vda3']
```
(note the unformatted list at the end)

It was probably meant to be something like
```suggestion
            .format(''.join(_formatted_list_output(no_tpm2_partitions)))
```
or put the join in the ``_formatted_list_output`` itself

repos/system_upgrade/common/actors/luksscanner/libraries/luksdump_parser.py
```diff
@@ -0,0 +1,195 @@
...
+        all lines are inserted.
+        """
+
+        def __init__(self, indented_line):
+            self.children = []
+            self.level = len(indented_line) - len(indented_line.lstrip())
+            self.text = indented_line.strip()
+
+        def add_children(self, nodes):
+            childlevel = nodes[0].level
```
dkubek, line 20: ``nodes[0]`` asumes non-empty list. This is probably not a problem, since we assume non-empty output from the command, but this fails otherwise

repos/system_upgrade/common/actors/inhibitwhenluks/actor.py
```diff
@@ -13,37 +13,9 @@ class InhibitWhenLuks(Actor):
     """
 
     name = 'check_luks_and_inhibit'
-    consumes = (StorageInfo, CephInfo)
-    produces = (Report,)
+    consumes = (CephInfo, LuksDumps, StorageInfo)
+    produces = (Report, TargetUserSpaceUpgradeTasks)
```
dkubek, line 17: Missing ``UpgradeInitramfsTasks``

dkubek:
I went over the code and up to some details it looks good. I manually tested that the upgrade is inhibited on 8->9. Also performed a successful upgrade on a VM with TPM.

repos/system_upgrade/common/actors/inhibitwhenluks/actor.py
pirat89: fixed. as you found it worked even without this msg produced, probably it's just seatbelt and dracut modules are automatically included when rpms are installed inside the container. but let's keep the seatbelt and make it active :-D fixed

repos/system_upgrade/common/actors/inhibitwhenluks/libraries/inhibitwhenluks.py
```diff
@@ -0,0 +1,165 @@
...
+
+    if luks1_partitions:
+
+        summary += (
+            '\n\nSince RHEL 8 the default format for LUKS encryption is LUKS2.'
+            ' Despite the old LUKS1 format is still supported on RHEL systems'
+            ' it has some limitations in comparison to LUKS2 and during initial'
+            ' investigation we have decided to support for upgrades only LUKS2'
+            ' format.'
+            ' The following LUKS1 partitions have been discovered on your system:{}'
```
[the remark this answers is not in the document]
pirat89: good point! updated

repos/system_upgrade/common/actors/luksscanner/libraries/luksscanner.py
```diff
@@ -0,0 +1,124 @@
...
+    try:
+        result = stdlib.run(['cryptsetup', 'luksDump', device_path])
+        luksdump_dict = LuksDumpParser.parse(result["stdout"].splitlines())
+
+        version = int(luksdump_dict["Version"]) if "Version" in luksdump_dict else None
+        uuid = luksdump_dict["UUID"] if "UUID" in luksdump_dict else None
+        if version is None or uuid is None:
+            api.current_logger().error(
+                'Failed to detect UUID or version from the output "cryptsetup luksDump {}" command'.format(device_path)
+            )
```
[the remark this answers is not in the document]
pirat89: yes, kind of it is. but I rather add the log so it's visible also in the log file. let's keep it, but add more details into the raised error.

repos/system_upgrade/common/actors/luksscanner/libraries/luksscanner.py
```diff
@@ -0,0 +1,124 @@
...
+                tokens=_get_tokens(device_path, luksdump_dict)
+        )
+
+    except (OSError, stdlib.CalledProcessError) as ex:
+        api.current_logger().error(
+                'Failed to execute "cryptsetup luksDump" command: {}'.format(ex)
+        )
+        raise StopActorExecutionError(
+            'Failed to execute "cryptsetup luksDump {}" command'.format(device_path)
+        )
```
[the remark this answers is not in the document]
pirat89: updated, added details into the raised error.

repos/system_upgrade/common/actors/luksscanner/libraries/luksdump_parser.py
pirat89: I see your point. seems to me ok as to my understanding it should by always non-empty and I guess we could have a different problems somewhere else otherwise. I am thinking to raise an exception to catch the error and act properly, but otoh I do not know what else to do and what error msg to provide to customer. As it's more hypothetic, let's just add a comment and see if it ever appears - as in both cases, I would stop the upgrade raising an error anyway right now.

pirat89: thinking about it twice. I do not know whether to use hard blocker (raising an error anyway) or whether to just log a warning and return from the function. let's add for now a comment note and discuss it later.

pirat89: hmmm..interesting. why it's not marked as obsoleted when I changed this part of the code? :eyes:

pirat89: anyway. note added.

repos/system_upgrade/common/actors/luksscanner/libraries/luksscanner.py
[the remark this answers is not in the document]
matejmatuska: Ok I agree, it was just a suggestion

matejmatuska (approved):
Looks alright now, I also went over the changes after @dkubek's review and seems that everything is addressed, but leaving that up to him to approve.

Just note that I dind't test this as it has already been tested by @pirat89 and @dkubek and there weren't functional changes since.

pirat89:
From manual testing (VM with emulated tpm2):
* LUKS1: inhibited
* converted to LUKS2 (without clevis+tpm2 binding): inhibited
* binded to clevis+tpm2: passed
* upgrade: passed

Adding some notes for simpler reproduce of steps I ddi:
* convert L1 -> L2, for usual minimal setup on VM:
```
# add cryptsetup to initramfs so the conversion from initramfs is possible
dracut -f --install cryptsetup
reboot

# after reboot, update kernel-cmdline: rd.break=initqueue
# once in emergency console (typically for minimal VM):
cryptsetup convert --type luks2 /dev/vda2
```

* bind clevis+tpm2
```
# install required rpms
dnf -y install clevis-luks clevis-dracut

# set the binding - note that possibly you will want to set also specific PCR register - in this case, using just default
clevis luks bind -yd /dev/vda2 tpm2 '{"hash":"sha256","key":"rsa"}'
dracut -f --regenerate-all
```

pirat89:
@danzatt :100:  :fireworks:  it's finally over! thanks for your contribution and help!!

PR #1202, opened 2024-04-17
[experimental] Upgrade dracut module: Update /usr mounting solution

pirat89:
Originally we had implemented our own mount_usr.sh script, which took care about mounting the /usr when it is present on separate partition / mountpoint. It took care also about LVM activation.

However, it has been problematic in various cases (e.g. when device needed more time for initialisation - e.g. when connected using FC). Let's use instead existing system solutions, starting the upgrade.target after initrd-fs.target (instead of just basic.target).

Jira-ref: RHEL-3344, RHEL-35446

Cooperating with: prajnoha, lnykryn, pvalena.

do_not_merge: just exprimenting right now. if we find working solution, I will update yet commit msg at minimum.

### 1. LVM Autoactivation
The device that is mounted at /usr might be an LVM partition, so the proposed solution needs to handle LVM activation (the old implementation already did so). To this end, we include the `69-dm-lvm.rules` udev rule into the upgrade initramfs. These rules autoactivate LVM volumes when a device is discovered by the kernel and the udev daemon is notified about them. This mechanism is borrowed from a standard boot process. For more details see [lvmautoactivation (7)](https://man.archlinux.org/man/core/lvm2/lvmautoactivation.7.en).

### 2. Mounting the filesystem according to /etc/fstab
The goal is to use a mounting mechanism that is as close as possible to an ordinary boot process of a RHEL system. Inside the initramfs, the boot process looks as follows (taken from [bootup](https://www.freedesktop.org/software/systemd/man/latest/bootup.html)):
```
                               : (beginning identical to above)
                               :
                               v
                         basic.target
                               |                       emergency.service
        ______________________/|                               |
       /                       |                               v
       |            initrd-root-device.target          emergency.target
       |                       |
       |                       v
       |                  sysroot.mount
       |                       |
       |                       v
       |             initrd-root-fs.target
       |                       |
       |                       v
       v            initrd-parse-etc.service
(custom initrd                 |
 services...)                  v
       |            (sysroot-usr.mount and
       |             various mounts marked
       |               with fstab option
       |              x-initrd.mount...)
       |                       |
       |                       v
       |                initrd-fs.target
       \______________________ |
                              \|
                               v
                          initrd.target
                               |
                               v
                     initrd-cleanup.service
                          isolates to
                    initrd-switch-root.target
                               |
                               v
        ______________________/|
       /                       v
       |        initrd-udevadm-cleanup-db.service
       v                       |
(custom initrd                 |
 services...)                  |
       \______________________ |
                              \|
                               v
                   initrd-switch-root.target
                               |
                               v
                   initrd-switch-root.service
                               |
                               v
                     Transition to Host OS
```
One of the most important parts is `initrd-parse-etc`, which calls `systemd-fstab-generator` to parse `/etc/fstab` and generate mount units that mount the filesystem tree as described by `/etc/fstab` when possible (device is available, mount target exists, etc.). However, triggering the service in initramfs will not help us, since the `/etc/fstab` in initramfs is different than the one used on the source system. Moreover, we a slightly different behavior than the one described by source system's fstab: instead of mounting device `X` at `/usr`, we want to mount it at `/sysroot/usr`.

The solution proposed by this PR is to manually run `systemd-fstab-generator` to generate corresponding mount units before reboot. These units are modified according to our needs (`/usr` -> `/sysroot/usr`), and, then we include them in our upgrade initramfs. These mount units are then mounted, since they are marked as a dependency of `local-fs.target`.

## Notes for reviewers
Modify the created upgrade boot entry by running the following command.
```
grubby --update-kernel=/boot/vmlinuz-upgrade.x86_64 --args rd.upgrade.break=leapp-pre-upgrade
```
After rebooting, you will be dropped into a shell that allows you to check the state of the system before proceeding with the upgrade. Check that everything is mounted as expected under `/sysroot`, and that the created mount units are active without any problems:
```
mount | grep sysroot
systemctl status '*.mount'  # Everything should be green
systemctl status '*.swap'  # Everything should be green
```

## Additional notes
 * discovered that we need to use `initrd-parse-etc.service`, but the inside the unit file:
   ```
   ExecStart=/usr/bin/systemctl --no-block start initrd-cleanup.service
   ```
   The `initrd-cleanup` service includes also switch-root, which we do not want to happen at all and due to isolation it basically kills out upgrade process.
 * the service has been changed in various RHEL systems, so creating own service would mean we should most likely copy couple of them (per each change in RHEL systems) and watch whether it's changed in future
 * going now with the original workaround #1218 . We are going to continue to investigate the proper solution for future releases in IPU 8 -> 9 )and newer upgrade paths)

Update:
* Discussed the issue with systemd guys, we have decided to override the initrd-cleanup service inside the upgrade initramfs. Several initial tests seems to provide positive results. (Just discovered a problem, I will investigate it later)
* Seems I forget to apply a change I did in the past locally as I could not get to the state where I get when I discovered the problem that should be fixed now. It's again in the state that only /usr is handled but not the rest of fstab. I need to check it one more time what is wrong with the upgrade service definition.


pirat89:
@MichalHe I have reproduced the problem with:

* fstab
```
#
# /etc/fstab
# Created by anaconda on Thu May  2 18:40:01 2024
#
# Accessible filesystems, by reference, are maintained under '/dev/disk'
# See man pages fstab(5), findfs(8), mount(8) and/or blkid(8) for more info
#
UUID=fc227809-1070-4d86-b676-54cb81a5455b /boot                   xfs     defaults        0 0
/dev/mapper/rhel01-usr  /usr                    xfs     defaults        0 0
/dev/mapper/rhel00-var  /var                    xfs     defaults        0 0
/dev/mapper/rhel-swap   swap                    swap    defaults        0 0
```

* partitioning:
```
# System bootloader configuration
bootloader --append=" crashkernel=auto" --location=mbr --boot-drive=vda
# Partition clearing information
clearpart --none --initlabel
# Disk partitioning information
part pv.1349 --fstype="lvmpv" --ondisk=vda --size=7172
part pv.146 --fstype="lvmpv" --ondisk=vda --size=6148
part /boot --fstype="xfs" --ondisk=vda --size=1024
part pv.726 --fstype="lvmpv" --ondisk=vda --size=8196
volgroup rhel00 --pesize=4096 pv.726
volgroup rhel01 --pesize=4096 pv.1349
volgroup rhel --pesize=4096 pv.146
logvol /  --fstype="xfs" --size=3072 --name=root --vgname=rhel
logvol /var  --fstype="xfs" --size=7168 --name=var --vgname=rhel00
logvol /usr  --fstype="xfs" --size=6144 --name=usr --vgname=rhel01
logvol swap  --fstype="swap" --size=2048 --name=swap --vgname=rhel
```

* lsblk
```
[root@localhost ~]# lsblk
NAME           MAJ:MIN RM SIZE RO TYPE MOUNTPOINT
vda            252:0    0  30G  0 disk 
├─vda1         252:1    0   1G  0 part /boot
├─vda2         252:2    0   8G  0 part 
│ └─rhel00-var 253:3    0   7G  0 lvm  /var
├─vda3         252:3    0   7G  0 part 
│ └─rhel01-usr 253:2    0   6G  0 lvm  /usr
├─vda4         252:4    0   1K  0 part 
└─vda5         252:5    0   6G  0 part 
  ├─rhel-root  253:0    0   3G  0 lvm  /
  └─rhel-swap  253:1    0   2G  0 lvm  [SWAP]
vdb            252:16   0  20G  0 disk 
```

MichalHe:
@pirat89 By default, systemd will mount only `/` and `/usr` from `/sysroot/etc/fstab`. If we want to have any other device mounted early by systemd, we need to add the `x-initrd.mount` option to corresponding fstab entries. Any such a unit will be picked up and mounted automatically. So, in the reproducer you have shared, I need to modify
```
/dev/mapper/rhel00-var  /var                    xfs     defaults        0 0
```
into
```
/dev/mapper/rhel00-var  /var                    xfs     defaults,x-initrd.mount        0 0
```

Also, if the device is an LVM's logical volume and the source system is using `rd.lvm.lv` cmdline args, we need to add anything we will be mounting the the list. So, I needed to add `rd.lvm.lv=rhel00/var` to the upgrade boot entry in order for the corresponding `/dev/mapper/...` entry to be present so that the unit generated by systemd could be activated. Alternatively, we can just remove all `rd.lvm.lv` cmdline args and dracut will make all lvm volumes available.

Otherwise, the upgrade went OK.

MichalHe:
So, the issue is essentially twofold:
1) We stop the standard boot process early, and, therefore, we might be stranded without necessary LVs. For example, if one uses `rd.lvm.lv=..` to instruct dracut to _only_ activate volumes for `/` and `/usr`, and there is a LV for `/var` none will take care of this for us. Luckily, the [lvmautoactivation](https://man.archlinux.org/man/core/lvm2/lvmautoactivation.7.en) manpage does a great job in explaining how the process works. So, to obtain some magic that will autoactivate LVM for us, we just need:
```
LEAPP_DRACUT_INSTALL_FILES="/usr/sbin/pvscan /usr/sbin/vgchange /usr/lib/udev/rules.d/69-dm-lvm.rules"
```
To tell dracut to include key LVM binaries, and an udev rule that takes care of triggering these commands. I have tried booting into the initramfs, and the lv for `/var` from @pirat89's example is activated, although the `rd.lvm.lv=` options are used for `/` and `/usr/


2) We want to ditch running `/mount -a` in a loop and, instead, use systemd mechanism
We will be hijacking `systemd-fstab-generator`, keeping the units, but modifying their mount targets to be prefixed with `/sysroot`. I have to still try this one out...

prajnoha:
@MichalHe, thanks for the great in-depth analysis. Just to understand the issue better - so the problem is if we're trying to access anything else than `/` and `/usr` (like the `/var`) with the leapp upgrade script inside initrd, right?

Is this just a new use case we need to support now (having `var` on a separate device)? Otherwise, how did this work before? Was leapp using only `/` and `/usr` before (without  `/var` being a separate mount point)? I'm a bit confused now...

pirat89:
> @MichalHe, thanks for the great in-depth analysis. Just to understand the issue better - so the problem is if we're trying to access anything else than `/` and `/usr` (like the `/var`) with the leapp upgrade script inside initrd, right?
> 
> Is this just a new use case we need to support now (having `var` on a separate device)? Otherwise, how did this work before? Was leapp using only `/` and `/usr` before (without  `/var` being a separate mount point)? I'm a bit confused now...

Previously, we activated lvm and all other mountpoints have been mounted later by `mount -a` - everything in /etc/fstab needs to be mounted. it's hard to upgrade the system when all system partitions are not mounted to perform all RPM operation.

prajnoha:
OK, I see - so dracut/systemd tandem in initrd only cares about `/` and `usr` and nothing else, because they are the only things needed to get the rootfs in the end and switch to that. Well, the solution that Michal commented about is actually how LVM autoactivates volumes when we are switched already from initrd to rootfs - which would mean bypassing and replacing dracut's hooks completely as a matter of fact. So the question now is whether we want to follow this path or whether we should update dracut itself to also care about other mount points than `/` and `/usr`.

prajnoha:
Side note: the reason that we're not using the usual autoactivation in dracut is historical. At first, we didn't have any event-based autoactivation, so dracut needed to come up with its own way. Then we added native event-based autoactivation to LVM, but only with a helper `lvmetad` daemon and we didn't want to include a daemon inside the dracut to only support autoactivation of a single VG/few VGs (with the LVs where / and /usr sit) and the LVM dracut hook worked, so there was no reason to replace it anyway. Now (since 2020 or 2021 I think), we removed lvmetad and we can do autoactivation just with the access to `/run` where we store helper files to track incoming PVs so we can autoactivate the VG once all needed PVs are present.

MichalHe:
@prajnoha Thanks. Anything that gets the job done is fine by me. I would just note that in case this change is implemented in dracut, we would be introducing code into dracut's codebase (a maintenance cost) that would be used by no one else but us (upgrades). How much work do you think would it be to get this into dracut? Is there a reason why is it a bad idea to bypass dracut's mechanisms?

@pirat89 Do you see any issues in case that this would be implemented in dracut w.r.t. to how leapp releases its RHEL builds? It would not be that we require dracut, we would have to require dracut with some minimal version, otherwise we would have to do some weird stuff when we detect the dracut version and use leapp's old mechanism or something.

I am also worried a bit about what if we decide, e.g., that we will support SANs.  I know a little to nothing about having a SAN attached to the system, so I blindly wonder whether a LVM could be set up on top of a SAN. Would implementing the solution in dracut be flexible enough? Our use case requires us to mount everything in fstab - is there a chance that a LVM activation requires some nontrivial systemd dependencies? 

Re-reading this message, I get the feeling that I am a proponent of going the "LVM-autoactivation" route. Please, do not be mistaken, I just want to make an informed decision :grin:

repos/system_upgrade/common/actors/initramfs/enable_lvm_autoactivation/actor.py
```diff
@@ -0,0 +1,21 @@
...
+    autoactivation in the initramfs.
+    """
+
+    name = 'enable_lvm_autoactivation'
+    consumes = ()
+    produces = (UpgradeInitramfsTasks, )
+    tags = (FactsPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        enable_lvm_autoactivation_lib.emit_lvm_autoactivation_instructions()
```
pirat89, line 21: Thinking about this actor, it's executed unconditionally. Do we want to make this happen on systems without LVM and in case the livemode is used? not a blocker, just making a note to think about it

repos/system_upgrade/common/actors/initramfs/mount_units_generator/libraries/mount_unit_generator.py
```diff
@@ -0,0 +1,235 @@
...
+        try:
+            run(['ln', '-s', fastlink_points_to, place_fastlink_at])
+
+            api.current_logger().debug(
+                'Dependency on {} created.'.format(unit_file)
+            )
+        except CalledProcessError as err:
+            err = 'Failed to required unit dependencies for local-fs.target in upgrade initramfs.'
+            details = {'details': str(err)}
+            raise StopActorExecutionError(err, details=details)
```
pirat89, line 144: @MichalHe  This seems little odd. Did you want to store the string into a different variable. as right now the msg and details are the same thing.

repos/system_upgrade/common/actors/initramfs/mount_units_generator/libraries/mount_unit_generator.py
```diff
@@ -0,0 +1,307 @@
...
+    unit files that could be put there. But most likely "local-fs.target.requires"
+    will always exists.
+    """
+    dir_list = [
+        'local-fs.target.requires',
+        'local-fs.target.wants',
+        'local-fs-pre.target.requires',
+        'local-fs-pre.target.wants',
+        'remote-fs.target.requires'
+        'remote-fs.target.wants',
```
matejmatuska, line 197: Missing comma:
```suggestion
        'remote-fs.target.requires',
        'remote-fs.target.wants',
```

pirat89: Holy macaroni!

pirat89:
Retesting summary (in-progress) (swap is always included in following tests)
* AWS 8 -> 9):
* AWS 9 -> 10):
* Azure (8->9): 
* Azure (9->10):
* GCP: :no_entry_sign: 

* tests with encrypted storage (8->9): :heavy_check_mark: 
* tests with encrypted storage (9->10):

* VMs with LVM - simple (root, boot) (8->9):  :heavy_check_mark: 
* VMs with LVM  - simple (root, boot) (9->10): 

* VMs with LVM (root, boot, usr, var) (8->9):  :heavy_check_mark: 
* VMs with LVM (root, boot, usr, var) (9->10): 

* VMs without LVM (8 ->9): 
* VMs without LVM (9 ->10): 

* VM (8->9->10): rootfs + /boot: 
* single MP (rootfs only, /boot is part of rootfs): :no_entry_sign: 


---
** :heavy_check_mark: works
** :x:  broken
** :no_entry_sign:  skipped
** nothing - not tested yet

repos/system_upgrade/common/actors/initramfs/mount_units_generator/libraries/mount_unit_generator.py
```diff
@@ -0,0 +1,307 @@
...
+    """
+    dir_list = [
+        'local-fs.target.requires',
+        'local-fs.target.wants',
+        'local-fs-pre.target.requires',
+        'local-fs-pre.target.wants',
+        'remote-fs.target.requires',
+        'remote-fs.target.wants',
+        'remote-fs-pre.target.requires'
+        'remote-fs-pre.target.wants',
```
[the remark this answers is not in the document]
pirat89: here you see it...one simple function that I decided to not cover by unit-tests, source of 3 issues

repos/system_upgrade/common/actors/initramfs/mount_units_generator/tests/test_mount_unit_generation.py
```diff
@@ -0,0 +1,269 @@
...
+
+@pytest.mark.parametrize('dirname', (
+    'local-fs.target.requires',
+    'local-fs.target.wants',
+    'local-fs-pre.target.requires',
+    'local-fs-pre.target.wants',
+    'remote-fs.target.requires'
+    'remote-fs.target.wants',
+    'remote-fs-pre.target.requires'
+    'remote-fs-pre.target.wants',
```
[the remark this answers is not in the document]
pirat89: omg....these commas are killing me :dizzy_face:  sorry guys.

PR #1207, opened 2024-04-23
Enhance grub2 install failure message.

abadger:
The new message informs the useir will happen (they will boot into the old RHEL's kernel) so they understand why they need to run the remediation.

jira: https://issues.redhat.com/browse/RHEL-29683


pirat89 (approved):
lgtm. thanks!!

PR #1210, opened 2024-04-25
mention `Report` in produces of transitionsystemdservicesstates

evgeni:
fixes upgrade warnings:

     leapp.workflow.Applications.transition_systemd_services_states: Actor is trying to produce a message of type "<class 'leapp.reporting.Report'>" without mentioning it explicitely in the actor's "produces" tuple. The message will be ignored


matejmatuska (approved):
I'll wait for the tests to finish and then merge.

PR #1212, opened 2024-04-25
OpenSSHConfigScanner: Include directive is supported since RHEL 8.6

Jakuje:
This issue could cause false positive reports when the user has the configuration options such as "Subsystem sftp" defined in included file only.

Resolves: RHEL-33902

### DIFF
--- a/repos/system_upgrade/common/actors/opensshconfigscanner/libraries/readopensshconfig.py
+++ b/repos/system_upgrade/common/actors/opensshconfigscanner/libraries/readopensshconfig.py
@@ -1,5 +1,9 @@
 import errno
+import glob
+import os
+import shlex
 
+from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.common.rpms import check_file_modification
 from leapp.libraries.stdlib import api
 from leapp.models import OpenSshConfig, OpenSshPermitRootLogin
@@ -12,14 +16,31 @@ def line_empty(line):
     return len(line) == 0 or line.startswith('\n') or line.startswith('#')
 
 
-def parse_config(config):
-    """Parse OpenSSH server configuration or the output of sshd test option."""
+def parse_config(config, base_config=None, current_cfg_depth=0):
+    """
+    Parse OpenSSH server configuration or the output of sshd test option.
 
-    # RHEL7 defaults
-    ret = OpenSshConfig(
-        permit_root_login=[],
-        deprecated_directives=[]
-    )
+    :param Optional[OpenSshConfig] base_config: Base configuration that is extended with configuration options from
+    current file.
+
+    :param int current_cfg_depth: Internal counter for how many includes were already followed.
+    """
+
+    if current_cfg_depth > 16:
+        # This should really never happen as it would mean the SSH server won't
+        # start anyway on the old system.
+        error = 'Too many recursive includes while parsing sshd_config'
+        api.current_logger().error(error)
+        return None
+
+    ret = base_config
+    if ret is None:
+        # RHEL7 defaults
+        ret = OpenSshConfig(
+            permit_root_login=[],
+            deprecated_directives=[]
+        )
+        # TODO(Jakuje): Do we need different defaults for RHEL8?
 
     in_match = None
     for line in config:
@@ -68,8 +89,26 @@ def parse_config(config):
                 # here we need to record all remaining items as command and arguments
                 ret.subsystem_sftp = ' '.join(el[2:])
 
+        elif el[0].lower() == 'include':
+            # recursively parse the given file or files referenced by this option
+            # the option can have several space-separated filenames with glob wildcards
+            for pattern in shlex.split(' '.join(el[1:])):
+                if pattern[0] != '/' and pattern[0] != '~':
+                    pattern = os.path.join('/etc/ssh/', pattern)
+                files_matching_include_pattern = glob.glob(pattern)
+                # OpenSSH sorts the files lexicographically
+                files_matching_include_pattern.sort()
+                for included_config_file in files_matching_include_pattern:
+                    output = read_sshd_config(included_config_file)
+                    if parse_config(output, base_config=ret, current_cfg_depth=current_cfg_depth + 1) is None:
+                        raise StopActorExecutionError(
+                            'Failed to parse sshd configuration file',
+                            details={'details': 'Too many recursive includes while parsing {}.'
+                                     .format(included_config_file)}
+                        )
+
         elif el[0].lower() in DEPRECATED_DIRECTIVES:
-            # Filter out duplicit occurrences of the same deprecated directive
+            # Filter out duplicate occurrences of the same deprecated directive
             if el[0].lower() not in ret.deprecated_directives:
                 # Use the directive in the form as found in config for user convenience
                 ret.deprecated_directives.append(el[0])
@@ -82,10 +121,10 @@ def produce_config(producer, config):
     producer(config)
 
 
-def read_sshd_config():
+def read_sshd_config(config):
     """Read the actual sshd configuration file."""
     try:
-        with open(CONFIG, 'r') as fd:
+        with open(config, 'r') as fd:
             return fd.readlines()
     except IOError as err:
         if err.errno != errno.ENOENT:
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/common/actors/opensshconfigscanner/libraries/readopensshconfig.py
```diff
@@ -12,14 +15,23 @@ def line_empty(line):
     return len(line) == 0 or line.startswith('\n') or line.startswith('#')
 
 
-def parse_config(config):
+def parse_config(config, ret=None, depth=0):
     """Parse OpenSSH server configuration or the output of sshd test option."""
```
MichalHe, line 19: This requires a better argument names and their docstrings.

```suggestion
def parse_config(config, default_return_val=None, current_cfg_depth=0):
    """
    Parse OpenSSH server configuration or the output of sshd test option.
    
    :param Optional[OpenSshConfig] default_return_val: Default return value returned in case <?>
    :param int current_cfg_depth: How many recursive config includes were already followed. 

    """
```

repos/system_upgrade/common/actors/opensshconfigscanner/libraries/readopensshconfig.py
```diff
@@ -68,6 +80,22 @@ def parse_config(config):
...
 
+        elif el[0].lower() == 'include':
+            # recursively parse the given file or files referenced by this option
+            pattern = el[1]
+            if pattern[0] != '/' and pattern[0] != '~':
+                pattern = os.path.join('/etc/ssh/', pattern)
+            # NOTE that OpenSSH sorts the files lexicographically
+            files = glob.glob(pattern)
+            files.sort()
+            for f in files:
```
MichalHe, line 91:
```suggestion
            for included_config_file in files_matching_include_pattern:
```

repos/system_upgrade/common/actors/opensshconfigscanner/libraries/readopensshconfig.py
```diff
@@ -68,6 +80,22 @@ def parse_config(config):
...
+                output = read_sshd_config(f)
+                if parse_config(output, ret, depth + 1) is None:
```
MichalHe, line 93: If a function defines a key-value argument, then it can cause future problems not calling it with as a key-value argument. For example, you can add a normaln (non-kwarg) argument and suddenly you break stuff you have believed should have not been broken because the remaining arguments were declared as key-value.
```suggestion
                if parse_config(output, red=ret, depth=depth + 1) is None:
```

repos/system_upgrade/common/actors/opensshconfigscanner/tests/test_readopensshconfig_opensshconfigscanner.py
```diff
@@ -143,12 +150,66 @@ def test_parse_config_deprecated():
 def test_parse_config_empty():
     output = parse_config([])
     assert isinstance(output, OpenSshConfig)
-    assert isinstance(output, OpenSshConfig)
     assert not output.permit_root_login
     assert output.use_privilege_separation is None
     assert output.protocol is None
 
 
+def test_parse_config_include():
```
MichalHe, line 158: Instead of using temporary files, you abstract away lines where you are reading from disk in the scanner, e.g.,
```python3
       with open(config, 'r') as fd:
            return fd.readlines()
```
would become a new function within th e actor. Then you can mock this function within your tests, avoiding interacting with filesystem altogether.

MichalHe (changes requested):
From the high-level perspective, these changes are fine. I left minor comments regarding variable naming. If you need help with avoiding touching the filesystem in the unit tests, I can happily render my assistance.

repos/system_upgrade/common/actors/opensshconfigscanner/libraries/readopensshconfig.py
Jakuje: Thank you for the suggestions! I would be for improving the naming of `ret` to `base_config` as this matches more what it is used for (extending previously created configuration structure).

Regarding to the `depth`, the description sounds good, but it should be only internal. Is there a good way to express this in the python/docstring? Start wit underscore?

repos/system_upgrade/common/actors/opensshconfigscanner/tests/test_readopensshconfig_opensshconfigscanner.py
Jakuje: Given that the actor goes over several files that are included from each other, I found it more useful to have the real directory structure instead of mocking the reading of different files. I would need to mock also the globing, which felt I would have to replace most of the logic in there.

MichalHe: @Jakuje Out of curiosity I have tried to rewrite the test without using filesystem https://github.com/MichalHe/leapp-repository/commit/a4d496f8c89d1a04e63f8042cdb9d9ed2bd8d2eb. What do you think about the approach? If you are adamant about using temporary files, then we can close this conversation and let this patch be merged with temporary files. I do not like relying on temporary files if it can be avoided in a sane fashion because the less we rely on anything the less things can go wrong. Maybe I am just paranoid because upgrades thought me not to trust any aspect of the system I am running on :grin: 

If you consider the approach proposed by me OK, then feel free to just cherry-pick the commit from my branch.

MichalHe:
@Jakuje Could the glob expression in the `Include` directive contain spaces? For example

```
Include "/home/my configs/sshd_config"
```
In such a case the logic would be faulty as it would try to use the pattern `/home/my`

repos/system_upgrade/common/actors/opensshconfigscanner/tests/test_readopensshconfig_opensshconfigscanner.py
Jakuje: sure. I will use your proposal. It looks cleaner. I got almost that far to implement it when I realized the glob needs to be mocked too and gave up.

Jakuje:
> @Jakuje Could the glob expression in the `Include` directive contain spaces? For example
> 
> ```
> Include "/home/my configs/sshd_config"
> ```
> 
> In such a case the logic would be faulty as it would try to use the pattern `/home/my`
> 
> A solution in such a case would be to use (this ensures that only the leftmost whitespace will be used for splitting)
> 
> ```python
> el = el.split(None, 1)
> ```

Good catch! You are right.  I will change it.

From the manual page of sshd_config, it can also accept several filenames one one line so that would be also good to test if this needs some special handling.

Jakuje:
> el = el.split(None, 1)

reading through the code, this wont work as in some other cases we already depend on the rest being split by spaces (such as match, subsystem). The way how the splitting is implemented in OpenSSH

https://github.com/openssh/openssh-portable/blob/master/misc.c#L2044

is something like the shell expansions so using `shlex` should do.

Pushed the updated change.

MichalHe:
I have tested the patch using the setup described in `RHEL-33902` and it works like a charm. Just waiting for the tests to finish and this can be merged.

repos/system_upgrade/common/actors/opensshconfigscanner/libraries/readopensshconfig.py
```diff
@@ -68,8 +89,26 @@ def parse_config(config):
...
+                if pattern[0] != '/' and pattern[0] != '~':
+                    pattern = os.path.join('/etc/ssh/', pattern)
+                files_matching_include_pattern = glob.glob(pattern)
+                # OpenSSH sorts the files lexicographically
+                files_matching_include_pattern.sort()
+                for included_config_file in files_matching_include_pattern:
+                    output = read_sshd_config(included_config_file)
+                    if parse_config(output, base_config=ret, current_cfg_depth=current_cfg_depth + 1) is None:
+                        raise StopActorExecutionError(
+                            'Failed to parse sshd configuration file: ',
```
MichalHe, line 105: Just a tiny detail: the message here seems to be intended to also include the filename of the problematic config (guessing by the trailing `: `). Maybe a `.format` is missing?

Jakuje: Yeah. Its now in the details so I am not sure if it would be useful to duplicate it also to the error title. What would be the best to be consistent with the rest of the messages?
I think I remember I heard keeping the title static and expanding the filenames in `details` is preferred. If so, please confirm and I will just remove the trailing `: ` from the title.

Jakuje: Removed trailing colon.

Rezney (approved):

PR #1217, opened 2024-05-03
Add additional KB resources

Rezney:
add aditional KB resources in a form of ExternalLink or error details as requested by support

### DIFF
--- a/repos/system_upgrade/common/actors/checkdetecteddevicesanddrivers/libraries/checkdddd.py
+++ b/repos/system_upgrade/common/actors/checkdetecteddevicesanddrivers/libraries/checkdddd.py
@@ -35,6 +35,16 @@ def create_inhibitors(inhibiting_entries):
                     source=get_source_major_version(),
                 )
             ),
+            reporting.ExternalLink(
+                url='https://access.redhat.com/solutions/6971716',
+                title='Leapp preupgrade getting "Inhibitor: Detected loaded kernel drivers which have been '
+                      'removed in RHEL 8. Upgrade cannot proceed." '
+            ),
+            reporting.ExternalLink(
+                url='https://access.redhat.com/solutions/5436131',
+                title='Leapp upgrade fail with error "Inhibitor: Detected loaded kernel drivers which '
+                      'have been removed in RHEL 8. Upgrade cannot proceed."'
+            ),
             reporting.Audience('sysadmin'),
             reporting.Groups([reporting.Groups.KERNEL, reporting.Groups.DRIVERS]),
             reporting.Severity(reporting.Severity.HIGH),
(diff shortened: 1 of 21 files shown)


repos/system_upgrade/common/actors/checkdetecteddevicesanddrivers/libraries/checkdddd.py
```diff
@@ -35,6 +35,16 @@ def create_inhibitors(inhibiting_entries):
                     source=get_source_major_version(),
                 )
             ),
+            reporting.ExternalLink(
```
mkluson, line 38: I would keep just one link in the tooling and possibly merge content of both solution.

repos/system_upgrade/common/actors/persistentnetnamesdisable/actor.py
```diff
@@ -50,6 +50,11 @@ def process(self):
                     title='How to perform an in-place upgrade to RHEL 8 when using kernel NIC names on RHEL 7',
                     url='https://access.redhat.com/solutions/4067471'
                 ),
+                reporting.ExternalLink(
+                    title='RHEL 8 to RHEL 9: inplace upgrade fails at '
```
mkluson, line 54: [nitpicking] usually you put `url` first

repos/system_upgrade/common/libraries/rhsm.py
```diff
@@ -85,7 +85,8 @@ def _handle_rhsm_exceptions(hint=None):
             details={
                 'details': str(e),
                 'stderr': e.stderr,
-                'hint': hint or _def_hint
+                'hint': hint or _def_hint,
+                'link': 'https://access.redhat.com/solutions/6138372'
```
mkluson, line 89: no `title`?

repos/system_upgrade/el7toel8/actors/checkbtrfs/actor.py
```diff
@@ -41,6 +41,10 @@ def process(self):
                             title='How do I prevent a kernel module from loading automatically?',
                             url='https://access.redhat.com/solutions/41278'
                         ),
+                        reporting.ExternalLink(
+                            title='Leapp upgrade fail with error "Inhibitor: Btrfs has been removed from RHEL8"',
```
mkluson, line 45: `url` is first in majority of cases

mkluson (commented):
Do we really want to have all the titles hard-coded in the tooling? I can imagine the titles will change over time and it will lead to unnecessary confusion. I would be ok to have just plain links.

repos/system_upgrade/common/actors/persistentnetnamesdisable/actor.py
pirat89: it's ok, the order is not important in this case. let's keep it

repos/system_upgrade/common/libraries/rhsm.py
pirat89: it's a detail for an error msg, not report. so not title.

pirat89 (approved):
lgtm. merging

pirat89:
@andywaltlova I guess this one is not important for you, as only external links have been added and no other change in reports happened. but to make it simple, setting the report label too.

PR #1220, opened 2024-05-04
Update PR bot welcome msg

pirat89:
I found useful to be able to run just failed tests (when they crashed due to infrastructure issues). Updating the welcome msg to provide this info and little bit restructured original statement, to make it more friendly for reading. I've updated the bot msg below manually to show the expected content.

### DIFF
--- a/.github/workflows/pr-welcome-msg.yml
+++ b/.github/workflows/pr-welcome-msg.yml
@@ -21,16 +21,21 @@ jobs:
             ## **Thank you for contributing to the Leapp project!**
             Please note that every PR needs to comply with the [Leapp Guidelines](https://leapp.readthedocs.io/en/latest/contributing.html#) and must pass all tests in order to be mergeable.
             If you want to request a review or rebuild a package in copr, you can use following commands as a comment:
-            - **review please @oamg/developers** to notify leapp developers of the review request
-            - **/packit copr-build** to submit a public copr build using packit
+            - **`review please @oamg/developers`** to notify leapp developers of the review request
+            - **`/packit copr-build`** to submit a public copr build using packit
 
-            Packit will automatically schedule regression tests for this PR's build and latest upstream leapp build. If you need a different version of leapp, e.g. from PR#42, use `/packit test oamg/leapp#42`
-            Note that first time contributors cannot run tests automatically - they will be started by a reviewer.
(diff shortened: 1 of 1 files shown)


pirat89:
@oamg/developers  note that `/packit retest-failed` command is not buggy: https://github.com/packit/packit-service/issues/1886 - so I can think about dropping it for now from the msg. wdyt?

r0x0d:
@pirat89, why not using dependabot or renovate to keep the dependencies up-to-date? With renovate, you can have more configurable settings for how many PRs it will open in a give period, I believe.

pirat89:
@r0x0d well, we can take a look. I am  not so familiar with it, but original approach to use dependabot in the project was to automatically inform about newer python packages available (when installing by pip to run unit-tests) and basically we didn't want any info about newer python apps available. if we can use it to track just particular deps, I think it could be potentially useful in future.

r0x0d:
> @r0x0d well, we can take a look. I am  not so familiar with it, but original approach to use dependabot in the project was to automatically inform about newer python packages available (when installing by pip to run unit-tests) and basically we didn't want any info about newer python apps available. if we can use it to track just particular deps, I think it could be potentially useful in future.

I will look into setting up renovate for leapp-repository. I will message you in slack to check which type of deps you want to get info.

evgeni:
> @r0x0d well, we can take a look. I am not so familiar with it, but original approach to use dependabot in the project was to automatically inform about newer python packages available (when installing by pip to run unit-tests) and basically we didn't want any info about newer python apps available. if we can use it to track just particular deps, I think it could be potentially useful in future.

dependabot can be configured to look at certain "ecosystems"/"package managers" only, so bump actions but not python, etc.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#example-of-a-basic-setup-for-three-package-managers

r0x0d:
Created https://github.com/oamg/leapp-repository/pull/1221

matejmatuska:
@pirat89 I know #1221 is already merged, just one thing to keep in mind is that when a GH Action is updated, we have to change the allowed version in repository setting too. That's what broke the pr-welcome-message-bot and shellcheck.

abadger (approved):

PR #1232, opened 2024-05-14
Make linter happy - fixing discovered potential problems

pirat89:
Issues:
* E0606 -> possibly-used-before-assignement
* W0135 -> contextmanager-generator-missing-cleanup

Note that in case of checktargetrepos actor we are introducing `https://red.ht/upgrading-rhel9-to-rhel10-main-official-doc` for IPU 9 -> 10. However, this shortened URL must be defined yet later. The doc does not exist yet, so no rush. I've added the URL to our list of used shortened URLs, marked as broken, so it's visible.

In case of `W0135`, linters expects try-finally right around `yield`. Checked reported context managers, usually it's FP as it doesn't take into acount:
* try-finally used aroud larger block of the code, it requires really try-finally around the `yield` (see `overlaygen.py:create_source_overlay()`)
* it does not take into account nested contextmanagers, like:
  ```python
  @contextlib.contextmanager
  def mycontectmgr():
    with another_context_mgr(...) as foo:
      yield foo
  ```
  which in such a case the try-final is handled by the nested `another_context_mgr`, which is the typical scenario in reported warnings.


MichalHe (approved):
LGTM

PR #1233, opened 2024-05-16
drop unused packager field from distro metadata

evgeni:
The actor was refactored to accept any `gpg-key` package, regardless of
the value in the packager field, but the data was never drop from the
JSON files.

Fixes: 4968bec73947fb83aeb2d89fe7e919fba2ca2776


pirat89:
Well, I remember I was thinking about that in the past, but decided to keep to to postpone what we will want to do about that. Whether to use the extra info about the packager or not at all in future. But thinking about that twice, I think the check is not neccessary and the complete drop makes sense. @oamg/developers  any opinions?

pirat89 (approved):
lgtm. postponing merging to give the team time to think about that.

PR #1234, opened 2024-05-16
properly indent the list of supported OSes

evgeni:
before:
```
Risk Factor: high (inhibitor)
Title: The installed OS version is not supported for the in-place upgrade to the target RHEL version
Summary: The supported OS releases for the upgrade process:
 RHEL 8.8
RHEL 8.10
RHEL-SAPHANA 8.8
RHEL-SAPHANA 8.10
```

after:
```
Risk Factor: high (inhibitor)
Title: The installed OS version is not supported for the in-place upgrade to the target RHEL version
Summary: The supported OS releases for the upgrade process:
    - RHEL 8.8
    - RHEL 8.10
    - RHEL-SAPHANA 8.8
    - RHEL-SAPHANA 8.10
```

### DIFF
--- a/repos/system_upgrade/common/actors/checkosrelease/libraries/checkosrelease.py
+++ b/repos/system_upgrade/common/actors/checkosrelease/libraries/checkosrelease.py
@@ -4,6 +4,7 @@ from leapp import reporting
 from leapp.libraries.common.config import version
 
 COMMON_REPORT_TAGS = [reporting.Groups.SANITY]
+FMT_LIST_SEPARATOR = '\n    - '
 
 related = [reporting.RelatedResource('file', '/etc/os-release')]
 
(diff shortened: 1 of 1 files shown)


pirat89:
@evgeni  I think it would be even better to used `FMT_LIST_SEPARATOR = '\n    - '` defined in many other actors, to have the list formatted in the same way as anywhere else. I am ok to go with this intendantion fix now but it would be pity to not change it already. wdyt?

pirat89 (approved):

evgeni:
Like this? Or less indentation. Wasn't sure :)

pirat89 (approved):
@evgeni  exactly like this! thanks!

pirat89:
@evgeni nice PR number :-D

PR #1235, opened 2024-05-16
report which OS release was detected as unsupported

evgeni:
includes #1234 to prevent merge conflicts

### DIFF
--- a/repos/system_upgrade/common/actors/checkosrelease/libraries/checkosrelease.py
+++ b/repos/system_upgrade/common/actors/checkosrelease/libraries/checkosrelease.py
@@ -30,13 +30,16 @@ def check_os_version():
         for rel in version.SUPPORTED_VERSIONS:
             for ver in version.SUPPORTED_VERSIONS[rel]:
                 supported_releases.append(rel.upper() + ' ' + ver)
+        current_release = ' '.join(version.current_version()).upper()
         reporting.create_report([
             reporting.Title(
                 'The installed OS version is not supported for the in-place upgrade to the target RHEL version'
             ),
             reporting.Summary(
                 'The supported OS releases for the upgrade process:'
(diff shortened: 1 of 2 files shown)


pirat89:
@evgeni lgtm. I like it. But thinking about that, when you already started these changes, that we could add also remediation instructions, somthing like:
```python
pkg_manager = 'yum' if version.get_source_major_version == '7' else 'dnf
reporting.Remediation(hint=(
    'Update your system using {pkg_manager} to the version of the system supported for the in-place upgrade.'
    .format(pkg_manager=pkg_manager)
))
```

evgeni:
Personally, I wouldn't do that.
The unsupported systems will be Alma, Rocky etc, and they *can't* be upgraded using `dnf`, so posting that remediation would be wrong

pirat89:
Interesting. Oki, let's skip it for now.

pirat89 (approved):

pirat89:
well, when we add check for the `rhel` system, then we can clearly skip this whole report from my pov. e.g. for centos stream this report does not provide any value at all. and I have no idea about requirements from almalinux/rocky at all. so from my pov, it's matter of time we update this particular actor. but it's for me interesting that almalinux could not use yum/dnf for the update of the system. that's for me hoenstly little bit weird. so I guess I am missing here something.

evgeni:
Well, can you use yum/dnf to upgrade RHEL from 8 to 9? Alma/Rocky/Oracle has the same issue. You gotta Leapp major versions.

pirat89:
no, but you can use DNF do update to newer minor version. This is about trying to upgrade FROM unsupported version of the system.

evgeni:
Ooooh. Now I get what you wanted to do!
```
if (on 8.9) {
    tell user to update to 8.10 first using dnf
}

PR #1243, opened 2024-05-27
BZ#2283067 - don't remove Tomcat during Satellite 7to8 upgrade

evgeni:
This was previously added to workaround issues with the packages in RHEL
8.9, but now that 8.10 is released it's not required anymore and
actually breaks the upgrade.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/satellite_upgrade_facts/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/satellite_upgrade_facts/actor.py
@@ -1,7 +1,7 @@
 import os
 
 from leapp.actors import Actor
-from leapp.libraries.common.config import architecture
+from leapp.libraries.common.config import architecture, version
 from leapp.libraries.common.rpms import has_package
 from leapp.libraries.stdlib import run
 from leapp.models import (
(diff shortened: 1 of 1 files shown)


pirat89 (approved):
@evgeni  in these cases it's usually better to use a condition, so the action happens for particular upgrade path and it's ignored otherwise. The benefit is, that if someone want to try in future upgrade to 8.9, e.g. for testing purposes to see specific behaviour between the use of 8.9 & 8.10 repositories, they will not have to manually update in such a case ..../to_remove file to ensure the package is removed during the upgrade. But as this is satellite specific and upgrade to 8.9 is not supported anymore, I have no problem to trop it. But I will postpone the merge for a while in case you would like to add the condition instead of removal. Let me know please, thanks. for the condition, I would use e.g.:

```python
from leapp.libraries.common.config import version
if version.matches_target_version('8.9'):
```

If it was problem for 8.9 only. In case the same problem was for 8.8, you can specify:
```
if version.matches_target_version('8.8', '8.9'):
```

evgeni:
Yeah, the problem was 8.9-only as the tomcat/pki-servlet/candlepin packages were in flux. It's now all (hopefully) settled in 8.10.

I'll add that suggestion!

evgeni:
> Yeah, the problem was 8.9-only as the tomcat/pki-servlet/candlepin packages were in flux. It's now all (hopefully) settled in 8.10.

Glad I've actually tested the patch… 8.8 and 8.9 is affected. Patch updated :)

pirat89:
Checking failed tests, problems are caused by recent RHUI changes. Change in this PR is not related to these fails. Waiting for last 2 tests that are still running which should not be affected by RHUI.

PR #1245, opened 2024-05-28
Enable IPU path 8.10 -> 9.5

matejmatuska:
* Enable 9.5 target
* Add RHEL 9.5 product certificates


repos/system_upgrade/common/files/upgrade_paths.json
```diff
@@ -2,15 +2,15 @@
...
+        "8": ["9.2", "9.4", "9.5"]
     },
     "saphana": {
         "7.9": ["8.10", "8.8"],
         "7": ["8.10", "8.8"],
         "8.8": ["9.2"],
-        "8.10": ["9.4"],
-        "8": ["9.4", "9.2"]
+        "8.10": ["9.4", "9.5"],
+        "8": ["9.4", "9.2", "9.5"]
```
pirat89, line 14: @matejmatuska  sap hana is not supported for odd. it sould stay untouched.

pirat89 (changes requested):

pirat89 (approved):
lgtm

pirat89:
@matejmatuska  seems we cannot allow it yet. so let's drop the change in tests and put it into a separate PR.

matejmatuska:
> @matejmatuska seems we cannot allow it yet. so let's drop the change in tests and put it into a separate PR.

Ok, I dropped the commit.

pirat89 (approved):
lgtm

PR #1256, opened 2024-06-20
Fix IPU being blocked by resource limitations

tomasfratrik:
First resource limit is maximum number of open file descriptors limit, second one being limit for maximum writable file size.

With resource lib we set the limit for the current process (tested with my testing scripts) and its children processes. By testing pre-upgrade and upgrade I found a threshold of max. open file descriptors to be around 512, (the default value for ulimit -n is 1024) so a small multiplier of the default value shouldn't be a problem to set. With writing size there shouldn't be a problem to set it to "infinity".

Both problems from the tickets were solved, but there could be other side effects.

Resolves: RHEL-26459 and RHEL-16881

### DIFF
--- a/commands/command_utils.py
+++ b/commands/command_utils.py
@@ -1,6 +1,7 @@
 import json
 import os
 import re
+import resource
 
 from leapp.exceptions import CommandError
 from leapp.utils import path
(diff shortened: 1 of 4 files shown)


matejmatuska:
> I don't really like the redundancy of code here, but I'm not sure where should the common logic be placed.

I think `commands/command_utils.py` is an option. There is already some code the commands share.

tomasfratrik:
> > I don't really like the redundancy of code here, but I'm not sure where should the common logic be placed.
> 
> I think `commands/command_utils.py` is an option. There is already some code the commands share.

Thanks, I put it there

pirat89:
all standard tests passed. executing extended tests

pirat89:
ah.. I thought it's different PR :D mama mia!

matejmatuska (changes requested):
Tested for both maximum file size and maximum open fd's, works as expected.

However unit tests are missing.

tomasfratrik:
Latest push squashes commits.

tomasfratrik:
Unit tests were added.

matejmatuska (approved):
LGTM

NOTE: If this breaks local tests in containers for you, just rebuild them (make `clean_containers` and then it will be rebuilt the next time you run `make test_container`)

PR #1258, opened 2024-06-28
Fix storage scanner parsing error

tomasfratrik:
The given error arises when storage scanner actor attempts to parse output from command `pvs` while output containts separator symbol within itself. Changed the format of physical volume as was suggested in the ticket so it contains separator `:`, and was given following error while preupgrade and upgrade.
(e.g: format of physical volume, contains the symbol `:`  `/dev/disk/by-path/pci-0000:00:03.0-part2)

```
  File "/etc/leapp/repos.d/system_upgrade/common/actors/storagescanner/libraries/storagescanner.py", line 195, in _get_pvs_info
    pv, vg, fmt, attr, psize, pfree = entry
ValueError: too many values to unpack (expected 6)
```

Therefore changing the separation symbol to `|`, and running preupgrade and upgrade from RHEL8 to RHEL9 didn't raise any errors.

Jira: RHEL-34570

### DIFF
--- a/repos/system_upgrade/common/actors/storagescanner/libraries/storagescanner.py
+++ b/repos/system_upgrade/common/actors/storagescanner/libraries/storagescanner.py
@@ -206,7 +206,7 @@ def _get_lsblk_info():
 @aslist
 def _get_pvs_info():
     """ Collect storage info from pvs command """
-    for entry in _get_cmd_output(['pvs', '--noheadings', '--separator', r':'], ':', 6):
+    for entry in _get_cmd_output(['pvs', '--noheadings', '--separator', r'|'], '|', 6):
         pv, vg, fmt, attr, psize, pfree = entry
         yield PvsEntry(
             pv=pv,
@@ -220,7 +220,7 @@ def _get_pvs_info():
 @aslist
 def _get_vgs_info():
     """ Collect storage info from vgs command """
-    for entry in _get_cmd_output(['vgs', '--noheadings', '--separator', r':'], ':', 7):
+    for entry in _get_cmd_output(['vgs', '--noheadings', '--separator', r'|'], '|', 7):
         vg, pv, lv, sn, attr, vsize, vfree = entry
         yield VgsEntry(
             vg=vg,
@@ -235,7 +235,7 @@ def _get_vgs_info():
 @aslist
 def _get_lvdisplay_info():
     """ Collect storage info from lvdisplay command """
-    for entry in _get_cmd_output(['lvdisplay', '-C', '--noheadings', '--separator', r':'], ':', 12):
+    for entry in _get_cmd_output(['lvdisplay', '-C', '--noheadings', '--separator', r'|'], '|', 12):
         lv, vg, attr, lsize, pool, origin, data, meta, move, log, cpy_sync, convert = entry
         yield LvdisplayEntry(
             lv=lv,


pirat89:
@tomasfratrik rebase against master. tests should be fixed now

repos/system_upgrade/common/actors/storagescanner/libraries/storagescanner.py
```diff
@@ -191,7 +191,7 @@ def _get_lsblk_info():
 @aslist
 def _get_pvs_info():
     """ Collect storage info from pvs command """
-    for entry in _get_cmd_output(['pvs', '--noheadings', '--separator', r':'], ':', 6):
+    for entry in _get_cmd_output(['pvs', '--noheadings', '--separator', r'|'], ':', 6):
```
matejmatuska, line 194:
```suggestion
    for entry in _get_cmd_output(['pvs', '--noheadings', '--separator', r'|'], '|', 6):
```
I think you forgot to modify the separator as well, same in the other 2 calls.

matejmatuska (changes requested):

repos/system_upgrade/common/actors/storagescanner/libraries/storagescanner.py
matejmatuska: The pv is incorrectly parsed without that:
```
"pvs": [
        {
            "attr": "",
            "fmt": "03.0-part2|rhel_ibm-p8-kvm-03-guest-02|lvm2|a--|<39.00g|0",
            "pfree": "",
            "psize": "",
            "pv": "/dev/disk/by-path/pci-0000",
            "vg": "00"
        }
    ],

```

tomasfratrik: Completly missed that, thanks!
Run it without to get the same message u got:
```
"pvs": [
        {
            "attr": "",
            "fmt": "03.0-part2|rhel_ibm-p8-kvm-03-guest-02|lvm2|a--|<39.00g|0",
            "pfree": "",
            "psize": "",
            "pv": "/dev/disk/by-path/pci-0000",
            "vg": "00"
        }
    ],
```
After fix the result should be now correct:
```
     "pvs": [
        {
            "attr": "a--",
            "fmt": "lvm2",
            "pfree": "0",
            "psize": "<39.00g",
            "pv": "/dev/disk/by-path/pci-0000:00:03.0-part2",
            "vg": "rhel_ibm-p8-kvm-03-guest-02"
        }
    ],
```
```
$ pvs -a
  PV                                       VG                          Fmt  Attr PSize   PFree
  /dev/disk/by-path/pci-0000:00:03.0-part1                                  ---       0     0
  /dev/disk/by-path/pci-0000:00:03.0-part2 rhel_ibm-p8-kvm-03-guest-02 lvm2 a--  <39.00g    0
```

matejmatuska:
@tomasfratrik Works as expected now, please squash the commits and I can approve this.

tomasfratrik:
The latest push squashes commits.

MichalHe:
@tomasfratrik Hmm, if you step back and read the commit message with fresh eyes, it is not clear what does "this issue" refers to:
> Output separators were before set to ':', but the output values (e.g.
> from command 'pvs -a') could also contain this symbol,
> therefore changing separator symbol fixes this issue.

Could you please rephrase the commit message describes the problem without ambiguous references? 

Below, you can find my suggestion, but I feel that it is a bit too detailed, maybe there is a shorter one.
> The storagescanner actor uses `:` as a separator when parsing the output of, e.g., `pvs -a`. However, this is problematic as, e.g., PCI paths in /dev/disk/by-path might contain `:` in their names. Since we split outputs of executed commands using this separator, having more `:` in the output than expected causes the actor to crash. This commit, therefore, changes the separator to `|`.

tomasfratrik:
Yes you are right about the commit message. From 'far

PR #1259, opened 2024-07-01
codespell: Use specific version instead of master

matejmatuska:
Let's not blindly use the master ref, use a specific version instead.

This should fix the failing Codespell action. It's failing because only v2 is allowed to run in this repo (as set in repository settings).


fernflower (approved):
lgtm

pirat89:
merging. this change does not have an effect on the failed tests.

PR #1260, opened 2024-07-03
Fix detection when a bootable partition in on RAID

mscherer:
On my server, leapp preupgrade fail with the following error:

	    self.run()
	  File "/usr/lib/python2.7/site-packages/leapp/repository/actor_definition.py", line 74, in _do_run
	  File "/usr/lib/python2.7/site-packages/leapp/actors/__init__.py", line 289, in run
	    self.process(*args)
	  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/scangrubdevpartitionlayout/actor.py", line 18, in process
	    scan_layout_lib.scan_grub_device_partition_layout()
	  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/scangrubdevpartitionlayout/libraries/scan_layout.py", line 91, in scan_grub_device_partition_layout
	    dev_info = get_partition_layout(device)
	  File "/usr/share/leapp-repository/repositories/system_upgrade/el7toel8/actors/scangrubdevpartitionlayout/libraries/scan_layout.py", line 79, in get_partition_layout
	    part_start = int(part_info[2]) if len(part_info) == len(part_all_attrs) else int(part_info[1])
	ValueError: invalid literal for int() with base 10: '*'

This is caused by the following line:

	/dev/sda1   *        2048     1026047      512000   fd  Linux raid autodetect

I have my server on EL7 with / using a Linux RAID so `len(part_info) != len(part_all_attrs)`, hence why it try to convert '*' to int.

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/scangrubdevpartitionlayout/libraries/scan_layout.py
+++ b/repos/system_upgrade/el7toel8/actors/scangrubdevpartitionlayout/libraries/scan_layout.py
@@ -64,7 +64,6 @@ def get_partition_layout(device):
         if not line.startswith('Device'):
             continue
 
-        part_all_attrs = split_on_space_segments(line)
         break
 
     partitions = []
@@ -75,7 +74,7 @@ def get_partition_layout(device):
 
         # If the partition is not bootable, the Boot column might be empty
         part_device = part_info[0]
-        part_start = int(part_info[2]) if len(part_info) == len(part_all_attrs) else int(part_info[1])
+        part_start = int(part_info[2]) if part_info[1] == '*' else int(part_info[1])
         partitions.append(PartitionInfo(part_device=part_device, start_offset=part_start*unit))
 
     return GRUBDevicePartitionLayout(device=device, partitions=partitions)
--- a/repos/system_upgrade/el7toel8/actors/scangrubdevpartitionlayout/tests/test_scan_partition_layout.py
+++ b/repos/system_upgrade/el7toel8/actors/scangrubdevpartitionlayout/tests/test_scan_partition_layout.py
@@ -31,7 +31,8 @@ Partition = namedtuple('Partition', ['name', 'start_offset'])
         )
     ]
 )
-def test_get_partition_layout(monkeypatch, devices):
+@pytest.mark.parametrize('fs', ('Linux', 'Linux raid autodetect'))
+def test_get_partition_layout(monkeypatch, devices, fs):
     device_to_fdisk_output = {}
     for device in devices:
         fdisk_output = [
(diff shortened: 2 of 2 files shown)


fernflower:
Could  you please add a unit test?

repos/system_upgrade/el7toel8/actors/scangrubdevpartitionlayout/tests/test_scan_partition_layout.py
```diff
@@ -32,6 +32,10 @@
     ]
 )
 def test_get_partition_layout(monkeypatch, devices):
+    for f in ['Linux', 'Linux raid autodetect']:
+        inner_test_get_partition_layout(monkeypatch, devices, f)
+
+def inner_test_get_partition_layout(monkeypatch, devices, fs):
```
pirat89, line 38: more elegant would be pytest parametrize:

```
@pytest.mark.parametrize('fs', ('Linux', 'Linux raid autodetect'))
def test_get_partition_layout(monkeypatch, devices, fs):
```

pirat89: @mscherer  thanks! running initial tests.

pirat89:
@mscherer  unit test is now reproducing original issue :) btw, in case you would like to test locally, you can try (in the main dir) `TEST_CONTAINER=rhel7 make test_container` or you can use `test_container_no_lint` skip linters. so you do not need to wait for us to retrigger unit-tests execution every time. see `make help` for more if interested.

fernflower:
Thanks for the unit test!

pirat89:
Hi @mscherer , actually it's now problem on our side. Most likely we will ping you during the next week, asking for rebase when tests are fixed, so we can see that there is not introduced another regression. the code looks good to me, just want to be sure that it will be final solution and we will not need to investigate similar issue again soon. And we are also little bit busy last week. 

Downstream ticket with basically same issue has been reported today:
* https://issues.redhat.com/browse/RHEL-47038

repos/system_upgrade/el7toel8/actors/scangrubdevpartitionlayout/libraries/scan_layout.py
```diff
@@ -64,7 +64,6 @@ def get_partition_layout(device):
         if not line.startswith('Device'):
             continue
 
-        part_all_attrs = split_on_space_segments(line)
         break
```
pirat89, line 67: @mscherer  actually, checking the code, it seems this whole loop can be removed, as the only purpose is to get `part_all_attrs` variable.

pirat89:
Hi @mscherer, tests in master branch should be fixed now. Can you please rebase your branch?

pirat89 (approved):
lgtm and all tests passed. Thank you for your contribution!

PR #1262, opened 2024-07-04
GPG check: do not raise an error when TargetUserSpaceInfo is missing

pirat89:
Previously, if the upgrade has been inhibited during
    TargetTransactionFactsCollectionPhase
usually because we could not create (for whatever reason) the target userspace container, the actor checking rpm gpg keys failed with the `Could not check for valid GPG keys` error. This has confused many users as they couldn't know that this is impacted by the problem described in an inhibitor that is below this error.

As it's for sure that the upgrade cannot continue when the target user space container has not been created (the TargetUserSpaceInfo msg is missing), we consider it safe to stop the gpg check here silently just with a warning msg instead of raising the error - as this check is important only in case we could actually upgrade.

All other possible raised errors are presereved.

jira: https://issues.redhat.com/browse/RHEL-30573

### DIFF
--- a/repos/system_upgrade/common/actors/missinggpgkeysinhibitor/libraries/missinggpgkey.py
+++ b/repos/system_upgrade/common/actors/missinggpgkeysinhibitor/libraries/missinggpgkey.py
@@ -7,7 +7,7 @@ import tempfile
 from six.moves import urllib
 
 from leapp import reporting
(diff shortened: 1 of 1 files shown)


Rezney:
Closing in favor of https://github.com/oamg/leapp-repository/pull/1269 as we could not push to the branch.

PR #1266, opened 2024-07-17
Fix ModelViolationError while parsing repo files

tomasfratrik:
This error occurs when repo file has invalid definition, specifically the 'name' entry of the config file.

After this, we get rid of the traceback error, and the following errors are shown
```
2024-06-17 13:35:00.974356 [ERROR] Actor: system_facts
Message: Invalid repository definition: invalid-repo in: /etc/yum.repos.d/invalid.repo: The value of "name" field is None, but this is not allowed
Summary:
    Hint: For more directions on how to resolve the issue, see: https://access.redhat.com/solutions/6969001.
2024-06-17 13:35:01.545642 [ERROR] Actor: repositories_blacklist
Message: Actor didn't receive required messages: RepositoriesFact
```

Jira: RHEL-19249

### DIFF
--- a/repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
+++ b/repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
@@ -217,7 +217,18 @@ def get_sysctls_status():
 
 def get_repositories_status():
     """ Get a basic information about YUM repositories installed in the system """
-    return RepositoriesFacts(repositories=repofileutils.get_parsed_repofiles())
+    try:
+        return RepositoriesFacts(repositories=repofileutils.get_parsed_repofiles())
+    except repofileutils.InvalidRepoDefinition as e:
+        raise StopActorExecutionError(
+            message=str(e),
+            details={
+                'hint': 'For more directions on how to resolve the issue, see: {url}.'
+                        .format(
+                            url='https://access.redhat.com/solutions/6969001'
+                        )
+            }
+        )
 
 
 def get_selinux_status():
--- a/repos/system_upgrade/common/libraries/repofileutils.py
+++ b/repos/system_upgrade/common/libraries/repofileutils.py
@@ -11,6 +11,16 @@ except ImportError:
     api.current_logger().warning('repofileutils.py: failed to import dnf')
 
 
+class InvalidRepoDefinition(Exception):
+    """Raised when a repository definition is invalid."""
+    def __init__(self, msg, repofile, repoid):
+        message = 'Invalid repository definition: {repoid} in: {repofile}: {msg}'.format(
+                    repoid=repoid, repofile=repofile, msg=msg)
+        super(InvalidRepoDefinition, self).__init__(message)
+        self.repofile = repofile
+        self.repoid = repoid
+
+
 def _parse_repository(repoid, repo_data):
     def asbool(x):
         return x == '1'
@@ -33,12 +43,17 @@ def parse_repofile(repofile):
     :param repofile: Path to the repo file
     :type repofile: str
     :rtype: RepositoryFile
+    :raises InvalidRepoDefinition: If the repository definition is invalid,
+        this can for example occur if 'name' field in repository is missing or it is invalid.
     """
     data = []
     with open(repofile, mode='r') as fp:
         cp = utils.parse_config(fp, strict=False)
         for repoid in cp.sections():
-            data.append(_parse_repository(repoid, dict(cp.items(repoid))))
+            try:
+                data.append(_parse_repository(repoid, dict(cp.items(repoid))))
+            except fields.ModelViolationError as e:
+                raise InvalidRepoDefinition(e, repofile=repofile, repoid=repoid)
     return RepositoryFile(file=repofile, data=data)
 
 
--- a/repos/system_upgrade/common/libraries/tests/test_repofileutils.py
+++ b/repos/system_upgrade/common/libraries/tests/test_repofileutils.py
@@ -1,7 +1,10 @@
 import json
 import os
 
+import pytest
+
 from leapp.libraries.common import repofileutils
+from leapp.models.fields import ModelViolationError
 
 CUR_DIR = os.path.dirname(os.path.abspath(__file__))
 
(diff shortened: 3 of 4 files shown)


repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -217,7 +217,21 @@ def get_sysctls_status():
...
+            message='Invalid repository definition',
+            details={
+                'details': str(e),
+                'file': e.repofile,
+                'repoid': e.repoid,
+                'hint': 'To resolve this issue, please visit: {url}'
+                        .format(
+                            url='https://access.redhat.com/solutions/3185891'
+                        )
+            }
```
matejmatuska, line 233: There is a problem with this approach, if the `InvalidRepoDefinition` exception is raised and not caught and handled the way it's handled here, there will be no message (and possibly no other info besides traceback, not sure how the kwargs would be displayed).

You could move the message definition into the InvalidRepoDefinition constructor, or at least to the ctor call site. Also the message could contain the repoid and repofile directly in the string too, e.g. `Invalid repository definition: <repoid> in: <filepath>`. But it's good having them available as members still.

Personally I like the way it's done here: https://github.com/oamg/leapp-repository/blob/236483d5478ce845726ee05b7ca3e080105cdb80/repos/system_upgrade/el7toel8/libraries/isccfg.py#L11-L23
but keeping that up to you.

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -217,7 +217,21 @@ def get_sysctls_status():
...
+        raise StopActorExecutionError(
+            message='Invalid repository definition',
+            details={
+                'details': str(e),
+                'file': e.repofile,
+                'repoid': e.repoid,
+                'hint': 'To resolve this issue, please visit: {url}'
+                        .format(
+                            url='https://access.redhat.com/solutions/3185891'
+                        )
```
matejmatuska, line 232: The link in the hint seems to be incorrect (maybe a typo)?

Also the wording of the hint might be misleading for the users, visiting the link will to resolve the issue :), something like
`For more directions on how to resolve the issue, see <link>.` is more accurate.
Of course it depends on what is actually at the correct link.

matejmatuska (changes requested):
The overall approach is ok, but there are some problems.

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
tomasfratrik: Changed the wording, and corrected the link.

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
tomasfratrik: It makes more sense! Put it there, and again tested it.

repos/system_upgrade/common/libraries/repofileutils.py
```diff
@@ -11,6 +11,15 @@
     api.current_logger().warning('repofileutils.py: failed to import dnf')
 
 
+class InvalidRepoDefinition(Exception):
+    def __init__(self, msg, repofile=None, repoid=None):
```
matejmatuska, line 15: I am not sure if there is any benefit to have the default arguments here, "None" would look weird in the message. I would leave this up to the caller. Wdyt?

matejmatuska (changes requested):
Just one small suggestion :) code-wise ok, I am going to test it now

matejmatuska (changes requested):
The PR description is outdated, from my testing on RHEL 8 the output looks like this, which is good:
```
2024-08-02 12:17:22.052435 [ERROR] Actor: system_facts
Message: Invalid repository definition: copr:copr.fedorainfracloud.org:group_oamg:leapp in: /etc/yum.repos.d/leapp-copr.repo: The value of "name" field is None, but this is not allowed
Summary:
    Hint: For more directions on how to resolve the issue, see: https://access.redhat.com/solutions/6969001.
2024-08-02 12:17:33.302747 [ERROR] Actor: repositories_blacklist
Message: Actor didn't receive required messages: RepositoriesFacts
```
After the PR desc is updated, squash the commits and briefly describe the changes in the final one, otherwise LGTM.

DO NOT MERGE YET - FOR NOW IT'S NOT TARGETED FOR THE UPCOMING RELEASE

repos/system_upgrade/common/libraries/repofileutils.py
tomasfratrik: You are right

tomasfratrik:
The latest push squashes commits.

fernflower:
Could you please add a unit test?

matejmatuska:
> Could you please add a unit test?

Agree, I don't know how I forgot that, sorry.

tomasfratrik:
> Could you please add a unit test?

I added it!

matejmatuska (approved):
Tests has been added and it works as expected. LGTM

tomasfratrik:
The latest push squashed commits.

repos/system_upgrade/common/libraries/repofileutils.py
```diff
@@ -38,7 +47,10 @@ def parse_repofile(repofile):
     with open(repofile, mode='r') as fp:
         cp = utils.parse_config(fp, strict=False)
         for repoid in cp.sections():
-            data.append(_parse_repository(repoid, dict(cp.items(repoid))))
+            try:
+                data.append(_parse_repository(repoid, dict(cp.items(repoid))))
+            except Exception as e:
```
pirat89, line 52: @tomasfratrik  this is too generic and usually it's not good practice. Specify expected exceptions instead. We know about `ModelViolationError`. Add anything else what you found that could really happen. So far, we know from the history just about this one.

repos/system_upgrade/common/libraries/repofileutils.py
```diff
@@ -11,6 +11,15 @@
     api.current_logger().warning('repofileutils.py: failed to import dnf')
 
 
+class InvalidRepoDefinition(Exception):
+    def __init__(self, msg, repofile, repoid):
```
matejmatuska, line 15: Since this is public and in a shared library, add a docstring describing what is the exception used for.

Also update the docstring in the `parse_repofile` function to document that it can raise this exception and when.

matejmatuska (changes requested):
@pirat89 pointed out that a unit test was added, but only for the system_facts actor. Tests are still missing for the `parse_repofile` and potentially `_parse_repository` functions.

Also docstring have to be included on public members in shared libraries (it would be okay in an actor lib).

repos/system_upgrade/common/libraries/repofileutils.py
tomasfratrik: Sorry forgot to reply, from testing, only ModelViolationError was present

repos/system_upgrade/common/libraries/repofileutils.py
tomasfratrik: Tests were added and docstrings were updated.

repos/system_upgrade/common/libraries/tests/test_repofileutils.py
```diff
@@ -12,6 +15,23 @@ def test_invert_dict():
     assert inv_dict == {'a': [1], 'b': [1, 2]}
 
 
+def test__parse_repository_missing_name():
+    repoid = 'missing-name'
+    data = {'baseurl': 'http://example.com', 'enabled': '1', 'gpgcheck': '1'}
```
matejmatuska, line 20: I think it wouldn't hurt to add some more test cases.

Since you test the `_parse_repository_function` directly, you have more flexibility to test unexpected inputs. 
I would for example add a test with `'name'=None`, some test to make sure invalid type is caught (e.g. `'baseurl'=True`) and also a test with `repoid=None`. 

The inputs right now should always be a string, but this will catch errors if we change the function in the future or call it from somewhere else.

matejmatuska (changes requested):
I would add a little more test cases, otherwise ok.

repos/system_upgrade/common/libraries/tests/test_repofileutils.py
tomasfratrik: I added those test cases.

matejmatuska (approved):
Integration tests on RHEL7 are broken everywhere, not cause by these changes.

LGTM

pirat89:
@matejmatuska seems that the non-els error we discussed has been caused just by infra issue. merging.

PR #1268, opened 2024-07-20
Make sure nobody is trying to make NetworkManager use dhclient in RHEL 10

lkundrak:
dhclient is going away. NetworkManager uses a different DHCP client library even if there's a configuration that would previously make it use dhclient. Make sure the users know what are they doing on upgrade to RHEL 10.


lkundrak:
> Hi! Thank you for your patch. Although it is obvious what the code does, but would you mind adding a unit test? I am not a fan of testing a code that is obvious, however, since it is a piece that can inhibit the entire upgrade, it should be covered.

Absolutely. I think I need a little help tho, I'm not sure what to test here. Should I just synthesize NetworkManagerConfig somehow and test that the right inhibitor report gets emitted?

Is there an example that does something similar I could steal from? (no shame!)

lkundrak:
> > Hi! Thank you for your patch. Although it is obvious what the code does, but would you mind adding a unit test? I am not a fan of testing a code that is obvious, however, since it is a piece that can inhibit the entire upgrade, it should be covered.
> 
> Absolutely. I think I need a little help tho, I'm not sure what to test here. Should I just synthesize NetworkManagerConfig somehow and test that the right inhibitor report gets emitted?
> 
> Is there an example that does something similar I could steal from? (no shame!)

Managed to do it all without any help from an adult and even the linter is happy woo hoo

MichalHe (approved):
Nice job! Tests are passing, and I gave this patch a go manually:
- having `dhcp=dhclient` in the `[main]` section of the NetworkManager config inhibits the upgrade
- not having `dhcp=dhclient` does not inhibit the upgrade

I will keep this unmerged until tomorrow to get a confirmation that this PR is clear from QE perspective.

Thank you for your wonderful contribution :bowing_man:

PR #1270, opened 2024-07-22
Inhibit incompatible ARM upgrades

dkubek:
Due to an incompatibility of RHEL8 bootloader with newer versions of kernel on RHEL9 since version 9.5 the upgrade cannot be performed as the old bootloader is used to load the new versions of kernel during the upgrade.

TODO:
  - [x] add ~KCS~ docs link

JIRA: [41193](https://issues.redhat.com/browse/RHEL-41193), [52993](https://issues.redhat.com/browse/RHEL-52993)

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/checkarmbootloader/actor.py
+++ b/repos/system_upgrade/el8toel9/actors/checkarmbootloader/actor.py
@@ -1,0 +1,25 @@
+import leapp.libraries.actor.checkarmbootloader as checkarmbootloader
+from leapp.actors import Actor
+from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckArmBootloader(Actor):
+    """
+    Inhibit ARM system upgrades on path with incompatible kernel/bootloader
+
+    Due to an incompatibility of RHEL8 bootloader with newer versions of kernel
+    on RHEL9 since version 9.5, the upgrade cannot be performed as the old
+    bootloader cannot load the new kernel when entering the interim phase.
+
+    This is temporary workaround until the issue is resolved.
+
+    """
+
+    name = 'check_arm_bootloader'
+    consumes = ()
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag,)
+
+    def process(self):
+        checkarmbootloader.process()
--- a/repos/system_upgrade/el8toel9/actors/checkarmbootloader/libraries/checkarmbootloader.py
+++ b/repos/system_upgrade/el8toel9/actors/checkarmbootloader/libraries/checkarmbootloader.py
@@ -1,0 +1,54 @@
+from leapp import reporting
+from leapp.libraries.common.config.architecture import ARCH_ARM64, matches_architecture
+from leapp.libraries.common.config.version import get_source_version, get_target_version, matches_target_version
+from leapp.libraries.stdlib import api
+
+
+def _inhibit_upgrade():
+    title = 'Upgrade RHEL {} to RHEL {} not possible for ARM machines.'.format(
+        get_source_version(), get_target_version())
+    summary = (
+        'Due to the incompatibility of the RHEL 8 bootloader with a newer version of kernel on RHEL {}'
+        ' for ARM machines, the direct upgrade cannot be performed to this RHEL'
+        ' system version now. The fix is not expected to be delivered during the RHEL 9.5 lifetime.'
+        .format(get_target_version())
+    )
+
+    reporting.create_report([
+        reporting.Title(title),
+        reporting.Summary(summary),
+        reporting.ExternalLink(
+            title='Known issues for the RHEL 8.10 to RHEL 9.5 upgrade',
+            url='https://red.ht/upgrading-rhel8-to-rhel9-known-issues'),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.INHIBITOR]),
+        reporting.Groups([reporting.Groups.SANITY]),
+        reporting.Remediation(hint=(
+            'To upgrade to the RHEL {} version, first in-place upgrade to RHEL 9.4 instead'
+            ' using the leapp `--target=9.4` option. After you finish the upgrade - including'
+            ' all required manual post-upgrade steps as well -'
+            '  update to the newer minor version using the dnf tool. In case of using Red Hat'
+            ' subscription-manager, do not forget to change the lock version to the newer one'
+            ' or unset the version lock before using DNF to be able to perform the minor version update.'
+            ' You can use e.g. `subscription-manager release --unset` command for that.'
+            .format(get_target_version())
+        )),
+    ])
+
+
+def process():
+    """
+    Check whether the upgrade path will use a target kernel compatible with the source bootloader on ARM systems
+    """
+
+    if not matches_architecture(ARCH_ARM64):
+        api.current_logger().info('Architecture not ARM. Skipping bootloader check.')
+        return
+
+    if matches_target_version('< 9.5'):
+        api.current_logger().info((
+            'Upgrade on ARM architecture on a compatible path ({} to {}). '
+            'Skipping bootloader check.').format(get_source_version(), get_target_version()))
+        return
+
+    _inhibit_upgrade()
--- a/repos/system_upgrade/el8toel9/actors/checkarmbootloader/tests/test_checkarmbootloader.py
+++ b/repos/system_upgrade/el8toel9/actors/checkarmbootloader/tests/test_checkarmbootloader.py
@@ -1,0 +1,67 @@
+import pytest
+
+from leapp import reporting
+from leapp.libraries.actor import checkarmbootloader
+from leapp.libraries.common.config.architecture import ARCH_ARM64, ARCH_SUPPORTED
+from leapp.libraries.common.testutils import create_report_mocked, CurrentActorMocked, logger_mocked
+from leapp.libraries.stdlib import api
+from leapp.utils.report import is_inhibitor
+
+
+@pytest.mark.parametrize("arch", [arch for arch in ARCH_SUPPORTED if not arch == ARCH_ARM64])
+def test_not_x86_64_passes(monkeypatch, arch):
+    """
+    Test no report is generated on an architecture different from ARM
+    """
+
+    monkeypatch.setattr(reporting, "create_report", create_report_mocked())
+    monkeypatch.setattr(api, 'current_logger', logger_mocked())
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(arch=arch))
+
+    checkarmbootloader.process()
+
+    assert 'Architecture not ARM.' in api.current_logger.infomsg[0]
+    assert not reporting.create_report.called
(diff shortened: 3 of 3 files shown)


repos/system_upgrade/el8toel9/actors/checkarmbootloader/libraries/checkarmbootloader.py
```diff
@@ -0,0 +1,47 @@
...
+        reporting.Remediation(hint=(
+            'To upgrade to a RHEL version 9.5 and up, first perform an upgrade '
+            'to version 9.4 using the leapp `--target` option and then perform '
+            'a minor dnf update after the upgrade')),
+    ])
+
+
+def process():
+    """
+    Check whether ARM system uses path with compatible kernel/bootloader.
```
MichalHe, line 33:
```suggestion
    Check whether the upgrade path will use a target kernel compatible with the source bootloader on ARM systems.
```

repos/system_upgrade/el8toel9/actors/checkarmbootloader/libraries/checkarmbootloader.py
```diff
@@ -0,0 +1,47 @@
+from leapp import reporting
+from leapp.libraries.common.config.architecture import ARCH_ARM64, matches_architecture
+from leapp.libraries.common.config.version import get_source_version, get_target_version
+from leapp.libraries.stdlib import api
+
+
+def _inhibit_upgrade():
+    title = 'Upgrade {} to {} unsupported for ARM machines.'.format(
```
MichalHe, line 8:
```suggestion
    title = 'Upgrade from RHEL {} to RHEL {} is unsupported for ARM machines.'.format(
```

MichalHe: Not sure whether "unsupported" is correct word - it is just not possible.  @Rezney

repos/system_upgrade/el8toel9/actors/checkarmbootloader/actor.py
```diff
@@ -0,0 +1,22 @@
...
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckArmBootloader(Actor):
+    """
+    Inhibit ARM system upgrades on path with incompatible kernel/bootloader
+
+    Due to an incompatibility of RHEL8 bootloader with newer versions of kernel
+    on RHEL9 since version 9.5 the upgrade cannot be performed as the old
+    bootloader is used to load the new versions of kernel during the upgrade.
```
MichalHe, line 13: Yes, but this kinda does not describes the issue. The old loader is always used to load the new kernel. Maybe rephrase it to something like `the old bootloader cannot load the new kernel`.

MichalHe (commented):
Cool patch! Let's add the KCS. I will attempt to test it and approve accordingly.

repos/system_upgrade/el8toel9/actors/checkarmbootloader/libraries/checkarmbootloader.py
dkubek: Much better wording. Sure!

repos/system_upgrade/el8toel9/actors/checkarmbootloader/libraries/checkarmbootloader.py
dkubek: We might go with something like "impossible", "not possible", "unfeasible"  as well ...

repos/system_upgrade/el8toel9/actors/checkarmbootloader/actor.py
dkubek: Sure. Rephrased!

repos/system_upgrade/el8toel9/actors/checkarmbootloader/libraries/checkarmbootloader.py
dkubek: not possible it is then!

repos/system_upgrade/el8toel9/actors/checkarmbootloader/libraries/checkarmbootloader.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    Check whether the upgrade path will use a target kernel compatible with the source bootloader on ARM systems
+    """
+
+    if not matches_architecture(ARCH_ARM64):
+        api.current_logger().info('Architecture not ARM. Skipping bootloader check.')
+        return
+
+    target_major, target_minor = tuple(map(int, get_target_version().split('.')))
+    if (target_major, target_minor) < (9, 5):
```
pirat89, line 40: * from leapp.libraries.common.config.version import matches_target_version

```suggestion
    if matches_target_version('< 9.5'):
```

repos/system_upgrade/el8toel9/actors/checkarmbootloader/actor.py
```diff
@@ -0,0 +1,22 @@
...
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckArmBootloader(Actor):
+    """
+    Inhibit ARM system upgrades on path with incompatible kernel/bootloader
+
+    Due to an incompatibility of RHEL8 bootloader with newer versions of kernel
+    on RHEL9 since version 9.5, the upgrade cannot be performed as the old
+    bootloader cannot load the new kernel when entering the interim phase.
```
pirat89, line 13: It's better to make it clear that we do not expect this to be there forever

```suggestion
    bootloader cannot load the new kernel when entering the interim phase.
    
    This is temporary workaround until the issue is resolved.
```

repos/system_upgrade/el8toel9/actors/checkarmbootloader/libraries/checkarmbootloader.py
```diff
@@ -0,0 +1,46 @@
+from leapp import reporting
+from leapp.libraries.common.config.architecture import ARCH_ARM64, matches_architecture
+from leapp.libraries.common.config.version import get_source_version, get_target_version
+from leapp.libraries.stdlib import api
+
+
+def _inhibit_upgrade():
+    title = 'Upgrade RHEL {} to RHEL {} not possible for ARM machines.'.format(
+        get_source_version(), get_target_version())
```
pirat89, line 9: nitpick; you can keep the original if you want to.

Officially the upgrade to RHEL 9.5 is supported only from RHEL 8.10, so from this point I would even completely strip the info from what system they are upgrading. I was also thinking to replace target version directly by 9.5 string, but as people will start soon to test upgrade 8.10 -> 9.6, let's keep it like this until we apply the proper fix - which we know, it must be delivered for 9.6 upgrades for sure. Once the fix is applied, this one actor will be removed.

```suggestion
    title = 'Cannot upgrade to RHEL {} on ARM machines.'.format(get_target_version())
```

repos/system_upgrade/el8toel9/actors/checkarmbootloader/libraries/checkarmbootloader.py
```diff
@@ -0,0 +1,46 @@
...
+    summary = ('Due to the incompatibility of the RHEL8 bootloader with a '
+               'newer version of kernel on RHEL9 a direct upgrade '
+               'cannot be performed for versions 9.5 and up.')
```
pirat89, line 12: This could be little bit disturbing for users as they will read it as: _upgrades to 9.5 and higher will not be possible on arm_. From this point we want to make it more clear that this is temporary solution but
* we do not want to make a promise neither - we should not regarding RH policies
* and we want to be clear that we do not expect any hotfix for RHEL 9.5 neither

```suggestion
    summary = (
        'Due to the incompatibility of the RHEL 8 bootloader with a newer version of kernel on RHEL {}'
        ' for ARM machines, the direct upgrade cannot be performed to this RHEL'
        ' system version now. The fix is not expected to be delivered during the RHEL 9.5 lifetime.'
        .format(get_target_version())
    )        
```

repos/system_upgrade/el8toel9/actors/checkarmbootloader/libraries/checkarmbootloader.py
```diff
@@ -0,0 +1,46 @@
...
+        reporting.ExternalLink(
+            title='Known issues for the RHEL 8.10 to RHEL 9.5 upgrade',
+            url='https://red.ht/upgrading-rhel8-to-rhel9-known-issues'),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.INHIBITOR]),
+        reporting.Groups([reporting.Groups.SANITY]),
+        reporting.Remediation(hint=(
+            'To upgrade to a RHEL version 9.5 and up, first perform an upgrade '
+            'to version 9.4 using the leapp `--target` option and then perform '
+            'a minor dnf update after the upgrade')),
```
pirat89, line 26:
```suggestion
            'To upgrade to the RHEL {} version, first in-place upgrade to RHEL 9.4 instead'
            ' using the leapp `--target=9.4` option. After you finish the upgrade - including'
            ' all required manual post-upgrade steps as well -'
            '  update to the newer minor version using the dnf tool. In case of using Red Hat'
            ' subscription-manager, do not forget to change the lock version to the newer one'
            ' or unset the version lock before using DNF to be able to perform the minor version update.'
            ' You can use e.g. `subscription-manager release --unset` command for that.'
            .format(get_target_version())
        )),
```

pirat89 (changes requested):
Suggested some minor changes in report and one code improvement to use existing functions for checking the system versions.

repos/system_upgrade/el8toel9/actors/checkarmbootloader/libraries/checkarmbootloader.py
dkubek: Much better! I was suspicious we had to have something like that

repos/system_upgrade/el8toel9/actors/checkarmbootloader/actor.py
dkubek: Sure! Fixed

repos/system_upgrade/el8toel9/actors/checkarmbootloader/libraries/checkarmbootloader.py
dkubek: I will keep it as is. When executed as supposed to nothing happens. When this somehow gets run when we don't expect it to, we get more information immediately

repos/system_upgrade/el8toel9/actors/checkarmbootloader/libraries/checkarmbootloader.py
dkubek: Yes I agree. Fixed!

repos/system_upgrade/el8toel9/actors/checkarmbootloader/libraries/checkarmbootloader.py
dkubek: Thanks a lot for expanding this! Fixed

pirat89 (approved):
lgtm and based on the simplicity of the solution and provided tests I believe it works as expected.

PR #1272, opened 2024-08-01
docs: Initial docs structure and infrastructure

matejmatuska:
Introduce the *initial* version of leapp-repository documentation and the required architecture (adapted from the framework docs). Most of the leapp-repository related documents previously in leapp documentation are migrated. Several documents are written from scratch.

---

NOTE: sphinx-autodoc is NOT included, because the way it works is by importing python modules, which are (apart from commands) not used in this repository, therefore it doesn't work.

In the future we could use sphinx-autoapi (https://sphinx-autoapi.readthedocs.io/en/latest/) which autogenerates docs by actually parsing source code. From my testing it will work with some tweaks.

---

TODO:
- [x] Add initial infrastructure
- [X] Migrate the docs regarding this project from leapp (framework) docs
- [ ] Add AutoAPI to generate code docs (see above) - will be done in a follow up


docs/Makefile
```diff
@@ -0,0 +1,31 @@
...
+SPHINXPROJ    = Leapp repository
+SOURCEDIR     = source
+BUILDDIR      = build
+
+all: clean html
+
+help:
+	@source venv/bin/activate && \
+	$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+livehtml:
```
pirat89, line 13:
```suggestion
livehtml: html
```

this will automatically build html without the need to think about it

pirat89 (approved):
Yaaahoooooo!!!

PR #1274, opened 2024-08-08
Explicitly show the directory creation.

lonicerae:
It's better to show the directory creation when running leapp. This can improve the readability of the debug file.


pirat89 (approved):
make sense. lgtm. thanks for the contribution

PR #1275, opened 2024-08-08
Workaround for ARM Upgrades from RHEL8 to RHEL9.5 onwards (GRUB and kernel incompatibility)

dkubek:
This actor addresses an issue encountered during the upgrade process on ARM systems. Specifically, the problem arises due to an incompatibility between the GRUB bootloader used in RHEL 8 and the newer kernels from RHEL 9.5 onwards. When the kernel of the target system is loaded using the bootloader from the source system, this incompatibility causes the bootloader to crash, halting the upgrade.

To mitigate this issue, the following steps are implemented:

Before the Upgrade (handled by ``AddArmBootloaderWorkaround``):

* Create a new Upgrade EFI entry:
  - A new EFI boot entry is created and populated with the updated RHEL 9 bootloader that is compatible with the new kernel.

* Preserve the original EFI boot entry and GRUB configuration:
  - The original EFI boot entry and GRUB configuration remain unchanged to ensure system stability.

After the Upgrade (handled by ``RemoveUpgradeEFIEntry``):

* Remove the upgrade EFI boot entry:
  - The temporary EFI boot entry created for the upgrade is removed to restore the system to its pre-upgrade state.

TODO:

- [ ] Add description of the grub library to docs

JIRA: [41193](https://issues.redhat.com/browse/RHEL-41193)

### DIFF
--- a/repos/system_upgrade/common/libraries/grub.py
+++ b/repos/system_upgrade/common/libraries/grub.py
@@ -1,9 +1,203 @@
 import os
+import re
 
 from leapp.exceptions import StopActorExecution
 from leapp.libraries.common import mdraid
 from leapp.libraries.stdlib import api, CalledProcessError, run
 from leapp.utils.deprecation import deprecated
+
+EFI_MOUNTPOINT = '/boot/efi/'
+"""The path to the required mountpoint for ESP."""
+
+GRUB2_BIOS_ENTRYPOINT = '/boot/grub2'
+"""The entrypoint path of the BIOS GRUB2"""
+
+GRUB2_BIOS_ENV_FILE = os.path.join(GRUB2_BIOS_ENTRYPOINT, 'grubenv')
+"""The path to the env file for GRUB2 in BIOS"""
+
+
+def canonical_path_to_efi_format(canonical_path):
+    r"""Transform the canonical path to the UEFI format.
+
+    e.g. /boot/efi/EFI/redhat/shimx64.efi -> \EFI\redhat\shimx64.efi
+    (just single backslash; so the string needs to be put into apostrophes
+    when used for /usr/sbin/efibootmgr cmd)
+
+    The path has to start with /boot/efi otherwise the path is invalid for UEFI.
+    """
+
+    # We want to keep the last "/" of the EFI_MOUNTPOINT
+    return canonical_path.replace(EFI_MOUNTPOINT[:-1], "").replace("/", "\\")
+
+
+class EFIBootLoaderEntry(object):
+    """
+    Representation of an UEFI boot loader entry.
+    """
+    # pylint: disable=eq-without-hash
+
+    def __init__(self, boot_number, label, active, efi_bin_source):
+        self.boot_number = boot_number
+        """Expected string, e.g. '0001'. """
+
+        self.label = label
+        """Label of the UEFI entry. E.g. 'Redhat'"""
+
+        self.active = active
+        """True when the UEFI entry is active (asterisk is present next to the boot number)"""
+
+        self.efi_bin_source = efi_bin_source
+        """Source of the UEFI binary.
+
+        It could contain various values, e.g.:
+            FvVol(7cb8bdc9-f8eb-4f34-aaea-3ee4af6516a1)/FvFile(462caa21-7614-4503-836e-8ab6f4662331)
+            HD(1,GPT,28c77f6b-3cd0-4b22-985f-c99903835d79,0x800,0x12c000)/File(\\EFI\\redhat\\shimx64.efi)
+            PciRoot(0x0)/Pci(0x2,0x3)/Pci(0x0,0x0)N.....YM....R,Y.
+        """
+
+    def __eq__(self, other):
+        return all(
+            [
+                self.boot_number == other.boot_number,
+                self.label == other.label,
+                self.active == other.active,
+                self.efi_bin_source == other.efi_bin_source,
+            ]
+        )
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    def __repr__(self):
+        return 'EFIBootLoaderEntry({boot_number}, {label}, {active}, {efi_bin_source})'.format(
+            boot_number=repr(self.boot_number),
+            label=repr(self.label),
+            active=repr(self.active),
+            efi_bin_source=repr(self.efi_bin_source)
+        )
+
+    def is_referring_to_file(self):
+        """Return True when the boot source is a file.
+
+        Some sources could refer e.g. to PXE boot. Return true if the source
+        refers to a file ("ends with /File(...path...)")
+
+        Does not matter whether the file exists or not.
+        """
+        return '/File(\\' in self.efi_bin_source
+
+    @staticmethod
+    def _efi_path_to_canonical(efi_path):
+        return os.path.join(EFI_MOUNTPOINT, efi_path.replace("\\", "/").lstrip("/"))
+
+    def get_canonical_path(self):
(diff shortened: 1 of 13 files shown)


repos/system_upgrade/el8toel9/actors/removeupgradeefientry/tests/test_removeupgradeefientry.py
```diff
@@ -0,0 +1,103 @@
...
+    monkeypatch.setattr(removeupgradeefientry, 'run', mock_run)
+
+    removeupgradeefientry.remove_upgrade_efi_entry()
+
+    assert ['/bin/mount', '/boot'] in run_calls
+    assert ['/bin/mount', '/boot/efi'] in run_calls
+    assert ['/usr/sbin/efibootmgr', '--delete-bootnum', '--bootnum', '0002'] in run_calls
+    assert ['rm', '-rf', removeupgradeefientry.LEAPP_EFIDIR_CANONICAL_PATH] in run_calls
+    assert ['/usr/sbin/efibootmgr', '--bootnext', '0001'] in run_calls
+    assert ['/bin/mount', '-a'] in run_calls
```
matejmatuska, line 103: The order of these calls is important, I would suggest popping the lists from `run_calls` when comparing them to also test that.

repos/system_upgrade/el8toel9/actors/addarmbootloaderworkaround/tests/test_addarmbootloaderworkaround.py
```diff
@@ -0,0 +1,159 @@
...
+import pytest
+
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.actor import addupgradebootloader
+from leapp.libraries.common.testutils import CurrentActorMocked, logger_mocked, make_OSError
+from leapp.libraries.stdlib import api
+from leapp.models import TargetUserSpaceInfo
+
+
+class IsolatedActionsMocked(object):
```
matejmatuska, line 12: If I am not missing something this is unused.

repos/system_upgrade/el8toel9/actors/addarmbootloaderworkaround/tests/test_addarmbootloaderworkaround.py
matejmatuska, line 1: The `process()` and `_add_upgrade_boot_entry()` functions are untested, I would expect some tests for them.

repos/system_upgrade/common/libraries/grub.py
```diff
@@ -1,10 +1,207 @@
...
+    when used for /usr/sbin/efibootmgr cmd)
+
+    The path has to start with /boot/efi otherwise the path is invalid for UEFI.
+    """
+
+    # We want to keep the last "/" of the EFI_MOUNTPOINT
+    return canonical_path.replace(EFI_MOUNTPOINT[:-1], "").replace("/", "\\")
+
+
+class EFIBootLoader(object):
```
matejmatuska, line 33: What about:
```suggestion
class EFIBootLoaderEntry(object):
```
or does it conflict with the model?

repos/system_upgrade/common/libraries/grub.py
```diff
@@ -1,10 +1,207 @@
...
+        self.current_bootnum = self._parse_key_value(bootmgr_output, 'BootCurrent')
+
+        if self.current_bootnum is None:
+            raise StopActorExecution('UEFI: Unable to detect current boot number.')
+
+    def _parse_next_bootnum(self, bootmgr_output):
+        # e.g.: BootNext: 0002
+        self.next_bootnum = self._parse_key_value(bootmgr_output, 'BootNext')
+
+        if self.current_bootnum is None:
```
matejmatuska, line 184: Maybe a typo?
```suggestion
        if self.next_bootnum is None:
```

Either way is it guaranteed that BootNext is always present? I quickly checked on my system and it's not there.

repos/system_upgrade/common/libraries/grub.py
```diff
@@ -1,10 +1,207 @@
...
+            raise StopActorExecution('UEFI: Unable to detect next boot number.')
+
+    def _parse_boot_order(self, bootmgr_output):
+        # e.g.:  BootOrder: 0001,0002,0000,0003
+        for line in bootmgr_output.splitlines():
+            if line.startswith('BootOrder:'):
+                self.boot_order = tuple(line.split(':')[1].strip().split(','))
+                return
+
+        raise StopActorExecution('UEFI: Unable to detect current boot order.')
```
matejmatuska, line 194: Not a blocker, but why not reuse the `_parse_key_value()` function here?

matejmatuska (changes requested):
A few comments, other than that it's OK.

repos/system_upgrade/common/libraries/grub.py
matejmatuska: Looking at the test, it seems like it's assumed it can be None, but then I am not sure why the error is raised?  Is it because it's assumed one would use the current_bootnum as a fallback?

repos/system_upgrade/common/libraries/grub.py
```diff
@@ -33,39 +230,126 @@ def blk_dev_from_partition(partition):
...
+
+    if not os.path.exists(EFI_MOUNTPOINT) or not os.path.ismount(EFI_MOUNTPOINT):
+        raise StopActorExecution(
+            'The UEFI has been detected but the ESP is not mounted in /boot/efi as required.'
+        )
+
+    return _get_partition('/boot/efi/')
+
+
+def get_blk_device(device):
```
matejmatuska, line 307: How is this different from the existing `blk_dev_from_partition()` function?
It could use the docstring though :)

repos/system_upgrade/el8toel9/actors/removeupgradeefientry/tests/test_removeupgradeefientry.py
dkubek: Good idea, thanks! Changed.

repos/system_upgrade/el8toel9/actors/addarmbootloaderworkaround/tests/test_addarmbootloaderworkaround.py
dkubek: It was a leftover from some test that were deleted at the time. I've made it relevant again with new tests.

repos/system_upgrade/el8toel9/actors/addarmbootloaderworkaround/tests/test_addarmbootloaderworkaround.py
dkubek: Added!

repos/system_upgrade/common/libraries/grub.py
dkubek: Sure thing! I liked the shorter name, but this is clearer. Changed!

repos/system_upgrade/common/libraries/grub.py
dkubek: OOps! copy-paste error! Nice catch! Fixed

repos/system_upgrade/common/libraries/grub.py
dkubek: Yeah... why not? Added! Thanks

repos/system_upgrade/common/libraries/grub.py
dkubek: another refactoring leftover. Thanks!

matejmatuska (approved):
Code-wise LGTM

PR #1276, opened 2024-08-09
Fix mocks that cause tests to fail outside of containers

matejmatuska:
See commit messages for details.

To reproduce you can use, for example: `PYTHON_VENV=python3.6 REPOSITORIES='common,el8toel9' make install-deps-fedora test_no_lint`

### DIFF
--- a/repos/system_upgrade/common/libraries/grub.py
+++ b/repos/system_upgrade/common/libraries/grub.py
@@ -36,7 +36,7 @@ def blk_dev_from_partition(partition):
         api.current_logger().warning(
             'Could not get parent device of {} partition'.format(partition)
         )
-        raise StopActorExecution()
+        raise StopActorExecution()  # TODO: return some meaningful value/error
     # lsblk "-s" option prints dependencies in inverse order, so the parent device will always
     # be the last or the only device.
     # Command result example:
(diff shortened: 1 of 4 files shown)


repos/system_upgrade/common/libraries/grub.py
```diff
@@ -36,7 +36,7 @@ def blk_dev_from_partition(partition):
         api.current_logger().warning(
             'Could not get parent device of {} partition'.format(partition)
         )
-        raise StopActorExecution()
+        raise StopActorExecution()  # TODO: return some meaningful value/error
```
pirat89, line 39: :+1:  :heavy_plus_sign:  that's bug introduced basically when the private actor's library has been moved to shared library instead. Checking the library, it should be fixed everywhere, but let's keep it for another PR. Good catch!

pirat89:
@matejmatuska on my machine, there is still one test failing:
```
================================================================================================================== FAILURES ==================================================================================================================
________________________________________________________________________________________________________________ test_nm_conn ________________________________________________________________________________________________________________

monkeypatch = <_pytest.monkeypatch.MonkeyPatch object at 0x7f6ac3dbb1d0>

    @pytest.mark.skipif(not nmconnscanner.libnm_available, reason="NetworkManager g-ir not installed")
    def test_nm_conn(monkeypatch):
        """
        Check a basic keyfile
        """
    
        monkeypatch.setattr(nmconnscanner.os, "listdir", _listdir_nm_conn)
        monkeypatch.setattr(api, "produce", produce_mocked())
        monkeypatch.setattr(nmconnscanner.GLib.KeyFile, "load_from_file", _load_from_file)
        nmconnscanner.process()
    
        assert api.produce.called == 1
        assert len(api.produce.model_instances) == 1
        nm_conn = api.produce.model_instances[0]
        assert isinstance(nm_conn, NetworkManagerConnection)
        assert nm_conn.filename == "/etc/NetworkManager/system-connections/conn1.nmconnection"
        assert len(nm_conn.settings) == 3
        assert nm_conn.settings[0].name == "connection"
>       assert len(nm_conn.settings[0].properties) == 4
E       assert 3 == 4
E         +3
E         -4

repos/system_upgrade/el8toel9/actors/networkmanagerconnectionscanner/tests/unit_test_networkmanagerconnectionscanner.py:78: AssertionError
```

Based on the skip condition, it seems it's still reading something from the system. In container it does not appear as the test is skipped.

pirat89 (approved):
lgtm. discovere one additional problem with nmconnscanner tests but keeping that for separate PR in future (see my comment).

PR #1278, opened 2024-08-12
targetuserspacecreator: stop skipping `directory-hash`

t184256:
If https://github.com/oamg/leapp/pull/880 was the only FD leak,
we should be able to remove this workaround.


pirat89 (approved):
tested manually and we can really drop it once the leapp stdlib is fixed.

The change has been introduced during the work on the current release so it negates original changelog.

PR #1281, opened 2024-08-16
First draft of running linting and unit tests separately.

abadger:
This way we still get unit test output even if the linters fail.

We might be able to do better than this.  For instance, since everything is static analysis, we might be able to run this once on each version of python we care about with all of the repositories enabled instead of running it once for each combination of python version and upgrade target.


pirat89 (approved):
I like it. LGTM and it works. merging

PR #1285, opened 2024-08-20
don't require all versions to be defined for obsoleted keys

evgeni:
in releases where we do not have any obsoleted keys, we still had to
define an entry (with an empty list), as otherwise the code would fail

instead, we can catch the KeyError and carry on as nothing happened


pirat89 (approved):
make sense. lgtm. thanks

PR #1290, opened 2024-08-29
ReadOfKernelArgsError: fix no attribute error

yuravk:
- AttributeError: module 'leapp.reporting' has no attribute 'Hints'


pirat89 (approved):
@yuravk   thank you for the contribution! I see we really missed that one. Could you also add unit-test to cover this part of the code? it's not blocker, but we see it has been missing all the time, which is one of reason why we missed this.

PR #1292, opened 2024-09-02
Update references from master branch to main

vojtechsokol:
Focus was on making the CI and GitHub actions work after the default
branch was switched from master to main.

See: OAMG-4907


matejmatuska (approved):
LGTM, all actions passed. However the commit message could mention it's about Github actions (or CI).

PR #1294, opened 2024-09-05
overlaygen: cap max sparse file size

MichalHe:
If the source system has a very large disk attached, leapp would attempt to create similarly large sparse file, possibly hitting the max file size limits of the filesystem where these images are created. This patch caps the maximum size of any of the created sparse files to 1TB, preventing such issues.

Jira ref: RHEL-57064

### DIFF
--- a/repos/system_upgrade/common/libraries/overlaygen.py
+++ b/repos/system_upgrade/common/libraries/overlaygen.py
@@ -68,6 +68,27 @@ or close to that size, stay always with this minimal protected size defined by
 this constant.
 """
 
+_MAX_DISK_IMAGE_SIZE_MB = 2**20  # 1*TB
+"""
+Maximum size of the created (sparse) images.
+
+Defaults to 1TB. If a disk with capacity larger than _MAX_DISK_IMAGE_SIZE_MB
+is mounted on the system, the corresponding image used to store overlay
+modifications will be capped to _MAX_DISK_IMAGE_SIZE_MB.
+
+Engineering rationale:
+   This constant was introduced to prevent leapp from creating files that are
+   virtually larger than the maximum file size supported by the file system.
+   E.g. if the source system hosts /var/lib/leapp on EXT4, then we cannot
+   create a file larger than 16TB.
+   We create these "disk images" to be able to verify the system has enough
+   disk space to perform the RPM upgrade transaction. From our experience,
+   we are not aware of any system which could have installed so much content
+   by RPMs that we would need 1TB of the free space on a single FS. Therefore,
+   we consider this value as safe while preventing us from exceeding FS
+   limits.
+"""
+
 
 MountPoints = namedtuple('MountPoints', ['fs_file', 'fs_vfstype'])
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -379,7 +379,7 @@ test_no_lint:
 	cd repos/system_upgrade/el7toel8/; \
 	snactor workflow sanity-check ipu && \
 	cd - && \
-	$(_PYTHON_VENV) -m pytest $(REPORT_ARG) $(TEST_PATHS) $(LIBRARY_PATH) $(PYTEST_ARGS)
+	$(_PYTHON_VENV) -m pytest $(REPORT_ARG) $(LIBRARY_PATH) $(PYTEST_ARGS)
```
pirat89, line 382: good catch!

repos/system_upgrade/common/libraries/overlaygen.py
```diff
@@ -68,6 +68,21 @@
...
+
+Defaults to 1TB. If a disk with capacity larger than _MAX_DISK_IMAGE_SIZE_MB
+is mounted on the system, the corresponding image used to store overlay
+modifications will be capped to _MAX_DISK_IMAGE_SIZE_MB.
+
+Engineering rationale:
+   This constant was introduced to prevent leapp from creating files that are
+   virtually larger than the maximum file size supported by the file system.
+   E.g. if the source system hosts /var/lib/leapp on EXT4, then we cannot
+   create a file larger than 16TB.
```
pirat89, line 83: I would extend the reasoning why we chose 1TB.

```suggestion
   This constant was introduced to prevent leapp from creating files that are
   virtually larger than the maximum file size supported by the file system.
   E.g. if the source system hosts /var/lib/leapp on EXT4, then we cannot
   create a file larger than 16TB.
   We create these "disk images" to be able to verify the system has enough
   disk space to perform the RPM upgrade transaction. From our experience,
   we are not aware of any system which could have installed so much content
   by RPMs, that we would need 1TB of the free space on a single FS. So we consider
   this cap to be safe enough and it does not cost us too much.
```

Makefile
MichalHe: Actually, I committed this by accident - if you do `make test_no_lint` you want to also test actors. I have removed the `$(TEST_PATHS)` because there was no way (at least obvious to me) to test just common libraries.

repos/system_upgrade/common/libraries/overlaygen.py
```diff
@@ -68,6 +68,21 @@
 this constant.
 """
 
+_MAX_DISK_IMAGE_SIZE_MB = 1024**2  # 1*TB
```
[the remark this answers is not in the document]
MichalHe: I don't care as long as I can easily parse the resulting number. Therefore, the suggested solution is overall more preferred, given my indifference and your preference (even if slight) :grin:

Makefile
pirat89: I am actually puzzled now :-D because I checked the Makefile as the change looked weird to me originally but going through it, it started to make sense to me as a good step. Reading your explanation and the Makefile again, I am now trying to find out why I thought that  :D
 :face_in_clouds:  + :palm_tree:  -->  :facepalm:
 
 Even Homer nods :shrug:

pirat89:
Tested. works as expected.
* prepare machine
```
dd if=/dev/zero of=/root/tbdisk.img bs=1M count=0 seek=10875978  # ~11TB
mkfs.xfs  -l size=32m -f /root/tbdisk.img
mkdir /mnt/big_disk
echo >>/etc/fstab "/root/tbdisk.img /mnt/big_disk xfs loop 0 0"
systemctl daemon-reload
mount -a
```

* original / upstream build (trying to create ~11TB img: `10260080` MiB)
```bash
# leapp-inspector actors --actor target_userspace_creator --terminal-like | grep -e "- \['/bin/dd'," | grep big_disk
    - ['/bin/dd', 'if=/dev/zero', 'of=/var/lib/leapp/scratch/diskimages/root_mnt_big_disk', 'bs=1M', 'count=0', 'seek=10260080']
```

* PR build (trying to create just 1TB img: `1048576` MiB):
```bash
# leapp-inspector actors --actor target_userspace_creator --terminal-like | grep -e "- \['/bin/dd'," | grep big_disk
    - ['/bin/dd', 'if=/dev/zero', 'of=/var/lib/leapp/scratch/diskimages/root_mnt_big_disk', 'bs=1M', 'count=0', 'seek=1048576']
```

pirat89 (approved):
lgtm and works as expected (see comment above)

PR #1296, opened 2024-09-17
pes_events_scanner: fix processing of RpmTransactionTasks

pirat89:
RpmTransactionTasks messages have higher priority than instructions based on PES data. Previously, if multiple such messages existed with duplicate instructions, this could lead to the crash of the actor - especially in case when an existing package has been asked to be removed several times. Ensure the occurance of each instruction is unique (list -> set).

jira: RHEL-50076

Note: just written on knees without testing. Unit tests should be updated too.

### DIFF
--- a/repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
+++ b/repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
@@ -78,19 +78,22 @@ def get_installed_pkgs():
 
 def get_transaction_configuration():
     """
-    Get pkgs to install, keep and remove from the user configuration files in /etc/leapp/transaction/.
+    Get pkgs to install, keep and remove from RpmTransactionTasks messages.
 
-    These configuration files have higher priority than PES data.
-    :return: RpmTransactionTasks model instance
+    Note these messages reflects inputs from various actors and configuration
+    files in /etc/leapp/transaction/. As these are explicit instruction, they
+    have higher priority than instructions from PES data.
+
+    :return: TransactionConfiguration
     """
-    transaction_configuration = TransactionConfiguration(to_install=[], to_remove=[], to_keep=[])
+    transaction_configuration = TransactionConfiguration(to_install=set(), to_remove=set(), to_keep=set())
 
     _Pkg = partial(Package, repository=None, modulestream=None)
 
     for tasks in api.consume(RpmTransactionTasks):
-        transaction_configuration.to_install.extend(_Pkg(name=pkg_name) for pkg_name in tasks.to_install)
-        transaction_configuration.to_remove.extend(_Pkg(name=pkg_name) for pkg_name in tasks.to_remove)
-        transaction_configuration.to_keep.extend(_Pkg(name=pkg_name) for pkg_name in tasks.to_keep)
+        transaction_configuration.to_install.update(_Pkg(name=pkg_name) for pkg_name in tasks.to_install)
+        transaction_configuration.to_remove.update(_Pkg(name=pkg_name) for pkg_name in tasks.to_remove)
+        transaction_configuration.to_keep.update(_Pkg(name=pkg_name) for pkg_name in tasks.to_keep)
     return transaction_configuration
 
 


MichalHe:
@pirat89 Although this fixes the crash, I believe that transaction configuration is handled in a weird manner altogether. If user says `to_remove=['tomcat']` then the output of the `pes_events_scanner` actor will not contain `tomcat` in its `to_remove` fields -- see the failing unit test. Not sure whether this is intended (unlikely) but we (e.g. me) can fix this problem within this PR.

pirat89:
> @pirat89 Although this fixes the crash, I believe that transaction configuration is handled in a weird manner altogether. If user says `to_remove=['tomcat']` then the output of the `pes_events_scanner` actor will not contain `tomcat` in its `to_remove` fields -- see the failing unit test. Not sure whether this is intended (unlikely) but we (e.g. me) can fix this problem within this PR.

you are right that this sounds weird. feel free to fix it properly in this PR - or close this one and open new one if you want. this has been really a draft to ilustrate what I found  and initial possible fix of the problem. based on your finding, it seems there is something dirty still around.

MichalHe:
@pirat89 I added a post-processing step that will add transaction configuration's fields to the scanner's output.

MichalHe:
I have tested the patch on a setup where I introduced an actor that produces:
```python
self.produce(RpmTransactionTasks(to_remove=['tomcat', 'tomcat']))
self.produce(RpmTransactionTasks(to_remove=['tomcat']))
```

Using current upstream build, the upgrade crashed with the following error:
```
2025-01-09 16:31:15.324 ERROR    PID: 45901 leapp.workflow.FactsCollection: Actor pes_events_scanner has crashed: Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/leapp/repository/actor_definition.py", line 75, in _do_run
  File "/usr/lib/python3.6/site-packages/leapp/actors/__init__.py", line 316, in run
    self.process(*args)
  File "/etc/leapp/repos.d/system_upgrade/common/actors/peseventsscanner/actor.py", line 41, in process
    process()
  File "/etc/leapp/repos.d/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py", line 512, in process
    source_pkgs = apply_transaction_configuration(source_pkgs)
  File "/etc/leapp/repos.d/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py", line 482, in apply_transaction_configuration
    source_pkgs_with_conf_applied.remove(pkg_name_to_pkg_info_map[pkg.name])
KeyError: tomcat:rhel-8-for-x86_64-appstream-rpms
```

Using build based on this PR results with no errors and the upgrade continues normally, without raising the above error.

Hence, I believe this PR is ready for review.

repos/system_upgrade/common/actors/peseventsscanner/libraries/pes_events_scanner.py
```diff
@@ -501,38 +503,87 @@ def remove_leapp_related_events(events):
...
+def include_instructions_from_transaction_configuration(rpm_tasks, transaction_configuration, installed_pkgs):
+    """
+    Extend current rpm_tasks applying data from transaction_configuration
+
+    :param PESRpmTransactionTasks rpm_tasks: Currently calculated rpm tasks based on PES data.
+    :param TransactionConfiguration transaction_configuration: Tasked configurated by user manually.
+    :param set(str) installed_pkgs: Set of distribution signed packages installed on the system.
+    :returns: updated tasks respecting configuration changes made by user
+    :rtype: PESRpmTransactionTasks
+    """
```
pirat89, line 515: @MichalHe I have updated the docstring of the function so it's easier to go through just inside the scope of the function without the need to check what are the input arguments.

pirat89:
giving +1 for @MichalHe changes. considering it good to go, but as original author of the PR, waiting for someone else to review it and merge it - even when my changes are minor in this PR.

MichalHe:
The failing 9>10 tests are not related to this PR. Since two of us are OK with this PR, I am going to merge this.

MichalHe (approved):
looks OK to me, hence we have agreement of 2 developers. e2e tests are passing, so lets merge this

PR #1297, opened 2024-09-25
Skip checking files under .../directory-hash/ dir

pirat89:
* The main reason for this change is to improve performance and reduce flood of logs for the content that does not seem to be important to check for the upgrade process.

The directory has been relatively recently added to ca-certificates rpm on EL 9+ systems mostly to improve performance of OpenSSL and the content does not seem to be important for the IPU process. The high number of files takes too much time to evaluate and causes flood of logs that are not important.

This is updated solution that we drop originally: 60f500e59bb92


pirat89:
The original solution checked the relative path of each file (one by one) instead of the directory. Seems to me that skipping the whole directory in this case it's a better option.

pirat89:
@oamg/developers  review please. the IPU 7 - 8 are failing due to test issue, the 8.10 -> 9.5 failed due to infra issue.

Rezney (approved):

PR #1299, opened 2024-09-27
pylint: exclude rule: too-many-positional-arguments (code: R0917)

pirat89:
New version of Pylint have the rule for checking of positional arguments - complaining when more than 4 positional arguments exists. We do not want to refactor the code to make it happy and the default value cannot be set right now - that's planned for future Pylint versions at this moment. So excluding this rule.

For more info:
  * https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/too-many-positional-arguments.html

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -41,6 +41,8 @@ disable=
  consider-using-from-import,
  use-list-literal,
  use-dict-literal,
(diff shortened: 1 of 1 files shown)


evgeni (approved):

pirat89:
@evgeni  thanks! merging

PR #1301, opened 2024-10-10
Replace mirror.centos.org with vault.centos.org Centos 7 Containerfile

matejmatuska:
As mirror.centos.org is dead, replace mirrorlist with baseurl pointing to vault.centos.org in utils/container-builds/Containerfile.centos7.

To test, remove delete the `leapp-repo-build-el7` container image (`podman rmi leapp-repo-build-el7`) and run `BUILD_CONTAINER=el7 make build_container`. The image should be successfully created, whereas before it would fail.

### DIFF
--- a/utils/container-builds/Containerfile.centos7
+++ b/utils/container-builds/Containerfile.centos7
@@ -2,6 +2,11 @@ FROM centos:7
 
 VOLUME /repo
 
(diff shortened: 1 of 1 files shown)


pirat89 (approved):
works as expected :100:

PR #1304, opened 2024-11-05
Add el9toel10 actor to handle symlink -> directory with ruby IRB.

jackorp:
The `/usr/share/ruby/irb` path is a symlink in RHEL 9, but a regular directory in RHEL 10.

Simply remove the symlink to then let the RPM mechanism create the correct directory on update.

Users should not expect to ever retain anything in this directory.

---

Fix for [OAMG-12187](https://issues.redhat.com/browse/OAMG-12187) .

instead of `readlink` present in el8->el9 actor, I just used `test -L` since here we only care if there is any symlink or not, contents or validity of where it is pointing is not important for the upgrade.

I left the rest as-is from a copy paste of the actor, I think naming and such still pretty much fits.

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/registerrubyirbadjustment/actor.py
+++ b/repos/system_upgrade/el8toel9/actors/registerrubyirbadjustment/actor.py
@@ -5,7 +5,16 @@ from leapp.tags import FactsPhaseTag, IPUWorkflowTag
 
 class RegisterRubyIRBAdjustment(Actor):
     """
-    Registers a workaround which will adjust the Ruby IRB directories during the upgrade.
+    Register a workaround to allow rubygem-irb's directory -> symlink conversion.
+
+    The /usr/share/ruby/irb has been moved from a directory to a symlink
+    in RHEL 9 and this conversion was not handled on RPM level.
+    This leads to DNF reporting package file conflicts when a major upgrade
+    is attempted and rubygem-irb (or ruby-irb) is installed.
+
+    Register "handlerubyirbsymlink" script that removes the directory prior
+    to DNF upgrade and allows it to create the expected symlink in place of
+    the removed directory.
     """
 
     name = 'register_ruby_irb_adjustment'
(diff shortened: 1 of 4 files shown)


jackorp:

pirat89 (commented):
@jackorp  checking the solution, wouldn't be better to generalize it and put it to the common repo instead of have two almost same solution for 8 -> 9 & 9 -> 10?

jackorp:
> checking the solution, wouldn't be better to generalize it and put it to the common repo instead of have two almost same solution for 8 -> 9 & 9 -> 10?

That depends, are there other packages that stand to gain from such work? Is that a refactor you'd consider worthwhile endeavor?

From my view of a packager who fixes the upgrade for just this problematic part of the package, I prefer simpler copy paste for these 2 upgrade cases, but I don't know if this is now more common problem or still applies only to a package or few. Ruby is back in line with Fedora with this and hopefully that's the end of the broken symlink story for ruby WRT LEAPP...

I am also going off of the comment on bugzilla regarding not needing generic solution https://bugzilla.redhat.com/show_bug.cgi?id=2030627#c13 If the world around changed in the time being making that comment not accurate then we can figure something out.

pirat89:
> > checking the solution, wouldn't be better to generalize it and put it to the common repo instead of have two almost same solution for 8 -> 9 & 9 -> 10?
> 
> That depends, are there other packages that stand to gain from such work? Is that a refactor you'd consider worthwhile endeavor?
> 
> From my view of a packager who fixes the upgrade for just this problematic part of the package, I prefer simpler copy paste for these 2 upgrade cases, but I don't know if this is now more common problem or still applies only to a package or few. Ruby is back in line with Fedora with this and hopefully that's the end of the broken symlink story for ruby WRT LEAPP...
> 
> I am also going off of the comment on bugzilla regarding not needing generic solution https://bugzilla.redhat.com/show_bug.cgi?id=2030627#c13 If the world around changed in the time being making that comment not accurate then we can figure something out.

I think that you are speaking about something different than me. The comment in BZ is not related anyhow to what I am suggesting :)

pirat89:
Discussed on the call. We will keep the actors separate, as the reason for both IPUs is different for this handling, but docstring should be updated in both actors to make it clear.

jackorp:
@pirat89 updated the docstrings.I did the docstring update for el8->el9 in separate commit as I don't see it as something hide-able in the el9toel10 commit. And squashing is easier than splitting imo...

pirat89 (approved):

pirat89:
@jackorp  merged! thanks for the contribution!!

PR #1306, opened 2024-11-12
IPU 7 -> 8: scangrubdevpartitionlayout: Skip warning msgs

pirat89:
The fdisk output can contain warning msgs when a partition is not alligned on physical sector boundary, like:
    Partition 4 does not start on physical sector boundary.
We know that in case of MBR the line we expect to parse always starts with canonical path. So let's skip all lines which does not start with '/'.

jira: https://issues.redhat.com/browse/RHEL-50947


andywaltlova (approved):

pirat89:
@andywaltlova  thanks for the prompt review! :heart:

PR #1309, opened 2024-11-14
Enable IPU for EL 9.6 (and drop EL 8.8/9.2)

pirat89:
* Add product certificates for RHEL 9.6
* Introduce upgrade path 8.10 -> 9.6
* Drop IPUs related to EL 8.8 and 9.2
  * This will not be supported in this release.
  * Keeping for now still IPU 8.10 -> 9.5 as it is a fresh release so it has a value for us to run tests there. We will drop it later during this lifecycle (CTC-2?).
* Update tests in packit
  * Note that tests for 9.6 could be failing for a while until composes are created.

jira: RHEL-67621


tomasfratrik (approved):
lgtm!

PR #1312, opened 2024-11-19
feat(net-naming-scheme): enable by default

MichalHe:
This commit enables the use of net.naming-scheme for 8>9 upgrades by default. The previously used environmental variable `LEAPP_USE_NET_NAMING_SCHEMES` is replaced with `LEAPP_DISABLE_NET_NAMING_SCHEMES` with inverse semantics.


matejmatuska:
Tested:
On RHEL 8 with `LEAPP_DISABLE_NET_NAMING_SCHEMES=0`, the actor does run:
```
Skipping generation of .link files renaming NICs as net.naming-scheme {LEAPP_DISABLE_NET_NAMING_SCHEMES != 1} is enabled and upgrade is 8>9
```

On RHEL 7, 8 with LEAPP_DISABLE_NET_NAMING_SCHEMES=1 and 9 it's not logger and correctly disabled.

LGTM.

PR #1313, opened 2024-11-20
fix(add_upgrade_boot_entry): convert arg list into a tuple

MichalHe:
Convert collected rd.lvm args into a tuple before trying to make a set with one of the elements being the args. As list is not hashable, this causes the actor to crash.

### DIFF
--- a/repos/system_upgrade/common/actors/addupgradebootentry/tests/unit_test_addupgradebootentry.py
+++ b/repos/system_upgrade/common/actors/addupgradebootentry/tests/unit_test_addupgradebootentry.py
@@ -273,7 +273,7 @@ def test_get_rdlvm_arg_values(monkeypatch):
 
     args = addupgradebootentry._get_rdlvm_arg_values()
 
-    assert tuple(args) == ('A', 'B')
+    assert args == ('A', 'B')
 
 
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/common/actors/addupgradebootentry/tests/unit_test_addupgradebootentry.py
```diff
@@ -273,7 +273,7 @@ def test_get_rdlvm_arg_values(monkeypatch):
 
     args = addupgradebootentry._get_rdlvm_arg_values()
 
-    assert args == ['A', 'B']
+    assert tuple(args) == ('A', 'B')
```
evgeni, line 276: `_get_rdlvm_arg_values` should be already returning a `tuple`, so shouldn't you be testing:
```suggestion
    assert args == ('A', 'B')
```

otherwise you're asserting that the return value of `_get_rdlvm_arg_values` can be converted to a tuple matching `(A, B)`

pirat89: good catch!

pirat89 (approved):

evgeni (approved):

PR #1321, opened 2024-12-18
Add actor for mysql migration RHEL9->RHEL10

SlouchyButton:
Basic actor that just checks that the mysql-server package is installed and inform user if so.

There are incompatibilities, so there is an article attached describing the process (the article is WIP)

jira: RHEL-62689

### DIFF
--- a/repos/system_upgrade/el7toel8/actors/firewalldfactsactor/actor.py
+++ b/repos/system_upgrade/el7toel8/actors/firewalldfactsactor/actor.py
@@ -1,5 +1,6 @@
 import os
 from xml.etree import ElementTree
+from xml.etree.ElementTree import ParseError
 
 from leapp.actors import Actor
 from leapp.libraries.actor import firewalldfactsactor
@@ -26,14 +27,14 @@ class FirewalldFactsActor(Actor):
             tree = ElementTree.parse('/etc/firewalld/lockdown-whitelist.xml')
             root = tree.getroot()
             facts.firewall_config_command = firewalldfactsactor.getLockdownFirewallConfigCommand(root)
-        except IOError:
+        except (ParseError, IOError):
             pass
 
         try:
             tree = ElementTree.parse('/etc/firewalld/direct.xml')
             root = tree.getroot()
             facts.ebtablesTablesInUse = firewalldfactsactor.getEbtablesTablesInUse(root)
-        except IOError:
+        except (ParseError, IOError):
             pass
 
         ipsetTypesInUse = set()
@@ -46,7 +47,7 @@ class FirewalldFactsActor(Actor):
                     tree = ElementTree.parse(os.path.join(directory, filename))
                     root = tree.getroot()
                     ipsetTypesInUse |= set(firewalldfactsactor.getIpsetTypesInUse(root))
-                except IOError:
+                except (ParseError, IOError):
                     pass
             facts.ipsetTypesInUse = list(ipsetTypesInUse)
         except OSError:
--- a/repos/system_upgrade/el7toel8/actors/firewalldfactsactor/tests/unit_test_firewalldfactsactor.py
+++ b/repos/system_upgrade/el7toel8/actors/firewalldfactsactor/tests/unit_test_firewalldfactsactor.py
@@ -1,6 +1,8 @@
 import xml.etree.ElementTree as ElementTree
+from xml.etree.ElementTree import ParseError
 
 from leapp.libraries.actor import firewalldfactsactor
+from leapp.models import FirewalldFacts
 
 
 def test_firewalldfactsactor_direct():
@@ -30,6 +32,13 @@ def test_firewalldfactsactor_direct():
         '''
     )
     assert set(firewalldfactsactor.getEbtablesTablesInUse(root)) == set(['broute', 'nat'])
+
+    # emulate a parse error
+    facts = FirewalldFacts()
+    try:
+        raise ParseError()
+    except ParseError:
+        assert not facts.ebtablesTablesInUse
 
 
 def test_firewalldfactsactor_firewallConfigCommand():


SlouchyButton:
Added configuration checks, so the user is informed that current MySQL configuration won't be compatible with newer MySQL present in RHEL 10.

When incompatible config is detected, severity is also raised to HIGH, as proceeding will result in `mysqld` failing to launch until issues with config are resolved.

pirat89:
@SlouchyButton  hi \o thanks for the contribution! regarding the severity, if I understand it right, the worst scenario here is that mysqldb will not be launched on the new system - will not be working - until they fix the configuration.

- High - Very likely to result in a deteriorated system state.
- Medium - Can impact both the system and applications.
- Low - Should not impact the system but can have an impact on applications.
- Info - Informational with no expected impact to the system or applications.

SlouchyButton:
Hi @pirat89 
Thank you, taking a look at this. I was originally planning on requesting a review today anyway.
Regarding the severity, what u describe makes sense to me. I'll lower the severity - keep the severity same for all cases detected by the actor.

I also have another question regarding tests. Are they required? There is a small sanity test similar to other our database components, which only tests the detection of the package on the system.

Testing other parts of the actor feels a bit problematic to me, as it is starting new commands with subprocess, or reading files (for configuration). Is it mandatory to _somehow_ test even this functionality, and if so what would be your proposed approach for this?

Thank you!

repos/system_upgrade/el9toel10/actors/mysqlcheck/libraries/mysqlcheck.py
```diff
@@ -0,0 +1,123 @@
+from leapp import reporting
+from leapp.libraries.common.rpms import has_package
+from leapp.libraries.stdlib import api
+from leapp.models import DistributionSignedRPM
+
+import subprocess
```
pirat89, line 6: the use of subprocess is forbidden in the project. use the run function instead:
```
from leapp.libraries.stdlib import api, CalledProcessError,  run
```

catch CalledProcessError and OSError exceptions when calling `run`.  Here is the definition of the function:
* https://github.com/oamg/leapp/blob/52cd433d58c50c972cc38e3010c0b8dedf2b88f4/leapp/libraries/stdlib/__init__.py#L149

You can check other actors which use the run function for a hint.

pirat89:
@SlouchyButton I made just a short overview the last time, we will try to do proper review the next week. Regarding the subprocess, use the `run` function instead. See my comment above. Regarding unit tests. Mocking `run` is usually simple (just set the function to return an output you would expect to obtain - e.g. from a file). Similar for the reading of files. Similar can be done for the `open` builtin function. We can help you with that if you want.

Of course, when some unit-tests are pretty tricky to write and their value would not be high, it is ok to mention it in a comment (in the test script file) for others and keep the testing on integration tests, etc. The full coverage is not required. Also, in such a case, you could still add files with example outputs under `tests/files/` directory so people can see some examples of outputs at least. And it could help them if they would like to create some tests based on these IO data.

I will cover such tips into the guidelines/tutorials we are writting in these days.

SlouchyButton:
>  Regarding the subprocess, use the `run` function instead. See my comment above.

Oh, I must have missed it in the docs, will change this.

> I will cover such tips into the guidelines/tutorials we are writting in these days.

That would be great, the documentation/guidelines are great, but they sometimes do feel like they miss some edge cases. Originally I was wanting to do multiple reports but eventually figured you can't do that from one actor and can't have more actors in one folder. I think this isn't explicitely mentioned so maybe some guidelines (best practices) how to report multiple things relating to one task (package upgrade) might help others.

> Update: but I would suggest anyway to postpone writing of another unit-tests until someone do a better review - so they will not need to be rewritten in case of bigger changes.

Right that's a good point.

pirat89:
> > Regarding the subprocess, use the `run` function instead. See my comment above.
> 
> Oh, I must have missed it in the docs, will change this.
> 
> > I will cover such tips into the guidelines/tutorials we are writing in these days.
> 
> That would be great, the documentation/guidelines are great, but they sometimes do feel like they miss some edge cases. Originally I was wanting to do multiple reports but eventually figured you can't do that from one actor and can't have more actors in one folder. I think this isn't explicitly mentioned so maybe some guidelines (best practices) how to report multiple things relating to one task (package upgrade) might help others.
> 
> > Update: but I would suggest anyway to postpone writing of another unit-tests until someone do a better review - so they will not need to be rewritten in case of bigger changes.
> 
> Right that's a good point.

well, you can create e.g. `mysql` directory and create several actors inside (each one in a separate dir). let's maybe rather sync the next week to tell us what everything you would like to do, so we can guide you what is the best way to do it :)

repos/system_upgrade/el9toel10/actors/mysqlcheck/libraries/mysqlcheck.py
```diff
@@ -0,0 +1,123 @@
...
+                'It seems that some config options currently used for MySQL server'
+                ' will be removed in updated MySQL server.\n'
+                'If you proceed with the update now, without addressing this issue,'
+                ' the MySQL server will fail to start until the config is fixed.\n'
+                'Detected deprecated options:\n'
+            )
+            for arg in self.found_options:
+                self.report_server_inst_summary += f"{arg} in MySQL config file\n"
+            for arg in self.found_arguments:
+                self.report_server_inst_summary += f"{arg} in SystemD service override\n"
```
pirat89, line 73: This will lead to "lot of" (depending on number of possible entries) text redundancy which is not accepted well by customers. I expect this will not generate usually too many lines, however, it's better to introduce the meaning of the list and then just simply list all the stuff. Something like:
```
Following configured MySQL options will are not valid anymore for MySQL on the target system:
    - option1
    - option2
    - ...
Following systemd...:
    - ...
    - ...
```

Also, define please the `FMT_LIST_SEPARATOR = '\n    - '` and use it for the listings so it's consistent between actors. We plan to document this and most likely create a function in the shared library - refactoring all actors to use it. But let's go with the simple formatting now.

repos/system_upgrade/el9toel10/actors/mysqlcheck/libraries/mysqlcheck.py
```diff
@@ -0,0 +1,123 @@
...
+        if self.found_arguments or self.found_options:
+            self.report_severity = reporting.Severity.HIGH
+            self.report_title = 'MySQL (mysql-server) seems to be using deprecated config options'
+            self.report_server_inst_summary += (
+                '\nWarning:\n'
+                'It seems that some config options currently used for MySQL server'
+                ' will be removed in updated MySQL server.\n'
+                'If you proceed with the update now, without addressing this issue,'
+                ' the MySQL server will fail to start until the config is fixed.\n'
+                'Detected deprecated options:\n'
```
pirat89, line 68: Using `\n` in the text is not acturally so good as the formatting could look weird in various webUIs. Also the Warning line is not good I would omit it. Basically the hole report is kind of the warning. Now you need to just provide information:
* What is the problem in genera?
* What does it mean for readers? What do you want them to understand? 
* What specifically we discovered that should concern the user? (the list of problematic entities)

In the remediation instructions:
* What should I do?
* Can resolve it before the upgrade?
* Do I have to do something after the upgrade?
* .....

Note that current summary could be confusing for the user as whatever they do, they will see this msg but remediation instructions says they have to deal with it after the upgrade (and the preupgrade part suggest only the backup of data - btw, do you mean actually dump the database?).

repos/system_upgrade/el9toel10/actors/mysqlcheck/libraries/mysqlcheck.py
```diff
@@ -0,0 +1,123 @@
...
+        # The syntax '--old' is deprecated and will be removed in a future release.
+        out = subprocess.run(['mysqld', '--validate-config', '--log-error-verbosity=2'],
+                             capture_output=True,
+                             check=False)
+
+        stderr = out.stderr.decode("utf-8")
+        if 'deprecated' in stderr:
+            self.found_options = {arg for arg
+                                  in self.REMOVED_ARGS
+                                  if arg in stderr}
```
pirat89, line 98: Use the `run` function from the leapp stdlib. Following code is untested, writting it just on my knees now (expecting the run function has been imported already):

 ```suggestion
        stderr = run(['mysqld', '--validate-config', '--log-error-verbosity=2'], split=False, checked=False)['stderr']
        if 'deprecated' in stderr:
            self.found_options = {
                arg for arg in self.REMOVED_ARGS if arg in stderr
            }
```

File: repos/system_upgrade/el9toel10/actors/mysqlcheck/actor.py [hunk 5]
```diff
@@ -0,0 +1,21 @@
...
+    """
+    Actor checking for presence of MySQL installation.
+
+    Provides user with information related to upgrading systems
+    with MySQL installed.
+    """
+    name = 'mysql_check'
+    consumes = (DistributionSignedRPM,)
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
```
pirat89, line 17: The actor is interacting with the system. It needs to be split in two actors:
* scanner scanner - that will collect all important data (facts). Based on the actor code, it seems you want a Model like this (I expect that most likely you will want to change a little bit):
```
class MySQLConfiguration(Model):
    """
    describe the content of the model
    """
    
    removed_options = fields.List(fields.String, default=[])
    """
    Configured options which are removed in RHEL 10 MySQL
    """
    
    systemd_services = fields.List(fields.String, default=[])
    """
    ...
    """
```
* the checker - that consumes particular leapp message(s) and evaluate the situation.

repos/system_upgrade/el9toel10/actors/mysqlcheck/actor.py
```diff
@@ -0,0 +1,21 @@
...
+from leapp.models import DistributionSignedRPM, Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class MySQLCheck(Actor):
+    """
+    Actor checking for presence of MySQL installation.
+
+    Provides user with information related to upgrading systems
+    with MySQL installed.
```
pirat89, line 12: Ideally, describe what the actor is really doing in details.

pirat89 (changes requested):
This will need several changes. Also, checking the current solution, I think that it should be pretty simple to cover the complete functionality by unit-tests. let's sync about it yet.

repos/system_upgrade/el9toel10/actors/mysqlcheck/libraries/mysqlcheck.py
pirat89: btw, `update` -> `upgrade` base on RHEL docs:
* update RPMs using `dnf update` - staying on same RHEL X major version
* upgrade -> upgrading to the next system major version

SlouchyButton: `\n` resolved (removed where not necessary).

Wording might still be improved on.

SlouchyButton:
Most of the changes should be hopefully done. Missing more thorough description of actors and tests.

SlouchyButton:
With this, all points from review should be hopefully resolved :crossed_fingers:, so I think this is ready for a second round of review.

PR #1323, opened 2025-01-08
Fix unreadable output in upgrade log

dkubek:
This commit resolves an issue where unwanted escape sequences (e.g., ANSI codes) appear in the output of certain commands like `dnf` during upgrades.

The issue seems to arise because, starting with version 242, `systemd-nspawn` introduced new pseudo-TTY capabilities (see the `Input/Output Options` section in `systemd-nspawn(1)`). As a result, commands run within container may include these escape sequences.

To address this, pseudo-TTY support is explicitly disabled in `systemd-nspawn` for upgrades on RHEL9 and later.

JIRA: [RHEL-69829](https://issues.redhat.com/browse/RHEL-69829)

### DIFF
--- a/repos/system_upgrade/common/libraries/mounting.py
+++ b/repos/system_upgrade/common/libraries/mounting.py
@@ -5,7 +5,7 @@ import shutil
 from collections import namedtuple
 
 from leapp.libraries.common.config import get_all_envs
-from leapp.libraries.common.config.version import get_source_major_version
+from leapp.libraries.common.config.version import get_source_major_version, matches_source_version
 from leapp.libraries.stdlib import api, CalledProcessError, run
 
 # Using ALWAYS_BIND will crash the upgrade process if the file does not exist.
@@ -88,6 +88,9 @@ class IsolationType(object):
                 # in such a case, just add line into the previous solution..
                 # TODO: the same about --capability=all
                 final_cmd += ['--keep-unit', '--capability=all']
+            if matches_source_version('>= 9.0'):
+                # Disable pseudo-TTY in container
+                final_cmd += ['--pipe']
             return final_cmd + ['-D', self.target] + binds + setenvs + cmd
 
     class CHROOT(_Implementation):


dkubek:
Also note that there is a warning in the man page for using ``--pipe``

```
Note that the pipe mode should be used carefully, as passing arbitrary file
descriptors to less trusted container payloads might open up unwanted interfaces for access by
the container payload. For example, if a passed file descriptor refers to a TTY of some form,
APIs such as TIOCSTI may be used to synthesize input that might be used for escaping the
container. Hence pipe mode should only be used if the payload is sufficiently trusted or when the
standard input/output/error output file descriptors are known safe, for example pipes.
```

Is this applicable to us?

dkubek:
Another consideration, this issue will arise on any system with ``nspawn`` version at least 242. Should we check instead the ``systemd-nspawn`` version when adding the argument instead of checking the source version? I am also not sure how feasible would that be at this level.

pirat89:
> Another consideration, this issue will arise on any system with `nspawn` version at least 242. Should we check instead the `systemd-nspawn` version when adding the argument instead of checking the source version? I am also not sure how feasible would that be at this level.

@dkubek great job! Checking the system version is ok. Systemd will not be rebased in RHEL 8 and we are sure that RHEL 9 will not contain any older systemd version. Staying with the system version check is ok from this POV. Let's keep it simple.

pirat89:
> Also note that there is a warning in the man page for using `--pipe`
> 
> ```
> Note that the pipe mode should be used carefully, as passing arbitrary file
> descriptors to less trusted container payloads might open up unwanted interfaces for access by
> the container payload. For example, if a passed file descriptor refers to a TTY of some form,
> APIs such as TIOCSTI may be used to synthesize input that might be used for escaping the
> container. Hence pipe mode should only be used if the payload is sufficiently trusted or when the
> standard input/output/error output file descriptors are known safe, for example pipes.
> ```
> 
> Is this applicable to us?

hmm.. not sure I understand it right. As far as I understand this, it should not affect us as we consider the container fully trusted as we created them and we know what we are executing. The question could be what if a user create custom actor calling custom / third party apps. But that is not different from the situation when we would like to operate without container at all and we use the container just to be able to affect the host system using tooling from the target OS. If someone is able to put malign content into the container, why they would need to escape from it when they could affect the system from inside already?

pirat89 (approved):
lgtm and works as expected \o/ (tested manually). good job!

PR #1326, opened 2025-01-10
redhatsignedrpmcheck: Add remediation hint and URL

pirat89:
Users do not have an idea what they should do about the third party packages - and quite often they do not understand the report itself. Adding the remediation hint with a link to related KB solution to help them to understand what they can do about this.

Note the actor needs significant changes to cover also other non-RHEL distributions. I decided to keep the change simple and resolve that in a follow up when we enable upgrades on other distributions as well.

jira: RHEL-44596

### DIFF
--- a/repos/system_upgrade/common/actors/redhatsignedrpmcheck/libraries/redhatsignedrpmcheck.py
+++ b/repos/system_upgrade/common/actors/redhatsignedrpmcheck/libraries/redhatsignedrpmcheck.py
@@ -3,8 +3,6 @@ from leapp.libraries.stdlib import api
 from leapp.libraries.stdlib.config import is_verbose
 from leapp.models import InstalledUnsignedRPM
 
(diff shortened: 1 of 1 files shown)


tomasfratrik:
The url link 'https://red.ht/customize-rhel-upgrade-actors' forwards me just to 'redhat.com', tried other 'red.ht' links and they forwarded me correctly.

pirat89:
@tomasfratrik  that's ok now, as the shortened URLs will be set/updated later. there are number of URLs now that do not even exist yet as we are waiting for RHEL 10 docs to be able to set them. this one is not such a case, but it will be set later when we go through the list of shortened URLs.

usually we parefer to have the links set before the merge, but with the new RHEL major version release, we are in situation where we cannot do it manytimes in advance and we will have to go anyway through all links in few months to set them. so it's ok

tomasfratrik:
Now I understand, makes sense. Otherways I tested it and it lgtm.

PR #1327, opened 2025-01-10
Generalize check of the third party RPMs

pirat89:
This is initial draft to generalize current check of installed third party RPMs that are not part of the installed system distribution.

Original actor has been written only for RHEL systems and expected that the only vendor as such is Red Hat. However, in case of other distributions it's not true. So such a report could be confusing for users and could set wrong expectations.

List of changes:
 * Rename and move redhatsignedrpmcheck actor to <NEW_NAME>
 * Update docstrings
 * Update the report, respecting name of the installed system
 * ...

Draft note: This is an initial draft I made when I crossed this code. Just sharing so someone else do not have to start from zero. It's still just a total draft however. 

jira: OAMG-12489

TODO:
- [ ] respecting name of the installed system in the report
- [ ] rename currently consumed leapp message - as the name is misleading - it contains more than just "unsigned" rpms
- [ ] update the title (keep it distro agnostic?)
- [ ] pick the name of the new actor - current one is a draft but i do not like so much.. maybe `distrorpmcheck`?
- [ ] update tests


pirat89:
Note this PR is not planned to be part of the upcoming release and I do not plan to even work on it in upcoming weeks.

matejmatuska:
This work is continued in #1402. Closing.

PR #1328, opened 2025-01-14
Verify supported target OS version in actors

tomasfratrik:
Originally when user specified the target system release using `--target` CLI option the verification has been performed immediately as only supported releases have been listed as possible choices for this option. The benefit of this solution was that users did not have to wait for all other checks to realize they execute leapp probably incorrectly. Unfortunately,
 * number of users do not understand why only some versions are supported
 * users upgrading with via various webUIs presenting only leapp reports
   could not see the error message available in terminal

To resolve this problem the checks are moved into actors so in case of specified unsupported target version the information is present in generated leapp reports.

Current behaviour is like this:
 * in case of invalid input (incorrect format of input data) the hard error is raised as before immediately. Malformed input data will not be processed anyhow by any actors
 * report error when the specified target major version is not direct successor of the current system version. I.e. specify 10.0 when upgrading from RHEL 8 (only RHEL 9 is acceptable).
   * this prevents number of cryptic errors as actors are not prepared for this situation
 * report standard inhibitor if the target release is not in the defined upgrade path, unless LEAPP_UNSUPPORTED=1
 * running leapp in unsupported (devel) mode skips the inhibitor and entire report

Additional changes:
* Update error message when format of target version is incorrect to clarify the expected version format

### NOTE
Note this solution introduces also `IPUPaths` msg which is by design marked already as deprecated. Read the commit msg for detail information.

jira: RHEL-51072

### DIFF
--- a/commands/command_utils.py
+++ b/commands/command_utils.py
@@ -28,7 +28,10 @@ def check_version(version):
     :return: release tuple
     """
     if not re.match(VERSION_REGEX, version):
-        raise CommandError('Unexpected format of target version: {}'.format(version))
+        raise CommandError(
+            "Unexpected format of target version: {}. "
+            "The required format is 'X.Y' (major and minor version).".format(version)
+        )
     return version.split('.')
 
 
@@ -126,7 +129,6 @@ def vet_upgrade_path(args):
     Make sure the user requested upgrade_path is a supported one.
     If LEAPP_DEVEL_TARGET_RELEASE is set then it's value is not vetted against upgrade_paths_map but used as is.
 
-    :raises: `CommandError` if the specified upgrade_path is not supported
     :return: `tuple` (target_release, flavor)
     """
     flavor = get_upgrade_flavour()
@@ -135,13 +137,6 @@ def vet_upgrade_path(args):
         check_version(env_version_override)
         return (env_version_override, flavor)
     target_release = args.target or get_target_version(flavor)
-    supported_target_versions = get_supported_target_versions(flavor)
-    if target_release not in supported_target_versions:
-        raise CommandError(
-                "Upgrade to {to} for {flavor} upgrade path is not supported, possible choices are {choices}".format(
-                    to=target_release,
-                    flavor=flavor,
-                    choices=','.join(supported_target_versions)))
     return (target_release, flavor)
 
 
--- a/repos/system_upgrade/common/actors/checktargetversion/tests/test_checktargetversion.py
+++ b/repos/system_upgrade/common/actors/checktargetversion/tests/test_checktargetversion.py
@@ -1,0 +1,90 @@
+import os
+
+import pytest
+
+from leapp import reporting
+from leapp.libraries.actor import checktargetversion
+from leapp.libraries.common.testutils import create_report_mocked, CurrentActorMocked, logger_mocked
+from leapp.libraries.stdlib import api
+from leapp.models import IPUPath, IPUPaths
+from leapp.utils.deprecation import suppress_deprecation
+from leapp.utils.report import is_inhibitor
+
+
+# It must be in a function so we can suppress the deprecation warning in tests.
+@suppress_deprecation(IPUPaths)
+def _get_upgrade_paths_data():
+    return IPUPaths(data=[
+        IPUPath(source_version='7.9', target_versions=['8.10']),
+        IPUPath(source_version='8.10', target_versions=['9.4', '9.5', '9.6']),
+        IPUPath(source_version='9.6', target_versions=['10.0']),
+        IPUPath(source_version='7', target_versions=['8.10']),
+        IPUPath(source_version='8', target_versions=['9.4', '9.5', '9.6']),
+        IPUPath(source_version='9', target_versions=['10.0'])
+    ])
+
+
+@pytest.fixture
+def setup_monkeypatch(monkeypatch):
+    """Fixture to set up common monkeypatches."""
+
+    def _setup(source_version, target_version, leapp_unsupported='0'):
+        curr_actor_mocked = CurrentActorMocked(
+            src_ver=source_version,
+            dst_ver=target_version,
+            envars={'LEAPP_UNSUPPORTED': leapp_unsupported},
+            msgs=[_get_upgrade_paths_data()]
+        )
+        monkeypatch.setattr(api, 'current_actor', curr_actor_mocked)
+        monkeypatch.setattr(api, 'current_logger', logger_mocked())
+        monkeypatch.setattr(reporting, 'create_report', create_report_mocked())
+    return _setup
+
+
+@pytest.mark.parametrize(('source_version', 'target_version', 'leapp_unsupported'), [
+    # LEAPP_UNSUPPORTED=0
+    ('7.9', '9.0', '0'),
+    ('8.10', '9.0', '0'),
+    ('9.6', '10.1', '0'),
+    ('7', '9.0', '0'),
+    ('8', '9.0', '0'),
+    ('9', '10.1', '0'),
+    # LEAPP_UNSUPPORTED=1
+    ('7.9', '9.0', '1'),
+    ('8.10', '9.0', '1'),
+    ('9.6', '10.1', '1'),
+    ('7', '9.0', '1'),
+    ('8', '9.0', '1'),
+    ('9', '10.1', '1'),
+])
+def test_unsuppoted_paths(setup_monkeypatch, source_version, target_version, leapp_unsupported):
+    setup_monkeypatch(source_version, target_version, leapp_unsupported)
+
+    if leapp_unsupported == '1':
+        checktargetversion.process()
+        assert reporting.create_report.called == 0
+        assert api.current_logger.warnmsg
+    else:
+        checktargetversion.process()
+        assert reporting.create_report.called == 1
+        assert is_inhibitor(reporting.create_report.report_fields)
+
+
+@pytest.mark.parametrize(('source_version', 'target_version'), [
+    ('7.9', '8.10'),
+    ('8.10', '9.4'),
+    ('8.10', '9.5'),
+    ('8.10', '9.6'),
+    ('9.6', '10.0'),
+    ('7', '8.10'),
+    ('8', '9.4'),
+    ('8', '9.5'),
+    ('8', '9.6'),
+    ('9', '10.0'),
+])
+def test_supported_paths(setup_monkeypatch, source_version, target_version):
+    setup_monkeypatch(source_version, target_version, leapp_unsupported='0')
+
+    checktargetversion.process()
+    assert reporting.create_report.called == 0
+    assert api.current_logger.infomsg
--- a/repos/system_upgrade/common/actors/ipuworkflowconfig/libraries/ipuworkflowconfig.py
+++ b/repos/system_upgrade/common/actors/ipuworkflowconfig/libraries/ipuworkflowconfig.py
@@ -64,17 +64,41 @@ def get_os_release(path):
                                       details={'details': str(e)})
 
 
+def check_target_major_version(curr_version, target_version):
+    required_major_version = int(curr_version.split('.')[0]) + 1
+    specified_major_version = int(target_version.split('.')[0])
+    if specified_major_version != required_major_version:
+        raise StopActorExecutionError(
+            message='Specified invalid major version of the target system',
+            details={
+                'Specified target major version': str(specified_major_version),
+                'Required target major version': str(required_major_version),
+                'hint': (
+                    'The in-place upgrade is possible only to the next system'
+                    ' major version: {ver}. Specify a valid version of the'
+                    ' target system when running leapp.'
+                    ' For more information about supported in-place upgrade paths'
+                    ' follow: https://access.redhat.com/articles/4263361'
+                    .format(ver=required_major_version)
+                )
+            }
+        )
+
+
 def produce_ipu_config(actor):
     flavour = os.environ.get('LEAPP_UPGRADE_PATH_FLAVOUR')
     target_version = os.environ.get('LEAPP_UPGRADE_PATH_TARGET_RELEASE')
     os_release = get_os_release('/etc/os-release')
+    source_version = os_release.version_id
+
+    check_target_major_version(source_version, target_version)
 
     actor.produce(IPUConfig(
         leapp_env_vars=get_env_vars(),
         os_release=os_release,
         architecture=platform.machine(),
         version=Version(
-            source=os_release.version_id,
+            source=source_version,
             target=target_version
         ),
         kernel=get_booted_kernel(),
--- a/commands/preupgrade/__init__.py
+++ b/commands/preupgrade/__init__.py
@@ -28,8 +28,7 @@ from leapp.utils.output import beautify_actor_exception, report_errors, report_i
              choices=['ga', 'e4s', 'eus', 'aus'],
              value_type=str.lower)  # This allows the choices to be case insensitive
 @command_opt('iso', help='Use provided target RHEL installation image to perform the in-place upgrade.')
-@command_opt('target', choices=command_utils.get_supported_target_versions(),
-             help='Specify RHEL version to upgrade to for {} detected upgrade flavour'.format(
+@command_opt('target', help='Specify RHEL version to upgrade to for {} detected upgrade flavour'.format(
                  command_utils.get_upgrade_flavour()))
 @command_opt('report-schema', help='Specify report schema version for leapp-report.json',
              choices=['1.0.0', '1.1.0', '1.2.0'], default=get_config().get('report', 'schema'))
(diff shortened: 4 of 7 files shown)


tomasfratrik:
This implementation could still be moved to its own actor if anyone wants that.
When upgrading from 8.10 with `LEAPP_UNSUPPORTED=1` leapp upgrade --target=9.2
No inhibitor was invoked, but after upgrade I was still in 9.6 version, so putting this to draft.

tomasfratrik:
I will add tests, after we finalize this.

tomasfratrik:
I will be moving this implementation to its own actor.

tomasfratrik:
I put his actor into Facts phase because when I upgraded for example with `--target 10.0`, the wouldn't be inhibitor, but 3 errors instead. Those 3:
```
Following errors occurred and the upgrade cannot continue:
    1. Actor: repository_mapping
       Message: The repository mapping file is invalid: the JSON does not match required schema (wrong field type/value): The value of "mapping" field is None, but this is not allowed
    2. Actor: repositories_blacklist
       Message: Actor didn't receive required messages: RepositoriesMapping
    3. Actor: pes_events_scanner
       Message: Cannot parse RepositoriesMapping data properly
```

When I added this before actor checkphase, now there are still those errors, but Inhibitor is triggered successfully.

tomasfratrik:
Latest push's fixed tests misspells

pirat89 (changes requested):
I tried to explain all wanted changes in the comments, with added suggestions for updated reports. Also, you are right about the misleading errors when incorrect major version is specified. It really bad from my pov as we are aware that this is happening from time to time to people and they could be very confused about that. Let's raise the `StopActorExecutionError` in such a case to protect sanity of readers. Add function like this into the ipuworkflowconfig:

```python
def check_target_major_version(curr_version, target_version):
    required_major_version = int(curr_version.split('.')[0]) + 1
    specified_major_version = int(target_version.split('.')[0])
    if specified_major_version != required_major_version:
        raise StopActorExecutionError(
            message='Specified invalid major version of the target system',
            details={
                'Specified targed major version': str(specified_major_version),
                'Required target major version': str(required_major_version),
                'hint': (
                    'The in-place upgrade is possible only to the next system'
                    ' major version: {ver}. Current solution depends heavily on this'
                    ' value and violation of this rule leads to various errors which'
                    ' could be very confusing.'
                    .format(ver=required_major_version)
                )
            }
        )
```

and call it before producing the `IPUConfig`:
```python
    check_target_major_version(os_release.version_id, target_version)
```

repos/system_upgrade/common/actors/verifytargetversion/libraries/verifytargetversion.py
```diff
@@ -0,0 +1,42 @@
...
+from leapp.cli.commands import command_utils
+from leapp.exceptions import StopActorExecution
+from leapp.libraries.stdlib import api
+
+FMT_LIST_SEPARATOR = '\n    - '
+
+
+def process():
+    flavour = os.environ.get('LEAPP_UPGRADE_PATH_FLAVOUR')
+    target_version = os.environ.get('LEAPP_UPGRADE_PATH_TARGET_RELEASE')
```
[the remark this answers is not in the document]
tomasfratrik: Done

repos/system_upgrade/common/actors/verifytargetversion/libraries/verifytargetversion.py
```diff
@@ -0,0 +1,42 @@
...
+
+    unsupported = target_version not in supported_target_versions and os.environ.get('LEAPP_UNSUPPORTED') != '1'
+
+    if unsupported:
+        hint = (
+                'Upgrade to a supported version\n'
+                'Supported target versions for your source version are: {sep}{ver}\n'
+                'Alternatively, if you want to upgrade to an unsupported version, '
+                'set the `LEAPP_UNSUPPORTED=1` environment variable.'
+                .format(sep=FMT_LIST_SEPARATOR, ver=FMT_LIST_SEPARATOR.join(supported_target_versions))
```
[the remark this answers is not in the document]
tomasfratrik: Done

repos/system_upgrade/common/actors/verifytargetversion/libraries/verifytargetversion.py
```diff
@@ -0,0 +1,42 @@
...
+        )
+
+        reporting.create_report([
+            reporting.Title('Target version is not supported'),
+            reporting.Summary(
+                'You are trying to upgrade to an unsupported target\n'
+                'Those are the targets without EUS (Extended Update Support)'
```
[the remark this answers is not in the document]
tomasfratrik: done, yes makes more sense to put the list of supported versions here

repos/system_upgrade/common/actors/verifytargetversion/libraries/verifytargetversion.py
```diff
@@ -0,0 +1,42 @@
...
+            ),
+            reporting.Groups([reporting.Groups.INHIBITOR]),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Remediation(hint=hint),
+            reporting.ExternalLink(
+                url='https://access.redhat.com/articles/4263361',
+                title='Learn more about supported in-place upgrade paths'
```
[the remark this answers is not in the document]
tomasfratrik: makes sense done

repos/system_upgrade/common/actors/verifytargetversion/libraries/verifytargetversion.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def process():
+    flavour = os.environ.get('LEAPP_UPGRADE_PATH_FLAVOUR')
+    target_version = os.environ.get('LEAPP_UPGRADE_PATH_TARGET_RELEASE')
+    supported_target_versions = command_utils.get_supported_target_versions(flavour)
+
+    unsupported = target_version not in supported_target_versions and os.environ.get('LEAPP_UNSUPPORTED') != '1'
+
+    if unsupported:
```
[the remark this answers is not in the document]
tomasfratrik: Yes completly missed that case... fixed

commands/command_utils.py
```diff
@@ -28,7 +28,7 @@ def check_version(version):
     :return: release tuple
     """
     if not re.match(VERSION_REGEX, version):
-        raise CommandError('Unexpected format of target version: {}'.format(version))
+        raise CommandError("Unexpected format of target version: {}. The correct format should be 'X.Y' (major and minor version).".format(version))
```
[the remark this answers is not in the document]
tomasfratrik: done

tomasfratrik:
In the `check_target_major_version(...)` function I don't really get this sentence 
```
Current solution depends heavily on this'
' value and violation of this rule leads to various errors which'
' could be very confusing.'
```

pirat89:
> ```python
> targed
> ```

Feel free to reword it. The msg is that this value is crucial for the functionality of actors and so we must stop it here.

repos/system_upgrade/common/actors/verifytargetversion/tests/test_verifytargetversion.py
```diff
@@ -0,0 +1,97 @@
...
+        "8.10": ["9.4", "9.5", "9.6"],
+        "9.6": ["10.0"],
+        "7": ["8.10"],
+        "8": ["9.4", "9.5", "9.6"],
+        "9": ["10.0"]
+}
+
+
+def _get_env_mocked(key, default=None):
+    return os.environ.get(key, default)
```
pirat89, line 23: do not work with envars in unit tests inside `os.environ`

repos/system_upgrade/common/actors/verifytargetversion/tests/test_verifytargetversion.py
```diff
@@ -0,0 +1,97 @@
...
+
+@pytest.fixture
+def setup_monkeypatch(monkeypatch):
+    """Fixture to set up common monkeypatches."""
+
+    def _setup(source_version, target_version, leapp_unsupported="0"):
+        monkeypatch.setattr(verifytargetversion, 'get_env', lambda key, default=None: _get_env_mocked(key, default))
+        monkeypatch.setattr(verifytargetversion, 'get_target_version', lambda: target_version)
+        monkeypatch.setenv('LEAPP_UPGRADE_PATH_FLAVOUR', 'default')
+        monkeypatch.setenv('LEAPP_UPGRADE_PATH_TARGET_RELEASE', target_version)
```
pirat89, line 38: no need to set this value based on the code

repos/system_upgrade/common/actors/verifytargetversion/actor.py
```diff
@@ -0,0 +1,21 @@
...
+    Verify if the target system version is supported by the upgrade process.
+
+    Invoke inhibitor if the target system is not supported.
+    Allow unsupported target if `LEAPP_UNSUPPORTED=1` is set.
+    """
+
+    name = 'verify_target_version'
+    consumes = ()
+    produces = (Report,)
+    tags = (FactsPhaseTag, IPUWorkflowTag)
```
pirat89, line 18: Check actors should be executed during `ChecksPhaseTag`

repos/system_upgrade/common/actors/verifytargetversion/libraries/verifytargetversion.py
```diff
@@ -0,0 +1,54 @@
...
+            ),
+            reporting.Groups([reporting.Groups.INHIBITOR]),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Remediation(hint=hint),
+            reporting.ExternalLink(
+                url='https://access.redhat.com/articles/4263361',
+                title='Supported in-place upgrade paths for Red Hat Enterprise Linux'
+            )
+        ])
+        raise StopActorExecution()
```
pirat89, line 53: seems my long comment about this has been dropped by GH :/ but this can be written in more simple way if initial conditions are negated.

I will push updated commit in ~15min with all changes

repos/system_upgrade/common/actors/ipuworkflowconfig/libraries/ipuworkflowconfig.py
```diff
@@ -64,17 +64,40 @@ def get_os_release(path):
...
+        raise StopActorExecutionError(
+            message='Specified invalid major version of the target system',
+            details={
+                'Specified targed major version': str(specified_major_version),
+                'Required target major version': str(required_major_version),
+                'hint': (
+                    'The in-place upgrade is possible only to the next system'
+                    ' major version: {ver}. Current solution depends heavily on this'
+                    ' value and violation of this rule leads to various errors which'
+                    ' could be very confusing.'
```
MichalHe, line 80: No need to give an explanation that reveals internals of leapp. Upgrading is possible only to the next major version, period.

repos/system_upgrade/common/actors/verifytargetversion/libraries/verifytargetversion.py
```diff
@@ -0,0 +1,54 @@
...
+FMT_LIST_SEPARATOR = '\n    - '
+
+
+def process():
+    flavour = get_env('LEAPP_UPGRADE_PATH_FLAVOUR')
+    target_version = get_target_version()
+    supported_target_versions = command_utils.get_supported_target_versions(flavour)
+
+    leapp_unsupported = get_env('LEAPP_UNSUPPORTED', '0') == '1'
+    unsupported = target_version not in supported_target_versions
```
MichalHe, line 17: Gets confusing having `leapp_unsupported` and `unsupported`, probably needs a better naming.
```suggestion
    is_target_version_unsupported = target_version not in supported_target_versions
```
or
```suggestion
    is_target_unsupported = target_version not in supported_target_versions
```

repos/system_upgrade/common/actors/verifytargetversion/libraries/verifytargetversion.py
```diff
@@ -0,0 +1,54 @@
...
+
+FMT_LIST_SEPARATOR = '\n    - '
+
+
+def process():
+    flavour = get_env('LEAPP_UPGRADE_PATH_FLAVOUR')
+    target_version = get_target_version()
+    supported_target_versions = command_utils.get_supported_target_versions(flavour)
+
+    leapp_unsupported = get_env('LEAPP_UNSUPPORTED', '0') == '1'
```
MichalHe, line 16: See my comment bellow
```suggestion
    is_leapp_unsupported_set = get_env('LEAPP_UNSUPPORTED', '0') == '1'
```

repos/system_upgrade/common/actors/verifytargetversion/libraries/verifytargetversion.py
```diff
@@ -0,0 +1,54 @@
+from leapp import reporting
+from leapp.cli.commands import command_utils
```
pirat89, line 2: actors should not import anything from command utils

repos/system_upgrade/common/actors/verifytargetversion/libraries/verifytargetversion.py
```diff
@@ -0,0 +1,54 @@
...
+
+def process():
+    flavour = get_env('LEAPP_UPGRADE_PATH_FLAVOUR')
+    target_version = get_target_version()
+    supported_target_versions = command_utils.get_supported_target_versions(flavour)
+
+    leapp_unsupported = get_env('LEAPP_UNSUPPORTED', '0') == '1'
+    unsupported = target_version not in supported_target_versions
+
+    if unsupported and leapp_unsupported:
```
MichalHe, line 19:
```suggestion
    if is_leapp_unsupported_set:
        log target if unsuppoted
        return
    if unsupported:
        emit inhibitor
        raise StopActorExecution()
```

repos/system_upgrade/common/actors/verifytargetversion/actor.py
```diff
@@ -0,0 +1,21 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import verifytargetversion
+from leapp.models import Report
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+
+
+class VerifyTargetVersion(Actor):
```
pirat89, line 7: let's rather use `CheckTargetVersion` instead of `Verify` to correspond with usual naming used for actors in these repositories.

repos/system_upgrade/common/actors/verifytargetversion/libraries/verifytargetversion.py
pirat89: covered

repos/system_upgrade/common/actors/ipuworkflowconfig/libraries/ipuworkflowconfig.py
pirat89: Actually good point. Dropping the explanation from the error msg. I will explain that in the comment.

repos/system_upgrade/common/actors/verifytargetversion/libraries/verifytargetversion.py
pirat89: refactored completely

repos/system_upgrade/common/actors/verifytargetversion/libraries/verifytargetversion.py
pirat89: refactored completely

repos/system_upgrade/common/actors/checktargetversion/tests/test_checktargetversion.py
```diff
@@ -0,0 +1,89 @@
...
+    setup_monkeypatch(source_version, target_version, leapp_unsupported)
+
+    if leapp_unsupported == '1':
+        checktargetversion.process()
+        assert reporting.create_report.called == 0
+        assert api.current_logger.warnmsg
+    else:
+        checktargetversion.process()
+        assert reporting.create_report.called == 1
+        assert 'inhibitor' in reporting.create_report.report_fields['groups']
```
MichalHe, line 69: IMHO, we should use:
```
from leapp.utils.report import is_inhibitor
```
Reaching into the internals of the report might cause problems if we change their internal structure in the future (hopefully not)

MichalHe (commented):
I have a single minor comment that needs to be addressed (now, or in a separate PR). Once I test this, I can will approve, since there are no blockers code-wise

repos/system_upgrade/common/actors/checktargetversion/tests/test_checktargetversion.py
pirat89: good catch! thanks! I am going to address it now

pirat89: done

pirat89 (approved):
lgtm, but as significant co-author I will wait for second green tick :)

MichalHe:
I have tested upgrading from 8.10 to various target versions with the following results:
| Target version | Result  |
| ---------------------| ---- |
| 10.7 | Inhibit :crossed_swords: |
| 9.7 | Inhibit :crossed_swords: |
| 9.5 | No inhibit :heavy_check_mark: |
|9.2 | Inhibit :crossed_swords: |

Using `LEAPP_UNSUPPORTE=1`, I can attempt upgrading to `9.2` without any inhibitors.

MichalHe (approved):
lgtm

PR #1330, opened 2025-01-15
Fix remediation message in the networkdeprecations actor

bengal:
The message should suggest to remove "dhclient", not "internal".

@lkundrak


pirat89:
@bengal  good catch! thanks.

pirat89 (approved):

PR #1332, opened 2025-01-21
CheckNetworkDeprecations9to10: Check for ifcfg artifacts

lkundrak:
Ifcfg support is going away in RHEL 10.

Let's add a few checks and suggestions so that those poor souls who happen particularly messed up configurations will get a hopefully somewhat helpful heads-up.

https://issues.redhat.com/browse/RHEL-58972


matejmatuska:
Hi @lkundrak, can you please rebase and fix the linter suggestions?

PR #1338, opened 2025-02-04
OpenSslEnginesCheck: New actor for OpenSSL engines (9->10) 

Jakuje:
The OpenSSL engines are deprecated in OpenSSL 3.0 (RHEL9) and in RHEL10 we removed existing pkcs11 engine and do not support building new engines. This check builds on top of existing 8to9 check OpenSslConfigCheck that I wrote years back that was doing some incompatibilities checks when updating from openssl 1  to 3.

Now, we really do not want users to upgrade with existing engines configured, and especially not with the ones that we removed from distribution (pkcs11), which might break the OpenSSL.

RHEL-78396


Jakuje:
I see there are some other common openssl actors already in `repos/system_upgrade/common/actors/openssl/` so I will probably move the `OpenSslConfigScanner` in there in the next iteration.

repos/system_upgrade/el9toel10/actors/opensslenginescheck/libraries/opensslenginescheck.py
```diff
@@ -0,0 +1,154 @@
...
+def check_openssl_engines(config):
+    """
+    Check there are no engines configured in openssl.cnf
+
+    In case the of the pkcs11 engine, stop the upgrade as the package was removed from RHEL 10.
+
+    In case of other (usually third-party) engines, just warn about them being deprecated,
+    and suggest removal.
+    """
+    init_block = _openssl_find_block(config, "openssl_init")
```
neverpanic, line 73: The name `openssl_init` can actually differ, too; what will not differ is that `openssl_conf` in the default (i.e., first) section will always contain the name of the init section.

repos/system_upgrade/el9toel10/actors/opensslenginescheck/libraries/opensslenginescheck.py
```diff
@@ -0,0 +1,154 @@
...
+            continue
+
+        enabled_engines.append(name)
+
+    if "pkcs11" in enabled_engines:
+        reporting.create_report([
+            reporting.Title('There is pkcs11 engine configured in openssl.cnf'),
+            reporting.Summary(
+                'The OpenSSL configuration file `/etc/pki/tls/openssl.cnf` contains the '
+                'initialization of pkcs11 engin. The pkcs11 engine was removed from RHEL 10 '
```
neverpanic, line 125: type: "engine" instead of "engin".

repos/system_upgrade/el9toel10/actors/opensslenginescheck/libraries/opensslenginescheck.py
```diff
@@ -0,0 +1,154 @@
...
+    engines_block = _openssl_find_block(config, engines_pair.value)
+    if not engines_block:
+        # No engines no problem
+        return
+
+    enabled_engines = []
+    # Iterate over engines directives -- they point to another block
+    for engine in engines_block.pairs:
+        name = engine.key
+        engine_block = _openssl_find_block(config, engine.value)
```
neverpanic, line 110: Now, technically the name of the engine configuration block is arbitrary, too.
For example, this is a valid configuration that load the pkcs11 engine:

```conf
openssl_conf = openssl_init

[openssl_init]
engines = engines_sect

[engines_sect]
cxpf11 = cxpf11_sect

[cxpf11_sect]
engine_id = pkcs11
dynamic_path = /usr/lib64/engines-3/libpkcs11.so
MODULE_PATH = opensc-pkcs11.so
init = 1
```

I'm not sure whether we actually want to go all the way and detect this, too. It feels like over-complicating things.

repos/system_upgrade/el9toel10/actors/opensslenginescheck/libraries/opensslenginescheck.py
Jakuje: Good catch! We have the right value in `config.openssl_conf` so I will use that one.

repos/system_upgrade/el9toel10/actors/opensslenginescheck/libraries/opensslenginescheck.py
Jakuje: Fixing. Thanks!

repos/system_upgrade/el9toel10/actors/opensslenginescheck/libraries/opensslenginescheck.py
Jakuje: My reading of the manual page was that this is just an "alternative name".  

>  engine_id
> This is used to specify an alternate name, overriding the default name specified in the list of engines. If present, it must be first.  For example:

You are right it overrides the original id, but the search for the pkcs11 is mostly additional check -- we really care if there are any engines or not and if they are we are going to say the user to fix their config.

I would probably like to avoid going to the engine_id, but it should not be that hard to add.

neverpanic (approved):

Jakuje:
Ugh. Looking at the other actors, there is #1131 which basically always backs up the modified openssl conf and uses the rpm one on the target system, which means that any changes to the OpenSSL configuration files we made with 8->9 upgrade nor this one nor the one proposed from CRYPTO-15581 would have any effect. So should we just kill these actors and let the actor in handle it all?

https://github.com/oamg/leapp-repository/blob/main/repos/system_upgrade/common/actors/openssl/migrateopensslconf/actor.py

@pirat89 ?

Jakuje:
Sounds like this actor checking the configured engines should be still usable. Even though we will get the new configuration automatically, warning the user if they have the engines hardcoded is still useful information.

repos/system_upgrade/el9toel10/actors/opensslenginescheck/libraries/opensslenginescheck.py
```diff
@@ -0,0 +1,154 @@
...
+            ),
+            reporting.Groups([reporting.Groups.INHIBITOR]),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Groups([
+                    reporting.Groups.SECURITY,
+                    reporting.Groups.NETWORK,
+                    reporting.Groups.SERVICES
+            ]),
+        ] + resources)
+        return
```
beldmit, line 94: `openssl_conf` indicates the name of the specific section. It can be anything like `openssl_conf=blahblah`.

Ideally we should look for the `engines` section name from the section specified in `openssl_conf=some_section`.

Jakuje: That is exactly what is done here. The `config.openssl_conf` has the name of the first first key=value pair.

beldmit: Oh. Sorry. LGTM then

pirat89 (changes requested):
discussed with @Jakuje , we need better clarification of the impact and remediation instruction regarding pkcs11-engine. Currently we do not know how to properly instruct users about expected impact and actions that should be done.

> so should we just kill these actors and let the actor in handle it all?

@neverpanic it makes sense to me still to report this msg, just not sure about <paragraph-above>. maybe it would make a sense to merge few reports together in future but let's keep it split for this release as time is running.

pirat89 (approved):
lgtm! note that IPU 9 -> 10 is not covered yet by CI. I will do some manual tests after the merge together with other related changes before the release. Usually I would do ti before the merge, but I want to test it with other related stuff at once and it seems that everything is ok regarding the code and component tests. So merging now.

Thank you for the contribution!

PR #1340, opened 2025-02-07
[WIP] Update CA bundle path in krb5 config (el9to10)

jrisc:
jira: RHEL-65265

### DIFF
--- a/repos/system_upgrade/el9toel10/actors/krb5conf/convertkrb5conf/actor.py
+++ b/repos/system_upgrade/el9toel10/actors/krb5conf/convertkrb5conf/actor.py
@@ -1,0 +1,18 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import convertkrb5conf
+from leapp.models import OutdatedKrb5confLocation
+from leapp.tags import IPUWorkflowTag, PreparationPhaseTag
+
+
+class ConvertKrb5conf(Actor):
+    """
+    Update krb5 configuration file
+    """
+
+    name = 'convert_krb5_conf'
+    consumes = (OutdatedKrb5confLocation,)
+    produces = ()
+    tags = (PreparationPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        convertkrb5conf.process()
--- a/repos/system_upgrade/el9toel10/actors/krb5conf/convertkrb5conf/libraries/convertpamuserdb.py
+++ b/repos/system_upgrade/el9toel10/actors/krb5conf/convertkrb5conf/libraries/convertpamuserdb.py
@@ -1,0 +1,19 @@
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.stdlib import api
+from leapp.models import OutdatedKrb5confLocation
+
+def _convert_krb5conf(conf_file):
+    with open(conf_file) as f:
+        text = f.read().replace('/etc/ssl/certs/ca-certificates.crt',
+                                '/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem')
+    with open(conf_file, 'w') as f:
+        f.write(text)
+
+def process():
+    msg = next(api.consume(OutdatedKrb5confLocation), None)
+    if not msg:
+        raise StopActorExecutionError('Expected OutdatedKrb5confLocation, but got None')
+
+    if msg.locations:
+        for location in msg.locations:
+            _convert_krb5conf(location)
--- a/repos/system_upgrade/el9toel10/actors/krb5conf/checkkrb5conf/actor.py
+++ b/repos/system_upgrade/el9toel10/actors/krb5conf/checkkrb5conf/actor.py
@@ -1,0 +1,18 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import checkkrb5conf
+from leapp.models import OutdatedKrb5confLocation, Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckKrb5conf(Actor):
+    """
+    Create report with the location of oudated krb5 configuration files
+    """
+
+    name = 'check_krb5_conf'
+    consumes = (OutdatedKrb5confLocation,)
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        checkkrb5conf.process()
--- a/repos/system_upgrade/el9toel10/actors/krb5conf/checkkrb5conf/libraries/checkkrb5conf.py
+++ b/repos/system_upgrade/el9toel10/actors/krb5conf/checkkrb5conf/libraries/checkkrb5conf.py
@@ -1,0 +1,28 @@
+from leapp import reporting
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.stdlib import api
+from leapp.models import OutdatedKrb5confLocation
+
+FMT_LIST_SEPARATOR = "\n    - "
+
+
+def process():
+    msg = next(api.consume(OutdatedKrb5confLocation), None)
+    if not msg:
+        raise StopActorExecutionError('Expected OutdatedKrb5confLocation, but got None')
+
+    if msg.locations:
+        reporting.create_report([
+            reporting.Title('MIT krb5 configuration file(s) will be updated to point the new X.509 CA bundle file'),
+            reporting.Summary(
+                'On RHEL 10, the location of the reference X.509 CA bundle '
+                'file was modified. The following MIT krb5 configuration files '
+                'have to be updated to point to the new bundle file:'
+                '{sep}{locations}'.format(sep=FMT_LIST_SEPARATOR, locations=FMT_LIST_SEPARATOR.join(msg.locations))),
+            reporting.Severity(reporting.Severity.INFO),
+            reporting.Groups([reporting.Groups.SECURITY, reporting.Groups.AUTHENTICATION])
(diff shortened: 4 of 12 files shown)


pirat89:
@jrisc I am aware that for first contributors tests are not performed automatically and it could you to wait for our approval always. You can execute tests locally in container using Makefile from the main dir:
```shell
TEST_CONTAINER=rhel9 make test_container
```

see `make help` for more prepared stuff. Note that I realized that `rhel9` container is not documented in the help page, but it's working otherwise.

repos/system_upgrade/el9toel10/actors/krb5conf/convertkrb5conf/actor.py
```diff
@@ -0,0 +1,18 @@
...
+
+class ConvertKrb5conf(Actor):
+    """
+    Update krb5 configuration file
+    """
+
+    name = 'convert_krb5_conf'
+    consumes = (OutdatedKrb5confLocation,)
+    produces = ()
+    tags = (PreparationPhaseTag, IPUWorkflowTag)
```
pirat89, line 15: @jrisc  This is weird. I would expect this actor to be performed during the application phase (`ApplicationsPhaseTag`). Is there any specific reason why this needs to be performed before the RPM upgrade transaction is performed? Additional changes in this PR can be connected based on the outcome of this discussion

pirat89 (commented):
The code looks good in general. I found just one thing which is weird to me and based on the outcome of the discussion there could be another things to change or not.

repos/system_upgrade/el9toel10/actors/krb5conf/convertkrb5conf/actor.py
jrisc: Most of the code in this PR is taken from the RHEL10 `pamuserdb` actors. This tag is used by `convertpamuserdb`:
https://github.com/oamg/leapp-repository/blob/main/repos/system_upgrade/el9toel10/actors/pamuserdb/convertpamuserdb/actor.py#L15

But I don't know what these tags are referring to. So if you think `ApplicationsPhaseTag`, I will update the PR accordingly.

pirat89: Well, [PreparationPhase](https://github.com/oamg/leapp-repository/blob/065dcb4c904d471ec3c2378bc03240ac771f74ce/repos/system_upgrade/common/workflows/inplace_upgrade.py#L160) is used for things that must happen before the rpm transaction - to ensure the "smooth" execution. If this has no impact on the rpm transaction, then the [applicationphase](https://github.com/oamg/leapp-repository/blob/065dcb4c904d471ec3c2378bc03240ac771f74ce/repos/system_upgrade/common/workflows/inplace_upgrade.py#L177) is the right one. in such a case there are more things that needs to be changed. let's sync tomorrow.

jrisc: RPM update and major version upgrade should not affect the krb5 configuration in `/etc/krb5.conf.d/`, . On the other hand, `/etc/krb5.conf` is never overwritten by RPM update (`%config(noreplace)` in RPM spec), but I am not sure if this is the case for RHEL major version upgrade too.

Anyway, if `/etc/krb5.conf` is overwritten on upgrade, there will be no `pkinit_anchors` parameter left to modify, and if it is not overwritten, then probably it's for the best to not modify it before the RPM transaction is successful. Because if the RPM transaction fails, we could end up with a reference to `/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem` in `/etc/krb5.conf`, while this file is not installed yet.

So yes, probably it is better to use `ApplicationsPhaseTag`. What else should be changed in this case?

pirat89: basically after the RPM transaction phase, starting with `ApplicationsPhaseTag`, actors must not crash or stop with an error - like `StopActorExecutionError` - it's better to have partially broken system than much more broken system, so the logic is to run over dead bodies. So the way is just log error and end silently - so customers better chance to still boot the system and fix it manually.

repos/system_upgrade/el9toel10/actors/krb5conf/convertkrb5conf/libraries/convertpamuserdb.py
```diff
@@ -0,0 +1,21 @@
...
+        text = f.read().replace('/etc/ssl/certs/ca-certificates.crt',
+                                '/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem')
+    with open(conf_file, 'w') as f:
+        f.write(text)
+
+
+def process():
+    msg = next(api.consume(OutdatedKrb5confLocation), None)
+    if not msg:
+        raise StopActorExecutionError('Expected OutdatedKrb5confLocation, but got None')
```
pirat89, line 17: Prevent stopping the upgrade process in this phase. Just log the error and end the actor.
```suggestion
    if not msg:
        api.current_logger().error(
            'Expected OutdatedKrb5confLocation, but got None. '
            'Cannot apply possibly needed changes in kerberos configuration files.'
        )
        return
```

repos/system_upgrade/el9toel10/actors/krb5conf/convertkrb5conf/libraries/convertpamuserdb.py
```diff
@@ -0,0 +1,21 @@
...
+from leapp.libraries.stdlib import api
+from leapp.models import OutdatedKrb5confLocation
+
+
+def _convert_krb5conf(conf_file):
+    with open(conf_file) as f:
+        text = f.read().replace('/etc/ssl/certs/ca-certificates.crt',
+                                '/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem')
+    with open(conf_file, 'w') as f:
+        f.write(text)
```
pirat89, line 11: if we are sure that all files wil be always present and `rw` operation will allowed, it's ok. otherwise we should cover it by try-except:
```
try:
     ...
except IOError as e:
    # log error
    return  # give it a chance to fix other files still
```

jrisc: Since the file list is supposed to be collected just before by the `scankrb5conf` actor, I think we can expect the files to still be there at this point.

I just realize that I forgot to rename this file though... :sweat_smile:

repos/system_upgrade/el9toel10/actors/krb5conf/convertkrb5conf/actor.py
jrisc: I replaced `PreparationPhaseTag` by `ApplicationsPhaseTag` in 724f6bef4a9e5343c9caceecf5491b87c4ab7a4b.

repos/system_upgrade/el9toel10/actors/krb5conf/convertkrb5conf/libraries/convertpamuserdb.py
jrisc: Is Leapp running as root? If that's the case, I don't expect it to face any write permissions problems. However, if one of the krb5 config file in `/etc/krb5.conf.d/` is not publicly readable (which makes no sense, but is technically possible), it will probably cause the `scankrb5conf` actor to fail. Do you think I should just log a warning and skip the erroneous file in this case?

pirat89: you are right. yes, it is running as root always and as we could read it before, we should be able to read it later as well - DAC is ignored by root, selinux will be always in disabled or permissive mode, and I do not expect someone would set an ACL on these files (like making them immutable). So only thing here would be if a file is removed or renamed during the RPM transaction.

repos/system_upgrade/el9toel10/actors/krb5conf/convertkrb5conf/libraries/convertkrb5conf.py
```diff
@@ -0,0 +1,25 @@
+from leapp.exceptions import StopActorExecutionError
```
pirat89, line 1: forgotten unused import

jrisc:
@pirat89 It seems that failing tests are cause by the absence of RPM database. Is there a way to handle this case, or it just cannot be tested?

MichalHe (changes requested):
The solution looks code-wise very good, thank you. I left some minor comments intended to ease with maintenance of the code.

repos/system_upgrade/el9toel10/actors/krb5conf/convertkrb5conf/libraries/convertkrb5conf.py
```diff
@@ -0,0 +1,37 @@
+from leapp.libraries.stdlib import api, CalledProcessError, run
+from leapp.models import OutdatedKrb5conf
+
+
+def _backup_krb5conf(conf_file):
+    try:
+        run(['/usr/bin/cp', conf_file, conf_file + '.leappsave'])
```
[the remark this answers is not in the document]
MichalHe: Discussed with @pirat89 that we are doing similar backups on other places, so you can ignore this comment. Sorry, my mistake.

jrisc:
@MichalHe Thank you for completing the PR. I've just squashed the commits and force-pushed them.

MichalHe (approved):
LGTM

PR #1343, opened 2025-02-11
el9to10: actors: mysql

SlouchyButton:
RHEL 9 and RHEL 10 have different MySQL versions and user should be
informed about differences and required steps to migrate their data
and configuration safely. In case of MySQL configuration, number of
deprecated options have been removed in MySQL on RHEL 10 which would
prevent run of the mysql server if present. In such a case users
need to drop all deprecated options. Also the mysqld service could
be executed with deprecated arguments which are no longer valid
on RHEL 10, so possible customized service should be detected also.

jira: RHEL-62689


pirat89 (approved):

pirat89:
merge time \o/ thanks.

PR #1344, opened 2025-02-12
Drop 9.5 and add 9to10 tests

matejmatuska:


repos/system_upgrade/common/libraries/config/version.py
```diff
@@ -19,7 +19,7 @@
     # Note: 'rhel-alt' is detected when on 'rhel' with kernel 4.x
     '7': {'rhel': ['7.9'], 'rhel-alt': [], 'rhel-saphana': ['7.9']},
     '8': {'rhel': ['8.10'], 'rhel-saphana': ['8.10']},
-    '9': {'rhel': ['9.4', '9.5', '9.6'], 'rhel-saphana': ['9.4', '9.6']},
+    '9': {'rhel': ['9.4', '9.6'], 'rhel-saphana': ['9.4', '9.6']},
```
pirat89, line 22:
```suggestion
    '9': {'rhel': ['9.6'], 'rhel-saphana': ['9.6']},
```

@matejmatuska  upgrade from 9.4 will not be allowed. we had it here for the testing purposes only. this covers source systems from which it is possible to upgrade

pirat89 (approved):
lgtm. keeping the merge after tests are finished.

PR #1346, opened 2025-02-13
checksaphana: Update link name to upgrade guide

matejmatuska:
The actual link is updated by changing the shortened URL target.

Also clarified one report a little, it might not be obvious that >=9.2 also covers >= 10.0.

### DIFF
--- a/repos/system_upgrade/common/actors/checksaphana/libraries/checksaphana.py
+++ b/repos/system_upgrade/common/actors/checksaphana/libraries/checksaphana.py
@@ -24,7 +24,7 @@ SAP_HANA_RHEL90_MINIMAL_VERSION_STRING = 'HANA 2.0 SPS05 rev 59.04 or later, or 
 
 def _report_skip_check():
     summary = (
-        'For the target RHEL releases >=8.8 and >=9.2 '
+        'For the target RHEL releases >=8.8, >=9.2 and >=10.0 '
         'the leapp utility does not check RHEL and SAP HANA 2.0 '
         'versions compatibility. Please ensure your SAP HANA 2.0 '
         'is supported on the target RHEL release and '
@@ -187,8 +187,6 @@ def version2_check(info):
             continue
         if version.matches_target_version('> 8.6', '< 9.0') or version.matches_target_version('> 9.0'):
             # if a target release is >=8.8 or >=9.2, the SAP HANA and RHEL versions compatibility is not checked
-            # TODO(mmatuska): We don't know whether the check will be skipped on RHEL 10 (9to10) yet,
-            # update the method accoridingly then
             _report_skip_check()
             return
         # if a starget release is 8.6 or 9.0 we still check SAP HANA and RHEL versions compatibility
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/common/actors/checksaphana/libraries/checksaphana.py
```diff
@@ -247,10 +245,10 @@ def platform_check():
             title='How to in-place upgrade SAP environments from RHEL 7 to RHEL 8'),
```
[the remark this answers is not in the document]
matejmatuska: Yes, why not? We are not dropping anything for 7->8 at the moment.

repos/system_upgrade/common/actors/checksaphana/libraries/checksaphana.py
```diff
@@ -247,10 +245,10 @@ def platform_check():
...
         '9': reporting.ExternalLink(
             url='https://red.ht/how-to-in-place-upgrade-sap-environments-from-rhel-8-to-rhel-9',
-            title='How to in-place upgrade SAP environments from RHEL 8 to RHEL 9'),
+            title='Upgrading SAP environments from RHEL 8 to RHEL 9'),
         '10': reporting.ExternalLink(
             url='https://red.ht/how-to-in-place-upgrade-sap-environments-from-rhel-9-to-rhel-10',
```
alexxa, line 250: the title change presumes the link change too. Thefore we have 404 on it now. 
plus it seems that shortcut was applicable to the previous doc portal, but not new one. 

The correct link for 8to9 is https://docs.redhat.com/en/documentation/red_hat_enterprise_linux_for_sap_solutions/9/html/upgrading_sap_environments_from_rhel_8_to_rhel_9/index

and similarly for 7to8 https://docs.redhat.com/en/documentation/red_hat_enterprise_linux_for_sap_solutions/8/html/upgrading_sap_environments_from_rhel_7_to_rhel_8/index

alexxa: and the same would be for 9to10, ie. https://docs.redhat.com/en/documentation/red_hat_enterprise_linux_for_sap_solutions/10/html/upgrading_sap_environments_from_rhel_9_to_rhel_10/index

matejmatuska: The links are changed outside of the code, we just change the redirects. This is so that we can change the redirects anytime and don't have to wait for a release.

pirat89: @alexxa yeah, about changed target URLs we know already. We will be fixing them during the next week together with other shortened URLs. But it seems that title should be updated also for 7 -> 8

matejmatuska: Yes I will update 7->8

pirat89: and shortened URLs are expected to stay same. we will update only targets where they should point to.

pirat89 (approved):
lgtm. thanks!!

pirat89:
The failed test is unrelated to the change. It's incorrectly configured el9 machine. merging.

PR #1347, opened 2025-02-13
networkdeprecations: Update summary to include list of files (#1332 follow-up)

matejmatuska:
The legacy files are currently only listed via
`RelatedResource('file')`, however this type of `RelatedResource` is currently not shown in leapp-report.txt.

This patch adds a list of the files to the report summary.

This is a follow-up to #1332.


pirat89 (approved):
lgtm

PR #1349, opened 2025-02-14
data: update leapp data files for final build

pirat89:
Updated PES and DDDD data files with numerous changes related for IPU 9 -> 10.


matejmatuska (approved):
Looks OK. Waiting for tests.

PR #1350, opened 2025-02-27
Livemode: various improvements to make the feature more accessible to users

MichalHe:
## Livemode improvements
This is a first part of the work necessary to make the  `livemode` feature more accessible, hoping to make it the default -in the future. There are multiple quality-of-live implemented as part of this PR (kept as separate commits):
- CLI: introduce `--enable-experimental-feature` that allows users to simply enable entire group of actors
- exclude DNF cache from squashfs to reduce the size of the created image. The DNF cache that contains packages to be installed during the upgrade is always read from `/sysroot/var/lib/leapp`, and therefore, it would not be read from squashfs'ed `/var/lib/leapp`
- configuration: use configuration capabilities provided by the framework instead of ad-hoc INI file
- bugfix: do not crash with an unhandled exception when trying to enable a systemd service inside the squashfs image that already appears to be enabled

## Why are we introducing a new command line option `--enable-experimental-feature`?
1) `--whitelist-experimental` is a quite bad user experience. Why should a user know actor names? The user is more interested in functionality, not in leapp-repository's internals.
2) Some features like `livemode` use a bunch of actors to facilitate the feature, therefore, the user would need to list all of these actors explicitly and he/she better not forget any of them.

Why don't we just use a single experimental actor that produces a message which will "enable" (if the message is received) all of the remaining actors? The problem is that anyone can produce such a message. We could write the remaining experimental actors such that they check for a combination of receiving a special message + checking that `LEAPP_EXPERIMENTAL` is set, but this solution is cumbersome and far from perfect (user can enable a different experimental actor, and, unintentionally enable livemode if some actor is producing the special message). Finally, there is no way for actors to check whether a specific experimental actor has been enabled, if I am not wrong -- `LEAPP_EXPERIMENTAL` is binary information (yes/no). 

## Example of livemode configuration
```
livemode:
  additional_packages : [ vim ]
  autostart_upgrade_after_reboot : false
  setup_network_manager : true
  setup_passwordless_root : true
  setup_opensshd_using_auth_keys : /root/.ssh/authorized_keys
```

jira-ref: RHELMISC-10648

### DIFF
--- a/commands/preupgrade/__init__.py
+++ b/commands/preupgrade/__init__.py
@@ -14,6 +14,10 @@ from leapp.utils.output import beautify_actor_exception, report_errors, report_i
 
 @command('preupgrade', help='Generate preupgrade report')
 @command_opt('whitelist-experimental', action='append', metavar='ActorName', help='Enables experimental actors')
+@command_opt('enable-experimental-feature', action='append', metavar='Feature',
+             help=('Enable experimental feature. '
+                   'Available experimental features: {}').format(util.get_help_str_with_avail_experimental_features()),
+             choices=list(util.EXPERIMENTAL_FEATURES), default=[])
 @command_opt('debug', is_flag=True, help='Enable debug mode', inherit=False)
 @command_opt('verbose', is_flag=True, help='Enable verbose logging', inherit=False)
 @command_opt('no-rhsm', is_flag=True, help='Use only custom repositories and skip actions'
--- a/commands/upgrade/util.py
+++ b/commands/upgrade/util.py
@@ -16,7 +16,6 @@ from leapp.utils.audit import get_checkpoints, get_connection, get_messages
 from leapp.utils.output import report_unsupported
 from leapp.utils.report import fetch_upgrade_report_messages, generate_report_file
 
-
 EXPERIMENTAL_FEATURES = {
     'livemode': [
         'live_image_generator',
--- a/commands/upgrade/util.py
+++ b/commands/upgrade/util.py
@@ -17,6 +17,25 @@ from leapp.utils.output import report_unsupported
 from leapp.utils.report import fetch_upgrade_report_messages, generate_report_file
 
 
+EXPERIMENTAL_FEATURES = {
+    'livemode': [
+        'live_image_generator',
+        'live_mode_config_scanner',
+        'live_mode_reporter',
+        'prepare_live_image',
+        'emit_livemode_requirements',
+        'remove_live_image',
+    ]
+}
+""" Maps experimental features to a set of experimental actors that need to be enabled. """
+
+
+def get_help_str_with_avail_experimental_features():
+    if EXPERIMENTAL_FEATURES:
+        return ', '.join(EXPERIMENTAL_FEATURES)
+    return 'There are no experimental features available'
+
+
 def disable_database_sync():
     def disable_db_sync_decorator(f):
         @functools.wraps(f)
@@ -184,11 +203,25 @@ def handle_output_level(args):
 # the latest supported release because of target_version discovery attempt.
 def prepare_configuration(args):
     """Returns a configuration dict object while setting a few env vars as a side-effect"""
+
     if args.whitelist_experimental:
         args.whitelist_experimental = list(itertools.chain(*[i.split(',') for i in args.whitelist_experimental]))
         os.environ['LEAPP_EXPERIMENTAL'] = '1'
     else:
         os.environ['LEAPP_EXPERIMENTAL'] = '0'
+        args.whitelist_experimental = []
+
+    for experimental_feature in set(args.enable_experimental_feature):
+        # It might happen that there are no experimental features, which would allow user
+        # to pass us any string as an experimental feature.
+        if experimental_feature not in EXPERIMENTAL_FEATURES:
+            continue
+
+        actors_needed_for_feature = EXPERIMENTAL_FEATURES[experimental_feature]
+        args.whitelist_experimental.extend(actors_needed_for_feature)
+    if args.enable_experimental_feature:
+        os.environ['LEAPP_EXPERIMENTAL'] = '1'
+
     os.environ['LEAPP_UNSUPPORTED'] = '0' if os.getenv('LEAPP_UNSUPPORTED', '0') == '0' else '1'
     if args.no_rhsm:
         os.environ['LEAPP_NO_RHSM'] = '1'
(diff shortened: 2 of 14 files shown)


MichalHe:
## Why are we introducing a new command line option `--enable-experimental-feature`?
1) `--whitelist-experimental` is a quite bad user experience. Why should a user know actor names? The user is more interested in functionality, not in leapp-repository's internals.
2) Some features like `livemode` use a bunch of actors to facilitate the feature, therefore, the user would need to list all of these actors explicitly and he/she better not forget any of them.

Why don't we just use a single experimental actor that produces a message which will "enable" (if the message is received) all of the remaining actors? The problem is that anyone can produce such a message. We could write the remaining experimental actors such that they check for a combination of receiving a special message + checking that `LEAPP_EXPERIMENTAL` is set, but this solution is cumbersome and far from perfect (user can enable a different experimental actor, and, unintentionally enable livemode if some actor is producing the special message). Finally, there is no way for actors to check whether a specific experimental actor has been enabled, if I am not wrong -- `LEAPP_EXPERIMENTAL` is binary information (yes/no).

MichalHe:
## Consuming user-specified configuration
### No config provided (i.e., defaults)
Produced livemode config without modifying configuration:
```
{
    "additional_packages": [],
    "autostart_upgrade_after_reboot": true,
    "capture_upgrade_strace_into": "",
    "dracut_network": "",
    "is_enabled": true,
    "setup_network_manager": false,
    "setup_opensshd_with_auth_keys": "",
    "setup_passwordless_root": false,
    "squashfs_fullpath": "/var/lib/leapp/live-upgrade.img",
    "url_to_load_squashfs_from": ""
}
```
### Modified configuration 
Using the following configuration:
```
livemode:
  additional_packages : [ vim ]
  autostart_upgrade_after_reboot : false
  setup_network_manager : true
  setup_passwordless_root : true
  setup_opensshd_using_auth_keys : /root/.ssh/authorized_keys
```
Resulted in producing the following configuration message:
```
{
    "additional_packages": [
        "vim"
    ],
    "autostart_upgrade_after_reboot": false,
    "capture_upgrade_strace_into": "",
    "dracut_network": "",
    "is_enabled": true,
    "setup_network_manager": true,
    "setup_opensshd_with_auth_keys": "/root/.ssh/authorized_keys",
    "setup_passwordless_root": true,
    "squashfs_fullpath": "/var/lib/leapp/live-upgrade.img",
    "url_to_load_squashfs_from": ""
}
```

commands/upgrade/util.py
```diff
@@ -184,11 +196,20 @@ def handle_output_level(args):
...
         os.environ['LEAPP_EXPERIMENTAL'] = '1'
     else:
         os.environ['LEAPP_EXPERIMENTAL'] = '0'
+        args.whitelist_experimental = []
+
+    for experimental_feature in set(args.enable_experimental_feature):
+        actors_needed_for_feature = EXPERIMENTAL_FEATURES[experimental_feature]
+        args.whitelist_experimental.extend(actors_needed_for_feature)
+    if args.enable_experimental_feature:
+        os.environ['LEAPP_EXPERIMENTAL'] = '1'
```
pirat89, line 211: @MichalHe  I am thinking about the situation when there will be no experimental feature at all - which is actually possible state in future. In such a case, there could be any value set as for the option it would lead to: `choices=[]` which would mean that any value would be possible to set based on the current framework behaviour (note I give it a thought to allow to set empty choices, but the behaviour then would be kind of problematic from my POV, leaving up to discussion a possible change in the framework).

In such a case `EXPERIMENTAL_FEATURES[experimental_feature]` would end with `KeyError` So it would be better to use `get(experimental_feature, None)` and add another condition to check whether it's `None` before trying to modify `whitelist_experimental`.

I have been thinking how to deal properly with LEAPP_EXPERIMENTAL in such a case when a nonexisting value is present, as it would lead to actually no experimental code to be executed anyway when `whitelist_experimental` list is not extended. First, I thought that possibly we should not set the value to `'1'` in such a case. However, people know they are already asking for an experimental content when they use the option at the first place, so let's stick with the current condition when speaking about the `LEAPP_EXPERIMENTAL` envar.

pirat89 (changes requested):
See my comments below for more details. Also
* the orig config file is not removed (should be the new one part of the installed rpm or we will keep on user to create it?...)

* the config scanner should be possibly updated? or the logic around the `--enable-experimental-feature` as it still requires set envar `LEAPP_DEVEL_ENABLE_LIVE_MODE=1`?  Or is the requirement of this envar still needed? See the `should_scan_config` function. From the description and name of the option (and PR description) I would expect the envar would not be needed anymore. Let's sync about it yet. Some additional info:  

cmd:
```
 leapp upgrade --debug --no-rhsm --enable-experimental-feature livemode
```

result:
```
Actor: live_mode_config_scanner
Executed: True
Phase: FactsCollection
Started: 2025-04-06T19:19:15.541892Z
Produced messages:
    ----
Executed shell commands:
    ----
Logs:
--- {"level": "DEBUG", "message": "2025-04-06 21:19:15.573 DEBUG    PID: 93812 leapp.workflow.FactsCollection.live_mode_config_scanner: Will not scan livemode config - the live mode is not enabled."}
```

commands/upgrade/util.py
MichalHe: I have extended to code to first check whether `experimental_feature` is present in `EXPERIMENTAL_FEATURES` (if not, we skip processing it).

commands/preupgrade/__init__.py
```diff
@@ -14,6 +14,8 @@
 
 @command('preupgrade', help='Generate preupgrade report')
 @command_opt('whitelist-experimental', action='append', metavar='ActorName', help='Enables experimental actors')
+@command_opt('enable-experimental-feature', action='append', metavar='Feature',
+             help='Enable experimental feature', choices=list(util.EXPERIMENTAL_FEATURES))
```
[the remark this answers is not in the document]
MichalHe: I have updated the help message to contain a list of available experimental features. If there are no experimental features available, the `--help` contains:
```
  --enable-experimental-feature Feature
                        Enable experimental feature. Available experimental
                        features: There are no experimental features available
```
If there are some experimental features (such as livemode) we get:
```
  --enable-experimental-feature Feature
                        Enable experimental feature. Available experimental
                        features: livemode
```

Furthermore, I have verified that running `leapp upgrade --enable-experimental-feature xyz` does not produce any errors when there are no experimental features available.

pirat89 (approved):
thanks! lgtm and so far it seems working based on manual testing

note: failed IPU 7 -> 8 tests are irrelevant - it's already in the removal process

PR #1351, opened 2025-03-03
Correct the list of leapp and leapp-repository rpms, that should be preserved during the 9to10 upgrade

yuravk:
Upgrade 9to10 fails because leapp-upgrade-el9toel10 is removed.
```
Mar 03 11:21:03 localhost upgrade[906]: ====> * scan_installed_target_kernel_version
Mar 03 11:21:03 localhost upgrade[906]:         Scan for the version of the newly installed kernel
Mar 03 11:21:03 localhost upgrade[33445]: Process Process-200:
Mar 03 11:21:03 localhost upgrade[33445]: Traceback (most recent call last):
...
Mar 03 11:21:03 localhost upgrade[33445]:     os.chdir(os.path.join(self._repo_dir, self._directory))
Mar 03 11:21:03 localhost upgrade[33445]: FileNotFoundError: [Errno 2] No such file or directory: '/etc/leapp/repos.d/system_upgrade/common/actors/scaninstalledtargetkernelversion'
```

The reason of that is incorrect excludes set into YUM configuration:
```
2025-03-03 12:21:02.529 DEBUG    PID: 941 leapp.workflow.RPMUpgrade.dnf_upgrade_transaction: External command has finished: ['dnf', 'config-manager', '--save', '--setopt', 'exclude=leapp-upgrade-el8toel9,python3-leapp,leapp,snactor']
2025-03-03 12:21:02.534 DEBUG    PID: 941 leapp.workflow.RPMUpgrade.dnf_upgrade_transaction: The DNF configuration has been updated to exclude leapp packages.
```


pirat89:
Hi \o thank you for the contribution and great catch! You are right, this should be covered. however, I am surprised it actually happened as el10 does not contain any package which could replace it and *-deps package should cover all dependencies to prevent libsolver wanting to remove the leapp-upgrade-el9toel10 package when calculating the DNF transaction. Tere is possibly something weird with the target compose if this problem is happening. But we want this seatbelt definitely applied.

repos/system_upgrade/common/libraries/dnfconfig.py
```diff
@@ -19,6 +19,8 @@ def get_leapp_packages():
     generic = ['leapp', 'snactor']
     if get_source_major_version() == '7':
         return generic + ['python2-leapp', 'leapp-upgrade-el7toel8']
+    if get_source_major_version() == '9':
+        return generic + ['python3-leapp', 'leapp-upgrade-el9toel10']
 
     return generic + ['python3-leapp', 'leapp-upgrade-el8toel9']
```
pirat89, line 25: the problem the root cause of the whole problem - that  the function is duplicating definition of related packages again. to be compatible with previous versions, let's keep the function (we can deprecated later) but use instead `leapp.libraries.common.rpms.get_leapp_packages` to obtain the list of the framework, repository, and tools packages for the source OS major version. The function is written in the way that does already the thing, so something like:

```python
from leapp.libraries.common import rpms

def get_leapp_packages():
    return rpms.get_leapp_packages()
```

should do the thing and we drop the re-definition

pirat89 (changes requested):
It's basically ok but the root cause why this problem is still present -  see my comment - so let's rather replace the function in the way that we drop the problem.

I would also deprecate this function at all, but for the simplicity I am ok to keep the function still present for now. We will mark it as deprecated later.

repos/system_upgrade/common/libraries/dnfconfig.py
pirat89: hmmm....silly me - I should posptone review at 1am. we do not need to even deprecate it. we can just remove it and import the other one...

```
from leapp.libraries.common.rpms import get_leapp_packages
```

as the default behaviour of the other function reflects the original purpose of this one, it's backward compatible so we do not need to even deal with the deprecation...

pirat89 (approved):
lgtm. thanks for the contribution!

PR #1353, opened 2025-03-04
fix(userspacegen): add exeception handling when failing to swap RHUI clients

MichalHe:
**Note: This PR has been created as part of teaching git on MUNI (how pull-requests and origins work)**

Leapp swaps to RHUI target clients (uninstall source client+install target client) in the scratch container to gain repository access. To perform this step atomically, without loosing repository access in between, `dnf shell` is invoked. The instruction as to which RPMs should be uninstalled and which should be installed are given on stdin of the `dnf shell` command. Currently, if we fail to swap clients we crash with an unhandled exception `CalledProcessError` that contains no information about what went wrong since the actual performed transaction is hidden within the stdin of the process. This patch adds error handling, so that we can tell that we have failed to swap RHUI clients. Leapp's logs also contain the full DNF transaction as well as possible explanations on why we failed.

jira-ref: RHEL-77945

### DIFF
--- a/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
+++ b/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
@@ -1239,7 +1239,35 @@ def setup_target_rhui_access_if_needed(context, indata):
         'shell'
     ]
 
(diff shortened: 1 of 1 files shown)


TomasTomecek (approved):
Looks good to me, please fix the linters and we can merge.

pirat89 (approved):
lgtm. merging

PR #1357, opened 2025-03-10
docs(livemode): add livemode documentation

MichalHe:
This commit introduces a high-level documentation of the livemode
feature. The following is documented:
 - how the upgrade process differs from the standard one when using
   livemode
 - why is the feature useful, i.e., why would anyone use it
 - how to enable and use the feature
 - what configuration options are available, including an example of
   configuration

Requires #1350

jira-ref: RHELMISC-9703


docs/source/livemode.md
```diff
@@ -0,0 +1,87 @@
+# Livemode - having a more familiar, possibly interactive, upgrade environment
+**Note that the livemode feature is experimental.**
```
pirat89, line 2: Note that the suggestion is littlebit malformed and not sure how to correctly input triple backticks here :) but you get the idea I believe.

```suggestion
```{warning}
Note that the livemode feature is experimental.
\`\`\`
```

docs/source/livemode.md
```diff
@@ -0,0 +1,87 @@
...
+
+Leapp features an experimental feature named _livemode_ that partially replaces
```
pirat89, line 4: The sentense is kind of hard to get, let's simplify it to something like

```suggestion
_livemode_ is an experimental feature that partially replaces
```

pirat89:
@MichalHe I think we should create a dedicated section for experimental features. We should not put it into docs just like that but rather having a dedicated section for it. I am thinking about putting it under:
```
 * Configuring in-place upgrade
    * ... 
    * Experimental Features
      * LiveMode
```

When we move it in future from the section experimental features somewhere else, we can set a rule for relocation from the org URL to the new one. I see now that configuration section is not "separated" correctly, so it would need small adjustment as well. wdyt?

docs/source/livemode.md
MichalHe: Suggestion applied to the `index.rst` of `experimental-features` subsection.

docs/source/livemode.md
MichalHe: Suggestion applied :+1:

MichalHe:
@pirat89 Done, I have restructured the documentation to include a section named _Experimental features_ under the _Configuring IPU_ section as you have suggested. I have also adopted the "camel case" naming of the feature in the documentation _LiveMode_.

pirat89 (approved):
nice! merging

PR #1363, opened 2025-04-14
Handle missing OS minor version gracefully

MichalHe:
Handle missing OS minor version gracefully, instead of crashing at various points of execution: when checking the version format in the CLI, or when using config.version library functions.

The proposed solution adds a special object into  `upgrade_paths.json` called `_virtual_versions` (under the upgrade paths for `CentOS`). When running on CentOS, we do not require the `MAJOR.MINOR` version format anymore. Furthermore, whenever the (old) code uses `config.version` library functions, we use minor versions defined in `_virtual_versions` to use a complete version for version comparisons. This approach guarantees predictable and correct behavior of version-specific checks. Other solutions were discussed on a team cabal, and we found edge cases when the code could exhibit unexpected behavior, such as version-specific workaround code not being executed (or being executed) at all, although it should be. Therefore, this solution, although a bit labor intensive, seems like the only reliable way to make the already written code work on CentOS.

**TODO**:
- [x] Check that version.matches_version and other affected library functions behave as expected, i.e., they use virtual versions

Jira-ref: RHEL-80334


repos/system_upgrade/common/libraries/config/version.py
```diff
@@ -141,17 +141,17 @@ def _version_to_tuple(version):
...
+    """Raise ``ValueError`` if provided versions are not strings in the form ``<integer>.<integer>``."""
+    version_format_regex = re.compile('^([1-9]\d*)\.(\d+)$')
+    for version in versions:
+        if not re.match(version_format_regex, version):
             raise ValueError("Versions have to be in the form of '<integer>.<integer>' "
                              "but provided was '{}'".format(versions))
 
 
-def _simple_versions(versions):
+def _are_comparision_operators_used(versions):
```
pirat89, line 152: do you mean `comparison` ? :)

commands/command_utils.py
```diff
@@ -16,26 +18,52 @@
...
+    MAJOR_MINOR = _VersionFormat('MAJOR_VER.MINOR_VER', re.compile(r"^([1-9]\d*)\.(\d+)$"))
+
+
+class _VersionKind(str, Enum):
+    """ Enum encoding information whether the given OS version is source or target. """
+    SOURCE = 'source'
+    TARGET = 'target'
+
+
+class RHELDistros(str, Enum):
```
pirat89, line 36: I would not use the `RHELDistros` name in this context. Let's use rather something like `DistroIDs`

commands/command_utils.py
```diff
@@ -71,15 +99,33 @@ def get_upgrade_flavour():
     return LEAPP_UPGRADE_FLAVOUR_DEFAULT
 
 
-def _retrieve_os_release_contents(_os_release_path='/etc/os-release'):
+@functools.cache
```
pirat89, line 102: @MichalHe `functools.cache` is defined since Python 3.9, but the code is executed with Python 3.6 as well:
* https://docs.python.org/3/library/functools.html#functools.cache

Is there a high enough number of calls to have a benefit from using the cache wrapper at all?

repos/system_upgrade/common/libraries/config/version.py
MichalHe: Should, be fixed now

commands/command_utils.py
MichalHe: Fixed, thanks.

commands/command_utils.py
MichalHe: Thank you for noticing, I have dropped the use of `functools`.

> Is there a high enough number of calls to have a benefit from using the cache wrapper at all?

Since it prevents us from accessing the filesystem (systemcalls are usually expensive), then the cost of the wrapper is small. I somehow think that the performance of the tool (how fast can I iterate) is something that I don' enjoy about the development experience, so I try to take into consideration also that aspect of the code.

MichalHe:
@oamg/developers review please

commands/command_utils.py
```diff
@@ -16,26 +18,52 @@
...
+_DISTRO_VERSION_FORMATS = {
+    DistroIDs.RHEL: VersionFormats.MAJOR_MINOR,
+    DistroIDs.CENTOS: VersionFormats.MAJOR_ONLY,
+}
+"""
+Maps distro ID to the expected OS version format.
+
+If a distro is not listed in the dictionary, then VersionFormats.MAJOR_MINOR
+is used as a default.
+"""
```
evgeni, line 50: How can consumers extend these lists (without forking the codebase)?
While this probably won't matter much (for Alma and friends) if the fallback is MAJOR_MINOR, but still curious :)

MichalHe: Hmm, community projects can always create an upstream PR extending the list, and there should be no problem in merging such a change. Not sure whether implementing something that provides extensibility is worth the effort, since it is not as simple as 'writing an actor' due to the code being executed very early.

commands/command_utils.py
```diff
@@ -71,15 +99,32 @@ def get_upgrade_flavour():
...
+        for line in lines:
+            if '=' not in line:
+                continue
+
+            key_val_pair = line.strip().split('=', 1)
+
+            if strip_double_quotes:
+                key_val_pair = strip_quotes_from_field_value(key_val_pair)
+
+            os_release_contents[key_val_pair[0]] = key_val_pair[1]
```
pirat89, line 125: nitpick, but I would do rather something like this:
```suggestion
            key, val = line.strip().split('=', 1)

            if strip_double_quotes:
                val = strip_quotes_from_field_value(val)

            os_release_contents[key] = val
```

The nested `strip_quotes_from_field_value` function would need to be updated as well then. Possibly it could be also renamed to `strip_quotes`.

repos/system_upgrade/common/files/upgrade_paths.json
```diff
@@ -21,6 +21,11 @@
         "default": {
             "8": ["9"],
             "9": ["10"]
+        },
+        "_virtual_versions": {
+            "8": "8.10",
+            "9": "9.5",
+            "10": "10.0"
```
pirat89, line 28: Can be changed later, but based on the merge time, we can virtual versions already 9.6, 10.0. It would be possible even 9.7, 10.1 as Centos Stream contains already changes for new minor releases, we can stay with 9.6, 10.0.

This does not block the merge, it's a nitpick now.

repos/system_upgrade/common/actors/ipuworkflowconfig/libraries/ipuworkflowconfig.py
```diff
@@ -92,8 +93,7 @@ def load_upgrade_paths_definitions(paths_definition_file):
     return definitions
 
 
-def load_raw_upgrade_paths_for_distro_and_flavour(distro_id, flavour, paths_definition_file='upgrade_paths.json'):
-    all_definitions = load_upgrade_paths_definitions(paths_definition_file)
+def extract_paths_upgrade_paths_for_distro_and_flavour(all_definitions, distro_id, flavour):
```
pirat89, line 96: nitpick, but why `extract_paths_upgrade_paths_for_distro_and_flavour` instead of `extract_upgrade_paths_for_distro_and_flavour`? Reading the name, it's a little bit confusing

repos/system_upgrade/common/actors/ipuworkflowconfig/libraries/ipuworkflowconfig.py
```diff
@@ -117,6 +117,39 @@ def construct_models_for_paths_matching_source_major(raw_paths, src_major_versio
     return multipaths_matching_source
 
 
+def construct_virtual_versions(all_upgrade_path_defs, distro_id, source_version, target_version):
+    if distro_id.lower() != 'centos':
+        return (source_version, target_version)
+
+    centos_upgrade_paths = all_upgrade_path_defs.get('centos', {})
+    if not centos_upgrade_paths:
+        raise StopActorExecutionError('There are no upgrade paths are defined for CentOS.')
```
pirat89, line 126: typo:

```suggestion
        raise StopActorExecutionError('There are no upgrade paths defined for CentOS.')
```

repos/system_upgrade/common/models/ipuconfig.py
```diff
@@ -33,6 +33,21 @@ class Version(Model):
...
+    virtual_source_version = fields.String()
+    """
+    Source OS version used when checking whether to execute version-dependent code.
+
+    On RHEL and other systems that have version of the form MINOR.MAJOR, `virtual_source_version`
+    matches `source_version`.
+
+    CentOS has version of the form MAJOR, lacking the minor version number. The
+    `virtual_source_version` value is obtained by combining CentOS major
+    version number with a minor version number stored internally in leapp.
```
pirat89, line 45: This could be confusing as there is not such a value stored in leapp (project).

```suggestion
    version number with a minor version number stored internally in the upgrade_paths.json file.
```

repos/system_upgrade/common/files/upgrade_paths.json
MichalHe: I have modified virtual version for CentOS 9 to 9.6. Virtual version for CentOS 10 is unchanged.

repos/system_upgrade/common/actors/ipuworkflowconfig/libraries/ipuworkflowconfig.py
MichalHe: There is no reason, I'm just blind :). Should be fixed.

repos/system_upgrade/common/actors/ipuworkflowconfig/libraries/ipuworkflowconfig.py
MichalHe: Thank you, should be fixed.

pirat89 (approved):
thanks! lgtm. I will do yet some manual testing before merging but everything seems right and it's covered by unit tests. The failed tests do are not releated to this change.

MichalHe:
@pirat89 Ups, I wanted to merge PR1378, but I have not noticed that I switched to the wrong tab (with this PR). Sorry, I did not intent to do so. Nothing should be seriously broken (I hope). Let me know if you discover any issues, I will promptly do a follow up PR.

pirat89:
@MichalHe  np :D I just wanted to merge it now.

PR #1364, opened 2025-04-14
[DRAFT] Drop IPU 7 -> 8 related upstream CI tests

tomasfratrik:
RHELMISC-11004


pirat89 (approved):
lgtm. It seems to me even that the reuse-copr-build.yml file is no longer used anymore. At minimum, it does not contain even epel-9* chroot defined. We could possibly remove it in future, but let's keep it for now and resolved that in a followup PR if it's true.

PR #1368, opened 2025-04-23
Add upstream doc about running single actor

karolinku:
JIRA: RHELMISC-11596


pirat89:
taking a short look, some sections will need to be rewritten. let's take a look at it after we finish the work on other PRs for upcoming release.

PR #1369, opened 2025-04-25
Fix lint in Makefile for docs

karolinku:
isort is failing when getting empty list of arguments, what takes place when only doc files was modified in a commit. This fix handles that issue.

JIRA: RHELMISC-11679

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -356,16 +356,17 @@ lint:
 		echo "--- Linting done. ---"; \
 	fi
 
-	if [[  "`git rev-parse --abbrev-ref HEAD`" != "$(MASTER_BRANCH)" ]] && [[ -n "`git diff $(MASTER_BRANCH) --name-only --diff-filter AMR`" ]]; then \
+	if [[  "`git rev-parse --abbrev-ref HEAD`" != "$(MASTER_BRANCH)" ]]; then \
 		. $(VENVNAME)/bin/activate; \
-		git diff $(MASTER_BRANCH) --name-only --diff-filter AMR | grep -v "^docs/" | xargs isort -c --diff || \
-		{ \
+		files_to_sort=`git diff main --name-only --diff-filter AMR | grep -v "^docs/"`; \
+		if [ -n "$$files_to_sort" ]; then \
+			echo "$$files_to_sort" | xargs isort -c --diff || { \
 			echo; \
-			echo "------------------------------------------------------------------------------"; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -358,15 +358,17 @@ lint:
...
+        if [ ! -z $files_to_sort ]; then \
+            echo $files_to_sort | xargs isort -c --diff; \
+            echo "--- isort check done. ---"; \
+        else \
+            echo; \
+            echo "Hint: Apply the required changes.";\
+            echo "      Execute the following command to apply them automatically: make lint_fix"; \
+            exit 1; \
+        fi \
+    fi
```
pirat89, line 371: This is actually leading to basically same result.  The code is saying: fail if the list of files is empty. It's because the condition is based on the emptiness of the file list for the isort instead of the isort output.

One solution is to update the condition on line 359 to reflect the change on line 361. so:
```
	if [[  "`git rev-parse --abbrev-ref HEAD`" != "$(MASTER_BRANCH)" ]] && [[ -n "`git diff $(MASTER_BRANCH) --name-only --diff-filter AMR | grep -v "^docs/"`" ]]; then \
```

another option is the modified solution introduced in this PR, moving the condition inside / dropping the first check (using just one generated list for everything) - kindof pseudocode:
```
if [[  "`git rev-parse --abbrev-ref HEAD`" != "$(MASTER_BRANCH)" ]]; then \
    files_to_sort=$(git diff main --name-only --diff-filter AMR | grep -v "^docs/"); \
    if [ -n "$files_to_sort" ]; then \
        echo $files_to_sort | xargs isort -c --diff || { \
         .... al these echos ...
         exit 1; \
        } && echo "--- isort check done. ---"; \
    fi
```

I like more the second one style, as in the first case the problem could occur in future again if the "filter" is updated just in one of places.

karolinku: The issue occurred when `| xargs isort -c --diff` was reached with empty output, it is not accepted by isort. My solution runs isort only when there is anything to proceed.

More over, I wanted to remove bash constructions like `A || B && C`, because they behaviour is not very clear and and understandable at first glance.

pirat89: > The issue occurred when | xargs isort -c --diff was reached with empty outpu

I know, what I wrote is not saying anything else :)

pirat89 (approved):
lgtm and works (tested manually)

PR #1370, opened 2025-04-28
Improve report of unsupported network configuration

tomasfratrik:
* Remove KCS from summary since it is present in the related links.
* Generalize KCS title when using Kernel-Assigned NIC Names.
* Show KCS about unsupported network configuration device types only when IPU 8 -> 9.

Jira: RHEL-77175


pirat89 (approved):

pirat89 (approved):

pirat89 (changes requested):
unti tests are failing

pirat89 (approved):
everything looks good now

PR #1372, opened 2025-04-30
Fix parsing of the kernel cmdline

pirat89:
Original parsing has been splitting string tokens by "=" without a limit, means that e.g.
   root=UUID=some-id
has been split to
   (root, UUID, some-id)
and and stored just "root" and "UUID" strings in the parsed msg, instead of "root" and "UUID=some-id".

So split the key=value just based on the first occurence of "=" instead.

jira: RHELMISC-12490

TODO:
- [x] add the missing unit-tests

### DIFF
--- a/repos/system_upgrade/common/actors/scankernelcmdline/tests/test_scankernelcmdline.py
+++ b/repos/system_upgrade/common/actors/scankernelcmdline/tests/test_scankernelcmdline.py
@@ -1,0 +1,67 @@
+import pytest
+
+from leapp.libraries.actor import scankernelcmdline
+from leapp.libraries.common.testutils import CurrentActorMocked, produce_mocked
+from leapp.libraries.stdlib import api, CalledProcessError
+from leapp.models import KernelCmdline, KernelCmdlineArg
+
+
+def mock_cmd_output():
+    expected_cmd_output = (
+        'BOOT_IMAGE=(hd0,msdos1)/vmlinuz-xxx root=UUID=some_uid ro console=tty0'
+        ' console=ttyS0,115200 rd_NO_PLYMOUTH biosdevname=0 net.ifnames=0 crashkernel=auto'
+        )
+    return expected_cmd_output
+
+
+def test_cmdline_output(monkeypatch):
+
+    monkeypatch.setattr(scankernelcmdline, 'get_cmdline_input', mock_cmd_output)
+    current_actor = CurrentActorMocked(src_ver='8.10', dst_ver='9.6')
+    monkeypatch.setattr(api, 'current_actor', current_actor)
+    monkeypatch.setattr(api, 'produce', produce_mocked())
+
+    scankernelcmdline.parse_cmdline_input()
+
+    expected_params = [KernelCmdlineArg(key=k, value=v) for k, v in [
+        ('BOOT_IMAGE', '(hd0,msdos1)/vmlinuz-xxx'),
+        ('root', 'UUID=some_uid'),
+        ('ro', None),
+        ('console', 'tty0'),
+        ('console', 'ttyS0,115200'),
+        ('rd_NO_PLYMOUTH', None),
+        ('biosdevname', '0'),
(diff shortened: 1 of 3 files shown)


repos/system_upgrade/common/actors/scankernelcmdline/tests/test_scankernelcmdline.py
```diff
@@ -0,0 +1,37 @@
...
+        ('console', 'ttyS0,115200'),
+        ('rd_NO_PLYMOUTH', None),
+        ('biosdevname', '0'),
+        ('net.ifnames', '0'),
+        ('crashkernel', 'auto')]]
+
+    expected_output_msg = KernelCmdline(parameters=expected_params)
+
+    assert api.produce.model_instances
+    assert expected_output_msg == api.produce.model_instances[0]
```
pirat89, line 37: @karolinku  current test is good, but it's not sufficient. This test would not actually reveal the bug that is fixed by this PR. I suggest either to update the test or create another onw that will test the bug specifically. Both ways are ok for me.

Also, current test is not covering obtaining the information about the kernel cmdline from the system - it tests only the parsing of the obtained data. I am ok with that, but I would add at least a comment to the test that this is not tested with a short explanation why.  In case you would like to cover it, I suggest to move that call also to the library to be able to cover it by unit-tests (so it could be mocked). But as I mentioned, the comment in this case would be enough.

repos/system_upgrade/common/actors/scankernelcmdline/tests/test_scankernelcmdline.py
```diff
@@ -0,0 +1,63 @@
+from leapp.libraries.actor import scankernelcmdline
+from leapp.libraries.common.testutils import CurrentActorMocked, produce_mocked
+from leapp.libraries.stdlib import api, CalledProcessError
+from leapp.models import KernelCmdline, KernelCmdlineArg
+
+
+def mock_cmd_output():
+
+    mocked_cmd = 'BOOT_IMAGE=(hd0,msdos1)/vmlinuz-xxx root=UUID=some_uid ro console=tty0 '\
+        'console=ttyS0,115200 rd_NO_PLYMOUTH biosdevname=0 net.ifnames=0 crashkernel=auto'
```
pirat89, line 10: Omit using backslach to join multiple lines at any cost - consider them forbidden in python code in this project. Use rather brackets:
```suggestion
    mocked_cmd = (
        'BOOT_IMAGE=(hd0,msdos1)/vmlinuz-xxx root=UUID=some_uid ro console=tty0 '
        'console=ttyS0,115200 rd_NO_PLYMOUTH biosdevname=0 net.ifnames=0 crashkernel=auto'
    )
```
the same below

repos/system_upgrade/common/actors/scankernelcmdline/tests/test_scankernelcmdline.py
```diff
@@ -0,0 +1,63 @@
...
+
+def test_cmdline_run_failed(monkeypatch):
+
+    def run_mocked_error(cmd, **kwargs):
+        assert cmd == ['cat', '/proc/cmdline']
+        raise CalledProcessError("Mocked error raised", cmd, "result")
+
+    monkeypatch.setattr(scankernelcmdline, 'run', run_mocked_error)
+    cmd_output = scankernelcmdline.get_cmdline_input()
+    assert cmd_output == ''
```
pirat89, line 63: The same should be done for `OSError`.  I suggest to parametrize the test.

pirat89: information about exception that should be raised. You can get inspiration here: https://github.com/oamg/leapp-repository/blob/4e87908b2e17d87b69bcec29ba1e9b777439a0fe/repos/system_upgrade/common/actors/systemd/scansystemdtarget/tests/test_scansystemdtarget.py#L89

karolinku (approved):

PR #1362, opened 2025-04-10
Add handling of shorten PCI ID and lowercases

karolinku:
HW data file is inconsistent and sometimes contains both upper/lower cases so this patch is unifying it. Also PCI ID can be passed as full ID: Vendor:Device:SVendor:SDevice, but also in shorten version: Vendor:Device. This patch introduce handling of both cases.

JIRA: RHEL-72544

PR #1376, opened 2025-05-07
Prevent device list from containing incorrent content

karolinku:
```
Some PCI devices got from lspci output do not have to neccessarily
provide SVendor and SDevice fields. PCI ID of such devices is
composed in the pci_device_scanner actor just from:
    Vendor:Device
instead of the full id:
    Vendor:Device:SVendor:SDevice

The recent change comparing such devices with bundled DDDD
data drops last two fragments from the composed PCI ID,
which led to the situation when the `shortened_pci_id`
resulted in an empty string and it has been incorrectly
matched with one of (consider random) drivers and possibly
inhibited the in-place upgrade.

So let's ensure that we do not match any PCI device with
any entry from DDDD undefined PCI ID.
```

Related PR: #1362

Thanks to @yuravk for the reported problem.  

JIRA: RHEL-72544


pirat89 (approved):
lgtm and it seems to be working now correctly. tested manually

PR #1377, opened 2025-05-07
Deprecate is_rhel_alt from share libraries

pirat89:
The function has no value anymore as RHEL-ALT 7 is EOL for years and RHEL 7 is not longer maintained in this project neither. I have considered the direct removal as this function is really old (it's not even used anywhere in the project anymore).

But as it is just one small function without additional impact and another rhel-alt artifacts became deprecated this month, I consider it better to fulfill our promise and add it on the deprecation list for 6months instead.


MichalHe (approved):
Since E2E tests are passing (except one unrelated failure), and the change does not affect functionality, I say LGTM!

PR #1378, opened 2025-05-07
Handle RPM GPG keys per distribution

pirat89:
Originally we handled rpm-gpg keys only for RHEL distribution. As we want to generalize the solution to be more distro agnostic, correct handling of RPM GPG keys is in this case important.

 The original trusted rpm-gpg directories under `files/rpm-gpg/` is not flexible when considering other distributions as well. The new path pattern will be `files/distro/<DISTRO>/rpm-gpg/`.


MichalHe (commented):
Code-wise the PR looks good. I have verified that all proposed GPG keys are indeed the GPG keys provided by [centos-gpg-keys-10.0-8.el10.noarch.rpm](https://mirror.stream.centos.org/10-stream/BaseOS/x86_64/os/Packages/centos-gpg-keys-10.0-8.el10.noarch.rpm) and [centos-gpg-keys-9.0-26.el9.noarch.rpm](https://mirror.stream.centos.org/9-stream/BaseOS/x86_64/os/Packages/centos-gpg-keys-9.0-26.el9.noarch.rpm). I am going to manually test that GPG keys are being imported during the upgrade. If anything goes well, I will approve and merge.

MichalHe:
Running a 9>10 upgrade manually, the following line can be found in the upgrade log:
```
... Info: Importing GPG keys from: /etc/leapp/repos.d/system_upgrade/common/files/distro/rhel/rpm-gpg/10/RPM-GPG-KEY-redhat-release
```

The upgrade proceeds without any issues. Since the above path is correct, I approve

MichalHe:
The failed e2e test seems to be unrelated to this PR. Merging :saluting_face:

MichalHe (approved):

PR #1382, opened 2025-05-30
Makefile: Non-zero exit code on failed tests in container

pirat89:
Previously all exit codes from execution of tests inside container have been ignored and make command ended with zero always.


.github/workflows/unit-tests.yml
```diff
@@ -9,43 +9,43 @@ on:
 
 jobs:
   test:
-    name: Run unit tests in containers
+    name: 'Run: ${{ matrix.scenarios.name }}'
```
karolinku, line 12: I like this idea of shorten test name

PR #1383, opened 2025-06-04
support rhel9 upgrade to rhel10  using rhui on alibaba cloud

jinkkkang:
support rhel9 upgrade to rhel10  using rhui on alibaba cloud
we test the  rhel 9.6  x86_64&arm64 and the results were ok

there also a PR for update repo.json

### DIFF
--- a/repos/system_upgrade/common/libraries/rhui.py
+++ b/repos/system_upgrade/common/libraries/rhui.py
@@ -348,6 +348,13 @@ RHUI_SETUPS = {
                         ('content.crt', RHUI_PKI_PRODUCT_DIR)
                       ],
                       os_version='9'),
+        mk_rhui_setup(clients={'aliyun_rhui_rhel10'}, leapp_pkg='leapp-rhui-alibaba',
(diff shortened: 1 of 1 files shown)


repos/system_upgrade/common/libraries/rhui.py
```diff
@@ -566,15 +580,29 @@ def mk_rhui_setup(clients=None, leapp_pkg='', mandatory_files=None, optional_fil
...
+            'target_pkg': 'aliyun_rhui_rhel10',
+            'leapp_pkg': 'leapp-rhui-alibaba',
+            'leapp_pkg_repo': 'leapp-alibaba.repo',
+            'files_map': [
+                ('content.crt', RHUI_PKI_PRODUCT_DIR),
+                ('key.pem', RHUI_PKI_DIR),
+                ('leapp-alibaba.repo', YUM_REPOS_PATH)
+            ],
+        },
+    }
```
MichalHe, line 595: This is no longer necessary - the dictionary should not be used anywhere if I am right. Nevertheless, your change is OK so no need to remove it.

MichalHe (approved):

PR #1387, opened 2025-06-15
libs(rhui): add rhel10 targets for upgrades

MichalHe:
Adds RHEL 10 system definitions to leapp's internal database so that upgrades are possible. Naturally, a necessary dependence is having RHEL 9 `leapp-rhui-{cloud_provider}` packages prepared. 

Also implements a quality-of-life improvement that prevents leapp from "scanning" source/debug repositories as they will not be used during the upgrade. This prevents silly crashes when the cloud provider forgot to sync debug/source repositories (`dnf repolist` which we are running would cause `CalledProcessError` and ultimately crash the entire upgrade).

Jira-ref: RHEL-64911, RHEL-64910

### DIFF
--- a/etc/leapp/files/repomap.json
+++ b/etc/leapp/files/repomap.json
@@ -1,5 +1,5 @@
 {
-    "datetime": "202505201636Z",
+    "datetime": "202507171303Z",
     "version_format": "1.3.0",
     "provided_data_streams": [
         "4.0"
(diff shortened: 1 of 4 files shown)


pirat89 (approved):
lgtm. Commit msgs are clearly explaining the changes and I checked repomap twice. We are aware of some discrepancies regarding ARM architecture on Azure (missing repos) but that can be handled later for the final builds or a future release. I haven't tested the change manually.

PR #1389, opened 2025-06-17
Remove obsolete workflows

danmyway:
* tmt-tests.yml and reuse-copr-build.yml were replaced by the Packit workflows ages ago
* removed the obsolete workflows


pirat89 (approved):
true story! lgtm! Thanks for the help with cleaning!

PR #1392, opened 2025-07-01
Inhibit upgrade on systems with cgroupsv1 on 9to10

matejmatuska:
cgroups-v1 were deprecated in RHEL 9 and are unsupported in RHEL 10. The kernel cmdline options to switch to cgroups-v1 are ignored. Users have to migrate to cgroups-v2 in order to upgrade.

Both legacy and hybrid modes are unsupported. Only unified hierarchy (cgroups-v2) is supported. More info on how the modes are switched is at: https://www.freedesktop.org/software/systemd/man/247/systemd.html#systemd.unified_cgroup_hierarchy and https://www.freedesktop.org/software/systemd/man/247/systemd.html#systemd.legacy_systemd_cgroup_controller.

Jira: RHEL-81212

### DIFF
--- a/repos/system_upgrade/el9toel10/actors/inhibitcgroupsv1/actor.py
+++ b/repos/system_upgrade/el9toel10/actors/inhibitcgroupsv1/actor.py
@@ -14,7 +14,7 @@ class InhibitCgroupsv1(Actor):
     hierarchy (cgroups-v2) is supported.
     """
 
-    name = 'inhibit_cgroupsv1'
+    name = "inhibit_cgroupsv1"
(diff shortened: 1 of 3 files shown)


matejmatuska:
Another thing up for discussion is the remediation command, I see a case for removing it because of the need to specify a kernel and I am not sure what would be the right option, I've kept it to ALL for now.

pirat89 (approved):
lgtm and tests seems ok. I haven't tried it manually but I do not expect problems at this moment. we plan to test it in upcoming days and we need to merge it now. So, let's be brave!

PR #1394, opened 2025-07-01
Point to leapp-repository contribution guidelines

bocekm:
Instead of the leapp framework contribution guidelines.


matejmatuska (approved):
LGTM, I tried all the links.

Only possible change I see is to link to stable instead of latest docs, but given this is where the development happens I think latest is the better choice.

PR #1395, opened 2025-07-02
feat(almalinux): add new mappings for repositories with almalinux distro ID

yuravk:
The PR is to just make visible changes of new mappings for repositories with `almalinux` distro ID.

That's in scope of [Allow upgrades of almalinux (AlmaLinux OS) #1391](https://github.com/oamg/leapp-repository/pull/1391)

### DIFF
--- a/etc/leapp/files/repomap.json
+++ b/etc/leapp/files/repomap.json
@@ -308,6 +308,38 @@
                     "arch": "aarch64",
                     "channel": "ga",
                     "repo_type": "rpm",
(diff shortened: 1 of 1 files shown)


pirat89:
@yuravk mapping added in #1480 - keeping extras and rs repos unmapped as in this PR until we implement new format that will allow to handle 1:0 mapping between various distros. We will share more info about the format change in 1-2 months (currently it's just in messy notes from brainstorming)

PR #1398, opened 2025-07-03
Read the DNF config by module.py library

karolinku:
The DNF configuration has not been loaded when trying to get
information about available module streams (library module.py).
This causes a traceback e.g. on systems which must access DNF
repositories via a proxy.
This patch introduces loading the DNF configuration before trying
to access remote resources.

Jira: RHEL-39095

### DIFF
--- a/repos/system_upgrade/common/actors/rpmscanner/libraries/rpmscanner.py
+++ b/repos/system_upgrade/common/actors/rpmscanner/libraries/rpmscanner.py
@@ -25,6 +25,8 @@ except ImportError:
 
 def _get_package_repository_data_yum():
     yum_base = yum.YumBase()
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/common/actors/rpmscanner/libraries/rpmscanner.py
```diff
@@ -25,6 +25,8 @@
 
 def _get_package_repository_data_yum():
     yum_base = yum.YumBase()
+    # DNF configuration is not loaded here, since no impact for operations
+    # done by the actor is observed here
```
pirat89, line 29: this was meant for `_get_package_repository_data_dnf()` function :) but that's a nitpick. no need to change it

PR #1401, opened 2025-07-07
Disable localpkg_gpgcheck parameter from plugin data

karolinku:
This commit is setting the localpkg_gpgcheck DNF option within the
rhel-upgrade plugin to 0 (disabled).
The upgrade process has been blocked with "Error: GPG check FAILED"
when dnf has been configured to apply gpg check also on local packages
(localpkg_gpgcheck=1). That's because the bundled leapp*-deps meta
packages, which are managing Leapp and leapp-repository dependencies
during the transition to the new system, are not signed by the Red
Hat key. Therefore, this option needs to be disabled.

Jira: RHEL-47472


pirat89 (approved):
lgtm! tested manually for IPU 8 -> 9 it works as expected. it cab be merged after tests are finished.

PR #1403, opened 2025-07-09
Makefile: Partially fix ACTOR tests and enable them in containers

matejmatuska:
The ACTOR variable for running tests of a single actor has long been unusable, git bisect led me to dc3abf63e581c1a5e24c9 as the commit where it was (first?) broken due to the same model module being defined in both el7toel8 and el8toel9 repos.

This patch unfortunately doesn't fix that, but works around it by enabling the utils/actor_path.py script to search in specific repositories. These are passed from the Makefile via REPOSITORIES variable.

As the Makefile rules for containerized tests already use repositories, the ACTOR variable is just passed along.

NOTE: the code in actor_path.py is ugly and uses private APIs, however that's nothing new :).

---
Yes, there is also `dev_test_no_lint` which is still faster, but currently doesn't work in containers.


Makefile
```diff
@@ -12,9 +12,18 @@ REPOS_PATH=repos
...
 LIBRARY_PATH=
 REPORT_ARG=
-REPOSITORIES ?= $(shell ls $(_SYSUPG_REPOS) | xargs echo | tr " " ",")
-SYSUPG_TEST_PATHS=$(shell echo $(REPOSITORIES) | sed -r "s|(,\\|^)| $(_SYSUPG_REPOS)/|g")
-TEST_PATHS:=commands repos/common $(SYSUPG_TEST_PATHS)
+
+ifdef ACTOR
+	# If REPOSITORIES is set, search for the actor only in the specified repos,
+	# if REPOSITORIES is not set i.e. it's empty, search in all repos, this is
+	# broken due to name collisions in repositories (FIXME)
```
MichalHe, line 19:
```suggestion
	# If REPOSITORIES is set, the `utils/actor_path.py` script searches for the actor only in the specified repos.
	# if REPOSITORIES is not set i.e. it's empty, all repositories are searched -- this is
	# broken due to name collisions in repositories (FIXME)
```

MichalHe: This will make it a bit clearer that the script `utils/actor_path.py` has the last argument optional (if it is missing, it will search everything)

MichalHe (commented):
Code looks OK, I have provided only a minor suggestion regarding the comment. I will give Approve once I run it a couple of times.

Makefile
```diff
@@ -375,11 +379,18 @@ lint_fix:
 	echo "--- isort inplace fixing done. ---;"
 
 test_no_lint:
+	@if [ -z "$(REPOSITORIES)" -a -n "$(ACTOR)" ]; then \
+		printf "\033[0;31mWARNING\033[0m: Running tests with ACTOR without specifying REPOSITORIES is currently broken, specify REPOSITORIES with only one elXtoelY repository (e.g. REPOSITORIES=common,el8toel9)\n" 2>&1; \
```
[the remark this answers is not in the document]
matejmatuska: Done, but it's not pretty and there is not really a nicer way.

pirat89 (changes requested):
This change breaks the execution for linters at this moment. See, it's trying to use python3.6 for el9toel10 actors which are python3.9 compatible only:

```
Process Process-13:
    self.run()
  File "/payload/tut/lib/python3.6/site-packages/leapp/repository/actor_definition.py", line 32, in inspect_actor
    definition.load()
  File "/payload/tut/lib/python3.6/site-packages/leapp/repository/actor_definition.py", line 199, in load
    self._module = load_module(importer, name)
  File "/payload/tut/lib/python3.6/site-packages/leapp/compat.py", line 86, in load_module
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 678, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/payload/repos/system_upgrade/el9toel10/actors/mysql/scanmysql/actor.py", line 2, in <module>
    from leapp.libraries.actor import scanmysql
  File "/payload/repos/system_upgrade/el9toel10/actors/mysql/scanmysql/libraries/scanmysql.py", line 35, in <module>
    def _check_incompatible_config() -> set[str]:
TypeError: 'type' object is not subscriptable
2025-08-08 15:50:53.225 ERROR    PID: 952 leapp.repository.system_upgrade_el9toel10: Process inspecting actor in actors/mysql/scanmysql failed with 1
Inspection of actor in actors/mysql/scanmysql failed
make: *** [Makefile:381: test_no_lint] Error 1
```

matejmatuska:
@pirat89 I fixed the sanity check in a new commit, tests now pass.

matejmatuska:
@pirat89 It should be resolved now, I added a manual call to `snactor repo findin the container ` which generates `~/.config/leapp/repos.json` which is required by the `utils/actor_path.py` script to link leapp repositories.

matejmatuska:
Okay I pushed a fix, but it's ugly. At this point I think we should extract a lot of the containerized tests related code into a script or separate Makefile maybe. It's become a undebuggable mess.

The error is still being printed, but there is a line in yellow informing that the test is skipped.

Makefile
```diff
@@ -370,14 +375,28 @@ lint_fix:
...
+	for dir in $$(echo $(REPOSITORIES) | tr "," " "); do \
+		echo "Running sanity-check in $(_SYSUPG_REPOS)/$$dir"; \
+		(cd $(_SYSUPG_REPOS)/$$dir && snactor workflow sanity-check ipu); \
+	done
+
+	@echo "==================== unit tests ======================" 2>&1;
+# the below commands need to be one shell invocation for the early exit to work;
+# note: need to store the paths into separate var as it here as it's lazily
+# evaluated on each use :), using ?= for the assignment does not help for
+# some reason
```
matejmatuska: These have to be unindented, otherwise they are treated as bash not make comments and get echoed by make :/. Another solution is to prefix all of them with the '@'.

PeterMocary (commented):
The issue I mentioned in my above comment was fixed.

I tried the case where one wants to test an actor that exists only in one of the upgrade path repositories. The tests are now executed only for the repository which has the specified actor while also informing about the skipped unit tests run when actor could not be found in the repository.

Running the ``TEST_CONTAINER=el9 ACTOR=sssd_facts make test_container_no_lint``
resulted in skipped unit tests for el8toel9 repository while for the el9toel10 the tests were run as expected because the actor sssd_facts exists there. The makefile did not report error at the end.
```
==================== unit tests ======================
Failed to find the sssd_facts actor in the common,el8toel9 repositories: ERROR:__read_error_messages_above_this_one_on_stderr__
Skipping unit tests, could not find the 'sssd_facts' actor in common,el8toel9 repositories
```
I've also tried it with a common actor, an actor that is only present on el8toel9 repository and with the test container el8. Everything worked as expected.

LGTM!

Makefile
pirat89: btw, if you use spaces instead of tabs, it will be also processed as comment. but rather do it this way, as otherwise it could happen that someone by mistake add spaces instead of tabs to a real code/recipe.

matejmatuska: I didn't know that, but it makes sense. I will leave it as it is.

PR #1404, opened 2025-07-10
docs: Add inhibiting tutorial and PR workflow guidelines

matejmatuska:


MichalHe (changes requested):
I have added some clarity suggestions and typos. Do not let my review discourage you, it is a solid and valuable piece of work.

PR #1405, opened 2025-07-11
Fix: skip checking ownership of files in .../directory-hash/ dir

matejmatuska:
This check has been reintroduced in 87013d25b5aa3 in #1297, however the "root" directory during traversal is:
`/var/lib/leapp/el10userspace/etc/pki/ca-trust/extracted/pem/directory-hash`. 
The skip condition looks for '/directory-hash/' which is false and the files are still checked.

I also added a test, which is not nice, but is only designed to ensure the existing behaviour keeps working. It also doesn't cover all paths.

From pdb:
```
> /etc/leapp/repos.d/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py(315)_get_files_owned_by_rpms()
-> if '/directory-hash/' in root:
display root: '/var/lib/leapp/el10userspace/etc/pki/ca-trust/extracted/pem/directory-hash'
display files: ['5931b5bc.0', 'ACCVRAIZ1.pem', 'a94d09e5.0', '3c9a4d3b.0', '228f89db.0', 'AC_RAIZ_FNMT-RCM.pem', 'cd8c0d63.0', 'b936d1c6.0', ... ]
```

On a minimal install this saves around 450 rpm ownership checks:
Before the change:
```
# grep directory-hash /var/log/leapp/leapp-preupgrade.log | grep owned | wc -l
445
```
after:
```
# grep directory-hash /var/log/leapp/leapp-preupgrade.log | grep owned | wc -l
0
```
and:
```
# grep directory-hash /var/log/leapp/leapp-upgrade.log | grep SKIP
2025-07-11 13:22:08.879 DEBUG    PID: 243344 leapp.workflow.TargetTransactionFactsCollection.target_userspace_creator: SKIP files in the /var/lib/leapp/el10userspace/etc/pki/ca-trust/extracted/pem/directory-hash directory: Not important for the IPU.
```

### DIFF
--- a/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
+++ b/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
@@ -294,6 +294,8 @@ def _get_files_owned_by_rpms(context, dirpath, pkgs=None, recursive=False):
     """
     Return the list of file names inside dirpath owned by RPMs.
 
+    The returned paths are relative to the dirpath.
+
     This is important e.g. in case of RHUI which installs specific repo files
     in the yum.repos.d directory.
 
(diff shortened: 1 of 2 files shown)


matejmatuska:
Might add more tests later, covering at least the nonrecursive path. The path where `pkgs` are set is unused.

PR #1406, opened 2025-07-15
Adjust DNF stream variable

karolinku:
CentOS Stream systems the `stream` DNF variable is used instead of the `releasever`, which cause an issue when creating target repos, becuase the $stream variable is handled incorrectly. This patch is adjusting the content of /etc/dnf/vars/stream in scratch container, so $stream var is replaced with correct target version of system.

Jira: RHEL-95982

### DIFF
--- a/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
+++ b/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
@@ -1091,6 +1091,26 @@ def _install_custom_repofiles(context, custom_repofiles):
         context.copy_to(rfile.file, _dst_path)
 
 
+def adjust_dnf_stream_variable(context, varfile='/etc/dnf/vars/stream'):
+    """
+    Adjust the version in the dnf 'stream' variable to the target version.
+
+    URLs in CentOS Stream repofiles contain the $stream variable which,
+    if not adjusted, retains the value from the source system making
+    the URLs point to repos for the source version. This function adjusts
+    the variable so that the URLs point to the target version repos.
+    """
+
+    target_version = get_target_major_version()
+    try:
+        with context.open(varfile, 'w') as f:
+            f.write(target_version + '-stream\n')
+    except (FileNotFoundError, OSError) as e:
+        raise StopActorExecutionError(
+            message='Failed to adjust dnf variable in {} to "{}".'.format(varfile, target_version + '-stream'),
+            details={'details': str(e)})
+
+
 def _gather_target_repositories(context, indata, prod_cert_path):
     """
     This is wrapper function to gather the target repoids.
@@ -1108,6 +1128,9 @@ def _gather_target_repositories(context, indata, prod_cert_path):
     """
     rhsm.set_container_mode(context)
     rhsm.switch_certificate(context, indata.rhsm_info, prod_cert_path)
+
+    if api.current_actor().configuration.os_release.release_id == 'centos':
+        adjust_dnf_stream_variable(context)
 
     _install_custom_repofiles(context, indata.custom_repofiles)
     return gather_target_repositories(context, indata)
--- a/repos/system_upgrade/common/actors/targetuserspacecreator/tests/unit_test_targetuserspacecreator.py
+++ b/repos/system_upgrade/common/actors/targetuserspacecreator/tests/unit_test_targetuserspacecreator.py
@@ -3,6 +3,7 @@ from __future__ import division, print_function
 import os
 import subprocess
 import sys
+import tempfile
 from collections import namedtuple
 
 import pytest
@@ -39,6 +40,9 @@ class MockedMountingBase(object):
     def __init__(self, **dummy_kwargs):
         self.called_copytree_from = []
         self.target = ''
+
+    def open(self, fullpath, *args, **kwargs):
+        return open(self, fullpath, *args, **kwargs)
 
     def copytree_from(self, src, dst):
         self.called_copytree_from.append((src, dst))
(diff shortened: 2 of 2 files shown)


repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -1075,6 +1075,31 @@ def _install_custom_repofiles(context, custom_repofiles):
...
+
+    varfile = '/etc/dnf/vars/stream'
+    target_version = get_target_major_version()
+    try:
+        with context.open(varfile, 'w') as f:
+            f.write(target_version + '-stream')
+    except FileNotFoundError as e:
+        raise StopActorExecutionError(
+            message='File {varfile} not found. Error: {e}'
+            .format(varfile=varfile, e=e))
```
pirat89, line 1096: put please details about the error into the details of the exception.

matejmatuska (changes requested):
The code is generally fine, there are just some small comments.
There are some errors in the tests that need to be addressed though.

There is a small typo in the commit message (becuase).

repos/system_upgrade/common/actors/targetuserspacecreator/tests/unit_test_targetuserspacecreator.py
```diff
@@ -1325,3 +1329,27 @@ def has_dbgmsg(substr):
...
+    monkeypatch.setattr(userspacegen.api, 'current_actor', CurrentActorMocked())
+    monkeypatch.setattr(userspacegen, 'get_target_major_version', lambda: '10')
+
+    with tempfile.NamedTemporaryFile(mode='w+') as tmpf:
+        tmpf.write('10-stream\n')
+        tmpf.seek(0)
+        content = tmpf.read()
+
+    userspacegen.adjust_dnf_stream_variable(MockedMountingBase, tmpf.name)
+    assert content == '10-stream\n'
```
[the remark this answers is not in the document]
karolinku: yes, you are right, it will better indicates file adjustment if first put there incorrect content. 
also, flush() is needed here to save file content immediately.

repos/system_upgrade/common/actors/targetuserspacecreator/tests/unit_test_targetuserspacecreator.py
```diff
@@ -1325,3 +1329,27 @@ def has_dbgmsg(substr):
...
+    assert content == '10-stream'
+
+
+def test_failing_stream_varfile_write(monkeypatch):
+    monkeypatch.setattr(userspacegen.api, 'current_actor', CurrentActorMocked())
+    monkeypatch.setattr(userspacegen, 'get_target_major_version', lambda: '10')
+
+    with pytest.raises(StopActorExecutionError) as err:
+        userspacegen.adjust_dnf_stream_variable(MockedMountingBase, '/path/not/exists')
+    assert 'Failed to adjust dnf variable' in str(err.value)
```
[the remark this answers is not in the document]
matejmatuska: Sorry, you are right, seems like both are valid with this context manager.

repos/system_upgrade/common/actors/targetuserspacecreator/tests/unit_test_targetuserspacecreator.py
```diff
@@ -1325,3 +1329,60 @@ def has_dbgmsg(substr):
...
+    def mock_gather_target_repositories(MockedMountingBase, testInData):
+        pass
+
+    monkeypatch.setattr(userspacegen.api, 'current_actor', CurrentActorMocked(release_id=distro))
+    monkeypatch.setattr(userspacegen, 'get_target_major_version', lambda: '10')
+    monkeypatch.setattr(rhsm, 'set_container_mode', mock_set_container_mode)
+    monkeypatch.setattr(rhsm, 'switch_certificate', mock_switch_certificate)
+    monkeypatch.setattr(userspacegen, '_install_custom_repofiles', mock__install_custom_repofiles)
+    monkeypatch.setattr(userspacegen, 'adjust_dnf_stream_variable', mock_adjust_stream_variable)
+    monkeypatch.setattr(userspacegen, 'gather_target_repositories', mock_gather_target_repositories)
```
matejmatuska, line 1383: wrt the `pass` functions, they are alright, though you can make this a lot shorter using a single function. If you want to keep it at it is, I would at least ask to use the real parameter names instead of `testInData`.

wrt the `mock_adjust_stream_variable`, I would assert whether the varfile is indeed /etc/dnf/vars/stream.

```suggestion
    def do_nothing(*args, **kwargs):
        pass

    def mock_adjust_stream_variable(context, varfile='/etc/dnf/vars/stream'):
        assert varfile = '/etc/dnf/vars/stream'
        nonlocal adjust_called
        adjust_called = True

    monkeypatch.setattr(userspacegen.api, 'current_actor', CurrentActorMocked(release_id=distro))
    monkeypatch.setattr(userspacegen, 'get_target_major_version', lambda: '10')
    monkeypatch.setattr(rhsm, 'set_container_mode', do_nothing)
    monkeypatch.setattr(rhsm, 'switch_certificate', do_nothing)
    monkeypatch.setattr(userspacegen, '_install_custom_repofiles', do_nothing)
    monkeypatch.setattr(userspacegen, 'adjust_dnf_stream_variable', mock_adjust_stream_variable)
    monkeypatch.setattr(userspacegen, 'gather_target_repositories', do_nothing)
```

matejmatuska (changes requested):
Just some suggestion, but I would really put the assert varfile == ... there.

repos/system_upgrade/common/actors/targetuserspacecreator/tests/unit_test_targetuserspacecreator.py
karolinku: I removed assertion  `assert varfile = '/etc/dnf/vars/stream'` from mock_adjust_stream_variable because I make this value default. I'm not sure if assert is needed here, however it doesn't harm

matejmatuska: Yeah, right now it shouldn't fail unless a non-default incorrect argument is passed, but in the future the code may get refactored. It doesn't do any harm to keep there.

MichalHe:
I have tested the path on CentoOS 9 > 10 upgrade. Without this patch, the upgrade incorrectly downloads `el9` content. With this patch, the upgrade correctly downloads `el10` content. Code-wise, the patch look good.

LGTM

PR #1407, opened 2025-07-15
Skip RHSM related actions on non-RHEL systems

matejmatuska:
Non-RHEL systems do not use Red Hat Subscription Manager (RHSM),
therefore all RHSM related actions should be skipped on such systems.

Changes:
- The --no-rhsm option is implied on non-RHEL systems and LEAPP_NO_RHSM is
forcefully set to 1.
- On non-RHEL distros skip rhsm actions which are not skipped on RHEL even with --no-rhsm.
- Do not report unhandled RHSM release on non-RHEL systems

There are still reports that need to be adjusted, this is just "functional" changes.

Jira: RHEL-95975

### DIFF
--- a/commands/command_utils.py
+++ b/commands/command_utils.py
@@ -136,6 +136,16 @@ def get_os_release_version_id(filepath):
     return _retrieve_os_release_contents(_os_release_path=filepath).get('VERSION_ID', '')
 
 
+def get_distro_id():
+    """
+    Retrieve the OS release ID from /etc/os-release.
+
+    :return: The OS release ID from /etc/os-release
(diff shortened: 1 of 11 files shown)


pirat89:
linters are unhappy :crying_cat_face: 
but it looks good otherwise :)

matejmatuska:
> linters are unhappy 😿 but it looks good otherwise :)

@pirat89 They are happy now, I kept the commits separate, the last one in particular. But I can squash it or feel free to squash it during a merge.

MichalHe (approved):

PR #1408, opened 2025-07-18
Remove obsolete CS RPM GPG key during IPU 9to10

matejmatuska:
Similarly to RHEL upgrades, we want to remove obsolete GPG keys, e.g. keys using SHA1. On CS9 (and also CS8) the gpg-pubkey-8483c65d-5ccc5b19 key imported from /etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial is used. This key uses SHA1 and makes CS9 -> CS10 IPUs fail due to SHA1 being prohibited.

This patch adds the key to the list of keys to remove during IPU.

NOTE: The key used on CS10 is
/etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial-SHA256 which uses SHA256, but after importing it it's also named gpg-pubkey-8483c65d-5ccc5b19. This isn't a problem because the SHA1 one is removed before the SHA256 one is imported, but it's something to keep in mind if there is CS10 -> CS11 sometime in he future.


matejmatuska:
This should be pretty easy to test, but other patches, most importantly #1406, but possibly also #1407, are needed to hit the issue.

Without this patch:
```
error: Verifying a signature using certificate 99DB70FAE1D7CE227FB6488205B555B38483C65D (CentOS (CentOS Official Signing Key) <security@centos.org>):
              1. Certificate 05B555B38483C65D invalid: policy violation
                  because: No binding signature at time 2025-06-05T19:03:38Z
                  because: Policy rejected non-revocation signature (PositiveCertification) requiring second pre-image resistance
                  because: SHA1 is not considered secure
              2. Certificate 05B555B38483C65D invalid: policy violation
                  because: No binding signature at time 2025-07-16T09:07:59Z
                  because: Policy rejected non-revocation signature (PositiveCertification) requiring second pre-image resistance
                  because: SHA1 is not considered secure
            error: Verifying a signature using certificate 99DB70FAE1D7CE227FB6488205B555B38483C65D (CentOS (CentOS Official Signing Key) <security@centos.org>):

...
a ton of these warnings 
...
            Public key for wget-1.24.5-5.el10.x86_64.rpm is not trusted. Failing package is: wget-1.24.5-5.el10.x86_64
             GPG Keys are configured as: file:///etc/pki/rpm-gpg/RPM-GPG-KEY-centosofficial-SHA256
            Error: GPG check FAILED
```

With this patch the errors should be gone.

pirat89:
corrected jira ticket in description :)

PR #1410, opened 2025-08-04
Add fapolicy rule for LEAPP during package installation

karolinku:
When fapolicy is running on system, the upgrade fails with error related to not sufficient permissions to /var/ilb/leapp/. To solve this issue, there needs to be added new rule to fapolicyd, so this change creates new rule and restart the service after any Leapp' packages transactions.

Jira: RHEL-50847

### DIFF
--- a/packaging/leapp-repository.spec
+++ b/packaging/leapp-repository.spec
@@ -258,6 +258,11 @@ install -m 0755 -d %{buildroot}%{leapp_python_sitelib}/leapp/cli/
 cp -r commands %{buildroot}%{leapp_python_sitelib}/leapp/cli/
 rm -rf %{buildroot}%{leapp_python_sitelib}/leapp/cli/commands/tests
 
+# install rules necessary for fapolicy
+mkdir -p %{buildroot}/etc/fapolicyd/rules.d/
+echo "allow perm=any all : dir=/var/lib/leapp/" > %{buildroot}/etc/fapolicyd/rules.d/31-fapolicy.rules
+chmod 644 %{buildroot}/etc/fapolicyd/rules.d/31-fapolicy.rules
+
 # Remove irrelevant repositories - We don't want to ship them for the particular
 # RHEL version
 for i in el7toel8 el8toel9 el9toel10;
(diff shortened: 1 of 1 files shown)


karolinku:
Approach presented in first commit is facing the ownership issue around rule file. If LEAPP creates /etc/fapolicyd/rules.d/31-fapolicy.rules first, it claims the ownership to the localization, so later when trying to install fapolicyd on system, it creates conflict error. On the other hand, if claiming only exact /etc/fapolicyd/rules.d/31-fapolicy.rules first in %files, the directory structure persists after uninstallation of LEAPP package (but not the rule itself). Both this issues are ommited with %post and %postun scriplets.

pirat89 (changes requested):
Current approach is not good by design. The original was closer to expected solution. See the comments for more details.

pirat89:
I see the point about the file conflict. We discussed that with @karolinku  over slack, writing it here also for others. The file conflict for dirs is happening as /etc/fapolicyd and /etc/fapolicyd/rules.d directories have a different attributes than we set in the SPEC file. it means, we would need to set the same access right (chmod), ownership, ... so directories between both rpms are identical. However here is a problem that the ownership of dirs is in this case `root:fapolicyd`, which means that we would have to create also the `fapolicyd` group, etc.. which does not seem right to me.

I am discussing with fapolicyd what is proper guideline for the packaging. We have 2 options at this moment from my POV:
* a) create the rules file without owning parent directories on the path; it's not good and most likely we will need to change it in future, but it would be a possibility
* b) create subpackage `leapp-upgrade-*-fapolicyd` which will require `fapolicyd`, set posttrans scriptlet and deliver the rules file; AND set [rich dependency](https://rpm-software-management.github.io/rpm/manual/boolean_dependencies.html) in `leapp-upgrade-elXtoelY` on the subpackage so it's required only when fapolicyd is installed. Like:
```
Requires: (%{lpr_name}-fapolicyd = %{version}-%{release} if fapolicyd)
```  
Note that we will need to extend the list of leapp-repository packages in the `rpms.py` library.

I am in favor of  `b)` as it's quite easy to do it and it looks as a quite clean solution.

Update: discussing it with fapolicyd guys, **b)** looks at this moment as a good approach from our side. So let's go this way.

pirat89:
The accessrights of the config file & ownership should be set as below:
```
-rw-r--r--. 1 root fapolicyd 
```

pirat89 (approved):
tested manually it works; the rebase did not change the content of this PR so no need to wait for the final test results

PR #1411, opened 2025-08-06
Update upgrade path: Add 9.7-10.1

karolinku:
Adding upgrade path 9.7 -> 10.1 for:
- RHEL
- RHEL with SAP HANA
- CentOS
- AlmaLinux

Jira: RHEL-107228


repos/system_upgrade/common/files/upgrade_paths.json
```diff
@@ -14,6 +15,7 @@
             "8.10": ["9.6", "9.4"],
             "8": ["9.6", "9.4"],
             "9.6": ["10.0"],
+            "9.7": ["10.1"],
```
pirat89, line 18: @karolinku that's incorrect. saphana is only and only for even releases. so saphnana should not be modified anyhow at this moment.

repos/system_upgrade/common/files/upgrade_paths.json
```diff
@@ -25,13 +27,15 @@
         "_virtual_versions": {
             "8": "8.10",
             "9": "9.6",
+            "9.7": ["10.1"],
```
pirat89, line 30: this is also invalid. check virtual versions around. virtual version != upgrade path. It's virtual adding of a minor version to centos major version.

repos/system_upgrade/common/files/upgrade_paths.json
```diff
@@ -25,13 +27,15 @@
...
             "10": "10.0"
```
pirat89, line 31: here should be 10.1

pirat89 (changes requested):
3 changes required. Also update please the commit message. CS does not have minor releases, so adding IPU Path 9.7 -> 10.1 is not for CS. You can mention that you are changing the virtual version for centos streams to 9.7 and 10.1

PR #1413, opened 2025-08-06
feat(fast_lint): execute isort

MichalHe:
Execute isort when running the `fast_lint` target. Therefore, the target now matches the `lint` target, allowing the developer to see the results of running all of our linters on his/hers changes in a quick manner.


pirat89 (approved):
lgtm and make sense.

PR #1415, opened 2025-08-06
Enable new default IPU path 8.10 -> 9.7

pirat89:
Adding the upgrade path 8.10 -> 9.7 for RHEL and Almalinux distros, making this upgrade path default now.

Jira: RHEL-107228


matejmatuska (approved):
LGTM, looks alright and I also did a sanity upgrade on RHEL 8.10 and RHEL 9.7 is correctly chosen as a target, whereas without the patch it was 9.6.

PR #1426, opened 2025-08-21
Update CurrentActorMocked to 8->9 upgrade path

matejmatuska:
Currently, the default upgrade path is 7.8->8.1 which has been dropped from the repo. Several tests make assumptions about these default and in some cases this means that code path for other general upgrade paths (8->9, 9->10) are left untested e.g. `modify_userspace_for_livemode`.

This patch changes the default path to 8.10->9.6.

Jira: RHELMISC-16538

### DIFF
--- a/repos/system_upgrade/common/actors/repositoriesblacklist/libraries/repositoriesblacklist.py
+++ b/repos/system_upgrade/common/actors/repositoriesblacklist/libraries/repositoriesblacklist.py
@@ -6,7 +6,6 @@ from leapp.models import CustomTargetRepository, RepositoriesBlacklisted, Reposi
 
 # {OS_MAJOR_VERSION: PESID}
 UNSUPPORTED_PESIDS = {
-    "7": "rhel7-optional",
     "8": "rhel8-CRB",
     "9": "rhel9-CRB",
     "10": "rhel10-CRB"
@@ -28,9 +27,8 @@ def _report_using_unsupported_repos(repos):
 
 
 def _report_excluded_repos(repos):
-    optional_repository_name = 'optional' if get_source_major_version() == '7' else 'CRB'
     api.current_logger().info(
-        "The {0} repository is not enabled. Excluding {1} from the upgrade".format(optional_repository_name, repos)
+        "The CRB repository is not enabled. Excluding {} from the upgrade".format(repos)
     )
 
     report = [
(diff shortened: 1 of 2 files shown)


MichalHe (changes requested):
The default version values for `CurrentActorMocked` have been updated, and appropriate changes to tests were done.

**Requested changes:** I suggest to squash the commit that updates `CurrentActorMocked` with updates to the tests. However, refactoring commits (such as changes performed to `scancpu.py`) should be kept as a separate commits. After that, I shall approve.

**Some feedback** In some cases, such as tests for the PesEventScanner actor, there was fundamentally no need to update the code. The code tests the logic of the actor, which should behave the same regardless of the source/version target pair, i.e., the test fulfills its function even if we manually set source/target version to `7.9`/`8.10`. So I would just change how `CurrentActorMocked` is initialized, and add a comment that the test is "upgrade path agnostic".

MichalHe:
There are couple of tests that still use `src_ver='7.*'`. However, these tests check whether some core algorithm (that is reused in every upgrade path) works as expected. In other words, it makes no sense to update the `src_ver` since the algorithm should behave the same regardless of whether we are upgrading `7>8` or `8>9` -- we only care that the core logic is correct.

WDYT about squashing commits that only update tests? (i.e., commits `65fc397eed10188c90523d7f11755289668e2047, 1f03f5b63568c5d2304303594363f13fed658b9d, 7ef525f06823f9d35b5634919e030ed23ce53cb2, b90fcdd99e4720616fd326117e73317e3eac4cf0, d51eaeca35020401b1018ecfce2886f6cd30322c`)?

LGTM!

matejmatuska:
> There are couple of tests that still use `src_ver='7.*'`. However, these tests check whether some core algorithm (that is reused in every upgrade path) works as expected. In other words, it makes no sense to update the `src_ver` since the algorithm should behave the same regardless of whether we are upgrading `7>8` or `8>9` -- we only care that the core logic is correct.
> 
> WDYT about squashing commits that only update tests? (i.e., commits `65fc397eed10188c90523d7f11755289668e2047, 1f03f5b63568c5d2304303594363f13fed658b9d, 7ef525f06823f9d35b5634919e030ed23ce53cb2, b90fcdd99e4720616fd326117e73317e3eac4cf0, d51eaeca35020401b1018ecfce2886f6cd30322c`)?
> 
> LGTM!

@MichalHe I squashed a little more of them, basically all that only contained test updates and *minor* refactors like renaming. 
The first commit contains a list of all of them. 
The bigger refactors are each in their own commit.

Also resolved the conflict.

MichalHe (approved):

PR #1431, opened 2025-09-15
chore(RHINENG-19596): Rebrand Insights to Lightspeed

mhuth:


pirat89 (approved):

repos/system_upgrade/common/actors/checkinsightsautoregister/libraries/checkinsightsautoregister.py
```diff
@@ -24,9 +24,9 @@ def _ensure_package(package):
 def _report_registration_info(installing_client):
     pkg_msg = " The '{}' package required for the registration will be installed during the upgrade."
 
-    title = "Automatic registration into Red Hat Insights"
+    title = "Automatic registration into Red Hat Lightspeed"
```
pirat89, line 27: @mhuth I just realized the modified title of the report which leads to a different key of the report. I suggest to add few lines below following line to ensure the report key stays as it is:
```
        reporting.Key('693963253195f418526f045b6d630a1f4c7a193d')
```

pirat89 (changes requested):

repos/system_upgrade/common/actors/checkinsightsautoregister/libraries/checkinsightsautoregister.py
mhuth: Done!

pirat89 (approved):
lgtm. thanks!

PR #1434, opened 2025-10-04
packit: refactor actions and jobs to remove duplication

tomasfratrik:
Refactored the Packit config to use a shared default-actions block instead 
of repeating the same steps in each job.

Noticed that some 9to10 individual tests are still tagged as 8to9. not sure 
if that's intentional.


pirat89:
For the reference, here is ticket for the missing documentation in packit https://github.com/packit/packit.dev/issues/443

PR #1443, opened 2025-10-24
Fix unsupported source version crash

PeterMocary:
This PR contains the solution to the root cause of the problem from ticket RHEL-120252 in the first commit. The second commit (now removed) proposes an improvement to the fallback mechanism described below. Lets decide if this improvement should be included or not. If it is not included, the upgrade paths map should be updated for the Alma Linux with the original fallback definitions same as for RHEL (now added as second commit).

I've tried to run preupgrade with the following changes on Alma Linux 9.6 and RHEL 9.6.

### Commit#1: fix unsupported source version crash

This patch fixes a crash when executing upgrade on a system with version
that is not defined in the supported upgrade paths map and when the
fallback mechanism for this particular case fails as well. The upgrade
process is now terminated with an error instead of a trace back.

### Commit#2: define fallback entries in the upgrade paths map for AlmaLinux 

The upgrade paths map contains fallback entries which make sure a target
upgrade version is found regardless of the source minor version. This patch
defines fallback entries for AlmaLinux.

<s>
<h3>Commit#2: improve fallback mechanism for source version matching</h3>

When searching in the upgrade paths map by the system version, we had
implemented a fallback that allowed the upgrade process to find a target
version even if the minor version of the system did not match exactly
with the ones defined in the map.

The fallback mechanism was implemented using additional definitions in
the map which did not have a minor version explicitly defined and thus
would be only used as a fallback when the system version did not match
any minor version.

This patch changes this fallback mechanism to instead search in the map
by the latest minor version of the system's major version. The
additional records are removed from the map as well.
</s>

Jira: RHEL-120252


PeterMocary:
Based on discussion with the team, we decided to not include the improvement at this time. I dropped the improvement commit and defined the original fallback mechanism for the Alma linux upgrade paths as well. This makes sure the error won't be hit for any distribution and the unsupported source version will result into an inhibitor informing the user about unsupported upgrade path.

PeterMocary:
Rebased.

matejmatuska (dismissed):
Tested, works as expected, when the fallback paths are not defined for `almalinux`, the upgrade raises the error. When they are, the upgrade is inhibited due to unsupported version.

However please edit the commit message to include that almalinux paths have changed. The other suggestions are just nitpicks.

I can approve then, but probably not merge as we want to prioritize #1438 and merging this would create conflicts.

PeterMocary:
I've separated the upgrade paths map changes into its own commit to distinguish them from the traceback crash fix.

pirat89 (changes requested):
lgtm. just blocking the merge until #1438  is merged due to conflicts between PRs and #1438 is much more complex. The failed tests are caused by internal issue.

pirat89 (approved):
lgtm and **I believe it will work** (I haven't tested it manually 🙈 )

PR #1444, opened 2025-10-30
mount_unit_gen: bind mount /sysroot/boot to /boot

MichalHe:
Our changes towards using systemd-fstab-generator in the upgrade initramfs caused that we are mounting almost all partitions, including /boot (the actual mount target is /sysroot/boot) early in the boot process. When upgrading with FIPS, the dracut fips module tries to mount the device where the boot partition resides to check the integrity of the kernel, however, it fails as the boot block device is already mounted by us. This patch therefore introduces a static unit that bindmounts What=/sysroot/boot to Where=/boot, making the contents of /boot available to the fips module. Thefore, upgrades with FIPS enabled should be again possible.

Jira-ref: RHEL-123886


karolinku:
tested manually, works good. other than one comment, lgtm

karolinku (approved):

PR #1448, opened 2025-11-13
removeobsoletegpgkeys: Adjust for conversions

matejmatuska:
When doing upgrade + conversion, removing obsolete GPG from the current
(or target for that matter) distro doesn't make sense, because we are
moving to a different distro.
Instead, all distro provided keys from the source distro need to be
removed, as the target distro uses it's own keys. Those are imported
elsewhere later during the upgrade.

A new list is added to the gpg-signatures.json maps, which contains the
names of the fake RPMs "generated" upon importing a GPG key into the RPM
DB.
These are in the order the key IDs ("keys" in the map) are in, however
the mapping is not always 1:1 (e.g. the Centos SIG Extras keys).
Some are also missing, see TODO below.

The key id could be mapped to the RPM names, however since the RPM NVR
format is:
`gpg-pubkey-<last 8 chars from key ID>-<creation time of the signature packet>`
there could be a collision between the key IDs.

I addded annotations to the keys at:
https://github.com/oamg/leapp-repository/wiki/gpg%E2%80%90signatures.json-key-annotations.

Jira: RHEL-110190


repos/system_upgrade/common/files/distro/centos/gpg-signatures.json
```diff
@@ -1,8 +1,14 @@
 {
     "keys": [
         "24c6a8a7f4a80eb5",
+        "4eb84e71f2ee9d55",
```
matejmatuska: I will probably still add these to key-rpms, but I have yet to find the files somewhere.

matejmatuska:
Hello @yuravk, in this PR I extended the RPM GPG keys config for all distros, including Alma Linux, with a list of names of the fake RPMs which are created when the keys are imported to RPM DB (gpg-pubkey-XXXXXX-XXXXXX) to complement the existing list of fingerprints.

However, I am not sure what keys some of the fingerprints for Alma Linux correspond to.
If you can point me to the GPG key files (e.g. in AL compose), I could include them in this PR.

Specifically I am interested in these:
- `51d6647ec21ad6ea`
- `429785e181b961a5`
- `d07bf2a08d50eb66`

yuravk:
> Hello @yuravk, in this PR I extended the RPM GPG keys config for all distros, including Alma Linux, with a list of names of the fake RPMs which are created when the keys are imported to RPM DB (gpg-pubkey-XXXXXX-XXXXXX) to complement the existing list of fingerprints.
> 
> However, I am not sure what keys some of the fingerprints for Alma Linux correspond to. If you can point me to the GPG key files (e.g. in AL compose), I could include them in this PR.
> 
> Specifically I am interested in these:
> 
> * `51d6647ec21ad6ea`
> * `429785e181b961a5`
> * `d07bf2a08d50eb66`

Hello @matejmatuska 

I believe they are:
* `429785e181b961a5` - https://repo.almalinux.org/elevate/RPM-GPG-KEY-ELevate (AlmaLinux ELevate key)
* `51d6647ec21ad6ea` - https://repo.almalinux.org/almalinux/RPM-GPG-KEY-AlmaLinux-legacy (AlmaLinux 8 old key)
* `d07bf2a08d50eb66` - https://repo.tuxcare.com/tuxcare/RPM-GPG-KEY-TuxCare (cloudlinux TuxCare key)

pirat89 (changes requested):
It looks good. there are just some stylistic problems `'` x `"`, one code simplification, and forgotten update of actor description. it looks good otherwise. I haven't tested it yet. Keeping it for the final rewview.

pirat89 (approved):
lgtm and seems to be working - tested manually for CS -> CS , CS -> RHEL (I forgot to make a snapshot of msgs from testing)

PR #1453, opened 2025-11-20
refactor handling of multipath during upgrade

MichalHe:
handle multipath devices in upgrade initramfs

Previously, multipath configs were scanned (and optionally) modified
only for the 8>9 upgrade paths. Known backward-compatibility issues were
fixed in the ApplicationsPhase, but any support for mounting multipath
devices was lacking. This patch refactors the handling of multipath so
that:
1) all multipath-related configs are copied into target uspace
2) any changes done in the ApplicationsPhase are performed to these
   copies instead
3) our modified multipath configs are included into the initramfs
4) dracut 'multipath' module is enabled when building upgrade initramfs
5) system configs are patched so that in the end they will look the same
   as if modified in the ApplicationPhase. This is achieved by simply
   copying our modified configs and replacing the old system ones.

These changes avoid reaching into the config-parsing code. An ugly
consequence of this is that some models (MultipathConfFacts8to9) needed
to be placed into the common repository, although they are only used for
8>9 upgrades. To address this ugliness would require designing and
producing a message that would contain the entire contents of the
primary multipath config (if we want to avoid parsing it twice).

jira-ref: RHEL-14712, RHEL-124306 (dev task)


MichalHe:
I have performed preliminary testing on RHEL 9 > RHEL 10. Multipath devices are present in the upgrade initramfs, however, we need to also enable multipathd.service so that in case of failover we can transition to multipath siblings seemlessly.

MichalHe:
I have tested the patch on a setup that uses a separate multipath device mounted at `/var`. The device did not have LVM on top. I found a problem regarding our bindmouting of /boot - see #1457, which needs to be merged. Rebooting into our upgrade environment, multipath device nodes were created, and we were able to successfully mount the devices using our mount units.

The only problem is that `multipathd.service` started and then it was shut down, because of [this line](https://github.com/dracutdevs/dracut/blob/5d2bda46f4e75e85445ee4d3bd3f68bf966287b9/modules.d/90multipath/multipathd.service#L11) . Therefore, I suggest we programmatically remove the line from the `multipathd.service` in our upgrade environment (in a separate PR).

repos/system_upgrade/common/actors/multipath/config_reader/tests/test_multipath_conf_read_8to9.py
```diff
@@ -100,45 +104,36 @@ def test_parse_config():
...
+    monkeypatch.setattr(multipathconfread, 'is_processable', lambda: True)
 
-    facts = multipathconfread.get_multipath_conf_facts(os.path.join(TEST_DIR, 'empty_dir.conf'))
-    assert facts
-    assert len(facts.configs) == 1
-    assert_config(facts.configs[0], empty_dir_conf)
+    produce_mock = produce_mocked()
+    monkeypatch.setattr(api, 'produce', produce_mock)
 
+    actor_mock = CurrentActorMocked()
```
pirat89: Let's rather set the IPU path, so if there are any changes in future about defaults, the test is not negatively impacted.

```suggestion
    actor_mock = CurrentActorMocked(src_ver='8.10', dst_ver='9.6')
```

Note that there is not a test for 9 -> 10 checking that 8to9 msg is not produced. MultipathInfo should be produced in both cases.

repos/system_upgrade/common/models/multipath.py
```diff
@@ -0,0 +1,66 @@
...
+
+    config_dir = fields.Nullable(fields.String())
+    """ Value of config_dir in the defaults section. None if not set. """
+
+
+class UpdatedMultipathConfig(Model):
+    """ Information about multipath config that needed to be modified for the target system. """
+    topic = SystemInfoTopic
+
+    path = fields.String()
```
pirat89: @MichalHe  nitpick (not blocking), from my experience it's better to have one msg with a list inside then mutliple msgs with a simple string (it's more compact for reading). if we expect this msg could be produced by multiple actors in future, then having one msg per one file does not make such a big difference. otherwise it's usually better to have one msg with a list.

Note that emitting new msg per each single item is implemented in several actors already, so keeping the decision on you what you prefer. Just adding my experience from debugging that using lists is otherwise a better pattern from my pov.

pirat89 (changes requested):
I have not been able to finish the review completely yet, but I discovered some problems (one seems to be fatal) so I am submitting at least this partial review already to make you aware.

repos/system_upgrade/common/actors/multipath/config_reader/tests/test_multipath_conf_read_8to9.py
MichalHe: I have:
- made the tests explicit for the 8>9 upgrade path
- added a new test for the 9>10 upgrade patch that checks whether only `MultipathInfo` is produced

pirat89 (dismissed):
LGTM and the test results seems good. @MichalHe thanks for sharing them and extensive testing!

I just moved 2 actors from ChecksPhase to TargetTransactionChecksPhase to keep the earlier phase based just on leapp messages without interaction with the system. But we can start a discussion about that policy, to see whether we want to still protect this or not.

@MichalHe  feel free to squash commits in advance but I can use the squash method when merging, so no problem if you keep it as it is. I will keep tests to be finished and merge it then.

pirat89 (approved):
re-approve :D after fixing a bug I made by myself 🙈

PR #1457, opened 2025-11-27
boot: fix deps when bindmounting /boot to /sysroot/boot

MichalHe:
When /boot is a separate partition, we bindmount what=/sysroot/boot to /boot, so that we can perform necessary checks when booting with FIPS enabled. However, the current solution contains incorrect unit dependencies: it requires sysroot-boot.target. There is no such target, and the correct value should be sysroot-boot.mount. This patch corrects the dependencies.

Jira: RHEL-123886

### DIFF
--- a/repos/system_upgrade/common/actors/initramfs/mount_units_generator/files/bundled_units/boot.mount
+++ b/repos/system_upgrade/common/actors/initramfs/mount_units_generator/files/bundled_units/boot.mount
@@ -1,8 +1,8 @@
 [Unit]
 DefaultDependencies=no
 Before=local-fs.target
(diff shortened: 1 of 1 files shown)


matejmatuska (approved):
You are right.

pirat89:
@MichalHe  you are right. It's not described well. They found the race condition later, after the ticket was created, but forget to update it. Let's then keep it as part of the current solution and merge it. I will interlink it with the original ticket in the PR description and let's keep the rest as it is.

@matejmatuska yy. it's good to go

pirat89 (approved):

PR #1460, opened 2025-12-04
add handling for LVM configuration

PeterMocary:
The relevant user LVM configuration is now copied into the target userspace
container along with enabling LVM dracut module for upgrade initramfs creation.
The LVM configuration is copied into the target userspace container when
lvm2 package in installed. Based on the configuration, the devices file
is also copied in when present and enabled. The --nolvmconf option used when
executing dracut is changed into --lvmconf instead if the lvm configuration is
copied into the target userspace container.

Jira: RHEL-14712

### DIFF
--- a/repos/system_upgrade/common/actors/checklvm/libraries/checklvm.py
+++ b/repos/system_upgrade/common/actors/checklvm/libraries/checklvm.py
@@ -1,0 +1,74 @@
+import os
+
+from leapp import reporting
+from leapp.libraries.common.rpms import has_package
+from leapp.libraries.stdlib import api
+from leapp.models import (
+    CopyFile,
+    DistributionSignedRPM,
+    DracutModule,
+    LVMConfig,
+    TargetUserSpaceUpgradeTasks,
+    UpgradeInitramfsTasks
+)
+
+LVM_CONFIG_PATH = '/etc/lvm/lvm.conf'
+LVM_DEVICES_FILE_PATH_PREFIX = '/etc/lvm/devices'
+
+
+def _report_filter_detection():
+    title = 'LVM filter definition detected.'
+    summary = (
+        'Beginning with RHEL 9, LVM devices file is used by default to select devices used by '
+        f'LVM. Since leapp detected the use of LVM filter in the {LVM_CONFIG_PATH} configuration '
+        'file, the configuration won\'t be modified to use devices file during the upgrade and '
+        'the LVM filter will remain in use after the upgrade.'
+    )
+
+    remediation_hint = (
+        'While not required, switching to the LVM devices file from the LVM filter is possible '
+        'using the following command. The command uses the existing LVM filter to create the system.devices '
+        'file which is then used instead of the LVM filter. Before running the command, '
+        f'make sure that \'use_devicesfile=1\' is set in {LVM_CONFIG_PATH}.'
+    )
+    remediation_command = ['vgimportdevices']
+
+    reporting.create_report([
+        reporting.Title(title),
+        reporting.Summary(summary),
+        reporting.Remediation(hint=remediation_hint, commands=[remediation_command]),
+        reporting.ExternalLink(
+            title='Limiting LVM device visibility and usage',
+            url='https://red.ht/limiting-lvm-devices-visibility-and-usage',
+        ),
+        reporting.Severity(reporting.Severity.INFO),
+    ])
+
+
+def check_lvm():
+    if not has_package(DistributionSignedRPM, 'lvm2'):
+        return
+
+    lvm_config = next(api.consume(LVMConfig), None)
+    if not lvm_config:
+        return
+
+    lvm_devices_file_path = os.path.join(LVM_DEVICES_FILE_PATH_PREFIX, lvm_config.devices.devicesfile)
+    lvm_devices_file_exists = os.path.isfile(lvm_devices_file_path)
+
+    filters_used = not lvm_config.devices.use_devicesfile or not lvm_devices_file_exists
+    if filters_used:
+        _report_filter_detection()
+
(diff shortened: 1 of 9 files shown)


repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
```diff
@@ -363,20 +364,29 @@ def generate_initram_disk(context):
...
+
+    env_variables = ' '.join(env_variables)
+    env_variables = env_variables.format(
+        kernel_version=_get_target_kernel_version(context),
+        dracut_modules=fmt_module_list(initramfs_includes.dracut_modules),
+        kernel_modules=fmt_module_list(initramfs_includes.kernel_modules),
+        arch=api.current_actor().configuration.architecture,
+        files=' '.join(initramfs_includes.files)
+    )
+    cmd = os.path.join('/', INITRAM_GEN_SCRIPT_NAME)
```
PeterMocary: This is not very clean solution. I'd suggest expanding on the tasks mechanism with the addition of enabling environment variables (similarly to copying files for example). I didn't try to implement this yet, so it could be an improvement for later.

repos/system_upgrade/common/actors/scanlvmconfig/libraries/scanner.py
```diff
@@ -0,0 +1,56 @@
...
+    if not os.path.isfile(LVM_CONFIG_PATH):
+        api.current_logger().debug('The "{}" is not present on the system.'.format(LVM_CONFIG_PATH))
+        return
+
+    lvm_config_lines = _read_config_lines(LVM_CONFIG_PATH)
+    devices_section = _lvm_config_devices_parser(lvm_config_lines)
+
+    lvm_config_devices = LVMConfigDevicesSection(use_devicesfile=int(get_source_major_version()) > 8)
+    if 'filter' in devices_section:
+        lvm_config_devices.filter = re.findall(r'"([^"]*)"', devices_section['filter'])
```
MichalHe: I believe that `re.findall` might return `re.Match` objects instead of a list of strings.

MichalHe: Also, please consider whether the use of regular expression could be avoided (if the code would be simple). Otherwise, add a comment on what exactly is the regex supposed to match, so devs don't have to parse it every time they look at the code.

repos/system_upgrade/common/actors/scanlvmconfig/libraries/scanner.py
```diff
@@ -0,0 +1,56 @@
...
+
+    if 'devicesfile' in devices_section:
+        lvm_config_devices.devicesfile = devices_section['devicesfile']
```
MichalHe:
```suggestion
lvm_config_devices.devicesfile = devices_section.get('devicesfile')
```

MichalHe (commented):
Solution-wise good: nice, simple, clear. I will test it later, on a setup similar to the one described in the ticket.

repos/system_upgrade/common/actors/scanlvmconfig/libraries/scanner.py
```diff
@@ -0,0 +1,56 @@
...
+
+    if 'use_devicesfile' in devices_section and devices_section['use_devicesfile'] in ['0', '1']:
```
matejmatuska: I don't think this condition is necessary.
```suggestion
    if 'use_devicesfile' in devices_section:
```

matejmatuska (commented):
Just some minor things

repos/system_upgrade/common/models/lvmconfig.py
```diff
@@ -0,0 +1,33 @@
...
+    that the default value changed on the RHEL 9 to True.
+    """
+
+    devicesfile = fields.String(default="system.devices")
+    """
+    Defines the name of the devices file that should be used. The default devices
+    file is located in '/etc/lvm/devices/system.devices'.
+    """
+
+    filter = fields.List(fields.String(), default=['a|.*|'])
```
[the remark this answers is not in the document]
PeterMocary: Removed due to not being used at all.

repos/system_upgrade/common/actors/checklvm/libraries/checklvm.py
```diff
@@ -0,0 +1,74 @@
...
+LVM_CONFIG_PATH = '/etc/lvm/lvm.conf'
+LVM_DEVICES_FILE_PATH_PREFIX = '/etc/lvm/devices'
+
+
+def _report_filter_detection():
+    title = 'LVM filter definition detected.'
+    summary = (
+        'RHEL 9 and above uses the LVM devices file by default to select devices used by LVM. '
+        f'Since this system has LVM filter defined in the {LVM_CONFIG_PATH}, it will be '
+        'used after the upgrade as well.'
```
matejmatuska: My suggestion (emphasize that we wont modify the config):

'Beginning with RHEL 9, LVM devicesfile is used by default to select devices used by LVM. Since leapp detected the use of LVM filter in the {LVM_CONFIG_PATH} configuration file, the configuration won't be modified to use devicesfile during the upgrade and the LVM filter will remain in use.'

matejmatuska (commented):
Suggestions for the report.

PeterMocary:
Fixed unit tests.

MichalHe (approved):
I have performed manual testing that imply the patch is working as expected (lvm conf is present in the upgrade initramfs). E2E tests are passing (ignoring the on-demand ones), and the patch is good code-wise. Therefore, LGTM!

PR #1466, opened 2025-12-10
Replace EFI boot entry during conversion

matejmatuska:
After an upgrade+conversion we want the EFI boot entry to have the name of the target OS (e.g. Red Hat Enterprise Linux, when converting from CS->RHEL). The source distro boot entry and also EFI directory (e.g. `/boot/efi/EFI/centos/`) should be removed.

This PR:
- splits out the functions from `common.grub` library that are not related to grub to 2 new libraries:
  - `common.efi` - functions related to UEFI; functions raise `EFIError` exception on errors instead
  - `common.partitons` (name suggestions are) welcome - operations on disks/parititions; this is used by both efi and grub lib; raise `StorageScanError` on errors
- functions from the `common.grub` library that are replaced by new libraries have been marked deprecated  
- refactor existing efi functions in the addarmbootloaderworkaround and removeupgradebootentry actors into the new efi lib
- adds a new actor that:
  - Removes the EFI dir of the source distro (e.g. `/boot/efi/EFI/centos/`).
    - If not empty create a post upgrade report to handle the leftover files manually.
  - Removes the EFI boot entry of the source distro.
  - Adds a new EFI boot entry for the target distro.
    - The contents of the target distro EFI dir are brought in by the grub-efi-x64 and shim-x64 packages.
- adds another actor that inhibits conversions on systems with secure boot enabled
  - At least for now, I am going with an inhibitor as this current solution doesn't work with secureboot enabled. The problem probably is, in my understanding, that the shim and grub only contain keys for verifying signatures of the source distro. Our upgrade boot entry uses the target distros kernel, which is signed by the target distro. And so the shim/grub of the source distro rejects it.
  - **A possible solution** IMO is to create an EFI boot entry for the upgrade pointing to the target distro's shim, similarly to how it's done in the addarmbootloaderworkaround actor.

Jira: RHEL-133549

### TODO
- [x] ARM - should work
- [X] secure boot - For now inhibited, see explanation above.
- [x] secure boot inhibitor - Need to get a better link for the report.

### DIFF
--- a/repos/system_upgrade/common/actors/convert/updateefi/libraries/updateefi.py
+++ b/repos/system_upgrade/common/actors/convert/updateefi/libraries/updateefi.py
@@ -1,0 +1,230 @@
+import errno
+import os
+
+from leapp import reporting
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.common import efi
+from leapp.libraries.common.config import architecture, get_source_distro_id, get_target_distro_id, is_conversion
+from leapp.libraries.common.distro import distro_id_to_pretty_name, get_distro_efidir_canon_path
+from leapp.libraries.stdlib import api
+
+
+def _get_target_efi_bin_path():
+    # Sorted by priority.
+    # NOTE: The shim-x64 package providing the shimx64.efi binary can be removed when
+    # not using secure boot, grubx64.efi should always be present (provided by
+    # grub-efi-x64).
+    # WARN: However it is expected to have the shim installed on the system to comply
+    # with the official guidelines.
+    #
+    # TODO: There are usually 2 more shim* files which appear unused on a fresh system:
+    # - shim.efi - seems like it's the same as shimx64.efi
+    # - shim64-<distro>.efi - ???
+    # What about them?
+    efibins_by_arch = {
+        architecture.ARCH_X86_64: ("shimx64.efi", "grubx64.efi"),
+        architecture.ARCH_ARM64: ("shimaa64.efi", "grubaa64.efi"),
+    }
+
+    arch = api.current_actor().configuration.architecture
+    for filename in efibins_by_arch[arch]:
+        efi_dir = get_distro_efidir_canon_path(get_target_distro_id())
+        canon_path = os.path.join(efi_dir, filename)
+        if os.path.exists(canon_path):
+            return efi.canonical_path_to_efi_format(canon_path)
+
+    return None
+
+
+def _add_boot_entry_for_target(efibootinfo):
+    """
+    Create a new UEFI bootloader entry for the target system.
+
+    Return the newly created bootloader entry.
+    """
+    efi_bin_path = _get_target_efi_bin_path()
+    if not efi_bin_path:
+        # this is a fatal error as at least one of the possible EFI binaries
+        # should be present
+        raise efi.EFIError("Unable to detect any UEFI binary file.")
+
+    label = distro_id_to_pretty_name(get_target_distro_id())
+
+    existing_entry = efi.get_boot_entry(efibootinfo, label, efi_bin_path)
+    if existing_entry:
+        api.current_logger().debug(
+            "The '{}' UEFI bootloader entry is already present.".format(label)
+        )
+        return existing_entry
+
+    return efi.add_boot_entry(label, efi_bin_path)
+
+
+def _remove_boot_entry_for_source(efibootinfo):
+    efibootinfo_fresh = efi.EFIBootInfo()
+    source_entry = efibootinfo_fresh.entries.get(efibootinfo.current_bootnum, None)
+
+    if not source_entry:
+        api.current_logger().debug(
+            "The currently booted source distro EFI boot entry has been already"
+            " removed since the target entry has been added, skipping removal."
+        )
+        return
+
+    original_source_entry = efibootinfo.entries[source_entry.boot_number]
+
+    if source_entry != original_source_entry:
+        api.current_logger().debug(
+            "The boot entry with current bootnum has changed since the target"
+            " distro entry has been added, skipping removal."
+        )
+        return
+
+    efi.remove_boot_entry(source_entry.boot_number)
+
+
+def _try_remove_source_efi_dir():
+    """
+    Try to remove the source distro EFI directory
+
+    The directory is not reported if it's not empty to preserve potential
+    custom files. In such a case a post upgrade report is produced informing
+    user to handle the leftover files.
+    """
+    efi_dir_source = get_distro_efidir_canon_path(get_source_distro_id())
+    if not os.path.exists(efi_dir_source):
+        api.current_logger().debug(
+            "Source distro EFI directory at {} does not exist, skipping removal.".format(efi_dir_source)
+        )
+        return
+
+    target_efi_dir = get_distro_efidir_canon_path(get_target_distro_id())
+    if efi_dir_source == target_efi_dir:
+        api.current_logger().debug(
+            "Source and target distros use the same '{}' EFI directory.".format(efi_dir_source)
+        )
+        return
+
+    try:
+        os.rmdir(efi_dir_source)
+        api.current_logger().debug(
+            "Deleted source system EFI directory at {}".format(efi_dir_source)
+        )
+    except FileNotFoundError:
+        api.current_logger().debug(
+            "Couldn't remove the source system EFI directory at {}: the directory no longer exists".format(
+                efi_dir_source
+            )
+        )
+    except OSError as e:
+        if e.errno == errno.ENOTEMPTY:
+            api.current_logger().debug(
+                "Didn't remove the source EFI directory {}, it does not exist".format(
+                    efi_dir_source
+                )
+            )
+            summary = (
+                "During the upgrade, the EFI binaries and grub configuration files"
+                f" were migrated from the source OS EFI directory {efi_dir_source}"
+                f" to the target OS EFI directory {target_efi_dir}."
+                f" Leftover files were detected in {target_efi_dir}, review them"
+                " and migrate them manually."
+            )
+            reporting.create_report([
+                reporting.Title("Review leftover files in the source OS EFI directory"),
+                reporting.Summary(summary),
+                reporting.Groups([
+                    reporting.Groups.BOOT,
+                    reporting.Groups.POST,
+                ]),
+                reporting.Severity(reporting.Severity.LOW),
+            ])
+        else:
+            api.current_logger().error(
+                "Failed to remove the source system EFI directory at {}: {}".format(
+                    efi_dir_source, e
+                )
+            )
+            summary = (
+                f"Removal of the source system EFI directory at {efi_dir_source} failed."
+                " Remove the directory manually if present."
+            )
+            reporting.create_report([
+                reporting.Title("Failed to remove source system EFI directory"),
+                reporting.Summary(summary),
+                reporting.Groups([
+                    reporting.Groups.BOOT,
+                    reporting.Groups.FAILURE,
+                    reporting.Groups.POST,
+                ]),
+                reporting.Severity(reporting.Severity.LOW),
+            ])
+
+
+def _replace_boot_entries():
+    try:
+        efibootinfo = efi.EFIBootInfo()
+        target_entry = _add_boot_entry_for_target(efibootinfo)
+        # NOTE: this isn't strictly necessary as UEFI should set the next entry
+        # to be the first in the BootOrder. This is a workaround to make sure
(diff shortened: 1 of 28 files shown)


repos/system_upgrade/common/actors/convert/securebootinbihit/libraries/securebootinhibit.py
```diff
@@ -0,0 +1,30 @@
...
+            details={'details': 'Actor did not receive FirmwareFacts message.'}
+        )
+
+    if ff.firmware == 'efi' and ff.secureboot_enabled:
+        reporting.create_report([
+            reporting.Title('SecureBoot is enabled'),
+            reporting.Summary((
+                'Upgrades with conversions are currently not possible when SecureBoot'
+                ' is enabled in the UEFI.'
+            )),
```
pirat89:
```suggestion
            reporting.Title('Detected enabled Secure Boot when trying to convert the system'),
            reporting.Summary((
                'The conversion to a different linux distribution is not possible when the Secure Boot'
                ' is enabled. Artifacts in the target linux distribution use are signed by different keys'
                ' that are not accepted by current linux distribution.'
            )),
```

pirat89 (changes requested):
my apology for partial review - maybe is valid, I haven't finished the full review yet so I am still considering some stuff. But sharing as for sure some parts will need to be made differently.

repos/system_upgrade/common/actors/convert/securebootinbihit/libraries/securebootinhibit.py
pirat89: @matejmatuska  btw, typo in dirname: securebootin**bihit** -> securebootin**hibit**

repos/system_upgrade/common/libraries/firmware/efi.py
```diff
@@ -0,0 +1,325 @@
...
+        "/usr/sbin/efibootmgr",
+        "--create",
+        "--disk",
+        efi_dev,
+        "--part",
+        str(esp_number),
+        "--loader",
+        efi_bin_path,
+        "--label",
+        label,
```
pirat89: nitpick: I will be old school, and say that this is for me more readable :) but I am ok with original formatting as well.
```suggestion
        "/usr/sbin/efibootmgr",
        "--create",
        "--disk", efi_dev,
        "--part", str(esp_number),
        "--loader", efi_bin_path,
        "--label", label,
```

repos/system_upgrade/common/libraries/distro.py
```diff
@@ -208,3 +208,16 @@ def get_distro_repoids(context, distro, major_version, arch):
...
+
+
+def distro_id_to_pretty_name(distro_id):
+    """
+    Get pretty name for the given distro id.
+
+    The pretty name is what is found in the NAME field of /etc/os-release.
+    """
+    return {
+        "rhel": "Red Hat Enterprise Linux",
```
pirat89: didn't we agree to use just `"RHEL` acronym actually? I dunno honestly, but as we usually do not use the full name in reports, I would expect to use `RHEL` instead.

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -297,10 +297,26 @@ def _get_firewall_status(service_name):
...
     firmware = 'efi' if os.path.isdir('/sys/firmware/efi') else 'bios'
+
+    ppc64le_opal = None
     if architecture.matches_architecture(architecture.ARCH_PPC64LE):
-        ppc64le_opal = bool(os.path.isdir('/sys/firmware/opal/'))
-        return FirmwareFacts(firmware=firmware, ppc64le_opal=ppc64le_opal)
-    return FirmwareFacts(firmware=firmware)
+        ppc64le_opal = os.path.isdir('/sys/firmware/opal/')
+
+    is_secureboot = None
```
pirat89: @matejmatuska  why not make it False by default already when allowed value is boolean?

repos/system_upgrade/common/libraries/distro.py
pirat89: Just thinking whether it will not be a little bit confusing with anothe rmap we plan to deliver. But I possibly overthinking it right now. Let's discuss this one later.

repos/system_upgrade/common/libraries/grub.py
```diff
@@ -16,20 +17,26 @@
 """The path to the env file for GRUB2 in BIOS"""
 
 
+@deprecated(
+    since="2025-12-04",
+    message="Replaced by canonical_path_to_efi_format in the firmware.efi library",
+)
```
pirat89: do not forget to update docs about all deprecations

repos/system_upgrade/common/models/firmwarefacts.py
```diff
@@ -9,3 +9,5 @@ class FirmwareFacts(Model):
     """ System firmware interface (BIOS or EFI) """
     ppc64le_opal = fields.Nullable(fields.Boolean())
     """ Check OPAL presence to identify ppc64le bare metal systems """
+    secureboot_enabled = fields.Nullable(fields.Boolean())
+    """ Whether SecureBoot is enabled, always False on BIOS systems """
```
pirat89: do we expect anything than True/False to make it nullable?

repos/system_upgrade/el8toel9/actors/addarmbootloaderworkaround/libraries/addupgradebootloader.py
```diff
@@ -19,9 +13,8 @@
 ARM_SHIM_PACKAGE_NAME = 'shim-aa64'
 ARM_GRUB_PACKAGE_NAME = 'grub2-efi-aa64'
 
-EFI_MOUNTPOINT = '/boot/efi/'
-LEAPP_EFIDIR_CANONICAL_PATH = os.path.join(EFI_MOUNTPOINT, 'EFI/leapp/')
-RHEL_EFIDIR_CANONICAL_PATH = os.path.join(EFI_MOUNTPOINT, 'EFI/redhat/')
+LEAPP_EFIDIR_CANONICAL_PATH = os.path.join(efi.EFI_MOUNTPOINT, 'EFI/leapp/')
+RHEL_EFIDIR_CANONICAL_PATH = os.path.join(efi.EFI_MOUNTPOINT, 'EFI/redhat/')
```
pirat89: @matejmatuska  Hmm, in this way, it would be working just for ARM on RHEL. Shouldn't it be dynamic? having a function in efi lib saying what is the expected efidir path? Also it could be broken in case of conversion. and now I do not remember architectures of other distros. so pointing this out at least

repos/system_upgrade/common/actors/convert/securebootinbihit/tests/test_securebootinhibit.py
```diff
@@ -0,0 +1,58 @@
...
+            FirmwareFacts(firmware='efi', ppc64le_opal=None, secureboot_enabled=True),
+            False,
+            False
+        ),
+        # bios is ok during conversion too
+        (
+            FirmwareFacts(firmware='bios', ppc64le_opal=None, secureboot_enabled=True),
+            True,
+            False
+        ),
```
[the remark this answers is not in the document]
matejmatuska: good catch

repos/system_upgrade/common/actors/convert/securebootinbihit/libraries/securebootinhibit.py
matejmatuska: Thanks, the report was still a little bit WIP.

repos/system_upgrade/common/libraries/firmware/efi.py
```diff
@@ -0,0 +1,325 @@
...
+    ]
+
+    try:
+        run(cmd)
+    except CalledProcessError as e:
+        raise EFIError(
+            "Unable to add a new UEFI bootloader entry '{}' for EFI binary at {}.".format(
+                label, efi_bin_path
+            )
```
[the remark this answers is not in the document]
matejmatuska: Thanks, I am still forgetting that ocasionaly.

repos/system_upgrade/common/models/firmwarefacts.py
matejmatuska: Not really, but same for the `ppc64le_opal` and yet it's set to `None` when not on ppcle. So it means sonmething like "not applicable".
I did the same with the secureboot which is "not applicable" i.e. `None` on BIOS systems.

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
matejmatuska: I replied to the comment in the model.

repos/system_upgrade/common/libraries/grub.py
matejmatuska: Thanks, I would.

repos/system_upgrade/el8toel9/actors/addarmbootloaderworkaround/libraries/addupgradebootloader.py
matejmatuska: Right, definitely good point about the distro name.

repos/system_upgrade/el8toel9/actors/addarmbootloaderworkaround/libraries/addupgradebootloader.py
```diff
@@ -59,21 +52,31 @@ def process():
...
+            )
 
         efibootentry_fields = ['boot_number', 'label', 'active', 'efi_bin_source']
         api.produce(
             ArmWorkaroundEFIBootloaderInfo(
                 original_entry=EFIBootEntry(**{f: getattr(current_boot_entry, f) for f in efibootentry_fields}),
                 upgrade_entry=EFIBootEntry(**{f: getattr(upgrade_boot_entry, f) for f in efibootentry_fields}),
                 upgrade_bls_dir=UPGRADE_BLS_DIR,
-                upgrade_entry_efi_path=os.path.join(EFI_MOUNTPOINT, LEAPP_EFIDIR_CANONICAL_PATH),
+                upgrade_entry_efi_path=os.path.join(efi.EFI_MOUNTPOINT, LEAPP_EFIDIR_CANONICAL_PATH),
```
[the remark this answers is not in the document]
matejmatuska: It is, good catch

repos/system_upgrade/common/libraries/firmware/efi.py
```diff
@@ -0,0 +1,325 @@
...
+
+
+class EFIError(Exception):
+    """
+    Exception raised when EFI operation failed.
+    """
+
+
+def canonical_path_to_efi_format(canonical_path):
+    r"""Transform the canonical path to the UEFI format.
```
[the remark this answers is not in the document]
matejmatuska: Ah nevermind the r has to be there for the backslashes.

repos/system_upgrade/common/libraries/firmware/efi.py
```diff
@@ -0,0 +1,325 @@
...
+
+
+def get_efi_device():
+    """Get the block device on which GRUB is installed."""
+
+    return blk_dev_from_partition(get_efi_partition())
+
+
+# TODO maybe rename this to find_boot_entry
+def get_boot_entry(efibootinfo, label, efi_bin_path):
```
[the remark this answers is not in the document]
matejmatuska: I kept it as it is

repos/system_upgrade/common/libraries/distro.py
pirat89: it's possible :D

repos/system_upgrade/common/actors/convert/updateefi/libraries/updateefi.py
```diff
@@ -0,0 +1,191 @@
...
+    """
+    Get canonical path to the distro EFI directory in the EFI mountpoint.
+
+    NOTE: The path might be incorrect for distros not properly enabled for IPU,
+    when enabling new distros in the codebase, make sure the path is correct.
+    """
+    if distro_id == "rhel":
+        return os.path.join(efi.EFI_MOUNTPOINT, "EFI", "redhat")
+
+    return os.path.join(efi.EFI_MOUNTPOINT, "EFI", distro_id)
```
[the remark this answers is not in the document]
matejmatuska: Okay, I will move it to the distro lib for now, but I think we should have some followup where we move it to some config file along with the other "static" information about distro. It would make enabling of other distros nicer.

pirat89 (approved):
lgtm and it seems to be working - however we have not been able to check all calls from actors yet to potentially catch all possible exceptions. Due to high amount of changes, we decided to merge it in this current state and cover remaining possible overlooked calls in a followup PR later this week.

PR #1476, opened 2026-01-15
Enable LiveMode for non-x86_64 architectures

MichalHe:
This patch contains the necessary changes to enable LiveMode upgrades. All non-x86_64 upgrade paths were tested, motivating the changes in this PR. Primary obstacle was the non-existence of the `dmidecode` package on other architectures, other "required" packages were removed as they are not necessary to perform an upgrade.

Jira-ref: RHEL-141861


pirat89 (approved):
lgtm and it make sense. I haven't tested it manually but it seems straightforward and the feature is still experimental.

PR #1495, opened 2026-02-11
feat(net-naming-scheme): enable by default on CS 9to10

matejmatuska:
This was done for RHEL 9to10 in 91f37a3913c1608b24c9aa583ac3b13a08aace71.

Jira: RHEL-148291


pirat89 (approved):
lgtm and works as expected.

On the upgraded system:
```
systemd-udevd[395]: Using default interface naming scheme 'rhel-9.0'.
```

PR #1499, opened 2026-02-23
Add APIs to get  distro "report" names

matejmatuska:
Since we plan to change the report texts to respect distro name, there is a need for a convenient API to retrieve them and use them for formatting. For that a new global is introduced in the distro library, `DISTRO_REPORT_NAMES`. More details in the docstrings, see test for example usage.

Jira: RHEL-130948

NOTE: I am still thinking it might be nicer to use a metaclass instead of the class and a class instead of the global var, it is easier to deprecate. Or replace the global var with a funciton

### DIFF
--- a/repos/system_upgrade/common/libraries/distro.py
+++ b/repos/system_upgrade/common/libraries/distro.py
@@ -1,9 +1,10 @@
 import json
 import os
+from collections.abc import Mapping
 
 from leapp.exceptions import StopActorExecutionError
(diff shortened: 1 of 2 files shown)


pirat89 (changes requested):
I like it. just squash commits and fix typos.

PeterMocary (approved):
LGTM! Note that there is a typo in the commit message "can be directly be used".

matejmatuska:
Rebased

pirat89 (approved):

PR #1501, opened 2026-02-25
Enable logrotate.timer during 8to9 IPU

matejmatuska:
On el8 logrotate uses cron, on el9 a systemd timer is used. This is not automatically migrated during an IPU, this patch adds a actor that does so.

Jira: RHEL-17361


pirat89 (approved):
lgtm. keep merging after manual test.

PR #1504, opened 2026-03-04
livemode: copy upgrade kernel's hmac into /boot

MichalHe:
Copying kernel's HMAC is necessary in order to boot with FIPS mode enabled. Standard (old) upgrade process copies the kernel HMAC file already, livemode did not in order to keep the initial implementation simple. This change adds copying of kernel HMAC also for livemode.

Jira-ref: RHEL-129571


pirat89 (approved):
lgtm - straighforward solution and based on logs above it works. merging

PR #1506, opened 2026-03-09
Add new Leapp actor to preinstall gnome-settings-daemon-server-defaults on "Server with GUI" el9 -> el10 upgrades

tpopela:
Prior to RHEL 10 we didn't have a way to enforce some of the desktop environment changes to be only Server specific (such as disabling the power button handling or disabling automatic suspendand) hence we had to enable everything for everyone. In RHEL 10 we've added a new subpackage of gnome-settings-daemon, the gnome-settings-daemon-server-defaults which is only preinstalled on "Server with GUI" product and on nothing else. But this only works for clean installations as the comps groups are not synced when doing the upgrades with Leapp. To fix that we've created this Leapp actor which will detect if the graphical-server-environment comps group was installed and only in the case it will preinstall the server-defaults subpackage.

Developed by Claude Code.

Jira: RHEL-148697

### DIFF
--- a/repos/system_upgrade/common/actors/scandnfcomps/actor.py
+++ b/repos/system_upgrade/common/actors/scandnfcomps/actor.py
@@ -1,0 +1,22 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import scandnfcomps
+from leapp.models import InstalledDNFComps
+from leapp.tags import FactsPhaseTag, IPUWorkflowTag
+
+
+class ScanDNFComps(Actor):
+    """
+    Scan installed DNF comps (environments and groups).
+
+    Collects information about DNF package environments and groups that are
+    currently installed on the source system and produce InstalledDNFComps
+    message.
+    """
+
+    name = 'scan_dnf_comps'
+    consumes = ()
+    produces = (InstalledDNFComps,)
+    tags = (FactsPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        scandnfcomps.process()
--- a/repos/system_upgrade/common/actors/scandnfcomps/libraries/scandnfcomps.py
+++ b/repos/system_upgrade/common/actors/scandnfcomps/libraries/scandnfcomps.py
@@ -1,0 +1,87 @@
+import warnings
+
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.common.dnflibs import create_dnf_base, DNFRepoError
+from leapp.libraries.stdlib import api
+from leapp.models import DNFEnvironment, DNFGroup, InstalledDNFComps
+
+try:
+    import dnf
+except ImportError:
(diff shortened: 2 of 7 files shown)


pirat89 (changes requested):
Discussed over slack. Current implementation seems to be working, however we do not allow interraction with the system during checksphase. Also I think that it makes sense to scan information about installed comps, so I decided to implement that part. However, current design (mainly common dnf-related libraries) seems to be a little bit not-ideal for that purposes. So I plan to do the groundwork in #1507  - once the PR is merged, we can updated this one.

matejmatuska (approved):
LGTM.

I also run a test, after preupgrade the messages are correctly produced:
```
Stamp: 2026-06-11T08:59:57.017981Z
Actor: gnome_settings_daemon_check
Phase: Checks
Type: RpmTransactionTasks
Message_data:
{
    "local_rpms": [],
    "modules_to_enable": [],
    "modules_to_reset": [],
    "to_install": [
        "gnome-settings-daemon-server-defaults"
    ],
    "to_keep": [],
    "to_remove": [],
    "to_upgrade": []
}
```

```
Stamp: 2026-06-11T08:59:48.508385Z
Actor: scan_dnf_comps
Phase: FactsCollection
Type: InstalledDNFComps
Message_data:
{
    "environments": [
        {
            "id": "graphical-server-environment",
            "name": "Server with GUI"
        }
    ],
    "groups": [
        {
            "id": "base",
            "name": "Base"
        },
        {
            "id": "base-x",
            "name": "base-x"
        },
        {
            "id": "container-management",
            "name": "Container Management"
        },
        {
            "id": "core",
            "name": "Core"
        },
        {
            "id": "fonts",
            "name": "Fonts"
        },
        {
            "id": "gnome-desktop",
            "name": "GNOME"
        },
        {
            "id": "guest-desktop-agents",
            "name": "Guest Desktop Agents"
        },
        {
            "id": "hardware-monitoring",
            "name": "Hardware Monitoring Utilities"
        },
        {
            "id": "hardware-support",
            "name": "Hardware Support"
        },
        {
            "id": "headless-management",
            "name": "Headless Management"
        },
        {
            "id": "internet-browser",
            "name": "Internet Browser"
        },
        {
            "id": "multimedia",
            "name": "Multimedia"
        },
        {
            "id": "networkmanager-submodules",
            "name": "Common NetworkManager submodules"
        },
        {
            "id": "platform-kvm",
            "name": "KVM platform specific packages"
        },
        {
            "id": "print-client",
            "name": "Printing Client"
        },
        {
            "id": "server-product",
            "name": "Server product core"
        },
        {
            "id": "standard",
            "name": "Standard"
        }
    ]
}
```

and the env group is installed one the upgraded system.

pirat89 (approved):

PR #1507, opened 2026-03-10
Redesign dnf shared libraries

pirat89:
## Summary / tl;dr
Consolidate DNF-related libraries (dnfconfig, dnfplugin, module) into a new dnflibs package for better organization. The old modules remain as deprecated shims using @deprecated decorator, ensuring full backward compatibility while encouraging migration to the new libraries. New libraries are
* dnflibs
* dnflibs.dnfmodule
* dnflibs.dnfconfig
* dnflibs.dnfplugin

The code in the project is updated to use the new libraries. Original `StopActorExecutionError` exceptions in these libs are replaced by new ones, providing more possibilities to deal with issues in callers.

Introduced new public function for a correct DNF base initialization with the filled sack.

This PR is prerequisite for #1506. 

## Changes
- Create repos/system_upgrade/common/libraries/dnflibs/ package
- Move dnfconfig.py to dnflibs/dnfconfig.py
- Move dnfplugin.py to dnflibs/dnfplugin.py
- Move module.py    to dnflibs/dnfmodule.py (renamed)
- Improve the test coverage in new libs
- Mark original libraries and their content as deprecated
- Update imports in the code to use the new libraries instead
- Update the documentation to reflect deprecated functionality
- Drop obsoleted dead code
- Introduce new `leapp_tmpdir` fixture in the `conftest.py` file. The fixture reflexes earlier team discussion about creating temp files/dirs in unit-tests just under `/tmp` and using the `leapptest-` prefix for them.
- Extend methods in testutils.logger_mocked to accept kwargs. 

## New exceptions hierarchy

```
StopActorExecutionError (from leapp.exceptions)             
├── DNFError (dnflibs/__init__)
│   ├── DNFRepoError (dnflibs/__init__)
│   ├── InvalidDNFConfig (dnflibs/dnfconfig)
│   ├── CannotObtainDNFConfig (dnflibs/dnfconfig)
│   ├── CannotUpdateDNFConfig (dnflibs/dnfconfig)
│   ├── DNFPluginInstallError (dnflibs/dnfplugin)
│   ├── DNFExecutionError (dnflibs/dnfplugin)
│   └── DNFUpgradeTransactionError (dnflibs/dnfplugin)
└── RegisteredWorkaroundApplicationError (dnflibs/dnfplugin)
```

---
### TODO
- [x]  find agreement on the naming :)
- [x] drop dead code & deprecated functionality
- [x] replace StopActorExecutionError in libs by new exceptions (inherit from StopActorExecutionError)
- [x] introduce public function to initialize `dnf.Base` objects correctly 
- [x] update unit tests
- [ ] implement scanner of DNF comps
  - due to the current size of the PR, this is going to be split into a separate one


repos/system_upgrade/common/actors/rpmscanner/libraries/rpmscanner.py
```diff
@@ -1,8 +1,8 @@
 import warnings
 
 from leapp.exceptions import StopActorExecutionError
-from leapp.libraries.common import module as module_lib
 from leapp.libraries.common import rpms
+from leapp.libraries.common.dnflibs import dnfmodule as module_lib
```
PeterMocary: Nitpick: I'd drop the as module_lib since now it has distinguishable name. Same in tests.

repos/system_upgrade/common/libraries/dnflibs/dnfmodule.py
```diff
@@ -0,0 +1,119 @@
...
+    import hawkey
+except ImportError:
+    hawkey = None
+    warnings.warn('Could not import the `hawkey` python module.', ImportWarning)
+
+
+class DNFRepoError(StopActorExecutionError):
+    """
+    Used when DNF fails to load repositories.
+    """
```
PeterMocary: Nitpick: Maybe the exceptions could be in a separate file so that they are accessible for all dnf libraries for future use. It seems cleaner to me, but currently I don't see much similarities and if it isn't expected to extend these libraries significantly, it doesn't matter.

repos/system_upgrade/common/libraries/dnflibs/dnfmodule.py
```diff
@@ -0,0 +1,119 @@
...
+        return []
+    return module_base.get_modules('*')[0]
+
+
+def get_enabled_modules():
+    """
+    Return currently enabled module streams as a list of libdnf.module.ModulePackage objects.
+
+    The function return an empty list if the DNF python module is not present.
+    """
```
PeterMocary, line 55: We should update doc strings of _all exposed functions_ similar to the `get_modules` function. Namely, the raises is missing here as it also calls the `_create_or_get_dnf_base()`. 

Same for the `map_installed_rpms_to_modules()`. I see that the `dnfconfig.py` wasn't updated with new exceptions yet, but same there as well as in the `dnfplugin.py`.

PeterMocary (commented):
I went through the changes just to get up to speed. 

- The naming seems reasonable considering we don't want to shadow `dnf`.
- One of the commit messages has typo "dnfplufin" -> "dnfplugin".
- Make sure to update all of the doc strings, but it is still WIP so you probably plan to anyway.

repos/system_upgrade/common/libraries/dnflibs/dnfmodule.py
pirat89: yy. update of docstring is still in plan. I plan to update also docstrings in callers to include information about possibly raised exceptions there as well if they are not caught and they are passed to actors.

repos/system_upgrade/common/libraries/dnflibs/dnfmodule.py
pirat89: I am still thinking about putting them into the `__init__.py` (a.k.a `dnflibs`). Keeping this step however for later at this moment.

pirat89: @PeterMocary  actually, wdyt about introducing sth like `DNFGeneralError` ? it does not have to be necessarily in this PR. just thinking that if someone would not be curious about various DNF problems, but wants to distinguish between DNF problems related exceptions and too much generic `StopActorExecutionError`, then it will be better this way. But possibly I am overthinking it now at this late hour.

repos/system_upgrade/common/libraries/dnflibs/dnfmodule.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def get_modules(base=None):
+    """
+    Return info about all module streams as a list of libdnf.module.ModulePackage objects.
+
+    The function return an empty list if the DNF python module is not present.
+
+    :param base: If it is set, use it instead of creating a new one.
+    :type base: dnf.Base
```
PeterMocary:
```suggestion
    :type base: dnf.Base or None
```

repos/system_upgrade/common/libraries/dnflibs/dnfmodule.py
```diff
@@ -0,0 +1,87 @@
...
+def get_enabled_modules():
+    """
+    Return currently enabled module streams as a list of libdnf.module.ModulePackage objects.
+
+    The function return an empty list if the DNF python module is not present.
+
+    .. seealso::
+        :func:`get_modules` for exceptions raised during module discovery.
+        :func:`create_dnf_base` for exceptions raised when creating dnf.Base
+    """
```
PeterMocary, line 55: Nitpick: I'd add it everywhere for the consistency, but as we discussed, moving to the typing syntax in future basically removes the need to make these perfect. The description states it clearly, so I'm fine with leaving it as is.
```suggestion
    Return currently enabled module streams as a list of libdnf.module.ModulePackage objects.

    The function returns an empty list if the DNF python module is not present.
    
    :returns: Currently enabled module streams
    :rtype: List[libdnf.module.ModulePackage]

    .. seealso::
        :func:`get_modules` for exceptions raised during module discovery.
        :func:`create_dnf_base` for exceptions raised when creating dnf.Base
    """
```

repos/system_upgrade/common/actors/rpmscanner/libraries/rpmscanner.py
pirat89: fixed

repos/system_upgrade/common/libraries/dnflibs/dnfmodule.py
pirat89: updated

repos/system_upgrade/common/libraries/dnflibs/dnfmodule.py
pirat89: Created the `DNFError` exception - all other DNF related exceptions are based on it. I keep the exceptions separated now in own libs. we can still discuss it later or change it in a followup PRs (until we introduce new public release).

@oamg/developers  more opinions? :)

repos/system_upgrade/common/libraries/dnflibs/tests/test_dnflibs.py
```diff
@@ -0,0 +1,118 @@
...
+    ('Generic error without repo information', 'unknown repo'),
+])
+def test_create_dnf_base_repo_error(monkeypatch, error_message, expected_repoid):
+    class RepoError(Exception):
+        pass
+
+    class MockExceptions:
+        pass
+
+    MockExceptions.RepoError = RepoError
```
PeterMocary, line 80: Nitpick: Couldn't this be defined in the class?


```suggestion
    class MockExceptions:
        RepoError = RepoError
```

repos/system_upgrade/common/libraries/dnflibs/tests/test_dnfmodule.py
```diff
@@ -0,0 +1,275 @@
...
+
+There is actually not a good coverage we could do for this library via
+unit-tests. These tests seem to have a low value at all as there is no
+a good way to cover this library without complete mocking of actually whole
+DNF & hawkey libs. At this point, we keep focus on integration tests executed
+in CI pipelines.
+
+I am still keeping these unit-tests present for an elementary functionality
+check - but mainly because I expect low maintenance cost for them as the library
+is kind of finished - we can expect minimal changes in future until its removal.
```
PeterMocary: Nitpick: I'd keep it simple.
```suggestion
Achieving better unit-test coverage for this library would require extensive mocking of DNF and hawkey libs. Thus, we check only the basic functionality here, and rely on integration tests that provide additional coverage.
```

repos/system_upgrade/common/libraries/dnflibs/tests/test_dnfmodule.py
```diff
@@ -0,0 +1,275 @@
...
+
+    assert len(modules) == 1
+    assert modules[0].getName() == 'nodejs'
+
+
+def test_get_modules_no_get_modules_method(monkeypatch):
+    """
+    Test the compatibility for DNF versions without module stream functionality.
+
+    This is know a case for RHEL 7, which we do not support in upstream anymore.
```
PeterMocary:
```suggestion
    This is a known case for RHEL 7, which we do not support in upstream anymore.
```

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
```diff
@@ -0,0 +1,610 @@
...
+    """
+    base_cmd = ['rpmdb', '--rebuilddb']
+    cmd = base_cmd if not root else base_cmd + ['-r', root]
+    context.call(cmd)
+
+
+def build_plugin_data(target_repoids, debug, test, tasks, on_aws):
+    """
+    Generates a dictionary with the DNF plugin data.
+    """
```
PeterMocary, line 112: Missing the docstring with parameter types and return type. Same for other public functions.

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
```diff
@@ -0,0 +1,610 @@
...
+    raise DNFUpgradeTransactionError(message=message, details=details)
+
+
+def _transaction(context, stage, target_repoids, tasks, plugin_info, xfs_info,
+                 test=False, cmd_prefix=None, on_aws=False):
+    """
+    Perform the actual DNF rpm download via our DNF plugin
+    """
+
+    # we do not want
```
PeterMocary: Nitpick: Seems like a leftover, we can remove it.

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
```diff
@@ -0,0 +1,610 @@
...
+
+        disable_plugins = []
+        if plugin_info:
+            for info in plugin_info:
+                if stage in info.disable_in:
+                    disable_plugins += [info.name]
+
+        dnfconfig.exclude_leapp_rpms(context, disable_plugins)
+        _transaction(
+            context=context, stage='check', target_repoids=target_repoids, plugin_info=plugin_info, xfs_info=xfs_info,
```
PeterMocary: Nitpick:
```suggestion
            context=context, stage=stage, target_repoids=target_repoids, plugin_info=plugin_info, xfs_info=xfs_info,
```

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
```diff
@@ -0,0 +1,610 @@
...
+
+        # we have to ensure the leapp packages will stay untouched
+        # Note: this is the most probably duplicate action - it should be already
+        # set like that, however seatbelt is a good thing.
+        dnfconfig.exclude_leapp_rpms(context, disable_plugins)
+
+        if int(get_target_major_version()) >= 9:
+            _rebuild_rpm_db(context, root='/installroot')
+        _transaction(
+            context=context, stage='upgrade', target_repoids=target_repoids, plugin_info=plugin_info,
```
PeterMocary: Nitpick:

```suggestion
            context=context, stage=stage, target_repoids=target_repoids, plugin_info=plugin_info,
```

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
```diff
@@ -0,0 +1,610 @@
...
+        disable_plugins = []
+        if plugin_info:
+            for info in plugin_info:
+                if stage in info.disable_in:
+                    disable_plugins += [info.name]
+
+        apply_workarounds(overlay.nspawn())
+        dnfconfig.exclude_leapp_rpms(context, disable_plugins)
+        _transaction(
+            context=context, stage='download', target_repoids=target_repoids, plugin_info=plugin_info, tasks=tasks,
```
PeterMocary: Nitpick:

```suggestion
            context=context, stage=stage, target_repoids=target_repoids, plugin_info=plugin_info, tasks=tasks,
```

repos/system_upgrade/common/libraries/dnflibs/tests/test_dnfplugin.py
```diff
@@ -0,0 +1,290 @@
...
+    target_plugin_dir = os.path.join(
+        target_basedir,
+        dnfplugin._DNF_PLUGIN_PATHS[target_version].lstrip('/')
+    )
+    os.makedirs(os.path.dirname(target_plugin_dir), exist_ok=True)
+
+    dnfplugin.install(target_basedir)
+
+    target_plugin_path = os.path.join(os.path.dirname(target_plugin_dir), dnfplugin.DNF_PLUGIN_NAME)
+    assert os.path.exists(target_plugin_path)
```
PeterMocary: Nitpick: We could simply mock the `shutil.copy2` and verify whether it gets correctly formed arguments. For the failure test, similarly, mock the `shutil.copy2` and make it raise exception and verify how install reacts.

I don't know if actually doing the copy is necessary. Apart from making the test work regardless of the copy function used, I don't see any advantage. Of course, my suggestion trusts that `shutil.copy2` works when given correct arguments, but I'd say that it is [unit-tested in `shutil` ](https://github.com/python/cpython/blob/f6d16a0d708c611dd96bb739f0896a0b6d7e04e8/Lib/test/test_shutil.py#L1388)and we can rely on that.

What do you think?

conftest.py
```diff
@@ -96,3 +100,21 @@ def pytest_runtestloop(session):
...
+    that follows the project's naming convention for test directories.
+
+    :returns: Path to the temporary directory
+    :rtype: str
+    """
+    tmpdir = tempfile.mkdtemp(prefix='leapptest-', dir='/tmp')
+    try:
+        yield tmpdir
+    finally:
+        shutil.rmtree(tmpdir)
```
PeterMocary: Could we use [TemporaryDirectory](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) instead? It also has the `dir` parameter, but I don't think it is really necessary. 

```suggestion
@pytest.fixture
def leapp_tmpdir():
    """
    Create a temporary directory with 'leapptest-' prefix in /tmp.

    This fixture automatically creates and cleans up a temporary directory
    that follows the project's naming convention for test directories.

    :returns: Path to the temporary directory
    :rtype: str
    """
    @pytest.fixture
    def leapp_tmpdir():
        with tempfile.TemporaryDirectory(prefix='leapptest-') as tmpdir:
            yield tmpdir
```

Alternatively, we should consider the [builtin pytest's `tmp_path` fixture](https://python-basics-tutorial.readthedocs.io/en/latest/test/pytest/builtin-fixtures.html#tmp-path-and-tmp-path-factory) if possible.

PeterMocary (changes requested):
I checked against original code and couldn't find any unintended functionality changes. Exceptions were reworked, and are being raised where required.

I've found mostly nitpicks, but the new fixture needs some more attention, thus I'm requesting changes.

repos/system_upgrade/common/libraries/dnflibs/dnfmodule.py
PeterMocary: As we discussed offline, I'm okay with it. :+1:

repos/system_upgrade/common/libraries/dnflibs/tests/test_dnfmodule.py
pirat89: @PeterMocary  actually the suggested message is different. there is already extensive mocking of everything - which is the reason why tests have low value - and it's not possible to write unit-tests in a better way without this extensive mocking, where basically almost everything in the code is mocked.

repos/system_upgrade/common/libraries/dnflibs/tests/test_dnfmodule.py
pirat89: actually, it was a different typo :D `s/know/now/` but this works too

repos/system_upgrade/common/libraries/dnflibs/dnfmodule.py
pirat89: will be updated

repos/system_upgrade/common/libraries/dnflibs/dnfmodule.py
pirat89: will be updated

repos/system_upgrade/common/libraries/dnflibs/tests/test_dnfmodule.py
pirat89: I made it shorter.

repos/system_upgrade/common/libraries/dnflibs/tests/test_dnfmodule.py
pirat89: will be updated

repos/system_upgrade/common/libraries/dnflibs/tests/test_dnflibs.py
pirat89: indeed! good point!

pirat89: @PeterMocary  interesting again, not at all :D it must be separated:
```       
>       class MockExceptions:

repos/system_upgrade/common/libraries/dnflibs/tests/test_dnflibs.py:77: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    class MockExceptions:
>       RepoError = RepoError
E       NameError: name 'RepoError' is not defined

repos/system_upgrade/common/libraries/dnflibs/tests/test_dnflibs.py:78: NameError
```

it's without a problem in newer python when running interactively, but for some reason, it is broken when running tests. So let's keep the original way to not spend time on it.

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
pirat89: good catch.

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
pirat89: good catch.

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
pirat89: all docstring are updated.

conftest.py
pirat89: @PeterMocary `tmp_path` is available since python 3.8. But we require python 3.6+ compatibility. But `tempfile.TemporaryDirectory` is ok as well.

pirat89: it will be updated with next push

repos/system_upgrade/common/libraries/dnflibs/tests/test_dnfplugin.py
pirat89: I like it. Just be aware that this way, we make the test more dependent on the code. The current way is more robust, but I like this more simple way. So let's do it simply.

pirat89:
@PeterMocary  all changes delivered.

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
```diff
@@ -0,0 +1,679 @@
...
+        )
+
+
+def _rebuild_rpm_db(context, root=None):
+    """
+    Convert rpmdb from BerkeleyDB to Sqlite
+
+    :param target_basedir: The target userspace container base directory
+    :type target_basedir: str
+    """
```
PeterMocary, line 90: This param is incorrect.

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
```diff
@@ -0,0 +1,679 @@
...
+    }
+    return data
+
+
+def create_config(context, target_repoids, debug, test, tasks, on_aws=False):
+    """
+    Creates the configuration data file for our DNF plugin.
+
+    :param context: Execution context (defaults to NotIsolatedActions)
+    :type context: mounting.IsolatedActions or None
```
PeterMocary: Do we really want None here? It seems like it would just crash. Same for the other context params through out the PR.

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
```diff
@@ -0,0 +1,679 @@
...
+    :param target_userspace_info: Information about the target userspace
+    :type target_userspace_info: TargetUserSpaceInfo
+    :param storage_info: Storage and filesystem information
+    :type storage_info: StorageInfo
+    :param used_repos: Target repositories to use for the upgrade
+    :type used_repos: UsedTargetRepositories
+    :param tasks: Explicit RPM transaction tasks for the DNF plugin
+    :type tasks: FilteredRpmTransactionTasks
+    :param plugin_info: DNF plugin configuration
+    :type plugin_info: DNFPluginTask
```
PeterMocary: Based on the usage it seems like this is a `List[DNFPluginTask]` or some iterable.

repos/system_upgrade/common/libraries/dnflibs/dnfconfig.py
```diff
@@ -0,0 +1,138 @@
...
+    Just like str.split(), but remove ambient whitespaces from all items
+    """
+    return [item.strip() for item in data.split(sep, maxsplit)]
+
+
+def _get_main_dump(context, disable_plugins):
+    """
+    Return the dnf configuration dump of main options for the given context.
+
+    Returns the list of lines after the line with "[main]" section
```
PeterMocary: Actually returns dict not list.

repos/system_upgrade/common/libraries/dnflibs/dnfconfig.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    pkgs = _strip_split(_get_main_dump(context, disable_plugins).get('exclude', ''), ',')
+    return [i for i in pkgs if i]
+
+
+def _set_excluded_pkgs(context, pkglist, disable_plugins):
+    """
+    Configure DNF to exclude packages in the given list
+
+    Raise the CalledProcessError on error.
```
PeterMocary: This is outdated

repos/system_upgrade/common/libraries/dnflibs/dnfconfig.py
```diff
@@ -0,0 +1,138 @@
...
+
+    This has to be called several times to ensure that our RPMs are not removed
+    or updated (replaced) during the IPU. The action should happen inside
+        - the target userspace container
+        - on the host system
+    So user will have to drop these packages from the exclude after the
+    upgrade.
+
+    :param context: An instance of a mounting.IsolatedActions class
+    :type context: mounting.IsolatedActions class
```
PeterMocary:
```suggestion
    :type context: mounting.IsolatedActions
```

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
```diff
@@ -0,0 +1,679 @@
...
+        finally:
+            if stage == 'check':
+                backup_debug_data(context=context)
+
+
+@contextlib.contextmanager
+def _prepare_transaction(used_repos, target_userspace_info, binds=()):
+    """ Creates the transaction environment needed for the target userspace DNF execution  """
+    target_repoids = set()
+    for message in used_repos:
```
PeterMocary, line 340: Based on this, `used_repos` in function calling this one should have `Iterable[UsedTargetRepositories]` instead of just `UsedTargetRepositories`

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
```diff
@@ -0,0 +1,679 @@
...
+                         plugin_info,
+                         target_iso=None,
+                         on_aws=False):
+    """
+    Perform RPM download including the transaction test using dnf with our plugin
+
+    :param target_userspace_info: Information about the target userspace
+    :type target_userspace_info: TargetUserSpaceInfo
+    :param storage_info: Storage and filesystem information
+    :type storage_info: StorageInfo
```
PeterMocary, line 598: Nitpick: the order doesn't match function signature. I don't really know if that matters much, just pointing it out. Same for  `perform_dry_run` and `perform_transaction_check`

PeterMocary (changes requested):
I went through the docstrings and found some discrepancies, most of which should be addressed.

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
pirat89: @PeterMocary  the order is not important. it usually matches as it's the way we naturally write write it. in this case, I copied it.

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
pirat89: hmm...a good question. I checked that with AI and it seems that typically it's a good habit to mention it explicitly. The modern way is like this:
```
    :type context: mounting.IsolatedActions | None
```

I will update it in the code where `or None` is mentioned. Otherwise I will keep it for the "mass docstrings update" PR that we will do in close future.

repos/system_upgrade/common/libraries/dnflibs/dnfmodule.py
```diff
@@ -0,0 +1,90 @@
...
+
+    module_base = dnf.module.module_base.ModuleBase(base)
+    # this method is absent on RHEL 7, in which case there are no modules anyway
+    # note the method could be removed in future as well - so keep the check
+    if not hasattr(module_base, 'get_modules'):
+        return []
+    return module_base.get_modules('*')[0]
+
+
+def get_enabled_modules():
```
matejmatuska, line 43: Since this function uses `get_modules` which as the base as an argument, wouldn't it be nicer if this function all did have base as an argument?

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
pirat89: iterable is more exact regarding the code implementation, but I see we give it a list everywhere in callers, so I will use the list instead.

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
```diff
@@ -0,0 +1,679 @@
...
+        )
+        details = {'hint': hint, 'Disk Requirements': '\n'.join(missing_space)}
+
+    raise DNFUpgradeTransactionError(message=message, details=details)
+
+
+def _transaction(context, stage, target_repoids, tasks, plugin_info, xfs_info,
+                 test=False, cmd_prefix=None, on_aws=False):
+    """
+    Perform the actual DNF rpm download via our DNF plugin
```
matejmatuska, line 262: Missing params/return val in the docstring. Not a blocker, because they are mostly self explanatory, but would be nice to have.

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
matejmatuska: Yeah the | is what's used in type hints, but Peter means that if you pass in None, the function will crash on context.makedirs() because context will be None.

It should be just:
`:type context: mounting.IsolatedActions | None`

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
```diff
@@ -0,0 +1,679 @@
...
+        )
+        json.dump(config_data, f, sort_keys=True, indent=2)
+
+
+def backup_config(context):
+    """
+    Backs up the configuration data used for the plugin.
+
+    :param context: Execution context (defaults to NotIsolatedActions)
+    :type context: mounting.IsolatedActions or None
```
matejmatuska: Same here, if context is None, the function crashes.
```suggestion
    :type context: mounting.IsolatedActions
```

repos/system_upgrade/common/libraries/dnflibs/dnfconfig.py
pirat89: and the docstring is actually wrong also from the other POV :D but it's private func so I missed that completely. fixing it

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
```diff
@@ -0,0 +1,679 @@
...
+    """
+    context.copy_from(DNF_PLUGIN_DATA_PATH, DNF_PLUGIN_DATA_LOG_PATH)
+
+
+def backup_debug_data(context):
+    """
+    Performs the backup of DNF debug data
+
+    :param context: Execution context (defaults to NotIsolatedActions)
```
matejmatuska:
```suggestion
    :param context: Execution context
```
It doesn't default to anything

repos/system_upgrade/common/libraries/dnflibs/dnfconfig.py
```diff
@@ -0,0 +1,138 @@
...
+class CannotObtainDNFConfig(DNFError):
+    """
+    Raised when cannot obtain DNF configuration.
+    """
+
+
+class CannotUpdateDNFConfig(DNFError):
+    """
+    Raised when the DNF configuration could not be updated.
+    """
```
matejmatuska, line 21: I would maybe introduce one more parent excetion for these, e.g. `DNFConfigError` if the caller doesn't want to handle all of these separately.

pirat89:
@PeterMocary  thanks for the properly done review!! :rocket:  everything should be addressed now.

matejmatuska (changes requested):
I did a quick review, a few minor problems and suggestions.

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
matejmatuska: This is still true even though it says outdated.

repos/system_upgrade/common/libraries/dnflibs/dnfmodule.py
pirat89: actually I would rather drop it everywhere in the lib :) but for the compatibility I kept it in the lib. I do not expectt anyone to set it, so I rather kept this function as it was. Let's add a parameter if someone wants it.

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
pirat89: it's private function, where we consider docstrings just as optional thing. Incorrect docstrings need to be fixed in private functions as well, but as we do not expect anyone to use private functions, it's just on us whether we add the info for ourselves or not. at this point, as the docstring is correct, just params, etc. are not part of it, I plan to keep it as it is. if we want to change it for private docstrings, it can wait for the "mass docstring update" PR. same for all other docstrings for private stuff

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
pirat89: aah. good point! the fix is on the way.

repos/system_upgrade/common/libraries/dnflibs/dnfconfig.py
pirat89: @matejmatuska  I was thinking about that for a while and to me it would start to be over-engineered. Let's wait if we or someone finds it useful and consider the current state good enough otherwise / for now.

pirat89:
@matejmatuska some required changes delivered, some are declined. -- sorry, I closed already all of them as resolved; I could at least keep these that have not been delivered opened.

matejmatuska (dismissed):
@pirat89 Okay, code looks alright now, didn't test. Let's keep the full review on @PeterMocary if that's okay.

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
```diff
@@ -0,0 +1,681 @@
...
+    }
+    return data
+
+
+def create_config(context, target_repoids, debug, test, tasks, on_aws=False):
+    """
+    Creates the configuration data file for our DNF plugin.
+
+    :param context: Execution context (defaults to NotIsolatedActions)
+    :type context: mounting.IsolatedActions | None
```
matejmatuska: My bad, this still has to go:
```suggestion
    :type context: mounting.IsolatedActions
```

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
```diff
@@ -0,0 +1,681 @@
...
+            target_repoids=target_repoids, debug=debug, test=test, tasks=tasks, on_aws=on_aws
+        )
+        json.dump(config_data, f, sort_keys=True, indent=2)
+
+
+def backup_config(context):
+    """
+    Backs up the configuration data used for the plugin.
+
+    :param context: Execution context (defaults to NotIsolatedActions)
```
matejmatuska:
```suggestion
    :param context: Execution context
```

matejmatuska (commented):
My bad, some of the comment are still not resolved

PeterMocary (commented):
LGTM! I've even tried to upgrade 9to10 manually with this patch. After the comments from @matejmatuska are adressed we can merge.

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
pirat89: I have deja vu, exactly these two changes I made, but I see they are not applied (Oo). Sorry guys, most likely PEBKAC

repos/system_upgrade/common/libraries/dnflibs/dnfplugin.py
pirat89: droped together with info about "default to NotIsolatedActions" - which is incorrect as well.

pirat89:
@matejmatuska  all should be addressed now!

@matejmatuska @PeterMocary  thank you guys for good reviews!

PeterMocary (approved):

matejmatuska (approved):

PR #1509, opened 2026-03-13
fix(livemode): make live.dir is relative to device root

MichalHe:
Booting from an squashfs image (from a local block dev) requires to parameters:
- identifier of the device where the squashfs image resides
- path to the squashfs image on the given device (aka live.dir) Therefore, live.dir needs to be set up with a path that is relative to the device where the image resides. This patch fixes the current behavior where the path is being taken as it is, i.e., relative to the root of the source system.

Jira-ref: RHEL-104379


pirat89 (approved):
lgtm and it seems to be working as expected (tested manually)

PR #1511, opened 2026-03-18
fix(overlaygen): exclude hugetlbfs from overlay mounts

holser:
hugetlbfs should not be mounted in the overlay environment during the upgrade. It's not a regular mount point. This patch adds it to the OVERLAY_DO_NOT_MOUNT exclusion list.

Resolves: RHEL-156902


pirat89 (approved):
thanks for the contribution. lgtm. postponing the merge until it's resolved the question of the ticket testing

PR #1513, opened 2026-03-27
fix(rhui): consider source clients always signed

MichalHe:
Previously, all RHUI clients mentioned in the cloud map in the rhui.py library were considered as signed by RH. However, configs allow using clients that are not known to leapp, and, therefore, such clients would not be considered as signed. Consequently, these packages would not be removed, causing the upgrade to fail. This patch changes this behavior so that all source clients mentioned in the RHUIInfo message are considered as signed.

Tests were generated using Cursor.


matejmatuska (dismissed):
I think this should be covered in unit tests.

matejmatuska (commented):
LGTM, works as expected.

matejmatuska (approved):
LGTM

PR #1514, opened 2026-03-27
[selinux][avc] leapp_resume.service: set install type context

pirat89:
The original fix to deal with AVC during execution of leapp when booting to the upgraded system appeared to leave still a space to generate new AVC messages during some operations. From our side, the only way how we could be sure that we do not hit any additional error on top of what root user would see, is to set an unconfined context.

However, after discussion with SELinux guys, we decided to not do that yet. Currently dealing with the RH Insights client, it has been discovered that bug is in new SELinux policies for the client and not in leapp-repository. So for now let's keep the situation as it is and wait for the proper fix in other components.

Also after the discussion, we are going to switch to the install_exec type context which should fits to our process better than the original one. We will monitor the situation around AVC messages and make possible further changes based on the feedback.

Jira: RHEL-161684

### DIFF
--- a/repos/system_upgrade/common/actors/createresumeservice/files/leapp_resume.service
+++ b/repos/system_upgrade/common/actors/createresumeservice/files/leapp_resume.service
@@ -8,8 +8,15 @@ Wants=network-online.target
 
 [Service]
 Type=oneshot
-# FIXME: this is temporary workaround for Python3
-ExecStart=/bin/sh -c "/root/tmp_leapp_py3/leapp3 upgrade --resume"
+# NOTE(pstodulk): After discussion with the SELinux team, we try to stay
+# out of unconfined mode for now.
+#SELinuxContext=unconfined_u:unconfined_r:unconfined_t:s0
+
+# NOTE(pstodulk): Correction of the selinux type context. This one should fit to leapp
+# Ignore the exit code as this cmd is expected to fail if selinux is disabled
+ExecStartPre=-chcon -t install_exec_t /root/tmp_leapp_py3/leapp3
+ExecStart=/root/tmp_leapp_py3/leapp3 upgrade --resume
 StandardOutput=journal+console
-# FIXME: this shouldn't be needed, but Satellite upgrade runs installer, and that's slow
+
+# NOTE(pstodulk): some actions can take a lot of time (e.g. satellite installer)
 TimeoutStartSec=infinity


repos/system_upgrade/common/actors/createresumeservice/files/leapp_resume.service
```diff
@@ -8,8 +8,14 @@ Wants=network-online.target
 
 [Service]
 Type=oneshot
-# FIXME: this is temporary workaround for Python3
-ExecStart=/bin/sh -c "/root/tmp_leapp_py3/leapp3 upgrade --resume"
+# NOTE(pstodulk): After discussion with the SELinux team, we try to stay
+# out of unconfined mode for now.
+#SELinuxContext=unconfined_u:unconfined_r:unconfined_t:s0
+
+# NOTE(pstodulk): Correct the selinux type context. This one should fit to leapp
```
matejmatuska: "Correct" sounds like it's a TODO

pirat89: Good point, I will change it.

pirat89: done

PeterMocary (dismissed):
LGTM!

The `/root/tmp_leapp_py3/leapp3` has the context correctly changed:
```bash
ls -Z /root/tmp_leapp_py3/leapp3
system_u:object_r:install_exec_t:s0 /root/tmp_leapp_py3/leapp3
```

PeterMocary (changes requested):
I was too hasty to approve, when running on a system with selinux disabled the service fails to start. 

```
journalctl -u leapp_resume.service
May 25 19:59:45 leapp-20260525212508 systemd[1]: Starting Temporary Leapp service which resumes execution after reboot...
May 25 19:59:45 leapp-20260525212508 chcon[1344]: chcon: can't apply partial context to unlabeled file '/root/tmp_leapp_py3/leapp3'
May 25 19:59:45 leapp-20260525212508 systemd[1]: leapp_resume.service: Control process exited, code=exited, status=1/FAILURE
May 25 19:59:45 leapp-20260525212508 systemd[1]: leapp_resume.service: Failed with result 'exit-code'.
May 25 19:59:45 leapp-20260525212508 systemd[1]: Failed to start Temporary Leapp service which resumes execution after reboot.
```

pirat89:
@PeterMocary good catch!! You are right, I will fix it

pirat89:
@PeterMocary  fixed

PeterMocary (approved):
Now the service starts even if the `chcon` command fails:

```
May 26 09:29:30 leapp-20260526110131 systemd[1]: Starting Temporary Leapp service which resumes execution after reboot...
May 26 09:29:30 leapp-20260526110131 chcon[1358]: chcon: can't apply partial context to unlabeled file '/root/tmp_leapp_py3/leapp3'
May 26 09:29:31 leapp-20260526110131 leapp3[1361]: 2026-05-26 09:29:31.690 INFO     PID: 1361 leapp: Logging has been initialized
May 26 09:29:31 leapp-20260526110131 leapp3[1361]: 2026-05-26 09:29:31.694 INFO     PID: 1361 leapp: Resuming execution after phase: Finalization
[...]
```

PR #1517, opened 2026-04-02
Ensure that greenboot rpms are removed during IPU

pirat89:
The greenboot packages have a value just for rpm-ostree based systems. However, if someone install them anyway (e.g. configuration mistake..), they would damage the upgraded systems - mainly in case IPU 8.10 -> 9.8+. Systemd fails with error:
    ```
    Failed to isolate the default target.
    ```

Due to bugs in greenboot scriptlets, that do not count with DNF execution inside container, nor with the update of greenboot packages from 0.14.x to 0.16.x and newer.

As there is not value to have these packages on rpm based system at all, just remove them during the upgrade to stay on a safe side.


matejmatuska (approved):
Makes sense

PR #1510, opened 2026-03-17
fix quoting in list representation of commands

PeterMocary:
Removed redundant quotes included due to incorrect handling of command conversion to string representation in leapp framework's reporting module. The reporting module is fixed by RHEL-156521 and this patch is a followup that fixes usage of the module so that the commands are propagated to the leapp-report.txt correctly.

### Example with both leapp and leapp-repository PRs

leapp-report.txt
```
Risk Factor: high (inhibitor) 
Title: Upgrade requires links in root directory to be relative 
Summary: After rebooting, parts of the upgrade process can fail if symbolic links in / point to absolute paths. 
Please change these links to relative ones. 
Related links: 
    - leapp upgrade stops with Inhibitor "Upgrade requires links in root directory to be relative": https://access.redhat.com/solutions/6989732 
Remediation: [command] sh -c 'ln -snf usr/bin /bin && ln -snf usr/lib /lib' 
Key: 3d895ad37ceaf4157864d439edb6bd75562061fa 
---------------------------------------- 
```

leapp-report.json
```
    {
      "audience": "sysadmin",
      "detail": {
        "external": [
          {
            "title": "leapp upgrade stops with Inhibitor \"Upgrade requires links in root directory to be relative\"",
            "url": "https://access.redhat.com/solutions/6989732"
          }
        ],
        "remediations": [
          {
            "context": [
              "sh",
              "-c",
              "ln -snf usr/bin /bin && ln -snf usr/lib /lib"
            ],
            "type": "command"
          }
        ]
      },
      "flags": [
        "inhibitor"
      ],
      "key": "3d895ad37ceaf4157864d439edb6bd75562061fa", 
      "severity": "high", 
      "summary": "After rebooting, parts of the upgrade process can fail if symbolic links in / point to absolute paths.\nPlease change these links to relative ones.", 
      "title": "Upgrade requires links in root directory to be relative", 
      "timeStamp": "2026-03-17T12:23:38.556773Z", 
      "hostname": "leapp-20260313125304", 
      "actor": "check_root_symlinks", 
      "id": "7d784af5714dfbe4a1d13e17ab97163a2139bf32336dd7c665756eca5d5b41d9", 
      "tags": [] 
    }, 
```

Both have correct equivalent representation of the command.

Jira: RHEL-155517, RHEL-155515

PR #1427, opened 2025-08-21
Remove 7->8 related code

matejmatuska:
Don't worry I will split it to more PRs :D

PR #1518, opened 2026-04-02
fix(checkyumpluginsenabled): correctly create remediation command

PeterMocary:
The remediation command for the inhibitor triggered when required dnf plugins are not enabled was created incorrectly. This patch fixes the command creation.

This is a regression that was introduced during removal of 7to8 related code [here](https://github.com/oamg/leapp-repository/pull/1427/changes/84cefba4c2e03a261f46effe507cac1ae25ecc87#diff-8a05c97750aae910c79df435fed901fccc03458198b3de72545e9e0c2a52d02fL48-L50). It even slipped me when updating the commands recently ([PR#1510](https://github.com/oamg/leapp-repository/pull/1510)) since it is just a few missing commas.


matejmatuska (approved):
Good catch!

PR #1519, opened 2026-04-10
Add scripts for generating type stubs reflecting runtime module layout

matejmatuska:
Due to the various dynamic import mechanism that the leapp framework uses to import leapp entities, development tools such as language servers or type checkers work very unreliably if at all.

This PR adds scripts for generation of type stubs for leapp framework components, namely: actor libs, common libs, models, tags, topics (no actors, they are never imported). These generated type stub files are stored in a `typings` directory in a layout which simulates the import hierarchy at runtime, i.e. a common library defined at `repos/system_upgrade/common/libraries/grub.py` has a type stub at `typings/leapp/libraries/common/grub.pyi`.

When tools such as the Pylance language server (the VS code extension for Python) or pyright (upstream of Pylance) are pointed at the type stub directory, they prefer it over the real source code. These stubs are then used as a source of information which brings (more like fixes) several **features**: 
- imports are resolved correctly
- autocompletion
  - when autocompleting something it can be auto-imported
- type information (mostly for models) and docstrings on mouse hover or other such places
- better type checking (mostly for models)
- method signature help

There are also some **limitations**, however no regressions, I believe:
- go-to-definition and go-to-references or similar navigation features navigate to the type stub file instead of the real file
- the content of the typings directory is currently static, i.e. you have to live with some errors until you regenerate it again
  - generation takes about 5-6s on my machine
  - this could be further sped up by only regenerating what changed based on e.g. git rev

### Usage
1. If you use pyright or pylance (VS code extension) the included pyrightconfig.json should get picked up by that and the `typings` directory will be used automatically. If you use anything else e.g. jedi, mypy, etc. You might need to point to the `typings` directory, but seems like the name is standard.
2. Run `make stubgen` to generate.
3. Enjoy

### Technical notes
All of the stubs are generated using `mypy.stubgen`, except for the model ones. Those are generated using a custom script with the ast module, because of how models are implemented (metaclasses and class properties). This allows offering much better type information because the scripts converts the leapp field types to actual standard python type hints (whereas mypy only generates `_typeshed.Incomplete`). Also, docstrings for the fields are preserved (though due to them not being standardized the tools might not show them). 

There is also a `.stub_tmp` directory which is a temporary dictionary to where the direct outputs of the generators are put before they are assembled. The assembly is a process in which multiple stubs are combined to a single `__init__.py` for a flat module hierarchy and import statements are inserted, this applies to models, tags and topics. Then all the stubs are moved to the correct places.     

### Other solutions considered
I tried to run the stub generation tool (`mypy.stubgen`) after actually loading the upgrade leapp repositories via leapp, however due to how the imports work and the different dynamic import schemes used for models, libraries and tags and topics, this requires a lot of workarounds and usage of the lower level stubgen apis such as `InspectionStubGenerator`. 
It is also slower due to the loading of repositories.

### Examples
- Autocompletion:
image

- Imports are resolved (note the warning is that it could not be resolved from source, i.e. it is resolved from the type stub file)
image

- Signature hint with types and docstring:
image
  
  - In VS code:
2026-02-25T18:25:42,728273978+01:00


utils/stubgen/gen_stubs.py
```diff
@@ -0,0 +1,104 @@
...
+# see gen_stubs.sh
+# All the generation is kept in a python script to save some time on
+# interpreter executions and initialization.
+
+import argparse
+import sys
+from pathlib import Path
+
+import gen_model_stubs
+from mypy import stubgen
```
pirat89: @matejmatuska  This can fail if mypy is not present on the system and if I see right, there is nothing informing people about that in the Makefile - nor rule for installation of required deps. I think that it will be nice to add some checks to Makefile as we do e.g. in `install-deps-fedora`. it's possible to do a check whether the `mypy` module can be imported or not. for sure, I would expect some errors handling the import problem here as well as it's not standard library.

(this is not full review)

utils/stubgen/gen_stubs.py
```diff
@@ -0,0 +1,104 @@
...
+        doc_dir="",
+        interpreter="/usr/bin/python",
+    )
+
+    try:
+        stubgen.generate_stubs(opts)
+    except SystemExit as e:
+        # stubgen calls sys.exit(0) on success; we catch it to prevent
+        # the entire script from exiting.
+        print("here")
```
pirat89: forgotten debug print? :)

utils/stubgen/gen_stubs.py
matejmatuska: Okay, makes sense, I added a check for an import error, now it fails like this without mypy.
```> make stubgen                                                        gen_stubs_v2 U 0
./utils/stubgen/build_stubs.sh .stubs_tmp typings
Error: Failed to import mypy.stubgen make sure mypy is installed.
Error: Stub generation failed.
make: *** [Makefile:553: stubgen] Error 1
```
One error is from the generation script and the other one from the "build" script.

pirat89 (changes requested):
just preventive block - to no forget to fix commits before the merge :-D

matejmatuska:
Just to clarify, this doesn't make mypy work very good. mypy is not very good at merging definitions from the source file with definitions in the type stub file (pyright is quite good at that). Furthermore the leapp-repository codebase as well as the leapp framework codebase have almost no type annotations.

It is possible to make it work relatively OK, however there is a lot of noise.

That said it works quite nice with pyright, which was the primary goal of the work. We can expand in the future.

docs/source/tutorials/index.rst
```diff
@@ -18,6 +18,7 @@ write leapp actors for **In-Place Upgrades (IPU)** with the leapp framework.
     :glob:
 
     setup-devel-env
+    setup-devel-tools
```
pirat89: @matejmatuska  thematically  it should be be part of setup-devel-env

docs/source/tutorials/setup-devel-tools.md
```diff
@@ -0,0 +1,19 @@
+# Making development tools work
```
pirat89: We should use a different title. What about something like: `Tweaks for type checking and editors`?

pirat89:
> Just to clarify, this doesn't make mypy work very good. mypy is not very good at merging definitions from the source file with definitions in the type stub file (pyright is quite good at that). Furthermore the leapp-repository codebase as well as the leapp framework codebase have almost no type annotations.
> 
> It is possible to make it work relatively OK, however there is a lot of noise.
> 
> That said it works quite nice with pyright, which was the primary goal of the work. We can expand in the future.

good to know!

docs/source/tutorials/index.rst
matejmatuska: Yeah I was thinking about that, but I am not sure if it fits there, the dev env is means something a little different there.

docs/source/tutorials/setup-devel-tools.md
matejmatuska: Yeah could be that.

utils/stubgen/gen_stubs.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/python
```
MichalHe, line 1: This requires mypy to be installed on the system globally. The following change allows installing mypy just into development venv (or globally, its up to the user). 
```suggestion
#!/usr/bin/env python3
```

MichalHe (commented):
LGTM. I ask only for one quality-of-life change and we can merge it. Also I think commits could/should be squashed into one.

docs/source/tutorials/index.rst
MichalHe: The PR now modifies `setup-devel-env`, but the index change still refers to `setup-devel-tools`.

MichalHe (changes requested):

docs/source/tutorials/index.rst
matejmatuska: Thanks, good catch

matejmatuska:
@MichalHe I believe everything is addressed and I also squashed the commits, I think it's ready to merge

MichalHe (approved):
I actively use the generated type stubs, they are very helpful. Shame there are almost no type hints in the codebase, it would be even better.

LGTM

pirat89:
> I actively use the generated type stubs, they are very helpful. Shame there are almost no type hints in the codebase, it would be even better.

I think that we will have a time to improve it the next release :)

PR #1522, opened 2026-04-14
python test deps: update required py packages per python

pirat89:
tl;dr; We haven't updated the requirements.txt file for a long time
(basically since IPU 7 -> 8; if we do not count some minor updates).
Let's update dependencies to use newer python modules for newer
versions of python.

* Pytest
  Originally we required pytest v6.2.5 for any python 3.6+. This is
  pretty old requirement coming from time of python ~ 3.8 and recently
  a https://github.com/advisories/GHSA-6w46-j5rx-g56g occuared, which is fix in pytest v9.0.3.
  So let's update ranges to actually allow use of newer pytest on newer
  systems:

        +----------------+---------------+
        | pytest version | python range  |
        +----------------+---------------+
        | 6.2.5          | 3.6.x - 3.7.x |
        | 8.3.5          | 3.8.x - 3.9.x |
        | 9.0.3          | 3.10+         |
        +----------------+---------------+

* Pyudev
  The original v0.22 is used up to CS 9 (pyton 3.9-). Use 0.24.1 for
  python 3.10+ (reflects CS 10 as well)

* Distro
  The original v1.5.0 can be used till CS 9 (python 3.9 included) safely
  as nowadays. Let's use version 1.9.0 for Python 3.10+

* Mock
   Drop the dependency + update imports in unit tests to use mock from std library.
   Covers https://github.com/oamg/leapp-repository/pull/1523

* Drop unsued dependencies (old python that is no longer supported
  in the project).

### TODO:
- [x] discover why unit tests for python3.9 el8toel9 are randomly failing on import
  - just when I decided that it was pebkac, I was able to reproduce it. it seems that there is a race condition when using newer python modules.
  - It seems to be problem between Python 3.9 and Pytest 8. Both have changes related to imports and mainly pytest is the source of problems for projects which have own handling of imports implemented. Seems that Pytest 7 works well on Python 3.9. I will give it several rounds before marking it ready for merge.
  - Confirmed. Pytest 7.x.x is safe to use with Python 3.9

### DIFF
--- a/commands/tests/test_upgrade_paths.py
+++ b/commands/tests/test_upgrade_paths.py
@@ -1,7 +1,7 @@
 import os
 import resource
+import unittest.mock as mock
 
-import mock
 import pytest
 
 from leapp.cli.commands import command_utils
--- a/repos/system_upgrade/common/actors/biosdevname/tests/test_biosdevname.py
+++ b/repos/system_upgrade/common/actors/biosdevname/tests/test_biosdevname.py
@@ -1,7 +1,8 @@
+from unittest.mock import mock_open, patch
+
 import pytest
 import pyudev
 import six
-from mock import mock_open, patch
 
 from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.actor import biosdevname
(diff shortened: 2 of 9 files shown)


pirat89:
one of tests is randomly failing with updated deps. maybe a race condition?  executed 4 times, 2times failed.

matejmatuska (approved):
LGTM! Let's wait for the tests

PR #1523, opened 2026-04-15
Drop dependency on mock

matejmatuska:
The mock library is now a part of the Python standard library since Python 3.3 and is available as the unittest.mock. The 'mock' library is only a backport which is not needed anymore.

Also remove some unnecessary imports.


pirat89:
@matejmatuska nice! I cherrypicked the commit to #1522 so it's resolved together with other updates in requirements.txt file.

PR #1524, opened 2026-04-16
mounting: ignore failures when callling mount -a

MichalHe:
The `mount -a` calls are implemented just as a safeguard. However, the semantics of `nofail` as seen by `mount` are different as the ones seen by `systemd` in that they can result in a non-zero exit code e.g. if the block device does not contain a valid filesystem. In contrast, systemd threads `nofail` so that the unit does not fail (even if the filesystem is broken). This patch adds error handling that ignores non-zero exit codes.

Jira-ref: RHEL-56176


pirat89:
@MichalHe I am not 100% sure now whether `mount -a` in dracut module is ignored safely or not. So I would rather make it clear. wdyt?

repos/system_upgrade/common/actors/initramfs/mount_units_generator/libraries/mount_unit_generator.py
```diff
@@ -110,6 +110,29 @@ def prefix_all_mount_units_with_sysroot(dir_containing_units):
...
+    for line in lines:
+        if line.startswith('Options'):
+            line_fragments = line.split('=', 1)
+            if len(line_fragments) <= 1:
+                continue  # Should not happen
+            used_options = [opt.strip() for opt in line_fragments[1].split(',')]
+            if 'nofail' in used_options:
+                has_nofail = True
+                break
+    return has_nofail
```
pirat89: @MichalHe nitpick, what about this?

```suggestion
    for line in lines:
        if line.startswith('Options'):
            line_fragments = line.split('=', 1)
            if len(line_fragments) <= 1:
                continue  # Should not happen
            used_options = [opt.strip() for opt in line_fragments[1].split(',')]
            if 'nofail' in used_options:
                return True
    return False
```

MichalHe: thank you, suggestion applied.

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
```diff
@@ -236,7 +236,8 @@ do_upgrade() {
     # NOTE: in case we would need to run leapp before pivot, we would need to
     #       specify where the root is, e.g. --root=/sysroot
     # TODO: update: systemd-nspawn
-    /usr/bin/systemd-nspawn $NSPAWN_OPTS -D "$NEWROOT" /usr/bin/bash -c "mount -a; $LEAPPBIN upgrade --resume $args"
+    # shellcheck disable=SC2086  # we do not quote NSPAWN_OPTS for a purpose here
```
PeterMocary: Nitpick: I'd change the comment so that it states the purpose.

```suggestion
    # shellcheck disable=SC2086  # The NSPAWN_OPTS and args variables are not quoted since they are supposed to expand into multiple arguments
```

Same thing for the other `systemd-nspawn` command.

repos/system_upgrade/common/actors/initramfs/mount_units_generator/tests/test_mount_unit_generation.py
```diff
@@ -326,3 +326,19 @@ def listdir_mocked(path):
...
+
+
+@pytest.mark.parametrize(
+    ('unit_filename', 'is_nofail'),
+    [
+        ('unit.mount', False),
+        ('unit-nofail.mount', True),
+    ]
+)
+def test_is_mount_unit_is_nofail(monkeypatch, unit_filename, is_nofail):
```
PeterMocary: Nitpick: The test name seems malformed?
```suggestion
def test_is_unit_marked_with_nofail(monkeypatch, unit_filename, is_nofail):
```

repos/system_upgrade/common/actors/initramfs/mount_units_generator/tests/test_mount_unit_generation.py
```diff
@@ -326,3 +326,19 @@ def listdir_mocked(path):
...
+
+TEST_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'files')
+
+
+@pytest.mark.parametrize(
+    ('unit_filename', 'is_nofail'),
+    [
+        ('unit.mount', False),
+        ('unit-nofail.mount', True),
+    ]
```
PeterMocary, line 340: Nitpick: We could add here the case where the file does not exist.

repos/system_upgrade/common/actors/removeupgradebootentry/libraries/removeupgradebootentry.py
```diff
@@ -45,9 +45,10 @@ def remove_boot_entry():
 
     # TODO: Move calling `mount -a` to a separate actor as it is not really related to removing the upgrade boot entry.
     #       It's worth to call it after removing the boot entry to avoid boot loop in case mounting fails.
-    run([
-        '/bin/mount', '-a'
-    ])
+    try:
+        run(['/bin/mount', '-a'])
+    except CalledProcessError:
+        pass
```
PeterMocary, line 57: Nitpick: To me, seeing except pass in code raises questions, so maybe adding a comment here would make it better for people who will read it without the context of this PR description.

Same for the other similar change in el8toel9 repository.

repos/system_upgrade/common/actors/initramfs/mount_units_generator/libraries/mount_unit_generator.py
```diff
@@ -110,6 +110,27 @@ def prefix_all_mount_units_with_sysroot(dir_containing_units):
...
 
+def is_unit_marked_with_nofail(unit_path: str) -> bool:
+    try:
+        with open(unit_path) as in_file:
+            lines = [line.strip() for line in in_file.readlines()]
+    except OSError:
+        api.current_logger().debug(
+            'Could not read the unit {} to determine whether it is nofail.'.format(unit_path)
+        )
+        return False  # This way, the unit will end up as .required, so it is a safe choice
```
PeterMocary: I don't understand this comment, it seems like the result of this function will impact the unit by modifying it with .required, but I don't see where would that happen. 

I see that the function call result is used only to skip fixing of symlinks. When the is_nofail is false it even skips the evaluation of `will_units_be_required = target_dir.endswith('.requires')`  (which checks for `.requires` not `.required`) and proceeds to fix the symlinks.

I might be just missing the big picture here, but it would be great if you could explain this a bit more.

PeterMocary (commented):
I've found mostly nitpicks (just ignore ones you don't like), but I'm not entirely sure about the `.required` comment in the `is_unit_marked_with_nofail` (see in comments).

repos/system_upgrade/common/actors/initramfs/mount_units_generator/libraries/mount_unit_generator.py
matejmatuska: "Fixing" symlinks is a strong word here, we are not really fixing them 100% properly. 
What is happening here is that when a unit is `nofail` and we are currently "fixing" (creating) symlinks in the `<target>.requires` directory, this unit is omitted (and will end up in a `<target>.wants` directory as `nofail` units should).

If the function returns `False` (the unit is not `nofail`), the symlink is correctly created in `.requires` and even in `.wants` but that's not an issue apparently as that was already happening for `sysroot-boot.unit`.

The ".required" in the comment is not really a thing. I guess @MichalHe just created a past tense of ".requires" :)

repos/system_upgrade/common/actors/removeupgradebootentry/libraries/removeupgradebootentry.py
pirat89: maybe we could add also warning log as well. in future, we might remove all `mount -a` calls.

matejmatuska:
TLDR: manual test passed
I did a test with reproducer from the ticket.
The unit is correctly created in the `.wants` dir:
```
2026-04-24 12:27:53.732 DEBUG    PID: 112095 leapp.workflow.InterimPreparation.mount_unit_generator: External command has finished: ['ln', '-s', '../sysroot-mnt.mount', '/var/lib/leapp/tmp_systemd_fstab_e7m2_skw/local-fs.target.wants/sysroot-mnt.mount']
```
while other units are in `.required` as expected:
```
2026-04-24 12:27:53.719 DEBUG    PID: 112095 leapp.workflow.InterimPreparation.mount_unit_generator: External command has finished: ['ln', '-s', '../boot.mount', '/var/lib/leapp/tmp_systemd_fstab_e7m2_skw/local-fs.target.requires/boot.mount']
```
after a reboot to the upgrade initramfs the `sysroot-mnt.mount` unit fails, because the device has corrupted superblock, but the boot is successful:
```
Apr 24 12:46:04 localhost systemd[1]: Mounted /sysroot.
Apr 24 12:46:04 localhost systemd[1]: Reached target Initrd Root File System.
Apr 24 12:46:04 localhost systemd[1]: Mounting /sysroot/boot...
Apr 24 12:46:04 localhost systemd[1]: Mounting /sysroot/mnt...
Apr 24 12:46:04 localhost systemd[1]: Starting Mountpoints Configured in the Real Root...
Apr 24 12:46:04 localhost kernel: XFS (vda1): Mounting V5 Filesystem 023e51f5-b590-4de5-ab21-cf33eca191b3
Apr 24 12:46:04 localhost kernel: XFS (vdb): Invalid superblock magic number
Apr 24 12:46:04 localhost mount[649]: mount: /sysroot/mnt: wrong fs type, bad option, bad superblock on /dev/vdb, missing codepage or helper program, or other error.
Apr 24 12:46:04 localhost systemd[1]: sysroot-mnt.mount: Mount process exited, code=exited, status=32/n/a
Apr 24 12:46:04 localhost systemd[1]: sysroot-mnt.mount: Failed with result 'exit-code'.
Apr 24 12:46:04 localhost systemd[1]: Failed to mount /sysroot/mnt.
Apr 24 12:46:04 localhost kernel: XFS (vda1): Ending clean mount
Apr 24 12:46:04 localhost kernel: xfs filesystem being mounted at /sysroot/boot supports timestamps until 2038 (0x7fffffff)
Apr 24 12:46:04 localhost systemd[1]: Mounted /sysroot/boot.
Apr 24 12:46:04 localhost systemd[1]: Mounting /boot...
Apr 24 12:46:04 localhost systemd[1]: initrd-parse-etc.service: Deactivated successfully.
Apr 24 12:46:04 localhost systemd[1]: Finished Mountpoints Configured in the Real Root.
Apr 24 12:46:04 localhost systemd[1]: Mounted /boot.
Apr 24 12:46:04 localhost systemd[1]: Reached target Initrd File Systems.
Apr 24 12:46:04 localhost systemd[1]: Reached target Local File Systems.
Apr 24 12:46:04 localhost systemd[1]: Reached target System Initialization.
Apr 24 12:46:04 localhost systemd[1]: Reached target Basic System.
Apr 24 12:46:04 localhost systemd[1]: Reached target System Upgrade.
Apr 24 12:46:04 localhost systemd[1]: Starting dracut mount hook...
Apr 24 12:46:04 localhost systemd[1]: Started Hardware RNG Entropy Gatherer Daemon.
Apr 24 12:46:04 localhost systemd[1]: Starting System Upgrade...
```  
Now the upgrade actually get's to any mount -a call. The `mount -a` calls in initramfs are correctly ignored at both places:
```
Apr 24 12:46:04 localhost systemd[1]: Starting System Upgrade...
Apr 24 12:46:04 localhost upgrade[668]: starting upgrade hook
Apr 24 12:46:04 localhost upgrade[668]: /bin/upgrade: line 20: /sysroot/var/tmp/system-upgrade.state: Read-only file system
Apr 24 12:46:04 localhost systemd[1]: Finished dracut mount hook.
Apr 24 12:46:04 localhost systemd[1]: Starting dracut pre-pivot and cleanup hook...
Apr 24 12:46:04 localhost kernel: xfs filesystem being remounted at /sysroot supports timestamps until 2038 (0x7fffffff)
Apr 24 12:46:04 localhost systemd[1]: Finished dracut pre-pivot and cleanup hook.
Apr 24 12:46:04 localhost systemd[1]: Starting Cleaning Up and Shutting Down Daemons...
Apr 24 12:46:04 localhost systemd[1]: initrd-cleanup.service: Deactivated successfully.
Apr 24 12:46:04 localhost systemd[1]: Finished Cleaning Up and Shutting Down Daemons.
Apr 24 12:46:04 localhost upgrade[734]: mount: /mnt: wrong fs type, bad option, bad superblock on /dev/vdb, missing codepage or helper program, or other error.
Apr 24 12:46:04 localhost kernel: XFS (vdb): Invalid superblock magic number
``` 
and
```
Apr 24 12:51:53 localhost upgrade[36802]: mount: /mnt: wrong fs type, bad option, bad superblock on /dev/vdb, missing codepage or helper program, or other error.
Apr 24 12:51:53 localhost kernel: XFS (vdb): Invalid superblock magic number
Apr 24 12:51:53 localhost upgrade[36801]: 2026-04-24 12:51:53.919 INFO     PID: 1 leapp: Logging has been initialized
Apr 24 12:51:53 localhost upgrade[36801]: 2026-04-24 12:51:53.920 INFO     PID: 1 leapp: Resuming execution after phase: RPMUpgrade
```
then in remove_upgrade_boot_entry which  fails due to the corrupted superblock,and leapp correctly ignores the error and continues with upgrade:
```
Apr 24 12:46:13 localhost upgrade[1293]: 2026-04-24 12:46:13.833 DEBUG    PID: 556 leapp.workflow.InitRamStart.remove_upgrade_boot_entry: External command has started: ['/bin/mount', '-a']
Apr 24 12:46:13 localhost kernel: XFS (vdb): Invalid superblock magic number
Apr 24 12:46:13 localhost upgrade[1293]: mount: /mnt: wrong fs type, bad option, bad superblock on /dev/vdb, missing codepage or helper program, or other error.
Apr 24 12:46:13 localhost upgrade[1293]: 2026-04-24 12:46:13.865 DEBUG    PID: 556 leapp.workflow.InitRamStart.remove_upgrade_boot_entry: Command ['/bin/mount', '-a'] failed with exit code 32.
Apr 24 12:46:13 localhost upgrade[1293]: 2026-04-24 12:46:13.866 DEBUG    PID: 556 leapp.workflow.InitRamStart.remove_upgrade_boot_entry: External command has finished: ['/bin/mount', '-a']
```

matejmatuska:
I will do a code review pass also, please address @PeterMocary's review in the meantime I think we can merge if I don't find some hidden blocker.

repos/system_upgrade/common/actors/commonleappdracutmodules/files/dracut/85sys-upgrade-redhat/do-upgrade.sh
MichalHe: Thank you. Fixed :+1:

repos/system_upgrade/common/actors/initramfs/mount_units_generator/tests/test_mount_unit_generation.py
MichalHe: I have added the test case. Originally, I have deemed the test case as needless since we generate the units, list what we have generated, and only then check whether these units are nofail, i.e., there is no room for the units to disappear (except for bad disks etc.).

repos/system_upgrade/common/actors/initramfs/mount_units_generator/tests/test_mount_unit_generation.py
MichalHe: Thanks, fixed.

repos/system_upgrade/common/actors/removeupgradebootentry/libraries/removeupgradebootentry.py
MichalHe: I have added both a warning and a comment explaining why we are ignoring the error.

repos/system_upgrade/common/actors/initramfs/mount_units_generator/libraries/mount_unit_generator.py
MichalHe: I have removed the past tense, which was used to make the comment more human. Technically, the symlink is created in `.requires` which makes the unit _required_ by the target.

I feel like there are no code changes (e.g. doc strings) required from me here. Am I correct? @PeterMocary

matejmatuska (approved):
Look good and also works as expected, test results above. Approving

PeterMocary (approved):

PR #1525, opened 2026-04-17
repofileutils: disable ConfigParser interpolation when parsing .repo files

prjagtap-rh:
repofileutils: disable ConfigParser interpolation when parsing .repo files

This occurs when parsing repository definitions containing percent encoded values (e.g. '%252B'in baseurls). Python's ConfigParser enables interpolation by default and treats '%' as a special token, which causes valid URLs to fail during parsing.

Disable interpolation by using parse_config(..., no_interpolation=True) so that repository values are handled literally, consistent with DNF  or YUM behavior.

Jira-ref: RHEL-162328, RHEL-148631

### DIFF
--- a/repos/system_upgrade/common/libraries/utils.py
+++ b/repos/system_upgrade/common/libraries/utils.py
@@ -10,7 +10,7 @@ from leapp.libraries.stdlib import api, CalledProcessError, config, run, STDOUT
 from leapp.utils.deprecation import deprecated
 
 
-def parse_config(cfg=None, strict=True):
+def parse_config(cfg=None, strict=True, no_interpolation=False):
     """
     Applies a workaround to parse a config file using py3 AND py2
 
@@ -18,10 +18,20 @@ def parse_config(cfg=None, strict=True):
     the old ones (Py2) obsoletes, these function was created to use the
     ConfigParser on Py2 and Py3
 
+    When no_interpolation is True, values are read literally. Use this for
+    inputs that may contain percent-encoded URLs (e.g. yum .repo baseurl),
+    since default ConfigParser treats ``%`` as interpolation syntax.
+
     :type cfg: str
     :type strict: bool
+    :type no_interpolation: bool
     """
-    if six.PY3:
+    if no_interpolation:
+        if six.PY3:
+            parser = six.moves.configparser.RawConfigParser(strict=strict)  # pylint: disable=unexpected-keyword-arg
+        else:
+            parser = six.moves.configparser.RawConfigParser()
+    elif six.PY3:
         parser = six.moves.configparser.ConfigParser(strict=strict)  # pylint: disable=unexpected-keyword-arg
     else:
         parser = six.moves.configparser.ConfigParser()
--- a/repos/system_upgrade/common/libraries/utils.py
+++ b/repos/system_upgrade/common/libraries/utils.py
@@ -1,8 +1,7 @@
 import functools
 import os
 import sys
-
-import six
+from configparser import ConfigParser, RawConfigParser
 
 from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.common import mounting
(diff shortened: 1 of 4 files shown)


repos/system_upgrade/common/libraries/utils.py
```diff
@@ -10,18 +10,28 @@
...
     :type cfg: str
     :type strict: bool
+    :type no_interpolation: bool
     """
-    if six.PY3:
+    if no_interpolation:
+        if six.PY3:
+            parser = six.moves.configparser.RawConfigParser(strict=strict)  # pylint: disable=unexpected-keyword-arg
+        else:
+            parser = six.moves.configparser.RawConfigParser()
```
pirat89: no need to check for PY3 anymore as the code will be always executed with python3. the sifx library does not have to be even used here at all. drop the check for py3

prjagtap-rh: Made required changes

repos/system_upgrade/common/libraries/utils.py
```diff
@@ -1,45 +1,37 @@
...
-    ConfigParser has a new def to read strings/files in Py3, making
-    the old ones (Py2) obsoletes, these function was created to use the
-    ConfigParser on Py2 and Py3
+    When no_interpolation is True, values are read literally. Use this for
+    inputs that may contain percent-encoded URLs (e.g. yum .repo baseurl),
+    since default ConfigParser treats ``%`` as interpolation syntax.
 
     :type cfg: str
     :type strict: bool
+    :type no_interpolation: bool
```
matejmatuska: Let's describe a bit more the internals, to provide more context to the caller. Also document the option in the docstring:

```suggestion
    When no_interpolation is True, RawConfigParser is used internally to
    read values literally. Use this fort inputs that may contain URLs
    which might contain '%' sign due to URL encoding (e.g. yum .repo
    baseurl).
    If set to False, the default ConfigParser is used which treats '%' 
    as interpolation syntax.

   :type cfg: str
   :type strict: bool
   :param no_interpolation: If True, values are read literally
   :type no_interpolation: bool
```

matejmatuska (changes requested):
Add a unit test for parse_repofile() to test that URL encoding characters (e.g. the percent sign) are parsed properly.

You can also add a test for parse_config() function.

I did a quick manual test using the reproducer in the ticket and this seems to fix it.

repos/system_upgrade/common/libraries/utils.py
```diff
@@ -1,45 +1,41 @@
...
     """
-    Applies a workaround to parse a config file using py3 AND py2
+    Parse a config file or string
 
-    ConfigParser has a new def to read strings/files in Py3, making
-    the old ones (Py2) obsoletes, these function was created to use the
-    ConfigParser on Py2 and Py3
+    When no_interpolation is True, RawConfigParser is used internally to
+    read values literally. Use this fort inputs that may contain URLs
+    which might contain '%' sign due to URL encoding (e.g. yum .repo
```
PeterMocary: Found a typo.

Nitpick:  I'd also reword it a bit.
```suggestion
    read values literally. Use this for inputs that may contain URLs
    with '%' sign for URL encoding (e.g. yum .repo
```

repos/system_upgrade/common/libraries/utils.py
```diff
@@ -1,45 +1,41 @@
...
+    baseurl).
+    If set to False, the default ConfigParser is used which treats '%'
+    as interpolation syntax.
 
-    :type cfg: str
-    :type strict: bool
+   :type cfg: str
+   :type strict: bool
```
PeterMocary: While at it, I'd also include the `:param` description for both of these and the `cfg` can be a file-like object or a string.

Since this is simply a wrapper, I'd include the relevant information from the `configparser` docs in short.

repos/system_upgrade/common/libraries/utils.py
```diff
@@ -1,45 +1,41 @@
...
 from leapp.libraries.common import mounting
 from leapp.libraries.stdlib import api, CalledProcessError, config, run, STDOUT
 from leapp.utils.deprecation import deprecated
 
 
-def parse_config(cfg=None, strict=True):
+def parse_config(cfg=None, strict=True, no_interpolation=False):
     """
-    Applies a workaround to parse a config file using py3 AND py2
+    Parse a config file or string
```
PeterMocary: Nitpick: I'd  put the information about it being able to consume a file-like object or its contents as a string for the description of the `cfg` argument.
```suggestion
    Parse a config file
```

PeterMocary (commented):
LGTM! Just some minor tweaks to the doc string. You can decide whether to include the nitpicks. :)

matejmatuska (approved):
LGTM!

PR #1526, opened 2026-04-22
Live mode: update 'do-upgrade.sh' to reflect upstream changes

PeterMocary:
### fix(LiveMode): update handling of .leapp_upgrade_failed file
In LiveMode, existence of .leapp_upgrade_failed file was treated
differently than in the custom initramfs solution leapp uses by default.
This patch updates LiveMode so that it treats the file the same way as
the default solution.

Reflects changes from commit https://github.com/oamg/leapp-repository/commit/1f8b8f3.

### fix(LiveMode): update regex used for extraction of OS major version
Updates the regex used to extract OS major version so that it is able to
extract version without a minor version which is useful for CentOS
version. As part of this patch, unused parts of code were removed to
keep the file cleaner. The removed parts were a logging solution that
was not used in LiveMode and happened to be here as a copy of the
original dracut's version of the do-upgrade.sh file.

Reflects changes from commit https://github.com/oamg/leapp-repository/commit/64ec2ec60eac7abd4910c5b2a1a43794d3df11cf.

### fix(LiveMode): ignore failures when calling mount -a
This patch makes mount treat nofail option same as systemd for LiveMode.

Reflect changes from commit https://github.com/oamg/leapp-repository/commit/966e7854ec63a977656743b53b1b2811f255a006.

### fix(LiveMode): improve do-upgrade.sh output when .noreboot file exists
The output of do-upgrade.sh suggested that a reboot is going to happen
even though it wasn't as .noreboot file stopped it. This patch makes
sure the output is not misleading users.

Update: Squashed the commits, leaving the messages here for more info.

---

Jira: RHEL-104384, RHEL-56040

---
I'm thinking about a good way to merge these files so that we have the code in one place. 
Update: I don't think there is a good way to merge them, we would either create a third file with common code or create one file with increased amount of if/else statements executing actions that depend on the mode. Lets leave this for later when live mode is finished -- for now we will have to remember to update both.

### DIFF
--- a/repos/system_upgrade/common/actors/livemode/modify_userspace_for_livemode/files/do-upgrade.sh
+++ b/repos/system_upgrade/common/actors/livemode/modify_userspace_for_livemode/files/do-upgrade.sh
@@ -7,13 +7,13 @@ warn() {
 
 get_rhel_major_release() {
     local os_version
-    os_version=$(grep -o '^VERSION="[0-9][0-9]*\.' /etc/os-release | grep -o '[0-9]*')
+    os_version=$(grep -o '^VERSION="[0-9][0-9]*' /etc/os-release | grep -o '[0-9]*')
     [ -z "$os_version" ] && {
         # This should not happen as /etc/initrd-release is supposed to have API
         # stability, but check is better than broken system.
         warn "Cannot determine the major RHEL version."
         warn "The upgrade environment cannot be setup reliably."
-        echo "Content of the /etc/initrd-release:"
+        echo "Content of the /etc/os-release:"
         cat /etc/os-release
         exit 1
     }
@@ -26,6 +26,7 @@ export RHEL_OS_MAJOR_RELEASE
 export LEAPPBIN=/usr/bin/leapp
 export LEAPPHOME=/root/tmp_leapp_py3
 export LEAPP3_BIN=$LEAPPHOME/leapp3
+export LEAPP_FAILED_FLAG_FILE="/root/tmp_leapp_py3/.leapp_upgrade_failed"
 
 # this was initially a dracut script, hence $NEWROOT.
 # the rootfs is mounted on /run/upgrade when booted with dmsquash-live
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/common/actors/livemode/modify_userspace_for_livemode/files/do-upgrade.sh
```diff
@@ -7,13 +7,13 @@ warn() {
...
     local os_version
-    os_version=$(grep -o '^VERSION="[0-9][0-9]*\.' /etc/os-release | grep -o '[0-9]*')
+    os_version=$(grep -o '^VERSION="[0-9][0-9]*' /etc/os-release | grep -o '[0-9]*')
     [ -z "$os_version" ] && {
         # This should not happen as /etc/initrd-release is supposed to have API
         # stability, but check is better than broken system.
         warn "Cannot determine the major RHEL version."
         warn "The upgrade environment cannot be setup reliably."
-        echo "Content of the /etc/initrd-release:"
+        echo "Content of the /etc/os-release:"
```
matejmatuska, line 16: I think /etc/os-release might be a symlink to /etc/initrd-release. In which case I am not sure I would change this.

repos/system_upgrade/common/actors/livemode/modify_userspace_for_livemode/files/do-upgrade.sh
```diff
@@ -265,14 +113,14 @@ do_upgrade() {
...
+        echo >&2 "Warning: Leapp upgrade failed and there is an issue blocking the upgrade."
+        echo >&2 "Please file a support case with /var/log/leapp/leapp-upgrade.log attached."
+
         "$NEWROOT/bin/touch" "$NEWROOT$LEAPP_FAILED_FLAG_FILE"
     fi
 
-    # Dump debug data in case something went wrong
-    ##if want_inband_dump "$rv"; then
-    ##    collect_and_dump_debug_data
-    ##fi
```
matejmatuska, line 274: Context from an discussion off Github: The ibdmp code is dropped because this code, that is the only way to enable it, is commented out because it depends on `getarg`.
We are going with removal because nobody noticed it's isn't working yet and even in the "old" do-upgrade script it's most probably not used at all.

matejmatuska (commented):
It looks alright, just one comment. I am fairly confident it will work correctly, but I will do simple test before merging.

repos/system_upgrade/common/actors/livemode/modify_userspace_for_livemode/files/do-upgrade.sh
matejmatuska: this is not true when the /sysroot/.noreboot file is present

matejmatuska (changes requested):

repos/system_upgrade/common/actors/livemode/modify_userspace_for_livemode/files/do-upgrade.sh
PeterMocary: Updated.

repos/system_upgrade/common/actors/livemode/modify_userspace_for_livemode/files/do-upgrade.sh
PeterMocary: I've checked and it does not seem to be.
```
[root@leapp-20260504092408 ~]# ls -lah /etc/os-release
lrwxrwxrwx. 1 root root 21 Apr 20 12:59 /etc/os-release -> ../usr/lib/os-release
[root@leapp-20260504092408 ~]# ls -lah ../usr/lib/os-release
-rw-r--r--. 1 root root 627 Apr 20 12:59 ../usr/lib/os-release
```

matejmatuska: And does /etc/initrd-release even exist?

PeterMocary: No it does not exist.

```
[root@leapp-20260504134553 ~]# ls -lah /etc/initrd-release
ls: cannot access '/etc/initrd-release': No such file or directory
```

matejmatuska: Great, then it's good you changed it.

matejmatuska (approved):
LGTM, tested also with a corrupted disk mounted and it's ignored as expected.

PeterMocary:
Squashed the commits.

PR #1527, opened 2026-04-23
el8toel9: networkdeprecations: warn about ifcfg devices not controlled by NM

ihuguet:
Network devices currently configured via ifcfg have to be configured by NetworkManager in EL9. If NetworkManager is disabled, connectivity will be lost. Generate a report with HIGH severity in that case.

If devices are configured with NM_CONTROLLED=no, then ignore it, as they are being explicitly excluded from NetworkManager control.

Assisted-by: Claude 4.6 Opus

Jira: https://redhat.atlassian.net/browse/RHEL-133966

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/networkdeprecations/libraries/networkdeprecations.py
+++ b/repos/system_upgrade/el8toel9/actors/networkdeprecations/libraries/networkdeprecations.py
@@ -1,12 +1,13 @@
 from leapp import reporting
 from leapp.libraries.stdlib import api
-from leapp.models import IfCfg, NetworkManagerConnection
+from leapp.models import IfCfg, NetworkManagerConnection, SystemdServicesInfoSource
 
 FMT_LIST_SEPARATOR = '\n    - '
 
 
 def process():
     wep_files = []
+    nm_controlled_files = []
 
     # Scan NetworkManager native keyfile connections
     for nmconn in api.consume(NetworkManagerConnection):
@@ -26,6 +27,8 @@ def process():
         if ifcfg.secrets is not None:
             props = props + ifcfg.secrets
 
+        nm_controlled = True
+
         for prop in props:
             name = prop.name
             value = prop.value
@@ -39,6 +42,14 @@ def process():
             if name in ('KEY_PASSPHRASE1', 'KEY1', 'WEP_KEY_FLAGS'):
                 wep_files.append(ifcfg.filename)
                 continue
+
+            # NM_CONTROLLED
+            if name == 'NM_CONTROLLED' and value.lower() in ("no", "false", "f", "n", "0"):
+                nm_controlled = False
+                continue
+
+        if nm_controlled:
+            nm_controlled_files.append(ifcfg.filename)
 
     if wep_files:
         title = 'Wireless networks using unsupported WEP encryption detected'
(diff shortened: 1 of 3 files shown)


repos/system_upgrade/el8toel9/actors/networkdeprecations/libraries/networkdeprecations.py
```diff
@@ -65,3 +76,42 @@ def process():
...
+            'Either enable the NetworkManager service before upgrading, or add'
+            ' NM_CONTROLLED=no to the ifcfg files that should not be managed'
+            ' by NetworkManager.'
+        )
+        reporting.create_report([
+            reporting.Title(title),
+            reporting.Summary(summary),
+            reporting.Remediation(hint=remediation),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Groups([reporting.Groups.NETWORK, reporting.Groups.SERVICES]),
```
pirat89, line 102: @ihuguet  it seems to me that this report is supposed to stop the upgrade to be sure the networking is working correctly after the upgrade. correct? in such a case you need this:


```suggestion
            reporting.Groups([reporting.Groups.NETWORK, reporting.Groups.SERVICES]),
            reporting.Groups([reporting.Groups.INHIBITOR]),
```

ihuguet: Changed

pirat89: @ihuguet possibly the change has been dropped by mistake during rebase? I can see you force-pushed changes but the list of changes is actually empty.

ihuguet: Sorry, my fault, I didn't amend correctly the commit. Now it's OK.

pirat89:
@ihuguet  `make lint_fix` should resolve the linter issue. The code LGTM otherwise, but I haven't tried it yet. Let's fix it and we can then run the CI then.

ihuguet:
> @ihuguet `make lint_fix` should resolve the linter issue. The code LGTM otherwise, but I haven't tried it yet. Let's fix it and we can then run the CI then.

`make lint_fix` didn't change anything, but I did the change manually

pirat89 (approved):
lgtm. I want to do some manual testing yet (unless someone is faster) and we can merge it.

repos/system_upgrade/el8toel9/actors/networkdeprecations/libraries/networkdeprecations.py
```diff
@@ -65,3 +76,43 @@ def process():
...
+            ' upgrade, legacy network scripts will no longer be available and'
+            ' NetworkManager will be the only way to manage networking. These'
+            ' connections will not be active unless NetworkManager is enabled.'
+            ' Files with the problematic configuration:{}'
+        ).format(
+            ''.join(['{}{}'.format(FMT_LIST_SEPARATOR, f) for f in nm_controlled_files])
+        )
+        remediation = (
+            'Either enable the NetworkManager service before upgrading, or add'
+            ' NM_CONTROLLED=no to the ifcfg files that should not be managed'
```
tyll, line 94: This seems bad advice to me. What is going to parse these files on RHEL 9?
To me it seems that the right path is to require enabling NM (or to just enable it as part of the upgrade unconditionally). Does something ensure that NM will actually be installed on RHEL 9?

pirat89: if networkmanager is not present on rhe original system, there is no guarantee it will be present on the upgraded system, as NM is removable on RHEL 9. Maybe it's because of cloud-init? I am not SME on networking, but to me it seems that cloud-init and NM are in conflict on RHEL (especially in RHEL 9+).

PR #1528, opened 2026-04-27
leapp_resume.service: use multi-user.target instead of default.target

PeterMocary:
When selinux autorelabel is triggered, the `selinux-autorelabel-generator` links `default.target` to `selinux-autorelabel.target`. During boot, the relabel is performed and then the system is rebooted, resulting in an intermediate reboot before the system fully starts.

Since `leapp_resume.service` was linked into `default.target.wants/` and ordered `After=default.target`, it was started during this intermediate autorelabel boot. When the relabel finished and the reboot was triggered, `leapp_resume.service` was terminated, preventing post-upgrade tasks from completing. The service was then rerun on the next boot.

This patch makes `leapp_resume.service` rely on `multi-user.target`, instead of the `default.target`, which is not reached during the autorelabel boot, ensuring `leapp_resume.service` only runs on the actual first boot into the upgraded OS.

Jira: RHEL-60060

---

### Steps to reproduce:
1. Make sure the [journal is persistent](https://access.redhat.com/solutions/696893)
2. Enable selinux if not enabled already (verify with `sestatus`)
3. Enable the relabeling `touch /.autorelabel`

Without the patch, the `leapp_resume.service` is started in both last boots (this is where the persistent journal is useful).
With the patch, the `leapp_resume.service` is started only in the last reboot and not in the relabeling boot.

### DIFF
--- a/repos/system_upgrade/common/actors/createresumeservice/actor.py
+++ b/repos/system_upgrade/common/actors/createresumeservice/actor.py
@@ -22,18 +22,21 @@ class CreateSystemdResumeService(Actor):
     tags = (FinalizationPhaseTag, IPUWorkflowTag)
 
     def process(self):
-        service_name = 'leapp_resume.service'
         systemd_dir = '/etc/systemd/system'
+        service_name = 'leapp_resume.service'
+        target_name = 'multi-user.target'
 
         service_templ_fpath = self.get_file_path(service_name)
         shutil.copyfile(service_templ_fpath, os.path.join(systemd_dir, service_name))
 
-        service_path = '/etc/systemd/system/{}'.format(service_name)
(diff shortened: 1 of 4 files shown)


MichalHe (approved):
Code-wise LGTM. I have executed the upgrade, using `rd.upgrade.break=leapp-finish`. The following symlink is created:
```
# ls -trahl /sysroot/etc/systemd/system/multi-user.target.wants/leapp_resume.service
lrwxrwxrwx 1 root root 40 Apr 28 10:01 /sysroot/etc/systemd/system/multi-user.target.wants/leapp_resume.service -> /etc/systemd/system/leapp_resume.service
```
After rebooting into RHEL 10, the service is correctly removed. Inspecting the past 3 boots, the leapp_resume service is not being executed during relabeling.

PR #1530, opened 2026-05-04
Remove enforcing=1 from boot entries during IPU

prjagtap-rh:
Detect enforcing=1 on the running kernel (KernelCmdline) or in any bootloader
entry. Bootloader detection is collected by systemfacts via `grubby --info ALL`
and exposed as SELinuxFacts.enforcing_via_any_cmdline.

When enforcing=1 is present, checkselinux (Checks phase) emits
UpgradeKernelCmdlineArgTasks to remove enforcing=1 from the interim upgrade
boot entry, and TargetKernelCmdlineArgTasks to remove it from the target kernel
entry and update the default kernel command line during finalization.

An INFO report with remediation guidance to restore enforcing=1 after upgrade is also
produced. Original boot entries for existing kernels are not modified.

JIRA-ref: RHEL-162192

### DIFF
--- a/repos/system_upgrade/common/actors/checkselinux/libraries/checkselinux.py
+++ b/repos/system_upgrade/common/actors/checkselinux/libraries/checkselinux.py
@@ -1,10 +1,84 @@
 from leapp import reporting
+from leapp.libraries.common.config import architecture
 from leapp.libraries.common.config.version import get_target_major_version
 from leapp.libraries.common.distro import DISTRO_REPORT_NAMES
 from leapp.libraries.stdlib import api
-from leapp.models import KernelCmdlineArg, SELinuxFacts, SelinuxPermissiveDecision, SelinuxRelabelDecision
+from leapp.models import (
+    KernelCmdline,
+    KernelCmdlineArg,
+    SELinuxFacts,
+    SelinuxPermissiveDecision,
+    SelinuxRelabelDecision,
+    TargetKernelCmdlineArgTasks,
+    UpgradeKernelCmdlineArgTasks
+)
 
 DOC_URL = 'https://red.ht/rhel9-disabling-selinux'
+
+_ENFORCING_ONE_ARG = KernelCmdlineArg(key='enforcing', value='1')
+
+
+def _proc_cmdline_has_enforcing_one(kernel_cmdline):
+    """
+    Checks if the kernel command line contains enforcing=1.
+    """
+    if not kernel_cmdline:
+        return False
+    for param in kernel_cmdline.parameters:
+        if param.key == 'enforcing' and param.value == '1':
+            return True
+    return False
+
+
+def _request_removal_of_enforcing_one():
+    """
+    Schedule stripping of enforcing=1 from the upgrade and target boot entries.
+
+    The upgrade workflow runs with SELinux permissive, but if enforcing=1 is present on
+    the running kernel or in any bootloader entry it can override that. We produce:
+
+    * ``UpgradeKernelCmdlineArgTasks`` so the interim upgrade boot entry drops enforcing=1
+      (consumed by ``addupgradebootentry``).
+    * ``TargetKernelCmdlineArgTasks`` so the target kernel entry and the default kernel
+      command line configuration are updated during finalization (consumed by
+      ``kernelcmdlineconfig``), ensuring that future kernels also omit enforcing=1.
+
+    Original boot entries for existing (non-target) kernels are left untouched.
+
+    ``SELinuxFacts.enforcing_via_any_cmdline`` is populated earlier by the
+    ``systemfacts`` actor.
+    """
+    arch = api.current_actor().configuration.architecture
+    arch_msg = ' Then make sure to run zipl to update the boot menu.' if arch == architecture.ARCH_S390X else ''
+    doc_url = 'https://red.ht/rhel-{}-configure-kernel-cmdline'.format(get_target_major_version())
+    hint = (
+        'After the upgrade, add the `enforcing=1` kernel command line argument back if it is wanted.'
+        ' Run "grubby --update-kernel=ALL --args=enforcing=1" to enable enforcing on all boot entries'
+        f' and set it as default.{arch_msg}'
+        ' Follow the documentation in the attached link for more details.'
+    )
+
+    api.produce(UpgradeKernelCmdlineArgTasks(to_remove=[_ENFORCING_ONE_ARG]))
+    api.produce(TargetKernelCmdlineArgTasks(to_remove=[_ENFORCING_ONE_ARG]))
+    reporting.create_report([
+        reporting.Title('Kernel boot parameter enforcing=1 will be removed'),
+        reporting.Summary(
+            'The kernel parameter enforcing=1 forces SELinux enforcing mode at boot and overrides '
+            'the permissive configuration used during the upgrade. Leapp will remove enforcing=1 from '
+            'the upgrade and target kernel boot entries and update the default kernel command line '
+            'configuration so it does not persist after the upgrade. Existing boot entries for other '
+            'kernels will not be modified.'
+        ),
+        reporting.Remediation(hint=hint),
+        reporting.Severity(reporting.Severity.INFO),
+        reporting.Groups([
+            reporting.Groups.SELINUX,
+            reporting.Groups.BOOT,
+            reporting.Groups.KERNEL,
+            reporting.Groups.POST,
+        ]),
+        reporting.ExternalLink(url=doc_url, title='Configuring kernel command line parameters'),
+    ])
 
 
 def process():
@@ -48,6 +122,10 @@ def process():
             reporting.Groups([reporting.Groups.SELINUX, reporting.Groups.SECURITY])
         ])
         return
+
+    kernel_cmdline = next(api.consume(KernelCmdline), None)
+    if _proc_cmdline_has_enforcing_one(kernel_cmdline) or facts.enforcing_via_any_cmdline:
+        _request_removal_of_enforcing_one()
 
     if conf_status in ('enforcing', 'permissive'):
         api.produce(SelinuxRelabelDecision(set_relabel=True))
--- a/repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
+++ b/repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
@@ -230,6 +230,24 @@ def get_repositories_status():
             })
 
 
+def _bootloader_entries_contain_enforcing_one():
+    """
+    True if grubby reports enforcing=1 in any boot loader entry's kernel arguments.
+    """
+    try:
+        out = run(['/usr/sbin/grubby', '--info', 'ALL'], split=False)['stdout']
+    except (OSError, CalledProcessError):
+        api.current_logger().debug(
+            'grubby --info ALL failed; assuming no bootloader enforcing=1', exc_info=True
+        )
+        return False
+
+    for line in out.splitlines():
+        if line.startswith('args=') and 'enforcing=1' in line[len('args='):].strip('"').split():
+            return True
+    return False
+
+
 def get_selinux_status():
     """ Get SELinux status information """
     # will be None if something went wrong or contain SELinuxFacts otherwise
@@ -258,6 +276,8 @@ def get_selinux_status():
         outdata['runtime_mode'] = 'permissive'
         outdata['static_mode'] = 'disabled'
         outdata['policy'] = 'targeted'
+
+    outdata['enforcing_via_any_cmdline'] = _bootloader_entries_contain_enforcing_one()
 
     res = SELinuxFacts(**outdata)
     return res
--- a/repos/system_upgrade/common/actors/systemfacts/tests/test_systemfacts_selinux.py
+++ b/repos/system_upgrade/common/actors/systemfacts/tests/test_systemfacts_selinux.py
@@ -2,6 +2,7 @@ import warnings
 
 import pytest
 
+from leapp.libraries.actor import systemfacts
 from leapp.libraries.actor.systemfacts import get_selinux_status
 from leapp.models import SELinuxFacts
 
@@ -20,6 +21,23 @@ reason_to_skip_msg = "Selinux is not available"
 # FIXME: create valid tests...
 
 
+def _run_result(stdout):
+    return {'stdout': stdout, 'stderr': '', 'signal': None, 'exit_code': 0, 'pid': 1}
+
+
+@pytest.fixture(autouse=True)
+def stub_grubby_info_all_without_enforcing(monkeypatch):
+    """
+    Avoid invoking real grubby from get_selinux_status(); default: no enforcing=1 in boot entries.
+    """
+
+    def mocked_run(cmd, split=False, **dummy_kwargs):
+        assert cmd == ['/usr/sbin/grubby', '--info', 'ALL'], f"Unexpected grubby cmd: {cmd}"
+        return _run_result('index=0\nkernel="/boot/vmlinuz-1"\nargs="ro rhgb quiet"\nroot="UUID=xxx"\n')
+
+    monkeypatch.setattr(systemfacts, 'run', mocked_run)
+
+
 @pytest.mark.skipif(no_selinux, reason=reason_to_skip_msg)
 def test_selinux_enabled_enforcing(monkeypatch):
     """
@@ -34,7 +52,8 @@ def test_selinux_enabled_enforcing(monkeypatch):
                      'mls_enabled': True,
                      'enabled': True,
                      'runtime_mode': 'enforcing',
-                     'static_mode': 'enforcing'}
+                     'static_mode': 'enforcing',
+                     'enforcing_via_any_cmdline': False}
     assert SELinuxFacts(**expected_data) == get_selinux_status()
 
 
@@ -52,7 +71,8 @@ def test_selinux_enabled_permissive(monkeypatch):
                      'mls_enabled': True,
                      'enabled': True,
                      'runtime_mode': 'permissive',
-                     'static_mode': 'permissive'}
+                     'static_mode': 'permissive',
+                     'enforcing_via_any_cmdline': False}
     assert SELinuxFacts(**expected_data) == get_selinux_status()
 
 
@@ -70,7 +90,8 @@ def test_selinux_disabled(monkeypatch):
                      'mls_enabled': False,
                      'enabled': False,
                      'runtime_mode': 'permissive',
-                     'static_mode': 'permissive'}
+                     'static_mode': 'permissive',
+                     'enforcing_via_any_cmdline': False}
     assert SELinuxFacts(**expected_data) == get_selinux_status()
 
 
@@ -93,6 +114,38 @@ def test_selinux_disabled_no_config_file(monkeypatch):
                      'mls_enabled': False,
                      'enabled': False,
                      'runtime_mode': 'permissive',
-                     'static_mode': 'disabled'}
+                     'static_mode': 'disabled',
+                     'enforcing_via_any_cmdline': False}
 
     assert SELinuxFacts(**expected_data) == get_selinux_status()
+
+
+@pytest.mark.skipif(no_selinux, reason=reason_to_skip_msg)
+@pytest.mark.parametrize('grubby_stdout,expected', [
+    ('index=0\nargs="ro enforcing=1"\n', True),
+    ('index=0\nargs="ro enforcing=1 rhgb quiet"\n', True),
+    ('index=0\nargs="enforcing=1"\n', True),
+    ('index=0\nargs="ro enforcing=1"\nindex=1\nargs="ro rhgb quiet"\n', True),
+    ('index=0\nargs="ro rhgb quiet"\nindex=1\nargs="ro enforcing=1"\n', True),
+    ('index=0\nargs="ro enforcing=0"\n', False),
+    ('index=0\nargs="ro rhgb quiet"\n', False),
+    ('index=0\nargs="ro fooenforcing=1"\n', False),
+    ('index=0\nargs="ro enforcing=1bar"\n', False),
+    ('index=0\nargs="ro"\n', False),
+    ('index=0\nargs=""\n', False),
+])
+def test_bootloader_enforcing_one_detected(monkeypatch, grubby_stdout, expected):
+    monkeypatch.setattr(selinux, 'is_selinux_mls_enabled', lambda: 1)
+    monkeypatch.setattr(selinux, 'security_getenforce', lambda: 1)
+    monkeypatch.setattr(selinux, 'selinux_getenforcemode', lambda: [0, 1])
+    monkeypatch.setattr(selinux, 'is_selinux_enabled', lambda: 1)
+    monkeypatch.setattr(selinux, 'selinux_getpolicytype', lambda: [0, 'targeted'])
+
+    def mocked_run(cmd, split=False, **dummy_kwargs):
+        assert cmd == ['/usr/sbin/grubby', '--info', 'ALL'], f"Unexpected grubby cmd: {cmd}"
+        return _run_result(grubby_stdout)
+
+    monkeypatch.setattr(systemfacts, 'run', mocked_run)
+
+    fact = get_selinux_status()
+    assert fact.enforcing_via_any_cmdline is expected
--- a/repos/system_upgrade/common/models/selinuxfacts.py
+++ b/repos/system_upgrade/common/models/selinuxfacts.py
@@ -10,3 +10,7 @@ class SELinuxFacts(Model):
     enabled = fields.Boolean()
     policy = fields.StringEnum(['targeted', 'minimum', 'mls'])
     mls_enabled = fields.Boolean()
+    enforcing_via_any_cmdline = fields.Boolean(default=False)
+    """
+    True when any bootloader entries have set enforcing=1.
+    """
--- a/repos/system_upgrade/common/actors/checkselinux/actor.py
+++ b/repos/system_upgrade/common/actors/checkselinux/actor.py
@@ -1,6 +1,15 @@
 from leapp.actors import Actor
 from leapp.libraries.actor import checkselinux
-from leapp.models import KernelCmdlineArg, Report, SELinuxFacts, SelinuxPermissiveDecision, SelinuxRelabelDecision
+from leapp.models import (
+    KernelCmdline,
+    KernelCmdlineArg,
+    Report,
+    SELinuxFacts,
+    SelinuxPermissiveDecision,
+    SelinuxRelabelDecision,
+    TargetKernelCmdlineArgTasks,
+    UpgradeKernelCmdlineArgTasks
+)
 from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
 
 
@@ -13,12 +22,14 @@ class CheckSelinux(Actor):
     """
 
     name = 'check_se_linux'
-    consumes = (SELinuxFacts,)
+    consumes = (SELinuxFacts, KernelCmdline)
     produces = (
         KernelCmdlineArg,
         Report,
         SelinuxPermissiveDecision,
         SelinuxRelabelDecision,
+        TargetKernelCmdlineArgTasks,
+        UpgradeKernelCmdlineArgTasks,
     )
     tags = (ChecksPhaseTag, IPUWorkflowTag)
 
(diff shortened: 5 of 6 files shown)


repos/system_upgrade/common/actors/checkselinux/libraries/checkselinux.py
```diff
@@ -1,17 +1,101 @@
...
+    matches = re.match(r'^([\'"]?)(.*)\1$', data)
+    return matches.group(2)
+
+
+def _bootloader_entries_contain_enforcing_one():
+    """
+    Inspect every bootloader entry (grubby/BLS), not only the running kernel.
+    """
+    try:
+        out = run(['/usr/sbin/grubby', '--info', 'ALL'], split=False)['stdout']
```
pirat89: @prjagtap-rh  actors in checksphase must not interact with the system in the project. Not sure whether checking all entries make sense - I am doing now just a quick overview, not full review. Anyway, in case you are interested if any boot entry has set "enforcing=1", I would update the selinuxfacts model to have such an info.

repos/system_upgrade/common/actors/checkselinux/libraries/checkselinux.py
```diff
@@ -1,17 +1,101 @@
...
+    The upgrade workflow and dracut hook force permissive SELinux unless the kernel was
+    booted with enforcing=1. If that parameter remains in any grub entry after upgrade,
+    a later boot can force enforcing mode and break the upgraded system. The interim upgrade
+    boot entry is configured with enforcing=0 and strips enforcing=1 when copied from the
+    default entry; this scheduling ensures enforcing=1 is not kept on the upgraded system
+    (see Checks phase here, application on ALL kernels in kernelcmdlineconfig).
+    """
+    in_proc = _proc_cmdline_has_enforcing_one(kernel_cmdline)
+    in_bootloader = False if in_proc else _bootloader_entries_contain_enforcing_one()
+    if not in_proc and not in_bootloader:
```
pirat89: this is weird. Why to make 2 checks at all? 

```suggestion
    in_proc = _proc_cmdline_has_enforcing_one(kernel_cmdline)
    in_bootloader = _bootloader_entries_contain_enforcing_one()
    if not (in_proc or in_bootloader):
```

If it was to skip the check of other bootloader entries, it's not important as in such a case the grubby call would have to be called before the actor execution as the interaction with the system must not be performed in this actor.

repos/system_upgrade/common/actors/checkselinux/libraries/checkselinux.py
```diff
@@ -1,17 +1,101 @@
...
+        return
+
+    api.produce(TargetKernelCmdlineArgTasks(to_remove=[_ENFORCING_ONE_ARG]))
+    reporting.create_report([
+        reporting.Title('Kernel boot parameter enforcing=1 will be removed'),
+        reporting.Summary(
+            'The kernel parameter enforcing=1 forces SELinux enforcing mode at boot and overrides '
+            'the permissive configuration used during the upgrade. Leapp will remove enforcing=1 from '
+            'all boot loader entries during the upgrade finalization so it does not persist after the upgrade.'
```
pirat89: at this point, it will mean also that the default kernel cmdline arguments will be updated. So the report should contain also remediation instruction, that if they want set `enforcing=1` kernel arg as default after the upgrade, they need to update the /etc/kernel/cmdline file after the upgrade as well (good candidate for remediation hint). be aware that this file is not updated always. so the report will need to be updated based on the target major version & architecture (IIRC) - note that as the remediation is for the upgraded system, the report should have in such a case the post group set.

pirat89 (changes requested):
the PR needs some changes. I did a quick overview of the solution (not full review)

repos/system_upgrade/common/models/selinuxfacts.py
```diff
@@ -10,3 +10,5 @@ class SELinuxFacts(Model):
     enabled = fields.Boolean()
     policy = fields.StringEnum(['targeted', 'minimum', 'mls'])
     mls_enabled = fields.Boolean()
+    # True when grubby reports any boot entry whose kernel args contain enforcing=1 (see systemfacts collection).
+    bootloader_entries_have_enforcing_one = fields.Boolean(default=False)
```
pirat89: @prjagtap-rh  write please correct docstring instead:


```suggestion
    bootloader_entries_have_enforcing_one = fields.Boolean(default=False)
    """
    True when any bootloader entries have set enforcing=1.
    """
```

repos/system_upgrade/common/actors/checkselinux/libraries/checkselinux.py
```diff
@@ -1,17 +1,105 @@
...
+        return (
+            'After the upgrade on IBM Power (ppc64le), add enforcing=1 back to the kernel command line if you need it '
+            'for all boot entries (for example `grubby --update-kernel=ALL --args=enforcing=1`, or update '
+            '`/etc/default/grub` and regenerate the config where applicable).'
+        )
+    return (
+        'After the upgrade, add enforcing=1 back to the kernel command line if you need it on {arch} '
+        '(for example `grubby --update-kernel=ALL --args=enforcing=1` for all entries). If you manage the kernel '
+        'command line via `/etc/default/grub`, update it and regenerate the bootloader config.'
+        .format(arch=arch, **DISTRO_REPORT_NAMES)
```
PeterMocary: Probably a leftover - there is no distro name used, so this is redundant
```suggestion
        .format(arch=arch)
```

PeterMocary: I'd form the message first with the common parts and just then add the additional info per architecture. The end result will diverge less between the architectures than it is diverging now.

Something like this (there might be issues):
```suggestion
    arch = api.current_actor().configuration.architecture
    message =  (
        'After the upgrade on {arch}, add enforcing=1 back to the kernel command line if you need it.'
        ' Run "grubby --update-kernel=ALL --args=enforcing=1" to enable enforcing on all boot entries.'
    )
    if arch == architecture.ARCH_S390X:
        message += ' Make sure to run zipl to update the boot menu.'
        return message.format(arch='IBM Z (s390x)')
    message += (
        ' Alternatively, you can adjust "/etc/default/grub" and regenerate the bootloader config using'
        ' the "grub2-mkconfig -o /boot/grub2/grub.cfg" command.'
    )
    return message.format(arch='IBM Power (ppc64le)' if arch == architecture.ARCH_PPC64LE else arch)
```

PeterMocary: Maybe adding external link ([Changing kernel command-line parameters for all boot entries](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/managing_monitoring_and_updating_the_kernel/configuring-kernel-command-line-parameters_managing-monitoring-and-updating-the-kernel#changing-kernel-command-line-parameters-for-all-boot-entries_configuring-kernel-command-line-parameters)) to the report to accompany the hint would be beneficial, but might be confusing in this case...

repos/system_upgrade/common/actors/addupgradebootentry/libraries/addupgradebootentry.py
```diff
@@ -45,9 +45,13 @@ def collect_upgrade_kernel_args(livemode_enabled):
 
 
 def collect_undesired_args(livemode_enabled):
-    args = {}
+    """
+    Always remove ``enforcing=1`` so it is not carried onto the interim upgrade entry
+    Live mode additionally drops args that conflict with the squashfs-based boot path.
+    """
```
PeterMocary: About doc strings: I'd put in a short summary at the start e.g. "Forms a set of arguments that should be removed from the cmdline." and then the additional details about what arguments are removed always or conditionally and for what reason .I'd add the sphinx style documentation of arguments and return value while we are at it.

Here are PEP conventions for doc strings as a reference (https://peps.python.org/pep-0257/).

repos/system_upgrade/common/actors/addupgradebootentry/libraries/addupgradebootentry.py
```diff
@@ -45,9 +45,13 @@ def collect_upgrade_kernel_args(livemode_enabled):
...
+    args = {'enforcing': '1'}
```
pirat89: @prjagtap-rh  note that docstring format is:
```
title

body
```

pirat89: instead of this hardcoded solution, I prefer to implement requested removal based on `UpgradeKernelCmdlineArgTasks` instead of hardcoding it directly in the actor. Note the logic has been implemented in the meanwhile in the upstream, so rebasing this branch it can be just simply used. This actor then needs no changes in this PR, unless a bug is discovered.

repos/system_upgrade/common/actors/kernelcmdlineconfig/libraries/kernelcmdlineconfig.py
```diff
@@ -88,6 +88,24 @@ def retrieve_arguments_to_modify():
...
+        return
+    remove_str = format_kernelarg_msgs_for_grubby_cmd(kernelargs_msgs_to_remove)
+    grubby_cmd = ['grubby', '--update-kernel=ALL', '--remove-args', remove_str]
+    if configs_to_modify_explicitly:
+        for config_to_modify in configs_to_modify_explicitly:
+            run_grubby_cmd(grubby_cmd + ['-c', config_to_modify])
+    else:
+        run_grubby_cmd(grubby_cmd)
+
+
```
pirat89: this is wrong. we do not want to touch original boot entries. The only boot entries expected to be modified are those for the target kernel. Boot entries for the original system need to stay as they are. On top of that, defaults need to be updated as well (which is also handled by the actor already). At this moment I do not see any reason to modify this actor in this PR.

repos/system_upgrade/common/actors/systemfacts/tests/test_systemfacts_selinux.py
```diff
@@ -20,6 +21,24 @@
...
+    return {'stdout': stdout, 'stderr': '', 'signal': None, 'exit_code': 0, 'pid': 1}
+
+
+@pytest.fixture(autouse=True)
+def stub_grubby_info_all_without_enforcing(monkeypatch):
+    """
+    Avoid invoking real grubby from get_selinux_status(); default: no enforcing=1 in boot entries.
+    """
+
+    def fake_run(cmd, split=False, **_kwargs):  # pylint: disable=unused-argument
```
pirat89: note the use of dummy kw satisfies linters 
```suggestion
    def fake_run(cmd, split=False, **dummy_kwargs):
```

repos/system_upgrade/common/actors/systemfacts/tests/test_systemfacts_selinux.py
```diff
@@ -20,6 +21,24 @@
...
+        if len(cmd) >= 3 and cmd[0] == '/usr/sbin/grubby' and cmd[1] == '--info' and cmd[2] == 'ALL':
+            return _run_result('index=0\nkernel="/boot/vmlinuz-1"\nargs="ro rhgb quiet"\nroot="UUID=xxx"\n')
+        raise AssertionError('Unexpected run() in systemfacts SELinux tests: {!r}'.format(cmd))
```
pirat89: why so complicated?

```suggestion
        assert cmd == ['/usr/sbin/grubby', '--info', 'ALL'], f"Unexpected grubby cmd: {cmd}"
        return _run_result('index=0\nkernel="/boot/vmlinuz-1"\nargs="ro rhgb quiet"\nroot="UUID=xxx"\n')
```

Similar fake_run below. btw in this project we prefer `mocked` wording instead of `fake`, to be consistent.

repos/system_upgrade/common/actors/systemfacts/tests/test_systemfacts_selinux.py
```diff
@@ -93,6 +115,26 @@ def test_selinux_disabled_no_config_file(monkeypatch):
...
+    monkeypatch.setattr(selinux, 'is_selinux_mls_enabled', lambda: 1)
+    monkeypatch.setattr(selinux, 'security_getenforce', lambda: 1)
+    monkeypatch.setattr(selinux, 'selinux_getenforcemode', lambda: [0, 1])
+    monkeypatch.setattr(selinux, 'is_selinux_enabled', lambda: 1)
+    monkeypatch.setattr(selinux, 'selinux_getpolicytype', lambda: [0, 'targeted'])
+
+    def fake_run(cmd, split=False, **_kwargs):  # pylint: disable=unused-argument
+        if len(cmd) >= 3 and cmd[0] == '/usr/sbin/grubby' and cmd[1] == '--info' and cmd[2] == 'ALL':
+            return _run_result('index=0\nargs="ro enforcing=1"\n')
+        raise AssertionError(cmd)
```
pirat89: similar as above

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -230,6 +230,36 @@ def get_repositories_status():
             })
 
 
+def _parse_grubby_quoted_value(record):
+    """
+    Parse a grubby --info line: value after '=', strip one matching pair of surrounding quotes.
+    """
+    data = record.split('=', 1)[1]
+    matches = re.match(r'^([\'"]?)(.*)\1$', data)
+    return matches.group(2)
```
pirat89: why so somplicate? why not just something like:

```suggestion
def _parse_grubby_quoted_value(record):
    """
    Parse a grubby --info line: value after '=', strip one matching pair of surrounding quotes.
    """
    data = record.split('=', 1)[1]
    return data[1:-1]  # if nested quotes/apostrophes are expected
```

or if we know that only quotes are expected and nested quotes are not:

```suggestion
def _parse_grubby_quoted_value(record):
    """
    Parse a grubby --info line: value after '=', strip one matching pair of surrounding quotes.
    """
    return record.split('=', 1)[1].strip('"')
```

but in both cases, based on tests and and code in the caller function, I think this function can be just dropped as the parsing does not seem to be needed at all.

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -230,6 +230,36 @@ def get_repositories_status():
...
+            'grubby --info ALL failed; assuming no bootloader enforcing=1', exc_info=True
+        )
+        return False
+
+    for line in out.splitlines():
+        if line.startswith('args='):
+            argstr = _parse_grubby_quoted_value(line)
+            tokens = argstr.split()
+            if 'enforcing=1' in tokens:
+                return True
```
pirat89: based on what is expected, it can be simplified:


```suggestion
        if line.startswith('args=') and 'enforcing=1' in line.split():
                return True
```

As the parameters are not important here, proper parsing is also not needed, and only thing that we are interested about is to check whether `enforcing=1` is set. and for that it should be enough to tokenize the line.

repos/system_upgrade/common/actors/systemfacts/tests/test_systemfacts_selinux.py
```diff
@@ -93,6 +115,26 @@ def test_selinux_disabled_no_config_file(monkeypatch):
...
                      'runtime_mode': 'permissive',
-                     'static_mode': 'disabled'}
+                     'static_mode': 'disabled',
+                     'bootloader_entries_have_enforcing_one': False}
 
     assert SELinuxFacts(**expected_data) == get_selinux_status()
+
+
+@pytest.mark.skipif(no_selinux, reason=reason_to_skip_msg)
+def test_bootloader_enforcing_one_detected(monkeypatch):
```
pirat89: the function checks just one specific `args` line with enforcing=1. but it's not checking situation when multiple `args` lines is present with various setups, not a simple args line with various content, focusing also on strings like `fooenforcing=1`, `enforcing=0`, no enforcing specified at all, etc. The test should be extended - ideally parametrized with `@pytest.mark.parametrize` to check the behaviour for various inputs.

repos/system_upgrade/common/models/selinuxfacts.py
```diff
@@ -10,3 +10,7 @@ class SELinuxFacts(Model):
     enabled = fields.Boolean()
     policy = fields.StringEnum(['targeted', 'minimum', 'mls'])
     mls_enabled = fields.Boolean()
+    bootloader_entries_have_enforcing_one = fields.Boolean(default=False)
```
pirat89: thinking about the name twice, what about to use something like


```suggestion
    enforcing_via_any_cmdline = fields.Boolean(default=False)
```

@oamg/developers any idea about a better name?

pirat89 (changes requested):
Finished review. The solution should be modified to use `UpgradeKernelCmdlineArgTasks` and `TargetKernelCmdlineArgTasks` messages. Unless I miss something, only systemfacts & checkselinux actors should be updated.

repos/system_upgrade/common/models/selinuxfacts.py
pirat89: note that I am not sure my name is better :D just when I see it, I have the feel we should have a better name

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
PeterMocary: I agree, using regex here is not necessary. I'd drop the function and go with the simplified solution in the [other comment](https://github.com/oamg/leapp-repository/pull/1530/changes#r3248071567).

repos/system_upgrade/common/models/selinuxfacts.py
PeterMocary: I agree, using  `enforcing_in_any_cmdline` could be an option too.

repos/system_upgrade/common/actors/checkselinux/libraries/checkselinux.py
```diff
@@ -1,17 +1,105 @@
...
+    )
+
+
+def _request_removal_of_enforcing_one_from_target_cmdline(facts, kernel_cmdline):
+    """
+    Schedule stripping of enforcing=1 from all target boot entries (finalization phase).
+
```
PeterMocary: I'd include mention about the report in the first summary line.

```suggestion
def _request_removal_of_enforcing_one_from_target_cmdline(facts, kernel_cmdline):
    """
    Schedule stripping of enforcing=1 from all target boot entries and create a report about it.

```

repos/system_upgrade/common/actors/checkselinux/libraries/checkselinux.py
```diff
@@ -1,17 +1,105 @@
...
+        return False
+    for param in kernel_cmdline.parameters:
+        if param.key == 'enforcing' and param.value == '1':
+            return True
+    return False
+
+
+def _restore_enforcing_one_remediation_hint():
+    """
+    Instructions apply on the upgraded system (target architecture).
```
PeterMocary: This is not really stating what the function does. It should inform about the fact that it forms remediation hint for a specific report based on the system architecture.

PeterMocary (commented):
I think as well, as suggested by @pirat89, this can be solved simply along the lines of `systemfacts` actor provides information about `enforcing=1` on kernel cmdline to `checkselinux` actor which then uses the information to requests removal of the enforcing cmdline argument from both upgrade and target kernel cmdline using the Tasks messages.

I'll review again and test it after the suggestions are addressed.

repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
prjagtap-rh: @pirat89 
Regarding the suggestion to simplify _bootloader_entries_contain_enforcing_one() to just tokenize the line with line.split():

A plain line.split() doesn't work reliably because grubby wraps the args value in double quotes, 
e.g. args="ro enforcing=1". 

When we split the raw line, the quotes get added

```
'args="ro enforcing=1"'.split() → ['args="ro', 'enforcing=1"']  <-  trailing quote on last token = no match

'args="enforcing=1"'.split() → ['args="enforcing=1"']  <- prefix and quotes both attached = no match

'args="enforcing=1 ro"'.split() → ['args="enforcing=1', 'ro"'] <- prefix attached = no match

```

The only position where it accidentally works is when enforcing=1 is in the middle with args on both sides.

I have removed the separate _parse_grubby_quoted_value() function as suggested. 

Hence, replaced it with below
 ```

        if line.startswith('args=') and 'enforcing=1' in line[len('args='):].strip('"').split():
            return True
 
```

Comment (pirat89) [reply repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py]: @prjagtap-rh  good point!

File: repos/system_upgrade/common/actors/checkselinux/libraries/checkselinux.py [hunk 18]
```diff
@@ -1,17 +1,107 @@
...
+        ' Run "grubby --update-kernel=ALL --args=enforcing=1" to enable enforcing on all boot entries.'
+    )
+    if arch == architecture.ARCH_S390X:
+        message += ' Make sure to run zipl to update the boot menu.'
+        return message.format(arch='IBM Z (s390x)')
+    message += (
+        ' Alternatively, you can adjust "/etc/default/grub" and regenerate the bootloader config using'
+        ' the "grub2-mkconfig -o /boot/grub2/grub.cfg" command.'
+    )
+    return message.format(arch='IBM Power (ppc64le)' if arch == architecture.ARCH_PPC64LE else arch)
```
pirat89: The complexity just for the arch is not necessary. We do not use `IBM Z` and `IBM Power` names - also it would be inconsistent as architecture names for intel & arm are not reflected. From my pov the architecture name is useless in the report as it does not bring any relevant information on top of the message. drop it.

Also the suggestions about /etc/default/grub can lead to a different (no) results as if /etc/kernel/cmdline is set, it has precedence over /etc/default/grub on RHEL 10, and RHEL 9 in case of s390x. Instead of that, it's better to create new shortened URL for the official documentation to let people read it how to do so correctly. I created already shortened URLs for that:
* https://red.ht/rhel-9-configure-kernel-cmdline
* https://red.ht/rhel-10-configure-kernel-cmdline

So you could use later something like this to make it version agnostic:
```
URL = 'https://red.ht/rhel-{}-configure-kernel-cmdline'.format(get_target_major_version())
```

Based on the simplicity, I suggest to drop this nested function and make it part of the caller.

```suggestion
    arch_msg = ' Then make sure to run zipl to update the boot menu.' if arch == architecture.ARCH_S390X else ''
    message = (
        'After the upgrade, add the `enforcing=1` kernel command line argument back if it is wanted.'
        ' Run "grubby --update-kernel=ALL --args=enforcing=1" to enable enforcing on all boot entries'
        f' and set it as default.{arch}'
        ' Follow the attached URL to for more info.'
    )
    return message
```

repos/system_upgrade/common/actors/checkselinux/libraries/checkselinux.py
```diff
@@ -1,17 +1,107 @@
...
+
 
 def process():
     facts = next(api.consume(SELinuxFacts), None)
     if not facts:
         return
 
+    kernel_cmdline = next(api.consume(KernelCmdline), None)
+    _request_removal_of_enforcing_one(facts, kernel_cmdline)
+
```
pirat89: I missed that originally, but I just realized that this happens also when selinux is disabled. However, it should happen only if selinux is enabled. Also the `_request_removal_of_enforcing_one` function says by name that it is requesting the removal of enforcing option, but actually it checks the configuration and then it decides whether to request it or not.

I suggest to move checks out of the nested function - into the `process()` function, and keep inside just the rest.

repos/system_upgrade/common/models/selinuxfacts.py
```diff
@@ -10,3 +10,7 @@ class SELinuxFacts(Model):
     enabled = fields.Boolean()
     policy = fields.StringEnum(['targeted', 'minimum', 'mls'])
     mls_enabled = fields.Boolean()
+    enforcing_via_any_cmdline = fields.Boolean(default=False)
```
pirat89, line 13: I like the idea from @PeterMocary  to call it `enforcing_in_any_cmdline`. But this one can be kept as well. Keeping the decision on you.

pirat89 (changes requested):

repos/system_upgrade/common/models/selinuxfacts.py
prjagtap-rh: I will keep it as `enforcing_via_any_cmdline`

repos/system_upgrade/common/actors/systemfacts/tests/test_systemfacts_selinux.py
```diff
@@ -20,6 +21,23 @@
...
+def stub_grubby_info_all_without_enforcing(monkeypatch):
+    """
+    Avoid invoking real grubby from get_selinux_status(); default: no enforcing=1 in boot entries.
+    """
+
+    def mocked_run(cmd, split=False, **dummy_kwargs):
+        assert cmd == ['/usr/sbin/grubby', '--info', 'ALL'], f"Unexpected grubby cmd: {cmd}"
+        return _run_result('index=0\nkernel="/boot/vmlinuz-1"\nargs="ro rhgb quiet"\nroot="UUID=xxx"\n')
+
+    monkeypatch.setattr(systemfacts, 'run', mocked_run)
```
pirat89, line 38: this test seems to duplicate the test in the bottom

repos/system_upgrade/common/actors/kernelcmdlineconfig/tests/test_kernelcmdlineconfig.py
pirat89, line 1: @prjagtap-rh as there is no change in the code and it is not related to the rest of changes, I suggest one of followings:
* separate it into own commit
* drop it

Both are good for me.

pirat89:
@prjagtap-rh  last changes in tests, but the production code lgtm!

repos/system_upgrade/common/actors/systemfacts/tests/test_systemfacts_selinux.py
pirat89: Ah my mistake. I missed that it's a fixture actually. Resolved

pirat89 (approved):
I am going to test it manually yet and merge it later today

pirat89:
Everything seems to work as it should. Here are some outputs from my last test:
```
                          PRODUCED MESSAGES                           
Stamp: 2026-05-20T10:31:41.014063Z
Actor: system_facts
Phase: FactsCollection
Type: SELinuxFacts
Message_data:
{
    "enabled": true,
    "enforcing_via_any_cmdline": true,
    "mls_enabled": true,
    "policy": "targeted",
    "runtime_mode": "enforcing",
    "static_mode": "enforcing"
}

Stamp: 2026-05-20T10:31:46.290518Z
Actor: emit_net_naming_scheme
Phase: Checks
Type: TargetKernelCmdlineArgTasks
Message_data:
{
    "to_add": [
        {
            "key": "net.naming-scheme",
            "value": "rhel-9.0"
        }
    ],
    "to_remove": []
}
Stamp: 2026-05-20T10:31:44.174228Z
Actor: check_se_linux
Phase: Checks
Type: TargetKernelCmdlineArgTasks
Message_data:
{
    "to_add": [],
    "to_remove": [
        {
            "key": "enforcing",
            "value": "1"
        }
    ]
}

Stamp: 2026-05-20T10:31:46.289923Z
Actor: emit_net_naming_scheme
Phase: Checks
Type: UpgradeKernelCmdlineArgTasks
Message_data:
{
    "to_add": [
        {
            "key": "net.naming-scheme",
            "value": "rhel-9.0"
        }
    ],
    "to_remove": []
}
Stamp: 2026-05-20T10:31:44.173109Z
Actor: check_se_linux
Phase: Checks
Type: UpgradeKernelCmdlineArgTasks
Message_data:
{
    "to_add": [],
    "to_remove": [
        {
            "key": "enforcing",
            "value": "1"
        }
    ]
}


# grubby --info ALL | grep args
args="ro resume=/dev/mapper/rhel-swap rd.lvm.lv=rhel/swap crashkernel=2G-64G:256M,64G-:512M net.naming-scheme=rhel-9.0"
args="ro resume=/dev/mapper/rhel-swap rd.lvm.lv=rhel/swap crashkernel=2G-64G:256M,64G-:512M net.naming-scheme=rhel-9.0"
args="ro resume=/dev/mapper/rhel-swap rd.lvm.lv=rhel/swap crashkernel=2G-64G:256M,64G-:512M net.naming-scheme=rhel-9.0"

```

PR #1533, opened 2026-05-11
pulseaudiocheck: fix for users with "None" home directory

julusp:
If the user has single entry in /etc/passwd without any colons, the reported pw_dir is NoneType. This makes the os.path.join() crash as NoneType is unexpected.
While technically, entry without the colons is considered invalid and not POSIX compliant, it has been seen that some customers are using such entries and leapp is already taking this possibility into account.
This commit modifies the behaviour of the scanpulseaudio actor, so such entries are logged and skipped.

Jira: RHEL-174874


matejmatuska (commented):
LGTM, however since I did the unit test I will let someone else take a quick look.

matejmatuska:
I would squash this, but make sure to keep @julusp authorship.

PeterMocary (approved):
LGTM!

Tested with an empty home dir specified for `tstusr` in `/etc/passwd` and the user was skipped with the expected debug message.

matejmatuska:
I squashed it locally, when the actions finish we can merge.

PR #1534, opened 2026-05-11
conftest: exit actor context when switching repos

MichalHe:
Prior to this patch, the current_actor_context fixture was always exited when executing a new actor. Exiting the fixture causes sys.meta_path to be restored to a state when the fixture was first entered. Therefore, if we are running tests across multiple repositories, say common and el9toel10, then when we are executing the first actor from el9toel10 we execute `__exit__ `of the last executed actor from the common repository. So the following would happen:
1) Finish running the last actor named 'X' from the common repository.
2) Load el9toel10 repository, appending to sys.meta_path loaders for leapp.libraries.common, etc.
3) Discovery of the tests of the first actor from el9toel10.
4) Detect that we are executing a different actor than 'X', so call
   cleanup, which restores the state of sys.meta_path. Therefore, this effectively removes all loaders for el9toel10 repository from  sys.meta_path.
5) Execute tests for the first actor from el9toel10.

This patch ensures that we call `current_actor_context.__exit__` whenever we are about to start loading a new repository, preventing us from unwanted removal of the loaders in sys.meta_path (added by the loaded repository).

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -12,24 +12,42 @@ logging.getLogger("parso").setLevel(logging.INFO)
 
 
 def _load_and_add_repo(manager, repo_path):
-    repo = find_and_scan_repositories(
+    manager_with_new_repos = find_and_scan_repositories(
         repo_path,
         include_locals=True
     )
-    unloaded = set()
-    loaded = {r.repo_id for r in manager.repos}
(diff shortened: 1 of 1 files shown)


matejmatuska (approved):
Nice!

PR #1536, opened 2026-05-12
fix(target_initramfs): always regenerate initramfs

MichalHe:
Remove conditions that would skip target initramfs regeneration if there are no dracut or kernel modules to be included. Instead, regenerate the target initramfs always, as there might be new configs written by us after the upgrade RPM transaction installs the target kernel and generates a corresponding initramfs. The initramfs generated during RPM transaction, therefore lacks or contains invalid config files.

Jira: RHEL-151509


matejmatuska:
Is there any ticket associated with this? Or reproducer?

In general, I think it makes sense to do this.

pirat89:
We can link it to the ticket with multipath.

pirat89 (approved):

PR #1537, opened 2026-05-14
fix(breadcrumbs): use os-release pretty name for source_os

prjagtap-rh:
source_os was always formatted as RHEL plus version.source, so non-RHEL sources (e.g. CentOS Stream 9) appeared as "Red Hat Enterprise Linux 9".

Set source_os from IPUConfig.os_release.pretty_name when set, with the previous RHEL-style string as fallback. target_os unchanged.

Migration results and RHSM leapp facts pick up the same source_os value.

JIRA: RHEL-156580

### DIFF
--- a/commands/command_utils.py
+++ b/commands/command_utils.py
@@ -39,6 +39,20 @@ class DistroIDs(str, Enum):
     ALMALINUX = 'almalinux'
 
 
+DISTRO_NAMES = {
+    DistroIDs.RHEL: 'Red Hat Enterprise Linux',
+    DistroIDs.CENTOS: 'CentOS Stream',
+    DistroIDs.ALMALINUX: 'AlmaLinux',
+}
+"""
+Maps distro IDs to their user-facing display names (matching NAME in /etc/os-release).
+
+NOTE: keep in sync with distro_id_to_pretty_name() in
+repos/system_upgrade/common/libraries/distro.py the repo library
+is not importable from the CLI layer.
+"""
+
+
 _DISTRO_VERSION_FORMATS = {
     DistroIDs.RHEL: VersionFormats.MAJOR_MINOR,
     DistroIDs.CENTOS: VersionFormats.MAJOR_ONLY,
--- a/commands/upgrade/breadcrumbs.py
+++ b/commands/upgrade/breadcrumbs.py
@@ -6,6 +6,7 @@ from functools import wraps
 from itertools import chain
 
 from leapp import FULL_VERSION
+from leapp.cli.commands.command_utils import DISTRO_NAMES
 from leapp.libraries.stdlib.call import _call
 from leapp.utils.audit import get_messages
 
@@ -13,6 +14,20 @@ try:
     from json.decoder import JSONDecodeError  # pylint: disable=ungrouped-imports
 except ImportError:
     JSONDecodeError = ValueError
+
+
+def _get_os_name(ipu_cfg, direction):
+    """
+    Determine the OS display name from IPUConfig data.
+
+    :param direction: 'source' or 'target'
+    """
+    distro = (ipu_cfg.get('distro') or {}).get(direction, '')
+    version = (ipu_cfg.get('version') or {}).get(direction, '')
+    if not distro:
+        return 'N/A'
+    distro_name = DISTRO_NAMES.get(distro, 'unknown')
+    return '{} {}'.format(distro_name, version or 'N/A')
 
 
 def runs_in_container():
@@ -95,10 +110,9 @@ class _BreadCrumbs:
         self._crumbs['run_id'] = os.environ.get('LEAPP_EXECUTION_ID', 'N/A')
         self._crumbs['leapp_file_changes'].extend(self._verify_leapp_pkgs())
         messages = get_messages(('IPUConfig',), self._crumbs['run_id'])
-        versions = json.loads((messages or [{}])[0].get('message', {}).get(
-            'data', '{}')).get('version', {'target': 'N/A', 'source': 'N/A'})
-        self._crumbs['target_os'] = 'Red Hat Enterprise Linux {target}'.format(**versions)
-        self._crumbs['source_os'] = 'Red Hat Enterprise Linux {source}'.format(**versions)
+        ipu_cfg = json.loads((messages or [{}])[0].get('message', {}).get('data', '{}'))
+        self._crumbs['source_os'] = _get_os_name(ipu_cfg, 'source')
+        self._crumbs['target_os'] = _get_os_name(ipu_cfg, 'target')
         self._crumbs['activity_ended'] = datetime.datetime.utcnow().isoformat() + 'Z'
         self._crumbs['env'] = {k: v for k, v in os.environ.items() if k.startswith('LEAPP_')}
         try:
--- a/repos/system_upgrade/common/libraries/distro.py
+++ b/repos/system_upgrade/common/libraries/distro.py
@@ -227,6 +227,10 @@ def distro_id_to_pretty_name(distro_id):
     Get pretty name for the given distro id.
 
     The pretty name is what is found in the NAME field of /etc/os-release.
+
+    NOTE: a parallel mapping exists in commands/command_utils.py
+    (DISTRO_NAMES) for use in the CLI layer which cannot import
+    repo libraries. Keep both in sync when adding a new distro.
     """
     return {
         "rhel": "Red Hat Enterprise Linux",


commands/upgrade/breadcrumbs.py
```diff
@@ -95,10 +95,14 @@ def save(self):
...
         messages = get_messages(('IPUConfig',), self._crumbs['run_id'])
-        versions = json.loads((messages or [{}])[0].get('message', {}).get(
-            'data', '{}')).get('version', {'target': 'N/A', 'source': 'N/A'})
+        ipu_cfg = json.loads((messages or [{}])[0].get('message', {}).get('data', '{}'))
+        versions = ipu_cfg.get('version', {'target': 'N/A', 'source': 'N/A'})
         self._crumbs['target_os'] = 'Red Hat Enterprise Linux {target}'.format(**versions)
-        self._crumbs['source_os'] = 'Red Hat Enterprise Linux {source}'.format(**versions)
+        pretty = (ipu_cfg.get('os_release') or {}).get('pretty_name') or ''
+        if pretty:
+            self._crumbs['source_os'] = pretty
```
matejmatuska: the pretty name look like this:
RHEL - `Red Hat Enterprise Linux 9.8 (Plow)`
Centos - `CentOS Stream 9`
Alma - `AlmaLinux 9.7 (Moss Jungle Cat)`

Now to be honest I am not sure if it's a problem or not, but at least for RHEL I would definitely keep it as it was: `Red Hat Enterprise Linux X.X`.

commands/upgrade/breadcrumbs.py
```diff
@@ -95,10 +95,14 @@ def save(self):
...
+        else:
+            self._crumbs['source_os'] = 'Red Hat Enterprise Linux {source}'.format(**versions)
```
matejmatuska: This would set the source_os incorrectly on distros other than RHEL.

commands/upgrade/breadcrumbs.py
```diff
@@ -95,10 +95,14 @@ def save(self):
         self._crumbs['run_id'] = os.environ.get('LEAPP_EXECUTION_ID', 'N/A')
         self._crumbs['leapp_file_changes'].extend(self._verify_leapp_pkgs())
         messages = get_messages(('IPUConfig',), self._crumbs['run_id'])
-        versions = json.loads((messages or [{}])[0].get('message', {}).get(
-            'data', '{}')).get('version', {'target': 'N/A', 'source': 'N/A'})
+        ipu_cfg = json.loads((messages or [{}])[0].get('message', {}).get('data', '{}'))
+        versions = ipu_cfg.get('version', {'target': 'N/A', 'source': 'N/A'})
         self._crumbs['target_os'] = 'Red Hat Enterprise Linux {target}'.format(**versions)
```
matejmatuska: I know this isn't in the scope of the ticket, but I think we could also do this for the target while we are at it.

matejmatuska (changes requested):

commands/upgrade/breadcrumbs.py
```diff
@@ -15,6 +16,31 @@
...
+
+    Uses os_release NAME + VERSION_ID (e.g. "CentOS Stream 9",
+    "Red Hat Enterprise Linux 9.8") avoids PRETTY_NAME which
+    may contain codenames in parentheses.
+    """
+    os_release = ipu_cfg.get('os_release') or {}
+    name = os_release.get('name', '')
+    version_id = os_release.get('version_id', '')
+    if name and version_id:
+        return '{} {}'.format(name, version_id)
```
PeterMocary: I'm not sure if we really need to try using `os_release`. If we don't use it, these functions could be merged into one.

commands/upgrade/breadcrumbs.py
```diff
@@ -15,6 +16,31 @@
...
+    distro = (ipu_cfg.get('distro') or {}).get('source', 'rhel')
```
PeterMocary: I'm not sure if we want to default to something as now different distros are possible... I'd rather say that we don't know instead of guessing... Let's see what others think.

PeterMocary (dismissed):
I'm not sure about few things but otherwise works as expected.

Tested on:
Centos 9 to Centos 10
```
      "run_id": "ca903cc3-7f4c-4861-83f7-4d0f77a4c90c",
      "source_os": "CentOS Stream 9",
      "success": true,
      "target_os": "CentOS Stream 10",
```

Centos 9 to RHEL 10
```
      "run_id": "4a226f14-fcb9-460f-9ecc-2cf172aafc2f",
      "source_os": "CentOS Stream 9",
      "success": true,
      "target_os": "Red Hat Enterprise Linux 10.3",
      "version": "0.21.0"
```

RHEL 9 to RHEL 10
```
      "run_id": "04c55490-4492-4934-a4d1-e297aec7f6e0",
      "source_os": "Red Hat Enterprise Linux 9.8",
      "success": true,
      "target_os": "Red Hat Enterprise Linux 10.2",
      "version": "0.21.0"
```

RHEL 8 to RHEL 9
```
     "run_id": "1c6d6206-63bb-489a-9995-dbe7bbcbbcdd",
      "source_os": "Red Hat Enterprise Linux 8.10",
      "success": true,
      "target_os": "Red Hat Enterprise Linux 9.9",
      "version": "0.21.0"
```

commands/upgrade/breadcrumbs.py
```diff
@@ -15,6 +16,19 @@
...
+def _get_os_name(ipu_cfg, direction):
+    """
+    Determine the OS display name from IPUConfig data.
+
+    :param direction: 'source' or 'target'
+    """
+    distro = (ipu_cfg.get('distro') or {}).get(direction, '')
+    version = (ipu_cfg.get('version') or {}).get(direction, '')
+    if not distro or not version:
+        return 'unknown'
```
matejmatuska: Currently, when the version is None, it would return 'N/A' so the final string would be e.g. "Red Hat Enterprise Linux N/A" I think we should keep that for backwards compatibility. So if `distro is not None and version is None` I would do that.
 
When both are none I think maybe we should just do N/A also to keep things consistent.

commands/command_utils.py
```diff
@@ -39,6 +39,20 @@ class DistroIDs(str, Enum):
     ALMALINUX = 'almalinux'
 
 
+DISTRO_DISPLAY_NAMES = {
```
matejmatuska: Let's not call it display names, we've considered that before and didn't like it. I would use maybe just DISTRO_NAMES

commands/command_utils.py
```diff
@@ -39,6 +39,20 @@ class DistroIDs(str, Enum):
...
+    DistroIDs.RHEL: 'Red Hat Enterprise Linux',
+    DistroIDs.CENTOS: 'CentOS Stream',
+    DistroIDs.ALMALINUX: 'AlmaLinux',
+}
+"""
+Maps distro IDs to their user-facing display names (matching NAME in /etc/os-release).
+
+NOTE: keep in sync with distro_id_to_pretty_name() in
+repos/system_upgrade/common/libraries/distro.py the repo library
+is not importable from the CLI layer.
```
matejmatuska, line 52: I am not sure about this, the path can easily get out of sync, but then it's probably a good idea to have some warning here to keep them in sync.

matejmatuska (changes requested):
The approach is good, just small suggestions.

commands/command_utils.py
prjagtap-rh: @matejmatuska We already have notes here.
Also by warning do you mean to add test case where it would check DISTRO_NAMES with  repo-layer distro_id_to_pretty_name()

PeterMocary (approved):
LGTM!

I've tested again on RHEL8, RHEL9, upgraded RHEL9 and CentOS 9.

Upgrading RHEL8
```
"run_id": "9cf193d6-b53d-4b04-922a-0dc903105161",
"source_os": "Red Hat Enterprise Linux 8.10",
"success": true,
"target_os": "Red Hat Enterprise Linux 9.9",
"version": "0.21.0"
```

Upgrading RHEL9
```
"run_id": "a564d8f4-c607-47c6-9b50-f45bd2020a5e",
"source_os": "Red Hat Enterprise Linux 9.8",
"success": true,
"target_os": "Red Hat Enterprise Linux 10.2",
"version": "0.21.0"
```

Upgrading RHEL9 which was upgraded from RHEL 8
```
"run_id": "7dae9145-b176-4afb-bf0f-f6f3f2266c83",
"source_os": "Red Hat Enterprise Linux 9.9",
"success": true,
"target_os": "Red Hat Enterprise Linux 10.3",
"version": "0.21.0"
```

Upgrading CentOS 9
```
"run_id": "948dc48c-814d-46d0-8510-17fea5f3d059",
"source_os": "CentOS Stream 9",
"success": true,
"target_os": "CentOS Stream 10",
"version": "0.21.0"
```

Upgrading and converting CentOS 9
```
"run_id": "b3979e3b-1f4e-49c7-a758-2163c3c601d8",
"source_os": "CentOS Stream 9",
"success": true,
"target_os": "Red Hat Enterprise Linux 10.3",
"version": "0.21.0"
```

commands/command_utils.py
PeterMocary: I don't think we need a test here, @matejmatuska just wanted to point out that leaving references such as this one puts another layer of keeping things in sync. Now you need to keep the path and function name in sync even if the dict itself doesn't change. Thus, such references should be generally avoided in my opinion. However, it is helpful to keep a note here, I'd just strip the exact function and path.

```suggestion
NOTE: keep in sync with pretty names in the common distro library
which is not importable from the CLI layer.
```

I'm ok with the original as well, just making sure the point is heard for the future. :)

matejmatuska (approved):
LGTM

PR #1539, opened 2026-05-19
Merge and refactor of repo-related actors (1/2)

karolinku:
## Summary
Consolidate `repositoriesmapping`, `peseventsscanner`, `repositoriesblacklist`, and `setuptargetrepos `into a single actor: `targetcontentresolver`. New actor executes four stages (previously handled by separate actors) sequentially:
  - Load repository mapping (repomap.json) and produce RepositoriesMapping
  - Compute blocklisted target repos (e.g. unsupported CRB)
  - Process PES events to determine RPM transaction tasks
  - Build final list of target repositories

Data that previously flowed through the messages between these actors is now passed directly as function arguments, simplifying the workflow and eliminating the circular dependency workaround between libraries. The workflow is orchestrated from the `targetcontentresolver` library.

Related models & messages have been updated to reflect new changes.

Introduce priority order when handling conflicting requests; listed top-down where top has the highest priority:

  * excplicit user request (`--enablerepo <repoid>`; `CustomTargetRepositories`)
  * external request to block a repoid
  * external request to enable a repoid
  * internal request to block a repoid (applied for CRB repos)

External requests are made by other actors with `RepositoriesSetupTasks` and
`RepositoriesBlacklisted` messages.

Also several bugs and anti-patterns that we discovered are fixed now (mainly around generated reports).
Some changes require however more significant redesign of the new actor and we postpone such changes for future.

All tests have been adjusted with help of AI - code coverage is similar to the previous state.

## Details of additional changes

* Models:
  * The `RepositoriesBlacklisted` model has been deprecated. The original behaviour could not be preserved, so it now behaves as a task to block specified repositories. (documented)
  * Introduce `to_block` field in `RepositoriesSetupTasks`. It replaces one of original purposes of `RepositoriesBlacklisted` model.
  * Introduce `RepositoriesBlocklisted` model, which represents facts about blocklisted repositories.

* Reports:
  * Synchronize behaviour between systems with enabled CRB repositories and upgrades with explicitely enabled repositories.
  * Drop the `Using repository not supported by Red Hat` report. The report was poorly worded and it is confusing in the context of current changes and upgrades on other distributions.
  * Drop most of irrelevant CRB repositories from the report. The correct fix requires another level of the refactoring.

* Renaming of libraries (private actor libs):
  * `repositoriesblacklist` -> `repositoriesblocklist`
  * `repositoriesmapping` -> `repomap_loader`
  * `repomap` -> `repomap_calc`

* Code:
  * Adjust wording to use `blocklist` terminology instead of `blacklist`
  * The `targetcontentresolver` library acts as orchestrator in the actor.
    All messages are consumed in this actor and propagated to the rest of
    libraries as needed.
  * Originally most of actors required repositories mapping which has been used
    to initialize `RepomapDataHandler` - that is used for all repomap related
    operations. So provide `RepomapDataHandler` instead to drop duplicate
    processing.
  * Dropping an obsolete and dead code.
  * Most of docstrings have been updated and polished to satisfy expected format.
    But some are kept for future (due to number of changes and relations between
    commits).

Unit-tests have been adjusted with an assistance of AI.
Read the commit messages for more details about all changes.

Jira: RHEL-115867

---

## TODOLIST:
- [x] do not transmit RepositoriseMapping between libraries - rather work just with the `RepomapDataHandler`
- [x] most likely reports about blocklisted repositories will need to be polished and the logic around updated as well.
- [x] refactor setuptargetrepositories
  - NOTE: partially done. See the last commit. Another changes expected in future
- [x] refactor pes_events_scanner
  - NOTE: partially done. See the last commit. Another changes expected in future
- [x] use apostrophes / single-quotes in libraries - most of the code is written this way, but some libs use double quotes
- [x] update of unit-tests

---
## Code coverage

(I know that code coverage is kinda misleading info, but keeping it her for the comparison of original and current state)

```
┌──────────────────────────┬──┬───────────┬────────────┐
│           File           │  │ New Cover │ Orig Cover │
├──────────────────────────┼──┼───────────┼────────────┤
│ repositoriesblocklist.py │  │      100% │ 91%        │
├──────────────────────────┼──┼───────────┼────────────┤
│ repomap_calc.py          │  │      99%  │ 99%        │
├──────────────────────────┼──┼───────────┼────────────┤
│ targetcontentresolver.py │  │      93%  │ ---        │
├──────────────────────────┼──┼───────────┼────────────┤
│ repomap_loader.py        │  │      92%  │ 91%        │
├──────────────────────────┼──┼───────────┼────────────┤
│ pes_events_scanner.py    │  │      88%  │ 84%        │
├──────────────────────────┼──┼───────────┼────────────┤
│ setuptargetrepos.py      │  │      87%  │ 89%        │
├──────────────────────────┼──┼───────────┼────────────┤
│ pes_event_parsing.py     │  │      42%  │ 42%        │
├──────────────────────────┼──┼───────────┼────────────┤
│ TOTAL                    │  │      83%  │ ---        │
└──────────────────────────┴──┴───────────┴────────────┘

```


repos/system_upgrade/common/actors/targetcontentresolver/actor.py
```diff
@@ -0,0 +1,59 @@
...
+    consumes = (
+        CustomTargetRepository,
+        DistributionSignedRPM,
+        EnabledModules,
+        InstalledRPM,
+        RepositoriesFacts,
+        RepositoriesSetupTasks,
+        RHUIInfo,
+        RpmTransactionTasks,
+        UsedRepositories,
```
pirat89, line 51: discussed via call, the `RepositoriesBlacklisted` messages need to be still consumed as there is no other way how to explicitly forbid some repositories. We are aware that in the same time we would like to have also an information about blocked repositories as it could be helpful for purpose of finvestigation. Originally, the message server both purposes, which it cannot anymore. So the introduction of a new message providing such an information could help with that - if we agree that we want to keep this information as part of messages and do not want to keep it on logging.

Regarding that we have several options (just a simplified sketch from the brainstorming):

```
1) same behaviour                                                               
- RepositoriesBlacklisted                                                       
- RepositoriesBlacklistedFacts                                                  
                                                                                
2) introduce new msg and use old one as a fact                                  
- RepositoriesSetupTasks.blocklist                                              
- RepositoriesBlacklisted --- fact what was blacklisted                         
                                                                                
3)                                                                              
- RepositoriesBlacklisted - deprecated use, suggest ↓ instead                   
- RepositoriesSetupTasks.blocklist                                              
- RepositoriesBlocklisted - facts

4) just consume `RepositoriesBlacklisted` and do not produce any other msg
```

We will discuss yet what way to go, however, as discussed, feel free to implement one of solutions in advance to not be blocked on that.

repos/system_upgrade/common/actors/targetcontentresolver/libraries/repomap_calc.py
pirat89, line 1: Discussed via call, we do not expose this library. In case we would like to expose it, the refactoring would not be needed so much as now. The whole problem is that we want to keep it private as no other actors should use it. We consider the repomap schema dynamic and we do not want to be limited here to make significant (breaking changes.

repos/system_upgrade/common/actors/targetcontentresolver/libraries/pes_events_scanner.py
```diff
@@ -3,7 +3,7 @@
 
 from leapp import reporting
 from leapp.exceptions import StopActorExecutionError
-from leapp.libraries.actor import peseventsscanner_repomap
+from leapp.libraries.common import repomap as peseventsscanner_repomap
```
pirat89: we prefer to get rid of the library renaming. it's ok now as an interim step, but for the final version we want to get rid of these renames, so the unified name is used everywhere in this actor.

repos/system_upgrade/common/actors/targetcontentresolver/libraries/targetcontentresolver.py
pirat89, line 1: Exactly something like that has been in my mind to start with. I just suggest to rename other `process` functions in actor's libs so it's not confusing what is actually the main process function - keeping just this one as it is. We can analyze other functions/code that could be moved to this one lib so simplify the actor (if find anything) but it's ok otherwise to me.

repos/system_upgrade/common/actors/targetcontentresolver/libraries/setuptargetrepos.py
```diff
@@ -84,16 +60,25 @@ def _get_mapped_repoids(repomap, src_repoids):
 
 
 @suppress_deprecation(RHELTargetRepository)
```
pirat89, line 58: As part of the PR I suggest to drop all deprecated models related to these actors that can be already removed.

pirat89 (changes requested):
Summarizing our discussion on this change in comments and adding another nitpicks I found during brief review.

As another step - which I skipped during the first review as I focused now mainly just on changes and I did not go through the whole (unchanged) code -  we should check what other pieces of code we could simplify thanks to the merge. It's possible we will keep it basically just split in separate libs, but I would like to give it some time to see whether we can simplify some of libs (to e.g. deduplicate some logic).

repos/system_upgrade/common/actors/targetcontentresolver/libraries/pes_events_scanner.py
karolinku: renaming removed.

repos/system_upgrade/common/actors/targetcontentresolver/actor.py
karolinku: I restored RepositoriesBlacklisted message produce. Let's discuss further changes.

repos/system_upgrade/common/actors/targetcontentresolver/libraries/repomap_calc.py
karolinku: The library is now moved private.

repos/system_upgrade/common/actors/targetcontentresolver/actor.py
matejmatuska: I would go with 2 or maybe 3.

repos/system_upgrade/common/models/repositoriesblocklisted.py
```diff
@@ -1,11 +1,23 @@
...
     topic = SystemFactsTopic
 
     repoids = fields.List(fields.String(), default=[])
+
+
+class RepositoriesBlocklisted(RepositoriesBlacklisted):
+    """
+    Repository IDs that should be ignored by Leapp during the upgrade process.
+    """
+    topic = SystemFactsTopic
```
pirat89, line 36: I thought about the inheritance in the opposite way  - so when we remove the deprecated model, we can just remove the code instead of updating the new model again. Also I realied the message for the deprecation and docstrings need to be extended - as the original model is replaced by 2 new models, depends on the use case.

```suggestion
class RepositoriesBlocklisted(Model):
    """
    Provide list of repositories blocklisted during the in-place upgrade.
    
    This message is expected to be produced just by one actor,
    but it can be consumed by others to check what repositories have been ignored
    during the upgrade process. 
    """
    topic = SystemFactsTopic

    repoids = fields.List(fields.String(), default=[])
    """
    The list of blocklisted repoids.
    """

@deprecated(
    since='2026-06-01',
    message=(
        'The model has been deprecated and replaced by newer ones:'
        ' a) Use RepositoriesBlocklisted if you want to get information about blocklisted repositories.'
        ' b) Use RepositoriesSetupTasks if you want to block explicitly some repositories during the upgrade.'
    )
class RepositoriesBlacklisted(RepositoriesBlocklisted):
    """
    Repository IDs that should be ignored by Leapp during the upgrade process.
    """
```

repos/system_upgrade/common/actors/targetcontentresolver/actor.py
pirat89: Discussed on cabal - 3 is the way.

repos/system_upgrade/common/actors/targetcontentresolver/actor.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+    produces = (
+        ConsumedDataAsset,
+        PESRpmTransactionTasks,
+        Report,
+        RepositoriesBlacklisted,
```
pirat89: hmm.. Actually, IIRC we discussed that the deprecated message should stay in the use of the "task" instead of the "fact", as  we discussed that:
* hardly someone consumes it to see what is ignored
* and most likely if someone uses it they want to block some repositories.

Correct me if I remember it wrong.

repos/system_upgrade/common/actors/targetcontentresolver/libraries/targetcontentresolver.py
pirat89: @karolinku  this one is missed.

repos/system_upgrade/common/actors/targetcontentresolver/libraries/targetcontentresolver.py
```diff
@@ -0,0 +1,42 @@
...
+
+    blacklisted_repoids = repositoriesblacklist.process(repositories_map_msg)
+    external_setup_tasks = list(api.consume(RepositoriesSetupTasks))
+    custom_blocklist_repoids = set()
+    for task in external_setup_tasks:
+        custom_blocklist_repoids.update(task.blocklist)
+    full_blocklist_repoids = blacklisted_repoids.union(custom_blocklist_repoids)
+    if full_blocklist_repoids:
+        api.produce(RepositoriesBlocklisted(repoids=sorted(full_blocklist_repoids)))
+        api.produce(RepositoriesBlacklisted(repoids=sorted(full_blocklist_repoids)))
```
pirat89: I would expect to hide this logic inside the repositoriesblacklist library, so we get already the whole list of blocklisted repoids instead of doing another calculation here. The library is expected to deal with it completely already. If the reason for this is to load the external repo setup tasks just once, then load it first and give relevant arguments to the executed functions. In such a case, I would create named tuple `tasks` with `blocklist` and `to_enable` fields as set. The produce of messages can stay then in the blacklisted_repoids as well.

repos/system_upgrade/common/actors/targetcontentresolver/libraries/targetcontentresolver.py
```diff
@@ -0,0 +1,42 @@
...
+
+    setup_tasks = pes_events_scanner.process(repositories_map_msg, full_blocklist_repoids)
+
+    setuptargetrepos.process(repositories_map_msg, internal_setup_tasks=setup_tasks,
+                             blacklisted_repoids=full_blocklist_repoids, external_setup_tasks=external_setup_tasks)
```
pirat89: I think we need to come up with a better naming and actually refactor this part.
* external setup tasks can be ok. But do we use anything else than `to_enable` there? As blocklisted task is separated, it seems we could call it just `external_repoids_requests`
* In case of `internal_setup_tasks` we speak literally just  about `to_enable` repoids based on PES. The message RepositoriesSetupTask is not needed here then from my POV. Why we do not just provide the list of repositories to enable to simplify it? Then we could name it e.g. `pes_requested_repoids`

pirat89 (changes requested):
I am still going through the code. Leaving some comments for some parts I already think about to change. wdyt?

repos/system_upgrade/common/actors/targetcontentresolver/libraries/setuptargetrepos.py
pirat89: Discussed on a call. It has been unfortunately deprecated recently and new version is going to be published before the protection period ends. So keeping it still present.

repos/system_upgrade/common/actors/targetcontentresolver/actor.py
matejmatuska: Tbh I am not sure what we decided, but I think this is a strong argument and I would vote for this too.

repos/system_upgrade/common/actors/targetcontentresolver/libraries/setuptargetrepos.py
```diff
@@ -84,24 +58,35 @@ def _get_mapped_repoids(repomap, src_repoids):
...
-def process():
+def setup_target_repos(repositories_map_msg, pes_requested_repoids=None,
+                       blacklisted_repoids=None, external_repoids_requests=None):
```
matejmatuska: Why the default arguments? The function is only called in one place and we don't really expect that to change. The default arguments then require special handling in the code.

If it's because of tests, I would suggest the tests to just pass in `set()` instead of this.

repos/system_upgrade/common/actors/targetcontentresolver/libraries/setuptargetrepos.py
```diff
@@ -84,24 +58,35 @@ def _get_mapped_repoids(repomap, src_repoids):
...
     used_repoids_dict = _get_used_repo_dict()
     enabled_repoids = _get_enabled_repoids()
-    excluded_repoids = _get_blacklisted_repoids()
+    excluded_repoids = blacklisted_repoids if blacklisted_repoids is not None else set()
     custom_repos = _get_custom_target_repos()
     repoids_from_installed_packages = _get_repoids_from_installed_packages()
 
     # Setup repomap handler
-    repo_mappig_msg = next(api.consume(RepositoriesMapping), RepositoriesMapping())
+    repo_mapping_msg = repositories_map_msg
```
matejmatuska: Just use one of these.

repos/system_upgrade/common/actors/targetcontentresolver/libraries/pes_events_scanner.py
```diff
@@ -323,11 +320,14 @@ def make_summary_entry_for_skipped_pkg(pkg):
         api.current_logger().info(summary)
 
 
-def remove_new_packages_from_blacklisted_repos(source_pkgs, target_pkgs):
+def remove_new_packages_from_blacklisted_repos(source_pkgs, target_pkgs, blacklisted_repoids=None):
```
matejmatuska: Again, why the default argument?

pirat89:
At this moment I am changing the code withou updates of unit-tests. Keeping unit tests for later due to dynamic nature of modifications I do at this moment. In the next step I am going to update work with the repositories mapping.

pirat89:
Renamed libraries and test files to reflect new names:
* repomap -> repomap_calc
* repositoriesmapping -> repomap_loader
* repositoriesblacklist -> repositoriesblocklist

All imports and calls are adjusted. More details in commit messages.

What's missing?
* do not transmit RepositoriseMapping between libraries - rather work just with the `RepomapDataHandler`
  *  I am considering to rename it actually to something like: RepomapCalculations or something like that. WDYT? We can keep the name as well, I am just thinking about it.
* Deprecate RepositoriesMapping message - we have in the docstring that we can do any changes without deprecations and people should not use it, but I will feel better when it's officially marked as deprecated.
  * We discussed with @matejmatuska that it can be nice source of facts for debugging purposes, so we do not have to remove it any time soon. As well, maybe we can in the meanwhile (before the removal) introduce new types of models/messages, like models that can be only produced but cannot be consumed. But that's just for a future plans, after we deal with current goals with higher priority.
  * Thinking about this step more, I am starting to feel that we should keep it as it is. Maybe deprecating it to emphasize that nobody else should work with the message, but I will think about it yet. Realizing that the removal would just require to create similar classes inside actors anyway and this already helps us to deal with some checks.

* most likely reports about blocklisted repositories will need to be polished and the logic around updated as well.
* refactoring of setuptargetrepositories and pes_events_scanner libs
* update of unit-tests

repos/system_upgrade/common/actors/targetcontentresolver/actor.py
pirat89: Updated. It's used now as a task instead of fact.

repos/system_upgrade/common/actors/targetcontentresolver/libraries/repositoriesblocklist.py
```diff
@@ -0,0 +1,186 @@
...
+    # upgrade, we should ignore any other as well.
+    filtered_repos_to_exclude = repos_to_exclude - manually_enabled_repos
+
+    if manually_enabled_repos:
+        # FIXME(pstodulk): This is wrong. This is valid only for RHEL systems
+        # and the report itself is confusing - as it is actually only about
+        # CRB repositories and not any other.
+        _report_using_unsupported_repos(manually_enabled_repos)
+    if filtered_repos_to_exclude:
+        _report_excluded_repos(filtered_repos_to_exclude)
```
pirat89: hmm... actually, I understand now why it seems wrong. the report about excluded repos should be really in the entry function of this library or later. as it was really supposed to inform about all blocklisted repositories. let's think about them later.

pirat89 (dismissed):
Waiting for another approval as co-author of the changes. We fixed several problems we discovered, however, by design, fixing of some other issues we know about requires additional level of refactoring. Our thoughts about another refactoring steps are described in the last TODO commit. We consider the current state as a good enough step.

pirat89:
I realized I merged by mistake 2 commits which I wanted to keep separate. Fixed. The resulting code is 1:1 to the previous state when tests passed (except the one failing on infra issues)

pirat89 (approved):

repos/system_upgrade/common/actors/targetcontentresolver/libraries/repositoriesblocklist.py
```diff
@@ -0,0 +1,166 @@
...
+    api.current_logger().info(
+        'The CRB repository is not enabled. Excluding {} from the upgrade'.format(repos)
+    )
+
+    report = [
+        reporting.Title('Excluded target system repositories'),
+        reporting.Summary(
+            'The following repositories are not supported by '
+            'Red Hat and are excluded from the list of repositories '
+            'used during the upgrade.\n- {}'.format('\n- '.join(repos))
```
matejmatuska, line 18: Maybe we could use the format_list already?

repos/system_upgrade/common/actors/targetcontentresolver/tests/test_pes_event_scanner.py
matejmatuska, line 522: model_instances is now just a single instance so the loop does just one iteration, but not a blocker.

repos/system_upgrade/common/actors/targetcontentresolver/tests/test_repositoriesblocklist.py
```diff
@@ -0,0 +1,364 @@
...
+    repomap_opts_only.repositories[1].pesid = 'test_pesid'
+
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(
+        arch='x86_64', src_ver='8.10', dst_ver='9.6', dst_distro='rhel'
+    ))
+    monkeypatch.setattr(reporting, 'create_report', create_report_mocked())
+
+    handler = RepoMapDataHandler(repomap_opts_only)
+    result = repositoriesblocklist._calc_internal_blocklist(
+        handler, _NO_TASKS, enabled_repoids=set()
```
matejmatuska, line 81: I think this test would pass regardless of whether the crb pesid is in mapping or not, because `enabled_repois == set()` i.e. empty set which is disjoint with all sets so `_are_crb_repos_disabled` will return True either way.

repos/system_upgrade/common/actors/targetcontentresolver/tests/test_setuptargetrepos.py
```diff
@@ -77,19 +86,23 @@ def test_custom_repos(monkeypatch):
...
+    to the resulting TargetRepositories (and blocklist filtering is applied to them).
     """
-    repositories_setup_tasks = RepositoriesSetupTasks(to_enable=['rhel-8-server-rpms',
-                                                                 'rhel-8-blacklisted-rpms'])
-    repos_blacklisted = RepositoriesBlacklisted(repoids=['rhel-8-blacklisted-rpms'])
     repositories_mapping = RepositoriesMapping(mapping=[], repositories=[])
-    msgs = [repositories_setup_tasks, repos_blacklisted, repositories_mapping]
+    msgs = [repositories_mapping]
 
     monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(msgs=msgs))
```
matejmatuska, line 95: I think this is not necessary anymore

matejmatuska: Same in almost all of these tests.

matejmatuska (commented):
Okay I review the code, not tested yet.

Code looks okay I found a few minor improvements, but overal I think it looks good. I was "live" reviewing when this was prepared and also discussed a some of the decisions and the overal solution  so not much to add to that.

I can approve after I test it works ok

matejmatuska (approved):
I did basic testing, everything looks alright, CI is passing except for the one test on aws which is an infra issue. Merging.

matejmatuska:
The minor suggestions from my review will be addressed in a followup PR.

PR #1541, opened 2026-05-25
Add inhibitor for invalid fstab entries on pseudo-filesystem mountpoints

prjagtap-rh:
Detect /etc/fstab entries that incorrectly define pseudo-filesystem
mountpoints (/proc, /sys, /dev/shm, /run, etc.) with block devices or
wrong filesystem types. Such configurations have been invalid since
RHEL 7 but are typically ignored during normal boot. During the upgrade
the system applies these entries as configured, leading to failures.
    
Use an allowlist of valid virtual filesystem sources and types so
entries referencing UUID=, LABEL=, PARTUUID=, or /dev/* paths are
correctly detected and inhibited.
    
JIRA: RHEL-17842

### DIFF
--- a/repos/system_upgrade/common/actors/checkfstabapifsoverride/actor.py
+++ b/repos/system_upgrade/common/actors/checkfstabapifsoverride/actor.py
@@ -1,0 +1,26 @@
+from leapp.actors import Actor
+from leapp.libraries.actor.checkfstabapifsoverride import check_fstab_api_fs_override
+from leapp.models import StorageInfo
+from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckFstabApiFsOverride(Actor):
+    """
+    Inhibit upgrade if /etc/fstab contains invalid entries for pseudo-filesystem mountpoints.
+
+    Paths like /proc, /sys, /dev/shm, /run are managed by the kernel and
+    systemd and have specific requirements for their fstab entries. Defining
+    them with block devices (via /dev/*, UUID=, LABEL=, etc.) or wrong
+    filesystem types has been invalid since RHEL 7, but such configurations
+    are typically ignored during normal boot. During the upgrade, these entries
+    are applied as configured, which may lead to failures. This actor ensures
+    users correct these entries before proceeding.
+    """
+    name = 'check_fstab_api_fs_override'
+    consumes = (StorageInfo,)
+    produces = (Report,)
+    tags = (ChecksPhaseTag, IPUWorkflowTag)
+
+    def process(self):
+        check_fstab_api_fs_override()
--- a/repos/system_upgrade/common/actors/checkfstabapifsoverride/libraries/checkfstabapifsoverride.py
+++ b/repos/system_upgrade/common/actors/checkfstabapifsoverride/libraries/checkfstabapifsoverride.py
@@ -1,0 +1,80 @@
+from leapp import reporting
+from leapp.libraries.stdlib import api, format_list
+from leapp.models import StorageInfo
+
+API_FS_EXPECTED_TYPES = {
+    '/dev/shm': {'tmpfs'},
+    '/dev/pts': {'devpts'},
+    '/dev/mqueue': {'mqueue'},
+    '/dev/hugepages': {'hugetlbfs'},
+    '/proc': {'proc'},
+    '/sys': {'sysfs'},
+    '/sys/fs/cgroup': {'cgroup', 'cgroup2', 'tmpfs'},
+    '/sys/fs/selinux': {'selinuxfs'},
+    '/sys/kernel/debug': {'debugfs'},
+    '/sys/kernel/security': {'securityfs'},
+    '/run': {'tmpfs'},
+    '/run/lock': {'tmpfs'},
+}
+
+
+def check_fstab_api_fs_override():
+    """
+    Check fstab entries targeting pseudo-filesystem mountpoints for invalid sources or types.
+
+    An entry is flagged if its mount point is a known pseudo-filesystem path and
+    either the filesystem type or the source device does not match the expected
+    virtual filesystem set. The 'none' pseudo-source is also accepted.
+    """
+    storage_info = next(api.consume(StorageInfo), None)
+    if not storage_info:
+        return
+
+    bad_entries = []
+    for entry in storage_info.fstab:
+        mount_point = entry.fs_file
+        if mount_point != '/':
+            mount_point = mount_point.rstrip('/')
+
+        expected = API_FS_EXPECTED_TYPES.get(mount_point)
+        if not expected:
+            continue
+
+        if entry.fs_vfstype not in expected or entry.fs_spec not in (expected | {'none'}):
+            bad_entries.append(entry)
+
+    if not bad_entries:
+        return
+
+    entries_text = format_list(
+        [e.fs_file for e in bad_entries],
+    )
+
+    reporting.create_report([
+        reporting.Title(
+            'Detected invalid entries in /etc/fstab for pseudo-filesystem mountpoints'
+        ),
+        reporting.Summary(
+            'Detected /etc/fstab entries that incorrectly define pseudo-filesystem '
+            'mountpoints. These mountpoints (e.g. /proc, /sys, /dev/shm, /run) are '
+            'managed by the kernel and systemd, and their fstab entries must use the '
+            'correct virtual filesystem type and source. Defining them with a block '
+            'device (via /dev/*, UUID=, LABEL=, etc.) or a wrong filesystem type is '
+            'invalid. Such configurations were common on RHEL 6 and earlier and are '
+            'typically ignored during normal boot. However, during the upgrade the '
+            'system applies fstab entries as configured, which may lead to failures.\n'
+            'Problematic entries: {}'.format(entries_text)
+        ),
+        reporting.Remediation(
+            hint=(
+                'Remove or correct the problematic entries in /etc/fstab. If an entry '
+                'is required (e.g. to set specific mount options for compliance), ensure '
+                'both the source and the filesystem type match the expected virtual '
+                "filesystem (e.g. 'tmpfs /dev/shm tmpfs defaults 0 0').\n"
+                'After editing, run "mount -a" to verify.'
+            )
+        ),
+        reporting.RelatedResource('file', '/etc/fstab'),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.FILESYSTEM, reporting.Groups.INHIBITOR]),
+    ])
--- a/repos/system_upgrade/common/actors/checkfstabapifsoverride/tests/test_checkfstabapifsoverride.py
+++ b/repos/system_upgrade/common/actors/checkfstabapifsoverride/tests/test_checkfstabapifsoverride.py
@@ -1,0 +1,121 @@
+import pytest
+
+from leapp import reporting
+from leapp.libraries.actor.checkfstabapifsoverride import check_fstab_api_fs_override
+from leapp.libraries.common.testutils import create_report_mocked, CurrentActorMocked
+from leapp.libraries.stdlib import api
+from leapp.models import FstabEntry, StorageInfo
+
+
+def _fstab(fs_spec, fs_file, fs_vfstype, fs_mntops='defaults'):
+    return FstabEntry(
+        fs_spec=fs_spec, fs_file=fs_file, fs_vfstype=fs_vfstype,
+        fs_mntops=fs_mntops, fs_freq='0', fs_passno='0',
+    )
+
+
+# --- Entries that SHOULD inhibit (invalid API fs overrides) ---
+_SHM_LVM = _fstab('/dev/mapper/vg00-shm', '/dev/shm', 'xfs')
+_SHM_TMPFS_LABEL = _fstab('LABEL=mydata', '/dev/shm', 'tmpfs')
+_SHM_LVM_TRAILING = _fstab('/dev/mapper/vg00-shm', '/dev/shm/', 'xfs')
+_PROC_EXT4 = _fstab('/dev/sda5', '/proc', 'ext4')
+_SYS_LVM = _fstab('/dev/mapper/vg00-sys', '/sys', 'xfs')
+_SYS_LVM_SYSFS = _fstab('/dev/mapper/vg00-sys', '/sys', 'sysfs')
+_SYS_UUID = _fstab('UUID=023e51f5-b590-4de5-ab21-cf33ecasasas', '/sys', 'xfs')
+_SYS_UUID_SYSFS = _fstab('UUID=023e51f5-b590-4de5-ab21-cf33ecasasas', '/sys', 'sysfs')
+_RUN_PARTUUID = _fstab('PARTUUID=abcd-1234', '/run', 'ext4')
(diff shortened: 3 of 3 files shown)


repos/system_upgrade/common/actors/checkfstabapifsoverride/tests/test_checkfstabapifsoverride.py
```diff
@@ -0,0 +1,87 @@
...
+        ([_SYS_LVM], True),
+        ([_SYS_LVM_SYSFS], True),
+        ([_SYS_UUID], True),
+        ([_SYS_UUID_SYSFS], True),
+        ([_SHM_LABEL], True),
+        ([_RUN_PARTUUID], True),
+        ([_SHM_TMPFS, _PROC_PROC, _SYS_SYSFS], False),
+        ([_PROC_NONE], False),
+        ([_HOME_EXT4, _ROOT_XFS], False),
+        ([], False),
```
matejmatuska, line 83: I would add a bunch more tests for cases we don't want to inhibit, e.g. normal mount (not kernel api fs) with LABEL, UUID, etc.

Also maybe add some cases bind mounts? e.g. the mount opts contain bind, or the fs_vfstype is mount.

repos/system_upgrade/common/actors/checkfstabapifsoverride/actor.py
```diff
@@ -0,0 +1,22 @@
...
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckFstabApiFsOverride(Actor):
+    """
+    Inhibit upgrade if /etc/fstab overrides an API filesystem mountpoint with persistent storage.
+
+    Paths like /proc, /sys, /dev/shm must remain kernel-managed. Overriding
+    them with block devices (via /dev/*, UUID=, LABEL=, etc.) causes the
+    reboot phase to fail.
```
matejmatuska: Afaik some of the mountpoint we have there are no really API filesystems. E.g. `/run` is populated by userspace software, like systemd.

Also, I wouldn't outright say that they cause the reboot to fail, unless we are sure all of them do, otherwise it's misleading. So I would write that they *might* cause the reboot to fail.

matejmatuska: And also, the case in the ticket doesn't cause a fail on a normal reboot, it just fails and crashes leapp during after the reboot during an upgrade.

repos/system_upgrade/common/actors/checkfstabapifsoverride/libraries/checkfstabapifsoverride.py
```diff
@@ -0,0 +1,76 @@
...
+
+    if not bad_entries:
+        return
+
+    entries_text = ', '.join(f'{e.fs_file} ({e.fs_vfstype})' for e in bad_entries)
+
+    reporting.create_report([
+        reporting.Title(
+            'Persistent filesystem mounted over API filesystem mountpoint. '
+            "Upgrade can't proceed"
```
matejmatuska: Don't put phrases like "Upgrade can't proceed" into the report titles. I know there might be a few which do have that, but  in general we don't do it.

matejmatuska: Also, as I mentioned in other comment, I don't think its 100% correct to call the filesystem kernel APIs if some of them really aren't, e.g. /run.

I think maybe we could just call them pseudo-filesystems. Let's see what others think.

repos/system_upgrade/common/actors/checkfstabapifsoverride/libraries/checkfstabapifsoverride.py
```diff
@@ -0,0 +1,76 @@
...
+        ),
+        reporting.Summary(
+            'Detected /etc/fstab entries that incorrectly override kernel-managed API '
+            'filesystem mountpoints. This can be caused by mounting a block device '
+            '(via /dev/*, UUID=, LABEL=, etc.) or using a wrong filesystem type at '
+            'these paths. This will cause the upgrade to fail during the reboot phase.\n'
+            'Offending entries: {}'.format(entries_text)
```
matejmatuska: Again with the kernel API, but unsure about that. What I am sure is that the part about the upgrade failing is not needed.
```suggestion
            'Detected /etc/fstab entries that incorrectly override pseudo '
            'filesystem mountpoints. This can be caused by mounting a block device '
            '(via /dev/*, UUID=, LABEL=, etc.) or using a wrong filesystem type at '
            'these paths.'
            'Offending entries: {}'.format(entries_text)
```

repos/system_upgrade/common/actors/checkfstabapifsoverride/libraries/checkfstabapifsoverride.py
```diff
@@ -0,0 +1,76 @@
...
+        ),
+        reporting.Remediation(
+            hint=(
+                'Remove the offending entries from /etc/fstab. The kernel and systemd '
+                'mount these paths automatically at boot. If an entry is required for '
```
matejmatuska: Are you 100% sure this is true for all of them?

repos/system_upgrade/common/actors/checkfstabapifsoverride/libraries/checkfstabapifsoverride.py
```diff
@@ -0,0 +1,76 @@
...
+                "compliance, ensure both the source and type are the virtual filesystem "
+                "(e.g. 'tmpfs /dev/shm tmpfs defaults 0 0'), not a block device.\n"
+                'After editing, run "mount -a" to verify, then re-run leapp preupgrade.'
```
matejmatuska: definitely don't mention stuff like "then re-run leapp preupgrade." it is kind of expected about most of the reports, even more so if they are an inhibitor. It might even be misleading for the user, because they might have run leapp upgrade.
```suggestion
                'After editing, run "mount -a" to verify.'
```

repos/system_upgrade/common/actors/checkfstabapifsoverride/libraries/checkfstabapifsoverride.py
```diff
@@ -0,0 +1,76 @@
...
+            )
+        ),
+        reporting.RelatedResource('file', '/etc/fstab'),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.FILESYSTEM]),
+        reporting.Groups([reporting.Groups.INHIBITOR]),
```
matejmatuska: Not a blocker, but you can combine these into a single line:


```suggestion
        reporting.Groups([reporting.Groups.FILESYSTEM, reporting.Groups.INHIBITOR]),
```

repos/system_upgrade/common/actors/checkfstabapifsoverride/libraries/checkfstabapifsoverride.py
```diff
@@ -0,0 +1,76 @@
...
+        if not expected:
+            continue
+
+        if entry.fs_vfstype not in expected or entry.fs_spec not in (expected | {'none'}):
+            bad_entries.append(entry)
+
+    if not bad_entries:
+        return
+
+    entries_text = ', '.join(f'{e.fs_file} ({e.fs_vfstype})' for e in bad_entries)
```
matejmatuska: We now have a `format_list` function to format list of items in a report in a standard way. It was introduced in https://github.com/oamg/leapp/pull/913

matejmatuska (dismissed):
The logic looks good to me, the changes are mostly around wording.
**Don't act on the review yet**, let's wait for opinions from others, @pirat89 can you take a look? Also please go throught the logic, I am not that familiar with possible setups and edge cases. 

I also didn't test it yet, will do now.

repos/system_upgrade/common/actors/checkfstabapifsoverride/actor.py
pirat89: It's important to understand that problem is not just that upgrade fails, but that such a definition in FSTAB is invalid basically since RHEL 7 and only reason why it "works" is that such an invalid configurations are usually ignored. The problem it creates during the upgrade is that we tell the system to apply this mess actually as users configured. So we want to be sure that users get rid of it completely.

repos/system_upgrade/common/actors/checkfstabapifsoverride/libraries/checkfstabapifsoverride.py
pirat89: what about to call them just `API filesystem mountpoints` or `system managed API filesystem mountpoints`?  Maybe it would be better if the summary explains better about what we speak, so the term does not have to have the mandatory role here. We can say that these mountpoints are managed by kernel and systemd and they have specific requirements that must be fullfilled.

Pseudo filesystems sounds weird, but I found that the term is used in docs so it seems ok.

repos/system_upgrade/common/actors/checkfstabapifsoverride/libraries/checkfstabapifsoverride.py
pirat89: I suggest to rephrase the hint, saying that we suggest to remove or correct problematic entries

* I would not use word like `offending` as applications do not feel to be offended :-)  
* entries are expected nowadays to be listed if users want to set a specific options to these mounpoints; e.g. limit the size, etc. - e.g. to comply with a certification.

repos/system_upgrade/common/actors/checkfstabapifsoverride/libraries/checkfstabapifsoverride.py
pirat89: also note the title is incorrect as these filesystems are not mounted based on these invalid fstab entries - that's why the system can boot. they are usually ignored during standard system boot. I am more inclining to sth like:
* Detected invalid entries in /etc/fstab for API mountpoints
* Detected invalid entries in /etc/fstab for pseudo-filesystems
* ...

repos/system_upgrade/common/actors/checkfstabapifsoverride/libraries/checkfstabapifsoverride.py
pirat89: I am missing here also explanation why this problem exists. E.g. that such configurations were  common on old systems before the Systemd era, as this is typical mistake made by admins used to RHEL 6 and older systems.

repos/system_upgrade/common/actors/checkfstabapifsoverride/libraries/checkfstabapifsoverride.py
pirat89: also note that information like `The kernel and systemd mount these paths automatically at boot.` is not hint but summary. this is remediation instruction, not explanation of the problem. explanation of the problem is expected to be part of the summary.

repos/system_upgrade/common/actors/checkfstabapifsoverride/libraries/checkfstabapifsoverride.py
prjagtap-rh: @matejmatuska Yes you are right I referred `checkcifs.py` for title. Hence added cant upgrade. But is now corrected.

repos/system_upgrade/common/actors/checkfstabapifsoverride/libraries/checkfstabapifsoverride.py
prjagtap-rh: @matejmatuska thanks for this its of great help will fix it.

repos/system_upgrade/common/actors/checkfstabapifsoverride/libraries/checkfstabapifsoverride.py
```diff
@@ -0,0 +1,81 @@
...
+
+        if entry.fs_vfstype not in expected or entry.fs_spec not in (expected | {'none'}):
+            bad_entries.append(entry)
+
+    if not bad_entries:
+        return
+
+    entries_text = format_list(
+        ['{} ({})'.format(e.fs_file, e.fs_vfstype) for e in bad_entries],
+        callback_sort=None,
```
pirat89: @prjagtap-rh  why? what is the problem with the sorted content? unless there is a good reason for it, the content should be sorted by default - it's preferred way.

also, I think that fs_file should be enough to identify the entry. What additional information user get from the fs_vfstype? from my pov it's not saying what's expected, not whether it's actually wrong. I think that we can stay here with fs_file.

prjagtap-rh: My take on not sorting it so that customer would be able to find it easily. 
If sorted and there are multiple entries then it would be confusing.

Also reason I added filesystem to catch scenarios like below where both entries are wrong. [Its less likely case to have multiple entries]
```
UID=023e51f5-... /sys xfs defaults 0 0        
UUID=023e51f5-... /sys sysfs defaults 0 0

```
In such scenarios if filesystem is not added it would be confusing.

Here I tested it on test system by creating below entries in fstab[Considering worst case fstab file]
```
# Bad entries
/dev/mapper/vg00-shm /dev/shm xfs defaults 0 0
LABEL=mydata /dev/shm tmpfs defaults 0 0
/dev/mapper/vg00-shm /dev/shm/ xfs defaults 0 0
/dev/sda5 /proc ext4 defaults 0 0
/dev/mapper/vg00-sys /sys xfs defaults 0 0
/dev/mapper/vg00-sys /sys sysfs defaults 0 0
UUID=023e51f5-... /sys xfs defaults 0 0
UUID=023e51f5-... /sys sysfs defaults 0 0
PARTUUID=abcd-1234 /run ext4 defaults 0 0
UUID=aabbccdd-... /run ext4 defaults 0 0
UUID=12345678-... /dev/shm xfs defaults 0 0
/some/dir /proc none bind 0 0
```

# correct entries
tmpfs /dev/shm tmpfs defaults 0 0
proc /proc proc defaults 0 0
sysfs /sys sysfs defaults 0 0
none /proc proc defaults 0 0
tmpfs /run tmpfs defaults 0 0
/dev/mapper/fedora-home /home ext4 defaults 0 0
UUID=aabbccdd-... /var xfs defaults 0 0
LABEL=optdata /opt ext4 defaults 0 0
/exports/home /home none bind 0 0
/data/shared /mnt/shared none bind 0 0

This is how it reports
```
Problematic entries:  
   - /dev/shm (xfs) 
   - /dev/shm (tmpfs) 
   - /dev/shm/ (xfs) 
   - /proc (ext4) 
   - /sys (xfs) 
   - /sys (sysfs) 
   - /sys (xfs) 
   - /sys (sysfs) 
   - /run (ext4) 
   - /run (ext4) 
   - /dev/shm (xfs) 
   - /proc (none)
```

prjagtap-rh: As discussed on slack I agree this would be a corner case with the worst possible fstab.
Will `sort` and list `fs_file`

prjagtap-rh: @pirat89 Fixed

matejmatuska:
The unit tests actions say that your acount is suspended, I think it might be caused by GH Actions issue, from https://www.githubstatus.com/incidents/gnftqj9htp0g:
> We have identified the cause of the authentication issues affecting GitHub Actions and are actively working on mitigation

edit: seems like there are other issue in the other actions as well

pirat89 (approved):
lgtm and works as expected.

PR #1542, opened 2026-05-28
Add CLAUDE.md and YAML frontmatter to skills for Claude compatibility

karolinku:
CLAUDE.md imports AGENTS.md so Claude Code uses the same instructions as other agents without duplicating content. YAML frontmatter (name, description) is added to each SKILL.md to satisfy Claude Code's skill metadata requirements.


pirat89 (approved):
lgtm and it seems to be working (just experienting with claude right now)

PR #1543, opened 2026-05-29
Fix rt kernel detection on Centos Stream

matejmatuska:
The kernel detection function didn't take into account major-only versions as used for CentOS Stream and therefore it misidentified rt kernels as "ordinary" as can be seen in the KernelInfo models "type" field which is populated with the output from the function:
```
Actor: scan_source_kernel
Phase: FactsCollection
Type: KernelInfo
Message_data:
{
    "pkg": {
        "arch": "x86_64",
        "epoch": "0",
        "module": null,
        "name": "kernel-rt-core",
        "packager": "builder@centos.org",
        "pgpsig": "RSA/SHA256, Thu 21 May 2026 02:15:40 AM CEST, Key ID 05b555b38483c65d",
        "release": "708.el9",
        "repository": "rt",
        "stream": null,
        "version": "5.14.0"
    },
    "type": "ordinary",
    "uname_r": "5.14.0-708.el9.x86_64+rt"
}
```
Which caused the target system to default to a non-rt kernel.

With this PR the kernel is correctly identified as "realtime":
```
Actor: scan_source_kernel
Phase: FactsCollection
Type: KernelInfo
Message_data:
{
    "pkg": {
        "arch": "x86_64",
        "epoch": "0",
        "module": null,
        "name": "kernel-rt-core",
        "packager": "builder@centos.org",
        "pgpsig": "RSA/SHA256, Thu 21 May 2026 02:15:40 AM CEST, Key ID 05b555b38483c65d",
        "release": "708.el9",
        "repository": "rt",
        "stream": null,
        "version": "5.14.0"
    },
    "type": "realtime",
    "uname_r": "5.14.0-708.el9.x86_64+rt"
}
```
and the target system correctly defaults to rt kernel:
```
[root@localhost ~]# uname -r
6.12.0-231.el10.x86_64+rt
```
```
# cat /etc/sysconfig/kernel
# UPDATEDEFAULT specifies if kernel-install should make
# new kernels the default
UPDATEDEFAULT=yes

# DEFAULTKERNEL specifies the default kernel package type
DEFAULTKERNEL=kernel-rt-core
```
---
I also removed the deprecated `is_rhel_realtime()` function because it caused circular imports between version and kernel libs. It was also deprecated long enough. 

Jira: RHEL-161560

### DIFF
--- a/repos/system_upgrade/common/libraries/config/tests/test_version.py
+++ b/repos/system_upgrade/common/libraries/config/tests/test_version.py
@@ -132,19 +132,6 @@ def test_is_supported_version(monkeypatch, result, src_ver, is_saphana):
     assert version.is_supported_version() == result
 
 
-@pytest.mark.parametrize('result,kernel,release_id', [
-    (False, '3.10.0-790.el7.x86_64', 'rhel'),
-    (False, '3.10.0-790.el7.x86_64', 'fedora'),
-    (False, '3.10.0-790.35.2.rt666.1133.el7.x86_64', 'fedora'),
-    (True, '3.10.0-790.35.2.rt666.1133.el7.x86_64', 'rhel'),
-])
-@suppress_deprecation(version.is_rhel_realtime)
-def test_is_rhel_realtime(monkeypatch, result, kernel, release_id):
-    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(src_ver='7.9', kernel=kernel,
-                                                                 release_id=release_id))
-    assert version.is_rhel_realtime() == result
-
-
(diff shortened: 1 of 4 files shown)


matejmatuska:
Also tested during a conversion CS 9 -> RHEL 10.2, the outputs are the same except the kernel NEVRA of course:
```
[root@localhost ~]# uname -r
6.12.0-211.7.1.el10_2.x86_64+rt
```

pirat89 (approved):
nice bug :D I missed that we haven't used `matches_version` here at all. this bug is nice example why we should prefer the use of libraries functions over own solution always

PR #1546, opened 2026-06-01
Add inhibitor for incorrect /var/run symlink configuration

prjagtap-rh:
If /var/run is not correctly configured as a symbolic link to /run, the in-place upgrade can lead to boot failures, service startup failures, or login issues. A new ChecksPhase actor detects this unsupported configuration and inhibits the upgrade.

JIRA: RHEL-76846

### DIFF
--- a/repos/system_upgrade/common/actors/checkvarrunsymlink/libraries/checkvarrunsymlink.py
+++ b/repos/system_upgrade/common/actors/checkvarrunsymlink/libraries/checkvarrunsymlink.py
@@ -1,0 +1,49 @@
+from leapp import reporting
+from leapp.libraries.stdlib import api
+from leapp.models import TrackedFilesInfoSource
+
+VAR_RUN_PATH = '/var/run'
+
+
+def _get_real_path(tracked_files):
+    for finfo in tracked_files.files:
+        if finfo.path == VAR_RUN_PATH:
+            return finfo.real_path
+    return None
+
+
+def process():
+    tracked_files = next(api.consume(TrackedFilesInfoSource), None)
+    if not tracked_files:
+        api.current_logger().warning(
+            'The TrackedFilesInfoSource message is missing. Skipping the check of /var/run symlink.'
+        )
+        return
+
+    real_path = _get_real_path(tracked_files)
+    if real_path is None or real_path == '/run':
+        return
+
+    reporting.create_report([
+        reporting.Title('/var/run is not correctly configured'),
+        reporting.Summary(
+            'In modern Linux systems that use systemd, /run is a tmpfs filesystem'
+            ' managed by systemd and /var/run must be a symbolic link pointing to'
+            ' ../run for compatibility. The current system does not have /var/run'
+            ' configured as expected. This can cause boot failures, service startup'
+            ' failures, or login issues both on the current system and after an'
+            ' in-place upgrade.'
+        ),
+        reporting.Remediation(
+            hint=(
+                'Back up current /var/run directory if needed and replace it by'
+                ' symbolic link pointing to "../run".'
+            ),
+            commands=[
+                ['bash', '-c', 'mv /var/run /tmp/var_run.bak && ln -s ../run /var/run'],
+            ]
+        ),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.FILESYSTEM, reporting.Groups.INHIBITOR]),
+        reporting.RelatedResource('directory', VAR_RUN_PATH),
+    ])
--- a/repos/system_upgrade/common/actors/checkvarrunsymlink/actor.py
+++ b/repos/system_upgrade/common/actors/checkvarrunsymlink/actor.py
@@ -1,0 +1,25 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import checkvarrunsymlink
+from leapp.models import TrackedFilesInfoSource
+from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckVarRunSymlink(Actor):
+    """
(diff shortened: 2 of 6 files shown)


repos/system_upgrade/common/actors/checkvarrunsymlink/libraries/checkvarrunsymlink.py
```diff
@@ -0,0 +1,32 @@
...
+            '/var/run is not correctly configured as a symbolic link to /run. '
+            'This unsupported configuration can cause failures after an in-place '
+            'upgrade, such as boot failures, service startup failures, or login issues.'
+        ),
+        reporting.Remediation(
+            hint=(
+                'Move the existing /var/run and replace it with '
+                'a symbolic link:\n'
+                '1. mv /var/run /tmp/var_run.bak\n'
+                '2. ln -s ../run /var/run'
```
pirat89: commands like this in hints are not ideal. and when it's clear how to do so, provide the command using `commands` as expected. In this case, I would expect hint like:


```suggestion
                'Back up current /var/run directory if needed and replace it by symbolic link pointing to "../run".'
```

repos/system_upgrade/common/actors/checkvarrunsymlink/libraries/checkvarrunsymlink.py
```diff
@@ -0,0 +1,32 @@
...
+        return
+
+    reporting.create_report([
+        reporting.Title(
+            '/var/run is not correctly configured'
+        ),
+        reporting.Summary(
+            '/var/run is not correctly configured as a symbolic link to /run. '
+            'This unsupported configuration can cause failures after an in-place '
+            'upgrade, such as boot failures, service startup failures, or login issues.'
```
pirat89: The summary is not explaining the problem well. Why it is /var/run symlink to `../run` expected instead on RHEL system nowadays? and it's not just beucase of in-place upgrades

pirat89 (changes requested):

repos/system_upgrade/common/actors/checkvarrunsymlink/libraries/checkvarrunsymlink.py
prjagtap-rh: Fixed

repos/system_upgrade/common/actors/checkvarrunsymlink/libraries/checkvarrunsymlink.py
prjagtap-rh: Fixed

repos/system_upgrade/common/actors/checkvarrunsymlink/libraries/checkvarrunsymlink.py
```diff
@@ -0,0 +1,37 @@
+import os
+
+from leapp import reporting
+
+VAR_RUN_PATH = '/var/run'
+
+
+def process():
+    if os.path.islink(VAR_RUN_PATH) and os.path.realpath(VAR_RUN_PATH) == '/run':
```
pirat89: @prjagtap-rh  I am sorry I missed that originally, but the actor is performed in the checksphase, which means that it cannot interact with the system  (like `os.path.islink` etc.). "Check" actors can observe the world only from the consumed messages. Possibly you will be intrested about TrackedFilesInfoSource message and the scansourcefiles actor. Most likely you will need to update it, but it seems as suitable way to do so instead of creating new model and actor just for one file

repos/system_upgrade/common/actors/checkvarrunsymlink/libraries/checkvarrunsymlink.py
```diff
@@ -0,0 +1,37 @@
...
+            ' in-place upgrade.'
+        ),
+        reporting.Remediation(
+            hint=(
+                'Back up current /var/run directory if needed and replace it by'
+                ' symbolic link pointing to "../run".'
+            ),
+            commands=[
+                ['mv', '/var/run', '/tmp/var_run.bak'],
+                ['ln', '-s', '../run', '/var/run'],
```
pirat89: at this moment all commands must be in the format that only one command is present (due to other webUI plugins; I guess that cockpit one). It needs to be wrapped like `bash -c your-command` - see other actors for examples.

pirat89 (changes requested):

repos/system_upgrade/common/actors/checkvarrunsymlink/libraries/checkvarrunsymlink.py
prjagtap-rh: fixed

repos/system_upgrade/common/actors/checkvarrunsymlink/libraries/checkvarrunsymlink.py
```diff
@@ -0,0 +1,49 @@
...
+    tracked_files = next(api.consume(TrackedFilesInfoSource), None)
+    if not tracked_files:
+        api.current_logger().warning(
+            'The TrackedFilesInfoSource message is missing. Skipping the check of /var/run symlink.'
+        )
+        return
+
+    real_path = _get_real_path(tracked_files)
+    if real_path is None or real_path == '/run':
+        return
```
pirat89, line 25: as we expect the real path to be always set based on the implementation, it would be better to raise a warning. but it's a nitpick. no blocking the merge

pirat89 (approved):
lgtm and works as expected.

repos/system_upgrade/common/actors/checkvarrunsymlink/libraries/checkvarrunsymlink.py
prjagtap-rh: @pirat89 fixed https://github.com/oamg/leapp-repository/pull/1552

PR #1547, opened 2026-06-02
mount_unit_generator: Remove pseudo FS mount units from upgrade initramfs

holser:
systemd-fstab-generator converts every `/etc/fstab` entry into a systemd mount unit. MountUnitGenerator then prefixes each unit's Where= path with `/sysroot` before bundling them into the upgrade initramfs. For virtual/pseudo
filesystem entries such as hugetlbfs (e.g. `/dev/hugepages2M`, `/dev/hugepages1G`) the corresponding directories may not exist under `/sysroot` (`/sysroot/dev/hugepages2M`, `/sysconfig/dev/hugepages1G`).

As the main reasons for having pseudo FS entries in fstab is security hardening, performance tuning, application-specific mounts. Since these pseudo filesystems have no purpose during the in-place upgrade phase, drop their mount units entirely in upgrade phase before any further processing.

Resolves: RHEL-180965

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/repos/system_upgrade/common/actors/initramfs/mount_units_generator/libraries/mount_unit_generator.py
+++ b/repos/system_upgrade/common/actors/initramfs/mount_units_generator/libraries/mount_unit_generator.py
@@ -9,6 +9,24 @@ from leapp.models import LiveModeConfig, StorageInfo, TargetUserSpaceInfo, Upgra
 
 BIND_MOUNT_SYSROOT_BOOT_UNIT = 'boot.mount'
 
+PSEUDO_FS_TYPES = frozenset([
+    'hugetlbfs',
+    'tmpfs',
+    'devtmpfs',
+    'devpts',
+    'sysfs',
+    'proc',
+    'cgroup',
+    'cgroup2',
+    'securityfs',
+    'selinuxfs',
+    'debugfs',
+    'mqueue',
+    'pstore',
+    'efivarfs',
+    'bpf',
+])
+
 
 def run_systemd_fstab_generator(output_directory):
     api.current_logger().debug(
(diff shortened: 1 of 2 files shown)


pirat89:
@holser I am not sure that this is the right approach as users could have set specific (nondefault) options for these MPs which then are not ignored AFAIK (and we should reflect that in such a case for systems following STIG). The /dev/* and MPs like this should be however mounted most likely just without the /sysroot prefix, because these are bind mounted later into the sysroot container.

MichalHe:
What is important here is that the reason for having a VF mount point in /etc/fstab can be one of the following:
1) security hardening (e.g. procfs)
2) performance (e.g. hugetblfs)
3) application-specific (debugfs, etc.)

Since we set enforcing=0, there is no point taking about security. There are no running user applications, so 2) and 3) are also out of the picture.

That being said, we should at least try to mount these filesystems as it does not introduce almost any complexity. If we fail, we continue. So the solution should be more akin to what we do with units marked with `nofail`: if the filesystem is a virtual one, we make sure the units does not end up in `.requires` and only in `.wants`.

holser:
> What is important here is that the reason for having a VF mount point in /etc/fstab can be one of the following:
> 
> 1. security hardening (e.g. procfs)
> 2. performance (e.g. hugetblfs)
> 3. application-specific (debugfs, etc.)
> 
> Since we set enforcing=0, there is no point taking about security. There are no running user applications, so 2) and 3) are also out of the picture.
> 
> That being said, we should at least try to mount these filesystems as it does not introduce almost any complexity. If we fail, we continue. So the solution should be more akin to what we do with units marked with `nofail`: if the filesystem is a virtual one, we make sure the units does not end up in `.requires` and only in `.wants`.

I have changed the behaviour to what's suggested. Thank you!

PR #1550, opened 2026-06-04
fix upgrades using kernel with 64k page size

PeterMocary:
### fix support for kernels with 64k page size
On aarch64, dedicated kernel packages for systems using 64k memory pages
(kernel-64k and kernel-rt-64k) exist. However, for ppc64le the standard
kernel packages are used despite having 64k page size. The upgrade now
detects the kernel page size and selects the correct kernel packages for
both the target OS and the internal upgrade environment.

The kernel-rt-64k RPM's kernel-uname-r provide uses a different suffix
(+rt_64k) than what uname -r returns (+rt-64k). A normalization
workaround is included to handle both variants until the RPM metadata
or uname -r output is fixed.

- Add ``page_size`` field to the ``KernelInfo`` model
- Add ``KernelPageSize`` enum and ``determine_kernel_page_size()`` using
  ``getconf PAGE_SIZE`` to the kernel shared library
- Update ``scansourcekernel`` to detect and produce the ``page_size``
- Recognize ``+rt-64k`` uname-r suffix as realtime kernel type
- Add ``get_kernel_pkg_names()`` to resolve kernel package names based
  on type (ordinary/rt), page size (4k/64k), and architecture (special
  64k kernel RPMs only exist on aarch64)
- Add ``KernelPackageInfoError`` exception and update
  ``get_target_kernel_pkg_info()`` to raise it instead of letting
  ``CalledProcessError`` propagate
- Update ``commonleappdracutmodules`` to dynamically select kernel
  packages instead of hardcoding kernel/kernel-core/kernel-modules
- Update ``scaninstalledtargetkernelversion`` to use
  ``get_target_kernel_pkg_names()`` with fallback from rt to ordinary
  preserving page size
- Update ``upgradeinitramfsgenerator`` to query the correct kernel
  variant inside the target userspace container
- Update ``generate-initram.sh`` fallback to query all kernel variants
- Deprecate the unused ``CurrentKernel`` model in favor of ``KernelInfo``

### rework kernel package identification to use /lib/modules paths

The ``kernel-uname-r`` rpm provide uses a different suffix format than
the actual ``uname -r`` output for RT kernels with 64k page size:
``uname -r`` returns ``+rt-64k``  while the provide has ``+rt_64k``.
This mismatch causes the existing provide-based lookup to fail to
identify the running kernel's package and resulted in a normalization
fix that mitigates it.

This patch switches a different file-based approach which is more
reliable:
- ``get_kernel_pkg_info_for_uname_r()`` now queries
  ``rpm --whatprovides /lib/modules/<uname_r>`` to directly find the
  owning packages, then selects the ``-core`` package from the results
- ``get_uname_r_provided_by_kernel_pkg()`` now extracts the version
  from the package's ``/lib/modules/<version>/`` file listing instead
  of parsing provides, and accepts an optional context parameter for
  in-container queries
- Deprecate ``KERNEL_UNAME_R_PROVIDES`` and remove ``_normalize_rt64k_suffix``
  as they are no longer needed 

## Summary on what changed in kernel library (to form the changelog easier)
- added `determine_kernel_page_size` function
- added `get_target_kernel_pkg_names` function
- added `KernelPackageInfoError` that inherits from `StopExecutionError`
- deprecated `KERNEL_UNAME_R_PROVIDES` constant
- ` get_uname_r_provided_by_kernel_pkg`
  - uname_r is now retrieved from packages `/lib/modules/<uname_r>` instead of the RPM provides 
  - now returns empty string instead of propagating `CalledProcessError`
  - now takes an optional argument context
- `get_kernel_pkg_info`
  - now raises `KernelPackageInfoError` instead of propagating `CalledProcessError`
- `get_kernel_pkg_info_for_uname_r`
  - now raises `KernelPackageInfoError` instead of `StopActorExecutionError` and when rpm query fails
  - now matches the kernel pkg to uname_r base on the `/lib/modules/<uname_r>` instead of using RPM whatprovides

Jira: RHEL-111860

### DIFF
--- a/repos/system_upgrade/common/actors/commonleappdracutmodules/libraries/modscan.py
+++ b/repos/system_upgrade/common/actors/commonleappdracutmodules/libraries/modscan.py
@@ -1,12 +1,15 @@
 import os
 import re
 
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.common import kernel as kernel_lib
 from leapp.libraries.common.config import architecture, version
 from leapp.libraries.stdlib import api
 from leapp.utils.deprecation import suppress_deprecation
 
 from leapp.models import (  # isort:skip
     CopyFile,
+    KernelInfo,
     LiveModeConfig,
     RequiredUpgradeInitramPackages,  # deprecated
     UpgradeDracutModule,  # deprecated
@@ -28,9 +31,6 @@ _REQUIRED_PACKAGES = [
     'hostname',
     'iscsi-initiator-utils',
     'kbd',
-    'kernel',
-    'kernel-core',
-    'kernel-modules',
     'keyutils',
     'kmod',
     'lldpad',
@@ -95,6 +95,18 @@ def _create_initram_packages():
         required_pkgs.append('NetworkManager')
     if version.get_target_major_version() == '9':
         required_pkgs += ['policycoreutils', 'rng-tools']
+
+    kernel_info = next(api.consume(KernelInfo), None)
+    if not kernel_info:
+        raise StopActorExecutionError('Could not retrieve KernelInfo message.')
+    page_size = kernel_info.page_size
+    kernel_packages = kernel_lib.get_target_kernel_pkg_names(
+        kernel_lib.KernelType.ORDINARY,
+        page_size,
+        api.current_actor().configuration.architecture
+    )
+    required_pkgs.extend(kernel_packages)
+
     return (
         RequiredUpgradeInitramPackages(packages=required_pkgs),
         TargetUserSpaceUpgradeTasks(install_rpms=required_pkgs)
--- a/repos/system_upgrade/common/actors/commonleappdracutmodules/tests/test_modscan_commonleappdracutmodules.py
+++ b/repos/system_upgrade/common/actors/commonleappdracutmodules/tests/test_modscan_commonleappdracutmodules.py
@@ -4,6 +4,7 @@ from collections import namedtuple
 
 import pytest
 
+from leapp.exceptions import StopActorExecutionError
 from leapp.libraries.actor import modscan
 from leapp.libraries.common.config import architecture
 from leapp.libraries.common.testutils import CurrentActorMocked
@@ -11,10 +12,36 @@ from leapp.libraries.stdlib import api
 from leapp.utils.deprecation import suppress_deprecation
 
 from leapp.models import (  # isort:skip
+    KernelInfo,
+    RPM,
     RequiredUpgradeInitramPackages,  # deprecated
     UpgradeDracutModule,  # deprecated
     TargetUserSpaceUpgradeTasks,
     UpgradeInitramfsTasks
+)
+
+_KERNEL_INFO_4K = KernelInfo(
+    pkg=RPM(name='kernel-core', arch='x86_64', version='5.14.0', release='100.el9',
+            epoch='0', packager='', pgpsig='SIG'),
+    type='ordinary',
+    uname_r='5.14.0-100.el9.x86_64',
+    page_size='4k',
+)
+
+_KERNEL_INFO_64K = KernelInfo(
+    pkg=RPM(name='kernel-64k-core', arch='aarch64', version='5.14.0', release='100.el9',
+            epoch='0', packager='', pgpsig='SIG'),
+    type='ordinary',
+    uname_r='5.14.0-100.el9.aarch64+64k',
+    page_size='64k',
+)
+
+_KERNEL_INFO_64K_PPC = KernelInfo(
+    pkg=RPM(name='kernel-core', arch='ppc64le', version='5.14.0', release='100.el9',
+            epoch='0', packager='', pgpsig='SIG'),
+    type='ordinary',
+    uname_r='5.14.0-100.el9.ppc64le',
+    page_size='64k',
 )
 
 
@@ -74,8 +101,7 @@ def test_created_modules(monkeypatch):
     ('8.4', '9.0', architecture.ARCH_S390X),
 ))
 def test_required_packages(monkeypatch, src_ver, dst_ver, arch):
-    # the default set of required rpms should not contain biosdevname
-    for pkg in ['biosdevname', 'policycoreutils', 'rng-tools']:
+    for pkg in ['biosdevname', 'policycoreutils', 'rng-tools', 'kernel', 'kernel-core', 'kernel-modules']:
         assert pkg not in modscan._REQUIRED_PACKAGES
 
     required_packages = modscan._REQUIRED_PACKAGES[:]
@@ -83,17 +109,59 @@ def test_required_packages(monkeypatch, src_ver, dst_ver, arch):
         required_packages += ['policycoreutils', 'rng-tools']
     if arch == architecture.ARCH_X86_64:
         required_packages += ['biosdevname']
+    required_packages += ['kernel', 'kernel-core', 'kernel-modules']
 
-    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(src_ver=src_ver, dst_ver=dst_ver, arch=arch))
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(
+        src_ver=src_ver, dst_ver=dst_ver, arch=arch, msgs=[_KERNEL_INFO_4K]
+    ))
     old_initram, new_initram = modscan._create_initram_packages()
     assert (set(required_packages)
             == set(old_initram.packages)
             == set(new_initram.install_rpms))
 
 
+def test_required_packages_no_kernel_info(monkeypatch):
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(src_ver='9.8', dst_ver='10.2'))
+    with pytest.raises(StopActorExecutionError):
+        modscan._create_initram_packages()
+
+
+@pytest.mark.parametrize('kernel_info,arch,expected_kernel_pkgs,unexpected_kernel_pkgs', (
+    (_KERNEL_INFO_4K, architecture.ARCH_ARM64,
+     ['kernel', 'kernel-core', 'kernel-modules'],
+     ['kernel-64k', 'kernel-64k-core', 'kernel-64k-modules']),
+    (_KERNEL_INFO_64K, architecture.ARCH_ARM64,
+     ['kernel-64k', 'kernel-64k-core', 'kernel-64k-modules'],
+     ['kernel', 'kernel-core', 'kernel-modules']),
+    # ppc64le: 64k page size but standard kernel packages (no kernel-64k RPMs)
+    (_KERNEL_INFO_64K_PPC, architecture.ARCH_PPC64LE,
+     ['kernel', 'kernel-core', 'kernel-modules'],
+     ['kernel-64k', 'kernel-64k-core', 'kernel-64k-modules']),
+    # x86_64: 4k page size, standard kernel packages
+    (_KERNEL_INFO_4K, architecture.ARCH_X86_64,
+     ['kernel', 'kernel-core', 'kernel-modules'],
+     ['kernel-64k', 'kernel-64k-core', 'kernel-64k-modules']),
+    # s390x: 4k page size, standard kernel packages
+    (_KERNEL_INFO_4K, architecture.ARCH_S390X,
+     ['kernel', 'kernel-core', 'kernel-modules'],
+     ['kernel-64k', 'kernel-64k-core', 'kernel-64k-modules']),
+))
+def test_required_packages_page_size(monkeypatch, kernel_info, arch, expected_kernel_pkgs, unexpected_kernel_pkgs):
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(
+        src_ver='9.6', dst_ver='10.0', arch=arch, msgs=[kernel_info]
+    ))
+    old_initram, new_initram = modscan._create_initram_packages()
+    for pkg in expected_kernel_pkgs:
+        assert pkg in new_initram.install_rpms
+        assert pkg in old_initram.packages
+    for pkg in unexpected_kernel_pkgs:
+        assert pkg not in new_initram.install_rpms
+        assert pkg not in old_initram.packages
+
+
 @suppress_deprecation(UpgradeDracutModule)
 def test_process_produces_modules(monkeypatch):
-    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked())
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(msgs=[_KERNEL_INFO_4K]))
     messages = []
     monkeypatch.setattr(api, 'produce', lambda *x: messages.extend(x))
     monkeypatch.setattr(api, 'get_actor_folder_path', _files_get_folder_path)
--- a/repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/files/generate-initram.sh
+++ b/repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/files/generate-initram.sh
@@ -7,7 +7,10 @@ stage() {
 }
 
 get_kernel_version() {
-    rpm -qa kernel --qf '%{VERSION}-%{RELEASE}.%{ARCH}\n' | sort --version-sort | tail --lines=1
+    # The actor should always provide the version via LEAPP_KERNEL_VERSION and ensure
+    # only one kernel is installed in the container. This is simple fallback in case
+    # the version is not provided.
+    rpm -q --whatprovides kernel --qf '%{VERSION}-%{RELEASE}.%{ARCH}\n' | sort --version-sort | tail --lines=1
 }
 
 dracut_install_modules()
--- a/repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
+++ b/repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
@@ -4,6 +4,7 @@ import shutil
 from collections import namedtuple
 
 from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.common import kernel as kernel_lib
 from leapp.libraries.common import mounting
 from leapp.libraries.common.config.version import get_target_major_version
 from leapp.libraries.common.dnflibs import dnfplugin
@@ -12,6 +13,7 @@ from leapp.models import RequiredUpgradeInitramPackages  # deprecated
 from leapp.models import UpgradeDracutModule  # deprecated
 from leapp.models import (
     BootContent,
+    KernelInfo,
     LiveModeConfig,
     LVMConfig,
     RAIDInfo,
@@ -33,51 +35,40 @@ def _get_target_kernel_version(context):
     """
     Get the version of the most recent kernel version within the container.
     """
+    kernel_info = next(api.consume(KernelInfo), None)
+    if not kernel_info:
+        raise StopActorExecutionError('Could not retrieve KernelInfo message.')
+    page_size = kernel_info.page_size
+    arch = api.current_actor().configuration.architecture
+    kernel_pkg_names = kernel_lib.get_target_kernel_pkg_names(kernel_lib.KernelType.ORDINARY, page_size, arch)
+    kernel_core_pkg = kernel_pkg_names.core
 
-    kernel_version = None
     try:
-        # NOTE: Currently we install/use always kernel-core in the upgrade
-        # initramfs. We do not use currently any different kernel package
-        # in the container. Note this could change in future e.g. on aarch64
-        # for IPU 9 -> 10.
-        # TODO(pstodulk): Investigate situation on ARM systems. OAMG-11433
-        results = context.call(['rpm', '-qa', 'kernel-core'], split=True)['stdout']
+        results = context.call(['rpm', '-qa', kernel_core_pkg], split=True)['stdout']
     except CalledProcessError:
         raise StopActorExecutionError(
             'Cannot get version of the installed kernel inside container.',
             details={'Problem': 'Could not query the currently installed kernel inside container using rpm.'})
 
+    if not results:
+        raise StopActorExecutionError(
+            'Cannot get version of the installed kernel inside container.',
+            details={'Problem': 'An rpm query for the available kernels did not produce any results.'})
+
     if len(results) > 1:
-        # this is should not happen. It's hypothetic situation, which alone it's
-        # already error. So skipping more sophisticated implementation.
-        # The container is always created during the upgrade and as that we expect
-        # always one-and-only kernel installed.
         raise StopActorExecutionError(
             'Cannot get version of the installed kernel inside container.',
             details={'Problem': 'Detected unexpectedly multiple kernels inside target userspace container.'}
         )
 
-    # kernel version == version-release from package
-    kernel_version = '-'.join(results[0].rsplit("-", 2)[-2:])
-    api.current_logger().debug('Detected kernel version inside container: {}.'.format(kernel_version))
-
+    kernel_version = kernel_lib.get_uname_r_provided_by_kernel_pkg(results[0], context=context)
     if not kernel_version:
         raise StopActorExecutionError(
             'Cannot get version of the installed kernel inside container.',
-            details={'Problem': 'An rpm query for the available kernels did not produce any results.'})
+            details={'Problem': 'Failed to determine uname-r provided by {}.'.format(results[0])})
 
+    api.current_logger().debug('Detected kernel version inside container: {}.'.format(kernel_version))
     return kernel_version
-
-
-def _get_target_kernel_modules_dir(context):
-    """
-    Return the path where the custom kernel modules should be copied.
-    """
-
-    kernel_version = _get_target_kernel_version(context)
-    modules_dir = os.path.join('/', 'lib', 'modules', kernel_version, 'extra', 'leapp')
-
-    return modules_dir
 
 
 def _reinstall_leapp_repository_hint():
@@ -153,7 +144,7 @@ def copy_dracut_modules(context, modules):
     _copy_modules(context, modules, DRACUT_DIR, 'dracut')
 
 
-def copy_kernel_modules(context, modules):
+def copy_kernel_modules(context, modules, kernel_version):
     """
     Copy kernel modules into the target userspace.
 
@@ -161,7 +152,7 @@ def copy_kernel_modules(context, modules):
     StopActorExecutionError exception is raised.
     """
 
-    dst_dir = _get_target_kernel_modules_dir(context)
+    dst_dir = os.path.join('/', 'lib', 'modules', kernel_version, 'extra', 'leapp')
     _copy_modules(context, modules, dst_dir, 'kernel')
 
 
@@ -396,8 +387,10 @@ def generate_initram_disk(context):
     # Issue #645
     initramfs_includes = collect_initramfs_includes()
 
+    kernel_version = _get_target_kernel_version(context)
+
     copy_dracut_modules(context, initramfs_includes.dracut_modules)
-    copy_kernel_modules(context, initramfs_includes.kernel_modules)
+    copy_kernel_modules(context, initramfs_includes.kernel_modules, kernel_version)
 
     def fmt_module_list(module_list):
         return ','.join(mod.name for mod in module_list)
@@ -422,7 +415,7 @@ def generate_initram_disk(context):
 
     env_variables = ' '.join(env_variables)
     env_variables = env_variables.format(
-        kernel_version=_get_target_kernel_version(context),
+        kernel_version=kernel_version,
         dracut_modules=fmt_module_list(initramfs_includes.dracut_modules),
         kernel_modules=fmt_module_list(initramfs_includes.kernel_modules),
         arch=api.current_actor().configuration.architecture,
@@ -481,7 +474,7 @@ def _generate_livemode_initramfs(context, userspace_initramfs_dest, target_kerne
     initramfs_includes = collect_initramfs_includes()
 
     copy_dracut_modules(context, initramfs_includes.dracut_modules)
-    copy_kernel_modules(context, initramfs_includes.kernel_modules)
+    copy_kernel_modules(context, initramfs_includes.kernel_modules, target_kernel_ver)
 
     md_cmdline_conf = write_md_uuid_cmdline_conf(context)
     if md_cmdline_conf:
--- a/repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/tests/unit_test_upgradeinitramfsgenerator.py
+++ b/repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/tests/unit_test_upgradeinitramfsgenerator.py
@@ -600,8 +600,8 @@ def test_get_target_kernel_version_page_size(monkeypatch, page_size, arch, expec
         queried_cmds.append(cmd)
         if cmd[:2] == ['rpm', '-qa']:
             return {'stdout': [target_nevra]}
-        if cmd[:3] == ['rpm', '-q', '--provides']:
-            return {'stdout': ['kernel-uname-r = {}'.format(expected_uname_r)]}
+        if cmd[:3] == ['rpm', '-q', '-l']:
+            return {'stdout': ['/lib/modules/{}/vmlinuz'.format(expected_uname_r)]}
         return {'stdout': []}
 
     context.call = mock_call
@@ -609,7 +609,7 @@ def test_get_target_kernel_version_page_size(monkeypatch, page_size, arch, expec
     version = upgradeinitramfsgenerator._get_target_kernel_version(context)
     assert version == expected_uname_r
     assert queried_cmds[0] == ['rpm', '-qa', expected_pkg]
-    assert queried_cmds[1] == ['rpm', '-q', '--provides', target_nevra]
+    assert queried_cmds[1] == ['rpm', '-q', '-l', target_nevra]
 
 
 def test_get_target_kernel_version_no_kernel_info(monkeypatch):
@@ -644,8 +644,8 @@ def test_get_target_kernel_version_empty_uname_r(monkeypatch):
     def mock_call(cmd, *args, **kwargs):
         if cmd[:2] == ['rpm', '-qa']:
             return {'stdout': ['kernel-core-5.14.0-100.el9.x86_64']}
-        if cmd[:3] == ['rpm', '-q', '--provides']:
-            return {'stdout': ['some-other-provide = foo']}
+        if cmd[:3] == ['rpm', '-q', '-l']:
+            return {'stdout': ['/boot/vmlinuz-5.14.0-100.el9.x86_64']}
         return {'stdout': []}
 
     context.call = mock_call
--- a/repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/tests/unit_test_upgradeinitramfsgenerator.py
+++ b/repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/tests/unit_test_upgradeinitramfsgenerator.py
@@ -11,6 +11,8 @@ from leapp.utils.deprecation import suppress_deprecation
 
 from leapp.models import (  # isort:skip
     FIPSInfo,
+    KernelInfo,
+    RPM,
     RequiredUpgradeInitramPackages,  # deprecated
     UpgradeDracutModule,  # deprecated
     BootContent,
@@ -356,7 +358,7 @@ def test_generate_initram_disk(monkeypatch, input_msgs, dracut_modules, kernel_m
     curr_actor = CurrentActorMocked(msgs=input_msgs, arch=architecture.ARCH_X86_64)
     monkeypatch.setattr(upgradeinitramfsgenerator.api, 'current_actor', curr_actor)
     monkeypatch.setattr(upgradeinitramfsgenerator, 'copy_dracut_modules', MockedCopyArgs())
-    monkeypatch.setattr(upgradeinitramfsgenerator, '_get_target_kernel_version', lambda _: '')
+    monkeypatch.setattr(upgradeinitramfsgenerator, '_get_target_kernel_version', lambda _: '5.14.0-100.el9.x86_64')
     monkeypatch.setattr(upgradeinitramfsgenerator, 'copy_kernel_modules', MockedCopyArgs())
     monkeypatch.setattr(upgradeinitramfsgenerator, 'copy_boot_files', lambda dummy: None)
     monkeypatch.setattr(upgradeinitramfsgenerator, '_get_fspace', MockedGetFspace(2*2**30))
@@ -381,6 +383,8 @@ def test_generate_initram_disk(monkeypatch, input_msgs, dracut_modules, kernel_m
         assert module in _kernel_modules
         detected_kernel_modules.add(module)
     assert detected_kernel_modules == _kernel_modules
+
+    assert upgradeinitramfsgenerator.copy_kernel_modules.args[2] == '5.14.0-100.el9.x86_64'
 
     # TODO(pstodulk): this test is not created properly, as context.call check
     # is skipped completely. Testing will more convenient with fixed #376
@@ -423,26 +427,28 @@ def test_copy_modules_fail(monkeypatch, kind):
     context.copytree_to = copytree_to_error
     monkeypatch.setattr(os.path, 'exists', mock_context_path_exists)
     monkeypatch.setattr(upgradeinitramfsgenerator.api, 'current_logger', MockedLogger())
-    monkeypatch.setattr(upgradeinitramfsgenerator, '_get_target_kernel_modules_dir', lambda _: '/kernel_modules')
 
     module_class = None
     copy_fn = None
-    dst_path = None
+    dst_dir = None
     if kind == 'dracut':
         module_class = DracutModule
         copy_fn = upgradeinitramfsgenerator.copy_dracut_modules
-        dst_path = 'dracut'
+        dst_dir = 'dracut'
     elif kind == 'kernel':
         module_class = KernelModule
         copy_fn = upgradeinitramfsgenerator.copy_kernel_modules
-        dst_path = 'kernel_modules'
+        dst_dir = 'lib/modules/dummy/extra/leapp'
 
     modules = [module_class(name='foo', module_path='/path/foo')]
     with pytest.raises(StopActorExecutionError) as err:
-        copy_fn(context, modules)
+        if kind == 'dracut':
+            copy_fn(context, modules)
+        else:
+            copy_fn(context, modules, 'dummy')
     assert err.value.message.startswith('Failed to install {kind} modules'.format(kind=kind))
-    expected_err_log = 'Failed to copy {kind} module "foo" from "/path/foo" to "/base/dir/{dst_path}"'.format(
-            kind=kind, dst_path=dst_path)
+    expected_err_log = 'Failed to copy {kind} module "foo" from "/path/foo" to "/base/dir/{dst_dir}"'.format(
+            kind=kind, dst_dir=dst_dir)
     assert expected_err_log in upgradeinitramfsgenerator.api.current_logger.errmsg
 
 
@@ -456,7 +462,6 @@ def test_copy_modules_duplicate_skip(monkeypatch, kind):
 
     monkeypatch.setattr(os.path, 'exists', mock_context_path_exists)
     monkeypatch.setattr(upgradeinitramfsgenerator.api, 'current_logger', MockedLogger())
-    monkeypatch.setattr(upgradeinitramfsgenerator, '_get_target_kernel_modules_dir', lambda _: '/kernel_modules')
 
     module_class = None
     copy_fn = None
@@ -470,7 +475,10 @@ def test_copy_modules_duplicate_skip(monkeypatch, kind):
     module = module_class(name='foo', module_path='/path/foo')
     modules = [module, module]
 
-    copy_fn(context, modules)
+    if kind == 'dracut':
+        copy_fn(context, modules)
+    else:
+        copy_fn(context, modules, 'dummy')
 
     debugmsg = 'The foo {kind} module has been already installed. Skipping.'.format(kind=kind)
     assert context.content
@@ -558,3 +566,89 @@ def test_prepare_boot_files_for_livemode(monkeypatch):
     assert boot_content.kernel_path == '/boot/vmlinuz-upgrade.x86_64'
     assert boot_content.initram_path == '/boot/initramfs-upgrade.x86_64.img'
     assert boot_content.kernel_hmac_path == '/boot/.vmlinuz-upgrade.x86_64.hmac'
+
+
+def _make_kernel_info(page_size='4k', arch='aarch64'):
+    return KernelInfo(
+        pkg=RPM(name='kernel-core', arch=arch, version='5.14.0', release='100.el9',
+                epoch='0', packager='', pgpsig='SIG'),
+        type='ordinary',
+        uname_r='5.14.0-100.el9.{}'.format(arch),
+        page_size=page_size,
+    )
+
+
+@pytest.mark.parametrize('page_size,arch,expected_pkg,expected_uname_r', [
+    ('4k', 'aarch64', 'kernel-core', '5.14.0-100.el9.aarch64'),
+    ('4k', 'x86_64', 'kernel-core', '5.14.0-100.el9.x86_64'),
+    ('4k', 's390x', 'kernel-core', '5.14.0-100.el9.s390x'),
+    # aarch64: 64k page size and special kernel packages
+    ('64k', 'aarch64', 'kernel-64k-core', '5.14.0-100.el9.aarch64+64k'),
+    # ppc64le: 64k page size but standard kernel packages
+    ('64k', 'ppc64le', 'kernel-core', '5.14.0-100.el9.ppc64le'),
+])
+def test_get_target_kernel_version_page_size(monkeypatch, page_size, arch, expected_pkg, expected_uname_r):
+    kernel_info = _make_kernel_info(page_size, arch)
+    monkeypatch.setattr(upgradeinitramfsgenerator.api, 'current_actor',
+                        CurrentActorMocked(arch=arch, msgs=[kernel_info]))
+
+    context = MockedContext()
+    target_nevra = '{}-5.14.0-100.el9.{}'.format(expected_pkg, arch)
+    queried_cmds = []
+
+    def mock_call(cmd, *args, **kwargs):
+        queried_cmds.append(cmd)
+        if cmd[:2] == ['rpm', '-qa']:
+            return {'stdout': [target_nevra]}
+        if cmd[:3] == ['rpm', '-q', '--provides']:
+            return {'stdout': ['kernel-uname-r = {}'.format(expected_uname_r)]}
+        return {'stdout': []}
+
+    context.call = mock_call
+
+    version = upgradeinitramfsgenerator._get_target_kernel_version(context)
+    assert version == expected_uname_r
+    assert queried_cmds[0] == ['rpm', '-qa', expected_pkg]
+    assert queried_cmds[1] == ['rpm', '-q', '--provides', target_nevra]
+
+
+def test_get_target_kernel_version_no_kernel_info(monkeypatch):
+    monkeypatch.setattr(upgradeinitramfsgenerator.api, 'current_actor',
+                        CurrentActorMocked(msgs=[]))
+
+    context = MockedContext()
+
+    with pytest.raises(StopActorExecutionError, match='Could not retrieve KernelInfo message'):
+        upgradeinitramfsgenerator._get_target_kernel_version(context)
+
+
+def test_get_target_kernel_version_empty_results(monkeypatch):
+    kernel_info = _make_kernel_info(page_size='4k', arch='x86_64')
+    monkeypatch.setattr(upgradeinitramfsgenerator.api, 'current_actor',
+                        CurrentActorMocked(arch='x86_64', msgs=[kernel_info]))
+
+    context = MockedContext()
+    context.call = lambda cmd, *args, **kwargs: {'stdout': []}
+
+    with pytest.raises(StopActorExecutionError, match='Cannot get version of the installed kernel'):
+        upgradeinitramfsgenerator._get_target_kernel_version(context)
+
+
+def test_get_target_kernel_version_empty_uname_r(monkeypatch):
+    kernel_info = _make_kernel_info(page_size='4k', arch='x86_64')
+    monkeypatch.setattr(upgradeinitramfsgenerator.api, 'current_actor',
+                        CurrentActorMocked(arch='x86_64', msgs=[kernel_info]))
+
+    context = MockedContext()
+
+    def mock_call(cmd, *args, **kwargs):
+        if cmd[:2] == ['rpm', '-qa']:
+            return {'stdout': ['kernel-core-5.14.0-100.el9.x86_64']}
+        if cmd[:3] == ['rpm', '-q', '--provides']:
+            return {'stdout': ['some-other-provide = foo']}
+        return {'stdout': []}
+
+    context.call = mock_call
+
+    with pytest.raises(StopActorExecutionError, match='Cannot get version of the installed kernel'):
+        upgradeinitramfsgenerator._get_target_kernel_version(context)
--- a/repos/system_upgrade/common/actors/scaninstalledtargetkernelversion/actor.py
+++ b/repos/system_upgrade/common/actors/scaninstalledtargetkernelversion/actor.py
@@ -8,10 +8,19 @@ class ScanInstalledTargetKernelVersion(Actor):
     """
     Scan for the version of the newly installed kernel
 
-    This actor will query rpm for all kernel-core packages and reports the
-    first matching target system kernel RPM. In case the RHEL Real Time has
-    been detected on the original system, the kernel-rt-core rpm is searched.
-    If the rpm is missing, fallback for standard kernel RPM.
+    Based on the source kernel type and page size, this actor determines the
+    appropriate target kernel package (e.g. kernel-core, kernel-64k-core,
+    kernel-rt-core) and produces InstalledTargetKernelInfo containing the installed target kernel RPM. If the
+    expected variant is missing, it falls back to the ordinary kernel for
+    the same page size. Note that a fallback from a realtime to an ordinary
+    kernel may occur when the target realtime kernel package is not
+    available. This can happen if the user does not have a repository
+    providing the realtime kernel enabled, and there is currently no way
+    to verify package availability beforehand (the dnf plugin does not
+    propagate this information back to the actor). In such cases, the user
+    must enable the appropriate repository and install the realtime kernel
+    after the upgrade manually, which is preferred to ending up with an
+    unbootable system.
     """
 
     name = 'scan_installed_target_kernel_version'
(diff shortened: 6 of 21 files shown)


PeterMocary:
I discovered a problem with kernel RPM's metadata when compared to ``uname -r`` output. The suffix for ``uname -r`` does not match RPM's metadata (`kernel-uname-r`): `+rt-64k` vs `+rt_64k`.

```bash
[root@host ~]# uname -r
5.14.0-687.13.1.el9_8.aarch64+rt-64k
[root@host ~]# rpm -q --provides kernel-rt-64k-core | grep kernel-uname-r
kernel-uname-r = 5.14.0-687.13.1.el9_8.aarch64+rt_64k
```

FIX: I'm adding a simple normalization function that will make sure that the RPM will be matched to the `uname -r` output even if either change in the future. When the change happens, we should remove this normalization.

We should probably contact some kernel team and raise this issue. What do you think? @pirat89 @matejmatuska

repos/system_upgrade/common/libraries/kernel.py
```diff
@@ -103,12 +149,14 @@ def get_kernel_pkg_info_for_uname_r(uname_r):
 
     kernel_pkg_nevras = set(kernel_pkg_nevras)
 
+    uname_r = _normalize_rt64k_suffix(uname_r)
     for kernel_pkg_nevra in kernel_pkg_nevras:
         provided_uname = get_uname_r_provided_by_kernel_pkg(kernel_pkg_nevra)  # We know all packages provide a uname
         if not provided_uname:
             api.current_logger().warning('Failed to obtain uname-r provided by %s', kernel_pkg_nevra)
+        provided_uname = _normalize_rt64k_suffix(provided_uname)
```
matejmatuska: Missing continue?
```suggestion
        if not provided_uname:
            api.current_logger().warning('Failed to obtain uname-r provided by %s', kernel_pkg_nevra)
            continue
        provided_uname = _normalize_rt64k_suffix(provided_uname)
```

repos/system_upgrade/common/actors/scaninstalledtargetkernelversion/libraries/scankernel.py
```diff
@@ -73,14 +100,14 @@ def process():
...
-    target_ver = get_target_major_version()
-    target_kernel_pkg_name = get_kernel_pkg_name(target_ver, src_kernel_info.type)
+    target_kernel_pkg_name = get_kernel_pkg_name(src_kernel_info.type, src_kernel_info.page_size)
     target_kernel_nevra = get_target_kernel_package_nevra(target_kernel_pkg_name)
 
     if src_kernel_info.type != kernel_lib.KernelType.ORDINARY and not target_kernel_nevra:
-        api.current_logger().warning('The kernel-rt-core rpm from the target RHEL has not been detected. Switching '
-                                     'to non-preemptive kernel.')
-        target_kernel_pkg_name = get_kernel_pkg_name(target_ver, kernel_lib.KernelType.ORDINARY)
+        # Fallback to ordinary kernel if target kernel nevra could not be determined
```
matejmatuska, line 93: Currently this falls back from realtime to ordinary, should it also try to fallback from 64k to 4k? I am not sure if that's compatible or good idea, but we should consider it I think.

matejmatuska (changes requested):

repos/system_upgrade/common/actors/scaninstalledtargetkernelversion/libraries/scankernel.py
PeterMocary: I was thinking the same but in the end I decided based on this:

> It is not recommended to move between 4k and 64k page size kernels after the initial boot without reinstallation of the OS.

[Source](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/automatically_installing_rhel/installing-kernel-64k-on-arm-using-the-command-line_rhel-installer)

Not sure how relevant that is though, since it is literally in the tutorial article on how to do it.

repos/system_upgrade/common/libraries/kernel.py
PeterMocary: I missed that, it currently does not change the behavior, but it should be there regardless. I'll fix it.

repos/system_upgrade/common/actors/scaninstalledtargetkernelversion/libraries/scankernel.py
pirat89: > should it also try to fallback from 64k to 4k?

no, it should not. based on the discussion I had with the kernel team in the past. Such behaviour would be considered as unsupported from the POV.

the fallback kernel-rt -> kernel is there because many times users do not delivered repository with the RT content and we do not have now a possibility to check whether such package is there or not (missing data propagated from the dnf plugin back to actor). In such a case, they need to enable repo and install the content after the upgrade manually. (which is bettern than ending with unbootable system).

In case of 4k & 64k kernels it's different - both kernels are in the very same repo. So no fallback is needed.

pirat89:
> I discovered a problem with kernel RPM's metadata when compared to `uname -r` output. The suffix for `uname -r` does not match RPM's metadata (`kernel-uname-r`): `+rt-64k` vs `+rt_64k`.
> 
> ```shell
> [root@host ~]# uname -r
> 5.14.0-687.13.1.el9_8.aarch64+rt-64k
> [root@host ~]# rpm -q --provides kernel-rt-64k-core | grep kernel-uname-r
> kernel-uname-r = 5.14.0-687.13.1.el9_8.aarch64+rt_64k
> ```
> 
> FIX: I'm adding a simple normalization function that will make sure that the RPM will be matched to the `uname -r` output even if either change in the future. When the change happens, we should remove this normalization.
> 
> We should probably contact some kernel team and raise this issue. What do you think? @pirat89 @matejmatuska

Open a ticket in JIRA on kernel, explaining how it affects us and that we require unified way how to identify which rpm represents the booted kernel and what is the type (flavor) of the kernel - in the meaning of 4k/64k ordinary kernel on arm, "ordinary" kernel, Real Time kernel,  ... the real time kernel is not on arm as well - not sure whether in 64k and 4k variant in the same time. And it's possible that other kernel families will be comming (there is already the nvidia/voyager kernel).

Based on a  discussion I had recently, it's most likely due to RPM, but let's rather open a ticket so we get a clear guidance on that (or fix in the package).

repos/system_upgrade/common/models/installedtargetkernelversion.py
```diff
@@ -30,6 +28,9 @@ class KernelInfo(Model):
     # @FixMe(mhecko): I want to use kernel_lib.KernelType here, but I cannot import any library code (yet).
     # #               Figure out how to do it.
 
+    page_size = fields.StringEnum(['4k', '64k'], default='4k')
+    """Memory page size of the kernel. The 64k page size is available only on aarch64 for both kernel types."""
```
pirat89: @PeterMocary  note that ppc64le kernel has most likely also 64k pagesize, so this statement can be confusing. take a power machine and run `getconf PAGESIZE`

pirat89 (commented):
@PeterMocary @matejmatuska  checking the solution briefly, I am not sure whether it's complete. Does the solution includes specific kernel installed also for the upgrade procedure? So, when
* 4k kernel is installed on the source system, the upgrade must be peformed with the 4k kernel
* similarly for the 64k kernel

So if the system is booted with kernel-64k, it must not be upgraded with the kernel (default).

PeterMocary:
Opened a separate PR for the tests fix and removed the commit from here.

PeterMocary:
Updated the solution. It now handles detection of different page size correctly regardless of architecture (e.g. ppc64le) . The kernel used during the upgrade procedure is also adjusted based on the page size as @pirat89 suggested above. Apart from that the alternative `uname -r` to RPM matching is implemented in separate commit - it replaces the normalization hotfix.

## Testing
This is few runs from my testing. Will add more as soon as I run them.

### ARCH: aarch64 
#### KERNEL: kernel-64k
```bash
# BEFORE REBOOT
[root@ampere-mtsnow-altramax-02-vm-16 ~]# leapp-inspector messages --type KernelInfo
                          PRODUCED MESSAGES
Stamp: 2026-06-21T17:16:29.322504Z
Actor: scan_source_kernel
Phase: FactsCollection
Type: KernelInfo
Message_data:
{
    "page_size": "64k",
    "pkg": {
        "arch": "aarch64",
        "epoch": "0",
        "module": null,
        "name": "kernel-64k-core",
        "packager": "Red Hat, Inc. <http://bugzilla.redhat.com/bugzilla>",
        "pgpsig": "RSA/SHA256, Thu 11 Jun 2026 09:24:33 AM EDT, Key ID 199e2f91fd431d51",
        "release": "715.el9",
        "repository": "rhel-9-for-aarch64-baseos-rpms",
        "stream": null,
        "version": "5.14.0"
    },
    "type": "ordinary",
    "uname_r": "5.14.0-715.el9.aarch64+64k"
}

[root@ampere-mtsnow-altramax-02-vm-16 ~]# grep 'Detected kernel version inside container\|Building initram disk for kernel' /var/log/leapp/leapp-upgrade.log
2026-06-21 13:23:01.71  DEBUG    PID: 37616 leapp.workflow.InterimPreparation.upgrade_initramfs_generator: Detected kernel version inside container: 6.12.0-211.7.1.el10_2.aarch64+64k.
2026-06-21 13:23:01.193 DEBUG    PID: 37616 leapp.workflow.InterimPreparation.upgrade_initramfs_generator: Building initram disk for kernel: 6.12.0-211.7.1.el10_2.aarch64+64k
2026-06-21 13:23:33.153 DEBUG    PID: 37616 leapp.workflow.InterimPreparation.upgrade_initramfs_generator: Building initram disk for kernel: 6.12.0-211.7.1.el10_2.aarch64+64k finished

# AFTER UPGRADE
[root@ampere-mtsnow-altramax-02-vm-16 ~]# leapp-inspector messages --type InstalledTargetKernelInfo
                          PRODUCED MESSAGES
Stamp: 2026-06-21T17:48:10.436032Z
Actor: scan_installed_target_kernel_version
Phase: RPMUpgrade
Type: InstalledTargetKernelInfo
Message_data:
{
    "initramfs_path": "/boot/initramfs-6.12.0-211.7.1.el10_2.aarch64+64k.img",
    "kernel_img_path": "/boot/vmlinuz-6.12.0-211.7.1.el10_2.aarch64+64k",
    "pkg_nevra": "kernel-64k-core-6.12.0-211.7.1.el10_2.aarch64",
    "uname_r": "6.12.0-211.7.1.el10_2.aarch64+64k"
}

[root@ampere-mtsnow-altramax-02-vm-16 ~]# uname -r
6.12.0-211.7.1.el10_2.aarch64+64k
```

### ARCH: ppc64le 
#### KERNEL: kernel
```bash
#BEFORE REBOOT
[root@ibm-p9z-15-lp7 ~]# leapp-inspector messages --type KernelInfo
                          PRODUCED MESSAGES
Stamp: 2026-06-21T15:35:45.378449Z
Actor: scan_source_kernel
Phase: FactsCollection
Type: KernelInfo
Message_data:
{
    "page_size": "64k",
    "pkg": {
        "arch": "ppc64le",
        "epoch": "0",
        "module": null,
        "name": "kernel-core",
        "packager": "Red Hat, Inc. <http://bugzilla.redhat.com/bugzilla>",
        "pgpsig": "(none)",
        "release": "716.el9",
        "repository": "beaker-BaseOS",
        "stream": null,
        "version": "5.14.0"
    },
    "type": "ordinary",
    "uname_r": "5.14.0-716.el9.ppc64le"
}

[root@ibm-p9z-15-lp7 ~]# grep 'Detected kernel version inside container\|Building initram disk for kernel' /var/log/leapp/leapp-upgrade.log
2026-06-21 13:15:08.368 DEBUG    PID: 69065 leapp.workflow.InterimPreparation.upgrade_initramfs_generator: Detected kernel version inside container: 6.12.0-211.7.1.el10_2.ppc64le.
2026-06-21 13:15:10.868 DEBUG    PID: 69065 leapp.workflow.InterimPreparation.upgrade_initramfs_generator: Building initram disk for kernel: 6.12.0-211.7.1.el10_2.ppc64le
2026-06-21 13:16:04.379 DEBUG    PID: 69065 leapp.workflow.InterimPreparation.upgrade_initramfs_generator: Building initram disk for kernel: 6.12.0-211.7.1.el10_2.ppc64le finished

# AFTER UPGRADE
[root@ibm-p9z-15-lp7 ~]# leapp-inspector messages --type InstalledTargetKernelInfo
                          PRODUCED MESSAGES
Stamp: 2026-06-21T18:04:11.897360Z
Actor: scan_installed_target_kernel_version
Phase: RPMUpgrade
Type: InstalledTargetKernelInfo
Message_data:
{
    "initramfs_path": "/boot/initramfs-6.12.0-211.7.1.el10_2.ppc64le.img",
    "kernel_img_path": "/boot/vmlinuz-6.12.0-211.7.1.el10_2.ppc64le",
    "pkg_nevra": "kernel-core-6.12.0-211.7.1.el10_2.ppc64le",
    "uname_r": "6.12.0-211.7.1.el10_2.ppc64le"
}

[root@ibm-p9z-15-lp7 ~]# uname -r
6.12.0-211.7.1.el10_2.ppc64le
```

### ARCH: x86_64 
#### KERNEL: kernel
```bash
#BEFORE REBOOT
[root@leapp-rhel-20260621171437 ~]# leapp-inspector messages --type KernelInfo
                          PRODUCED MESSAGES
Stamp: 2026-06-21T15:24:21.943010Z
Actor: scan_source_kernel
Phase: FactsCollection
Type: KernelInfo
Message_data:
{
    "page_size": "4k",
    "pkg": {
        "arch": "x86_64",
        "epoch": "0",
        "module": null,
        "name": "kernel-core",
        "packager": "Red Hat, Inc. <http://bugzilla.redhat.com/bugzilla>",
        "pgpsig": "RSA/SHA256, Tue 10 Feb 2026 04:26:08 AM UTC, Key ID 199e2f91fd431d51",
        "release": "678.el9",
        "repository": "anaconda",
        "stream": null,
        "version": "5.14.0"
    },
    "type": "ordinary",
    "uname_r": "5.14.0-678.el9.x86_64"
}

[root@leapp-rhel-20260621171437 ~]#   grep 'Detected kernel version inside container\|Building initram disk for kernel' /var/log/leapp/leapp-upgrade.log
2026-06-21 15:41:53.531 DEBUG    PID: 100608 leapp.workflow.InterimPreparation.upgrade_initramfs_generator: Detected kernel version inside container: 6.12.0-211.7.1.el10_2.x86_64.
2026-06-21 15:41:53.689 DEBUG    PID: 100608 leapp.workflow.InterimPreparation.upgrade_initramfs_generator: Building initram disk for kernel: 6.12.0-211.7.1.el10_2.x86_64
2026-06-21 15:43:07.848 DEBUG    PID: 100608 leapp.workflow.InterimPreparation.upgrade_initramfs_generator: Building initram disk for kernel: 6.12.0-211.7.1.el10_2.x86_64 finished

#AFTER UPGRADE
[root@leapp-rhel-20260621171437 ~]# leapp-inspector messages --type InstalledTargetKernelInfo
                          PRODUCED MESSAGES
Stamp: 2026-06-21T16:01:56.810793Z
Actor: scan_installed_target_kernel_version
Phase: RPMUpgrade
Type: InstalledTargetKernelInfo
Message_data:
{
    "initramfs_path": "/boot/initramfs-6.12.0-211.7.1.el10_2.x86_64.img",
    "kernel_img_path": "/boot/vmlinuz-6.12.0-211.7.1.el10_2.x86_64",
    "pkg_nevra": "kernel-core-6.12.0-211.7.1.el10_2.x86_64",
    "uname_r": "6.12.0-211.7.1.el10_2.x86_64"
}

[root@leapp-rhel-20260621171437 ~]# uname -r
6.12.0-211.7.1.el10_2.x86_64
```

repos/system_upgrade/common/libraries/kernel.py
```diff
@@ -15,18 +24,32 @@ class KernelType:
     REALTIME = 'realtime'
 
 
+class KernelPageSize:
+    DEFAULT = '4k'
+    LARGE = '64k'
```
PeterMocary: We could probably change the naming to something more direct (e.g. ``SIZE_4k`` and ``SIZE_64k``).

PeterMocary:
Fixed failing unit tests.

repos/system_upgrade/common/actors/commonleappdracutmodules/libraries/modscan.py
```diff
@@ -95,6 +94,13 @@ def _create_initram_packages():
         required_pkgs.append('NetworkManager')
     if version.get_target_major_version() == '9':
         required_pkgs += ['policycoreutils', 'rng-tools']
+
+    kernel_info = next(api.consume(KernelInfo), None)
+    page_size = kernel_info.page_size if kernel_info else kernel_lib.KernelPageSize.DEFAULT
```
matejmatuska: I would drop this fallback, KernelInfo is either produced or the producing actor raises StopActorExecutionErr.

I think maybe it would be a good idea to even raise StopActorExecutionErr if there is no such message.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
```diff
@@ -25,61 +26,46 @@
...
 
 
 def _get_target_kernel_version(context):
     """
     Get the version of the most recent kernel version within the container.
     """
+    kernel_info = next(api.consume(KernelInfo), None)
+    page_size = kernel_info.page_size if kernel_info else kernel_lib.KernelPageSize.DEFAULT
+    arch = api.current_actor().configuration.architecture
+    kernel_core_pkg = kernel_lib.get_kernel_pkg_names(kernel_lib.KernelType.ORDINARY, page_size, arch).core
```
matejmatuska: Why not use the actual type here?
```suggestion
    kernel_core_pkg = kernel_lib.get_kernel_pkg_names(kernel_info.type, page_size, arch).core
```
if there is a reason why not, it should be commented here.

repos/system_upgrade/common/actors/commonleappdracutmodules/libraries/modscan.py
```diff
@@ -95,6 +94,13 @@ def _create_initram_packages():
...
+    kernel_packages = kernel_lib.get_kernel_pkg_names(kernel_lib.KernelType.ORDINARY, page_size,
```
matejmatuska: Same as one of the other comments:
```suggestion
    kernel_packages = kernel_lib.get_kernel_pkg_names(kernel_info.type, page_size,
```

repos/system_upgrade/common/libraries/kernel.py
```diff
@@ -45,70 +61,131 @@ def determine_kernel_type_from_uname(rhel_version, kernel_uname_r):
...
-    Raises ``StopActorExecutionError`` if no package provides given uname_r or if the internal rpm query fails.
-    :param str uname_r: NEVRA of an installed kernel package
+    Queries RPM for packages owning ``/lib/modules/<uname_r>`` and returns info about the ``-core``
+    kernel package among the results.
+
+    :param uname_r: Kernel release (uname-r)
+    :type uname_r: str
     :returns: Information about the kernel package providing given uname_r
     :rtype: KernelPkgInfo
+    :raises KernelPackageInfoError: If no package provides given uname_r or if the internal rpm query fails
```
matejmatuska, line 150: Given that the type of the raised exception is changed, you should deprecate the orig function and create a new one or if you don't want to do that just keep the StopActorExecutionError.

repos/system_upgrade/common/libraries/kernel.py
```diff
@@ -45,70 +61,131 @@ def determine_kernel_type_from_uname(rhel_version, kernel_uname_r):
...
+    """
+    try:
+        pkg_nevras = run(['rpm', '-q', '--whatprovides', '/lib/modules/{}'.format(uname_r)],
+                         split=True)['stdout']
+    except CalledProcessError:
+        raise KernelPackageInfoError(
+            message='Unable to obtain kernel information of the booted kernel.',
+            details={'Problem': 'No package owns /lib/modules/{}.'.format(uname_r)})
```
matejmatuska: keep the orig error, e.g.:

```suggestion
    except CalledProcessError as err:
        raise KernelPackageInfoError(
            message='Unable to obtain kernel information of the booted kernel.',
            details={'Problem': 'No package owns /lib/modules/{}: }{}'.format(uname_r, err)})
```

repos/system_upgrade/common/libraries/kernel.py
```diff
@@ -45,70 +61,131 @@ def determine_kernel_type_from_uname(rhel_version, kernel_uname_r):
...
     Query the RPM database for information about the given kernel package.
 
-    Calls the ``rpm`` command internally and might raise CalledProcessError if the rpm query fails.
-
-    :param str kernel_pkg_nevra: NEVRA of an installed kernel package
+    :param kernel_pkg_nevra: NEVRA of an installed kernel package
+    :type kernel_pkg_nevra: str
     :returns: Information about the given kernel package
     :rtype: KernelPkgInfo
+    :raises KernelPackageInfoError: If the rpm query fails
```
matejmatuska, line 126: This also changes the API (the orig function propagates CalledProcessError), see my other comment about deprecation in the func below.

repos/system_upgrade/common/libraries/kernel.py
```diff
@@ -45,70 +61,131 @@ def determine_kernel_type_from_uname(rhel_version, kernel_uname_r):
...
     """
     query_format = '%{NAME}|%{VERSION}|%{RELEASE}|%{ARCH}|'
-    pkg_info = run(['rpm', '-q', '--queryformat', query_format, kernel_pkg_nevra])['stdout'].strip().split('|')
+    try:
+        pkg_info = run(['rpm', '-q', '--queryformat', query_format, kernel_pkg_nevra])['stdout'].strip().split('|')
+    except CalledProcessError:
+        raise KernelPackageInfoError(
+            message='Unable to obtain kernel package information.',
+            details={'Problem': 'Failed to query RPM info for {}.'.format(kernel_pkg_nevra)})
```
matejmatuska: Again let's keep the error message:

```suggestion
    except CalledProcessError as err:
        raise KernelPackageInfoError(
            message='Unable to obtain kernel package information.',
            details={'Problem': 'Failed to query RPM info for {}: {}'.format(kernel_pkg_nevra, err)})
```

repos/system_upgrade/common/libraries/kernel.py
```diff
@@ -45,70 +61,131 @@ def determine_kernel_type_from_uname(rhel_version, kernel_uname_r):
...
+    :returns: Kernel page size
+    :rtype: str
+    """
+    try:
+        result = run(['getconf', 'PAGE_SIZE'])['stdout'].strip()
+        if result == '65536':
+            return KernelPageSize.LARGE
+    except CalledProcessError:
+        api.current_logger().warning('Failed to determine kernel page size via getconf, defaulting to 4k.')
+    return KernelPageSize.DEFAULT
```
matejmatuska: I am unsure if this is a good idea for a library function, the caller can never tell if it failed or not. I think it might be a better idea to return None in case of error and let the caller supply the defualt, wdyt?

repos/system_upgrade/common/libraries/kernel.py
```diff
@@ -45,70 +61,131 @@ def determine_kernel_type_from_uname(rhel_version, kernel_uname_r):
     return KernelType.ORDINARY
 
 
-def get_uname_r_provided_by_kernel_pkg(kernel_pkg_nevra):
+def determine_kernel_page_size() -> str:
+    """
+    Determine kernel page size for the running kernel using ``getconf PAGE_SIZE``.
+
+    :returns: Kernel page size
```
matejmatuska: Same as the above function, we should somehow clarify this:
```suggestion
    :returns: Kernel page size (values come exclusively from the KernelPageSize class)
```
I think the wording might be better, this is just a quick suggestion.

repos/system_upgrade/common/libraries/kernel.py
```diff
@@ -1,32 +1,47 @@
...
     """
     Determine kernel type from given kernel release (uname-r).
 
-    :param str rhel_version: Version of RHEL for which the kernel with the uname-r is targeted.
-    :param str kernel_uname_r: Kernel release (uname-r)
+    :param rhel_version: Version of RHEL for which the kernel with the uname-r is targeted.
+    :type rhel_version: str
+    :param kernel_uname_r: Kernel release (uname-r)
+    :type kernel_uname_r: str
     :returns: Kernel type based on a given uname_r
```
matejmatuska: let's make this clear:
```suggestion
    :returns: Kernel type based on a given uname_r (values come exclusively from the KernelType class)
```

repos/system_upgrade/common/libraries/kernel.py
matejmatuska: 100% agree, when reading the code it makes more sense to see the actual size. I would maybe call it `PAGE_SIZE_Xk`

repos/system_upgrade/common/actors/scaninstalledtargetkernelversion/actor.py
```diff
@@ -8,10 +8,11 @@ class ScanInstalledTargetKernelVersion(Actor):
     """
     Scan for the version of the newly installed kernel
 
-    This actor will query rpm for all kernel-core packages and reports the
-    first matching target system kernel RPM. In case the RHEL Real Time has
-    been detected on the original system, the kernel-rt-core rpm is searched.
-    If the rpm is missing, fallback for standard kernel RPM.
+    Based on the source kernel type and page size, this actor determines the
+    appropriate target kernel package (e.g. kernel-core, kernel-64k-core,
+    kernel-rt-core) and reports the installed target kernel RPM. If the
```
matejmatuska: "report" might be confusing as no reporting is done
```suggestion
    kernel-rt-core) and produces InstalledTargetKernelInfo containing the installed target kernel RPM. If the
```

repos/system_upgrade/common/actors/scaninstalledtargetkernelversion/libraries/scankernel.py
```diff
@@ -73,19 +80,25 @@ def process():
     if not src_kernel_info:
         return  # Will not happen, other actors would inhibit the upgrade
 
-    target_ver = get_target_major_version()
-    target_kernel_pkg_name = get_kernel_pkg_name(target_ver, src_kernel_info.type)
+    arch = api.current_actor().configuration.architecture
+    target_kernel_pkg_name = kernel_lib.get_kernel_pkg_names(src_kernel_info.type,
+                                                             src_kernel_info.page_size, arch).core
```
matejmatuska: Just a nitpick, I don't like how the `.core` is "hidden" there

repos/system_upgrade/common/actors/scaninstalledtargetkernelversion/libraries/scankernel.py
```diff
@@ -73,19 +80,25 @@ def process():
...
     target_kernel_nevra = get_target_kernel_package_nevra(target_kernel_pkg_name)
 
     if src_kernel_info.type != kernel_lib.KernelType.ORDINARY and not target_kernel_nevra:
-        api.current_logger().warning('The kernel-rt-core rpm from the target RHEL has not been detected. Switching '
-                                     'to non-preemptive kernel.')
-        target_kernel_pkg_name = get_kernel_pkg_name(target_ver, kernel_lib.KernelType.ORDINARY)
+        # Fallback to ordinary kernel if target kernel nevra could not be determined
+        api.current_logger().warning('The %s rpm from the target RHEL has not been detected. Switching '
+                                     'to non-preemptive kernel.', target_kernel_pkg_name)
+        target_kernel_pkg_name = kernel_lib.get_kernel_pkg_names(kernel_lib.KernelType.ORDINARY,
```
matejmatuska: again, comment why hardcoding ORDINARY

repos/system_upgrade/common/actors/scaninstalledtargetkernelversion/actor.py
```diff
@@ -8,10 +8,11 @@ class ScanInstalledTargetKernelVersion(Actor):
...
+    expected variant is missing, it falls back to the ordinary kernel for
+    the same page size.
```
matejmatuska: Paraphrasing the answer as to why fallback from rt->ordinary is okay from Petr in one of the discussions in this PR, sorry for the formatting:

```suggestion
    the same page size. Note that this means that a fallback from a realtime to "ordinary" kernel might happen. This is because the user might not have a repository providing the realtime kernel and there is no way to check whether such package is present there (missing data propagated from the dnf plugin back to actor). In such case, the user needs to enable the repository and install the content after the upgrade manually - which is preferred to ending up with an unbootable system.
```

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
```diff
@@ -25,61 +26,46 @@
...
-LEAPP_CMDLINE_CONF_PATH = '/etc/cmdline.d/50-leapp.conf'
 DEDICATED_LEAPP_PART_URL = 'https://access.redhat.com/solutions/7011704'
 
 
 def _get_target_kernel_version(context):
     """
     Get the version of the most recent kernel version within the container.
     """
+    kernel_info = next(api.consume(KernelInfo), None)
+    page_size = kernel_info.page_size if kernel_info else kernel_lib.KernelPageSize.DEFAULT
```
matejmatuska: Same as in the other actor, we expect kernel_info to always exist, I would drop the fallback.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/files/generate-initram.sh
```diff
@@ -7,7 +7,10 @@ stage() {
 }
 
 get_kernel_version() {
-    rpm -qa kernel --qf '%{VERSION}-%{RELEASE}.%{ARCH}\n' | sort --version-sort | tail --lines=1
+    # The actor should always provide the version via LEAPP_KERNEL_VERSION and ensure
+    # only one kernel is installed in the container. This is simple fallback in case
+    # the version is not provided.
+    rpm -qa kernel kernel-64k kernel-rt kernel-rt-64k --qf '%{VERSION}-%{RELEASE}.%{ARCH}\n' | sort --version-sort | tail --lines=1
```
matejmatuska: I think it might be enough to just query for kernel, wouldn't that include the 64k kernels too? Haven't tried yet though

matejmatuska (commented):
Some required changes, but overal I think the solution is good.
**I have yet to take a look at the unit tests and test it though.**

Another thing is, we haven't devided on the type hints yet, but since they do not break anything I am not against them, in fact the opposite.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
PeterMocary: I'm not sure if having a real time kernel during the upgrade procedure makes any difference. Since RT kernel is about the scheduling changes and performance it should not matter during the upgrade (i think). On top of that, it haven't been used before either.

On the other hand, having the correct page size is important since there might be addressing issues if 4k page size kernel is used with file system with 64k blocks (i think). On top of that, Red Hat does not recommend switching to different page size kernel without reinstalling the system ([source](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/automatically_installing_rhel/installing-kernel-64k-on-arm-using-the-command-line_rhel-installer)).

So the idea was to use ordinary kernel during the upgrade procedure (same as before) but use the same page size as the source kernel. The target kernel then respects both `KernelType` and `KernelPageSize`.

That being said, I don't really see any problem with using RT during the upgrade procedure so we could change it as you suggest.

repos/system_upgrade/common/libraries/kernel.py
PeterMocary: The new exception is derived from the `StopActorExecutionError` to match what we did in `dnflibs`. 

Regardless, the function signature is technically changed so deprecation should be done i guess. What do you think @pirat89?

I also changed the `get_uname_r_provided_by_kernel_pkg`. The context parameter was added there. It is backward compatible (same as in the above case), but signature was changed.

I think this library could use a bit of a unification of return values and exceptions, maybe some function and argument renaming. So I'd rather leave the deprecating on some refactoring that would unify these functions.

Discrepancies I see:
- `determine_kernel_type_from_uname`
  - function mane has only `uname` (`uname_r` is used in others)
  - argument called `rhel_version` (with todo for its renaming)
- `get_uname_r_provided_by_kernel_pkg`
  - already added context to the signature to make it usable with the initram context
  - returns empty string when `CalledProcessError` happens (others raise an exception)
- `get_kernel_pkg_info`
  - changed to raise `KernelPackageInfoError` when `CalledProcessError` happens.
  - could also use context (not needed now but to follow same structure in the library)
- `get_kernel_pkg_info_for_uname_r`
  - changed to raise `KernelPackageInfoError` when `CalledProcessError` happens and if kernel pkg info not found.
  - could also use context (not needed now but to follow same structure in the library)
- the new function `determine_kernel_page_size`
  - outputs warning on `CalledProcessError` and defaults to 4k page size. There is another [comment](https://github.com/oamg/leapp-repository/pull/1550#discussion_r3451418396) adressing this - we could either return `None` or add a new exception e.g. `KernelPageSizeError` to make it similar to other functions.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/files/generate-initram.sh
PeterMocary: Nope, that would need to be kernel* and then also other packages mix in.

```Bash
[root@ampere-mtsnow-altramax-02-vm-16 ~]# rpm -qa kernel kernel-64k
kernel-5.14.0-716.el9.aarch64
kernel-64k-5.14.0-715.el9.aarch64
kernel-64k-6.12.0-211.7.1.el10_2.aarch64
kernel-6.12.0-211.7.1.el10_2.aarch64
[root@ampere-mtsnow-altramax-02-vm-16 ~]# rpm -qa kernel
kernel-5.14.0-716.el9.aarch64
kernel-6.12.0-211.7.1.el10_2.aarch64
```

Note that this includes RT kernels despite it is currently not used in initram (being discussed in other comments). Regardless this is effectively dead code as the `LEAPP_KERNEL_VERSION` is always produced. But still, I'd keep it in as a safety net.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
matejmatuska: Let's see what @pirat89 says :)

matejmatuska: But if the ordinary kernel is used here, there should be a comment why IMO

repos/system_upgrade/common/libraries/kernel.py
matejmatuska: I want to note that I didn't notice that the exception is derived from StopActorExecErr, that's an important detail.

When a new parameter with a default value is added I don't think we should deprecated the function, this is done in Python's stdlib multiple times from what I noticed and there are no deprecations.

yes the lib could definitely use some unification of the exceptions and return vals.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/files/generate-initram.sh
matejmatuska: Right, sorry I somehow thought it uses `rpm -q --whatprovides kernel`, though now I am not sure why it doesn't

repos/system_upgrade/common/actors/commonleappdracutmodules/tests/test_modscan_commonleappdracutmodules.py
```diff
@@ -74,15 +100,16 @@ def check_module(mod):
...
+    for pkg in ['biosdevname', 'policycoreutils', 'rng-tools', 'kernel', 'kernel-core', 'kernel-modules']:
         assert pkg not in modscan._REQUIRED_PACKAGES
 
     required_packages = modscan._REQUIRED_PACKAGES[:]
     if dst_ver[0] == '9':
         required_packages += ['policycoreutils', 'rng-tools']
     if arch == architecture.ARCH_X86_64:
         required_packages += ['biosdevname']
+    # no KernelInfo message -> defaults to 4k
+    required_packages += ['kernel', 'kernel-core', 'kernel-modules']
```
matejmatuska, line 112: I think that `kernel` is not strictly required, kernel is a metapackage (afaik it even doesn't contain any files), but there is probably nothing wrong with this in a test.

repos/system_upgrade/common/actors/commonleappdracutmodules/tests/test_modscan_commonleappdracutmodules.py
```diff
@@ -91,6 +118,31 @@ def test_required_packages(monkeypatch, src_ver, dst_ver, arch):
...
+    (_KERNEL_INFO_4K, architecture.ARCH_ARM64,
+     ['kernel', 'kernel-core', 'kernel-modules'],
+     ['kernel-64k', 'kernel-64k-core', 'kernel-64k-modules']),
+    (_KERNEL_INFO_64K, architecture.ARCH_ARM64,
+     ['kernel-64k', 'kernel-64k-core', 'kernel-64k-modules'],
+     ['kernel', 'kernel-core', 'kernel-modules']),
+    # ppc64le: 64k page size but standard kernel packages (no kernel-64k RPMs)
+    (_KERNEL_INFO_64K_PPC, architecture.ARCH_PPC64LE,
+     ['kernel', 'kernel-core', 'kernel-modules'],
+     ['kernel-64k', 'kernel-64k-core', 'kernel-64k-modules']),
```
matejmatuska, line 139: I would add a test case for x86_64 and s390x as well. In the future there might be some new kernel configs that we might need to handle, so better futureproof the test.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/tests/unit_test_upgradeinitramfsgenerator.py
```diff
@@ -558,3 +492,97 @@ def create_upgrade_hmac_from_target_hmac_mock(uspace_kernel_hmac_path,
...
+        uname_r='5.14.0-100.el9.{}'.format(arch),
+        page_size=page_size,
+    )
+
+
+@pytest.mark.parametrize('page_size,arch,expected_pkg,expected_uname_r', [
+    ('4k', 'aarch64', 'kernel-core', '5.14.0-100.el9.aarch64'),
+    ('64k', 'aarch64', 'kernel-64k-core', '5.14.0-100.el9.aarch64+64k'),
+    # ppc64le: 64k page size but standard kernel packages
+    ('64k', 'ppc64le', 'kernel-core', '5.14.0-100.el9.ppc64le'),
```
matejmatuska, line 588: same as in the other test, add a test for x86_64 and s390x

repos/system_upgrade/common/actors/scaninstalledtargetkernelversion/tests/test_scaninstalledkernel_scaninstalledtargetkernelversion.py
```diff
@@ -44,38 +53,58 @@ def assert_produced_messages_are_correct(produced_messages, expected_target_nevr
 
 
 @pytest.mark.parametrize(
-    ('is_rt', 'expected_target_nevra', 'stdouts'),
+    ('kernel_type', 'page_size', 'arch', 'expected_target_nevra', 'stdouts', 'dst_ver'),
     [
-        (False, TARGET_KERNEL_NEVRA, {'kernel-core': [OLD_KERNEL_NEVRA, TARGET_KERNEL_NEVRA]}),
-        (False, TARGET_KERNEL_NEVRA, {'kernel-core': [TARGET_KERNEL_NEVRA, OLD_KERNEL_NEVRA]}),
-        (False, TARGET_KERNEL_NEVRA, {
+        (kernel_lib.KernelType.ORDINARY, kernel_lib.KernelPageSize.DEFAULT, architecture.ARCH_X86_64,
```
matejmatuska: Just a suggestion, maybe import the `KernelType` and `KernelPageSize` and the individual architectures directly so you don't have to always specify the library name, it would make it much more readable IMO.

matejmatuska: I also think it might be a good idea to use a different style of formatting to make this more readable, but that's of course subjective:

e.g.:
```python
        (
            kernel_lib.KernelType.ORDINARY, 
            kernel_lib.KernelPageSize.DEFAULT, 
            architecture.ARCH_X86_64,
            TARGET_KERNEL_NEVRA, 
            {'kernel-core': [OLD_KERNEL_NEVRA, TARGET_KERNEL_NEVRA]}, 
            '9.0'
         ),
```
and maybe also expand the stdouts dict.

repos/system_upgrade/common/libraries/tests/test_kernel_lib.py
```diff
@@ -36,64 +37,146 @@ def test_determine_kernel_type_from_uname(monkeypatch, version, uname_r, expecte
...
+        (KernelType.ORDINARY, KernelPageSize.DEFAULT, 'ppc64le', ('kernel', 'kernel-core', 'kernel-modules')),
+        (KernelType.REALTIME, KernelPageSize.DEFAULT, 'x86_64', ('kernel-rt', 'kernel-rt-core', 'kernel-rt-modules')),
+        # LARGE page size on aarch64: 64k variant packages
+        (KernelType.ORDINARY, KernelPageSize.LARGE, 'aarch64',
+         ('kernel-64k', 'kernel-64k-core', 'kernel-64k-modules')),
+        (KernelType.REALTIME, KernelPageSize.LARGE, 'aarch64',
+         ('kernel-rt-64k', 'kernel-rt-64k-core', 'kernel-rt-64k-modules')),
+        # LARGE page size on non-aarch64: standard packages (no kernel-64k RPMs exist)
+        (KernelType.ORDINARY, KernelPageSize.LARGE, 'ppc64le', ('kernel', 'kernel-core', 'kernel-modules')),
+        (KernelType.ORDINARY, KernelPageSize.LARGE, 'x86_64', ('kernel', 'kernel-core', 'kernel-modules')),
```
matejmatuska: I would add one case for s390x

matejmatuska (changes requested):
Finally reviewed also the unit tests, btw I accidentally used "Comment" instead of "Request changes" before.

repos/system_upgrade/common/models/installedtargetkernelversion.py
```diff
@@ -30,6 +28,9 @@ class KernelInfo(Model):
     # @FixMe(mhecko): I want to use kernel_lib.KernelType here, but I cannot import any library code (yet).
     # #               Figure out how to do it.
 
+    page_size = fields.StringEnum(['4k', '64k'], default='4k')
+    """Memory page size of the kernel."""
+
```
pirat89, line 42: @PeterMocary  It's expected to have the whitespaces around docstrings. The thing that it's confusing for people always it's one of reasons why we suggest to use the long format always:
```
"""
Docstring
"""
```
Where it's clear and no mistake is made.

repos/system_upgrade/common/actors/commonleappdracutmodules/tests/test_modscan_commonleappdracutmodules.py
PeterMocary: This was included before as well. I just modified it to allow other kernel types as well. The `get_kernel_pkg_names` function returns the meta package as well as the core and modules.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/files/generate-initram.sh
PeterMocary: Yes that would be able to find all of the kernel packages as well:

```Bash
[root@hpe-apollo-cn99xx-14-vm-05 ~]# rpm -q --whatprovides kernel
kernel-core-5.14.0-687.17.1.el9_8.aarch64
kernel-5.14.0-687.17.1.el9_8.aarch64
kernel-rt-64k-core-5.14.0-687.17.1.el9_8.aarch64
kernel-rt-core-5.14.0-687.17.1.el9_8.aarch64
kernel-64k-core-5.14.0-687.17.1.el9_8.aarch64
```

repos/system_upgrade/common/libraries/kernel.py
PeterMocary: Changed to `PAGE_SIZE_<size>K`.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
PeterMocary: Other relevant comments:
https://github.com/oamg/leapp-repository/pull/1550#discussion_r3451239006
https://github.com/oamg/leapp-repository/pull/1550#discussion_r3451501542

I resolved those, to not split the conversation into multiple threads.

repos/system_upgrade/common/actors/commonleappdracutmodules/libraries/modscan.py
PeterMocary: Resolving in favor of this thread https://github.com/oamg/leapp-repository/pull/1550#discussion_r3451230909. Let's keep the discussion in one place.

repos/system_upgrade/common/actors/scaninstalledtargetkernelversion/libraries/scankernel.py
PeterMocary: Resolving in favor of this thread https://github.com/oamg/leapp-repository/pull/1550#discussion_r3451230909. Let's keep the discussion in one place.

repos/system_upgrade/common/libraries/kernel.py
PeterMocary: Good point, mentioned it in https://github.com/oamg/leapp-repository/pull/1550#discussion_r3453252469. I'm resolving this so that the discussion does not get split.

repos/system_upgrade/common/libraries/kernel.py
PeterMocary: This is part of the discussion in https://github.com/oamg/leapp-repository/pull/1550#discussion_r3453252469. I'm resolving this so that the discussion is in one thread.

repos/system_upgrade/common/libraries/kernel.py
PeterMocary: Other relevant comments I resolved to keep the discussion in this thread:
https://github.com/oamg/leapp-repository/pull/1550#discussion_r3451336067
https://github.com/oamg/leapp-repository/pull/1550#discussion_r3451418396

All of these are mentioned in my summary here.

PeterMocary:
Fixed majority of the suggestions (all fixed comments are marked resolved). The major things left to solve are: 
- kernel library unification and possible deprecations
- whether to use source kernel during the upgrade as well

PeterMocary:
Rebased.

repos/system_upgrade/common/actors/initramfs/upgradeinitramfsgenerator/libraries/upgradeinitramfsgenerator.py
PeterMocary: Discussed on call, and https://github.com/oamg/leapp-repository/pull/1550#discussion_r3452903828 has basically correct reasoning. I'll add comments in as suggested by @matejmatuska.

repos/system_upgrade/common/libraries/kernel.py
```diff
@@ -45,70 +61,131 @@ def determine_kernel_type_from_uname(rhel_version, kernel_uname_r):
...
-
-    raise StopActorExecutionError(message='Unable to obtain kernel information of the booted kernel: no package is '
-                                          'providing the booted kernel release returned by uname.')
+    parts = ['kernel']
+    if kernel_type == KernelType.REALTIME:
+        parts.append('rt')
+    if kernel_page_size == KernelPageSize.PAGE_SIZE_64K and arch == arch_config.ARCH_ARM64:
+        parts.append('64k')
+    base = '-'.join(parts)
+    return KernelPkgNames(base=base, core='{}-core'.format(base), modules='{}-modules'.format(base))
```
pirat89, line 195: @PeterMocary  Why modules? there is also kernel-*-modules-core which is not covered, but where is "modules" used actually? Most of the time we are interested just about the core package.

pirat89: ah. `required_pkgs.extend(kernel_packages)` in modscan

repos/system_upgrade/common/actors/commonleappdracutmodules/tests/test_modscan_commonleappdracutmodules.py
pirat89: For this change let's keep it as it was previously to not introduce a problem so close to release. But I think we added it there just to be sure. Also, when you requier kernel (base package), the core and modules packages are installed always as well. So other two are kind of redundant in such a case. But having all of them noted should not be a problem.

repos/system_upgrade/common/actors/scaninstalledtargetkernelversion/libraries/scankernel.py
```diff
@@ -73,19 +80,27 @@ def process():
...
+    target_kernel_nevra = get_target_kernel_package_nevra(target_kernel_core_pkg)
 
     if src_kernel_info.type != kernel_lib.KernelType.ORDINARY and not target_kernel_nevra:
-        api.current_logger().warning('The kernel-rt-core rpm from the target RHEL has not been detected. Switching '
-                                     'to non-preemptive kernel.')
-        target_kernel_pkg_name = get_kernel_pkg_name(target_ver, kernel_lib.KernelType.ORDINARY)
-        target_kernel_nevra = get_target_kernel_package_nevra(target_kernel_pkg_name)
+        # Fallback to ordinary kernel if target kernel nevra could not be determined
+        api.current_logger().warning('The %s rpm from the target RHEL has not been detected. Switching '
+                                     'to non-preemptive kernel.', target_kernel_core_pkg)
```
pirat89, line 95: @PeterMocary  btw, the ordinary modern kernel is nowadays preemptive as well :) the difference between RT and ordinary kernel is that RT one is even more preemptive - reduce the worst waiting time to crazy values in comparison to the ordinary one.

repos/system_upgrade/common/libraries/kernel.py
pirat89: @PeterMocary  `get_uname_r_provided_by_kernel_pkg` - changed signature does not matter if the original call has the same results. Change in exception is not ideal but I think that in this case it's reasonable. But then it needs to be mentioned in changelogs :) ideally also in PR description.

But, `KERNEL_UNAME_R_PROVIDES` has been removed instead of marking it as deprecated :-P but let's keep it removed and just mention it in changelogs.  (UPDATE: it's deprecated now)

pirat89 (changes requested):
Discovered some nitpicks that are ok, but also one bug for IPU 8 -> 9 on ARM, where only 4k target kernel is installed instead of 64k one which leads to the emergency mode:

```
[  568.600105] upgrade[17638]: ============================================================
[  568.600142] upgrade[17638]:                            ERRORS
[  568.600179] upgrade[17638]: ============================================================
[  568.600215] upgrade[17638]: 2026-06-23 19:51:20.137473 [ERROR] Actor: target_initramfs_generator
[  568.600249] upgrade[17638]: Message: Cannot get version of the installed target OS kernel
[  568.600281] upgrade[17638]: Summary:
[  568.600314] upgrade[17638]:     Problem: Did not receive a message with installed target OS kernel version (InstalledTargetKernelVersion)
[  568.600347] upgrade[17638]: ============================================================
[  568.600379] upgrade[17638]:                        END OF ERRORS
[  568.600411] upgrade[17638]: ============================================================

```

Note the actor does not print any logs about the reason which we could improve I believe.

The fix seems to be easy, but keeping it for the morning. producing a task to install correct rpms and do not require upgrade of current one should resolve the problem.

repos/system_upgrade/common/libraries/kernel.py
```diff
@@ -45,70 +61,131 @@ def determine_kernel_type_from_uname(rhel_version, kernel_uname_r):
...
+        pkg_info = get_kernel_pkg_info(pkg_nevra)
+        if pkg_info.name.endswith('-core'):
+            return pkg_info
+
+    raise KernelPackageInfoError(
+        message='Unable to obtain kernel information of the booted kernel.',
+        details={'Problem': 'No installed package owning /lib/modules/{} is a kernel core package.'.format(uname_r)})
+
+
+def get_kernel_pkg_names(kernel_type: str, kernel_page_size: str, arch: str) -> KernelPkgNames:
```
pirat89: @PeterMocary  The function must accept the version of the system for which the pkg names are generated. Alternatively, you can rename it to `get_target_kernel_pkg_names` and you do not have to change the input params. Also, you would not need to change the logic in the function as el8 is the only exception at this very moment.

PeterMocary:
Renamed `get_kernel_pkg_names` to `get_target_kernel_pkg_names`.

PeterMocary:
Added KERNEL_UNAME_R_PROVIDES back in as deprecated.

repos/system_upgrade/common/actors/commonleappdracutmodules/libraries/modscan.py
```diff
@@ -95,6 +95,15 @@ def _create_initram_packages():
         required_pkgs.append('NetworkManager')
     if version.get_target_major_version() == '9':
         required_pkgs += ['policycoreutils', 'rng-tools']
+
+    kernel_info = next(api.consume(KernelInfo), None)
+    if not kernel_info:
+        raise StopActorExecutionError('Could not retrieve KernelInfo message.')
+    page_size = kernel_info.page_size
+    kernel_packages = kernel_lib.get_target_kernel_pkg_names(kernel_lib.KernelType.ORDINARY, page_size,
+                                                      api.current_actor().configuration.architecture)
```
pirat89: nitpitck for pylint :)
```suggestion
    kernel_packages = kernel_lib.get_target_kernel_pkg_names(
        kernel_lib.KernelType.ORDINARY,
        page_size,
        api.current_actor().configuration.architecture
    )
```

repos/system_upgrade/common/libraries/kernel.py
PeterMocary: Made it deprecated. Added a summary for the kernel.py relevant changes to the description for easier changelog creation for devs. Other points in this thread can be done as refactoring in follow up PRs.

PeterMocary:
Retested with the new RHEL8 changes:

- RHEL8 on aarch64 with kernel-64k:
```Bash
[root@hpe-apollo-cn99xx-14-vm-04 ~]# leapp-inspector messages --type KernelInfo
                          PRODUCED MESSAGES
Stamp: 2026-06-24T15:17:04.422089Z
Actor: scan_source_kernel
Phase: FactsCollection
Type: KernelInfo
Message_data:
{
    "page_size": "64k",
    "pkg": {
        "arch": "aarch64",
        "epoch": "0",
        "module": null,
        "name": "kernel-core",
        "packager": "Red Hat, Inc. <http://bugzilla.redhat.com/bugzilla>",
        "pgpsig": "RSA/SHA256, Sun 21 Jun 2026 07:49:45 AM EDT, Key ID 199e2f91fd431d51",
        "release": "553.137.1.el8_10",
        "repository": "beaker-BaseOS",
        "stream": null,
        "version": "4.18.0"
    },
    "type": "ordinary",
    "uname_r": "4.18.0-553.137.1.el8_10.aarch64"
}
[root@hpe-apollo-cn99xx-14-vm-04 ~]#  grep 'Detected kernel version inside container\|Building initram disk for kernel' /var/log/leapp/leapp-upgrade.log
2026-06-24 11:23:04.792 DEBUG    PID: 155674 leapp.workflow.InterimPreparation.upgrade_initramfs_generator: Detected kernel version inside container: 5.14.0-687.5.1.el9_8.aarch64+64k.
2026-06-24 11:23:05.12  DEBUG    PID: 155674 leapp.workflow.InterimPreparation.upgrade_initramfs_generator: Building initram disk for kernel: 5.14.0-687.5.1.el9_8.aarch64+64k
2026-06-24 11:24:23.70  DEBUG    PID: 155674 leapp.workflow.InterimPreparation.upgrade_initramfs_generator: Building initram disk for kernel: 5.14.0-687.5.1.el9_8.aarch64+64k finished
[root@hpe-apollo-cn99xx-14-vm-04 ~]# leapp-inspector messages --type InstalledTargetKernelInfo
                          PRODUCED MESSAGES
Stamp: 2026-06-24T15:41:48.697967Z
Actor: scan_installed_target_kernel_version
Phase: RPMUpgrade
Type: InstalledTargetKernelInfo
Message_data:
{
    "initramfs_path": "/boot/initramfs-5.14.0-687.5.1.el9_8.aarch64+64k.img",
    "kernel_img_path": "/boot/vmlinuz-5.14.0-687.5.1.el9_8.aarch64+64k",
    "pkg_nevra": "kernel-64k-core-5.14.0-687.5.1.el9_8.aarch64",
    "uname_r": "5.14.0-687.5.1.el9_8.aarch64+64k"
}
```

- RHEL9 on aarch64 with kernel-64k

```bash
[root@nvidia-grace-grace-07-vm-21 ~]# leapp-inspector messages --type KernelInfo
                          PRODUCED MESSAGES
Stamp: 2026-06-24T20:14:27.629529Z
Actor: scan_source_kernel
Phase: FactsCollection
Type: KernelInfo
Message_data:
{
    "page_size": "64k",
    "pkg": {
        "arch": "aarch64",
        "epoch": "0",
        "module": null,
        "name": "kernel-64k-core",
        "packager": "Red Hat, Inc. <http://bugzilla.redhat.com/bugzilla>",
        "pgpsig": "RSA/SHA256, Wed 24 Jun 2026 01:36:27 AM EDT, Key ID 199e2f91fd431d51",
        "release": "717.el9",
        "repository": "rhel-9-for-aarch64-baseos-rpms",
        "stream": null,
        "version": "5.14.0"
    },
    "type": "ordinary",
    "uname_r": "5.14.0-717.el9.aarch64+64k"
}
[root@nvidia-grace-grace-07-vm-21 ~]#  grep 'Detected kernel version inside container\|Building initram disk for kernel' /var/log/leapp/leapp-upgrade.log
2026-06-24 16:16:42.173 DEBUG    PID: 43896 leapp.workflow.InterimPreparation.upgrade_initramfs_generator: Detected kernel version inside container: 6.12.0-211.7.1.el10_2.aarch64+64k.
2026-06-24 16:16:42.236 DEBUG    PID: 43896 leapp.workflow.InterimPreparation.upgrade_initramfs_generator: Building initram disk for kernel: 6.12.0-211.7.1.el10_2.aarch64+64k
2026-06-24 16:17:06.420 DEBUG    PID: 43896 leapp.workflow.InterimPreparation.upgrade_initramfs_generator: Building initram disk for kernel: 6.12.0-211.7.1.el10_2.aarch64+64k finished
[root@nvidia-grace-grace-07-vm-21 ~]# leapp-inspector messages --type InstalledTargetKernelInfo
                          PRODUCED MESSAGES
Stamp: 2026-06-24T16:51:57.855843Z
Actor: scan_installed_target_kernel_version
Phase: RPMUpgrade
Type: InstalledTargetKernelInfo
Message_data:
{
    "initramfs_path": "/boot/initramfs-6.12.0-211.7.1.el10_2.aarch64+64k.img",
    "kernel_img_path": "/boot/vmlinuz-6.12.0-211.7.1.el10_2.aarch64+64k",
    "pkg_nevra": "kernel-64k-core-6.12.0-211.7.1.el10_2.aarch64",
    "uname_r": "6.12.0-211.7.1.el10_2.aarch64+64k"
}
```

pirat89 (approved):
lgtm now and it seems to work!

matejmatuska (approved):
LGTM now!

PR #1551, opened 2026-06-09
Add scandnfcomps actor to scan installed DNF comps

pirat89:
The actor scan installed DNF environments and groups (DNF comps)
which are represented by new InstalledDNFComps model.

The scan is performed using initialized dnf.Base to load all
available groups and envs and then it checks what is installed based
on the history.

About each group and environment, only ID and and (pretty) name
are collected. Skipping the scan of other data as we do not have a
use for that at this moment.

Unit-tests and part of docstrings are assisted by AI. All the other
code had to be manually updated unfortunately.

Related to jira: RHEL-148697


pirat89:
Closing in favor of #1506 - let's apply all changes there directly.

PR #1554, opened 2026-06-15
Add machine-id validation actors for in-place upgrade

prjagtap-rh:
Introduced a scanner/checker actor pair that verifies /etc/machine-id exists and contains a valid 32-character hexadecimal identifier. The upgrade is inhibited if the file is missing or malformed, as a valid machine-id is required by systemd and the target userspace container creation.

Jira: RHEL-50161

### DIFF
--- a/repos/system_upgrade/common/actors/checkmachineid/actor.py
+++ b/repos/system_upgrade/common/actors/checkmachineid/actor.py
@@ -1,0 +1,23 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import checkmachineid
+from leapp.models import MachineIdInfo
+from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckMachineId(Actor):
+    """
+    Check that /etc/machine-id contains a valid machine ID.
+
+    The /etc/machine-id file must contain exactly 32 lowercase hexadecimal
+    characters that may not be all zeros. If the file is missing or contains an
+    invalid ID, the system is in an unsupported state and the upgrade is inhibited.
+    """
+
+    name = 'check_machine_id'
+    consumes = (MachineIdInfo,)
+    produces = (Report,)
+    tags = (IPUWorkflowTag, ChecksPhaseTag)
+
+    def process(self):
+        checkmachineid.process()
--- a/repos/system_upgrade/common/actors/checkmachineid/libraries/checkmachineid.py
+++ b/repos/system_upgrade/common/actors/checkmachineid/libraries/checkmachineid.py
@@ -1,0 +1,46 @@
+import re
+
+from leapp import reporting
+from leapp.libraries.stdlib import api
+from leapp.models import MachineIdInfo
+
+_VALID_MACHINE_ID_RE = re.compile(r'^[0-9a-f]{32}$')
+_ALL_ZEROS = '0' * 32
+
+
+def process():
+    machine_id_info = next(api.consume(MachineIdInfo), None)
+    if not machine_id_info:
+        api.current_logger().warning(
+            'The MachineIdInfo message is missing. Skipping the check of /etc/machine-id.'
+        )
+        return
+
+    if (machine_id_info.machine_id
+            and _VALID_MACHINE_ID_RE.match(machine_id_info.machine_id)
+            and machine_id_info.machine_id != _ALL_ZEROS):
+        return
+
+    reporting.create_report([
+        reporting.Title('Missing or invalid /etc/machine-id file'),
+        reporting.Summary(
+            'The /etc/machine-id file is missing or does not contain a valid'
+            ' machine ID. A valid machine ID is a 32-character lowercase'
+            ' hexadecimal string that is not all zeros. Various system'
+            ' components rely on a valid machine ID and the upgrade cannot'
+            ' proceed without one.'
+        ),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.SANITY, reporting.Groups.INHIBITOR]),
+        reporting.Remediation(
+            hint=(
+                'Generate a valid machine ID by running `systemd-machine-id-setup`.'
+                ' See `man machine-id` and `man systemd-machine-id-setup` for details.'
+            )
+        ),
+        reporting.ExternalLink(
+            url='https://access.redhat.com/solutions/3600401',
+            title='How to reconfigure the machine-id?'
+        ),
+        reporting.RelatedResource('file', '/etc/machine-id'),
+    ])
--- a/repos/system_upgrade/common/actors/checkmachineid/tests/test_checkmachineid.py
+++ b/repos/system_upgrade/common/actors/checkmachineid/tests/test_checkmachineid.py
@@ -1,0 +1,39 @@
+import pytest
+
+from leapp import reporting
+from leapp.libraries.actor import checkmachineid
+from leapp.libraries.common.testutils import create_report_mocked, CurrentActorMocked
+from leapp.libraries.stdlib import api
+from leapp.models import MachineIdInfo
+from leapp.utils.report import is_inhibitor
+
+_VALID_MACHINE_ID = 'a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4'
+
+
+@pytest.mark.parametrize('machine_id,should_inhibit', [
+    (None, True),
+    ('', True),
+    ('abc123', True),
+    ('a' * 33, True),
+    ('z' * 32, True),
+    ('A1B2C3D4E5F6A1B2C3D4E5F6A1B2C3D4', True),
+    ('0' * 32, True),
+    ('a' * 31 + ' ', True),
+    (_VALID_MACHINE_ID, False),
+])
+def test_check_machine_id(monkeypatch, machine_id, should_inhibit):
+    monkeypatch.setattr(reporting, 'create_report', create_report_mocked())
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(
+        msgs=[MachineIdInfo(machine_id=machine_id)]
+    ))
+    checkmachineid.process()
+    assert bool(reporting.create_report.called) == should_inhibit
+    if should_inhibit:
+        assert is_inhibitor(reporting.create_report.report_fields)
+
+
+def test_check_machine_id_missing_message(monkeypatch):
+    monkeypatch.setattr(reporting, 'create_report', create_report_mocked())
+    monkeypatch.setattr(api, 'current_actor', CurrentActorMocked(msgs=[]))
+    checkmachineid.process()
+    assert not reporting.create_report.called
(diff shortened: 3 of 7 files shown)


repos/system_upgrade/common/actors/checkmachineid/libraries/checkmachineid.py
```diff
@@ -0,0 +1,42 @@
...
+def process():
+    machine_id_info = next(api.consume(MachineIdInfo), None)
+    if not machine_id_info:
+        api.current_logger().warning(
+            'The MachineIdInfo message is missing. Skipping the check of /etc/machine-id.'
+        )
+        return
+
+    if machine_id_info.machine_id and _VALID_MACHINE_ID_RE.match(machine_id_info.machine_id):
+        return
```
matejmatuska:
```suggestion
    if _VALID_MACHINE_ID_RE.match(machine_id_info.machine_id):
        return
```
If `machine_id_info.machine_id` is `None` and therefore evaluates to False ( `bool(None) == False`), the inhibitor report should still be created.

It doesn't really matter if the file is missing or couldn't be read because of some other error, if leapp couldn't read it, there is a good chance no one else can.

The report doesn't need to differentiate between a missing and invalid file, we just need it to be there and be in the current format.

matejmatuska: Note that this is not a full review, just something that is IMO 100% a blocker.

repos/system_upgrade/common/actors/checkmachineid/tests/test_checkmachineid.py
```diff
@@ -0,0 +1,36 @@
...
+
+_VALID_MACHINE_ID = 'a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4'
+
+
+@pytest.mark.parametrize('machine_id,should_inhibit', [
+    (None, True),
+    ('', True),
+    ('abc123', True),
+    ('a' * 33, True),
+    ('z' * 32, True),
```
karolinku, line 18: nit: we could test also case with uppercase (which is specifically forbidden by machine-id spec)

repos/system_upgrade/common/actors/scanmachineid/libraries/scanmachineid.py
```diff
@@ -0,0 +1,14 @@
...
+_MACHINE_ID_PATH = '/etc/machine-id'
+
+
+def process():
+    machine_id = None
+    try:
+        with open(_MACHINE_ID_PATH, 'r') as f:
+            machine_id = f.read().strip()
+    except OSError:
+        pass
```
karolinku: This is making the error silent. What about add simple warn/log that the file failed to be opened?

matejmatuska (dismissed):

repos/system_upgrade/common/actors/scanmachineid/tests/test_scanmachineid.py
```diff
@@ -0,0 +1,61 @@
...
+
+_VALID_MACHINE_ID = 'a1b2c3d4e5f6a1b2c3d4e5f6a1b2c3d4'
+
+
+class _MockOpen:
+    def __init__(self, content):
+        self._content = content
+
+    def __call__(self, *args, **kwargs):
+        return StringIO(self._content)
```
karolinku: Are there any reasons for using _MockOpen instead of monkeypatch?

repos/system_upgrade/common/actors/checkmachineid/libraries/checkmachineid.py
karolinku: agree, this was my concern too.

prjagtap-rh: @matejmatuska 
Agreed on the intent: None, missing, unreadable, and invalid should all result in the inhibitor. The current code already does that.

However, we can't simplify to just:

if _VALID_MACHINE_ID_RE.match(machine_id_info.machine_id):
because re.match(None) raises a TypeError, it doesn't return None/falsy:

repos/system_upgrade/common/actors/scanmachineid/tests/test_scanmachineid.py
prjagtap-rh: There was no specific reason as I had generated this test case using AI Skill and was testing same and seems it pulled MockOpen instead of monkeypatch

repos/system_upgrade/common/actors/scanmachineid/libraries/scanmachineid.py
prjagtap-rh: fixed

repos/system_upgrade/common/actors/checkmachineid/tests/test_checkmachineid.py
prjagtap-rh: fixed

repos/system_upgrade/common/actors/checkmachineid/libraries/checkmachineid.py
matejmatuska: I think the code is right, I must have misunderstood (it was midnight).
Sorry for the noise.

karolinku (dismissed):
lgtm now, however I tested it only on RHEL 8.

PeterMocary (changes requested):
I didn't test it, but note that the [man page](https://www.man7.org/linux/man-pages/man5/machine-id.5.html) states:

> The machine ID is a single newline-terminated, hexadecimal, 32-character,
> lowercase ID. When decoded from hexadecimal, this corresponds to a
> 16-byte/128-bit value. This ID may not be all zeros.

So my interpretation is that we need to check:
- single line
- **newline-terminated**
- hexadecimal characters
- 32 characters
- lowercase
- **not all zeros**

Since the content of `/etc/machine-id` is stripped and then passed to the check actor, we could possibly have a valid id terminated by a space and it would escape the inhibitor. Unless I'm missing something. Similarly, an ID with zeros only escapes the inhibitor while being invalid according to the man page.

I also see this:

> The machine ID may be set, for example when network booting, with
> the `systemd.machine_id=` kernel command line parameter or by
> passing the option `--machine-id=` to systemd. An ID specified in
> this manner has higher priority and will be used instead of the ID
> stored in `/etc/machine-id`.

However, it seems like it is always transfered into the file itself based on the initialization section of the same man page:
> After the machine ID is established, systemd will attempt to save it to /etc/machine-id. 

It might be worth checking how this really behaves, if you didn't already.




I'd fix the validation to match the man page, thus I'm requesting changes. I think the other ways of setting the machine id would not make a difference, right?

prjagtap-rh:
Thanks @PeterMocary  for thorough review. You're right on both counts  I have updated the validation to match the man page specification.

1. Changed .strip() to .rstrip('\n') in the scanner — now only the expected trailing newline is removed, so whitespace like a trailing space won't be silently hidden.
2. Added an explicit all-zeros check (!= '0' * 32) in the checker.
3. Updated the actor docstring and report summary to mention the all-zeros constraint.
4. Added tests for both cases.

Regarding systemd.machine_id= kernel parameter
I looked into this. The man page confirms that regardless of how the ID is initially set (kernel cmdline, --machine-id= option), systemd always writes it back to /etc/machine-id during initialization. So reading from the file is sufficient no kernel cmdline check needed.

repos/system_upgrade/common/actors/checkmachineid/libraries/checkmachineid.py
```diff
@@ -0,0 +1,46 @@
...
+
+    reporting.create_report([
+        reporting.Title('Missing or invalid /etc/machine-id file'),
+        reporting.Summary(
+            'The /etc/machine-id file is missing or does not contain a valid'
+            ' machine ID. A valid machine-id is a 32-character lowercase'
+            ' hexadecimal string that is not all zeros. It is required by'
+            ' systemd and is used during the upgrade to create the target'
+            ' userspace container.'
```
PeterMocary: We shouldn't include implementation details in the summary I think. The user does not need to know about the container at all. We should make the summary bit more high level.

```suggestion
            'The /etc/machine-id file is missing or does not contain a valid'
            ' machine ID. A valid machine-id is a 32-character lowercase'
            ' hexadecimal string that is not all zeros. It is'
            ' a fundamental system identifier that many core components'
            ' depend on. Performing an in-place upgrade with a malformed'
            ' or missing machine-id would lead to a failure.'
```
Note: the formatting might be off.

I'm not really happy with this formulation either. Do you have some better suggestion? @matejmatuska @karolinku @prjagtap-rh

repos/system_upgrade/common/actors/checkmachineid/actor.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class CheckMachineId(Actor):
+    """
+    Check that /etc/machine-id contains a valid machine ID.
+
+    The /etc/machine-id file must contain exactly 32 lowercase hexadecimal
+    characters followed by a newline (33 bytes total). The ID may not be
+    all zeros. If the file is missing or contains an invalid ID, the system
+    is in an unsupported state and the upgrade is inhibited.
```
PeterMocary: Nit: I'd not include the newline here as it is stripped during the scanning. Also I've tried to truncate it from the file and the preupgrade actually didn't fail.
```suggestion
    The /etc/machine-id file must contain exactly 32 lowercase hexadecimal
    characters that may not be all zeros. If the file is missing or contains an 
    invalid ID, the system is in an unsupported state and the upgrade is inhibited.
```

PeterMocary (commented):
I tested it on RHEL 8 with different configurations. Found out that the newline is not actually enforced as the man page implies (at least in the particular case that caused the upgrade to fail here).

So it LGTM, I just added some reformulation of the summary as I think the technical detail about upgrade itself is not really necessary to mention.

repos/system_upgrade/common/actors/checkmachineid/libraries/checkmachineid.py
prjagtap-rh: @PeterMocary 

        'The /etc/machine-id file is missing or does not contain a valid'
        ' machine ID. This is a fundamental system identifier that many'
        ' core components depend on. Performing an in-place upgrade with'
        ' a malformed or missing machine-id could lead to failures in'
        ' services that rely on it.'

PeterMocary: I don't want to complicate it with services as it might be misunderstood as systemd services. The machine id is required by systemd, journald but also it is tied to the dbus. On top of that, there is no list of things that depend on it.

Let's go with something simple. The following should be straight to the point while also providing valid reasoning behind the inhibitor.
```
  The /etc/machine-id file is missing or does not contain a valid
  machine ID. A valid machine ID is a 32-character lowercase
  hexadecimal string that is not all zeros. Various system
  components rely on a valid machine ID and the upgrade cannot
  proceed without one.
```

PeterMocary (approved):
LGTM! We can merge when tests finish.

PR #1560, opened 2026-06-24
Add rpmtoflatpak actor to help with RPMs -> Flatpak migrations

tpopela:
Starting with RHEL 10.2, we've switched the default delivery method for Firefox and Thunderbird from RPMs to Flatpaks (see RHEL-139533). To accommodate for that switch for upgrades, we check whether the firefox or thunderbird packages are preinstalled and if so, then we install the packages with flatpak-preinstall configuration files and run flatpak preinstall on it (the same workflow that Anaconda is doing).

We also limit the actor to work on subscribed systems, so we don't have to deal with credentials for the Red Hat Flatpak Registry (see RHEL-85004 for more information).

Developed by Claude Code.

Resolves: RHEL-53806
Resolves: RHEL-67925

### DIFF
--- a/repos/system_upgrade/el9toel10/actors/rpmtoflatpak/checkreportflatpak/actor.py
+++ b/repos/system_upgrade/el9toel10/actors/rpmtoflatpak/checkreportflatpak/actor.py
@@ -1,0 +1,28 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import checkreportflatpak
+from leapp.models import RpmToFlatpakFacts, RpmTransactionTasks
+from leapp.reporting import Report
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckReportFlatpak(Actor):
+    """
+    Report RPM-to-Flatpak migrations and schedule required packages for installation.
+
+    Consumes facts produced by ScanRpmToFlatpak. When RHSM is in use
+    (the default), produces RpmTransactionTasks to install the
+    redhat-flatpak-preinstall-* packages and flatpak itself as part of the
+    main upgrade transaction, and creates an informational report.
+
+    When RHSM is not in use (e.g. Satellite, RHUI, ISO), creates an
+    inhibitor report directing the user to install the affected applications
+    as Flatpaks manually after the upgrade.
+    """
+
+    name = 'check_report_flatpak'
+    consumes = (RpmToFlatpakFacts,)
+    produces = (Report, RpmTransactionTasks)
+    tags = (IPUWorkflowTag, ChecksPhaseTag)
+
+    def process(self):
+        checkreportflatpak.process()
--- a/repos/system_upgrade/el9toel10/actors/rpmtoflatpak/checkreportflatpak/libraries/checkreportflatpak.py
+++ b/repos/system_upgrade/el9toel10/actors/rpmtoflatpak/checkreportflatpak/libraries/checkreportflatpak.py
@@ -1,0 +1,84 @@
+from leapp import reporting
+from leapp.libraries.common.rhsm import skip_rhsm
+from leapp.libraries.stdlib import api
+from leapp.models import RpmToFlatpakFacts, RpmTransactionTasks
+
+_FLATPAK_PKG = 'flatpak'
+_FLATPAK_DOCS_URL = (
+    'https://docs.redhat.com/documentation/red_hat_enterprise_linux/10/html'
+    '/administering_rhel_by_using_the_gnome_desktop_environment/installing-applications-by-using-flatpak'
+)
+
+
+def _report_inhibitor(rpm_names):
+    pkg_list = ', '.join(rpm_names)
+
+    summary = (
+        'The following packages are installed as RPMs but are no longer shipped as RPMs '
+        'on the target system: {packages}. These packages are available as Flatpaks '
+        'on RHEL 10.\n\n'
+        'Automatic RPM-to-Flatpak migration is currently only supported on systems '
+        'using Red Hat Subscription Manager (RHSM). This system is not using RHSM '
+        '(LEAPP_NO_RHSM=1 is set).\n\n'
+        'After the upgrade, install the applications manually as Flatpaks.'
+    ).format(packages=pkg_list)
+
+    reporting.create_report([
+        reporting.Title(
+            'RPM-to-Flatpak migration is not supported without RHSM'
+        ),
+        reporting.Summary(summary),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.INHIBITOR, reporting.Groups.DESKTOP]),
+        reporting.Remediation(
+            hint='After the upgrade, install the applications as Flatpaks manually.'
+        ),
+        reporting.ExternalLink(
+            url=_FLATPAK_DOCS_URL,
+            title='Installing applications by using Flatpak'
+        ),
+    ])
+
+
+def _report_migration(rpm_names):
+    pkg_list = ', '.join(rpm_names)
+
+    summary = (
+        'The following packages are installed as RPMs but are no longer shipped as RPMs '
+        'on the target system. They will be migrated to their Flatpak equivalents '
+        'during the upgrade: {packages}.\n\n'
+        'The migration is performed by installing the corresponding '
+        'redhat-flatpak-preinstall-* packages and the flatpak package itself on the '
+        'target system as part of the upgrade transaction, then running '
+        '`flatpak preinstall`.'
+    ).format(packages=pkg_list)
+
+    reporting.create_report([
+        reporting.Title(
+            'RPM packages will be migrated to Flatpak'
+        ),
+        reporting.Summary(summary),
+        reporting.Severity(reporting.Severity.INFO),
+        reporting.Groups([reporting.Groups.DESKTOP]),
+        reporting.ExternalLink(
+            url=_FLATPAK_DOCS_URL,
+            title='Installing applications by using Flatpak'
+        ),
+    ])
+
+
+def process():
+    facts = next(api.consume(RpmToFlatpakFacts), None)
+    if not facts or not facts.packages:
+        return
+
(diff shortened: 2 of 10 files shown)


repos/system_upgrade/el9toel10/actors/rpmtoflatpak/scanrpmtoflatpak/libraries/scanrpmtoflatpak.py
```diff
@@ -0,0 +1,50 @@
+from leapp.libraries.common.rhsm import skip_rhsm
+from leapp.libraries.common.rpms import has_package
+from leapp.libraries.stdlib import api
+from leapp.models import DistributionSignedRPM, FlatpakMigrationPackage, RHSMInfo, RpmToFlatpakFacts
+
+# Maps RPM package names to their corresponding flatpak-preinstall package names.
+# The preinstall package triggers `flatpak preinstall` via RPM scriptlets on the
```
matejmatuska: Maybe I am missing something, but I installed both of the packages on a RHEL 10.2 system and there are no scriptlets reported by `rpm -q --scripts` and the flatpaks are not installed.

Furthermore, the migraterpmtoflatpak actor calls the command explicitly so I presume the comment is incorrect.

repos/system_upgrade/el9toel10/actors/rpmtoflatpak/checkreportflatpak/libraries/checkreportflatpak.py
```diff
@@ -0,0 +1,30 @@
...
+        'on the target system. They will be migrated to their Flatpak equivalents '
+        'during the upgrade: {packages}.\n\n'
+        'The migration is performed by installing the corresponding '
+        'redhat-flatpak-preinstall-* packages on the target system and running '
+        '`flatpak preinstall`.'
+    ).format(packages=pkg_list)
+
+    reporting.create_report([
+        reporting.Title(
+            'RPM packages will be migrated to Flatpak: {}'.format(pkg_list)
```
matejmatuska: Do not use formatting in the title, just a plain string. The reports have hashes computed by the framework and title is included in that computation.

repos/system_upgrade/el9toel10/actors/rpmtoflatpak/checkreportflatpak/libraries/checkreportflatpak.py
```diff
@@ -0,0 +1,30 @@
...
+        ),
+        reporting.Summary(summary),
+        reporting.Severity(reporting.Severity.INFO),
+        reporting.Groups([reporting.Groups.SERVICES]),
```
matejmatuska: reporting.Groups.DESKTOP also exists, but I am not sure if only desktop apps are subject to this.

repos/system_upgrade/el9toel10/actors/rpmtoflatpak/checkreportflatpak/actor.py
```diff
@@ -0,0 +1,21 @@
...
+    """
+    Report RPM-to-Flatpak migrations that will be performed during upgrade.
+
+    Consumes facts produced by ScanRpmToFlatpak and creates a report entry
+    informing the user which packages will be migrated from RPM to Flatpak.
+    """
+
+    name = 'check_report_flatpak'
+    consumes = (RpmToFlatpakFacts,)
+    produces = ()
```
matejmatuska: Report should be included in the produces if the actor produces a report (using reporting.create_report also counts):
```suggestion
    produces = (Report,)
```

repos/system_upgrade/el9toel10/actors/rpmtoflatpak/migraterpmtoflatpak/actor.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class MigrateRpmToFlatpak(Actor):
+    """
+    Migrate RPM packages to their Flatpak equivalents on the upgraded system.
+
+    Installs the redhat-flatpak-preinstall-* packages corresponding to each
+    RPM package that was detected by ScanRpmToFlatpak. Installing those packages
+    triggers `flatpak preinstall` via RPM scriptlets, which pulls in the Flatpak
+    application from the configured remote.
```
matejmatuska: again this does not seem to be the case

repos/system_upgrade/el9toel10/actors/rpmtoflatpak/migraterpmtoflatpak/libraries/migraterpmtoflatpak.py
```diff
@@ -0,0 +1,43 @@
+from leapp.libraries.stdlib import api, CalledProcessError, run
+from leapp.models import RpmToFlatpakFacts
+
+
+def _install_preinstall_packages(preinstall_pkgs):
+    cmd = ['dnf', 'install', '-y'] + preinstall_pkgs
```
matejmatuska: While I understand the intent, this is not allowed during the upgrade and wouldn't work, the correct way would be to request a package to be installed by producing RpmTransactionTasks message:
```
RpmTransactionTasks(to_install=<package>)
```
this includes the packages in the main upgrade transaction which is performed using dnf with carefully setup configuration/parameters. Actors in the ApplicationsPhase, such as this one, can then do further actions, such as call the `flatpak preinstall` command.

repos/system_upgrade/el9toel10/actors/rpmtoflatpak/migraterpmtoflatpak/libraries/migraterpmtoflatpak.py
```diff
@@ -0,0 +1,43 @@
...
+                ', '.join(preinstall_pkgs), e
+            )
+        )
+        return False
+    return True
+
+
+def _run_flatpak_preinstall():
+    try:
+        run(['flatpak', 'preinstall', '--system', '--noninteractive'])
```
matejmatuska, line 7: Maybe the scan actor should also ensure that flatpak is installed on the source system? 

Edit: It seems like it's an indirect dependency of the preinstall packages, either way I think it might make sense to request it's installation explicitly and inform the users about it in the report (I think the report you already have can cover it, no need to create a new one).

repos/system_upgrade/el9toel10/models/rpmtoflatpak.py
matejmatuska, line 1: We generally expect Model subclasses to have docstring, same for their fields.

repos/system_upgrade/el9toel10/actors/rpmtoflatpak/scanrpmtoflatpak/libraries/scanrpmtoflatpak.py
```diff
@@ -0,0 +1,50 @@
...
+                    rpm_name, preinstall_pkg
+                )
+            )
+    return packages
+
+
+def _has_active_subscription():
+    """Return True if RHSM is skipped (e.g. Satellite) or the system is registered."""
+    if skip_rhsm():
+        return True
```
matejmatuska: What the docstring says is not the case. What `skip_rhsm()` (triggered e.g. by `--skip-rhsm`) does, is exactly what is says, it tells leapp to not use RHSM. However, that doesn't mean the system is registered.
When RHSM is not used, there are various source the target content can be taken from, including Satellite, but also e.g. ISO, local mirrors or RHUI in the clouds, then there is also "custom repositories" where you can put any repositories in a repofile and leapp will use content from there. 

I am unsure in which of these environments the authentication mechanism proposed in RHEL-85004 works, but I think there is a good chance it's not all.

Therefore I think it might be a good idea to do a more extensive check of whether the machine is authenticated and if not, report that it has to be (even if using the podman login command) for the migration to work.
Is there any reliable way this can be done?

matejmatuska (changes requested):
There are some hard blockers that have to be resolved.

repos/system_upgrade/el9toel10/actors/rpmtoflatpak/scanrpmtoflatpak/libraries/scanrpmtoflatpak.py
tpopela: Yeah, Claude made that up as the process is completely different.

repos/system_upgrade/el9toel10/actors/rpmtoflatpak/checkreportflatpak/libraries/checkreportflatpak.py
tpopela: Flatpak are containerized desktop apps, so DESKTOP seems like a good fit.

repos/system_upgrade/el9toel10/actors/rpmtoflatpak/migraterpmtoflatpak/libraries/migraterpmtoflatpak.py
tpopela: Ack

repos/system_upgrade/el9toel10/actors/rpmtoflatpak/migraterpmtoflatpak/libraries/migraterpmtoflatpak.py
tpopela: Ack

repos/system_upgrade/el9toel10/actors/rpmtoflatpak/scanrpmtoflatpak/libraries/scanrpmtoflatpak.py
tpopela: We've talked about this with Matej on Slack and for now we've agreed on checking that `skip_rhsm == False` which means that the user didn't enforced skipping the RHSM. And in other situations we will point them at existing Flatpak documentation.

tpopela-rh:
This is moved to https://github.com/oamg/leapp-repository/pull/1600 (moving from my personal GitHub account to my work one).

PR #1561, opened 2026-07-01
Add basic Rocky Linux support for upgrades and conversions

karolinku:
Define Rocky Linux as a new distro: add 'rocky' in DistroIDs enum, add GPG signing keys and signatures, define upgrade paths (8->9, 9->10), map distro repofiles for versions 8/9/10, add distro package swap/remove rules for Rocky-to-RHEL conversions, add display name mappings, remove symlinks related to rhel/rocky packages.

Jira: RHEL-187004


repos/system_upgrade/common/actors/convert/updateefi/libraries/updateefi.py
matejmatuska: Another small typo:
```python
  "Didn't remove the source EFI directory {}, it does not exist".format(
```
should be:
```python
  "Coudln't remove the source EFI directory {}, it's not empty".format(
```

matejmatuska (commented):
Found one more small thing, other than that we can merge.

matejmatuska (dismissed):
LGTM

matejmatuska (approved):
Let's wait for the test and we can merge.

PR #1562, opened 2026-07-06
fix(ceph/luks): Match Ceph/LUKS volumes UUIDs

holser:
The LUKS inhibitor incorrectly compared device names UUIDs against
Ceph LV UUIDs, so encrypted Ceph OSDs were never recognized and always
triggered the upgrade inhibitor. Resolve by querying the actual LUKS UUID
via blkid for each encrypted Ceph OSD and comparing that against the
LUKS dumped UUIDs.

Resolves: rhbz#2264543

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/repos/system_upgrade/common/actors/cephvolumescan/libraries/cephvolumescan.py
+++ b/repos/system_upgrade/common/actors/cephvolumescan/libraries/cephvolumescan.py
@@ -57,6 +57,16 @@ def get_ceph_lvm_list():
     return json_output
 
 
+def _get_luks_uuid(lv_path):
+    try:
+        output = run(['blkid', '-s', 'UUID', '-o', 'value', lv_path])
+    except CalledProcessError:
+        api.current_logger().warning('Failed to get LUKS UUID for %s', lv_path)
+        return None
+    uuid = output['stdout'].strip()
+    return uuid if uuid else None
+
+
 def encrypted_osds_list():
     result = []
     if os.path.isfile(CEPH_CONF):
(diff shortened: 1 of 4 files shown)


holser:
Review please @oamg/developers

xbezdick:
I think I don't like the solution with - to --. Instead we should collect blkid uiid of the ceph luks device and compare  if luks_dump.uuid in ceph_vol instead of if luks_dump.device_name in ceph_vol:

repos/system_upgrade/common/actors/cephvolumescan/libraries/cephvolumescan.py
```diff
@@ -57,14 +57,29 @@ def get_ceph_lvm_list():
...
     result = []
     if os.path.isfile(CEPH_CONF):
         output = get_ceph_lvm_list()
         if output is not None:
             try:
                 for key in output:
-                    result.extend([element['lv_uuid'] for element in output[key] if element['tags']['ceph.encrypted']])
+                    for element in output[key]:
+                        if element['tags']['ceph.encrypted']:
+                            lv_path = '/dev/{}/{}'.format(element['vg_name'], element['lv_name'])
```
holser, line 79: sometimes we don't have "lv_path" in json generated by  `ceph-volume lvm list --format json` so we construct it anyway unconditionally

Though we could do 
```
if element['lv_path']:
    lv_path = '{}'.format(element['lv_path'])
else:
   lv_path = '/dev/{}/{}'.format(element['vg_name'], element['lv_name'])
```
but I would prefer lighter code where lv_path is built unconditionally.

holser:
> I think I don't like the solution with - to --. Instead we should collect blkid uiid of the ceph luks device and compare if luks_dump.uuid in ceph_vol instead of if luks_dump.device_name in ceph_vol:

I have reworked the solution using `blkid` approach. Thanks!

PR #1563, opened 2026-07-07
Increase free space check for / from 100MB to 200MB

holser:
The space_guard function checks available disk space before proceeding with the upgrade. The previous default of 100MB proved insufficient in some scenarios, leading to failed upgrades due to lack of disk space. Increase the threshold to 200MB to provide a safer margin for the upgrade process.

BZ#2278831

### DIFF
--- a/repos/system_upgrade/common/libraries/guards.py
+++ b/repos/system_upgrade/common/libraries/guards.py
@@ -43,7 +43,7 @@ def connection_guard(url='https://example.com'):
     return closure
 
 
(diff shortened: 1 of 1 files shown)


pirat89 (commented):
@holser  it seems to me that you are trying to fix non-existing problem. The ticket you refer to is speaking about pretty old leapp-repository version without the redesigned solution based on OVL images. The free space is checked by several other mechanisms since that time (100mib here does not mean 100 mib on the host system). Can you provide valid scenario when this piece of code is actually hit and have an effect?

PR #1564, opened 2026-07-07
fix(ceph/luks): Match Ceph/LUKS volumes UUIDs

holser:
The LUKS inhibitor incorrectly compared device names UUIDs against Ceph LV UUIDs, so encrypted Ceph OSDs were never recognized and always triggered the upgrade inhibitor. Resolve by querying the actual LUKS UUID via blkid for each encrypted Ceph OSD and comparing that against the LUKS dumped UUIDs.

Resolves: rhbz#2264543

### DIFF
--- a/repos/system_upgrade/common/actors/cephvolumescan/libraries/cephvolumescan.py
+++ b/repos/system_upgrade/common/actors/cephvolumescan/libraries/cephvolumescan.py
@@ -57,17 +57,32 @@ def get_ceph_lvm_list():
     return json_output
 
 
+def _get_luks_uuid(lv_path):
+    try:
+        output = run(['blkid', '-s', 'UUID', '-o', 'value', lv_path])
+    except CalledProcessError:
+        api.current_logger().warning('Failed to get LUKS UUID for %s', lv_path)
+        return None
+    uuid = output['stdout'].strip()
+    return uuid if uuid else None
+
+
 def encrypted_osds_list():
     result = []
-    if os.path.isfile(CEPH_CONF):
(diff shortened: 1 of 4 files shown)


repos/system_upgrade/common/actors/cephvolumescan/libraries/cephvolumescan.py
```diff
@@ -57,17 +57,32 @@ def get_ceph_lvm_list():
...
+        return result
+    output = get_ceph_lvm_list()
+    if output is None:
+        return result
+    try:
+        for key in output:
+            for element in output[key]:
+                if not element['tags']['ceph.encrypted']:
+                    continue
+                lv_path = '/dev/{}/{}'.format(element['vg_name'], element['lv_name'])
```
pirat89:
```suggestion
                lv_path = os.path.join('/dev', element['vg_name'], element['lv_name'])
```

holser: sometimes we don't have "lv_path" in json generated by  ceph-volume lvm list --format json so we construct it anyway unconditionally

Though we could do
```
if element['lv_path']:
    lv_path = '{}'.format(element['lv_path'])
else:
   lv_path = os.path.join('/dev', element['vg_name'], element['lv_name'])
```
but I would prefer lighter code where lv_path is built unconditionally.

holser:

pirat89 (approved):
lgtm and based on discussion with @holser  it passes the testing. merging

PR #1565, opened 2026-07-08
mount_unit_generator: Remove pseudo FS mount units from upgrade initramfs

holser:
systemd-fstab-generator converts every `/etc/fstab` entry into a systemd mount unit. MountUnitGenerator then prefixes each unit's Where= path with `/sysroot` before bundling them into the upgrade initramfs. For virtual/pseudo filesystem entries such as hugetlbfs (e.g. `/dev/hugepages2M`, `/dev/hugepages1G`) the corresponding directories may not exist under `/sysroot` (`/sysroot/dev/hugepages2M`, `/sysconfig/dev/hugepages1G`).

As the main reasons for having pseudo FS entries in fstab is security hardening, performance tuning, application-specific mounts. Since these pseudo filesystems have no purpose during the in-place upgrade phase, drop their mount units entirely before any further processing.

Resolves: RHEL-180965


holser:
I mistakenly deleted [PR1547](https://github.com/oamg/leapp-repository/pull/1547). This one is continuation.

MichalHe:
Closing this PR as the same problem has been addressed by: #1569. Thank you for proposing multiple solutions!

PR #1566, opened 2026-07-13
Drop resume kernel cmdline argument from upgrade boot entry

prjagtap-rh:
## The reason for the change

The upgrade initramfs excludes the dracut `resume` module (`-o resume`),
but the `resume=` kernel parameter is still copied into the upgrade boot
entry via `grubby --copy-default`. This causes the system to hang during
reboot while waiting for the resume device to appear. The issue is
reproducible when swap is on software RAID but can affect other
configurations as well.

## What has been changed

Added a new `CheckResumeKernelArg` actor (ChecksPhase) that removes the `resume=`
kernel command line parameter from the upgrade boot entry and restores it
on the target kernel entry after the upgrade.

## How it has been changed

The actor consumes `KernelCmdline`, filters for any `resume=` parameters,
and produces:
- `UpgradeKernelCmdlineArgTasks(to_remove=...)` to strip `resume=` from
  the upgrade boot entry
- `TargetKernelCmdlineArgTasks(to_add=...)` to restore `resume=` on the
  target kernel so hibernation keeps working after the upgrade

New files:
- `repos/system_upgrade/common/actors/checkresume/actor.py`
- `repos/system_upgrade/common/actors/checkresume/libraries/checkresume.py`
- `repos/system_upgrade/common/actors/checkresume/tests/test_checkresume.py`

## How to try/test the PR

1. Unit tests: `ACTOR=check_resume_kernel_arg make dev_test_no_lint` (9 tests covering
   UUID, /dev/mapper, /dev/md, bare key, multiple args, no resume, no cmdline)
2. Container tests: `TEST_CONTAINER=el9 make test_container`
3. Manual: On a system with `resume=` on the kernel cmdline, run leapp
   preupgrade and verify the report contains
   "Kernel boot parameter resume= will be removed during the upgrade"

## Reference to a Jira ticket
RHEL-95495

### DIFF
--- a/repos/system_upgrade/common/actors/checkresumekernelarg/libraries/checkresumekernelarg.py
+++ b/repos/system_upgrade/common/actors/checkresumekernelarg/libraries/checkresumekernelarg.py
@@ -1,0 +1,35 @@
+from leapp.libraries.stdlib import api
+from leapp.models import KernelCmdline, TargetKernelCmdlineArgTasks, UpgradeKernelCmdlineArgTasks
+
+
+def process():
+    """
+    Remove resume from the upgrade boot entry and restore it on the target.
+
+    The upgrade initramfs omits the dracut resume module, so the resume device
+    cannot be resolved during the upgrade boot. This can cause the system to
+    hang waiting for the device. Stripping resume argument from the upgrade kernel
+    command line avoids the hang. The value is added back to the target kernel
+    entry via TargetKernelCmdlineArgTasks so hibernation continues working
+    after the upgrade.
+    """
+    cmdline = next(api.consume(KernelCmdline), None)
+    if not cmdline:
+        api.current_logger().debug('No KernelCmdline message received, nothing to do.')
+        return
+
+    resume_args = [arg for arg in cmdline.parameters if arg.key == 'resume']
+    if not resume_args:
+        api.current_logger().debug('No resume argument found on the kernel command line.')
+        return
+
+    api.current_logger().info(
+        'Requesting removal of resume argument from the upgrade kernel command line: %s',
+        ['{}={}'.format(a.key, a.value or '') for a in resume_args]
+    )
+
+    api.produce(UpgradeKernelCmdlineArgTasks(to_remove=resume_args))
+    # When installing the target kernel RPM, the cmdline is copied from the booted system.
+    # As we remove the resume= args, we would accidentally remove them also from the
+    # target entry. Therefore, we add them back in here.
+    api.produce(TargetKernelCmdlineArgTasks(to_add=resume_args))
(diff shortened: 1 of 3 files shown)


repos/system_upgrade/common/actors/checkresume/actor.py
```diff
@@ -0,0 +1,25 @@
...
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckResume(Actor):
+    """
+    Remove the resume argument from the upgrade boot entry.
+
+    The resume argument points to a swap device used for hibernation. During the
+    upgrade the dracut resume module is excluded, so the resume device cannot be
+    resolved and the system can hang during reboot.
```
PeterMocary: The summary in this doc string only states the reasoning and does not include the solution. In my opinion, it should also include at the end something like the following sentence.

 "Therefore, we remove the resume kernel command line argument to prevent possible hang during the reboot."

repos/system_upgrade/common/actors/checkresume/actor.py
```diff
@@ -0,0 +1,25 @@
+from leapp.actors import Actor
+from leapp.libraries.actor import checkresume
+from leapp.models import KernelCmdline, TargetKernelCmdlineArgTasks, UpgradeKernelCmdlineArgTasks
+from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
+
+
+class CheckResume(Actor):
```
PeterMocary: This name doesn't tell much about the actor. I'd change it to `CheckResumeKernelArg` or something like that.

repos/system_upgrade/common/actors/checkresumekernelarg/libraries/checkresumekernelarg.py
```diff
@@ -0,0 +1,35 @@
...
+
+    resume_args = [arg for arg in cmdline.parameters if arg.key == 'resume']
+    if not resume_args:
+        api.current_logger().debug('No resume argument found on the kernel command line.')
+        return
+
+    api.current_logger().info(
+        'Requesting removal of resume argument from the upgrade kernel command line: %s',
+        ['{}={}'.format(a.key, a.value) for a in resume_args]
+    )
```
PeterMocary, line 29: I think that `resume=` (without an argument value) is essentially disabling the functionality. Based on [kernel's hibernate.c](https://raw.githubusercontent.com/torvalds/linux/master/kernel/power/hibernate.c) the `resume=` overwrites the `resume_file` with empty string and essentially stops `find_resume_device` with return `-ENOENT`. 

Removing it doesn't break anything and it should be removed IMO, I just checked it since we log `resume=None` in this log and I was not sure if it is valid. I'd just update this log to output empty string when value is `None` as `resume=None` is not really what is in the kernel command line args.

PeterMocary (commented):
LGTM! Just some nitpicks in comments.

I'll approve after manual testing.

PeterMocary (dismissed):
Tested on RHEL 8 and both `TargetKernelCmdlineArgTasks` and `UpgradeKernelCmdlineArgTasks` are produced correctly leading to removal for upgrade kernel arguments and addition to the target kernel arguments as expected.

PeterMocary:
Finalized the [actor renaming](https://github.com/oamg/leapp-repository/pull/1566#discussion_r3621243500) also in folder, file names, imports and commit message.

PeterMocary (approved):

PR #1569, opened 2026-07-13
mount_unit_generator: Treat pseudo FS mount units as nofail

holser:
systemd-fstab-generator converts every /etc/fstab entry into a systemd mount unit. MountUnitGenerator then prefixes each unit's Where= path with /sysroot before bundling them into the upgrade initramfs. For virtual/pseudo filesystem entries such as hugetlbfs (e.g. /dev/hugepages2M, /dev/hugepages1G) the corresponding directories may not exist under /sysroot.

Instead of removing these units entirely, treat them like nofail: keep the mount units but ensure they only end up in .wants directories (not .requires). This way systemd will attempt to mount the pseudo filesystems during the upgrade initramfs boot, but if a mount fails the boot continues. This is appropriate because the main reasons for having pseudo FS entries in fstab (security hardening, performance tuning, application-specific mounts) do not apply during the upgrade phase.

To make this work, fix_symlinks_in_targets now ensures .wants directories exist for each target that has a .requires directory, since systemd-fstab-generator only creates .wants for fstab entries with the nofail option.

Resolves: RHEL-180965, RHEL-16320

### DIFF
--- a/repos/system_upgrade/common/actors/initramfs/mount_units_generator/libraries/mount_unit_generator.py
+++ b/repos/system_upgrade/common/actors/initramfs/mount_units_generator/libraries/mount_unit_generator.py
@@ -9,6 +9,28 @@ from leapp.models import LiveModeConfig, StorageInfo, TargetUserSpaceInfo, Upgra
 
 BIND_MOUNT_SYSROOT_BOOT_UNIT = 'boot.mount'
 
+# Virtual/pseudo filesystem types that should be treated as nofail during the upgrade initramfs phase.
+# Their mount units are kept but placed only in .wants (not .requires) directories, so systemd
+# will attempt to mount them but continue booting if the mount fails (e.g. mount point does not
+# exist under /sysroot).
+PSEUDO_FS_TYPES = frozenset([
+    'hugetlbfs',
+    'tmpfs',
+    'devtmpfs',
+    'devpts',
+    'sysfs',
+    'proc',
+    'cgroup',
+    'cgroup2',
+    'securityfs',
+    'selinuxfs',
+    'debugfs',
+    'mqueue',
+    'pstore',
+    'efivarfs',
+    'bpf',
+])
+
 
 def run_systemd_fstab_generator(output_directory):
     api.current_logger().debug(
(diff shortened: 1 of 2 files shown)


holser:
It's alternative approach to [PR#1565](https://github.com/oamg/leapp-repository/pull/1565/)

MichalHe:
I have tested the patch on the following setup:
```
[root@leapp-rhel-20260708231442 ~]# cat /etc/fstab

/dev/mapper/rhel_ibm--p8--kvm--03--guest--02-root /                       xfs     defaults        0 0
UUID=3148e508-5d3b-489c-ba21-9e0bad8a2226 /boot                   xfs     defaults        0 0
/dev/mapper/rhel_ibm--p8--kvm--03--guest--02-swap none                    swap    defaults        0 0
#VAGRANT-BEGIN
# The contents below are automatically generated by Vagrant. Do not modify.
#VAGRANT-END
none /dev/hugepages1G hugetlbfs pagesize=1G 0 0
tmpfs /run/user/1000 tmpfs rw,nosuid,nodev,relatime,size=10240k,mode=700,uid=1000 0 0
```
The upgrade initramfs contains the following. Notice the missing simlink for the hugepages mount point from local-fs.target.requires.
```
-rw-r--r--   1 root     root          453 Dec 10  2024 usr/lib/systemd/system/local-fs-pre.target
-rw-r--r--   1 root     root          556 Dec 10  2024 usr/lib/systemd/system/local-fs.target
drwxr-xr-x   2 root     root            0 Jan 30 00:00 usr/lib/systemd/system/local-fs.target.requires
lrwxrwxrwx   1 root     root           13 Jan 30 00:00 usr/lib/systemd/system/local-fs.target.requires/boot.mount -> ../boot.mount
lrwxrwxrwx   1 root     root           21 Jan 30 00:00 usr/lib/systemd/system/local-fs.target.requires/sysroot-boot.mount -> ../sysroot-boot.mount
drwxr-xr-x   2 root     root            0 Jan 30 00:00 usr/lib/systemd/system/local-fs.target.wants
lrwxrwxrwx   1 root     root           13 Jan 30 00:00 usr/lib/systemd/system/local-fs.target.wants/boot.mount -> ../boot.mount
lrwxrwxrwx   1 root     root           21 Jan 30 00:00 usr/lib/systemd/system/local-fs.target.wants/sysroot-boot.mount -> ../sysroot-boot.mount
lrwxrwxrwx   1 root     root           32 Jan 30 00:00 usr/lib/systemd/system/local-fs.target.wants/sysroot-dev-hugepages1G.mount -> ../sysroot-dev-hugepages1G.mount
lrwxrwxrwx   1 root     root           32 Jan 30 00:00 usr/lib/systemd/system/local-fs.target.wants/sysroot-run-user-1000.mount -> ../sysroot-run-user-1000.mount
```

After booting into the upgrade log, we see the following error, but the upgrade continues on without crashing.
```
...
Jul 14 21:43:00 localhost systemd[1]: sysroot-dev-hugepages1G.mount: Failed with result 'exit-code'.
Jul 14 21:43:00 localhost systemd[1]: Failed to mount sysroot-dev-hugepages1G.mount - /sysroot/dev/hugepages1G.
Jul 14 21:43:00 localhost mount[716]: mount: /sysroot/run/user/1000: mount point does not exist.
Jul 14 21:43:00 localhost mount[716]:        dmesg(1) may have more information after failed mount system call.
Jul 14 21:43:00 localhost systemd[1]: sysroot-run-user-1000.mount: Mount process exited, code=exited, status=32/n/a
Jul 14 21:43:00 localhost systemd[1]: sysroot-run-user-1000.mount: Failed with result 'exit-code'.
Jul 14 21:43:00 localhost systemd[1]: Failed to mount sysroot-run-user-1000.mount - /sysroot/run/user/1000.
Jul 14 21:43:00 localhost kernel: XFS (vda1): Ending clean mount
Jul 14 21:43:00 localhost systemd[1]: Mounted sysroot-boot.mount - /sysroot/boot.
Jul 14 21:43:00 localhost systemd[1]: Mounting boot.mount - /boot...
...
```

matejmatuska (approved):
Code LGTM, @MichalHe has already tested. Approving.

PR #1572, opened 2026-07-20
Add Vim /etc/virc migration actor for el8toel9 upgrade

zdohnal:
Remove 'filetype plugin on' and 'let skip_defaults_vim=1' from /etc/virc during RHEL 8 to 9 in-place upgrade. Uses scanner/modifier pattern: vircscanner detects lines in FactsPhase, vircmodifier removes them in ApplicationsPhase.

Jira: RHEL-194363
Assisted-by: Claude Code

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/vircscanner/libraries/vircscanner.py
+++ b/repos/system_upgrade/el8toel9/actors/vircscanner/libraries/vircscanner.py
@@ -1,0 +1,59 @@
+import os
+
+from leapp import reporting
+from leapp.libraries.common.rpms import has_package
+from leapp.libraries.stdlib import api
+from leapp.models import DistributionSignedRPM, VircConfig
+
+VIRC_CONFIG = '/etc/virc'
+
+OPTIONS_TO_REMOVE = (
+    'filetype',
+    'let',
+)
+
+
+def _scan_virc(path):
+    """
+    Read virc and return the original (unstripped) lines whose stripped content
+    matches any entry in OPTIONS_TO_REMOVE.
+    """
+    try:
+        with open(path, 'r') as f:
+            lines = f.readlines()
+    except (OSError, IOError):
+        api.current_logger().warning('Could not read {}.'.format(path))
+        return []
+
+    return [line for line in lines if any(line.strip().startswith(option) for option in OPTIONS_TO_REMOVE)]
+
+
+def process(actor):
+    if not has_package(DistributionSignedRPM, 'vim-minimal'):
+        api.current_logger().debug('vim-minimal is not installed, skipping virc scan.')
+        return
+
+    if not os.path.isfile(VIRC_CONFIG):
+        api.current_logger().debug('{} does not exist, skipping virc scan.'.format(VIRC_CONFIG))
+        return
+
+    found = _scan_virc(VIRC_CONFIG)
+    if not found:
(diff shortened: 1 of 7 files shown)


MichalHe:
@zdohnal Hi, thank you for the patch. I have a question regarding robustness of the approach: Can we assume that users will not modify virc too much/at all? I am asking because the current implementation has two quirks, and it might not be worthwhile to address these and make it more complex:
- `filetype    plugin on`  will not be detected (multiple whitespaces between `filetype` and `plugin`)
- if there are multiple offending lines of the same type (e.g., `filetype ...`) will cause the same line to appear multiple times in the report.

zdohnal:
Hi @MichalHe,

thank you for the review! IMO that file is not modified that often, but still it is configuration file which is used for settings of Vi, so I would rather like to handle it.

For context, both of those lines were either in default virc which was shipped with RPM, or added by Vim actors in the past (el7->el8 migration), so what I find less probable is user played with those directives which were in the config at the start (except for commenting them out), so it is highly probably those directives looks the same.

But I can adjust the code if you want - actually it will be even easier since neither filetype or let commands are present for Vi... I'll update the MR.

repos/system_upgrade/el8toel9/actors/vircscanner/libraries/vircscanner.py
```diff
@@ -0,0 +1,59 @@
...
+from leapp import reporting
+from leapp.libraries.common.rpms import has_package
+from leapp.libraries.stdlib import api
+from leapp.models import DistributionSignedRPM, VircConfig
+
+VIRC_CONFIG = '/etc/virc'
+
+OPTIONS_TO_REMOVE = (
+    'filetype',
+    'let',
```
mhecko-rh, line 12: Hi @zdohnal, is removing of all of `let ...` lines intended? The user could have there, e.g., `let loaded_matchparen = 1` as suggested in this [KB](https://access.redhat.com/solutions/59694).

mhecko-rh (commented):
I have executed the upgrade with the patch. The actor produces the following message:
```
Actor: virc_scanner
Phase: FactsCollection
Type: VircConfig
Message_data:
{
    "lines_to_remove": [
        "filetype plugin oN\n",
        "let skip_defaults_vim=1\n"
    ],
    "path": "/etc/virc"
}
```

After the upgrade, `/etc/virc` contains none of these lines.

Without this patch using the same `/etc/virc, `vi` reports failures after upgrade as described in `RHEL-194363`.


I am waiting with approving to settle with the author whether removing all `let ..` lines is intended.

repos/system_upgrade/el8toel9/actors/vircscanner/libraries/vircscanner.py
zdohnal: @mhecko-rh yes, it is intended - command `let` is not available in Vi (vim-minimal). The KB you mention works with .vimrc, which is Vim (vim-enhanced) config file, so it is not affected.

mhecko-rh: Ah, I did not notice that, my bad, sorry.

mhecko-rh (approved):
I have manually tested the patch and it fixes the issue. LGTM. E2E tests are green, merging.

PR #1574, opened 2026-07-23
fix(arm_boot_workaround): run after transaction dry run

MichalHe:
Previously, the only message consumed by the add_arm_bootloader_workaround actor was TargetUserSpaceInfo, meaning that the actor could be executed at any time in the InterimPreparationPhase. This is problematic in case that the dnf_dry_run actor finds that the transaction cannot be performed, e.g., because of an insufficient disk space as we would incorrectly modify the EFI boot order.

Jira-ref: RHEL-213986

### DIFF
--- a/repos/system_upgrade/el8toel9/actors/addarmbootloaderworkaround/actor.py
+++ b/repos/system_upgrade/el8toel9/actors/addarmbootloaderworkaround/actor.py
@@ -2,7 +2,7 @@ from leapp.actors import Actor
 from leapp.libraries.actor import addupgradebootloader
 from leapp.libraries.common.config import architecture
 from leapp.libraries.common.config.version import matches_target_version
-from leapp.models import ArmWorkaroundEFIBootloaderInfo, TargetUserSpaceInfo
+from leapp.models import ArmWorkaroundEFIBootloaderInfo, TargetUserSpaceInfo, TransactionDryRun
 from leapp.tags import InterimPreparationPhaseTag, IPUWorkflowTag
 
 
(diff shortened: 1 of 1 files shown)


pirat89 (approved):
lgtm. The change makes a sense, but tests failed - I haven't checked yet for the reason, but I expect it will be an infra issue. let me rerun them again

pirat89:
@MichalHe  in one+ case it's infra issue, in some cases there is a separate bug:
```
2026-07-27 10:22:10.840 ERROR    PID: 14205 leapp.workflow.FactsCollection: Actor scan_source_boot_entry has crashed: Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/leapp/repository/actor_definition.py", line 75, in _do_run
  File "/usr/lib/python3.6/site-packages/leapp/actors/__init__.py", line 319, in run
    self.process(*args)
  File "/etc/leapp/repos.d/system_upgrade/common/actors/scan_source_boot_loader/actor.py", line 18, in process
    scan_source_boot_entry_lib.scan_default_source_boot_entry()
  File "/etc/leapp/repos.d/system_upgrade/common/actors/scan_source_boot_loader/libraries/scan_source_boot_entry.py", line 47, in scan_default_source_boot_entry
    initramfs_path = default_kernel_info['initrd']
KeyError: 'initrd'
```

which is caused by a bug in internal grubby build which is going to be fixed yet. I am giong to check with the bootloader team what's an ETA for the fix.

MichalHe:
I have tested the patch with a reproducer that resembles the problem described in the attached Jira. That is, I have modified leapp's dnf plugin library to always raise a `DNFUpgradeTransactionError`. When the actor does not consume `TransactionDryRun`, the `add_arm_bootloader_workaround` actor was executed and EFI boot order has been modified, although the transaction check crashed. With this patch, the `add_arm_bootloader_workaround` is no longer executed when the transaction check crashes.

pirat89:
@MichalHe  thanks. Merging, the pipeline should be almost fixed now for IPU 8 -> 9, but I know it's ok.

PR #1576, opened 2026-07-29
Add KCS article link to dynamic linker configuration report

prjagtap-rh:
## The reason for the change
Users experiencing a custom dynamic linker configuration hit a confusing
"Error: No handle specified" / "Unable to install RHEL X userspace packages"
failure during upgrade, with no guidance pointing them to the documented
resolution. The existing report ("Detected customized configuration for
dynamic linker") lacks a reference to the relevant KCS article.

## What has been changed
Added an external link to KCS article 7130046 in the dynamic linker
configuration report, and a corresponding test assertion.

## How it has been changed
- Added `reporting.ExternalLink(url=..., title=...)` to the report
  entries in `_report_custom_dynamic_linker_configuration()`.
- Added a test assertion verifying the KCS URL is present in the
  report's external links when a report is generated.

## Reference to a Jira ticket
RHEL-119943

### DIFF
--- a/repos/system_upgrade/common/actors/checkdynamiclinkerconfiguration/libraries/checkdynamiclinkerconfiguration.py
+++ b/repos/system_upgrade/common/actors/checkdynamiclinkerconfiguration/libraries/checkdynamiclinkerconfiguration.py
@@ -19,6 +19,11 @@ def _report_custom_dynamic_linker_configuration(summary):
         reporting.Remediation(hint=('Remove or revert the custom dynamic linker configurations and apply the changes '
                                     'using the ldconfig command. In case of possible active software collections we '
                                     'suggest disabling them persistently.')),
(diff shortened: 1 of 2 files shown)


PeterMocary (commented):
If I understand this correctly, user encounters an error for which there is an article with a solution. We also have a related high risk report and this change updates the report with the solution article for the error. I'm not sure if user would expect a solution to an error in the report for a seemingly different problem, given that the error itself is not really connecting them in the first place.

There is also this article about the report being changed so maybe we should include it as well.
https://access.redhat.com/solutions/7102641

PeterMocary (approved):
I looked into the Jira comments of this issue and found out @pirat89 explains there that users are expected to read the reports in any case. I think that this case highlights  how high risk reports are important although easier to overlook than inhibitors.

PR #1579, opened 2026-08-03
Handle inaccessible EFI runtime variables gracefully

prjagtap-rh:
## The reason for the change

On UEFI systems where EFI runtime variables are inaccessible (e.g. certain hypervisors, RT kernels, or firmware configurations), `mokutil --sb-state` exits with "EFI variables are not supported". This error was not handled by the `system_facts` actor, causing it to crash with a `StopActorExecutionError` and blocking the process entirely.

During conversions (CentOS to RHEL), this also prevented the `securebootinhibit` actor from distinguishing between "Secure Boot unsupported" and "EFI vars inaccessible", making it impossible to safely determine whether conversion could proceed.

## What has been changed

- **FirmwareFacts model**: added `efi_vars_accessible` field  (`Nullable(Boolean)`) to distinguish between inaccessible EFI vars and hardware that simply lacks Secure Boot support.
- **system_facts actor library**: `_get_secure_boot_state()` now returns a `(secureboot_enabled, efi_vars_accessible)` tuple and gracefully handles the "EFI variables are not supported" error instead of crashing.  `get_firmware()` unpacks the tuple accordingly.
- **securebootinhibit actor** (conversion-only, ChecksPhase): extended to consume the new `efi_vars_accessible` field.  When EFI vars are inaccessible, presents a Leapp Dialog asking the user whether Secure Boot is enabled. Uses fail-safe logic and produces an inhibitor report when the state cannot be confirmed as safe.
- **Tests**: updated existing `test_systemfacts.py` for tuple returns and added a new `test_get_secure_boot_state_efi_vars_unavailable` case.  Updated `test_securebootinhibit.py` with parametrized test cases covering the dialog paths (answer True, False, and None) and the "EFI vars accessible but no SB support" path.

## How it has been changed

1. **Separating EFI variable accessibility from Secure Boot support status.**
  Previously both "no SB support" and "EFI vars inaccessible" mapped to `secureboot_enabled=None`, making them indistinguishable.  The new `efi_vars_accessible` field resolves this ambiguity.
3. **Fail-safe dialog design (conversion path only).**
   The dialog asks "Is Secure Boot enabled?" (`confirm_secureboot_enabled`).  A blind True or no answer keeps the conversion inhibited (safe default).
   Only a deliberate False (user explicitly confirms SB is disabled) removes the inhibitor.  This protects copy-paste users who confirm without verifying.
4. **Inhibitor behavior.**
   Two outcomes when EFI vars are inaccessible:
   - User answers False (SB disabled): no inhibitor, conversion proceeds.
   - User answers True or does not answer: inhibitor titled "EFI runtime variables are not accessible" with a hint to verify the Secure Boot status in UEFI firmware settings.

### Crash fix (upgrades and conversions)
1. On a UEFI system with inaccessible EFI vars (e.g. RT kernel), run `leapp preupgrade` and verify `system_facts` no longer crashes. `FirmwareFacts` should have `efi_vars_accessible=False` and `secureboot_enabled=None`.

### Dialog logic (conversions only)
2. During a conversion (`leapp convert`), verify the inhibitor "EFI runtime variables are not accessible" is produced.
4. Answer False (confirm SB is disabled):
   `leapp answer --section secure_boot_inhibit.confirm_secureboot_enabled=False`
   Re-run and verify the inhibitor is not produced.
5. Answer True (confirm SB is enabled):
   `leapp answer --section secure_boot_inhibit.confirm_secureboot_enabled=True`
   Re-run and verify the inhibitor is still produced.
6. On a BIOS system, verify no inhibitor from `secure_boot_inhibit`.

## Reference to a Jira ticket

Jira: RHEL-160096

## Test results

see https://github.com/oamg/leapp-repository/pull/1579#issuecomment-5327515929

### DIFF
--- a/repos/system_upgrade/common/actors/convert/securebootinhibit/actor.py
+++ b/repos/system_upgrade/common/actors/convert/securebootinhibit/actor.py
@@ -1,4 +1,6 @@
 from leapp.actors import Actor
+from leapp.dialogs import Dialog
+from leapp.dialogs.components import BooleanComponent
 from leapp.libraries.actor import securebootinhibit
 from leapp.models import FirmwareFacts
 from leapp.reporting import Report
@@ -8,12 +10,45 @@ from leapp.tags import ChecksPhaseTag, IPUWorkflowTag
 class SecureBootInhibit(Actor):
     """
     Inhibit the conversion if SecureBoot is enabled.
+
+    When EFI runtime variables are inaccessible and the Secure Boot state
+    cannot be determined automatically, the user is asked to confirm the
+    state via a dialog.
     """
 
     name = 'secure_boot_inhibit'
     consumes = (FirmwareFacts,)
     produces = (Report,)
     tags = (IPUWorkflowTag, ChecksPhaseTag)
+    dialogs = (
+        Dialog(
+            scope='secure_boot_inhibit',
+            reason='Confirmation',
+            components=(
+                BooleanComponent(
+                    key='confirm_secureboot_enabled',
+                    label='Is Secure Boot enabled on this system?',
+                    description=(
+                        'Set to True if Secure Boot is enabled, False if disabled.'
+                        ' If unsure, check your UEFI firmware settings.'
+                    ),
+                    reason=(
+                        'EFI runtime variables are not accessible, so the Secure Boot'
+                        ' state cannot be determined automatically.'
+                    ),
+                ),
+            ),
+        ),
+    )
+
+    _asked_answer = False
+    _sb_answer = None
+
+    def get_sb_answer(self):
+        if not self._asked_answer:
+            self._asked_answer = True
+            self._sb_answer = self.get_answers(self.dialogs[0]).get('confirm_secureboot_enabled')
+        return self._sb_answer
 
     def process(self):
         securebootinhibit.process()
--- a/repos/system_upgrade/common/actors/convert/securebootinhibit/actor.py
+++ b/repos/system_upgrade/common/actors/convert/securebootinhibit/actor.py
@@ -33,8 +33,8 @@ class SecureBootInhibit(Actor):
                         ' If unsure, check your UEFI firmware settings.'
                     ),
                     reason=(
-                        'EFI runtime variables are not accessible, so the Secure Boot'
-                        ' state cannot be determined automatically.'
+                        'The Secure Boot state cannot be determined automatically'
+                        ' because UEFI runtime variables are not accessible.'
                     ),
                 ),
             ),
--- a/repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
+++ b/repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
@@ -5,6 +5,33 @@ from leapp.libraries.stdlib import api
 from leapp.models import FirmwareFacts
 
 
+def _report_secureboot_enabled(extra_hint=''):
+    hint = (
+        "Disable Secure Boot to be able to convert the system to"
+        " a different Linux distribution. Then re-enable Secure Boot"
+        " again after the conversion process is finished successfully."
+        " Check instructions for your current OS, or hypervisor in"
+        " case of virtual machines, for more information how to"
+        " disable Secure Boot."
+    )
+    if extra_hint:
+        hint += ' ' + extra_hint
+    reporting.create_report([
+        reporting.Title(
+            "Detected enabled Secure Boot when trying to convert the system"
+        ),
+        reporting.Summary(
+            "Conversion to a different Linux distribution is not possible"
+            " when the Secure Boot is enabled. Artifacts of the target"
+            " Linux distribution are signed by keys that are not accepted"
+            " by the source Linux distribution."
+        ),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.INHIBITOR, reporting.Groups.BOOT]),
+        reporting.Remediation(hint=hint),
+    ])
+
+
 def process():
     if not is_conversion():
         return
@@ -16,27 +43,61 @@ def process():
             details={"details": "Actor did not receive FirmwareFacts message."},
         )
 
-    if ff.firmware == "efi" and ff.secureboot_enabled:
-        report = [
-            reporting.Title(
-                "Detected enabled Secure Boot when trying to convert the system"
-            ),
-            reporting.Summary(
-                "Conversion to a different Linux distribution is not possible"
-                " when the Secure Boot is enabled. Artifacts of the target"
-                " Linux distribution are signed by keys that are not accepted"
-                " by the source Linux distribution."
-            ),
-            reporting.Severity(reporting.Severity.HIGH),
-            reporting.Groups([reporting.Groups.INHIBITOR, reporting.Groups.BOOT]),
-            # TODO some link
-            reporting.Remediation(
-                hint="Disable Secure Boot to be able to convert the system to"
-                " a different Linux distribution. Then re-enable Secure Boot"
-                " again after the upgrade process is finished successfully."
-                " Check instructions for your current OS, or hypervisor in"
-                " case of virtual machines, for more information how to"
-                " disable Secure Boot."
-            ),
-        ]
-        reporting.create_report(report)
+    if ff.firmware != 'efi':
+        return
+
+    secureboot_enabled = ff.secureboot_enabled
+
+    # Hardware does not support Secure Boot (EFI vars readable, state is None).
+    if secureboot_enabled is None and ff.efi_vars_accessible is True:
+        api.current_logger().info(
+            "EFI variables are accessible but Secure Boot is not supported"
+            " by the hardware; proceeding."
+        )
+        return
+
+    # EFI runtime variables inaccessible -> the user's answer is our only source.
+    answered_by_user = False
+    if secureboot_enabled is None:
+        secureboot_enabled = api.current_actor().get_sb_answer()
+        answered_by_user = True
+
+    if secureboot_enabled is True:
+        _report_secureboot_enabled(
+            extra_hint=(
+                "If Secure Boot is already disabled and the answer was"
+                " incorrect, change the value of 'confirm_secureboot_enabled'"
+                " from True to False in the answer file and re-run."
+            ) if answered_by_user else '',
+        )
+        return
+
+    if secureboot_enabled is False:
+        if answered_by_user:
+            api.current_logger().info(
+                "User confirmed Secure Boot is disabled; proceeding despite"
+                " inaccessible EFI variables."
+            )
+        return
+
+    # Still None -> user did not answer; state genuinely undeterminable.
+    reporting.create_report([
+        reporting.Title("Cannot determine the Secure Boot status"),
+        reporting.Summary(
+            "The system is booted in UEFI mode but the Secure Boot state"
+            " cannot be determined automatically because EFI runtime"
+            " variables are not accessible. The Secure Boot status is"
+            " important for the conversion process as converting with"
+            " Secure Boot enabled could lead to a system that fails"
+            " to boot."
+        ),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.INHIBITOR, reporting.Groups.BOOT]),
+        reporting.Remediation(
+            hint="Verify the Secure Boot status in your UEFI firmware"
+            " settings. If Secure Boot is enabled, disable it before"
+            " proceeding with the conversion. Then provide the Secure Boot"
+            " status by setting the value of 'confirm_secureboot_enabled'"
+            " to True or False in the answer file and re-run."
+        ),
+    ])
--- a/repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
+++ b/repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
@@ -5,17 +5,26 @@ from leapp.libraries.stdlib import api
 from leapp.models import FirmwareFacts
 
 
-def _report_secureboot_enabled(extra_hint=''):
+def _report_secureboot_enabled(answered_by_user):
+    answerfile_hint = ""
+    if answered_by_user:
+        # extra space in the beginning due to pasting inside the hint str
+        answerfile_hint = (
+            " "
+            "Then update your answer about the secure boot in the answerfile"
+            " to reflect the new state (for key: 'confirm_secureboot_enabled')."
+        )
+
     hint = (
-        "Disable Secure Boot to be able to convert the system to"
-        " a different Linux distribution. Then re-enable Secure Boot"
-        " again after the conversion process is finished successfully."
+        "To be able to convert the system to a different Linux distribution,"
+        " disable Secure Boot.{} Then re-enable Secure Boot"
+        " again after the upgrade process with conversion is finished successfully."
         " Check instructions for your current OS, or hypervisor in"
         " case of virtual machines, for more information how to"
         " disable Secure Boot."
+        .format(answerfile_hint)
     )
-    if extra_hint:
-        hint += ' ' + extra_hint
+
     reporting.create_report([
         reporting.Title(
             "Detected enabled Secure Boot when trying to convert the system"
@@ -43,33 +52,23 @@ def process():
             details={"details": "Actor did not receive FirmwareFacts message."},
         )
 
-    if ff.firmware != 'efi':
+    if ff.firmware != 'efi' or ff.secureboot_enabled is False:
         return
 
     secureboot_enabled = ff.secureboot_enabled
-
-    # Hardware does not support Secure Boot (EFI vars readable, state is None).
     if secureboot_enabled is None and ff.efi_vars_accessible is True:
-        api.current_logger().info(
-            "EFI variables are accessible but Secure Boot is not supported"
-            " by the hardware; proceeding."
-        )
+        # HW does not support Secure Boot (EFI vars readable, SB state is None).
+        api.current_logger().info("Secure Boot is not supported by the HW. Skipping.")
         return
 
-    # EFI runtime variables inaccessible -> the user's answer is our only source.
     answered_by_user = False
     if secureboot_enabled is None:
+        # Cannot determine the SB state (no EFI vars) -> get answer from user
         secureboot_enabled = api.current_actor().get_sb_answer()
         answered_by_user = True
 
     if secureboot_enabled is True:
-        _report_secureboot_enabled(
-            extra_hint=(
-                "If Secure Boot is already disabled and the answer was"
-                " incorrect, change the value of 'confirm_secureboot_enabled'"
-                " from True to False in the answer file and re-run."
-            ) if answered_by_user else '',
-        )
+        _report_secureboot_enabled(answered_by_user)
         return
 
     if secureboot_enabled is False:
@@ -87,9 +86,8 @@ def process():
             "The system is booted in UEFI mode but the Secure Boot state"
             " cannot be determined automatically because EFI runtime"
             " variables are not accessible. The Secure Boot status is"
-            " important for the conversion process as converting with"
-            " Secure Boot enabled could lead to a system that fails"
-            " to boot."
+            " important to be able to determine next steps for the in-place"
+            " upgrade process with the conversion."
         ),
         reporting.Severity(reporting.Severity.HIGH),
         reporting.Groups([reporting.Groups.INHIBITOR, reporting.Groups.BOOT]),
--- a/repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
+++ b/repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
@@ -6,38 +6,61 @@ from leapp.models import FirmwareFacts
 
 
 def _report_secureboot_enabled(answered_by_user):
-    answerfile_hint = ""
+    answerfile_hint = ''
     if answered_by_user:
         # extra space in the beginning due to pasting inside the hint str
         answerfile_hint = (
-            " "
-            "Then update your answer about the secure boot in the answerfile"
-            " to reflect the new state (for key: 'confirm_secureboot_enabled')."
+            ' '
+            'Then update your answer about the secure boot state in the answerfile'
+            ' to reflect the new state (for key: "confirm_secureboot_enabled").'
         )
 
     hint = (
-        "To be able to convert the system to a different Linux distribution,"
-        " disable Secure Boot.{} Then re-enable Secure Boot"
-        " again after the upgrade process with conversion is finished successfully."
-        " Check instructions for your current OS, or hypervisor in"
-        " case of virtual machines, for more information how to"
-        " disable Secure Boot."
+        'To be able to convert the system to a different Linux distribution,'
+        ' disable Secure Boot.{} Then re-enable Secure Boot'
+        ' again after the upgrade and conversion process is finished successfully.'
+        ' Check instructions for your current OS, or hypervisor in'
+        ' case of virtual machines, for more information how to'
+        ' disable Secure Boot.'
         .format(answerfile_hint)
     )
 
     reporting.create_report([
         reporting.Title(
-            "Detected enabled Secure Boot when trying to convert the system"
+            'Detected enabled Secure Boot when trying to convert the system'
         ),
         reporting.Summary(
-            "Conversion to a different Linux distribution is not possible"
-            " when the Secure Boot is enabled. Artifacts of the target"
-            " Linux distribution are signed by keys that are not accepted"
-            " by the source Linux distribution."
+            'Conversion to a different Linux distribution is not possible'
+            ' when the Secure Boot is enabled. Artifacts of the target'
+            ' Linux distribution are signed by keys that are not accepted'
+            ' by the source Linux distribution.'
         ),
         reporting.Severity(reporting.Severity.HIGH),
         reporting.Groups([reporting.Groups.INHIBITOR, reporting.Groups.BOOT]),
         reporting.Remediation(hint=hint),
+    ])
+
+
+def _report_missing_answer():
+    reporting.create_report([
+        reporting.Title('Cannot determine the Secure Boot state'),
+        reporting.Summary(
+            'The system is booted in UEFI mode but the Secure Boot state'
+            ' cannot be determined automatically because EFI runtime'
+            ' variables are not accessible. The information Secure Boot state'
+            ' is required to determine the next steps for the in-place upgrade'
+            ' and conversion process.'
+        ),
+
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.INHIBITOR, reporting.Groups.BOOT]),
+        reporting.Remediation(
+            hint='Verify the Secure Boot state in your UEFI firmware'
+            ' settings. If Secure Boot is enabled, disable it before'
+            ' proceeding with the upgrade and conversion. Then provide the Secure Boot'
+            ' state by setting the value of "confirm_secureboot_enabled"'
+            ' to True or False in the answer file and re-run.'
+        ),
     ])
 
 
@@ -48,8 +71,8 @@ def process():
     ff = next(api.consume(FirmwareFacts), None)
     if not ff:
         raise StopActorExecutionError(
-            "Could not identify system firmware",
-            details={"details": "Actor did not receive FirmwareFacts message."},
+            'Could not identify system firmware',
+            details={'details': 'Actor did not receive FirmwareFacts message.'},
         )
 
     if ff.firmware != 'efi' or ff.secureboot_enabled is False:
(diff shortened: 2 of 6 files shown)


repos/system_upgrade/common/actors/systemfacts/libraries/systemfacts.py
```diff
@@ -315,20 +315,25 @@ def _get_firewall_status(service_name):
 
 
 def _get_secure_boot_state():
+    """Returns (secureboot_enabled, efi_vars_accessible) tuple."""
```
pirat89: nintpick: Note that this does not say much about the function -  such a docstring has no value. I suggest to either skip it or prepare a proper docstring, informing when is what set. E.g. what states of variables are expected? when? Also, for new docstrings, we agreed to follow sphinx style.

repos/system_upgrade/common/actors/checkefivars/libraries/checkefivars.py
```diff
@@ -0,0 +1,42 @@
...
+            ' are not accessible. This prevents Leapp from determining'
+            ' the Secure Boot state automatically.'
+        ),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.INHIBITOR, reporting.Groups.BOOT]),
+        reporting.Remediation(
+            hint=(
+                'If Secure Boot is disabled, confirm by answering the dialog:\n'
+                '   # leapp answer --section'
+                ' check_efi_vars.confirm_secureboot_disabled=True'
```
pirat89: I do not like this hint:
* why adding the command here at all?
   * creating this copy-paste command, users could just copy paste it without reading what is expected from them. I can imagine people copying it even when the answer should be False. we discussed to do it like in case of VDO, where a hint created by an actor does not provide such a command. yes, the autogenerator for dialogs do so without a context. that's not good neither, but not so problematic, as in case we would change the dialog and the hint would become invalid.
* I suggest to keep it as a text, exaplaining what & why we want people to provide the information. without the command, asking them to answer generated user questions.

repos/system_upgrade/common/actors/checkefivars/actor.py
```diff
@@ -0,0 +1,49 @@
...
+                        ' cannot determine the Secure Boot state automatically.'
+                        ' If you are certain that Secure Boot is disabled,'
+                        ' confirm by setting this to True.'
+                    ),
+                    reason=(
+                        'Leapp needs to know the Secure Boot state to ensure'
+                        ' a safe upgrade. On this system EFI runtime variables'
+                        ' are not accessible, so the state must be confirmed'
+                        ' manually.'
+                    ),
```
pirat89: I suggest to go in the positive way: `confirm_secureboot_enabled` - as most of the time people have enabled secure boot nowadays whether they boot with UEFI. Also, if you check the code dependent on whether the secureboot is enabled or not, you will realize that more systems are blocked if they use secure boot (when performing the conversion) than systems with disabled secure boot. from that point, it's a bigger problem if customer says that secure boot is disabled (and it is active) than in the other case -- if you have disabled secure boot and says that it is active, you will not (most likely) be harmed during the process, which is a different situation than in case you say that secure boot is disabled, and it is not.

this will be a better solution for copy-pasters who blindly perform commands without understanding.

repos/system_upgrade/common/actors/checkefivars/actor.py
```diff
@@ -0,0 +1,49 @@
...
+            reason='Confirmation',
+            components=(
+                BooleanComponent(
+                    key='confirm_secureboot_disabled',
+                    label='Confirm that Secure Boot is disabled on this system.',
+                    description=(
+                        'EFI runtime variables are not accessible, so Leapp'
+                        ' cannot determine the Secure Boot state automatically.'
+                        ' If you are certain that Secure Boot is disabled,'
+                        ' confirm by setting this to True.'
```
pirat89: this is not good woringg. it suggests users that if they are not sure, they should answer False. We should be sure by the answer. You can write just: `If Secure Boot is ...., confirm it by setting the True, set False otherwise.` (mentioning false as well, as there are unfortunately plenty of sys admins who does not understand the boolean logic)

pirat89 (changes requested):

repos/system_upgrade/common/actors/checkefivars/libraries/checkefivars.py
prjagtap-rh: @pirat89 Instead of command does this looks good ?
```
            hint=(
                'Verify the Secure Boot status in your UEFI firmware settings.'
                ' If Secure Boot is enabled, disable it before proceeding'
                ' with the upgrade. Upgrading with Secure Boot enabled while'
                ' EFI variables are inaccessible may result in an unbootable'
                ' system.'
            )
```

repos/system_upgrade/common/actors/checkefivars/libraries/checkefivars.py
```diff
@@ -0,0 +1,58 @@
...
+def process():
+    ff = next(api.consume(FirmwareFacts), None)
+    if not ff:
+        return
+
+    if ff.firmware != 'efi':
+        return
+
+    if ff.efi_vars_accessible is not False:
+        return
```
pirat89: btw, nitpick, no need to change it, but nicer would be:


```suggestion
    if not ff or ff.firmware != 'efi' or ff.efi_vars_accessible is not False:
        return
```

however, the first condition is kind of suggesting a bug in the code because that situation should not happen at all.

repos/system_upgrade/common/actors/checkefivars/libraries/checkefivars.py
```diff
@@ -0,0 +1,58 @@
...
+        title = 'Secure Boot is enabled'
+        summary = (
+            'The user confirmed that Secure Boot is enabled on this system.'
+            ' The upgrade cannot proceed safely with Secure Boot enabled'
+            ' while EFI runtime variables are inaccessible.'
+        )
+        hint = (
+            'Disable Secure Boot in your UEFI firmware settings'
+            ' before proceeding with the upgrade.'
+        )
```
pirat89: checking this part again, I think it's missleading. let's sync about on a call, as this seems wrong to me. Why to suggest to user - based on the answer - whether should do something or not? this actor has no idea whether secure boot is problem or not. it's only about figuring out whether we have the information about the secure boot when it's relevant or not. As far as we have the answer, this does not make sense to me. but let's sync first.

repos/system_upgrade/common/actors/checkefivars/libraries/checkefivars.py
```diff
@@ -0,0 +1,58 @@
...
+    else:
+        title = 'EFI runtime variables are not accessible'
```
pirat89: Thinking about that, we should change the title to sth like `Cannot detect the Secure Boot status`

repos/system_upgrade/common/actors/checkefivars/libraries/checkefivars.py
pirat89:
```
Upgrading with Secure Boot enabled while'
                ' EFI variables are inaccessible may result in an unbootable'
                ' system.'
```
this does not make a sense. How the EFI variables accessibility affects the upgrade after this PR?

repos/system_upgrade/common/actors/checkefivars/libraries/checkefivars.py
```diff
@@ -0,0 +1,58 @@
...
+            'The system is booted in UEFI mode but EFI runtime variables'
+            ' are not accessible. This prevents Leapp from determining'
+            ' the Secure Boot state automatically.'
+        )
+        hint = (
+            'Verify the Secure Boot status in your UEFI firmware settings.'
+            ' If Secure Boot is enabled, disable it before proceeding'
+            ' with the upgrade. Upgrading with Secure Boot enabled while'
+            ' EFI variables are inaccessible may result in an unbootable'
+            ' system.'
```
pirat89: this does not make sense to me. How the EFI variables accessibility affects the upgrade after this PR?

repos/system_upgrade/common/actors/checkefivars/libraries/checkefivars.py
prjagtap-rh: @pirat89 You're right first condition should never happen in practice.

Can we do this instead of silently returning ?
```
if not ff:
    raise StopActorExecutionError(
        'Could not identify system firmware',
        details={'details': 'Actor did not receive FirmwareFacts message.'}
    )
```

repos/system_upgrade/common/actors/checkefivars/libraries/checkefivars.py
prjagtap-rh: @pirat89
I agree  this actor is overstepping its responsibility.

Here the idea was 
User says SB disabled → proceed (just log)
User says SB enabled → inhibit with "disable Secure Boot"
User doesn't answer → inhibit with "EFI vars inaccessible"

repos/system_upgrade/common/actors/checkefivars/libraries/checkefivars.py
pirat89: yes

repos/system_upgrade/common/actors/checkefivars/libraries/checkefivars.py
pirat89: I think you misunderstand what I am trying to say. The only inhibitor that makes a sense is for the missing answer. why there should be any other inhibitor? as I wrote, let's sync on a call to find the common ground.

repos/system_upgrade/common/actors/checkefivars/libraries/checkefivars.py
pirat89: or `determine`  is probably the better wording.

matejmatuska:
Keep in mind that the primary reason for examining the secureboot state was that so we can inhibit upgrades with conversions (the `securebootinhibit actor`). For that reason `securebootinhibit` needs to know the secureboot state either determined via mokutil or defined by the user via answerfile. So the answer needs to be propagated there.

repos/system_upgrade/common/models/firmwarefacts.py
```diff
@@ -18,3 +18,12 @@ class FirmwareFacts(Model):
     Note that some machines do not support SecureBoot at all - even for systems booted with UEFI.
     For systems booted with UEFI that does not support SecureBoot set None.
```
pirat89: I thought I mention it already, but I cannot find it, so writing it now at least. The original docstring is getting outdated with this change, as the list of situations when the `secureboot_enabled` field can be `None` is extended.

```suggestion
    The value can be None in these cases:

        * on systems that do not support Secure Boot (even when booted with UEFI)
        * on systems with disabled EFI variables (usually on Real Time systems due to effect on latency)
```

repos/system_upgrade/common/models/firmwarefacts.py
```diff
@@ -18,3 +18,12 @@ class FirmwareFacts(Model):
...
     """
+
+    efi_vars_accessible = fields.Nullable(fields.Boolean())
+    """
+    Whether EFI runtime variables are accessible.
+
+    True when mokutil can query EFI vars (even if SecureBoot is unsupported).
+    False when mokutil fails with "EFI variables are not supported".
+    None for BIOS systems or when mokutil is not installed.
```
pirat89: Note that information about the unsupported Secure Boot is not relevant here. However, it makes sense to document when it's useful to check/use this value.

```suggestion
    True if EFI runtime variables are accessible via mokutil.

    Checking this value is useful in on systems booted with UEFI when the information about
    the Secure Boot settings is not determined (`secureboot_enabled` is None). 

    Another values:
    
        * False if mokutil fails with "EFI variables are not supported".
        * None for BIOS systems or when mokutil is not installed.
```

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
```diff
@@ -5,38 +5,102 @@
...
+                ' by the source Linux distribution.'
+            ),
+            reporting.Severity(reporting.Severity.HIGH),
+            reporting.Groups([reporting.Groups.INHIBITOR, reporting.Groups.BOOT]),
+            reporting.Remediation(
+                hint='Disable Secure Boot in your UEFI firmware settings'
+                ' before proceeding with the conversion.'
+            ),
+        ])
+        return
```
pirat89: why to generate 2 reports about the same thing from two places? this is anti-pattern, there is already the code generating the report. in case there is a need to mention in the report that it's detected based on user's choices, it should be a separate note / paragraph. but in this case I do not see it anyhow valuable - what's the benefit to know that user told us that system is using secure boot, when the upgrade cannot be performed so there is not even a harm if they answer incorrectly? logging is enough in such a case from my POV.

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
```diff
@@ -5,38 +5,102 @@
...
             ),
-        ]
-        reporting.create_report(report)
+        ])
+        return
+
+    if ff.secureboot_enabled is False:
+        return
+
+    _resolve_unknown_sb_state(ff)
```
pirat89: this is kind of antipattern. there is already evaluation checking how is it going with the secure boot, and after that, there is another evaluation, partially copying the resolution that has been already performed. the code needs to be written in the way, that is going through all the checks and ends in the point when it's clear that we cannot perform the upgrade in the discovered state.

pirat89 (changes requested):
the securebootinhibit actor needs to be updated properly. we could say the solution works, but the implemention contains anti-patterns which should not be there.

repos/system_upgrade/common/models/firmwarefacts.py
prjagtap-rh: @pirat89 I missed updating firmwarefact.py and a small correction regarding BIOS based systems
```
    The value can be None in these cases:
        * on BIOS systems (mokutil is never called)
        * on systems that do not support SecureBoot (even when booted with UEFI)
        * on systems with disabled EFI variables (usually on Real Time systems due to effect on latency)
```

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
```diff
@@ -29,14 +32,55 @@ def process():
...
+            "User confirmed Secure Boot is disabled; proceeding despite"
+            " inaccessible EFI variables."
+        )
+        return
+
+    if sb_answer is True:
+        api.current_logger().info(
+            "User confirmed Secure Boot is enabled via dialog."
+        )
+
```
pirat89: this information is not used anyhow so it does not have an effect on the upgrade.

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
```diff
@@ -16,8 +16,11 @@ def process():
             details={"details": "Actor did not receive FirmwareFacts message."},
         )
 
-    if ff.firmware == "efi" and ff.secureboot_enabled:
-        report = [
+    if ff.firmware != 'efi':
+        return
+
+    if ff.secureboot_enabled is True:
```
pirat89: as we discussed, this condition is not sufficient anymore, that's why the user question needs to be involved in the decision flow as well.

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
```diff
@@ -29,14 +32,55 @@ def process():
...
             reporting.Groups([reporting.Groups.INHIBITOR, reporting.Groups.BOOT]),
-            # TODO some link
             reporting.Remediation(
                 hint="Disable Secure Boot to be able to convert the system to"
                 " a different Linux distribution. Then re-enable Secure Boot"
-                " again after the upgrade process is finished successfully."
+                " again after the conversion process is finished successfully."
                 " Check instructions for your current OS, or hypervisor in"
                 " case of virtual machines, for more information how to"
                 " disable Secure Boot."
```
pirat89: just to not forget about that, we discussed that the remediation needs to be also extended in case that the user question is answered - as if user answer the user question, saying the SB is enabled, his answer will be still there. IOW, when they disable SB, they need to correct their answer as well.

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
```diff
@@ -29,14 +32,55 @@ def process():
...
+        )
+        return
+
+    if sb_answer is True:
+        api.current_logger().info(
+            "User confirmed Secure Boot is enabled via dialog."
+        )
+
+    reporting.create_report([
+        reporting.Title("EFI runtime variables are not accessible"),
```
pirat89: as pointed previously, this title is not so much informative for user:

> Thinking about that, we should change the title to sth like `Cannot determine the Secure Boot status`

Also note that the summary does not explain so much why we need to detect it.  We should tell them something like that this information is important for the upgrade process and incorrect information about the secure boot could lead to the upgrade failure, etc.

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
```diff
@@ -29,14 +32,55 @@ def process():
...
+            " determined automatically."
+        ),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.INHIBITOR, reporting.Groups.BOOT]),
+        reporting.Remediation(
+            hint="Verify the Secure Boot status in your UEFI firmware settings."
+            " If Secure Boot is enabled, disable it before proceeding"
+            " with the conversion. Converting with Secure Boot enabled while"
+            " EFI variables are inaccessible may result in an unbootable"
+            " system."
```
pirat89: as mentioned before:
> this does not make sense to me. How the EFI variables accessibility affects the upgrade after this PR?

What is expected from user to do after reading the remediation? Note that based on the expected behaviour, even if they disable the secure boot, they will still see this message. Where is information about that we need them to provide their answers to the asked questions.

pirat89 (changes requested):
currently, when converting a system booted with UEFI without accessible EFI vars, the upgrade is always inhibited on the report about inaccessible efi vars instead of having behaviour affected by the missing.

also some of previously required changes have not been adressed. Possibly they were overlooked after moving the code to a different actor.

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
```diff
@@ -16,27 +43,61 @@ def process():
...
-        ]
-        reporting.create_report(report)
+    if ff.firmware != 'efi':
+        return
+
+    secureboot_enabled = ff.secureboot_enabled
+
+    # Hardware does not support Secure Boot (EFI vars readable, state is None).
+    if secureboot_enabled is None and ff.efi_vars_accessible is True:
+        api.current_logger().info(
```
pirat89: Comments out of scope - putting comments like this suggests that when I got already here, we are in the state that HW does not support SB - which is misleading as such a state is hit just when the if statement is true. It's better to to put the comment inside the statement to be explicit about the scope, or rephrase it:

* updating the scope
```suggestion
    if secureboot_enabled is None and ff.efi_vars_accessible is True:
        # HW does not support Secure Boot (EFI vars readable, state is None).
        api.current_logger().info(
```

* rephrasing the comment
```suggestion
    # Check whether the Secure Boot is supported on the HW
    if secureboot_enabled is None and ff.efi_vars_accessible is True:
        api.current_logger().info(
```

Same for a comment below.

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
```diff
@@ -5,6 +5,33 @@
...
+    hint = (
+        "Disable Secure Boot to be able to convert the system to"
+        " a different Linux distribution. Then re-enable Secure Boot"
+        " again after the conversion process is finished successfully."
+        " Check instructions for your current OS, or hypervisor in"
+        " case of virtual machines, for more information how to"
+        " disable Secure Boot."
+    )
+    if extra_hint:
+        hint += ' ' + extra_hint
```
pirat89: 2 things:
* do not concatanate strings like this, use rather formatting (`.format()` method) or format strings. so e.g.:
```python
hint = f'{hint} {extra_hint}'
```

* Reading the hint, the structure is not ideal. now it says:
  * without user questions: `do sth to continue. do sth after the upgrade`
  * with user questions: `do sth to continue. do sth after the upgrade. if user questions, do sth to continue`
 
the structure could be confusing for some users, as they could misunderstand that user questions are related still to the part before the upgrade, to be able to continue. So the structure should be rather something like: `do sth to continue + extra hint. do sth after the upgrade`.

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
```diff
@@ -16,27 +43,61 @@ def process():
...
+    if secureboot_enabled is None:
+        secureboot_enabled = api.current_actor().get_sb_answer()
+        answered_by_user = True
+
+    if secureboot_enabled is True:
+        _report_secureboot_enabled(
+            extra_hint=(
+                "If Secure Boot is already disabled and the answer was"
+                " incorrect, change the value of 'confirm_secureboot_enabled'"
+                " from True to False in the answer file and re-run."
```
pirat89: this is little bit confusing. `the answer was incorrect` - we expect the answer was correct. it can be possibly outdated by now but we do not know whether customer disabled SB already or not. it's better to write it in the way like: `after you disable SB, update the provided answer.` - like in the other inhibitor below.

pirat89 (changes requested):

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
```diff
@@ -5,6 +5,64 @@
...
+        answerfile_hint = (
+            " "
+            "Then update your answer about the secure boot in the answerfile"
+            " to reflect the new state (for key: 'confirm_secureboot_enabled')."
+        )
+
+    hint = (
+        "To be able to convert the system to a different Linux distribution,"
+        " disable Secure Boot.{} Then re-enable Secure Boot"
+        " again after the upgrade process with conversion is finished successfully."
```
matejmatuska: seems like there some typos

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
```diff
@@ -5,6 +5,64 @@
 from leapp.models import FirmwareFacts
 
 
+def _report_secureboot_enabled(answered_by_user):
+    answerfile_hint = ""
+    if answered_by_user:
+        # extra space in the beginning due to pasting inside the hint str
+        answerfile_hint = (
+            " "
+            "Then update your answer about the secure boot in the answerfile"
```
matejmatuska:
```suggestion
            "Then update your answer about the secure boot state in the answerfile"
```

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
```diff
@@ -5,6 +5,64 @@
...
+        ),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.INHIBITOR, reporting.Groups.BOOT]),
+        reporting.Remediation(hint=hint),
+    ])
+
+
+def _report_missing_answer():
+    reporting.create_report([
+        reporting.Title("Cannot determine the Secure Boot status"),
```
matejmatuska: I think the the" shouldn't be here, also let's use state instead of status, as we use that everywhere. 
```suggestion
        reporting.Title("Cannot determine Secure Boot state"),
```

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
```diff
@@ -5,6 +5,64 @@
...
+        reporting.Summary(
+            "The system is booted in UEFI mode but the Secure Boot state"
+            " cannot be determined automatically because EFI runtime"
+            " variables are not accessible. The Secure Boot status is"
+            " important to be able to determine next steps for the in-place"
+            " upgrade process with the conversion."
```
matejmatuska:
```suggestion
            "The system is booted in UEFI mode but the Secure Boot state"
            " cannot be determined automatically because EFI runtime"
            " variables are not accessible. The Secure Boot state is"
            " important to be able to determine next steps for the in-place"
            " upgrade process with the conversion."
```

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
```diff
@@ -5,6 +5,64 @@
...
+        ),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.INHIBITOR, reporting.Groups.BOOT]),
+        reporting.Remediation(
+            hint="Verify the Secure Boot status in your UEFI firmware"
+            " settings. If Secure Boot is enabled, disable it before"
+            " proceeding with the conversion. Then provide the Secure Boot"
+            " status by setting the value of 'confirm_secureboot_enabled'"
+            " to True or False in the answer file and re-run."
```
matejmatuska:
```suggestion
            hint="Verify the Secure Boot state in your UEFI firmware"
            " settings. If Secure Boot is enabled, disable it before"
            " proceeding with the conversion. Then provide the Secure Boot"
            " state by setting the value of 'confirm_secureboot_enabled'"
            " to True or False in the answer file and re-run."
```

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
```diff
@@ -16,27 +74,28 @@ def process():
...
+        api.current_logger().info("Secure Boot is not supported by the HW. Skipping.")
+        return
+
+    answered_by_user = False
+    if secureboot_enabled is None:
+        # Cannot determine the SB state (no EFI vars) -> get answer from user
+        secureboot_enabled = api.current_actor().get_sb_answer()
+        if secureboot_enabled is None:
+            _report_missing_answer()
+            continue
```
matejmatuska:
```suggestion
            return
```

repos/system_upgrade/common/actors/convert/securebootinhibit/actor.py
```diff
@@ -8,12 +10,45 @@
...
+                BooleanComponent(
+                    key='confirm_secureboot_enabled',
+                    label='Is Secure Boot enabled on this system?',
+                    description=(
+                        'Set to True if Secure Boot is enabled, False if disabled.'
+                        ' If unsure, check your UEFI firmware settings.'
+                    ),
+                    reason=(
+                        'EFI runtime variables are not accessible, so the Secure Boot'
+                        ' state cannot be determined automatically.'
```
matejmatuska: Let's get rid of the "so"
```suggestion
                        'The Secure Boot state cannot be determined automatically'
                        ' because UEFI runtime variables are not accessible.'
```

matejmatuska (changes requested):
Some small suggestions

repos/system_upgrade/common/actors/convert/securebootinhibit/libraries/securebootinhibit.py
pirat89: keeping `the` :)

pirat89:
Tested manually:
* uefi with SB with EFI vars: inhibit as usually; no dialog prompt
* uefi with SB without EFI vars (RT kernel):
  * without answer: user question required + inhibitor informing about the missing EFI vars & missing answer
     ```
        1. Cannot determine the Secure Boot state
        2. Missing required answers in the answer file

    # report:
    Risk Factor: high (inhibitor)
    Title: Cannot determine the Secure Boot state
    Summary: The system is booted in UEFI mode but the Secure Boot state cannot be determined automatically because EFI runtime variables are not accessible. The information Secure Boot state is required to determine the next steps for the in-place upgrade and conversion process.
    Remediation: [hint] Verify the Secure Boot state in your UEFI firmware settings. If Secure Boot is enabled, disable it before proceeding with the upgrade and conversion. Then provide the Secure Boot state by setting the value of "confirm_secureboot_enabled" to True or False in the answer file and re-run.
    Key: bd945b38c54375e2678fa95f42eef4eb8015f9a7

    # answerfile:
    [secure_boot_inhibit]
    # Title:              None
    # Reason:             Confirmation
    # =============== secure_boot_inhibit.confirm_secureboot_enabled ==============
    # Label:              Is Secure Boot enabled on this system?
    # Description:        Set to True if Secure Boot is enabled, False if disabled. If unsure, check your UEFI firmware settings.
    # Reason:             The Secure Boot state cannot be determined automatically because UEFI runtime variables are not accessible.
    # Type:               bool
    # Default:            None
    # Available choices: True/False
    # Unanswered question. Uncomment the following line with your answer
    # confirm_secureboot_enabled = 
     ```
  * with answer `True`: blocked + updated report
    ```
        1. Detected enabled Secure Boot when trying to convert the system

    # report
    Risk Factor: high (inhibitor)
    Title: Detected enabled Secure Boot when trying to convert the system
    Summary: Conversion to a different Linux distribution is not possible when the Secure Boot is enabled. Artifacts of the target Linux distribution are signed by keys that are not accepted by the source Linux distribution.
    Remediation: [hint] To be able to convert the system to a different Linux distribution, disable Secure Boot. Then update your answer about the secure boot state in the answerfile to reflect the new state (for key: "confirm_secureboot_enabled"). Then re-enable Secure Boot again after the upgrade and conversion process is finished successfully. Check instructions for your current OS, or hypervisor in case of virtual machines, for more information how to disable Secure Boot.
    Key: 7f1a9f7d93d919386bd27ff890b8c7c2f81c2639
    ```
  * with answer `False`: no inhibitor + log about user's answer

matejmatuska (approved):
LGTM now

pirat89 (approved):

PR #1581, opened 2026-08-12
Replace custom FMT_LIST_SEPARATOR formatting with stdlib format_list

prjagtap-rh:
## What has been changed
Replaced all custom FMT_LIST_SEPARATOR-based list formatting (and local helper functions like` _format_list`, `_formatted_list_output`, `__human_readable_list`, etc.) with the framework-provided format_list() function from leapp.libraries.stdlib. Modified files span across common/, el8toel9/, and el9toel10/ repositories, covering both actor libraries and shared libraries.

## How it has been changed
- Removed local `FMT_LIST_SEPARATOR` constant definitions and local
 formatting helper functions from each actor library and actor module.
- Added `from leapp.libraries.stdlib import format_list` imports.
- Replaced `.join()`-based and loop-based formatting calls with
 `format_list(items)` (or `format_list(items, callback_sort=None)`
 where ordering must be preserved).
- Used `FMT_LIST_SEPARATOR` from stdlib for the truncation suffix 
  in `checkthirdpartytargetpythonmodules` to stay in sync with format_list.
- Updated one test assertion in `component_test_sssdchecks.py` to match
 the new `format_list` output.

## Reason for the change
- **Standardization**: RHEL-126447 introduced `format_list` in the
 framework to provide a single, consistent way to format lists in
 reports, logs, and error messages. Each actor was previously
 reinventing this with its own `FMT_LIST_SEPARATOR` constant.
- **Sorted output**: `format_list` sorts by default, improving report
 readability for users. Previously most lists were unsorted.
- **Maintainability**: Reduces ~124 lines of duplicated formatting
 boilerplate across the codebase.

### DIFF
--- a/repos/system_upgrade/common/actors/checkcustommodifications/libraries/checkcustommodifications.py
+++ b/repos/system_upgrade/common/actors/checkcustommodifications/libraries/checkcustommodifications.py
@@ -1,25 +1,17 @@
 from leapp import reporting
-from leapp.libraries.stdlib import api
+from leapp.libraries.stdlib import api, format_list
 from leapp.models import CustomModifications
-
-FMT_LIST_SEPARATOR = "\n    - "
 
 
 def _pretty_files(messages):
     """
     Return formatted string of discovered files from obtained CustomModifications messages.
     """
-    flist = []
+    items = []
     for msg in messages:
         actor = ' (Actor: {})'.format(msg.actor_name) if msg.actor_name else ''
-        flist.append(
-            '{sep}{filename}{actor}'.format(
-                sep=FMT_LIST_SEPARATOR,
-                filename=msg.filename,
-                actor=actor
-            )
-        )
-    return ''.join(flist)
+        items.append('{}{}'.format(msg.filename, actor))
+    return format_list(items)
 
 
 def _is_modified_config(msg):
--- a/repos/system_upgrade/common/actors/checkdynamiclinkerconfiguration/libraries/checkdynamiclinkerconfiguration.py
+++ b/repos/system_upgrade/common/actors/checkdynamiclinkerconfiguration/libraries/checkdynamiclinkerconfiguration.py
@@ -1,13 +1,11 @@
 from leapp import reporting
-from leapp.libraries.stdlib import api
+from leapp.libraries.stdlib import api, format_list
 from leapp.models import DynamicLinkerConfiguration
 
 LD_SO_CONF_DIR = '/etc/ld.so.conf.d'
 LD_SO_CONF_MAIN = '/etc/ld.so.conf'
 LD_LIBRARY_PATH_VAR = 'LD_LIBRARY_PATH'
 LD_PRELOAD_VAR = 'LD_PRELOAD'
-FMT_LIST_SEPARATOR_1 = '\n- '
-FMT_LIST_SEPARATOR_2 = '\n    - '
 
 
 def _report_custom_dynamic_linker_configuration(summary):
@@ -39,9 +37,9 @@ def check_dynamic_linker_configuration():
     custom_configurations = ''
     if configuration.main_config.modified:
         custom_configurations += (
-            '{}The {} file has unexpected contents:{}{}'
-            .format(FMT_LIST_SEPARATOR_1, LD_SO_CONF_MAIN,
-                    FMT_LIST_SEPARATOR_2, FMT_LIST_SEPARATOR_2.join(configuration.main_config.modified_lines))
+            '\n- The {} file has unexpected contents:{}'
+            .format(LD_SO_CONF_MAIN,
+                    format_list(configuration.main_config.modified_lines, callback_sort=None))
         )
 
     custom_configs = []
@@ -51,15 +49,14 @@ def check_dynamic_linker_configuration():
 
     if custom_configs:
         custom_configurations += (
-            '{}The following drop in config files were marked as custom:{}{}'
-            .format(FMT_LIST_SEPARATOR_1, FMT_LIST_SEPARATOR_2, FMT_LIST_SEPARATOR_2.join(custom_configs))
+            '\n- The following drop in config files were marked as custom:{}'
+            .format(format_list(custom_configs))
         )
 
     if configuration.used_variables:
         custom_configurations += (
-            '{}The following variables contain unexpected dynamic linker configuration:{}{}'
-            .format(FMT_LIST_SEPARATOR_1, FMT_LIST_SEPARATOR_2,
-                    FMT_LIST_SEPARATOR_2.join(configuration.used_variables))
+            '\n- The following variables contain unexpected dynamic linker configuration:{}'
+            .format(format_list(configuration.used_variables))
         )
 
     if custom_configurations:
--- a/repos/system_upgrade/common/actors/checkosrelease/libraries/checkosrelease.py
+++ b/repos/system_upgrade/common/actors/checkosrelease/libraries/checkosrelease.py
@@ -3,9 +3,9 @@ import os
 from leapp import reporting
 from leapp.libraries.common.config import version
 from leapp.libraries.common.distro import DISTRO_REPORT_NAMES
+from leapp.libraries.stdlib import format_list
 
 COMMON_REPORT_TAGS = [reporting.Groups.SANITY]
-FMT_LIST_SEPARATOR = '\n    - '
 
 related = [reporting.RelatedResource('file', '/etc/os-release')]
 
@@ -41,9 +41,9 @@ def check_os_version():
             ),
             reporting.Summary(
                 'The supported OS releases for the upgrade process:'
-                '{}{}\n\nThe detected OS release is: {}'.format(FMT_LIST_SEPARATOR,
-                                                                FMT_LIST_SEPARATOR.join(supported_releases),
-                                                                current_release)
+                '{}\n\nThe detected OS release is: {}'.format(
+                    format_list(supported_releases, callback_sort=None),
+                    current_release)
             ),
             reporting.Severity(reporting.Severity.HIGH),
             reporting.Groups(COMMON_REPORT_TAGS),
--- a/repos/system_upgrade/common/actors/checkthirdpartytargetpythonmodules/libraries/checkthirdpartytargetpythonmodules.py
+++ b/repos/system_upgrade/common/actors/checkthirdpartytargetpythonmodules/libraries/checkthirdpartytargetpythonmodules.py
@@ -1,23 +1,19 @@
 from leapp import reporting
-from leapp.libraries.stdlib import api
+from leapp.libraries.stdlib import api, FMT_LIST_SEPARATOR, format_list
 from leapp.models import ThirdPartyTargetPythonModules
 
-FMT_LIST_SEPARATOR = '\n    - '
 MAX_REPORTED_ITEMS = 30
 
 
-def _formatted_list_output_with_max_items(input_list, sep=FMT_LIST_SEPARATOR, max_items=MAX_REPORTED_ITEMS):
+def _formatted_list_output_with_max_items(input_list, max_items=MAX_REPORTED_ITEMS):
     if not input_list:
         return ''
 
-    total_count = len(input_list)
-    items_to_show = input_list[:max_items]
-    formatted = ['{}{}'.format(sep, item) for item in items_to_show]
+    result = format_list(input_list, limit=max_items, callback_sort=None)
+    if len(input_list) > max_items:
+        result += '{}... and {} more'.format(FMT_LIST_SEPARATOR, len(input_list) - max_items)
 
-    if total_count > max_items:
-        formatted.append('{}... and {} more'.format(sep, total_count - max_items))
-
-    return ''.join(formatted)
+    return result
 
 
 def check_third_party_target_python_modules(third_party_target_python_modules):
--- a/repos/system_upgrade/common/actors/checkyumpluginsenabled/libraries/checkyumpluginsenabled.py
+++ b/repos/system_upgrade/common/actors/checkyumpluginsenabled/libraries/checkyumpluginsenabled.py
@@ -2,11 +2,11 @@ import os
 
 from leapp import reporting
 from leapp.libraries.common.rhsm import skip_rhsm
+from leapp.libraries.stdlib import format_list
 
 # If LEAPP_NO_RHSM is set, subscription-manager and product-id will not be
 # considered as required when checking whether the required plugins are enabled.
 REQUIRED_DNF_PLUGINS = {'subscription-manager', 'product-id'}
-FMT_LIST_SEPARATOR = '\n    - '
 
 
 def check_required_dnf_plugins_enabled(pkg_manager_info):
@@ -24,9 +24,7 @@ def check_required_dnf_plugins_enabled(pkg_manager_info):
         missing_required_plugins -= {'subscription-manager', 'product-id'}
 
     if missing_required_plugins:
-        missing_required_plugins_text = ''
-        for missing_plugin in missing_required_plugins:
-            missing_required_plugins_text += '{0}{1}'.format(FMT_LIST_SEPARATOR, missing_plugin)
+        missing_required_plugins_text = format_list(missing_required_plugins)
 
         # dnf_conf_path - enable/disable plugins globally
         # rhsm_plugin_conf, product_id_plugin_conf - plugins can be disabled individually
@@ -44,7 +42,7 @@ def check_required_dnf_plugins_enabled(pkg_manager_info):
         reporting.create_report([
             reporting.Title('Required DNF plugins are not being loaded.'),
             reporting.Summary(
-                'The following DNF plugins are not being loaded: {}'.format(missing_required_plugins_text)
+                'The following DNF plugins are not being loaded:{}'.format(missing_required_plugins_text)
             ),
             reporting.Remediation(
                 hint=(
--- a/repos/system_upgrade/common/actors/distributionsignedrpmcheck/libraries/distributionsignedrpmcheck.py
+++ b/repos/system_upgrade/common/actors/distributionsignedrpmcheck/libraries/distributionsignedrpmcheck.py
@@ -1,9 +1,7 @@
 from leapp import reporting
-from leapp.libraries.stdlib import api
+from leapp.libraries.stdlib import api, format_list
 from leapp.libraries.stdlib.config import is_verbose
 from leapp.models import ThirdPartyRPM
-
-FMT_LIST_SEPARATOR = "\n    - "
 
 
 def _generate_report(packages):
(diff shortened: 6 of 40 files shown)


repos/system_upgrade/common/actors/checkthirdpartytargetpythonmodules/libraries/checkthirdpartytargetpythonmodules.py
```diff
@@ -1,23 +1,19 @@
...
-    formatted = ['{}{}'.format(sep, item) for item in items_to_show]
+    result = format_list(input_list, limit=max_items)
+    if len(input_list) > max_items:
+        result += '\n    - ... and {} more'.format(len(input_list) - max_items)
 
-    if total_count > max_items:
-        formatted.append('{}... and {} more'.format(sep, total_count - max_items))
-
-    return ''.join(formatted)
+    return result
```
PeterMocary: Note that this version sorts the list and the original one just took the first _max_items_.

IIRC there were discussions whether to sort this list some time back and apparently it wasn't sorted in the end. Maybe @karolinku can confirm?

repos/system_upgrade/common/actors/checkdynamiclinkerconfiguration/libraries/checkdynamiclinkerconfiguration.py
```diff
@@ -39,9 +37,9 @@ def check_dynamic_linker_configuration():
     custom_configurations = ''
     if configuration.main_config.modified:
         custom_configurations += (
-            '{}The {} file has unexpected contents:{}{}'
-            .format(FMT_LIST_SEPARATOR_1, LD_SO_CONF_MAIN,
-                    FMT_LIST_SEPARATOR_2, FMT_LIST_SEPARATOR_2.join(configuration.main_config.modified_lines))
+            '\n- The {} file has unexpected contents:{}'
+            .format(LD_SO_CONF_MAIN,
+                    format_list(configuration.main_config.modified_lines))
         )
```
PeterMocary, line 43: Again, the format_list sorts these by default. I don't think it is a big deal but having it in the same order as found in the config file is better in my opinion.

PeterMocary (commented):
Found few concerns about sorting.

There is a missed report [here](https://github.com/oamg/leapp-repository/blob/main/repos/system_upgrade/common/actors/initramfs/checkinitramfstasks/libraries/checkinitramfstasks.py#L18)

We can use this for logs as well, for example in [rhsm.py:198](https://github.com/oamgleapp-repository/blob/59190847a8d4f87b5fa07758cbef576dd1521f6d/repos/system_upgrade/common/libraries/rhsm.py#L198), [rhsm.py:218](  https://github.com/oamg/leapp-repository/blob/main/repos/system_upgrade/common/libraries/rhsm.py#L218)

I didn't do a full review yet, and mind that the lists of missed places is naturally not exhaustive.

repos/system_upgrade/common/actors/checkcustommodifications/libraries/checkcustommodifications.py
```diff
@@ -1,25 +1,17 @@
...
-        flist.append(
-            '{sep}{filename}{actor}'.format(
-                sep=FMT_LIST_SEPARATOR,
-                filename=msg.filename,
-                actor=actor
-            )
-        )
-    return ''.join(flist)
+        items.append('{}{}'.format(msg.filename, actor))
+    return format_list(items, callback_sort=None)
```
PeterMocary: Nitpick: I wouldn't mid sorting here, as the order won't matter. We decided to have default sorted, so I'd sort most of the time unless it breaks some semantics of the message.

repos/system_upgrade/common/actors/checkdynamiclinkerconfiguration/libraries/checkdynamiclinkerconfiguration.py
```diff
@@ -51,15 +49,14 @@ def check_dynamic_linker_configuration():
...
+            .format(format_list(custom_configs, callback_sort=None))
         )
 
     if configuration.used_variables:
         custom_configurations += (
-            '{}The following variables contain unexpected dynamic linker configuration:{}{}'
-            .format(FMT_LIST_SEPARATOR_1, FMT_LIST_SEPARATOR_2,
-                    FMT_LIST_SEPARATOR_2.join(configuration.used_variables))
+            '\n- The following variables contain unexpected dynamic linker configuration:{}'
+            .format(format_list(configuration.used_variables, callback_sort=None))
```
PeterMocary: These should be sorted. It won't be confusing because there is no preliminary sorting as in the unexpected contents of a file where you would expect them to be ordered the same as they are in the file.

repos/system_upgrade/common/actors/checkosrelease/libraries/checkosrelease.py
```diff
@@ -41,9 +41,9 @@ def check_os_version():
             ),
             reporting.Summary(
                 'The supported OS releases for the upgrade process:'
-                '{}{}\n\nThe detected OS release is: {}'.format(FMT_LIST_SEPARATOR,
-                                                                FMT_LIST_SEPARATOR.join(supported_releases),
-                                                                current_release)
+                '{}\n\nThe detected OS release is: {}'.format(
+                    format_list(supported_releases, callback_sort=None),
+                    current_release)
```
PeterMocary, line 46: We could technically add sorting that would make sure the versions are sorted correctly, but default sorted doesn't do that well. It would need to be something like the following (using `partial` from `functools` to prefill the sorted):

```Python
my_sort = partial(sorted, key=lambda v: tuple(int(x) for x in v.split('.')))
format_list(supported_releases. callback_sort=my_sort)
```

However, we can expect the list to be sorted already so no need to do that really. Same would apply for target versions in `checktargetversion` actor but it is also presorted so not really needed.

repos/system_upgrade/el8toel9/actors/rocecheck/libraries/rocecheck.py
```diff
@@ -50,7 +45,7 @@ def _report_wrong_setup(roce):
         ' For more information, see: {url}'
```
PeterMocary, line 45: Nitpick: I'd move this to external link. Not really necessary to do it now though.

repos/system_upgrade/common/actors/checkdynamiclinkerconfiguration/libraries/checkdynamiclinkerconfiguration.py
```diff
@@ -39,9 +37,9 @@ def check_dynamic_linker_configuration():
     custom_configurations = ''
     if configuration.main_config.modified:
         custom_configurations += (
-            '{}The {} file has unexpected contents:{}{}'
-            .format(FMT_LIST_SEPARATOR_1, LD_SO_CONF_MAIN,
-                    FMT_LIST_SEPARATOR_2, FMT_LIST_SEPARATOR_2.join(configuration.main_config.modified_lines))
+            '\n- The {} file has unexpected contents:{}'
```
matejmatuska, line 40: I think you could've kept the second separator, why remove it?

repos/system_upgrade/el9toel10/actors/krb5conf/checkkrb5conf/libraries/checkkrb5conf.py
```diff
@@ -1,15 +1,13 @@
...
-FMT_LIST_SEPARATOR = "\n    - "
 
-
-def __human_readable_list(unmanaged_files):
-    if unmanaged_files:
-        return FMT_LIST_SEPARATOR + FMT_LIST_SEPARATOR.join(unmanaged_files)
+def __human_readable_list(items):
+    if items:
+        return format_list(items)
     return ''
```
PeterMocary: This wrapper is not necessary anymore, format_list returns empty string for empty list.

repos/system_upgrade/common/actors/checkthirdpartytargetpythonmodules/libraries/checkthirdpartytargetpythonmodules.py
```diff
@@ -1,23 +1,19 @@
...
+def _formatted_list_output_with_max_items(input_list, max_items=MAX_REPORTED_ITEMS):
     if not input_list:
         return ''
 
-    total_count = len(input_list)
-    items_to_show = input_list[:max_items]
-    formatted = ['{}{}'.format(sep, item) for item in items_to_show]
+    result = format_list(input_list, limit=max_items, callback_sort=None)
+    if len(input_list) > max_items:
+        result += '\n    - ... and {} more'.format(len(input_list) - max_items)
```
matejmatuska: The leapp stdlib still has the `FMT_LIST_SEPARATOR`, which the `format_list` function uses internally, I would use that instead of the `\n    -``:

```suggestion
        result += '{}... and {} more'.format(FMT_LIST_SEPARATOR, len(input_list) - max_items)
```
ofc import it first.

repos/system_upgrade/common/actors/checkyumpluginsenabled/libraries/checkyumpluginsenabled.py
matejmatuska: Not a big deal, but I just noticed there is an extra space:
```suggestion
                'The following DNF plugins are not being loaded:{}'.format(missing_required_plugins_text)
```

repos/system_upgrade/common/actors/distributionsignedrpmcheck/libraries/distributionsignedrpmcheck.py
matejmatuska: You can drop this sort, the formatting function does it.

repos/system_upgrade/el9toel10/actors/sssd/sssdchecks/libraries/sssdchecks.py
```diff
@@ -17,9 +16,8 @@ def check_config(model):
         'to reflect this by updating every mention of sss_ssh_knownhostsproxy by '
         'the corresponding mention of sss_ssh_knownhosts, even those commented out. '
         'SSSD\'s ssh service will be enabled if not already done.\n\n'
-        'The following files will be updated:{}{}'.format(
-            FMT_LIST_SEPARATOR,
-            FMT_LIST_SEPARATOR.join(model.sssd_config_files + model.ssh_config_files)
+        'The following files will be updated:{}'.format(
+            format_list(model.sssd_config_files + model.ssh_config_files, callback_sort=None)
```
PeterMocary, line 20: Nitpick: I'd probably call format_list twice here for each list and then combine the result in the summary so that both lists are sorted. This works as well though, so not really necessary to change.

repos/system_upgrade/common/actors/gpgpubkeycheck/libraries/gpgpubkeycheck.py
```diff
@@ -69,10 +66,9 @@ def _report_unexpected_keys(unexpected_fps):
         ' This can be caused e.g. by a manual intervention'
         ' or by malicious attempt to hijack the upgrade process.'
         ' The unexpected keys are the following:'
-        ' {sep}{key_list}'
+        ' {key_list}'
```
matejmatuska: Again, I think there is an extra space, but not a blocker:

```suggestion
        '{key_list}'
```

repos/system_upgrade/common/actors/gpgpubkeycheck/libraries/gpgpubkeycheck.py
matejmatuska, line 1: I am thinking about whether these reports should sort the lists, but effectively it doesn't really matter since all the items begin with a fingerprint which cannot be meaningfully sorted. Maybe a better idea would be to sort by the packager, but it's not a blocker, it wasn't like that before.

PeterMocary (changes requested):
So far looks ok to me, but there are some things to change, however not all comments need to be addressed (namely the ssh lists and source/target versions are ok as is - I just wanted to show my line of thinking here).

I found these places that were missed:

### Reports

1. [rhsm.py:228](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/libraries/rhsm.py#L228) — duplicate repos `reporting.Summary`
2. [userspacegen.py:799](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py#L799) — duplicate repos `reporting.Summary`
3. [checkdddd.py:35](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/actors/checkdetecteddevicesanddrivers/libraries/checkdddd.py#L35) — removed drivers `reporting.Summary`
4. [checkdddd.py:74](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/actors/checkdetecteddevicesanddrivers/libraries/checkdddd.py#L74) — removed devices `reporting.Summary`
5. [checkdddd.py:99](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/actors/checkdetecteddevicesanddrivers/libraries/checkdddd.py#L99) — removed processors `reporting.Summary`
6. [checkdddd.py:128](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/actors/checkdetecteddevicesanddrivers/libraries/checkdddd.py#L128) — deprecated drivers `reporting.Summary`
7. [checkdddd.py:151](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/actors/checkdetecteddevicesanddrivers/libraries/checkdddd.py#L151) — deprecated devices `reporting.Summary`
8. [checkdddd.py:174](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/actors/checkdetecteddevicesanddrivers/libraries/checkdddd.py#L174) — deprecated processors `reporting.Summary`
9. [repositoriesblocklist.py:18](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/actors/targetcontentresolver/libraries/repositoriesblocklist.py#L18) — excluded repos `reporting.Summary`
10. [checkinitramfstasks.py:18](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/actors/initramfs/checkinitramfstasks/libraries/checkinitramfstasks.py#L18) — conflicting initramfs modules `reporting.Summary`

### Logs

11. [rhsm.py:198](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/libraries/rhsm.py#L198) — available RHSM repos `logger.info`
12. [rhsm.py:221](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/libraries/rhsm.py#L221) — duplicate repos `logger.warning`
13. [userspacegen.py:791](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py#L791) — duplicate repos `logger.warning`
14. [rpmtransactionconfigtaskscollector.py:33](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/actors/rpmtransactionconfigtaskscollector/libraries/rpmtransactionconfigtaskscollector.py#L33) — filtered install packages `logger.debug`

repos/system_upgrade/common/actors/checkosrelease/libraries/checkosrelease.py
matejmatuska: The SUPPORTED_VERSIONS dict which is used here is deprecated and is going to be replaced. Aside from that, yeah we keep it sorted there and will keep doing that.

matejmatuska (changes requested):
I review at the same time as @PeterMocary hopefully I removed all duplicate comments. I found some mostly minor issues.

repos/system_upgrade/common/actors/checkdynamiclinkerconfiguration/libraries/checkdynamiclinkerconfiguration.py
prjagtap-rh: The "second separator" isn't removed/lost it's baked into format_list's behaviour.

repos/system_upgrade/common/actors/checkosrelease/libraries/checkosrelease.py
prjagtap-rh: @PeterMocary
Agreed, the default `sorted()` would break version ordering here so keeping `callback_sort=None` is intentional. Since the list is already pre-sorted, adding a custom version-aware sort would be unnecessary complexity.

repos/system_upgrade/el9toel10/actors/sssd/sssdchecks/libraries/sssdchecks.py
prjagtap-rh: @PeterMocary 
Acknowledged, keeping it as-is for now since it works. Can revisit if needed.

repos/system_upgrade/common/actors/checkdynamiclinkerconfiguration/libraries/checkdynamiclinkerconfiguration.py
PeterMocary: The formatting didn't change - there are 2 separators, first one `\n-` and the second one is default in `format_list` with 4 spaces before the bullet point `\n    -`. Thus, I'm marking as resolved.

PeterMocary (changes requested):
LGTM! However, one of the suggested places was missed and I've found some additional:

1. [checkinitramfstasks.py:18](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/actors/initramfs/checkinitramfstasks/libraries/checkinitramfstasks.py#L18) — conflicting initramfs modules `reporting.Summary` (mentioned before already)
2. [checksaphana.py:86](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/actors/checksaphana/libraries/checksaphana.py#L86) — SAP HANA instance list `reporting.Summary`
3. [unsupportedupgradecheck/actor.py:49](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/actors/unsupportedupgradecheck/actor.py#L49) — development variables list `reporting.Summary`
4. [userspacegen.py:1114](https://github.com/oamg/leapp-repository/blob/4963bc98/repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py#L1114) — missing custom repo IDs `reporting.Summary`

PeterMocary:
Removed unnecessary white space and fixed the 4 missed lists.

PeterMocary:
Turned the sorting off for lists where sorting does not make a difference.

repos/system_upgrade/common/actors/gpgpubkeycheck/libraries/gpgpubkeycheck.py
PeterMocary: I turned the sorting off. If the operation does not make sense it should not be done.

repos/system_upgrade/common/actors/gpgpubkeycheck/libraries/gpgpubkeycheck.py
PeterMocary: Fixed this and multiple other places in my recent changes.

PeterMocary (approved):
LGTM now, I've added the 4 missed places, fixed the white space and made lists with gpg keys unsorted as that doesn't make a difference really.

We should wait for second approve before merging - @matejmatuska can you review please?

matejmatuska (approved):
Seems like everything is adresssed and the new changes LGTM.

PR #1584, opened 2026-08-19
Warn when Postfix still uses Berkeley DB hash/btree maps

Amodak9801:
## Summary
- RHEL 10 removed Berkeley DB, so Postfix `hash:` / `btree:` lookup tables fail after a 9→10 in-place upgrade (`unsupported dictionary type: hash`).
- Add a Facts-phase scanner and a Checks-phase reporter that warn when `/etc/postfix/*.cf` still uses those map types. Systems already converted to `lmdb:` are left alone.
- Configuration is **not** rewritten automatically (Postfix layouts are too varied). The report points at [KCS 7131247](https://access.redhat.com/solutions/7131247).

Jira: [RHEL-114756](https://redhat.atlassian.net/browse/RHEL-114756)

## Test plan
- [ ] `pytest repos/system_upgrade/el9toel10/actors/postfix/scanpostfixbdb/tests repos/system_upgrade/el9toel10/actors/postfix/checkpostfixbdb/tests` (10 unit tests)
- [ ] On a RHEL 9 system with default Postfix (`alias_maps = hash:/etc/aliases`), `leapp preupgrade` shows a high-risk report with the KCS link
- [ ] After converting maps to `lmdb:`, the report is not produced

Made with [Cursor](https://cursor.com)

[RHEL-114756]: https://redhat.atlassian.net/browse/RHEL-114756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/repos/system_upgrade/el9toel10/actors/postfix/checkpostfixbdb/libraries/checkpostfixbdb.py
+++ b/repos/system_upgrade/el9toel10/actors/postfix/checkpostfixbdb/libraries/checkpostfixbdb.py
@@ -1,0 +1,58 @@
+from leapp import reporting
+from leapp.exceptions import StopActorExecutionError
+from leapp.libraries.common.distro import DISTRO_REPORT_NAMES
+from leapp.libraries.stdlib import api, format_list
+from leapp.models import PostfixBdbConfiguration
+
+REPORT_KCS_URL = 'https://access.redhat.com/solutions/7131247'
+
+
+def process():
+    msg = next(api.consume(PostfixBdbConfiguration), None)
+    if not msg:
+        raise StopActorExecutionError('Expected PostfixBdbConfiguration, but got None')
+
+    if not msg.postfix_present:
+        api.current_logger().debug('postfix package not found, no report generated')
+        return
+
+    if not msg.bdb_occurrences:
+        api.current_logger().debug(
+            'Postfix is installed but no hash:/btree: maps were found; no report generated'
+        )
+        return
+
+    summary = (
+        'RHEL 10 no longer provides Berkeley DB, so Postfix cannot use the '
+        'hash: or btree: lookup table types after the upgrade. Postfix will '
+        'fail to start or to rebuild aliases (unsupported dictionary type: hash) '
+        'until the maps are converted to lmdb. '
+        'The following configuration entries still use a Berkeley DB map type:'
+        '{occurrences}'
+    ).format(
+        occurrences=format_list(msg.bdb_occurrences, callback_sort=None),
+    )
+
+    hint = (
+        'Do not rely on Leapp to rewrite Postfix configuration. Before or after '
+        'the upgrade, replace hash: and btree: with lmdb: in /etc/postfix/main.cf '
+        '(and related files), set default_database_type = lmdb, and rebuild maps '
+        'with postmap/postalias. See the linked article for the full procedure. '
+        'Target system: {target_distro} 10.'
+    ).format_map(DISTRO_REPORT_NAMES)
+
+    reporting.create_report([
+        reporting.Title(
+            'Postfix Berkeley DB (hash/btree) maps are not available after the upgrade'
+        ),
+        reporting.Summary(summary),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.SERVICES]),
+        reporting.ExternalLink(
+            title='Postfix fails with unsupported dictionary type: hash after upgrading to RHEL 10',
+            url=REPORT_KCS_URL,
+        ),
+        reporting.RelatedResource('package', 'postfix'),
+        reporting.RelatedResource('file', '/etc/postfix/main.cf'),
+        reporting.Remediation(hint=hint),
+    ])
--- a/repos/system_upgrade/el9toel10/actors/postfix/scanpostfixbdb/libraries/scanpostfixbdb.py
+++ b/repos/system_upgrade/el9toel10/actors/postfix/scanpostfixbdb/libraries/scanpostfixbdb.py
@@ -1,0 +1,71 @@
+import os
+import re
+
+from leapp.libraries.common.rpms import has_package
+from leapp.libraries.stdlib import api
+from leapp.models import DistributionSignedRPM, PostfixBdbConfiguration
+
+DEFAULT_POSTFIX_PATHS = ['/etc/postfix']
+
+# Map type used as a prefix: hash:/etc/aliases, proxy:hash:/path, ...
+BDB_MAP_RE = re.compile(r'(?:^|[\s,=:])(hash|btree):', re.IGNORECASE)
+# Compiled default on RHEL 9 is hash; an explicit setting of hash/btree is also BDB.
+DEFAULT_DB_RE = re.compile(r'^\s*default_database_type\s*=\s*(hash|btree)\b', re.IGNORECASE)
+
+
+def _iter_config_files(paths):
+    """Yield Postfix *.cf files from the given files or directories."""
+    for path in paths:
+        if os.path.isdir(path):
+            try:
+                names = os.listdir(path)
+            except OSError as err:
+                api.current_logger().warning(
+                    'Could not list Postfix configuration directory {}: {}'.format(path, err)
+                )
+                continue
+            for name in sorted(names):
+                if name.endswith('.cf'):
+                    yield os.path.join(path, name)
+        elif os.path.isfile(path):
+            yield path
+
+
+def _scan_file(path):
+    """Return list of 'path: line' strings for Berkeley DB map settings."""
+    occurrences = []
+    try:
+        with open(path) as conf_file:
+            for line in conf_file:
+                stripped = line.strip()
+                if not stripped or stripped.startswith('#'):
+                    continue
+                if BDB_MAP_RE.search(stripped) or DEFAULT_DB_RE.search(stripped):
+                    occurrences.append('{}: {}'.format(path, stripped))
+    except OSError as err:
+        api.current_logger().warning(
+            'Could not read Postfix configuration file {}: {}'.format(path, err)
+        )
+    return occurrences
+
+
+def scan_postfix_configuration(conf_paths=None, _context=api):
+    """
+    Scan Postfix configuration for Berkeley DB (hash/btree) lookup tables.
+
+    :param conf_paths: Files or directories to scan. Defaults to /etc/postfix.
+    :return: PostfixBdbConfiguration
+    """
+    postfix_present = has_package(DistributionSignedRPM, 'postfix', context=_context)
+    if not postfix_present:
+        return PostfixBdbConfiguration(postfix_present=False, bdb_occurrences=[])
+
+    paths = DEFAULT_POSTFIX_PATHS if conf_paths is None else conf_paths
+    occurrences = []
+    for conf_file in _iter_config_files(paths):
+        occurrences.extend(_scan_file(conf_file))
+
+    return PostfixBdbConfiguration(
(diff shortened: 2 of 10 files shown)


Amodak9801:

repos/system_upgrade/el9toel10/actors/postfix/checkpostfixbdb/libraries/checkpostfixbdb.py
```diff
@@ -0,0 +1,58 @@
...
+        occurrences=format_list(msg.bdb_occurrences, callback_sort=None),
+    )
+
+    hint = (
+        'Do not rely on Leapp to rewrite Postfix configuration. Before or after '
+        'the upgrade, replace hash: and btree: with lmdb: in /etc/postfix/main.cf '
+        '(and related files), set default_database_type = lmdb, and rebuild maps '
+        'with postmap/postalias. See the linked article for the full procedure. '
+        'Target system: {target_distro} 10.'
+    ).format_map(DISTRO_REPORT_NAMES)
```
PeterMocary, line 42:
```suggestion
    hint = (
        'Replace hash: and btree: with lmdb: in /etc/postfix/main.cf '
        '(and related files), set default_database_type = lmdb, and rebuild maps '
        'with postmap/postalias. See the linked article for the full procedure.'
    )
```

repos/system_upgrade/el9toel10/actors/postfix/checkpostfixbdb/libraries/checkpostfixbdb.py
```diff
@@ -0,0 +1,58 @@
...
+    summary = (
+        'RHEL 10 no longer provides Berkeley DB, so Postfix cannot use the '
+        'hash: or btree: lookup table types after the upgrade. Postfix will '
+        'fail to start or to rebuild aliases (unsupported dictionary type: hash) '
+        'until the maps are converted to lmdb. '
+        'The following configuration entries still use a Berkeley DB map type:'
+        '{occurrences}'
+    ).format(
+        occurrences=format_list(msg.bdb_occurrences, callback_sort=None),
+    )
```
PeterMocary, line 34:
```suggestion
  summary = (
      '{target_distro} 10 no longer provides Berkeley DB, so Postfix cannot '
      'use the hash: or btree: lookup table types after the upgrade. Postfix will '
      'fail to start or to rebuild aliases (unsupported dictionary type: hash) '
      'until the maps are converted to lmdb. '
      'The following configuration entries still use a Berkeley DB map type:'
      '{occurrences}'
  ).format(
      occurrences=format_list(msg.bdb_occurrences, callback_sort=None),
      **DISTRO_REPORT_NAMES,
  )
```

PeterMocary: Is the list guaranteed to be exhaustive? If not, we should note that in the summary as well.

repos/system_upgrade/el9toel10/actors/postfix/scanpostfixbdb/libraries/scanpostfixbdb.py
```diff
@@ -0,0 +1,71 @@
...
+    occurrences = []
+    try:
+        with open(path) as conf_file:
+            for line in conf_file:
+                stripped = line.strip()
+                if not stripped or stripped.startswith('#'):
+                    continue
+                if BDB_MAP_RE.search(stripped) or DEFAULT_DB_RE.search(stripped):
+                    occurrences.append('{}: {}'.format(path, stripped))
+    except OSError as err:
```
PeterMocary, line 45: Wouldn't it be worth handling the UnicodeDecodeError as well since we are reading from a file?

repos/system_upgrade/el9toel10/actors/postfix/scanpostfixbdb/libraries/scanpostfixbdb.py
```diff
@@ -0,0 +1,71 @@
...
+            except OSError as err:
+                api.current_logger().warning(
+                    'Could not list Postfix configuration directory {}: {}'.format(path, err)
+                )
+                continue
+            for name in sorted(names):
+                if name.endswith('.cf'):
+                    yield os.path.join(path, name)
+        elif os.path.isfile(path):
+            yield path
```
PeterMocary, line 31: The doc string states only `*.cf` files are yielded but here it is not filtered.

repos/system_upgrade/el9toel10/actors/postfix/scanpostfixbdb/libraries/scanpostfixbdb.py
PeterMocary, line 1: This scanner wouldn't detect settings in `master.cf` since there is a different syntax used. Currently, if one would have a special configuration for a specific service which would involve the database types no longer supported in RHEL10, the same issue would be encountered without the report listing these. If such configuration would be the only occurrence of these database types, it wouldn't even be reported at all.

The `-o name=value` overrides  the  named main.cf configuration parameter from the `master.cf` as is stated in [man pages](https://www.postfix.org/master.5.html). 

Since detecting this is not a big change, I'd suggest detecting it.

repos/system_upgrade/el9toel10/actors/postfix/scanpostfixbdb/libraries/scanpostfixbdb.py
```diff
@@ -0,0 +1,71 @@
+import os
+import re
+
+from leapp.libraries.common.rpms import has_package
+from leapp.libraries.stdlib import api
+from leapp.models import DistributionSignedRPM, PostfixBdbConfiguration
+
+DEFAULT_POSTFIX_PATHS = ['/etc/postfix']
```
PeterMocary, line 8: This path is default, however, the configuration allows to specify alternative paths as well. On top of that, there is also [Postfix multi-instance](https://www.postfix.org/postmulti.1.html) which has multiple configuration paths. Thus, we would not be able to detect all possible occurrences. 

I don't want to extend the scanning, this is just an additional reason to add a note to the report about the fact that the list of detected occurrences is not exhaustive.

repos/system_upgrade/el9toel10/actors/postfix/checkpostfixbdb/libraries/checkpostfixbdb.py
```diff
@@ -0,0 +1,58 @@
...
+        ),
+        reporting.Summary(summary),
+        reporting.Severity(reporting.Severity.HIGH),
+        reporting.Groups([reporting.Groups.SERVICES]),
+        reporting.ExternalLink(
+            title='Postfix fails with unsupported dictionary type: hash after upgrading to RHEL 10',
+            url=REPORT_KCS_URL,
+        ),
+        reporting.RelatedResource('package', 'postfix'),
+        reporting.RelatedResource('file', '/etc/postfix/main.cf'),
```
PeterMocary, line 56: This might be misleading if the occurrences are in some other config file only (e.g. `master.cf`). I wouldn't include the file as related resource at all.

repos/system_upgrade/el9toel10/actors/postfix/scanpostfixbdb/libraries/scanpostfixbdb.py
```diff
@@ -0,0 +1,71 @@
...
+
+# Map type used as a prefix: hash:/etc/aliases, proxy:hash:/path, ...
+BDB_MAP_RE = re.compile(r'(?:^|[\s,=:])(hash|btree):', re.IGNORECASE)
+# Compiled default on RHEL 9 is hash; an explicit setting of hash/btree is also BDB.
+DEFAULT_DB_RE = re.compile(r'^\s*default_database_type\s*=\s*(hash|btree)\b', re.IGNORECASE)
```
PeterMocary, line 13: I'm not sure if the `re.IGNORECASE` wouldn't just allow for false positives. Is it stated somewhere in the documentation that these settings could use upper case at all?

repos/system_upgrade/el9toel10/actors/postfix/scanpostfixbdb/tests/test_scanpostfixbdb.py
PeterMocary, line 1: We usually unit-test the helper functions as well, currently, these are tested just indirectly. The `isdir` branch in `_iter_config_files` is not tested now. It is worth noting that there is also the `leapp_tmpdir` fixture one can use for testing with files and directories.

PeterMocary (changes requested):
Hello, thank you for the contribution! I went through the changes and prepared some suggestions we need to address before I can approve these changes. Among which is increasing the unit-test coverage a bit.

We should definitely make it clear that leapp can't detect all of the problematic occurrences in the configuration and that the user should make sure they migrate their entire configuration. Alternatively we could just stop detecting these occurrences all together, since we can't possibly provide an exhaustive list. However, if we decide to stick with the best effort detection, we need to focus on the `master.cf` file as well - it has different syntax which would not match the regex in current solution.

It is also worth noting how the configuration changes between the RHEL versions:

RHEL 9 (default):
- `default_database_type` not set in the `main.cf` 
- default for `postconf -d default_database_type` is `default_database_type = hash`
- `hash:` is used on multiple lines
- `master.cf` does not contain either `hash:` or `default_database_type`

```bash
[root@rhel9 ~]> cat /etc/postfix/main.cf | grep -E "hash:|btree:|lmdb:|default_database_type"
#mynetworks = hash:/etc/postfix/network_table
#relay_recipient_maps = hash:/etc/postfix/relay_recipients
alias_maps = hash:/etc/aliases
#alias_maps = hash:/etc/aliases, nis:mail.aliases
alias_database = hash:/etc/aliases
#alias_database = hash:/etc/aliases, hash:/opt/majordomo/aliases
```
RHEL 10 (default)
- `default_database_type` is set in the `main.cf` to `lmdb`
- default for `postconf -d default_database_type` is `default_database_type = hash`
- `lmdb:` is used on multiple lines
-  `hash:` is not used
- `master.cf` does not contain either

```bash
[root@rhel10 ]> cat /etc/postfix/main.cf | grep -E "hash:|btree:|lmdb:|default_database_type"
#mynetworks = hash:/etc/postfix/network_table
#relay_recipient_maps = hash:/etc/postfix/relay_recipients
alias_maps = lmdb:/etc/aliases
#alias_maps = hash:/etc/aliases, nis:mail.aliases
alias_database = lmdb:/etc/aliases
#alias_database = hash:/etc/aliases, hash:/opt/majordomo/aliases
default_database_type = lmdb
```
The hash entries are commented out in the configuration file as the output suggests.

Amodak9801:
Thanks for the review, @PeterMocary. Addressed in bb5bc103:

- Report summary now uses `{target_distro} 10` and states that the listed hits are **not exhaustive** (custom config paths and Postfix multi-instance are not scanned). Hint wording matches your suggestion (`lmdb:`, `default_database_type`, `postmap`/`postalias`).
- Scanner now detects `master.cf` `-o name=value` overrides, including `-o default_database_type=hash`.
- `UnicodeDecodeError` is handled when reading config files.
- `_iter_config_files` only yields `*.cf` files (directories and explicit paths).
- Dropped `RelatedResource` for `/etc/postfix/main.cf`.
- Dropped `re.IGNORECASE`; Postfix map types are lowercase, so `HASH:` is not treated as a hit.
- Unit tests cover the helpers directly (`isdir` branch, `leapp_tmpdir`), `master.cf` `-o` lines, decode errors, and the updated report text.

Did not extend scanning to custom paths or `postmulti` instances, per your note.

PR #1585, opened 2026-08-20
Add Oracle Linux  handling for upgrade and conversion

karolinku:
Enable Oracle Linux (ol) as a recognized IPU distribution: distro ID, GPG keys, repo files, upgrade paths, and OL→RHEL package swaps.

Some OL packages (higher epoch or unversioned Obsoletes/Provides) block their RHEL counterparts, so dnf treats the target RPM as already installed. Add swap_distro_packages_workaround to download the target builds and force-swap them in a single dnf shell transaction before the upgrade.

Jira: RHEL-187015, [RHEL-224276](https://redhat.atlassian.net/browse/RHEL-224276)

[RHEL-224276]: https://redhat.atlassian.net/browse/RHEL-224276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


repos/system_upgrade/common/actors/convert/swapdistropackagesworkaround/libraries/swapdistropackagesworkaround.py
```diff
@@ -0,0 +1,214 @@
...
+from leapp.libraries.common.config import get_source_distro_id, get_target_distro_id
+from leapp.libraries.common.config.version import get_target_major_version, get_target_version
+from leapp.libraries.common.gpg import is_nogpgcheck_set
+from leapp.libraries.stdlib import api, CalledProcessError, config
+from leapp.models import DistributionSignedRPM, DNFWorkaround, TargetUserSpaceInfo, UsedTargetRepositories
+
+
+SWAP_DATA_DIR = '/var/lib/leapp/swap-workaround-data'
+DNF_SHELL_INSTRUCTIONS_PATH = os.path.join(SWAP_DATA_DIR, 'dnfshellinstructions')
+_CONTAINER_DATA_DIR = '/swap-workaround-data'
```
pirat89: here

repos/system_upgrade/common/actors/convert/swapdistropackagesworkaround/libraries/swapdistropackagesworkaround.py
```diff
@@ -0,0 +1,214 @@
...
+    if packages_to_remove:
+        lines.append('remove {}'.format(' '.join(packages_to_remove)))
+    lines.append('run')
+    return '\n'.join(lines) + '\n'
+
+
+def _write_instructions(content):
+    with open(DNF_SHELL_INSTRUCTIONS_PATH, 'w') as f:
+        f.write(content)
+    api.current_logger().info(
```
pirat89: I would not print it to user, it's just for debugging purposes.

```suggestion
    api.current_logger().debug(
```

repos/system_upgrade/common/actors/convert/swapdistropackagesworkaround/libraries/swapdistropackagesworkaround.py
```diff
@@ -0,0 +1,214 @@
...
+            'Cannot prepare the distribution package swap workaround: '
+            'missing target userspace or target repositories information.'
+        )
+        return
+
+    packages_to_download = []
+    for entry in active_swaps:
+        packages_to_download.extend(entry['install'])
+    # de-duplicate while preserving order
+    packages_to_download = list(dict.fromkeys(packages_to_download))
```
pirat89: Checking the rest of the code, it seems safe to work with sets:

```suggestion
    packages_to_download = set()
    for entry in active_swaps:
        packages_to_download.update(entry['install'])
```

 The `list(dict.fromkeys(..))` construct is not ideal way how to deduplicate stuff. In case you have a `list` and want to sure you deduplicate everything in it, this is better way: `list(set(my_list))`

repos/system_upgrade/common/actors/convert/swapdistropackagesworkaround/libraries/swapdistropackagesworkaround.py
```diff
@@ -0,0 +1,214 @@
...
+    Pick the swap entries that apply to the current system.
+
+    An entry applies when any of the packages it removes - other than the target
+    packages it installs under a different name - is actually installed, i.e. a
+    source-distro specific package that needs to be swapped out is present.
+    """
+    active = []
+    for entry in swap_config:
+        triggers = [pkg for pkg in entry["remove"] if pkg not in entry["install"]]
+        if any(pkg in installed_pkgs for pkg in triggers):
```
pirat89: Discussed together, we could not came for the idea why the filtered `triggers` is needed and it's confusing part of the code. Let's simplify it unless we came up for the reason why to filter entries in "remove" lists (in such a case, let's add a comment to explaining it so we understand the case in future).

```suggestion
        if any(pkg in installed_pkgs for pkg in entry["remove"]):
```

Update please the docstring as well.

pirat89 (changes requested):

repos/system_upgrade/common/libraries/distro.py
```diff
@@ -349,4 +368,7 @@ def get_distro_efidir_canon_path(distro_id):
     if distro_id == "rhel":
         return os.path.join(efi.EFI_MOUNTPOINT, "EFI", "redhat")
 
+    if distro_id == "ol":
+        return os.path.join(efi.EFI_MOUNTPOINT, "EFI", "oracle")
```
matejmatuska: Seems like it's `redhat` on OL (I checked on OL 9.8).
I don't know about OL 8 (yet).

```suggestion
        return os.path.join(efi.EFI_MOUNTPOINT, "EFI", "redhat")
```

repos/system_upgrade/common/actors/convert/swapdistropackagesworkaround/libraries/swapdistropackagesworkaround.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+    Create a clean data directory for the downloaded RPMs and instructions.
+    """
+    if os.path.isdir(SWAP_DATA_DIR):
+        for name in os.listdir(SWAP_DATA_DIR):
+            path = os.path.join(SWAP_DATA_DIR, name)
+            if os.path.isfile(path):
+                os.unlink(path)
+    else:
+        os.makedirs(SWAP_DATA_DIR)
```
matejmatuska: IIUC the purpose of the function is to ensure that the SWAP_DATA_DIR exists and is empty, why not use shutil.rmtree and then mkdir unconditionally?

repos/system_upgrade/common/actors/convert/swapdistropackagesworkaround/libraries/swapdistropackagesworkaround.py
```diff
@@ -0,0 +1,214 @@
...
+        raise StopActorExecutionError(
+            "Cannot download packages for the distribution swap workaround: "
+            "no target repositories are available."
+        )
+
+    binds = ['{}:{}'.format(SWAP_DATA_DIR, _CONTAINER_DATA_DIR)]
+    repos_opt = list(itertools.chain(*[['--enablerepo', repo] for repo in target_repoids]))
+    cmd = [
+        'dnf',
+        'download',
```
matejmatuska, line 93: This is a part of `dnf-plugins-core` package, maybe we should add it as a dep? It think that it is expected to be installed though.

repos/system_upgrade/common/actors/convert/swapdistropackagesworkaround/libraries/swapdistropackagesworkaround.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+def process():
+    source_distro = get_source_distro_id()
+    target_distro = get_target_distro_id()
+    if source_distro == target_distro:
+        return
+
+    if source_distro not in _SUPPORTED_SOURCE_DISTROS:
+        return
```
matejmatuska: I am not sure what is the purpose of `_SUPPORTED_SOURCE_DISTROS`?
If a (source_distro, target_distro) is not in the config it, the next `if` will return early. I would remove this condition and `_SUPPORTED_SOURCE_DISTROS` and just add a log to the next one.

Let's keep the number of places where distro specific stuff to minimum, it's already across several files :/

repos/system_upgrade/common/libraries/distro.py
karolinku: you are right. I confirmed that OL8 has also "redhat".

repos/system_upgrade/common/actors/convert/swapdistropackagesworkaround/libraries/swapdistropackagesworkaround.py
karolinku: Discussed in side thread - the requirement will be added as always-installed in userspacegen creation actor. However, in most cases `dnf download` is pulled inexplicitely as dep anyway.

repos/system_upgrade/common/actors/targetuserspacecreator/libraries/userspacegen.py
```diff
@@ -101,7 +101,7 @@ def _consume_data(self):
         It doesn't consume TargetRepositories, which are consumed in the
         own function.
         """
-        self.packages = {'dnf', 'dnf-command(config-manager)', 'util-linux'}
+        self.packages = {'dnf', 'dnf-command(config-manager)', 'dnf-command(download)', 'util-linux'}
```
matejmatuska: I think a comment explaining why the download is there would be great, also mentioning it in the commit mesage.

repos/system_upgrade/common/actors/convert/swapdistropackages/libraries/swapdistropackages.py
```diff
@@ -73,6 +76,24 @@
...
+            "oracle-logos-ipa": "redhat-logos-ipa",
+            "oraclelinux-release": "redhat-release",
+        },
+        "remove": {
+            "oraclelinux-release-el*",
+            "oraclelinux-*-release-*",
+        },
+        "install": {
+            "plymouth-theme-spinner",
+        },
```
pirat89, line 98: actually, it would be usefull to have an info here why the installation of plymouth-theme-spinner is needed. currently I see it documented in tests.

repos/system_upgrade/common/files/upgrade_paths.json
```diff
@@ -48,5 +48,15 @@
             "8": ["9.6", "9.8", "9.9"],
             "9": ["10.3"]
         }
+    },
+    "ol": {
+        "default": {
+            "8.10": ["9.6", "9.8", "9.9"],
+            "9.6": ["10.0"],
+            "9.8": ["10.2"],
+            "9.9": ["10.3"],
```
pirat89, line 57: nitpick, OL has basically just the latest version, so they do not have 9.6 & 9.8. but it's ok to keep it as it is now, we will change it anyway soon due to 9.10 & 10.4

repos/system_upgrade/common/actors/convert/swapdistropackagesworkaround/libraries/swapdistropackagesworkaround.py
```diff
@@ -0,0 +1,228 @@
...
+    only accepts local packages while no transaction job exists yet, else it
+    rejects them (https://bugzilla.redhat.com/show_bug.cgi?id=1773483).
+    Install-first also lets the RHEL replacements provide the capabilities of
+    the removed packages, so the swap resolves as one atomic transaction.
+    """
+    lines = []
+    if rpm_paths:
+        lines.append('install {}'.format(' '.join(rpm_paths)))
+    if packages_to_remove:
+        lines.append('remove {}'.format(' '.join(packages_to_remove)))
```
pirat89, line 145: @karolinku  Currently, this result into a file like:
```
install rpm_path1 rpm_path2 rpm_path3 rpm_path4
remove rpm1 rpm2 rpm3
run

```

which is not so much friendly for reading. In this case I suggest to change it for instruction per line, so the result would be:
```
install /var/lib/leapp/swap-workaround-data/redhat-backgrounds-100.3-2.el10.noarch.rpm
install /var/lib/leapp/swap-workaround-data/redhat-indexhtml-10-4.el10_0.noarch.rpm
install /var/lib/leapp/swap-workaround-data/redhat-logos-100.3-2.el10.x86_64.rpm
remove oracle-backgrounds
remove oracle-indexhtml
remove oracle-logos
run
```

TBD
```suggestion
    if rpm_paths:
        lines.extend([f'install {rpm_path}' for rpm_path in rpm_paths])
    if packages_to_remove:
        lines.extend([f'remove {rpm}' for rpm in packages_to_remove])
```

pirat89: checked together. due to bug in DNF, all install packages must be listed on the first line.

repos/system_upgrade/common/tools/dnfshellswap
```diff
@@ -0,0 +1,34 @@
...
+instructions_file="$1"
+shift
+
+if [ ! -f "$instructions_file" ]; then
+    log_error "The dnf shell instructions file ${instructions_file} does not exist, nothing to do."
+    exit 1
+fi
+
+log_info "Applying distribution package swap using ${instructions_file}"
+dnf -y "$@" shell "$instructions_file"
```
pirat89:
```suggestion
dnf -y "$@" shell "$instructions_file" | tee /tmp/dnfshellswap.log
grep -qiE "error|conflict|broken|problem" /tmp/dnfshellswap.log && {
    log_error "Could not perform swap of packages"
    exit 1
}

exit 0
```

pirat89:
@karolinku  on OL 8 also: oraclelinux-release-el8-1.0-38.el8.x86_64

pirat89 (dismissed):
lgtm.  I am still running some tests, so I will merge it after I finish these runs, but I expect it will work based on preivous testing. Good job! 🚀

pirat89:
I found 2 another problems:
-  broken rpm after the conversion - most likely incorrect specification or handling of the GPG key replacements
- upgrade is broken for OL 8 -> RHEL 9 conversion when oracle-logos-httpd is installed - most likely it just needs to add a swap package into the dict

I do not consider the last problem blocking, but the first one could be. postponing the merge for now until we check that.

repos/system_upgrade/common/files/distro/ol/gpg-signatures.json
```diff
@@ -0,0 +1,7 @@
+{
+    "keys": {
+        "82562ea9ad986da3": [],
+        "bc4d06a08d8b756f": []
+    },
+    "obsoleted-keys": {}
+}
```
pirat89, line 7: I haven't tested it yet, but based on keys I found, this should be the content of the file:

```suggestion
{
    "keys": {
        "82562ea9ad986da3": ["gpg-pubkey-ad986da3-5cabf60d"],
        "bc4d06a08d8b756f": ["gpg-pubkey-8d8b756f-629e59ec"],
        "72f97b74ec551f03": ["gpg-pubkey-ec551f03-53619141"]
    },
    "obsoleted-keys": {
        "8": ["gpg-pubkey-ec551f03-53619141"],
        "9": ["gpg-pubkey-ad986da3-5cabf60d"],
        "10": []
    }
}
```

Note I have covered just keys for OL 7, 8, and 9. I haven't checked OL 10 keys yet. so there is possibly still missing something, especially for OL 9 -> OL 10.

pirat89:
lgtm. I will re-test it and merge it unless I find anything blocking. but based on changes, everything should be working now.

repos/system_upgrade/common/files/distro/ol/gpg-signatures.json
pirat89: applied

pirat89 (approved):
original problems are fixed. I discovered that OL 8 -> RHEL 9 conversion is failing due to crashing scriptlets related to some OL specific RPMs. But this is problem out of scope of this PR - we will address it separately. merging.

PR #1587, opened 2026-08-22
Make the RHEL 7 libssh2 Leapp upgrade path deterministic

dwivedipranav-dev:
## Summary
- Event 103 SPLIT (`libssh2` → `libssh` + modular `virt:rhel` `libssh2`) is not a regular RHEL 8 AppStream package. Hosts classified as `rhel7-base` vs `rhel7-optional` therefore finish with different RPMs (leftover el7 libssh2 + compat-openssl10 vs libssh only).
- Change event 103 to REPLACED (`libssh2` → `libssh` only) and add the same mapping for `rhel7-optional`.
Jira: https://redhat.atlassian.net/browse/RHEL-246089
## Test plan
- [ ] RHEL 7.9 host with `libssh2` from base: preupgrade maps it to `libssh` only
- [ ] Same NVR classified as `rhel7-optional`: same mapping
- [ ] After upgrade: `libssh` present, no `libssh2-*.el7`, no extra `compat-openssl10` pulled only by leftover libssh2


dwivedipranav-dev:

pirat89:
Thank you for the contribution, but the data for 7 -> 8 are rather archived. The project does not support upgrades 7 -> 8 anymore and any manual change would be rewritten with the next data update. We actually plan to filter the data in future to drop all el7 releated data at all. Closing.

PR #1588, opened 2026-08-26
mount_units_generator: Preserve parent units for nested dracut mounts

prjagtap-rh:
## The reason for the change

Systems with nested mounts under paths already mounted by dracut (e.g.
/usr/local or /usr/sap under /usr) need parent mount units present in
the upgrade initramfs so systemd can order parent-before-child mounts.
The mount unit generator previously removed parent units such as
usr.mount unconditionally, which breaks that ordering for nested mounts.

## What has been changed

- `remove_units_for_targets_that_are_already_mounted_by_dracut()` now
  skips removal of a parent mount unit when child mount units exist in
  the workspace.
- New helper `_has_child_mount_units()` detects child units using
  systemd unit naming (e.g. `usr-local.mount` is a child of `usr.mount`).
- Unit tests for the helper and for preserve/remove integration behavior.

## How it has been changed

Before deleting units that dracut already mounts (`-.mount`, `usr.mount`),
the generator scans the workspace for child mount unit files whose names
start with the parent prefix (e.g. `usr-` for `usr.mount`). If any exist,
the parent unit is kept and an info log is emitted; otherwise behavior is
unchanged. Child detection relies on systemd's path-to-unit-name 
convention rather than parsing Requires= dependencies.

## Jira
RHEL-192249

### DIFF
--- a/repos/system_upgrade/common/actors/initramfs/mount_units_generator/libraries/mount_unit_generator.py
+++ b/repos/system_upgrade/common/actors/initramfs/mount_units_generator/libraries/mount_unit_generator.py
@@ -314,6 +314,20 @@ def copy_units_into_system_location(upgrade_container_ctx, dir_with_our_mount_un
     return copied_files
 
 
+def _has_child_mount_units(dir_path, parent_unit_name):
+    """
+    Check if any mount unit file in dir_path is a child of parent_unit_name.
+
+    Uses systemd unit name convention: /usr/local -> usr-local.mount is a child
+    of /usr -> usr.mount because the name starts with 'usr-'.
+    """
+    parent_prefix = parent_unit_name[:-len('.mount')] + '-'
+    for filename in os.listdir(dir_path):
+        if filename.endswith('.mount') and filename.startswith(parent_prefix):
+            return True
+    return False
+
+
 def remove_units_for_targets_that_are_already_mounted_by_dracut(dir_with_our_mount_units):
     """
     Remove mount units for mount targets that are already mounted by dracut.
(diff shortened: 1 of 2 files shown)


repos/system_upgrade/common/actors/initramfs/mount_units_generator/libraries/mount_unit_generator.py
```diff
@@ -342,6 +361,14 @@ def remove_units_for_targets_that_are_already_mounted_by_dracut(dir_with_our_mou
             api.current_logger().debug('The {} unit does not exists, no need to remove it.'.format(unit))
             continue
 
+        if (unit.endswith('.mount') and unit != '-.mount'
+                and _has_child_mount_units(dir_with_our_mount_units, unit)):
```
MichalHe: The condition is unnecessarily complex.
```suggestion
        if (unit == 'usr.mount' and _has_child_mount_units(dir_with_our_mount_units, unit)):
```

MichalHe (changes requested):
I have tested the patch on a system that has a separate /usr and /usr/local partition. 
Without this patch, the `sysroot-usr.mount` is not included in the upgrade initramfs:
```
-rw-r--r--   1 root     root          353 Jan 30  2026 usr/lib/systemd/system/sysroot-boot.mount
-rw-r--r--   1 root     root          376 Jan 30  2026 usr/lib/systemd/system/sysroot-usr-local.mount
```


With this patch, the following mount units are included in the upgrade initramfs:
```
-rw-r--r--   1 root     root          353 Jan 30  2026 usr/lib/systemd/system/sysroot-boot.mount
-rw-r--r--   1 root     root          376 Jan 30  2026 usr/lib/systemd/system/sysroot-usr-local.mount
-rw-r--r--   1 root     root          358 Jan 30  2026 usr/lib/systemd/system/sysroot-usr.moun
```
The upgrade initramfs mounts all partitions without any problems (all mount units are Active)

repos/system_upgrade/common/actors/initramfs/mount_units_generator/libraries/mount_unit_generator.py
prjagtap-rh: Makes sense. Since the child-mount handling is specifically needed for usr.mount, I'll narrow the condition to unit == 'usr.mount' as suggested.

MichalHe (approved):
LGTM

MichalHe:
All is green, merging.

PR #1593, opened 2026-09-03
Inhibit upgrade and conversion when booted into UEK

prjagtap-rh:
## The reason for the change

Oracle Linux systems booted into the Unbreakable Enterprise Kernel (UEK) cannot
safely proceed with a Leapp in-place upgrade and conversion. The workflow must
detect this condition early and inhibit with clear, Oracle Linux-specific
remediation guidance.

## What has been changed

- Added the `check_uek_kernel` ChecksPhase actor under the conversion actors path.
- Updated shared kernel library logic to recognize `kernel-uek` RPMs when resolving
  the package providing a given kernel release.
- Added unit tests for the new actor and extended kernel library tests for UEK packages.

## How it has been changed

The actor reads the booted kernel release from actor configuration and checks
whether it contains `"uek"`. When UEK is detected, it produces a HIGH-severity
inhibitor report (KERNEL, BOOT groups) telling the operator that the upgrade and
conversion cannot proceed while the system is booted into UEK, with remediation
to boot into the standard Oracle Linux kernel.

Kernel package resolution in `get_kernel_pkg_info_for_uname_r` now accepts packages
ending in `-uek` in addition to `-core`.

## Prerequisite

Testing this PR requires the Oracle Linux upgrade and conversion handling
introduced in [#1585](https://github.com/oamg/leapp-repository/pull/1585). The prerequisite PR must be available/applied before testing the
`check_uek_kernel` actor as part of the Oracle Linux upgrade and conversion workflow.

---
Jira: RHEL-251975


repos/system_upgrade/common/libraries/kernel.py
```diff
@@ -158,16 +158,22 @@ def get_kernel_pkg_info_for_uname_r(uname_r: str) -> KernelPkgInfo:
...
     for pkg_nevra in pkg_nevras:
-        # Skip packages that are clearly not kernel core packages to avoid unnecessary rpm queries
-        if '-core' not in pkg_nevra or '-modules' in pkg_nevra:
+        # Skip modules immediately
+        if '-modules' in pkg_nevra:
+            continue
+
+        # Skip packages that are clearly not kernel core or uek packages to avoid unnecessary rpm queries
+        if '-core' not in pkg_nevra and '-uek' not in pkg_nevra:
             continue
```
PeterMocary: I'd like the skip to remain as one conditional. To me, it seems like `-modules` has a different reason to be skipped when it is divided. On top of that only the second if condition keeps the original reason in its comment.

```suggestion
        # Skip packages that are clearly not kernel core or uek packages to avoid unnecessary rpm queries
        not_kernel_core_package = ('-core' not in pkg_nevra and '-uek' not in pkg_nevra) or '-modules' in pkg_nevra or
        if not_kernel_core_package:
            continue
```

But I'm not sure yet if this is even correct.

repos/system_upgrade/common/actors/convert/checkuekkernel/libraries/checkuekkernel.py
```diff
@@ -0,0 +1,32 @@
...
+        reporting.Groups([reporting.Groups.KERNEL, reporting.Groups.BOOT]),
+        reporting.Groups([reporting.Groups.INHIBITOR]),
+        reporting.Remediation(
+            hint=(
+                'To proceed with the upgrade and conversion, boot into the standard Oracle Linux kernel. '
+                'Ensure that the kernel package is installed, set the standard Oracle Linux kernel '
+                'as the default boot kernel, reboot the system.'
+            )
+        ),
+    ])
```
PeterMocary, line 39: I'd change it to 'Red Hat compatible kernel' instead of 'standard Oracle Linux Kernel' since UEK could be also a 'standard' OL kernel. However, the RHCK might need to be defined somehow and I'm not sure if it is a common term. WDYT?

repos/system_upgrade/common/actors/convert/checkuekkernel/libraries/checkuekkernel.py
```diff
@@ -0,0 +1,32 @@
+from leapp import reporting
+from leapp.libraries.stdlib import api
+
+
+def process():
```
PeterMocary, line 8: Currently, the `checkuekkernel` actor is not exclusive to conversion. We know that UEK is not on RHEL so it is implicitly conversion only and on top of that OL only. I'd rather make it explicit by adding the conditional (same as other actors in convert).

However, it depends on what we want the behavior to be if the UEK ever gets introduced to RHEL. If we are fine with it hitting this inhibitor we should word it in a more generic way (e.g. not mentioning OL directly).  I think, it is better to hit this inhibitor rather than hope that the target kernel matching pairs it correctly - namely because the package naming is, in some cases, very unpredictable.

TL;DR: We need to either make this inhibitor gated to conversion (and maybe oracle linux) or word it more generic as it now can be hit by other distros if they introduce the uek named kernel.

repos/system_upgrade/common/actors/convert/checkuekkernel/libraries/checkuekkernel.py
```diff
@@ -0,0 +1,32 @@
...
+    """Inhibit the upgrade if the system is booted into UEK."""
+    uname_r = api.current_actor().configuration.kernel
+
+    if 'uek' not in uname_r:
+        api.current_logger().debug('Not a UEK kernel (%s). Skipping.', uname_r)
+        return
```
PeterMocary, line 18: Wouldn't it be better to match `-uek` to make sure it is what we expect? 

Although, this conditional basically duplicates the kernel detection. I'd expect the `KernelInfo` message to be used instead and the detection to be handled by the actor producing this message.

If `KernelInfo` is not aware of UEK correctly then for the conversion in which the UEK is used we have a message that is not reflecting the system correctly.

Edit: I see that it is uname r which doesnt have `-` in it so disregard that.

repos/system_upgrade/common/libraries/kernel.py
PeterMocary, line 25: I'd expect the UEK to be a new `KernelType`. I think that having a robust source kernel detection, even tough it won't be used to match target kernel is good thing.

repos/system_upgrade/common/libraries/kernel.py
```diff
@@ -158,16 +158,22 @@ def get_kernel_pkg_info_for_uname_r(uname_r: str) -> KernelPkgInfo:
...
         pkg_info = get_kernel_pkg_info(pkg_nevra)
-        if pkg_info.name.endswith('-core'):
+
+        if pkg_info.name.endswith(('-core', '-uek')):
             return pkg_info
```
PeterMocary, line 171: I don't think this is correct because i can see that `kernel-uek-core` exists and /lib/modules are owned by the core package.

```bash
[vagrant@oracle8 ~]$ rpm -qa kernel*
kernel-uek-core-5.15.0-323.211.3.5.el8uek.x86_64
kernel-tools-4.18.0-553.158.1.el8_10.x86_64
kernel-tools-libs-4.18.0-553.158.1.el8_10.x86_64
kernel-headers-4.18.0-553.158.1.el8_10.x86_64
kernel-modules-4.18.0-553.158.1.el8_10.x86_64
kernel-4.18.0-553.158.1.el8_10.x86_64
kernel-uek-5.15.0-323.211.3.5.el8uek.x86_64
kernel-uek-modules-5.15.0-323.211.3.5.el8uek.x86_64
kernel-core-4.18.0-553.158.1.el8_10.x86_64

[vagrant@oracle8 ~]$ uname -r
5.15.0-323.211.3.5.el8uek.x86_64

[vagrant@oracle8 ~]$ rpm -q --whatprovides /lib/modules/5.15.0-323.211.3.5.el8uek.x86_64
kernel-uek-core-5.15.0-323.211.3.5.el8uek.x86_64
```

So I'd expect no changes here.

PeterMocary (changes requested):
I don't think the UEK kernel detection is good enough. Although, it might work, the `KernelInfo` message won't be populated correctly. I'd instead introduce new kernel type (`KernelType`) into the source kernel detection and inhibit based on that.

The inhibitor wording should also change since currently it is accessible by any distro that introduces UEK so it shouldn't mention OL directly. Alternatively, it could be gated to conversion.

repos/system_upgrade/common/actors/convert/checkuekkernel/libraries/checkuekkernel.py
prjagtap-rh: Found a [doc](https://docs.oracle.com/en/operating-systems/oracle-linux/9/boot/boot-about_linux_kernels.html#topic_drz_k3d_ltb ) from Oracle which refers Red Hat Compatible Kernel (RHCK) and we can use term RHCK instead of Standard Oracle Linux Kernel.

repos/system_upgrade/common/libraries/kernel.py
prjagtap-rh: On UEK release 7 systems `/lib/modules/<uname_r>` is indeed owned by `kernel-uek-core`, Whereas on UEK release 6 system it owned by `kernel-uek` package.

```
[root@ol810 ~]# rpm -qa| grep kernel
kernel-modules-4.18.0-372.9.1.el8.x86_64
kernel-uek-5.4.17-2136.307.3.1.el8uek.x86_64
kernel-tools-4.18.0-553.5.1.el8_10.x86_64
kernel-uek-5.4.17-2136.332.5.2.el8uek.x86_64
kernel-modules-4.18.0-553.5.1.el8_10.x86_64
kernel-4.18.0-553.5.1.el8_10.x86_64
kernel-tools-libs-4.18.0-553.5.1.el8_10.x86_64
kernel-core-4.18.0-553.5.1.el8_10.x86_64
kernel-4.18.0-372.9.1.el8.x86_64
kernel-core-4.18.0-372.9.1.el8.x86_64

[root@ol810 ~]# rpm -qf /lib/modules/5.4.17-2136.332.5.2.el8uek.x86_64
kernel-uek-5.4.17-2136.332.5.2.el8uek.x86_64

[root@ol810 ~]# rpm -q --whatprovides /lib/modules/5.4.17-2136.332.5.2.el8uek.x86_64
kernel-uek-5.4.17-2136.332.5.2.el8uek.x86_64

```

The selection logic should still handle that case:
- `kernel-uek-core` passes the pre-filter because `-core` is present in the NEVRA (line 166).
- The final check is `pkg_info.name.endswith(('-core', '-uek'))`, so `kernel-uek-core` matches via the `-core` suffix (line 170).

repos/system_upgrade/common/actors/convert/checkuekkernel/libraries/checkuekkernel.py
prjagtap-rh: I would make the conversion scope explicit.
If UEK is ever introduced on RHEL or another supported source distribution, we can revisit the scope and make the inhibitor more generic at that point.

repos/system_upgrade/common/actors/convert/checkuekkernel/libraries/checkuekkernel.py
prjagtap-rh: First, regarding -uek: Oracle usually formats their UEK release strings without the dash (e.g., 5.4.17-2011.1.2.el8uek.x86_64), so checking for 'uek' rather than '-uek' is needed to catch strings like el8uek.

For the second point, I changed the implementation to consume the KernelInfo message instead of reading configuration.kernel directly. This avoids duplicating the kernel detection here. The UEK detection should be handled while producing KernelInfo, so that the message accurately represents the running kernel.

repos/system_upgrade/common/libraries/kernel.py
prjagtap-rh: I agree that having more robust source-kernel detection can be useful. However, for this change I don't think UEK needs to be introduced as a new KernelType.
I would prefer to keep KernelType unchanged and keep the UEK handling scoped to this conversion-specific check rather than expanding the kernel type model for this use case.

repos/system_upgrade/common/libraries/kernel.py
PeterMocary: If this is an outlier from what we expect, it should be put into a comment in the code at least. One could assume all of the uek packages have it the same and they could later come to the same conclusion as I did here. Therefore having this noted in the comments would be beneficial.

PR #1594, opened 2026-09-09
sssd_facts: skip non-UTF-8 files instead of crashing

dyangzon-coder:
## Summary
  - The `sssd_facts` actor crashes with an uncaught `UnicodeDecodeError` if any non-UTF-8/binary file (e.g. a leftover Vim swap file) is present in `/etc/sssd/conf.d/`, aborting `leapp preupgrade` with no
  indication of which file caused it.
  - Adds a targeted `except UnicodeDecodeError` handler that logs the offending file path and skips it, consistent with the existing `FileNotFoundError` handling, instead of crashing the whole actor.

  Fixes RHEL-216662. Reproduces the scenario from a real support case where a stray `.swp` file left behind by an interrupted `vim` edit blocked an RHEL 9.8 → 10.2 upgrade.

  ## Test plan
  - [x] Added `test_sssdfacts__non_utf8_file_is_skipped` - a single non-UTF-8 file no longer crashes `get_facts`.
  - [x] Added `test_sssdfacts__non_utf8_file_in_directory_is_skipped` - reproduces the real scenario: a binary `.swp` file alongside a valid `sssd.conf`; confirms the valid file is still found.
  - [x] `make test_container TEST_CONTAINER=el9` passes (lint + full test suite, exit code 0).
 

  Jira: RHEL-216662


pirat89 (dismissed):
Hello, thanks for the contribution. For unit tests, use please `leapp_tmpdir` instead of  `tmp_path` - see other unit-tests for examples.

Also, in the commit msg, use `Jira:` reference instead of `Fixes:`. lgtm otherwise

dyangzon-coder:
Hello,

Thanks for the review. I have addressed the following :

-  Updated the two new unit tests from `tmp_path` to `leapp_tmpdir`, matching the existing test conventions. 
-  Updated the commit trailer from Fixes: to `Jira: RHEL-216662`.

I also verified the changes with a local containerized run: `ACTOR=sssdfacts TEST_CONTAINER=el9 make test_container_no_lint`
- All 11 tests pass, including the two new ones.

I have force-pushed the updated commit.  Feel free to review when you have time. Thank you.

mhecko-rh (approved):
Code-wise OK. I have tested the patch with a swp file. Without this patch, leapp crashes with
```
...
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xbe in position 17: invalid start byte
```

With this patch, the upgrade proceeds without errors. Upgrade log (`leapp-upgrade.log`) contains the following:
```
2026-09-14 09:50:24.948 WARNING  PID: 88289 leapp.workflow.FactsCollection.sssd_facts: Skipping /etc/sssd/conf.d/.my.conf.swp because it does not appear to be a valid UTF-8 text file
```

LGTM.

PR #1595, opened 2026-09-09
Do not log ENOENT when removing leapp_resume.service

### LINKED ISSUE #1545: Error log message after a successful upgrade:Failed removing /etc/systemd/system/default.target.wants/leapp_resume.service
**Actual behavior**
After a successful Leapp upgrade and system reboot, the following error message appears in the /var/log/leapp/leapp-upgrade.log file: "Failed removing /etc/systemd/system/default.target.wants/leapp_resume.service".

This error occurs because the systemctl disable operation in the code already removes the /etc/systemd/system/default.target.wants/leapp_resume.service link. Consequently, the subsequent attempt to delete the same file triggers an error.

**To Reproduce**
Steps to reproduce the behavior

install 'leapp-upgrade'

setup 'leapp_upgrade_repositories.repo'

run 'leapp upgrade'

reboot

check /var/log/leapp/leapp-upgrade.log after reboot

**System information (please complete the following information):**

Amodak9801:
## Summary
- After a successful upgrade, `leapp-upgrade.log` could show `Failed removing .../leapp_resume.service` because `systemctl disable` already drops the wants symlink and the actor then `unlink`s the same path.
- Ignore `ENOENT` without logging it as a failure. Still remove the unit file, and also drop a leftover `default.target.wants` or `multi-user.target.wants` symlink if present. Other `OSError`s still raise.
- Replaced the destructive root-only test with non-destructive unit tests.

Fixes #1545

## Test plan
- [ ] `pytest repos/system_upgrade/common/actors/removeresumeservice/tests/test_removeresumeservice.py`
- [ ] Optional: after a real IPU reboot, confirm `/var/log/leapp/leapp-upgrade.log` no longer contains `Failed removing` for `leapp_resume.service`

Made with [Cursor](https://cursor.com)


Amodak9801:

PR #1597, opened 2026-09-15
Fix CVE-2026-75092

pirat89:
The leapp upgrade process is executed by root.
Unfortunately, it means also that all other
sub-processes are executed with root permissions.
In case of `mysqld --validate-config` command,
it can lead in some situations to a privilege
permissions when it is not executed by mysql user
or when the `--user=mysql` is not specified.

Adding the `--user=mysql` resolves this problem.

Resolves: RHEL-243195


pirat89:
@yuravk  headsup.

matejmatuska (approved):
LGTM

PR #1599, opened 2026-09-17
Bump release version to 0.25.1 in main

pirat89:
We have released 0.25.1 in the meanwhile, separated in the release/0.25 branch, so updating the spec and docs here as the main branch should be always the most up-to-date.


matejmatuska (approved):
LGTM

PR #1600, opened 2026-09-18
Add rpmtoflatpak actor to help with RPMs -> Flatpak migrations

tpopela-rh:
Starting with RHEL 10.2, we've switched the default delivery method for Firefox and Thunderbird from RPMs to Flatpaks (see RHEL-139533). To accommodate for that switch for upgrades, we check whether the firefox or thunderbird packages are preinstalled and if so, then we install the packages with flatpak-preinstall configuration files and run flatpak preinstall on it (the same workflow that Anaconda is doing).

We also limit the actor to work on subscribed systems, so we don't have to deal with credentials for the Red Hat Flatpak Registry (see RHEL-85004 for more information).

Developed by Claude Code.

Resolves: RHEL-53806
Resolves: RHEL-67925


tpopela-rh:
This is moved from https://github.com/oamg/leapp-repository/pull/1560 (moving from my personal GitHub account to my work one).

